PR #713, opened 2017-08-08
add ServerBuilder.grpcServiceUnder()

demongaorui:
One issue found in GrpcServiceBuilder.
in the comment of build() method, using serviceUnder() to bind grpc services to a prefix is recommended.
```
    /**
     * Constructs a new {@link GrpcService} that can be bound to
     * {@link com.linecorp.armeria.server.ServerBuilder}. As GRPC services themselves are mounted at a path that
     * corresponds to their protobuf package, you will almost always want to bind to a prefix, e.g. by using
     * {@link com.linecorp.armeria.server.ServerBuilder#serviceUnder(String, Service)}.
     */
```

While, when tried to using serviceUnder(), there happened "UNIMPLEMENTED, Method not found" exception.

After checking the related code, following approach seems to be the simplest solution.
```
    /**
     * Binds the specified {@link Service} under the specified directory.
     */
    public B serviceUnder(String pathPrefix, Service<HttpRequest, HttpResponse> service) {
        service(PathMapping.ofPrefix(pathPrefix, !(service instanceof GrpcService)), service);
        return self();
    }
```
But, this approach lead to circular dependency between core and grpc module. So added grpcServiceUnder() instead. Please help to reviw and share your opinions, thank you.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
@@ -207,6 +207,16 @@ abstract class AbstractVirtualHostBuilder<B extends AbstractVirtualHostBuilder> 
     }
 
     /**
+     * Binds the specified grpc {@link Service} under the specified directory.
+     */
+    public B grpcServiceUnder(
+            String pathPrefix,
+            Service<HttpRequest, HttpResponse> service) {
+        service(PathMapping.ofPrefix(pathPrefix, false), service);
+        return self();
+    }
+
+    /**
      * Binds the specified {@link Service} under the specified directory.
      */
     public B serviceUnder(String pathPrefix, Service<HttpRequest, HttpResponse> service) {
--- a/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
@@ -352,9 +352,26 @@ public final class ServerBuilder {
      * @throws IllegalStateException if the default {@link VirtualHost} has been set via
      *                               {@link #defaultVirtualHost(VirtualHost)} already
      */
-    public ServerBuilder serviceUnder(String pathPrefix, Service<HttpRequest, HttpResponse> service) {
+    public ServerBuilder serviceUnder(
+            String pathPrefix,
+            Service<HttpRequest, HttpResponse> service) {
         defaultVirtualHostBuilderUpdated();
         defaultVirtualHostBuilder.serviceUnder(pathPrefix, service);
+        return this;
+    }
+
+    /**
+     * Binds the specified grpc {@link Service} under the specified directory of the default
+     * {@link VirtualHost}.
+     *
+     * @throws IllegalStateException if the default {@link VirtualHost} has been set via
+     *                               {@link #defaultVirtualHost(VirtualHost)} already
+     */
+    public ServerBuilder grpcServiceUnder(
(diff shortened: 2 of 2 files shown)


anuraaga:
I've used serviceUnder without problem with armeria 0.51. Is this a different use case, or perhaps your version is outdated?

https://github.com/curioswitch/curiostack/blob/master/common/server/framework/src/main/java/org/curioswitch/common/server/framework/ServerModule.java#L340

demongaorui:
Oh, will try to upgrade from 0.5 to 0.51, if 0.51 works fine, then will close this PR.

demongaorui:
@anuraaga 
Somehow, with 0.51.0, server shut down right after start up.

Then, tried with 0.52.0, got the same error.

Init sb like:
```
String servePathPrefix =     '/' + MyServiceGrpc.SERVICE_NAME; // "/my.package.name.MyServiceGrpc"
sb.serviceUnder(
   servePathPrefix,
    new GrpcServiceBuilder()
           .addService(new MyService())
            .supportedSerializationFormats(GrpcSerializationFormats.values())
             .build())
```
Then when serve one request with path `/my.package.name.MyService/MyMethod`, PathMappingResult of the ctx [strip the prefix](https://github.com/line/armeria/blob/fbd4adc930828cf4db0a0dd12987d4ed7a24e0cd/core/src/main/java/com/linecorp/armeria/server/PrefixPathMapping.java#L57), then ctx.mappedPath() returns `MyMethod`, so method name becomes `MyMethod` when doing [GrpcRequestUtil.determineMethod(ctx)](https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java#L120), leading
[registry.lookupMethod(method)](https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java#L128) returns `null` because registry.methods using the full method(/my.package.name.MyService/MyMethod) as key in the map. So I think set stripPrefix as false could solve the problem.

[getGrpcPath()](https://github.com/curioswitch/curiostack/blob/master/common/server/framework/src/main/java/org/curioswitch/common/server/framework/config/ServerConfig.java#L66) in curiostack using default value "/api"(nothing related to grpc package and service name), this will not cause the stripPrefix issue, while how will you sending a grpc req to host:port/api? Cause grpc client(at lease object-c client) can only config host and port.

anuraaga:
Ah the example makes it clear. In that case, user's could just use `.service(PathMapping.ofPrefix(..., false), service)` manually - I don't think this is worth adding an API to the builder, especially one named `grpcFoo`.

`ServiceWithPathMapping` as discussed elsewhere could also take care of this for users.

As an aside, mounting on `/api` works with grpc-web and armeria-grpc client, but indeed for other clients it needs to be mounted at `/`.

demongaorui:
Oh~ You are right, `.service(PathMapping.ofPrefix(..., false), service)` is very handy. Thanks for sharing. 

Close this PR because problem solved. 

BTW, if I want to add some examples for general usage of armeria-grpc server, where should I check? Any doc contribution guide available?

trustin:
@demongaorui 

> BTW, if I want to add some examples for general usage of armeria-grpc server, where should I check? Any doc contribution guide available?

For now, all the content in our web site is generated by Sphinx. You will find the `.rst` files under the `site/` directory in this repository. `./gradlew site` will generate a site into `site/build/site/` so you can browse it from your browser. You could update the `server-grpc.rst` file to add more usages and tips. If you send a PR that improves our documentation, that would be very awesome! :-D

PR #717, opened 2017-08-09
Switch to micrometer-metrics

trustin:
Motivation:

There are two major metric collection libraries in Java: Dropwizard and
Prometheus. They both have their own advantages and disadvantages so we
cannot simply chose one over the other.

So far, we provided a dedicated decorator implementation for each
library, but this approach is not scalable. Also, it gives us hard time
for adding session-level metrics because the core API has to remain
detached from both libraries' API.

Modifications:

- Use micrometer-metrics as our standard metric collection API.
- Replace (Dropwizard|Prometheus)MetricCollecting(Client|Service) with
  MetricCollecting(Client|Service)
- Replace DropwizardMetricsCircuitBreakerListener with
  MetricCollectingCircuitBreakerListener
- Add Server.meterRegistry() and ClientFactory.meterRegistry()
- Add RequestContext.meterRegistry()
- Add CaffeineMetricSupport and migrate RouterCache to the new API

Result:

- We can support third-party metric collection libraries easily.
- We can add metrics everywhere easily.


trustin:
Please do not review Spring integration yet; it's not done yet. Other parts are ready for review.

imasahiro:
nit.  `There are two major matric collection libraries in Java: Dropwizard and`
`matric` -> `metric`.

How about extracting miscellaneous changes (license, gradle script update) and making another PR for that to focus on micrometer change?

trustin:
@imasahiro:
> How about extracting miscellaneous changes (license, gradle script update) and making another PR for that to focus on micrometer change?

Fixed

trustin:
@jmostella @imasahiro @tokuhirom Finished migrating Spring support. PTAL

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
```diff
@@ -0,0 +1,143 @@
...
+            if (!force) {
+                if (currentTimeNanos - lastStatsUpdateTime < UPDATE_INTERVAL_NANOS) {
+                    return;
+                }
+            }
+
+            cacheStats = cache.stats();
+            estimatedSize = cache.estimatedSize();
+
+            // Write the volatile field lastly so that cacheStats and estimatedSize are visible
```
anuraaga: last

core/src/main/java/com/linecorp/armeria/internal/metric/MicrometerUtil.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * A utility that prevents double instantiation of an object for a certain {@link MeterId}. This can be useful
+ * when you need to make sure you do not register the same {@link Meter} more than once.
+ *
+ * @see RequestMetricSupport
+ * @see CaffeineMetricSupport
+ */
+public final class MicrometerUtil {
+
+    private static final ConcurrentMap<MeterRegistry, BiMap<MeterId, Object>> map = new MapMaker().weakKeys()
```
anuraaga: Do you have a use case for using `BiMap` instead of `Map`?

core/src/main/java/com/linecorp/armeria/internal/metric/MicrometerUtil.java
```diff
@@ -0,0 +1,84 @@
...
+                                 BiFunction<MeterRegistry, MeterId, T> factory) {
+        requireNonNull(registry, "registry");
+        requireNonNull(id, "id");
+        requireNonNull(type, "type");
+        requireNonNull(factory, "factory");
+
+        final BiMap<MeterId, Object> objects = map.computeIfAbsent(registry, unused -> HashBiMap.create());
+        final Object object = objects.computeIfAbsent(id, i -> {
+            final Object newObject = factory.apply(registry, i);
+            checkState(type.isInstance(newObject), "factory returned an incompatible object: %s", newObject);
```
anuraaga: I believe you can use `final T newObject` and remove this type check.

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -0,0 +1,266 @@
...
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.util.MeterId;
+import io.netty.util.AttributeKey;
+
+/**
+ * (Internal use only) Collects the metric data and stores it into the {@link MeterRegistry}.
```
anuraaga: Do we need `Internal use only` even though this is in `internal` package?

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -192,35 +184,74 @@ public PathMapping pathStringMapping() {
...
-                                       List<MediaType> consumeTypes, List<MediaType> produceTypes) {
-        final StringJoiner methodNames = new StringJoiner("_");
-        supportedMethods.forEach(e -> methodNames.add(e.name()));
-        StringJoiner name = new StringJoiner(delim).add(prefix)
-                                                   .add(methodNames.toString());
+    private static String generateLoggerName(String prefix, Set<HttpMethod> supportedMethods,
+                                             List<MediaType> consumeTypes, List<MediaType> produceTypes) {
+        final StringJoiner name = new StringJoiner(".");
+        name.add(prefix);
+        name.add(Joiner.on('_').join(supportedMethods.stream().sorted().iterator()));
```
anuraaga: Can extract `Joiner.on` to a constant.

anuraaga: Or can use `stream().sorted().collect(Collectors.joining("_"))`

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -192,35 +184,74 @@ public PathMapping pathStringMapping() {
...
         }
         return name.toString();
     }
+
+    private static List<String> generateMetricName(List<String> prefix, Set<HttpMethod> supportedMethods,
+                                                   List<MediaType> consumeTypes, List<MediaType> produceTypes) {
+
+        final ImmutableList.Builder<String> builder = ImmutableList.builder();
+        builder.addAll(prefix);
+        builder.add(Joiner.on(',').join(supportedMethods.stream().sorted().iterator()));
```
anuraaga: Ditto (can extract method to join a set of HttpMethod)

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -609,11 +622,14 @@ public Server build() {
             virtualHosts = this.virtualHosts;
         }
 
+        final MeterRegistry meterRegistry = this.meterRegistry != null ? this.meterRegistry
```
anuraaga: firstNonNull or Optional

thrift/src/test/java/com/linecorp/armeria/it/metric/PrometheusMetricsIntegrationTest.java
```diff
@@ -16,76 +16,92 @@
...
-                       new PrometheusExporterHttpService(registry));
+                       new PrometheusExporterHttpService(prometheusRegistry));
         }
     };
 
+    private static MeterRegistry oldRegistry;
+
+    @BeforeClass
+    public static void setDefaultMeterRegistry() {
+        oldRegistry = ClientFactory.DEFAULT.meterRegistry();
```
anuraaga: It's possible in the future overwriting static fields in tests can be problematic if for example running tests in parallel.  If possible, I guess we shouldn't use the DEFAULT clientfactory in these tests

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
trustin: Fixed by extracting `Joiner`

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
trustin: Fixed by extracting `Joiner`

core/src/main/java/com/linecorp/armeria/internal/metric/MicrometerUtil.java
trustin: Oops! :-)

core/src/main/java/com/linecorp/armeria/internal/metric/MicrometerUtil.java
trustin: I was trying to do something with it but it wasn't needed after all.

core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java
```diff
@@ -0,0 +1,274 @@
...
+            return PROMETHEUS_SANITIZE_BODY_PATTERN.matcher(
+                    PROMETHEUS_SANITIZE_PREFIX_PATTERN.matcher(buf.toString())
+                                                      .replaceFirst("_")).replaceAll("_");
+        }
+
+        final StringBuilder buf = new StringBuilder();
+        for (String n : name) {
+            buf.append(n).append('.');
+        }
+        buf.setLength(buf.length() - 1);
```
trustin: Yes. Simplified using `String.join()`

core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java
```diff
@@ -0,0 +1,274 @@
...
+     */
+    public static String name(MeterRegistry registry, Iterable<String> name) {
+        return name(registry, MeterUnit.NONE, name);
+    }
+
+    /**
+     * Returns a newly-created {@link Meter} name by concatenating the name of the specified {@link MeterId}
+     * and the specified name parts.
+     */
+    public static String name(MeterRegistry registry, MeterUnit unit, MeterId id, String... name) {
```
trustin: Could you elaborate a little bit?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -224,20 +227,30 @@ public ClientFactoryBuilder connectionPoolListener(
...
+    public ClientFactoryBuilder meterRegistry(MeterRegistry meterRegistry) {
+        this.meterRegistry = requireNonNull(meterRegistry, "meterRegistry");
+        return this;
+    }
+
+    /**
      * Returns a newly-created {@link ClientFactory} based on the properties of this builder.
      */
     public ClientFactory build() {
+        final MeterRegistry meterRegistry = this.meterRegistry != null ? this.meterRegistry
```
trustin: What's the benefit of doing so? `firstNonNull()` will pre-instantiate the fallback registry instance unnecessarily and using `Optional` just defers the instantiation to `ServerConfig`.

thrift/src/test/java/com/linecorp/armeria/it/metric/PrometheusMetricsIntegrationTest.java
trustin: Great point. Fixed.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
```diff
@@ -0,0 +1,49 @@
...
+        this.id = requireNonNull(id, "id");
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        final String endpoints = name(registry, MeterUnit.NONE, id, "endpoints");
+        registry.gauge(endpoints, tags(id, "state", "healthy"), endpointGroup,
+                       unused -> endpointGroup.endpoints().size());
+        registry.gauge(endpoints, tags(id, "state", "unhealthy"), endpointGroup,
+                       unused -> endpointGroup.allServers.size() - endpointGroup.endpoints().size());
```
trustin: Implemented. :-)

trustin:
@jmostella @imasahiro Now a user can configure metric name generation by injecting `MeterIdFunctionFactory`.

core/src/test/java/com/linecorp/armeria/common/metric/MeterIdFunctionTest.java
```diff
@@ -0,0 +1,54 @@
...
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.prometheus.PrometheusMeterRegistry;
+import io.micrometer.core.instrument.util.MeterId;
+
+public class MeterIdFunctionTest {
+
+    @Test
+    public void testWithTags() {
+        final MeterIdFunction f = (registry, log) -> new MeterId("a", Tags.zip("b", "c"));
+        assertThat(f.withTags(Tags.zip("d", "e", "f", "g")).apply(null, null))
```
jmostella: To be instructive for implementors, maybe we should use meaningful labels.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
anuraaga: Ah good point on `firstNonNull`. For `Optional`, I was just thinking this would become `meterRegistry.orElseGet(SimpleMeterRegistry::new)`. My feeling on `Optional`is we prefer it to nullable types when possible, but their lower performance makes them impractical in some cases. But either way is fine.

core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java
anuraaga: Just meant `name(registry, unit, id, ImmutableList.copyOf(name)`. It avoids the duplicating of the logic `ImmutableList.<String>builder()...` at the expense of an extra copy.

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdFunction.java
```diff
@@ -0,0 +1,124 @@
...
+     * Returns a {@link MeterIdFunction} that returns a newly created {@link MeterId} which has
+     * the specified label added.
+     */
+    default MeterIdFunction withTags(String... keyValues) {
+        requireNonNull(keyValues, "keyValues");
+        return withTags(Tags.zip(keyValues));
+    }
+
+    /**
+     * Returns a {@link MeterIdFunction} that returns a newly created {@link MeterId    } which has
```
minwoox: nit: trailing spaces after MeterId

core/src/main/java/com/linecorp/armeria/client/metric/MetricCollectingClient.java
```diff
@@ -0,0 +1,76 @@
...
+        requireNonNull(meterIdFunction, "meterIdFunction");
+        return delegate -> new MetricCollectingClient<>(delegate, meterIdFunction);
+    }
+
+    private final MeterIdFunction meterIdFunction;
+
+    @SuppressWarnings("unchecked")
+    MetricCollectingClient(Client<I, O> delegate, MeterIdFunction meterIdFunction) {
+        super(delegate);
+        this.meterIdFunction = meterIdFunction;
```
minwoox: nit: requeireNonNull since this constructor is not private?

core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java
```diff
@@ -0,0 +1,269 @@
...
+        return Iterables.concat(tags, Tags.zip(keyValues));
+    }
+
+    /**
+     * Returns the value of the {@link Measurement} collected from the matching {@link Meter}.
+     *
+     * @throws NoSuchElementException if there's no such {@link Meter} in the specified {@link MeterRegistry}
+     */
+    public static double measure(MeterRegistry registry, String name, String... keyValues) {
+        requireNonNull(name, "name");
```
minwoox: Seems null check for the registry is missed for all of the methods in this class here

core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java
```diff
@@ -0,0 +1,269 @@
...
+            buf.setCharAt(buf.length() - 1, '}');
+        }
+
+        return buf.toString();
+    }
+
+    /**
+     * Returns whether the specified {@link MeterRegistry} implementation prefers base unit for values.
+     */
+    public static boolean prefersBaseUnit(MeterRegistry registry) {
```
minwoox: prometheusMeterRegistry? I have just imagined that adding other registry whose base is different from PrometheusMeterRegistry

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
trustin: I'm fine with using `Optional` but it seems like it's not recommended to use `Optional` as a parameter or a member field according to IntelliJ's inspection. Let me leave this as it is for now.

core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java
trustin: Ah, yes. Fixed. :-)

core/src/test/java/com/linecorp/armeria/common/metric/MeterIdFunctionTest.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java
trustin: 'base unit' means 'seconds' or 'bytes' without multipliers (nano, kilo, etc), so I think it's fine as it is.

core/src/main/java/com/linecorp/armeria/client/metric/MetricCollectingClient.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdFunction.java
trustin: Fixed. :-)

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdFunction.java
```diff
@@ -0,0 +1,124 @@
...
+    static MeterIdFunction ofDefault(String firstNamePart, String... otherNameParts) {
+        requireNonNull(firstNamePart, "firstNamePart");
+        requireNonNull(otherNameParts, "otherNameParts");
+
+        final List<String> name =
+                ImmutableList.<String>builder().add(firstNamePart).add(otherNameParts).build();
+
+        return (registry, log) -> {
+            final RequestContext ctx = log.context();
+            final Object requestEnvelope = log.requestHeaders();
```
trustin: Just a cruft from the old days when `log.requestHeaders()` was `log.requestEnvelope()`. Fixed.

hyangtack: I see! :-)

core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java
minwoox: Okay I misunderstood. Sorry about that~ :-)

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
```diff
@@ -16,101 +16,115 @@
...
-                return result;
+    private static final class CachingRouter<V> implements Router<V> {
+
+        private final Router<V> delegate;
+        private final Cache<PathMappingContext, V> cache;
+        private final Function<V, PathMapping> pathMappingResolver;
+
+        CachingRouter(Router<V> delegate, Cache<PathMappingContext, V> cache,
+                      Function<V, PathMapping> pathMappingResolver) {
+            this.delegate = delegate;
```
hyangtack: nit: `requireNonNull` for consistency? (Actually I think it's ok without that.)

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -192,35 +186,74 @@ public PathMapping pathStringMapping() {
...
+        final ImmutableList.Builder<String> builder = ImmutableList.builder();
+        builder.addAll(prefix);
+        builder.add(metricNameJoiner.join(supportedMethods.stream().sorted().iterator()));
+
+        // The following three cases should be different to each other.
+        // Each name would be produced as follows:
+        //
+        // consumeTypes: text/plain, text/html               -> "consumes:text_plain,text_html"
+        // consumeTypes: text/plain, produceTypes: text/html -> "consumes:text_plain", "produces:text_html"
+        // produceTypes: text/plain, text/html               -> "produces:text_plain,text_html"
```
hyangtack: This comment should be fixed. (`_` -> `/`)
```
        // consumeTypes: text/plain, text/html               -> "consumes:text/plain,text/html"
        // consumeTypes: text/plain, produceTypes: text/html -> "consumes:text/plain", "produces:text/html"
        // produceTypes: text/plain, text/html               -> "produces:text/plain,text/html"
```

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
trustin: Fixed

spring-boot/autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaMeterBindersTest.java
```diff
@@ -66,14 +80,26 @@ protected void doGet(ServiceRequestContext ctx, HttpRequest req,
...
-        assertThat(metricRegistry.getMetrics().get("jvm.buffers.direct.count")).isNotNull();
-        assertThat(metricRegistry.getMetrics().get("jvm.classloader.loaded")).isNotNull();
-        assertThat(metricRegistry.getMetrics().get("jvm.memory.heap.max")).isNotNull();
-        assertThat(metricRegistry.getMetrics().get("jvm.threads.daemon.count")).isNotNull();
-        assertThat(metricRegistry.getMetrics().get("system.cpu.load-average")).isNotNull();
+        registry.getMeters().forEach(m -> {
+            final String id = m.getName() +
+                              Streams.stream(m.getTags()).collect(toImmutableMap(Tag::getKey, Tag::getValue));
+            final Map<String, Double> values = measureAll(registry, m.getName(), m.getTags());
+            if (values.size() > 1) {
```
imasahiro: Can simplify this if-else block `logger.debug("{}={}", id, values);`?

trustin: Yes we can, but it's not pretty when it's just a single-value meter. `foo=[ bar ]` vs. `foo=bar`

imasahiro: I see. I'm worried about `values.size() == 0` case. It throws NoSuchElementException.
But, it is test code, so it is not big deal.

trustin: A `Meter.measure()` returning an empty collection would be a bug, so I think it's fine.

imasahiro: okay

trustin:
Thanks, everyone for reviewing.

PR #719, opened 2017-08-09
Clean up the build script

trustin:
- Sort the dependencies of core/build.gradle
- Fix the incorrect comment related with ProGuarding Caffeine


trustin:
Pretty much cosmetic change only. Will self-merge.

PR #720, opened 2017-08-10
Modify RetryStrategy to return backoff according to the response

### LINKED ISSUE #655: Provide a way to choose different backoff policy according to the response
A RetryingClient may want to have different backoff policy according to the response.
For example, a client may want to retry using exponential backoff when it gets `SERVICE_UNAVAILABLE`.
On the other hand, it may want to keep retrying without delay when it updates some field in a server, and gets revision conflicts.

minwoox:
Motivation:

A `RetryingClient` may want to have different `Backoff` by the responses.
For example, the client wants to retry using exponential backoff when it gets `SERVICE_UNAVAILABLE`.
On the other hand, it wants to keep retrying without delay when it gets revision conflicts.

Modifications:

- Make all of the implementations of `Backoff` stateless
- Change the `shouldRetry` method returng `Optional<Backoff>`

Result:

- Close #655
- Users can choose `Backoff` accroding to the response

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/ExponentialBackoff.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/ExponentialBackoff.java
@@ -20,7 +20,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import com.google.common.base.MoreObjects;
 
 final class ExponentialBackoff extends AbstractBackoff {
-    private long currentDelayMillis;
+    private final long initialDelayMillis;
     private final long maxDelayMillis;
     private final double multiplier;
 
@@ -30,19 +30,19 @@ final class ExponentialBackoff extends AbstractBackoff {
                       "initialDelayMillis: %s (expected: >= 0)", initialDelayMillis);
         checkArgument(initialDelayMillis <= maxDelayMillis, "maxDelayMillis: %s (expected: >= %s)",
                       maxDelayMillis, initialDelayMillis);
-        currentDelayMillis = initialDelayMillis;
+        this.initialDelayMillis = initialDelayMillis;
         this.maxDelayMillis = maxDelayMillis;
         this.multiplier = multiplier;
     }
 
     @Override
     protected long doNextDelayMillis(int numAttemptsSoFar) {
-        if (currentDelayMillis >= maxDelayMillis) {
-            return maxDelayMillis;
+        if (numAttemptsSoFar == 1) {
+            return initialDelayMillis;
         }
-        long nextDelay = currentDelayMillis;
-        currentDelayMillis = saturatedMultiply(currentDelayMillis, multiplier);
-        return nextDelay;
+        final long nextDelay =
+                saturatedMultiply(initialDelayMillis, Math.pow(multiplier, numAttemptsSoFar - 1));
+        return Math.min(nextDelay, maxDelayMillis);
     }
 
     private static long saturatedMultiply(long left, double right) {
@@ -53,7 +53,7 @@ final class ExponentialBackoff extends AbstractBackoff {
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this)
-                          .add("currentDelayMillis", currentDelayMillis)
+                          .add("initialDelayMillis", initialDelayMillis)
                           .add("maxDelayMillis", maxDelayMillis)
                           .add("multiplier", multiplier)
                           .toString();
--- a/core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
@@ -16,19 +16,16 @@
 
 package com.linecorp.armeria.client.retry;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
-import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-
-import com.google.common.collect.ImmutableList;
+import java.util.function.Function;
 
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
-import com.linecorp.armeria.common.HttpStatusClass;
 import com.linecorp.armeria.internal.HttpHeaderSubscriber;
 
 /**
@@ -37,28 +34,18 @@ import com.linecorp.armeria.internal.HttpHeaderSubscriber;
  */
 final class HttpStatusBasedRetryStrategy implements RetryStrategy<HttpRequest, HttpResponse> {
 
-    private final List<HttpStatus> retryStatuses;
+    private final Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction;
 
     /**
      * Creates a new instance.
      */
-    HttpStatusBasedRetryStrategy(Iterable<HttpStatus> retryStatuses) {
-        this.retryStatuses = validateStatus(ImmutableList.copyOf(retryStatuses));
-    }
-
-    private List<HttpStatus> validateStatus(List<HttpStatus> retryStatuses) {
-        requireNonNull(retryStatuses, "retryStatuses");
-        checkArgument(!retryStatuses.isEmpty(), "Need at least a status to retry");
-
-        for (HttpStatus retryStatus : retryStatuses) {
-            checkArgument(retryStatus.codeClass() != HttpStatusClass.INFORMATIONAL,
-                          "retryStatuses contains an informational status: %s", retryStatus);
-        }
-        return retryStatuses;
+    HttpStatusBasedRetryStrategy(Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction) {
+        this.statusBasedBackoffFunction = requireNonNull(
+                statusBasedBackoffFunction, "statusBasedBackoffFunction");
     }
 
     @Override
-    public CompletableFuture<Boolean> shouldRetry(HttpRequest request, HttpResponse response) {
+    public CompletableFuture<Optional<Backoff>> shouldRetry(HttpRequest request, HttpResponse response) {
         final CompletableFuture<HttpHeaders> future = new CompletableFuture<>();
         final HttpHeaderSubscriber subscriber = new HttpHeaderSubscriber(future);
         response.closeFuture().whenComplete(subscriber);
@@ -68,11 +55,10 @@ final class HttpStatusBasedRetryStrategy implements RetryStrategy<HttpRequest, H
             if (headers != null) {
                 final HttpStatus resStatus = headers.status();
                 if (resStatus != null) {
-                    return retryStatuses.stream().anyMatch(
-                            retryStatus -> resStatus.code() == retryStatus.code());
+                    return statusBasedBackoffFunction.apply(resStatus);
                 }
             }
-            return false;
+            return Optional.empty();
         });
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
@@ -17,12 +17,15 @@ package com.linecorp.armeria.client.retry;
 
 import static java.util.Objects.requireNonNull;
 
-import java.util.Arrays;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
 
+import com.linecorp.armeria.common.Flags;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.HttpStatusClass;
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Response;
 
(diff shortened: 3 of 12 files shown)


core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -34,19 +36,21 @@
...
+
     /**
      * A {@link RetryStrategy} that defines a retry should not be performed.
      */
     static <I extends Request, O extends Response> RetryStrategy<I, O> never() {
         return new RetryStrategy<I, O>() {
             @Override
-            public CompletableFuture<Boolean> shouldRetry(I request, O response) {
-                return CompletableFuture.completedFuture(false);
+            public CompletableFuture<Optional<Backoff>> shouldRetry(I request, O response) {
```
minwoox: @imasahiro Sorry, I have to change the api again in order to meet the needs in #655

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -66,21 +70,22 @@ public boolean shouldRetry(I request, Throwable throwable) {
...
+     * <p>If an {@link Exception} occurs while processing {@link Request} and {@link Response}, this method
      * should not complete the {@link CompletableFuture} with the {@link Exception}. The {@link Exception}
-     * needs to be dealt in the {@link RetryStrategy#shouldRetry(Request, Throwable)} method, and
-     * the {@link CompletableFuture} has to be completed with {@code false}.
-     * </p>
+     * needs to be dealt in the {@link RetryStrategy#shouldRetry(Request, Throwable)} method, so
+     * the {@link CompletableFuture} has to be completed with {@link Optional#EMPTY}.
      */
-    CompletableFuture<Boolean> shouldRetry(I request, O response);
+    CompletableFuture<Optional<Backoff>> shouldRetry(I request, O response);
```
minwoox: @imasahiro Sorry, I have to change the api again in order to meet the needs in #655

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -66,21 +70,22 @@ public boolean shouldRetry(I request, Throwable throwable) {
...
+     * <p>If an {@link Exception} occurs while processing {@link Request} and {@link Response}, this method
      * should not complete the {@link CompletableFuture} with the {@link Exception}. The {@link Exception}
-     * needs to be dealt in the {@link RetryStrategy#shouldRetry(Request, Throwable)} method, and
-     * the {@link CompletableFuture} has to be completed with {@code false}.
-     * </p>
+     * needs to be dealt in the {@link RetryStrategy#shouldRetry(Request, Throwable)} method, therefore
+     * the {@link CompletableFuture} has to be completed with {@link Optional#EMPTY}.
      */
-    CompletableFuture<Boolean> shouldRetry(I request, O response);
+    CompletableFuture<Optional<Backoff>> shouldRetry(I request, O response);
```
minwoox: @imasahiro Sorry, I have to change the api again in order to meet the needs in #655

core/src/main/java/com/linecorp/armeria/client/retry/ExponentialBackoff.java
```diff
@@ -30,19 +30,16 @@
...
     protected long doNextDelayMillis(int numAttemptsSoFar) {
-        if (currentDelayMillis >= maxDelayMillis) {
-            return maxDelayMillis;
-        }
-        long nextDelay = currentDelayMillis;
-        currentDelayMillis = saturatedMultiply(currentDelayMillis, multiplier);
-        return nextDelay;
+        final long nextDelay =
+                saturatedMultiply(initialDelayMillis, Math.pow(multiplier, numAttemptsSoFar - 1));
+        return nextDelay >= maxDelayMillis ? maxDelayMillis : nextDelay;
```
hyangtack: nit: How about using `Math.min(nextDelay, maxDelayMillis)` to easily understand?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -185,33 +180,30 @@ private static HttpHeaders getHttpHeaders(HttpResponse res) {
...
-                        nextDelay, TimeUnit.MILLISECONDS);
-            }
+        } catch (ResponseTimeoutException e) {
+            deferredRes.close(e);
+            rootReqDuplicator.close();
+            return;
+        }
+
+        final EventLoop eventLoop = ctx.contextAwareEventLoop();
+        if (nextDelay <= 0) {
```
hyangtack: In this case, `nextDelay` is always `0`, right? If so, how about rearranging the codes like below for readability?
```
final long nextDelay = getNextDelay(...);
if (nextDelay == 0) {
  eventLoop.submit(...);
  return;
} 
if (nextDelay > 0) {
  eventLoop.schedule(...);
  return;
}
try {
...
} catch (ResponseTimeoutException e) {
...
}
```

core/src/main/java/com/linecorp/armeria/client/retry/ExponentialBackoff.java
minwoox: Fixed. Thank you~ :-)

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
minwoox: `getNextDelay()` will throw a `ResponseTimeoutException` when the elapsed time is greater than `responseTimeout`. It returns `-1` when the number of attempts exceeds the `maxAttempts`. So try catch needs to wrap the getNextDelay().  :-)

hyangtack: OK! Sorry, I missed reading `responseTimeoutMillis` method.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -75,20 +74,20 @@ private static void setDeadlineOfThisRequest(ClientRequestContext ctx) {
...
-    Backoff newBackoff() {
-        Backoff backoff = backoffSupplier.get();
-        if (!backoff.as(AttemptLimitingBackoff.class).isPresent()) {
-            backoff = backoff.withMaxAttempts(defaultMaxAttempts);
+    private void initRetryStrategyState(ClientRequestContext ctx) {
+        final RetryStrategyState retryStrategyState = ctx.attr(LAST_RETRY_STRATEGY_STATE).get();
+        if (retryStrategyState != null) {
+            retryStrategyState.init();
+        } else {
+            ctx.attr(LAST_RETRY_STRATEGY_STATE).set(new RetryStrategyState(defaultMaxAttempts));
```
trustin: `ctx.attr(...)` creates a new `Attribute` instance when no attribute is set. We could extract it into a variable:

```java
final Attribute<RetryStrategyState> attr = ctx.attr(...);
final RetryStrategyState retryStrategyState = attr.get();
if (...) {
    ...
} else {
    attr.set(new RetryStrategyState(...));
}
```

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -38,11 +39,20 @@
 final class HttpStatusBasedRetryStrategy implements RetryStrategy<HttpRequest, HttpResponse> {
 
     private final List<HttpStatus> retryStatuses;
+    private final Backoff backoff;
```
trustin: Could we allow a user specify a different backoff for a different HTTP status? i.e. `Map<HttpStatus, Backoff>`?

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -66,21 +70,39 @@ public boolean shouldRetry(I request, Throwable throwable) {
...
     /**
-     * Returns {@link CompletableFuture} that contains a {@link Boolean} flag which indicates whether
-     * {@link RetryingClient} needs to retry or not.
-     * <p>
-     * If an {@link Exception} occurs while processing {@link Request} and {@link Response}, this method
+     * Returns the {@link RetryStrategy} that decides to retry the request using HTTP statuses with the
+     * specified {@link Backoff}.
+     */
+    static RetryStrategy<HttpRequest, HttpResponse> onStatus(Backoff backoff, HttpStatus... retryStatuses) {
+        return onStatus(backoff, Arrays.asList(requireNonNull(retryStatuses, "retryStatuses")));
```
trustin: Could we also add `RetryStrategy.onStatus(Map<HttpStatus, Backoff>)`?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
minwoox: Oops Thank you~~ fixed. :-)

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
minwoox: Added. Thank you~

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
minwoox: Added. Thank you~

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -37,28 +39,58 @@
...
+     * Creates a new instance.
+     */
+    HttpStatusBasedRetryStrategy(Backoff backoff, Iterable<HttpStatus> retryStatuses) {
+        this(statusWithBackoff(requireNonNull(backoff, "backoff"),
+                               requireNonNull(retryStatuses, "retryStatuses")));
+    }
+
+    private static Map<HttpStatus, Backoff> statusWithBackoff(Backoff backoff,
+                                                              Iterable<HttpStatus> retryStatuses) {
+        checkArgument(retryStatuses.iterator().hasNext(), "need at least a status to retry");
```
trustin: `!Iterables.isEmpty()`

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -37,28 +39,58 @@
  */
 final class HttpStatusBasedRetryStrategy implements RetryStrategy<HttpRequest, HttpResponse> {
 
-    private final List<HttpStatus> retryStatuses;
+    private Map<HttpStatus, Backoff> retryStatusesWithBackoff;
```
trustin: Could be `final`?

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -37,28 +39,58 @@
...
+
+    /**
+     * Creates a new instance.
+     */
+    HttpStatusBasedRetryStrategy(Map<HttpStatus, Backoff> retryStatusesWithBackoff) {
+        validateRetryStatusesWithBackoff(retryStatusesWithBackoff);
+        this.retryStatusesWithBackoff =
+                retryStatusesWithBackoff instanceof ImmutableMap ? retryStatusesWithBackoff
+                                                                 : ImmutableMap.copyOf(
+                                                                         retryStatusesWithBackoff);
```
trustin: What do you think about doing `ImmutableMap.copyOf()` before calling `validateRetryStatusesWithBackoff()` so that there's no chance for a user modifying the map during validation? (thread safety)

Please also note that `ImmutableMap.copyOf()` already handles the case where the specified map is an `ImmutableMap`.

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -37,28 +39,58 @@
...
+        requireNonNull(retryStatusesWithBackoff, "retryStatusesWithBackoff");
+        checkArgument(!retryStatusesWithBackoff.isEmpty(), "need at least a status to retry");
 
-        for (HttpStatus retryStatus : retryStatuses) {
+        for (HttpStatus retryStatus : retryStatusesWithBackoff.keySet()) {
             checkArgument(retryStatus.codeClass() != HttpStatusClass.INFORMATIONAL,
-                          "retryStatuses contains an informational status: %s", retryStatus);
+                          "contains an informational status: %s", retryStatus);
+            checkArgument(retryStatusesWithBackoff.get(retryStatus) != null,
+                          "backoff is null for the retryStatus: %s", retryStatusesWithBackoff.get(retryStatus));
```
trustin: If we do `ImmutableMap.copyOf()` before calling `validateRetryStatusesWithBackOff()`, we don't need to do null checks on the keys and values.

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -68,11 +100,13 @@
             if (headers != null) {
                 final HttpStatus resStatus = headers.status();
                 if (resStatus != null) {
-                    return retryStatuses.stream().anyMatch(
-                            retryStatus -> resStatus.code() == retryStatus.code());
+                    final Backoff backoff = retryStatusesWithBackoff.get(resStatus);
+                    if (backoff != null) {
+                        return Optional.of(backoff);
```
trustin: Could be: `return Optional.ofNullable(retryStatusesWithBackoff.get(resStatus));`

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -66,21 +71,48 @@ public boolean shouldRetry(I request, Throwable throwable) {
...
```
trustin: `Arrays.asList` -> `ImmutableList.copyOf()` ?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -44,25 +41,27 @@
     private static final AttributeKey<Long> RESPONSE_TIMEOUT_DEADLINE_NANOS =
             AttributeKey.valueOf("RESPONSE_TIMEOUT_DEADLINE_NANOS");
 
-    private final Supplier<? extends Backoff> backoffSupplier;
+    private static final AttributeKey<RetryStrategyState> LAST_RETRY_STRATEGY_STATE =
+            AttributeKey.valueOf("LAST_RETRY_STRATEGY_STATE");
```
trustin: - `AttributeKey.valueOf(RetryingClient.class, "LAST_RETRY_STRATEGY_STATE");`
- `AttributeKey.valueOf(RetryingClient.class, "RESPONSE_TIMEOUT_DEADLINE_NANOS");`

We may want to merge these two attributes into one such as `RetryingClientState`.

minwoox: That's a good point. They're merged. Thank you~ :-)

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -34,53 +37,90 @@
...
+     */
+    static RetryStrategy<HttpRequest, HttpResponse> onStatus(HttpStatus retryStatus) {
+        return onStatus(retryStatus, defaultBackoff);
+    }
+
+    /**
+     * Returns the {@link RetryStrategy} that decides to retry the request using HTTP statuses with the
+     * specified {@link Backoff}.
+     */
+    static RetryStrategy<HttpRequest, HttpResponse> onStatus(HttpStatus retryStatus, Backoff backoff) {
```
trustin: Would this method be useful? Just curious.

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -34,53 +37,90 @@
...
+    /**
+     * Returns the {@link RetryStrategy} that decides to retry the request using
+     * {@code statusBasedBackoffFunction}.
+     *
+     * @param statusBasedBackoffFunction the {@link Function} that returns {@code Optional<Backoff>} according
+     *                                   to the {@link HttpStatus}
+     */
+    static RetryStrategy<HttpRequest, HttpResponse> onStatus(
+            Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction) {
+        return new HttpStatusBasedRetryStrategy(statusBasedBackoffFunction);
```
trustin: requireNonNull

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -34,53 +37,90 @@
...
 
     /**
-     * Returns the {@link RetryStrategy} that decides to retry the request using HTTP statuses.
+     * Returns the {@link RetryStrategy} that retries the request with the specified {@code backoff}
+     * when the response matches {@link HttpStatusClass#SERVER_ERROR}.
      */
-    static RetryStrategy<HttpRequest, HttpResponse> onStatus(Iterable<HttpStatus> retryStatuses) {
-        return new HttpStatusBasedRetryStrategy(retryStatuses);
+    static RetryStrategy<HttpRequest, HttpResponse> onServerErrorStatus(Backoff backoff) {
+        requireNonNull(backoff);
```
trustin: `, "backoff"`

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -41,54 +38,43 @@
 public abstract class RetryingClient<I extends Request, O extends Response>
         extends SimpleDecoratingClient<I, O> {
 
-    private static final AttributeKey<Long> RESPONSE_TIMEOUT_DEADLINE_NANOS =
-            AttributeKey.valueOf("RESPONSE_TIMEOUT_DEADLINE_NANOS");
+    private static final AttributeKey<RetryingClientState> RETRYING_CLIENT_STATE =
+            AttributeKey.valueOf(RetryingClient.class, "RETRYING_CLIENT_STATE");
```
trustin: `"STATE"` should be enough because `valueOf(Class, String)` will use the FQCN as a prefix.

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
minwoox: This method is maybe, just too much. I will remove this and above one. Thank you~!

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
minwoox: check null in `HttpStatusBasedRetryStrategy` class with the same parameter name.

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
minwoox: Oops~ fixed. :-)

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
minwoox: Oops~ Thank you for letting me know~!

trustin:
Thanks, @minwoox !

PR #721, opened 2017-08-10
Implement Server#toString

imasahiro:


trustin:
@imasahiro Would you mind moving `toString()` below `numConnections()`?

trustin:
Thanks, @imasahiro !

PR #722, opened 2017-08-14
Propagate missing content-length header if Tomcat coyote response co…

imasahiro:
…ntains content-length header


tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -495,6 +495,10 @@ private static HttpHeaders convertResponse(Response coyoteRes) {
             headers.set(HttpHeaderNames.CONTENT_TYPE, contentType);
         }
 
+        if (coyoteRes.getContentLength() >= 0) {
```
kishida: If we can set the value as long, getContentLengthLong may be better.

PR #725, opened 2017-08-17
Correct typo

AmosDoan:
- Correct some typos in the javadoc comment of com.linecorp.armeria.common.util.Exceptions methods

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java
@@ -62,7 +62,7 @@ public final class Exceptions {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java
```diff
@@ -62,7 +62,7 @@ public static boolean isVerbose() {
     }
 
     /**
-     * Logs the specified exception if it is {@linkplain #isExpected(Throwable)} unexpected}.
+     * Logs the specified exception if it is unexpected.
```
trustin: Should be corrected as: `{@linkplain #isExpected(Throwable) unexpected}.`

trustin:
Thanks, @AmosDoan !

Issue #18, opened 2015-11-19
Throttling


freesoft:
Since Armeria is RPC library for client/server, it would be great if the library has integrated throttling feature so that server/client can properly manage incoming traffic if they have to. ( unless Armeria already have one, haven't seen it though )

trustin:
We have an option in `ServerConfig.maxPendingRequests()` which is not implemented yet. The plan is to have a configurable policy which tells Armeria what to do when the server is busy enough. (e.g. stop reading, returning 503, applying CoDel, etc)

PR #726, opened 2017-08-20
Implement a decorator that throttles an incoming request

imasahiro:
Motivations:
- A user somtimes want to throttle an incomming request to prevent a service
  from being overwhelmed by too many requests.

Modifications:
- Add RequestThrottling*

Results:
- Fixes https://github.com/line/armeria/issues/18


core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+package com.linecorp.armeria.server.throttling;
+
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Supplier;
+
+import com.linecorp.armeria.common.Request;
+
+/**
+ * Determines whether a failed request should be retried.
```
minwoox: a request should be throttled?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    static RequestThrottlingStrategy rateLimit(float rate) {
+        return rateLimit(rate, ThreadLocalRandom::current);
+    }
+
+    /**
+     * Returns a {@link RequestThrottlingStrategy} that throttles {@code rate} percent of incoming requests.
+     */
+    static RequestThrottlingStrategy rateLimit(float rate, Supplier<Random> randomSupplier) {
+        return request -> randomSupplier.get().nextFloat() > rate;
```
minwoox: requireNonNull(randomSupplier, "randomSupplier");

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingService.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Creates a new decorator using the specified {@link RequestThrottlingStrategy} instance.
+     *
+     * @param strategy The {@link RequestThrottlingStrategy} instance to be used
+     */
+    public static <I extends Request, O extends Response>
+    Function<Service<I, O>, RequestThrottlingService<I, O>>
+    newDecorator(RequestThrottlingStrategy strategy) {
+        return newDecorator((ctx, req) -> strategy);
```
minwoox: null check for the strategy?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingService.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private final RequestThrottlingMapping mapping;
+
+    /**
+     * Creates a new instance that decorates the specified {@link Service}.
+     */
+    protected RequestThrottlingService(Service<I, O> delegate,
+                                       RequestThrottlingMapping mapping) {
+        super(delegate);
+        this.mapping = mapping;
```
minwoox: null check for the mapping?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingService.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public O serve(ServiceRequestContext ctx, I req) throws Exception {
+        RequestThrottlingStrategy strategy = mapping.get(ctx, req);
+        if (strategy.shouldThrottle(req)) {
+            ctx.meterRegistry()
+               .counter("armeria_request_throttling", Tags.zip("name", strategy.name(),
+                                                               "type", "throttled_count"))
+               .increment();
+            throw new HttpResponseException(HttpStatus.TOO_MANY_REQUESTS);
```
minwoox: I am not very sure, but `TOO_MANY_REQUESTS` seems like used when "a client" send too many requests

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingMapping.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+@FunctionalInterface
+public interface RequestThrottlingMapping {
+
+    /**
+     * Creates a new {@link RequestThrottlingMapping} that binds one {@link RequestThrottlingStrategy}
+     * per method.
+     */
+    static RequestThrottlingMapping methodStrategy(Map<String, RequestThrottlingStrategy> strategies) {
+        return (ctx, req) -> {
```
minwoox: null check for the strategies?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Returns whether a request should be throttled according to the given request.
+     */
+    boolean shouldThrottle(Request request);
```
trustin: - `CompletableFuture<Boolean>`?
- `request` could be parameterized?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Returns a name of this {@link RequestThrottlingStrategy}.
+     */
+    default String name() {
+        return "";
```
trustin: Auto-generate with `AtomicInteger.incrementAndGet()`?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+     * Returns a {@link RequestThrottlingStrategy} that throttles {@code rate} percent of incoming requests.
+     */
+    static RequestThrottlingStrategy rateLimit(float rate) {
+        return rateLimit(rate, ThreadLocalRandom::current);
+    }
+
+    /**
+     * Returns a {@link RequestThrottlingStrategy} that throttles {@code rate} percent of incoming requests.
+     */
+    static RequestThrottlingStrategy rateLimit(float rate, Supplier<Random> randomSupplier) {
```
trustin: I'm not sure this will ever be used directly by a user. What a user wants would be to create a `RequestThrottlingStrategy` which can be turned on/off dynamically.

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingService.java
```diff
@@ -0,0 +1,81 @@
...
+        this.mapping = mapping;
+    }
+
+    @Override
+    public O serve(ServiceRequestContext ctx, I req) throws Exception {
+        RequestThrottlingStrategy strategy = mapping.get(ctx, req);
+        if (strategy.shouldThrottle(req)) {
+            ctx.meterRegistry()
+               .counter("armeria_request_throttling", Tags.zip("name", strategy.name(),
+                                                               "type", "throttled_count"))
```
trustin:
```java
.counter(MeterRegistryUtil.name(registry, MeterUnit.NONE_TOTAL, "armeria", "server", "throttledRequests"),
         Tags.zip("name", strategy.name()))
```

Also, please keep the reference to the `Counter` so we don't keep looking it up. e.g. A weak map of strategy and `Counter`.

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingService.java
trustin: - How about adding a protected callback method with default behavior, just like `HttpAuthService`?
- How about using `SERVICE_UNAVAILABLE`?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Determines whether a failed request should be retried.
+ */
+@FunctionalInterface
+public interface RequestThrottlingStrategy {
+    /**
+     * A {@link RequestThrottlingStrategy} that defines a throttling should not be performed.
+     */
+    RequestThrottlingStrategy NEVER = request -> false;
```
trustin: - How about making this a method? `never()`
- How about adding `always()` as well?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingStrategy.java
imasahiro: Added sample impl for RequestThrottlingMapping that can be turend on/off dynamically based on file. WDYT?
https://github.com/line/armeria/pull/726/files#diff-7a7f8ccd97b41bee2ddf18eda232edaf

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingMapping.java
```diff
@@ -0,0 +1,48 @@
...
+     * per method.
+     */
+    static RequestThrottlingMapping methodStrategy(Map<String, RequestThrottlingStrategy> strategies) {
+        requireNonNull(strategies, "strategies");
+        return (ctx, req) -> {
+            String method = req instanceof RpcRequest ? ((RpcRequest) req).method() : ctx.method().name();
+            return strategies.getOrDefault(method, RequestThrottlingStrategy.never());
+        };
+    }
+
```
trustin: How about adding the factory method for file-based one?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingMapping.java
```diff
@@ -0,0 +1,48 @@
...
+import com.linecorp.armeria.common.Request;
+import com.linecorp.armeria.common.RpcRequest;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+/**
+ * Returns a {@link RequestThrottlingStrategy} instance from remote invocation parameters.
+ */
+@FunctionalInterface
+public interface RequestThrottlingMapping {
```
trustin: I'm curious why we need this interface. Couldn't we just add `ServiceRequestContext` to `RequestThrottlingStrategy.shouldThrottle(..)` and compose strategies?

core/src/main/java/com/linecorp/armeria/server/throttling/RequestThrottlingStrategy.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+import com.linecorp.armeria.common.Request;
+
+/**
+ * Determines whether a request should be throttled.
+ */
+@FunctionalInterface
+public interface RequestThrottlingStrategy {
```
trustin: How about just `ThrottlingStrategy` just like `RetryStrategy`?

core/src/main/java/com/linecorp/armeria/server/throttling/FileBasedDynamicRequestThrottlingMapping.java
```diff
@@ -0,0 +1,132 @@
...
+            return RequestThrottlingStrategy.never();
+        }
+        switch (Ascii.toLowerCase(spec.get(0))) {
+            case "always":
+                return RequestThrottlingStrategy.always();
+            case "never":
+                return RequestThrottlingStrategy.never();
+            case "rate":
+                float rate = Floats.tryParse(spec.get(1));
+                return req -> CompletableFuture.completedFuture(rate > ThreadLocalRandom.current().nextFloat());
```
kojilin: static import?

core/src/main/java/com/linecorp/armeria/server/throttling/FileBasedDynamicRequestThrottlingMapping.java
```diff
@@ -0,0 +1,132 @@
...
+            default:
+                throw new IllegalArgumentException("Unknown key " + spec.get(0));
+        }
+    }
+
+    @Override
+    public RequestThrottlingStrategy get(ServiceRequestContext ctx, Request req) {
+        String method = req instanceof RpcRequest ? ((RpcRequest) req).method() : ctx.method().name();
```
kojilin: Can we do like path#method ? sometimes same method name may bind to different path. And I think it's good to using same pattern for http handler. e.g. `/api#POST`.

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingService.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Invoked when {@code req} is throttled. By default, this method responds with the
+     * {@link HttpStatus#SERVICE_UNAVAILABLE} status.
+     */
+    protected HttpResponse onFailure(ServiceRequestContext ctx, HttpRequest req, @Nullable Throwable cause)
+            throws Exception {
+        final DefaultHttpResponse res = new DefaultHttpResponse();
+        res.respond(HttpStatus.UNAUTHORIZED);
```
hyangtack: How about getting a throttled response status code from `ThrottlingStrategy`? A user might want to have their own throttling status code, so I think it would be better to be customizable.

core/src/main/java/com/linecorp/armeria/server/throttling/FileBasedDynamicThrottlingStrategy.java
```diff
@@ -0,0 +1,150 @@
...
+                return (ctx, req) -> completedFuture(rate > ThreadLocalRandom.current().nextFloat());
+            default:
+                throw new IllegalArgumentException("Unknown key " + spec.get(0));
+        }
+    }
+
+    @Override
+    public CompletableFuture<Boolean> shouldThrottle(ServiceRequestContext ctx, Request req) {
+        String method = req instanceof RpcRequest ? ((RpcRequest) req).method() : ctx.method().name();
+        return strategies.getOrDefault(method, ThrottlingStrategy.never()).shouldThrottle(ctx, req);
```
hyangtack: Just curious. Is this enough to throttle a request based on the request method? What do you think about adding a simple path pattern as a key?

core/src/main/java/com/linecorp/armeria/server/throttling/FileBasedDynamicThrottlingStrategy.java
```diff
@@ -0,0 +1,157 @@
...
+    public FileBasedDynamicThrottlingStrategy(String settingFile) {
+        this.settingFile = FileSystems.getDefault().getPath(settingFile);
+        try {
+            watchService = FileSystems.getDefault().newWatchService();
+            this.settingFile.getParent().register(watchService,
+                                                  new WatchEvent.Kind[] {
+                                                          StandardWatchEventKinds.ENTRY_CREATE,
+                                                          StandardWatchEventKinds.ENTRY_MODIFY
+                                                  },
+                                                  SensitivityWatchEventModifier.HIGH);
```
jmostella: It looks like this will poll the file system every 2 seconds. Maybe this excessive? or should be configurable at least?

imasahiro:
Added TokenBucketThrottlingStrategy that throttoles the request based on current qps.
Also added CircuitBreakerThrottlingStrategy that throttoles the request based on existing CircuitBreaker

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingStrategy.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * A {@link ThrottlingStrategy} that defines a throttling should not be performed.
+     */
+    public static <T extends Request> ThrottlingStrategy<T> never() {
+        return of((ctx, request) -> completedFuture(false));
+    }
+
+    /**
+     * A {@link ThrottlingStrategy} that defines a throttling should be performed.
```
trustin: - A -> Returns a singleton
- that defines ... -> that fails all requests.
- Could we make this method return a singleton?

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingStrategy.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Returns whether a request should be throttled according to the given request.
+     */
+    public abstract CompletableFuture<Boolean> shouldThrottle(ServiceRequestContext ctx, T request);
+
+    /**
+     * Returns a name of this {@link ThrottlingStrategy}.
+     */
+    public String name() {
+        return "strategy" + strategyId;
```
trustin: `"throttling-strategy-"`? (to be consistent with circuit breakers)

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingStrategy.java
```diff
@@ -0,0 +1,74 @@
...
+        return new ThrottlingStrategy<T>() {
+            @Override
+            public CompletableFuture<Boolean> shouldThrottle(ServiceRequestContext ctx, T request) {
+                return function.apply(ctx, request);
+            }
+        };
+    }
+
+    /**
+     * Returns whether a request should be throttled according to the given request.
```
trustin: How about: Returns whether a given request should be treated as failed before it is handled actually?

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingStrategy.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public abstract CompletableFuture<Boolean> shouldThrottle(ServiceRequestContext ctx, T request);
```
trustin: How about inverting the meaning of this method and renaming to `accept` or `canServe`? e.g. fail 20% of incoming requests vs accept 80% of incoming requests.

core/src/main/java/com/linecorp/armeria/server/throttling/TokenBucketThrottlingStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import com.google.common.math.LongMath;
+
+import com.linecorp.armeria.common.Request;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+/**
+ * A {@link ThrottlingStrategy} that throttles a request using
+ * {@see <a href="http://en.wikipedia.org/wiki/Token_bucket">Token Bucket</a>}.
```
trustin: `{@see}` is unnecessary?

core/src/main/java/com/linecorp/armeria/server/throttling/TokenBucketThrottlingStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+    private final int tokenBucketCapacity;
+    private final long refillInterval;
+    private volatile int tokensInBucket;
+    private volatile long nextRefillTime;
+
+    /**
+     * Creates a new strategy.
+     */
+    public TokenBucketThrottlingStrategy(int tokenBucketCapacity, int averageQPS, Duration refillInterval) {
+        checkArgument(tokenBucketCapacity > 0, "tokenBucketCapacity > 0");
```
trustin: `tokenBucketCapacity: %s (expected: > 0)", tokenBucketCapacity`

core/src/main/java/com/linecorp/armeria/server/throttling/TokenBucketThrottlingStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+        checkArgument(averageQPS > 0, "averageQPS > 0");
```
trustin: `averageQPS: %s (expected: > 0)", averageQPS`

core/src/main/java/com/linecorp/armeria/server/throttling/TokenBucketThrottlingStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public CompletableFuture<Boolean> shouldThrottle(ServiceRequestContext ctx, T request) {
+        refillTokenIfNeeded();
+        if (tokensInBucket < 1) { // bucket is empty.
+            return completedFuture(true);
+        }
+
+        tokensInBucket--;
```
trustin: Thread safety? Probably better using `AtomicInteger` and CaS?

core/src/main/java/com/linecorp/armeria/server/throttling/FileBasedDynamicThrottlingStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+    private Thread startWatching() {
+        Thread thread = new Thread(() -> {
+            try {
+                while (true) {
+                    try {
+                        WatchKey key = watchService.take();
+                        for (WatchEvent<?> event : key.pollEvents()) {
+                            // We only register "ENTRY_MODIFY" so the context is always a Path.
+                            final Path changed = (Path) event.context();
+                            if (changed.equals(settingFile.getFileName())) {
```
anuraaga: This comparison doesn't work when symlinks are involved (e.g., a Kubernetes config map where the specified human-readable filename like `settings.txt` will be a symlink to a timestamped uuid type filename like `settings.fq38rua3r908auf.txt`, and the event gets triggered with the timestamped file, not symlink file update.

I'm not sure if it's the best solution, but resolving the real paths during this check could help
https://github.com/curioswitch/curiostack/commit/0164c45c76df6897ed7bc5c7ebc2efe6da938272

core/src/main/java/com/linecorp/armeria/server/throttling/TokenBucketThrottlingStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.common.math.LongMath;
+
+import com.linecorp.armeria.common.Request;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+/**
+ * A {@link ThrottlingStrategy} that throttles a request using
+ * {@see <a href="http://en.wikipedia.org/wiki/Token_bucket">Token Bucket</a>}.
+ */
+public final class TokenBucketThrottlingStrategy<T extends Request> extends ThrottlingStrategy<T> {
```
anuraaga: You might be able to use Guava's `RateLimiter` and `return completedFuture(limiter.tryAcquire())`

core/src/main/java/com/linecorp/armeria/server/throttling/FileBasedDynamicThrottlingStrategy.java
```diff
@@ -0,0 +1,174 @@
...
+        return thread;
+    }
+
+    private void updateThrottlingSettings() {
+        Map<String, String> newStrategies;
+        try {
+            newStrategies = OBJECT_MAPPER.readValue(settingFile.toFile(),
+                                                    new TypeReference<Map<String, String>>() {});
+        } catch (IOException e) {
+            logger.warn("Could not read hosts file", e);
```
trustin: `hosts file -> the settings file: {}", settingFile, e` ?

imasahiro:
Pushed the changed but following FileBasedDynamicThrottlingStrategy.java comments are not reached yet.
https://github.com/line/armeria/pull/726#discussion_r143634419
https://github.com/line/armeria/pull/726#discussion_r143638201

> You may want to send a separate PR for strategy/mapping implementations because building a highly customizable strategy will make this PR take longer.

I feel, it better to split the PR like 1) add throttling feature with basic strategies, and 2) add other strategies like FileBasedDynamicThrottlingStrategy.. WDYT?

trustin:
> I feel, it better to split the PR like 1) add throttling feature with basic strategies, and 2) add other strategies like FileBasedDynamicThrottlingStrategy.. WDYT?

SGTM. Thanks for your patience!

core/src/main/java/com/linecorp/armeria/server/throttling/FileBasedDynamicThrottlingStrategy.java
imasahiro: Nice catch! And thanks for reference.
To focus on other basic throttoring, I removed FileBasedDynamicThrottlingStrategy.java from this PR  strategies. Let me try dig more by myself on next PR.

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Returns whether a given request should be treated as failed before it is handled actually.
+     */
+    public abstract CompletableFuture<Boolean> accept(ServiceRequestContext ctx, T request);
+
+    /**
+     * Returns the name of this {@link ThrottlingStrategy}.
+     */
+    public String name() {
+        return "throttling-strategy-" + strategyId;
```
anuraaga: Not sure but this might work and provide more info

`getClass().isAnonymousClass ? strategyid : getClass().getName`

core/src/test/java/com/linecorp/armeria/server/throttling/RateLimitingThrottlingStrategyTest.java
```diff
@@ -0,0 +1,56 @@
...
+    };
+
+    @Test
+    public void rateLimit() throws Exception {
+        HttpClient client = HttpClient.of(serverRule.uri("/"));
+        assertThat(client.get("/rate-limit").aggregate().get().status())
+                .isEqualTo(HttpStatus.OK);
+        client.get("/rate-limit").aggregate().get();
+        client.get("/rate-limit").aggregate().get();
+        // Reached to limit
```
anuraaga: This seems like it can be flaky? How about just using a unit test instead of server test?

core/src/test/java/com/linecorp/armeria/server/throttling/RateLimitingThrottlingStrategyTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class RateLimitingThrottlingStrategyTest {
+    @Test
+    public void rateLimit() throws Exception {
+        RateLimitingThrottlingStrategy<Request> strategy = new RateLimitingThrottlingStrategy<>(1);
+        assertThat(strategy.accept(null, null).get()).isEqualTo(true);
+        assertThat(strategy.accept(null, null).get()).isEqualTo(false);
+        assertThat(strategy.accept(null, null).get()).isEqualTo(false);
+
+        TimeUnit.SECONDS.sleep(1);
```
anuraaga: I don't think this changes the flakiness much, we should add a `@VisibleForTesting` constructor that can take a mocked ratelimiter, or at least use awaitility here instead of a fixed sleep

imasahiro: added VisibleForTesting constructor and inject mocked RateLimiter.

core/src/main/java/com/linecorp/armeria/server/throttling/RateLimitingThrottlingStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public final class RateLimitingThrottlingStrategy<T extends Request> extends ThrottlingStrategy<T> {
+    private final RateLimiter rateLimiter;
+
+    /**
+     * Creates a new strategy with specified name.
+     *
+     * @param requestPerSecond a number of requests per one second this {@link ThrottlingStrategy} accepts.
+     */
+    public RateLimitingThrottlingStrategy(int requestPerSecond, @Nullable String name) {
```
trustin: Should use `double`

core/src/main/java/com/linecorp/armeria/server/throttling/RateLimitingThrottlingStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+        checkArgument(requestPerSecond > 0, "requestPerSecond: %s (expected: > 0)", requestPerSecond);
+        rateLimiter = RateLimiter.create(requestPerSecond);
+    }
+
+    /**
+     * Creates a new strategy.
+     *
+     * @param requestPerSecond a number of requests per one second this {@link ThrottlingStrategy} accepts.
+     */
+    public RateLimitingThrottlingStrategy(int requestPerSecond) {
```
trustin: Should use `double`

trustin:
Thanks, @imasahiro !

imasahiro:
Thanks a lot, all reviewers!

PR #729, opened 2017-08-22
Update Micrometer to 0.10.0

trustin:
- Add common.metric.MeterId because the MeterId in Micrometer is not
  public anymore
- Remove MeterRegistryUtil
  - Remove MeterUnit, MeterRegistryUtil.name() and prefersBaseUnit()
    because Micrometer now has NamingConvention
  - Remove MeterRegistryUtil.tags() because Tags now has concat()
  - Move other methods to MoreMeters
- Add MoreNamingConventions which provides sensible default
  NamingConvention implementations for Dropwizard Metrics and Prometheus
- Replace PathMapping.metricName() with meterTag() so it is more aligned
  with Micrometer API


core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
```diff
@@ -0,0 +1,168 @@
...
+     * Returns a {@link Timer} with the default {@link Quantiles} configured.
+     */
+    public static Timer timerWithDefaultQuantiles(MeterRegistry registry, MeterId id) {
+        requireNonNull(registry, "registry");
+        requireNonNull(id, "id");
+        return registry.timerBuilder(id.name()).tags(id.tags()).quantiles(defaultQuantiles()).create();
+    }
+
+    private static Quantiles defaultQuantiles() {
+        // According to Jon Schneider of Micrometer:
```
imasahiro: Do you have any url link for this discussion?

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
```diff
@@ -0,0 +1,168 @@
...
+        return builder.build();
+    }
+
+    private static String measurementName(Object id, Measurement measurement) {
+        final StringBuilder buf = new StringBuilder();
+
+        // Append name.
+        final String name;
+        name = getName(id);
+        buf.append(name);
```
imasahiro: Can you append `name` to `buf` directly?

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
trustin: Unfortunately, it was a Slack chat. Let me state that it's from Slack.

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
trustin: Thanks! Fixed

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/MetricCollectingCircuitBreakerListener.java
```diff
@@ -33,29 +33,29 @@
...
- * <tr><td>{@code <name>_rejected_requests{name="<circuit breaker name>"}}</td>
+ * <tr><td>{@code <name>.rejectedRequests{name="<circuit breaker name>"}}</td>
  * <td>The number of requests rejected by the circuit breaker.</td></tr>
  *
  * </table>
  */
 public final class MetricCollectingCircuitBreakerListener implements CircuitBreakerListener {
 
-    private static final String DEFAULT_NAME = "armeria_client_circuit_breaker";
+    private static final String DEFAULT_NAME = "armeria.client.circuitBreaker";
```
jmostella: I think we should distinguish between ``tag``name and ``metric``name. 
maybe ``METRIC_NAME = ...``

trustin: Renamed to `DEFAULT_METER_NAME`. Thanks! Anything else?

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * {@link MeterRegistry} implementation for <a href="http://metrics.dropwizard.io/">Dropwizard Metrics</a>.
+ * This implementation adds more convenient constructors and sets more sensible default {@link NamingConvention}
+ * and {@link HierarchicalNameMapper} on top of the upstream implementation.
+ */
+@SuppressWarnings("ClassNameSameAsAncestorName")
+public class DropwizardMeterRegistry extends io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry {
+
+    private static final HierarchicalNameMapper DEFAULT_NAME_MAPPER = (name, tags) -> {
+        final Iterator<Tag> i = tags.iterator();
```
minwoox: Iterables.isEmpty(tags) ?

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+        Streams.stream(i).sorted(comparing(Tag::getKey)).forEach(tag -> {
+            buf.append('.').append(tag.getKey());
+            buf.append(':').append(tag.getValue());
+        });
+
+        return buf.toString();
+    };
+
+    /**
+     * Creates a new instance with the default {@link HierarchicalNameMapper}..
```
minwoox: two dots?

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public DropwizardMeterRegistry(HierarchicalNameMapper nameMapper) {
+        this(nameMapper, Clock.SYSTEM);
+    }
+
+    /**
+     * Creates a new instance with the specified {@link HierarchicalNameMapper}.
+     */
+    public DropwizardMeterRegistry(HierarchicalNameMapper nameMapper, Clock clock) {
+        super(nameMapper, clock);
```
minwoox: requireNonNull since super class are not checking them.. ?

core/src/main/java/com/linecorp/armeria/common/metric/MeterId.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+
+/**
+ * An immutable holder of {@link Meter} name and {@link Tag}s.
```
minwoox: a {@link Meter} name ?

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
trustin: It use the `Iterator` to create a stream from it so `tags.iterator()` isn't called twice. (`Streams.stream(i)`)

minwoox: I missed that part! Sorry~!

core/src/main/java/com/linecorp/armeria/common/metric/MeterId.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
trustin: Doh!

core/src/main/java/com/linecorp/armeria/common/metric/PrometheusMeterRegistry.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public PrometheusMeterRegistry(CollectorRegistry registry) {
+        this(registry, Clock.SYSTEM);
+    }
+
+    /**
+     * Creates a new instance with the specified {@link CollectorRegistry}.
+     */
+    public PrometheusMeterRegistry(CollectorRegistry registry, Clock clock) {
+        super(registry, clock);
```
minwoox: ditto!

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
```diff
@@ -0,0 +1,79 @@
...
+
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.NamingConvention;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.util.HierarchicalNameMapper;
+
+/**
+ * {@link MeterRegistry} implementation for <a href="http://metrics.dropwizard.io/">Dropwizard Metrics</a>.
+ * This implementation adds more convenient constructors and sets more sensible default {@link NamingConvention}
```
anuraaga: since on top of upstream implementation is sort of implied, can simplify to
`sets more sensible defaults for Naming Convention and HierarchicalNameMapper.`

core/src/main/java/com/linecorp/armeria/common/metric/PrometheusMeterRegistry.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/metric/MeterId.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private static ImmutableList<Tag> sort(List<Tag> tags) {
+        tags.sort(comparing(Tag::getKey));
+        return ImmutableList.copyOf(tags);
+    }
+
+    private static ImmutableList<Tag> copyAndSort(Iterable<Tag> tags) {
+        final List<Tag> list = new ArrayList<>();
+        tags.forEach(list::add);
```
anuraaga: Can use `Lists.newArrayList(Iterable)` throughout file.
Can also consider `Streams.stream(tags).sorted().collect(toImmutableList)`

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
```diff
@@ -0,0 +1,166 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.common.metric;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.Streams.stream;
```
anuraaga: Recommend being consistent in PR on whether using `Streams.stream` or static import

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Micrometer, as of 0.10.0, has a bug where Meter.getName() sometimes returns conventional name
+     * and sometimes not. We use reflection to get the consistent meter names as a workaround.
+     */
+    @SuppressWarnings("unchecked")
+    private static Map<Object, Meter> getMeterMap(MeterRegistry registry) {
+        try {
+            @SuppressWarnings("JavaReflectionMemberAccess")
+            final Field field = AbstractMeterRegistry.class.getDeclaredField("meterMap");
```
anuraaga: If I understand correctly, I think the JVM can optimize reflection much better, including inlining, if the field and method handles are class constants.

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
```diff
@@ -0,0 +1,166 @@
...
+            return (Map<Object, Meter>) field.get(registry);
+        } catch (Exception e) {
+            throw new IllegalStateException("failed to retrieve the meter map", e);
+        }
+    }
+
+    private static String getName(Object id) {
+        checkMeterIdType(id);
+        try {
+            final Method getName = id.getClass().getDeclaredMethod("getName");
```
anuraaga: Can you use `MeterId.getDeclaredMethod("getName")` instead? Then it can be a constant.

core/src/main/java/com/linecorp/armeria/common/metric/MoreNamingConventions.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Returns the {@link NamingConvention} of <a href="https://prometheus.io/">Prometheus</a>.
+     */
+    public static NamingConvention prometheus() {
+        return BetterPrometheusNamingConvention.INSTANCE;
+    }
+
+    private MoreNamingConventions() {}
+
+    private static final class BetterPrometheusNamingConvention implements NamingConvention {
```
anuraaga: How is it better? :D

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
```diff
@@ -0,0 +1,79 @@
...
+ * and {@link HierarchicalNameMapper} on top of the upstream implementation.
+ */
+@SuppressWarnings("ClassNameSameAsAncestorName")
+public class DropwizardMeterRegistry extends io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry {
```
anuraaga: Can this and the prometheus registery just be factories instead of inheritance?

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
trustin: Sadly, it's a package-private class so I can't access it statically.

trustin: Or.. can I just use `Class.forName()`? Let me try.

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
trustin: Didn't want to have a factory class for each registry type, because otherwise a user has to pull in both dropwizard and prometheus even if s/he doesn't use both.

core/src/main/java/com/linecorp/armeria/common/metric/MeterId.java
trustin: Now using `Lists.newArrayList(Iterable)` where necessary.

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
trustin: Good idea. Fixed

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
anuraaga: Ah I didn't mean combining them into the same class. But instead of using inheritance, I think this could just be simple factories like this, so no need to worry about the `ClassNameSameAsAncestorName` that can be a bit confusing.

```java
public final class DropwizardMeterRegistries {
  public static DropwizardMeterRegistry default() {
    return of(DEFAULT_NAME_MAPPER);
  }

  public static DropwizardMeterRegistery of(nameMapper) {
    return of(nameMapper, Clock.System);
  }

  public static DropwizardMeterRegistry of(nameMapper, clock) {
    DropwizardMeterRegistry registry = new DropwizardMeterRegistry(nameMapper, clock);
    registry.config().namingConvention(MoreNamingConventions.dropwizard());
  }
}
```

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/metric/MoreNamingConventions.java
trustin: Good question. Added class Javadoc :-)

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistry.java
trustin: Okie dokie

trustin: Added `DropwizardMeterRegistries` and `PrometheusMeterRegistries`. Method names are `newRegistry()` though.

trustin:
Thanks for reviewing everyone.

PR #731, opened 2017-08-23
Add service(ServiceWithPathMapping,...) to AbstractVirtualHostBuilder

trustin:
.. because ServerBuilder and AbstractVirtualHostBuilder should have the
same set of service-adding methods

/cc @demongaorui

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
@@ -254,6 +254,28 @@ abstract class AbstractVirtualHostBuilder<B extends AbstractVirtualHostBuilder> 
     }
 
     /**
+     * Binds the specified {@link ServiceWithPathMappings} at multiple {@link PathMapping}s.
+     */
+    public <T extends ServiceWithPathMappings<HttpRequest, HttpResponse>>
+    B service(T serviceWithPathMappings) {
+        return service(serviceWithPathMappings, Function.identity());
+    }
+
+    /**
+     * Decorates and binds the specified {@link ServiceWithPathMappings} at multiple {@link PathMapping}s.
+     */
+    public <T extends ServiceWithPathMappings<HttpRequest, HttpResponse>,
+            R extends Service<HttpRequest, HttpResponse>>
+    B service(T serviceWithPathMappings, Function<T, R> decorator) {
+        requireNonNull(serviceWithPathMappings, "serviceWithPathMappings");
+        requireNonNull(serviceWithPathMappings.pathMappings(), "serviceWithPathMappings.pathMappings()");
+        requireNonNull(decorator, "decorator");
+        final Service<HttpRequest, HttpResponse> decorated = decorator.apply(serviceWithPathMappings);
+        serviceWithPathMappings.pathMappings().forEach(pathMapping -> service(pathMapping, decorated));
+        return self();
+    }
+
+    /**
      * Binds the specified annotated service object under the path prefix {@code "/"}.
      */
     public B annotatedService(Object service) {
--- a/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
@@ -420,19 +420,21 @@ public final class ServerBuilder {
     }
 
     /**
-     * Binds the specified {@link ServiceWithPathMappings} at multiple {@link PathMapping}
+     * Binds the specified {@link ServiceWithPathMappings} at multiple {@link PathMapping}s
      * of the default {@link VirtualHost}.
      *
      * @throws IllegalStateException if the default {@link VirtualHost} has been set via
      *                               {@link #defaultVirtualHost(VirtualHost)} already
      */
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -420,32 +420,31 @@ public ServerBuilder service(PathMapping pathMapping, Service<HttpRequest, HttpR
...
      */
     public <T extends ServiceWithPathMappings<HttpRequest, HttpResponse>,
             R extends Service<HttpRequest, HttpResponse>>
-            ServerBuilder service(T serviceWithPathMapping, Function<T, R> decorator) {
-        requireNonNull(serviceWithPathMapping, "serviceWithPathMapping");
-        requireNonNull(serviceWithPathMapping.pathMappings(), "serviceWithPathMapping.pathMappings()");
-        requireNonNull(decorator, "decorator");
-        serviceWithPathMapping.pathMappings()
-                .forEach(pathMapping -> service(pathMapping, decorator.apply(serviceWithPathMapping)));
+    ServerBuilder service(T serviceWithPathMappings, Function<T, R> decorator) {
```
demongaorui: indention

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -254,6 +254,28 @@ public B service(PathMapping pathMapping, Service<HttpRequest, HttpResponse> ser
...
+    B service(T serviceWithPathMappings) {
+        return service(serviceWithPathMappings, Function.identity());
+    }
+
+    /**
+     * Decorates and binds the specified {@link ServiceWithPathMappings} at multiple {@link PathMapping}s.
+     */
+    public <T extends ServiceWithPathMappings<HttpRequest, HttpResponse>,
+            R extends Service<HttpRequest, HttpResponse>>
+    B service(T serviceWithPathMappings, Function<T, R> decorator) {
```
demongaorui: indention

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -254,6 +254,28 @@ public B service(PathMapping pathMapping, Service<HttpRequest, HttpResponse> ser
     }
 
     /**
+     * Binds the specified {@link ServiceWithPathMappings} at multiple {@link PathMapping}s.
+     */
+    public <T extends ServiceWithPathMappings<HttpRequest, HttpResponse>>
+    B service(T serviceWithPathMappings) {
```
demongaorui: indention

trustin:
> With gradle google java format plugin, we could apply format to all the codes with only one command ./gradlew :googleJavaFormat. Could you consider using it for format/style in Armeria?

No, because we are using slightly different convention internally. I would not recommend using automated code formatter for any conventions, because it often breaks handcrafted indentations, making the code look worse.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: Actually, we use this indentation for all other places where type parameter definitions are too long.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -420,32 +420,31 @@ public ServerBuilder service(PathMapping pathMapping, Service<HttpRequest, HttpR
...
+     * Binds the specified {@link ServiceWithPathMappings} at multiple {@link PathMapping}s
      * of the default {@link VirtualHost}.
      *
      * @throws IllegalStateException if the default {@link VirtualHost} has been set via
      *                               {@link #defaultVirtualHost(VirtualHost)} already
      */
     public <T extends ServiceWithPathMappings<HttpRequest, HttpResponse>>
-            ServerBuilder service(T serviceWithPathMapping) {
-        return service(serviceWithPathMapping, Function.identity());
+            ServerBuilder service(T serviceWithPathMappings) {
```
demongaorui: This indention is different with  service(T serviceWithPathMappings) in HostBuilder, is it OK?

demongaorui:
Actually, googleJavaFormat could handle the indention of codes/docs and auto format them pretty beautiful as long as codes/docs adhere to [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html). 

Comparing to resolve [Wrong Import Order] alike checkStyle issues and manual format check cost, some compromise for rare intentioned special indention cases are not that critical. Further more auto formatters are improving as well. Give it a try, maybe you will like it. 

While, if auto formatters have conflict with your Corp internal convention, then maybe find a way to auto apply your convention is better IMO. But this is a little out of scope of this PR. Could discussion this topic again in future.

For this PR, no more concerns, only https://github.com/line/armeria/pull/731#discussion_r134679934 left now 😆

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: Oops. Fixed :-)

PR #732, opened 2017-08-23
Update external Javadocs

trustin:
- Add micrometer
- Remove spring-metrics which is not used anymore
- Update the links of:
  - Caffeine
  - libthrift


site/src/apidocs/package-lists/micrometer/package-list
```diff
@@ -0,0 +1,22 @@
+io.micrometer.core.annotation
```
anuraaga: Just curious, is there a way to automatically generate this list?

trustin: Javadoc generates this file from `.class` files, which means it's not possible without compiling Micrometer. I think it's possible to write a build plugin/script that fetches the `package-list` files from the Javadoc site and caches them. But itch didn't make me go that far yet. :-)

PR #733, opened 2017-08-24
Use NoopMeterRegistry by default

trustin:
Motivation:

Using SimpleMeterRegistry by default doesn't make much sense because a
user will always want to switch to a better MeterRegistry implementation
such as DropwizardMeterRegistry or PrometheusMeterRegistry.

Modifications:

- Add NoopMeterRegistry and use it by default

Result:

- Less overhead unless a user explicitly requested metric recording


trustin:
Thanks, @minwoox !

PR #734, opened 2017-08-24
Update dependencies and fix checkstyle violations

trustin:
- Update dependencies
  - Brave 4.5.1 -> 4.5.2
  - Caffeine 2.5.3 -> 2.5.5
  - Checkstyle 8.0 -> 8.1
  - Dropwizard Metrics 3.2.3 -> 3.2.4
  - gRPC 1.4.0 -> 1.5.0
  - Guava 22.0 -> 23.0
  - JSON-unit 1.24.0 -> 1.25.0
  - Kafka 0.10.2.1 -> 0.11.0.0
  - Netty 4.1.13 -> 4.1.14
  - Reactive Streams 1.0.1-RC1 -> 1.0.1
  - Spring Boot 1.5.4 -> 1.5.6
  - Tomcat 8.5.16 -> 8.5.20 / 8.0.45 -> 8.0.46
- Fix new checkstyle violations reported by 8.1

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/MediaType.java
+++ b/core/src/main/java/com/linecorp/armeria/common/MediaType.java
@@ -422,7 +422,7 @@ public final class MediaType {
     public static final MediaType PDF = createConstant(APPLICATION_TYPE, "pdf");
     public static final MediaType POSTSCRIPT = createConstant(APPLICATION_TYPE, "postscript");
     /**
-     * <a href="http://tools.ietf.org/html/draft-rfernando-protocol-buffers-00">Protocol buffers</a>
+     * <a href="http://tools.ietf.org/html/draft-rfernando-protocol-buffers-00">Protocol buffers</a>.
      */
     public static final MediaType PROTOBUF = createConstant(APPLICATION_TYPE, "protobuf");
 
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
@@ -201,6 +201,8 @@ abstract class AbstractVirtualHostBuilder<B extends AbstractVirtualHostBuilder> 
     }
 
     /**
+     * Binds the specified {@link Service} at the specified path pattern.
+     *
      * @deprecated Use {@link #service(String, Service)} instead.
      */
     @Deprecated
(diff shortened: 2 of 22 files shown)


trustin:
@kawamuray I'm curious if it's OK to upgrade `org.apache.kafka:kafka-clients` to 0.11.0.0 from 0.10.2.1.

core/src/main/java/com/linecorp/armeria/common/MediaType.java
```diff
@@ -422,7 +422,7 @@ private static MediaType addKnownType(MediaType mediaType) {
     public static final MediaType PDF = createConstant(APPLICATION_TYPE, "pdf");
     public static final MediaType POSTSCRIPT = createConstant(APPLICATION_TYPE, "postscript");
     /**
-     * <a href="http://tools.ietf.org/html/draft-rfernando-protocol-buffers-00">Protocol buffers</a>
+     * <a href="http://tools.ietf.org/html/draft-rfernando-protocol-buffers-00">Protocol buffers</a>.
```
anuraaga: Would it make sense to ignore checkstyle for purely forked files like this one?

trustin: I wish I can but IIRC @hyangtack made some changes on top of the fork so it's not a pure fork anymore unfortunately.

kawamuray:
> I'm curious if it's OK to upgrade org.apache.kafka:kafka-clients to 0.11.0.0 from 0.10.2.1.

Yeah I think so.
There was an improvement regarding client's protocol compatibility which has been released with `0.10.2.0`: https://issues.apache.org/jira/browse/KAFKA-4462 .
And as same as the last time we bumped `kafka-clients`'s version, people still can downgrade clients by specifiying direct dependency in their `build.gradle` or `pom.xml` if necessary, so I think its' okay.

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -201,6 +201,8 @@ public B sslContext(
     }
 
     /**
+     * Binds the specified {@link Service} at the the specified path pattern.
```
minwoox: at the the :-)

trustin: Fixed :-)

PR #741, opened 2017-08-31
Do shaded tests on all Java modules

trustin:
We did not include non-published Java projects from shaded tests, but
there's no reason that keeps us from doing so. With this change, ':it'
is also shade-tested.


trustin:
Thanks!

PR #744, opened 2017-09-03
Fix DeferredStreamMessage may not call subscribeToDelegate method.

kojilin:
#delegate() and #subscribe() may call from different thread, and there is a timing issue that both method won't call subscribeToDelegate method. Apply this change will fix the problem and solve leaking subscriber issue too.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
@@ -35,6 +35,9 @@ import com.linecorp.armeria.common.util.CompletionActions;
  */
 public class DeferredStreamMessage<T> implements StreamMessage<T> {
 
+    private static final PendingSubscription<Object> NO_OP_PENDING_SUBSCRIPTION =
+            new PendingSubscription<>(null, null, false);
+
     @SuppressWarnings({ "AtomicFieldUpdaterIssues", "rawtypes" })
     private static final AtomicReferenceFieldUpdater<DeferredStreamMessage, StreamMessage> delegateUpdater =
             AtomicReferenceFieldUpdater.newUpdater(
@@ -173,12 +176,13 @@ public class DeferredStreamMessage<T> implements StreamMessage<T> {
         final PendingSubscription<? super T> newPendingSubscription =
                 new PendingSubscription<>(subscriber, executor, withPooledObjects);
 
+        if (!pendingSubscriptionUpdater.compareAndSet(this, null, newPendingSubscription)) {
+            failLateSubscriber(subscriber, executor,
+                               new IllegalStateException("subscribed by other subscriber already"));
+        }
+
         final StreamMessage<T> delegate = this.delegate;
         if (delegate == null) {
-            if (!pendingSubscriptionUpdater.compareAndSet(this, null, newPendingSubscription)) {
-                failLateSubscriber(subscriber, executor,
-                                   new IllegalStateException("subscribed by other subscriber already"));
-            }
             return;
         }
 
@@ -203,15 +207,23 @@ public class DeferredStreamMessage<T> implements StreamMessage<T> {
         }
     }
 
-    private void subscribeToDelegate(
-            StreamMessage<T> delegate, PendingSubscription<? super T> pendingSubscription) {
+    private void subscribeToDelegate(StreamMessage<T> delegate,
+                                     PendingSubscription<? super T> pendingSubscription) {
+
+        if (pendingSubscription == NO_OP_PENDING_SUBSCRIPTION ||
+            !pendingSubscriptionUpdater.compareAndSet(this,
+                                                      pendingSubscription,
+                                                      NO_OP_PENDING_SUBSCRIPTION)) {
+            return;
+        }
 
         final Subscriber<? super T> subscriber = pendingSubscription.subscriber;
         final Executor executor = pendingSubscription.executor;
+        final boolean withPooledObjects = pendingSubscription.withPooledObjects;
         if (executor == null) {
-            delegate.subscribe(subscriber, pendingSubscription.withPooledObjects);
+            delegate.subscribe(subscriber, withPooledObjects);
         } else {
-            delegate.subscribe(subscriber, executor, pendingSubscription.withPooledObjects);
+            delegate.subscribe(subscriber, executor, withPooledObjects);
         }
     }
 
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
@@ -35,7 +35,7 @@ import com.linecorp.armeria.common.util.CompletionActions;
  */
 public class DeferredStreamMessage<T> implements StreamMessage<T> {
 
-    private static final PendingSubscription<Object> NO_OP_PENDING_SUBSCRIPTION =
+    private static final PendingSubscription<Object> NO_OP_ALREADY_SUBSCRIBED =
             new PendingSubscription<>(null, null, false);
 
     @SuppressWarnings({ "AtomicFieldUpdaterIssues", "rawtypes" })
@@ -84,11 +84,7 @@ public class DeferredStreamMessage<T> implements StreamMessage<T> {
             }).exceptionally(CompletionActions::log);
         }
 
-        @SuppressWarnings("unchecked")
-        final PendingSubscription<? super T> ps = pendingSubscription;
-        if (ps != null) {
-            subscribeToDelegate(delegate, ps);
-        }
+        safeOnSubscribeToDelegate();
     }
 
     /**
@@ -179,14 +175,10 @@ public class DeferredStreamMessage<T> implements StreamMessage<T> {
         if (!pendingSubscriptionUpdater.compareAndSet(this, null, newPendingSubscription)) {
             failLateSubscriber(subscriber, executor,
                                new IllegalStateException("subscribed by other subscriber already"));
-        }
-
-        final StreamMessage<T> delegate = this.delegate;
-        if (delegate == null) {
             return;
         }
 
-        subscribeToDelegate(delegate, newPendingSubscription);
+        safeOnSubscribeToDelegate();
     }
 
     private void failLateSubscriber(Subscriber<? super T> subscriber, Executor executor, Throwable cause) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -35,22 +35,34 @@
  */
 public class DeferredStreamMessage<T> implements StreamMessage<T> {
 
+    private enum State {
+        NONE, DELEGATED, SUBSCRIBED, COMPLETE
```
trustin: - NONE -> INIT?
- COMPLETE -> SUBSCRIBED_AND_DELEGATED? (or add some Javadoc?)

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -81,10 +93,14 @@ protected void delegate(StreamMessage<T> delegate) {
...
 
-        @SuppressWarnings("unchecked")
-        final PendingSubscription<? super T> ps = pendingSubscription;
-        if (ps != null) {
-            subscribeToDelegate(delegate, ps);
+        for (;;) {
+            if (stateUpdater.compareAndSet(this, State.NONE, State.DELEGATED)) {
+                return;
+            }
+            if (stateUpdater.compareAndSet(this, State.SUBSCRIBED, State.COMPLETE)) {
```
trustin: Could you add some comments here about why this seemingly infinite loop will never loop infinitely?

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -169,20 +185,24 @@ private void subscribe0(Subscriber<? super T> subscriber, Executor executor, boo
...
+                               new IllegalStateException("subscribed by other subscriber already"));
             return;
         }
 
-        subscribeToDelegate(delegate, newPendingSubscription);
+        for (;;) {
+            if (stateUpdater.compareAndSet(this, State.NONE, State.SUBSCRIBED)) {
+                return;
+            }
+            if (stateUpdater.compareAndSet(this, State.DELEGATED, State.COMPLETE)) {
```
trustin: Could you add some comments here about why this seemingly infinite loop will never loop infinitely?

kojilin:
I can reproduce this by code like below, but it's really rare, sometimes it won't happened even run more that 30mins, sometime reproduces when just running in few seconds.

```
    @Test
    public void tryRace() throws InterruptedException {
        ExecutorService executorService = Executors.newFixedThreadPool(2);
        while (true) {
            StreamMessage<Object> delegate = mock(StreamMessage.class);
            when(delegate.closeFuture()).thenReturn(new CompletableFuture<>());
            Subscriber<Object> subscriber = mock(Subscriber.class);
            DeferredStreamMessage<Object> deferredStreamMessage = new DeferredStreamMessage<>();
            final CountDownLatch latch = new CountDownLatch(1);
            executorService.submit(() -> {
                latch.await();
                Thread.sleep(1);
                deferredStreamMessage.delegate(delegate);
                return null;
            });
            executorService.submit(() -> {
                latch.await();
                Thread.sleep(1);
                deferredStreamMessage.subscribe(subscriber);
                return null;
            });
            latch.countDown();

            await().atMost(5, TimeUnit.SECONDS)
                   .untilAsserted(() -> verify(delegate).subscribe(subscriber, false));
        }
    
```

kojilin:
updated, I'm trying to make test, but not have idea yet.

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -81,10 +93,19 @@ protected void delegate(StreamMessage<T> delegate) {
             }).exceptionally(CompletionActions::log);
         }
 
-        @SuppressWarnings("unchecked")
-        final PendingSubscription<? super T> ps = pendingSubscription;
-        if (ps != null) {
-            subscribeToDelegate(delegate, ps);
+        // There are 4 state in class.
```
trustin: Probably OK to remove this comment?

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -81,10 +93,19 @@ protected void delegate(StreamMessage<T> delegate) {
...
+        // If delegate is called first, then it will match INIT -> DELEGATED.
+        // If subscribe is called first, than it will match SUBSCRIBED -> SUBSCRIBED_AND_DELEGATED.
```
trustin: - subscribe -> subscribe()
- delegate -> delegate()
- then it will match -> the state will transit from ... to ...

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -169,20 +190,29 @@ private void subscribe0(Subscriber<? super T> subscriber, Executor executor, boo
...
-                                   new IllegalStateException("subscribed by other subscriber already"));
-            }
+        if (!subscriptionUpdater.compareAndSet(this, null, newPendingSubscription)) {
+            failLateSubscriber(subscriber, executor,
+                               new IllegalStateException("subscribed by other subscriber already"));
             return;
         }
 
-        subscribeToDelegate(delegate, newPendingSubscription);
+        // There are 4 state in class.
```
trustin: Probably OK to remove this comment?

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -169,20 +190,29 @@ private void subscribe0(Subscriber<? super T> subscriber, Executor executor, boo
...
+        // If subscribe is called first, then it will match INIT -> SUBSCRIBED.
+        // If delegate is called first, than it will match DELEGATED -> SUBSCRIBED_AND_DELEGATED.
```
trustin: - subscribe -> subscribe()
- delegate -> delegate()
- then it will match -> the state will transit from ... to ...

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -81,10 +93,19 @@ protected void delegate(StreamMessage<T> delegate) {
...
-        @SuppressWarnings("unchecked")
-        final PendingSubscription<? super T> ps = pendingSubscription;
-        if (ps != null) {
-            subscribeToDelegate(delegate, ps);
+        // There are 4 state in class.
+        // If delegate is called first, then it will match INIT -> DELEGATED.
+        // If subscribe is called first, than it will match SUBSCRIBED -> SUBSCRIBED_AND_DELEGATED.
+        // It's not possible to enter this for loop with DELEGATED/SUBSCRIBED_AND_DELEGATED state,
+        // because this method doesn't allow to set delegate twice.
+        for (;;) {
```
minwoox: What if this delegate() is called for a while after subscribe() is called?? Then, the thread will be consumed in the loop and cannot do anything.

minwoox: Oops~ my mistake forget about that~

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -203,16 +233,18 @@ private void failLateSubscriber(Subscriber<? super T> subscriber, Executor execu
...
+        final Executor executor = subscription.executor;
+        final boolean withPooledObjects = subscription.withPooledObjects;
         if (executor == null) {
-            delegate.subscribe(subscriber, pendingSubscription.withPooledObjects);
+            delegate.subscribe(subscriber, withPooledObjects);
         } else {
-            delegate.subscribe(subscriber, executor, pendingSubscription.withPooledObjects);
+            delegate.subscribe(subscriber, executor, withPooledObjects);
         }
+        subscriptionUpdater.compareAndSet(this, subscription,
```
minwoox: Do we need this here to update subscription?

minwoox:
I'm really sorry to tell you, but this just crossed my mind.
I think we can handle this without using states just change a little bit of subscribe0() method from the previous code.
<pre><code>
private void subscribe0(Subscriber<? super T> subscriber, Executor executor, boolean withPooledObjects) {
        if (abortPending) {
            failLateSubscriber(subscriber, executor, AbortedStreamException.get());
            return;
        }

        final PendingSubscription<? super T> newPendingSubscription =
                new PendingSubscription<>(subscriber, executor, withPooledObjects);

        if (!pendingSubscriptionUpdater.compareAndSet(this, null, newPendingSubscription)) {
            failLateSubscriber(subscriber, executor,
                               new IllegalStateException("subscribed by other subscriber already"));

        }
        final StreamMessage<T> delegate = this.delegate;
        if (delegate == null) {
            return;
        }

        subscribeToDelegate(delegate, newPendingSubscription);
    }
</code><pre>

and compareAndSet again in subscribeToDelegate().. What do you think about this?

kojilin:
updated, @minwoox thx for ur idea/recommendation.

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -35,6 +35,9 @@
  */
 public class DeferredStreamMessage<T> implements StreamMessage<T> {
 
+    private static final PendingSubscription<Object> NO_OP_PENDING_SUBSCRIPTION =
```
anuraaga: Maybe `NO_OP_ALREADY_SUBSCRIBED` ? As far as I can tell, it's not really pending after this is set, it's more of a marker that there won't be anymore subscriptions.

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -203,15 +207,23 @@ private void failLateSubscriber(Subscriber<? super T> subscriber, Executor execu
         }
     }
 
-    private void subscribeToDelegate(
-            StreamMessage<T> delegate, PendingSubscription<? super T> pendingSubscription) {
+    private void subscribeToDelegate(StreamMessage<T> delegate,
+                                     PendingSubscription<? super T> pendingSubscription) {
+
+        if (pendingSubscription == NO_OP_PENDING_SUBSCRIPTION ||
```
anuraaga: Should these two lines be comparing against `this.pendingSubscription`, not `pendingSubscription`? (renaming the parameter to `newPendingSubscription` would probably make it clearer either way).

trustin:
@imasahiro Any comments?

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
kojilin: not two lines. only first statement ?

```
 if (pendingSubscription == NO_OP_ALREADY_SUBSCRIBED ||
            !pendingSubscriptionUpdater.compareAndSet(this,
                                                      newPendingSubscription,
                                                      NO_OP_ALREADY_SUBSCRIBED)) {
            return;
        }
```

because compareAndSet with this.pendingSubscription will always be true, right?

kojilin:
updated, will 2nd commit make things better? and I found 1st commit forget to return when `!pendingSubscriptionUpdater.compareAndSet(this, null, newPendingSubscription)` . I think it should(?).

kojilin:
And I changed some at 2nd commit, so need recheck. thx.

trustin:
Thanks, @kojilin !

PR #745, opened 2017-09-05
Use NoopMeterRegistry by default (again)

trustin:
Motivation:

Using Metrics.globalRegistry can make Server or ClientFactory attempt to
register meters with duplicate IDs.

Modifications:

- Use NoopMeterRegistry.get() instead of Metrics.globalRegistry

Result:

Server or Client does not fail due to duplicate meter IDs.


trustin:
I just realized this may not be a correct fix. Will update.

PR #746, opened 2017-09-05
Enhance Annotated http service exception handling

kojilin:
* Return status code 400 when annotated http service throwing IllegalArgumentException.
Currently only param check by armeria will be treat as 400 error. But user may want to do some param check with throwing IllegalArgumentException.

* Return correct status code when user code throws HttpResponseException.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
@@ -21,6 +21,7 @@ import static java.util.Objects.requireNonNull;
 import java.util.Formatter;
 import java.util.Locale;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 
@@ -169,9 +170,13 @@ public interface HttpResponse extends Response, StreamMessage<HttpObject> {
     static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
         requireNonNull(stage, "stage");
         final DeferredHttpResponse res = new DeferredHttpResponse();
-        stage.whenComplete((delegate, cause) -> {
-            if (cause != null) {
-                res.close(cause);
+        stage.whenComplete((delegate, thrown) -> {
+            if (thrown != null) {
+                if (thrown instanceof CompletionException && thrown.getCause() != null) {
+                    res.close(thrown.getCause());
+                } else {
+                    res.close(thrown);
+                }
             } else if (delegate == null) {
                 res.close(new NullPointerException("delegate stage produced a null response: " + stage));
             } else {
--- a/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
@@ -21,11 +21,12 @@ import static java.util.Objects.requireNonNull;
 import java.util.Formatter;
 import java.util.Locale;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 
 import org.reactivestreams.Publisher;
+
+import com.google.common.base.Throwables;
 
 import com.linecorp.armeria.common.stream.StreamMessage;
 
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
```diff
@@ -482,6 +485,20 @@ public String post() {
...
+    public static class MyAnnotatedService10 {
+
+        @Get("/syncThrow")
+        public String sync() {
+            throw new IllegalArgumentException("foo");
+        }
+
+        @Get("/asyncThrow")
+        public CompletableFuture<String> async() {
+            throw new IllegalArgumentException("bar");
```
anuraaga: What about CompletableFuture.completedExceptionally?

kojilin: ah.. but then I need to think make it even in `HttpResponse.from(CompletionStage<? extends HttpResponse> stage)` should behave like this too.

trustin: How about adding another method that returns a failed `CompletableFuture` for completeness then?

kojilin:
updated

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -171,7 +173,12 @@ static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
         final DeferredHttpResponse res = new DeferredHttpResponse();
         stage.whenComplete((delegate, cause) -> {
             if (cause != null) {
-                res.close(cause);
+                Throwable rootCause = Throwables.getRootCause(cause);
```
kojilin: CompletableFuture should wrap exception with `CompletionException` . So make HttpResponse + handler can handle HttpResponseException.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpService.java
```diff
@@ -69,15 +71,31 @@ boolean overlaps(AnnotatedHttpService entry) {
 
     @Override
     public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
```
kojilin: move all 400 handling code here.

kojilin:
updated

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpService.java
```diff
@@ -69,15 +72,34 @@ boolean overlaps(AnnotatedHttpService entry) {
...
             return HttpResponse.ofFailure(new IllegalStateException(
                     "illegal return type: " + ret.getClass().getSimpleName()));
         }
 
         @SuppressWarnings("unchecked")
         CompletionStage<HttpResponse> castStage = (CompletionStage<HttpResponse>) ret;
-        return HttpResponse.from(castStage);
+        return HttpResponse.from(castStage.handle((httpResponse, throwable) -> {
+            if (throwable != null) {
+                if (throwable instanceof CompletionException &&
```
kojilin: I'm curious should I use Guava's `getRootCause` or just check 1 depth of cause.

trustin: That's a good question. +1 for using Guava.

trustin: Let me merge once you switch to Guava's `getRootCause()`

kojilin:
updated.

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -169,9 +170,13 @@ static HttpResponse ofFailed(Throwable cause) {
     static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
         requireNonNull(stage, "stage");
         final DeferredHttpResponse res = new DeferredHttpResponse();
-        stage.whenComplete((delegate, cause) -> {
-            if (cause != null) {
-                res.close(cause);
+        stage.whenComplete((delegate, thrown) -> {
+            if (thrown != null) {
+                if (thrown instanceof CompletionException && thrown.getCause() != null) {
```
trustin: Probably better using `Throwables.getRootCause(throwable)` here as well?

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -172,11 +173,7 @@ static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
         final DeferredHttpResponse res = new DeferredHttpResponse();
         stage.whenComplete((delegate, thrown) -> {
             if (thrown != null) {
-                if (thrown instanceof CompletionException && thrown.getCause() != null) {
-                    res.close(thrown.getCause());
-                } else {
-                    res.close(thrown);
-                }
+                res.close(Throwables.getRootCause(thrown));
```
trustin: `getRootCause()` can return `null` and you need to handle it.

kojilin: I think it's implementation will return self if cause is null.

trustin: Sorry, I misread the code. :-)

PR #747, opened 2017-09-05
Make CaffineMetricSupport more tolerant to double registration

trustin:
Motivation:

Currently, CaffeineMetricsSupport raises an IllegalStateException when a
user tries to call setup() with different Cache instances at the same
metric ID. This can result in the failure of Server startup when a user
reuses a registry or uses a Metrics.globalRegistry.

Modifications:

- Improve CaffeineMetricsSupport so that it aggregates the stats of all
  Cache instances instead of rejection

Result:

Less surprise to users

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
@@ -16,12 +16,24 @@
 
 package com.linecorp.armeria.internal.metric;
 
-import static com.google.common.base.Preconditions.checkState;
+import static com.linecorp.armeria.internal.metric.CaffeineMetricSupport.Type.EVICTION_COUNT;
+import static com.linecorp.armeria.internal.metric.CaffeineMetricSupport.Type.EVICTION_WEIGHT;
+import static com.linecorp.armeria.internal.metric.CaffeineMetricSupport.Type.HIT_COUNT;
+import static com.linecorp.armeria.internal.metric.CaffeineMetricSupport.Type.LOAD_FAILURE_COUNT;
+import static com.linecorp.armeria.internal.metric.CaffeineMetricSupport.Type.LOAD_SUCCESS_COUNT;
+import static com.linecorp.armeria.internal.metric.CaffeineMetricSupport.Type.MISS_COUNT;
+import static com.linecorp.armeria.internal.metric.CaffeineMetricSupport.Type.TOTAL_LOAD_TIME;
 import static java.util.Objects.requireNonNull;
 
+import java.lang.ref.WeakReference;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.function.DoubleSupplier;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.ToDoubleFunction;
+
+import javax.annotation.Nullable;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.LoadingCache;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
```diff
@@ -47,63 +59,149 @@ public static void setup(MeterRegistry registry, MeterId id, Cache<?, ?> cache)
...
+                                  func(EVICTION_WEIGHT, ref -> ref.cacheStats.evictionWeight()));
+            parent.gauge(id.name("estimatedSize"), id.tags(), this,
+                         func(null, ref -> ref.estimatedSize));
+        }
+
+        void add(Cache<?, ?> cache, Ticker ticker) {
+            synchronized (cacheRefs) {
+                for (CacheReference ref : cacheRefs) {
+                    if (ref.get() == cache) {
+                        // Do not aggregated more than once for the same instance.
```
minwoox: aggregate?

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
```diff
@@ -47,63 +59,149 @@ public static void setup(MeterRegistry registry, MeterId id, Cache<?, ?> cache)
...
+                                      func(LOAD_SUCCESS_COUNT, ref -> ref.cacheStats.loadSuccessCount()));
+
                 parent.more().counter(loads, id.tags("result", "failure"), this,
-                                      new CacheStatFunction(() -> cacheStats.loadFailureCount()));
+                                      func(LOAD_FAILURE_COUNT, ref -> ref.cacheStats.loadFailureCount()));
 
-                final DoubleSupplier totalLoadTimeSupplier;
-                totalLoadTimeSupplier = () -> cacheStats.totalLoadTime();
+                final ToDoubleFunction<CacheReference> totalLoadTimeSupplier;
+                totalLoadTimeSupplier = ref -> ref.cacheStats.totalLoadTime();
```
minwoox: nit: how about removing totalLoadTimeSupplier? or is this for the readability?

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
trustin: Inlined totalLoadTimeSupplier. It was extracted before to avoid the column limit (Checkstyle)

trustin:
Thanks for quick reviews.

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
```diff
@@ -47,63 +59,145 @@ public static void setup(MeterRegistry registry, MeterId id, Cache<?, ?> cache)
...
+    enum Type {
+        HIT_COUNT,
+        MISS_COUNT,
+        EVICTION_COUNT,
+        EVICTION_WEIGHT,
+        LOAD_SUCCESS_COUNT,
+        LOAD_FAILURE_COUNT,
+        TOTAL_LOAD_TIME;
+
+        static final int count = Type.values().length;
```
anuraaga: Just small note, this was confusing to me at first since some of the enum are also named count.

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
```diff
@@ -47,63 +59,145 @@ public static void setup(MeterRegistry registry, MeterId id, Cache<?, ?> cache)
...
-                                  new CacheStatFunction(() -> cacheStats.evictionCount()));
-            parent.more().counter(id.name("evictionWeight"), id.tags(), this,
-                                  new CacheStatFunction(() -> cacheStats.evictionWeight()));
-            parent.gauge(id.name("estimatedSize"), id.tags(), this,
-                         new CacheStatFunction(() -> estimatedSize));
+        private ToDoubleFunction<CaffeineMetrics> func(@Nullable Type type,
+                                                       ToDoubleFunction<CacheReference> valueFunction) {
+            return value -> {
+                double sum = 0;
+                synchronized (cacheRefs) {
```
anuraaga: Is this called every time the cache metrics are updated? Seems like a lot of overhead on the normal case to satisfy a minor use case.

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
trustin: Any suggestions for a better name? I used length first and then count. Didn't like any :-)

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
trustin: It's only called when metrics are collected. Which shouldn't be very often, like once a minute?

core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java
anuraaga: Since it's only used once, I would just remove the field.

PR #748, opened 2017-09-06
Fix large response handling issue with webapp container services

trustin:
Motivations:

Both TomcatService and JettyService do not handle a large response in
different ways.

TomcatService sets incorrect 'content-length' header when a response is
greater than its buffer size (8KiB).

JettyService fails with an IllegalStateException when ResourceHandler
tries to server a resource greater than 32KiB.

Modifications:

- Add the test cases for various large responses
- Make sure TomcatService and JettyService set the 'conntent-length'
  header correctly
- Do not mark Jetty Request as 'async-supported', because we don't.

Result:

- TomcatService and JettyService handles a large response correctly.


trustin:
@imasahiro This fixes the regression introduced by #722

trustin:
> nit: Could we make a change the same as TomcatService on the following line?

Fixed

testing-internal/src/main/java/com/linecorp/armeria/testing/internal/webapp/WebAppContainerTest.java
```diff
@@ -325,4 +328,28 @@ public void addressesAndPorts_localhost() throws Exception {
...
+    @Test
+    public void largeResponse() throws Exception {
+        testLarge("/jsp/large.jsp");
+    }
+
+    protected void testLarge(String path) throws IOException {
+        try (CloseableHttpClient hc = HttpClients.createMinimal()) {
+            try (CloseableHttpResponse res = hc.execute(new HttpGet(server().uri(path)))) {
+                assertThat(res.getStatusLine().toString()).isEqualTo("HTTP/1.1 200 OK");
+                assertThat(res.getFirstHeader(HttpHeaderNames.CONTENT_TYPE.toString()).getValue())
```
anuraaga: assert content length header?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -487,16 +487,17 @@ private static void convertHeaders(HttpHeaders headers, MimeHeaders cHeaders) {
...
         final String contentType = coyoteRes.getContentType();
         if (contentType != null && !contentType.isEmpty()) {
             headers.set(HttpHeaderNames.CONTENT_TYPE, contentType);
         }
 
-        if (coyoteRes.getContentLengthLong() >= 0) {
-            headers.setLong(HttpHeaderNames.CONTENT_LENGTH, coyoteRes.getContentLengthLong());
+        final String method = coyoteRes.getRequest().method().toString();
+        if (!"HEAD".equals(method)) {
+            headers.setLong(HttpHeaderNames.CONTENT_LENGTH, outputBuffer.getBytesWritten());
```
anuraaga: I found coyote response has getByteswritten as well, is this different? 

https://tomcat.apache.org/tomcat-5.5-doc/catalina/docs/api/org/apache/coyote/Response.html

testing-internal/src/main/java/com/linecorp/armeria/testing/internal/webapp/WebAppContainerTest.java
trustin: Good idea. Added.

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
trustin: Ah, they are same. Simplified!

anuraaga:
By the way, just to clarify, was the problem with Tomcat that it would switch to chunked mode with no content-length header when the response is over the buffer size?

trustin:
> was the problem with Tomcat that it would switch to chunked mode with no content-length header when the response is over the buffer size?

No, the change made by #722 was adding the content-length header with wrong value.

anuraaga:
Is it a bug in Tomcat? I'm surprised `getContentLength` wouldn't return a valid content length, regardless of buffering.

trustin:
I'm surprised as well. `getContentLength()` or `getContentLengthAsLong()` returned the length of the last chunk only.

PR #749, opened 2017-09-06
Rename PrometheusExporterHttpService to PrometheusExpositionService

trustin:
.. because exporter and exposition are different things, as documented
in:

- https://prometheus.io/docs/instrumenting/writing_exporters/
- https://prometheus.io/docs/instrumenting/exposition_formats/


trustin:
Thanks for reviewing!

PR #764, opened 2017-09-25
Work around IndexOutOfBoundsException in WindowSketchQuantiles.get()

trustin:
Related: https://github.com/micrometer-metrics/micrometer/issues/139

Motivation:

Specifying '1.0' as a monitored quantile makes WindowSketchQuantiles.get()
throw an IndexOutOfBoundsException.

Modifications:

- Exclude '1.0' from the monitored quantiles as a workaround

Result:

- No IIOBE


trustin:
On the other hand, it seems to me that using a `LinkedList` doesn't seem like a good idea.

rfkm:
nit; IIOBE -> IOOBE

trustin:
> nit; IIOBE -> IOOBE

Oops. :-)

PR #768, opened 2017-09-30
Switch from WindowSketchQuantiles to CKMSQuantiles

trustin:
.. until the efficienty issues with WindowSketchQuantiles are resolved

See https://github.com/micrometer-metrics/micrometer/issues/139

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
+++ b/core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
@@ -39,8 +39,8 @@ import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.stats.quantile.CKMSQuantiles;
 import io.micrometer.core.instrument.stats.quantile.Quantiles;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/common/metric/MoreMeters.java
```diff
@@ -102,7 +102,13 @@ private static Quantiles defaultQuantiles() {
         // (1) Frugal2U is by far the fastest, but can take a while to converge.
```
anuraaga: What do you think about just defaulting to this fastest one which should converge fine with reasonable traffic? I think it's uncommon to use armeria without multiple server instances, and since quantiles are not aggregated it seems very rare for them to be used in practice as opposed to aggregating in the metrics collector / server. Making the default have the least overhead seems reasonable if the use case is small.

trustin: This is just a temporary fix since `WindowSketchQuantiles` causes the server to struggle. We could switch to `Frugal2U` but I'm afraid I have enough information about 'how soon' it will converge, so I chose a safer one, which is basically same with what Prometheus does.

Micrometer 1.0.0-RC1 introduced percentile approximation to histograms, so we may want to explore that option when we upgrade to a newer Micrometer version: http://micrometer.io/docs/prometheus#_histograms (4.2.2) .. because it seems to bring both percentiles and aggregations.

PR #770, opened 2017-10-08
Use global MeterRegistry if MeterRegistry bean is not present

### LINKED ISSUE #767: Spring boot when default config will Caused by: java.util.NoSuchElementException: No value present
I upgrade 0.53.1 from 0.52 and use  armeria-spring-boot-starter,   but found when default config like
application.properties:
armeria.ports[0].port=8080
armeria.ports[0].protocol=HTTP

An unexpected close has occurred:
Caused by: java.util.NoSuchElementException: No value present

and I debug source code, found if add 

armeria.enableMetrics = false

then normal.

imasahiro:
Results:
- Fix #767


trustin:
Thanks, @imasahiro !

PR #775, opened 2017-10-13
Support Optional<T> type for annotated services

hyangtack:
Modifications:

- Change '@Optional' to '@Default' to avoid conflicting with java.util.Optional<T>
- Support 'Optional<T>' as a type of arguments


core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -467,10 +501,14 @@ ParameterType parameterType() {
             return parameterType;
         }
 
-        boolean isRequired() {
+        boolean required() {
             return required;
         }
 
+        public boolean optionalWrapped() {
```
trustin: It is often better adding `is` for boolean properties because an adjective is sometimes a verb in English. e.g. empty

hyangtack: Thanks! Fixed!

Issue #780, opened 2017-10-20
Client side Thrift callback not notified correctly for a void method with RetryingRpcClient


LiYing2010:
I am using the latest Armeria v0.53.2, but something is going wrong.

When I call a method via Thrift, if this method has not return value, for example:
`public void doSomething(p1, p2)`

The thrift client side will go into deadlock.

I did some research, traced into source code of Armeria, and find these source is not working well:
https://github.com/line/armeria/blob/54e124d73178609b4f0af6d833ea562a1ce15bd9/core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java#L117-L121

If the method has not return value, the "result" will be null, the RPC call is completed normally, but RetryingRpcClient process this case as an error, and it will call the "responseFuture.completeExceptionally(thrown)" branch.

trustin:
Changed the title because it's not a dead lock technically.

PR #781, opened 2017-10-20
Client side Thrift callback not notified correctly for a void method with RetryingRpcClient (#780)

LiYing2010:
See: https://github.com/line/armeria/issues/780

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
@@ -114,9 +114,14 @@ public final class RetryingRpcClient extends RetryingClient<RpcRequest, RpcRespo
                 }
             } else {
                 response.handle(voidFunction((result, thrown) -> {
-                    if (result != null) {
+                    // !!! DO NOT use "result != null" as the success/fail check condition !!!
+                    // !!! because the return value of the RPC target method may be void !!!
+                    // See: https://github.com/line/armeria/issues/780
+                    if (thrown == null) {
+                        // normal response
                         responseFuture.complete(result);
                     } else {
+                        // failed
                         responseFuture.completeExceptionally(thrown);
                     }
                 }));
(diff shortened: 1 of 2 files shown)


imasahiro:
Could you write test code for this on `com.linecorp.armeria.it.client.retry.RetryingRpcClientTest`?
Think you can use `DevNullService.Iface` for testing.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
```diff
@@ -114,9 +114,14 @@ private void doExecute0(ClientRequestContext ctx, RpcRequest req, DefaultRpcResp
                 }
             } else {
                 response.handle(voidFunction((result, thrown) -> {
-                    if (result != null) {
+                    // !!! DO NOT use "result != null" as the success/fail check condition !!!
```
anuraaga: IMO we can remove these comments since handle/whenComplete is a common pattern in the codebase and we wouldn't add this to every single call. Would be very cool if it's possible to make a checkstyle rule for this though.

With the new unit test, there shouldn't be a regression either way, thanks :)

LiYing2010: comments deleted

thrift/src/test/java/com/linecorp/armeria/it/client/retry/RetryingRpcClientTest.java
```diff
@@ -116,4 +124,46 @@ public void execute_reachedMaxAttempts() throws Exception {
...
         verify(serviceHandler, times(1)).hello("hello");
     }
+
+    @Test
+    public void execute_void() throws Exception {
+        RetryStrategy<RpcRequest, RpcResponse> retryStrategy = new MaxTimeRetryStrategy<>(2);
+
+        DevNullService.Iface client = new ClientBuilder(server.uri(BINARY, "/thrift-devnull"))
+                .decorator(RpcRequest.class, RpcResponse.class,
+                           RetryingRpcClient.newDecorator(retryStrategy, 3)
```
imasahiro: Could you reuse `ONLY_HANDLES_EXCEPTION` retry strategy than `MaxTimeRetryStrategy`?

LiYing2010: done

trustin:
Thanks a lot, @LiYing2010 !

PR #795, opened 2017-10-24
Migrate to Brave 4.9.0

### LINKED ISSUE #791: Upgrade Brave to 4.9
Brave 4.9 contains various breaking changes since 4.6.

anuraaga:
Fixes #791

This is a breaking change, users of the old `Span` API will either have to
- Update the official zipkin collector to a version that supports v2 API
- https://github.com/openzipkin/brave/tree/master/brave#zipkin-v1-setup
- Reimplement the reporter if using some non-official API (like a custom Thrift API)

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -90,7 +90,7 @@ io.prometheus:
   simpleclient_common: { version: '0.1.0' }
 
 io.zipkin.brave:
-  brave: { version: &BRAVE_VERSION '4.6.0' }
+  brave: { version: &BRAVE_VERSION '4.9.0' }
 
 it.unimi.dsi:
   fastutil: { version: '8.1.0' }
--- a/zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
+++ b/zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
@@ -35,6 +35,7 @@ import com.linecorp.armeria.client.ClientOptions;
 import com.linecorp.armeria.client.ClientRequestContext;
 import com.linecorp.armeria.client.DefaultClientRequestContext;
 import com.linecorp.armeria.client.Endpoint;
+import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
@@ -50,6 +51,7 @@ import brave.sampler.Sampler;
 import io.netty.channel.Channel;
 import io.netty.channel.DefaultEventLoop;
 import zipkin2.Span;
+import zipkin2.Span.Kind;
 
 public class HttpTracingClientTest {
 
@@ -65,6 +67,9 @@ public class HttpTracingClientTest {
         Span span = reporter.spans().take();
         assertThat(span.name()).isEqualTo(TEST_SPAN);
 
+        // check kind
+        assertThat(span.kind() == Kind.CLIENT);
+
         // only one span should be submitted
         assertThat(reporter.spans().poll(1, TimeUnit.SECONDS)).isNull();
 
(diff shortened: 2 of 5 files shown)


dependencies.yml
```diff
@@ -90,7 +90,7 @@ io.prometheus:
   simpleclient_common: { version: '0.1.0' }
 
 io.zipkin.brave:
-  brave: { version: &BRAVE_VERSION '4.6.0' }
+  brave: { version: &BRAVE_VERSION '4.9.0' }
```
adriancole: ps you can remove the transitive dep on io.zipkin.java:zipkin (as you aren't using it)

zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
```diff
@@ -64,24 +63,24 @@ public void shouldSubmitSpanWhenSampled() throws Exception {
...
-        // check annotation values
-        List<String> values = annotations.stream().map(anno -> anno.value).collect(Collectors.toList());
-        assertThat(values).containsExactlyInAnyOrder("cs", "cr");
+        // check tags
+        assertThat(span.tags()).containsAllEntriesOf(ImmutableMap.of(
+                "http.host", "localhost",
+                "http.method", "POST",
+                "http.path", "/",
+                "http.status_code", "-1",
+                "http.url", "none+h2c://localhost/"));
```
adriancole: interesting scheme!

zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
```diff
@@ -64,24 +63,24 @@ public void shouldSubmitSpanWhenSampled() throws Exception {
...
 
-        // check annotation values
-        List<String> values = annotations.stream().map(anno -> anno.value).collect(Collectors.toList());
-        assertThat(values).containsExactlyInAnyOrder("cs", "cr");
+        // check tags
+        assertThat(span.tags()).containsAllEntriesOf(ImmutableMap.of(
+                "http.host", "localhost",
+                "http.method", "POST",
+                "http.path", "/",
+                "http.status_code", "-1",
```
adriancole: maybe better to not put the status code tag instead of adding an invalid one?

zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
```diff
@@ -64,24 +63,24 @@ public void shouldSubmitSpanWhenSampled() throws Exception {
...
 
-        // check # of annotations
-        List<Annotation> annotations = span.annotations;
-        assertThat(annotations).hasSize(2);
+        // check # of annotations (zipkin2 format does not use them by default)
+        assertThat(span.annotations()).isEmpty();
 
-        // check annotation values
-        List<String> values = annotations.stream().map(anno -> anno.value).collect(Collectors.toList());
-        assertThat(values).containsExactlyInAnyOrder("cs", "cr");
```
adriancole: you can replace this check with span.kind()== Kind.CLIENT

dependencies.yml
anuraaga: Hmm - we are using `zipkin.Constants` / `zipkin.TraceKeys`. Is there a replacement for those in `zipkin2` or should we use something like this to ensure compatibility?

https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/tag/Tags.java

zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
anuraaga: Yeah in practice it wouldn't happen it was a bad mocking in the test. Cleaned it up and some inconsistencies too.

adriancole:
>
> Hmm - we are using zipkin.Constants / zipkin.TraceKeys. Is there a replacement for those in zipkin2 or should we use something like this to ensure compatibility?
>
> There's no replacement constant file, but we can make one if you like.
Notably, many of the constants are invalid (ex core annotations and address
ones), but the trace keys are still valid.

anuraaga:
@adriancole I think having well-defined constants in zipkin2 would be nice. Also, does that mean that `Constants.ERROR` is invalid now? That's the only constant we're using.

adriancole:
>
> @adriancole <https://github.com/adriancole> I think having well-defined
> constants in zipkin2 would be nice. Also, does that mean that
> Constants.ERROR is invalid now? That's the only constant we're using.
>
no "error" is fine.. the "cs" "cr" etc are the ones that are no longer
valid

anuraaga:
Got it - in that case having a file with the `TraceKeys` and `error` would be great :)

trustin:
Thanks a lot for reviewing, @adriancole, and thanks again for the upgrade, @anuraaga! Will review tomorrow. :-)

trustin:
@imasahiro Gentle ping

imasahiro:
Thanks @anuraaga and reviewers!

PR #797, opened 2017-10-25
Make EndpointSelector.select() aware of current request

### LINKED ISSUE #766: Make EndpointSelector.select() aware of current request
Some load-balancing mechanism uses a property of the current request as its input, e.g. the current user's ID and the current request's certain parameter. However, it's not possible to implement such a mechanism with our endpoint API at the moment because `EndpointSelector.select()` does not accept any parameters. I propose to add `ClientRequestContext` to the parameter list of:

- `EndpointSelector.select()`
- `EndpointGroupRegistry.selectNode()`
- `Endpoint.resolve()`

so that a user can inject some information which could be used for load-balancing into `ClientRequestContext` attribute map.

imasahiro:
Fixes #766

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointGroupRegistry.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointGroupRegistry.java
@@ -24,6 +24,7 @@ import java.util.regex.Pattern;
 
 import com.google.common.base.Ascii;
 
+import com.linecorp.armeria.client.ClientRequestContext;
 import com.linecorp.armeria.client.Endpoint;
 
 /**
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointGroupRegistry.java
```diff
@@ -103,16 +104,16 @@ public static EndpointGroup get(String groupName) {
 
     /**
      * Selects an {@link Endpoint} from the {@link EndpointGroup} associated with the specified
-     * case-insensitive {@code groupName}.
+     * {@link ClientRequestContext} and the specified case-insensitive {@code groupName}.
```
minwoox: Can remove `the specified`?

trustin:
@imasahiro We can merge this once @minwoox's comment is addressed.

core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointGroupRegistry.java
imasahiro: Removed

imasahiro:
@anuraaga Thanks for suggestions. I didn't dig it yet but maybe it can be possible if client decorator attaches user-id through e.g. request header? Let me think a bit in next PR.

anuraaga:
Cool - yeah I think it might not affect this API. Using a decorator to move a header into the context seems very heavyweight, but I think propagating through a threadlocal, something like this might be pretty reasonable

```java
try (withContextCustomizer(ctx -> ctx.attrs().set(USER_ID, userId)) {
  client.doRpc(request);
}
```

trustin:

PR #798, opened 2017-10-25
Add a simple benchmark for gRPC.

anuraaga:
It just sends and receives a real-world-esque proto (Github search API) but this should be a reasonable sanity check of the HTTP + marshalling logic, which is the core of gRPC.

Preliminary benchmarks show around
- 750 qps for armeria
- 1100 qps for upstream (with faster JIT warmup from eyeballing the warmup iterations)

It will be interesting to dig into the difference going forward - I don't like being 25% slower!


trustin:
@anuraaga Uh-oh! Let us investigate this for sure. Will review the PR tomorrow.

anuraaga:
I added some benchmarks with empty responses too (the actual payload to the client isn't totally empty though, it still includes message header and 0 for message length).

armeria: ~16K
upstream: ~30K

I put some profiler findings in #799

anuraaga:
Good call - fixed

core/src/jmh/java/com/linecorp/armeria/benchmarks/HttpServerBenchmark.java
```diff
@@ -0,0 +1,82 @@
+package com.linecorp.armeria.benchmarks;
```
trustin: Missing copyright headers.

grpc/build.gradle
```diff
@@ -69,5 +89,13 @@ idea {
         sourceDirs += file("${protobuf.generatedFilesBaseDir}/main/grpc");
         sourceDirs += file("${protobuf.generatedFilesBaseDir}/test/java");
         sourceDirs += file("${protobuf.generatedFilesBaseDir}/test/grpc");
+        sourceDirs += file("${protobuf.generatedFilesBaseDir}/jmh/java");
+        sourceDirs += file("${protobuf.generatedFilesBaseDir}/jmh/grpc");
     }
 }
+
+tasks.checkstyleJmh.enabled = false
```
trustin: Why were JMH classes disabled? Just curious.

anuraaga: Didn't want to deal with javadoc, but realized I could edit the checkstyle config so they're treated as tests, fixed.

trustin:
Thanks, @anuraaga !

imasahiro:
@anuraaga Just curious, how do I reproduce ` grpc/src/jmh/resources/github_search_response.binarypb`?

anuraaga:
@imasahiro I generated it by following the examples for the Github API here - https://developer.github.com/v3/search/

imasahiro:
Thanks for reference! I will check it~

PR #803, opened 2017-10-27
Wait until a pending connection completes before trying to acquire a …

### LINKED ISSUE #802: Armeria opens lots of files in synthetic benchmark
If I create a JMH benchmark which creates a single empty HTTP endpoint and single client that accesses it, and a benchmark which just hits the endpoint without blocking, the server spews exceptions of too many open files (on Linux). Is this expected? My intuition would be that the client and server have a single HTTP/2 connection, so there should only be one socket and therefore open file for this benchmark even though there are many concurrent requests because the benchmark doesn't block for the response between requests.

Exception
```
io.netty.channel.unix.Errors$NativeIoException: accept(..) failed: Too many open files
        at io.netty.channel.unix.Errors.newIOException(Errors.java:117)
        at io.netty.channel.unix.Socket.accept(Socket.java:314)
        at io.netty.channel.epoll.AbstractEpollServerChannel$EpollServerSocketUnsafe.epollInReady(AbstractEpollServerChannel.java:112)
        at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:404)
        at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:304)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
        at java.lang.Thread.run(Thread.java:748)
```

The benchmark code

https://github.com/anuraaga/armeria/blob/dev_rag2/core/src/jmh/java/com/linecorp/armeria/benchmarks/HttpServerBenchmark.java#L71

anuraaga:
…channel for the same key.

Fixes #802

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
+++ b/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
@@ -18,6 +18,7 @@ package com.linecorp.armeria.client.pool;
 import static java.util.Objects.requireNonNull;
 
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -55,6 +56,7 @@ public class DefaultKeyedChannelPool<K> implements KeyedChannelPool<K> {
     private final boolean healthCheckOnRelease;
 
     private final Map<K, Deque<Channel>> pool;
+    private final Map<K, Future<Channel>> pendingConnections;
 
     /**
      * Creates a new instance.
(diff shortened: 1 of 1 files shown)


anuraaga:
Admittedly this is somewhat of a strawman's proposal since I'm not too familiar with the code, but it seems to reason correctly to me, and it fixed the benchmark I was having troubles with.

I used a `HashMap` instead of `ConcurrentHashMap` since as far as I can tell calls are always synchronized in the event loop.

trustin:
> I used a `HashMap` instead of `ConcurrentHashMap` since as far as I can tell calls are always synchronized in the event loop.

Yes, future listeners are always invoked from the event loop thread, so it's safe to use `HashMap`.

trustin:
Thanks a lot, @anuraaga !

PR #807, opened 2017-10-30
Customize ClientRequestContext when creating requests

### LINKED ISSUE #805: ClientRequestContext should be customizable when creating requests.
Followup to #766, as discussed sometimes it's not enough to be able to load-balance based on request information, for example, requests should be balanced based on user-id but the request should contain an encrypted user token rather than the raw id for security purposes even for internal services.

One idea I had is propagating through a `ThreadLocal` - does this seem like a reasonable API?

```java
try (withContextCustomizer(ctx -> ctx.attrs().set(USER_ID, userId)) {
  client.doRpc(request);
}
```

imasahiro:
Fixes #805

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Clients.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Clients.java
@@ -15,18 +15,23 @@
  */
 package com.linecorp.armeria.client;
 
+import static com.linecorp.armeria.client.ClientRequestContext.HTTP_HEADERS;
+import static com.linecorp.armeria.client.DefaultClientRequestContext.THREAD_LOCAL_CONTEXT_CUSTOMIZER;
 import static java.util.Objects.requireNonNull;
 
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Proxy;
 import java.net.URI;
 import java.util.Optional;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
+import com.linecorp.armeria.common.DefaultHttpHeaders;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.util.SafeCloseable;
 
 import io.netty.util.AsciiString;
+import io.netty.util.Attribute;
 
 /**
  * Creates a new client that connects to a specified {@link URI}.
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/Clients.java
```diff
@@ -325,16 +329,52 @@ public static SafeCloseable withHttpHeader(AsciiString name, String value) {
...
+        return withContextCustomizer(ctx -> {
+            final Attribute<HttpHeaders> attr = ctx.attr(HTTP_HEADERS);
+            final HttpHeaders headers = attr.get();
+            attr.set(headerManipulator.apply(headers != null ? headers : new DefaultHttpHeaders()));
+        });
+    }
+
+    /**
+     * Sets the specified {@link ClientRequestContext} customization function in a thread-local variable so that
+     * the customized context are used when the client invokes a request from the current thread. Use the
```
anuraaga: is used

core/src/main/java/com/linecorp/armeria/client/Clients.java
```diff
@@ -325,16 +329,52 @@ public static SafeCloseable withHttpHeader(AsciiString name, String value) {
...
+     *     ctx.attr(USER_ID).set(userId);
+     *     ctx.attr(USER_PASSWORD).set(userPassword);
+     * }) {
+     *     client.executeSomething(..);
+     * }
+     * }</pre>
+     * You can also nest the request context customization:
+     * <pre>{@code
+     * try (SafeCloseable ignored = withContextCustomizer(ctx -> ctx.attr(USER_ID).set(userId)) {
+     *     for (String secret : secrets) {
```
anuraaga: I think the example makes more sense if you replace this with

```client.getSecret```

File: core/src/main/java/com/linecorp/armeria/client/UserClient.java [hunk 3]
```diff
@@ -163,22 +161,19 @@ protected final O execute(@Nullable EventLoop eventLoop,
         }
 
         try (SafeCloseable ignored = RequestContext.push(ctx)) {
```
anuraaga: I wasn't able to use UserClient in gRPC since it wasn't designed for streaming.

Can you update this, or maybe we can move to a request context constructor?

https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaChannel.java

imasahiro:
Yeah, I just pushed javadoc fix and am working on gRPC integration~

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -87,6 +90,15 @@ public DefaultClientRequestContext(
             headersCopy.set(headers);
             attr(HTTP_HEADERS).set(headersCopy);
         }
+        runThreadLocalContextCustomizer();
+    }
+
+    private void runThreadLocalContextCustomizer() {
+        final Consumer<ClientRequestContext> customizer = THREAD_LOCAL_CONTEXT_CUSTOMIZER.get();
+        if (customizer == null) {
```
anuraaga: Nit - != would remove one line of code

trustin:
Thanks, @imasahiro !

PR #808, opened 2017-10-30
Improve JMH integration

trustin:
- Apply JMH plugin for all Java projects
- Add JMH annotation processor to the dependencies so that a developer
  can launch a benchmark from his or her IDE
- Add 'jmh.*' properties for tuning JMH plugin configuration
  - Rename grpcJmh.upstream to jmh.grpcUpstream

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,16 +8,16 @@ buildscript {
     repositories {
         mavenCentral()
         jcenter()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
     }
     dependencies {
         classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.1'
         classpath 'com.google.gradle:osdetector-gradle-plugin:1.4.0'
+        classpath 'me.champeau.gradle:jmh-gradle-plugin:0.4.4'
         classpath 'org.yaml:snakeyaml:1.18'
     }
-}
-
-plugins {
-    id 'me.champeau.gradle.jmh' version '0.4.4' apply false
 }
 
 apply plugin: 'com.google.osdetector'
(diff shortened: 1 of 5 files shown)


trustin:
On the other hand, I was able to use https://github.com/trustin/perf-map-agent/ to resolve the symbols of Java application. However, I had to remove it from this PR since it's GPL'd.

After then, I realized I don't need `perf-map-agent` but just install the HSDIS plugin for OpenJDK: https://github.com/abak/openjdk-hsdis

trustin:
Ended up making some additional modifications to get it work on my Linux: https://github.com/trustin/openjdk-hsdis

trustin:
It seems like the annotation processor and the gradle plugin do not get along. Will investigate a little bit more..

trustin:
Removed the annotation processor because it causes double-generation of benchmark classes (one from IDE and the other from the plugin), making things go unexpectedly.

trustin:
Build failure unrelated with this PR. Filed a separate issue: #809

anuraaga:
FYI I have found that since Gradle 4.2, startup is blazingly fast, so I have set IntelliJ to Delegate builds to Gradle, which usually handles annotation processors smoothly.

Otherwise for Gradle / IntelliJ interop I was using these hacks but have their own problems, as does gradle-apt-plugin's approach too...

https://github.com/curioswitch/curiostack/blob/master/tools/gradle-plugins/gradle-curiostack-plugin/src/main/java/org/curioswitch/gradle/plugins/curiostack/CuriostackPlugin.java#L374

https://github.com/curioswitch/curiostack/blob/master/tools/gradle-plugins/gradle-curiostack-plugin/src/main/java/org/curioswitch/gradle/plugins/curiostack/CuriostackPlugin.java#L445

Issue #811, opened 2017-10-31
Zipkin trace doesn't open/close span correctly?


kojilin:
I found that sometimes my code has unpredictable `currentSpan` in some thread. I suspect that there is nested makeContextAware and cause unbalanced zipkin span scope open/close. Should we implement SpanContextUtil like below?

```
    public static void setupContext(RequestContext ctx, Span span, Tracer tracer) {
        ctx.onEnter(unused -> {
            Deque<SpanInScope> queue = ctx.attr(SPAN_IN_SCOPE_KEY).get();
            if (queue == null) {
                queue = new ArrayDeque<>();
                ctx.attr(SPAN_IN_SCOPE_KEY).set(queue);
            }
            queue.push(tracer.withSpanInScope(span));
        });
        ctx.onExit(unused -> {
            Deque<SpanInScope> spanInScopes = ctx.attr(SPAN_IN_SCOPE_KEY).get();
            if (spanInScopes != null) {
                SpanInScope currentSpan = spanInScopes.pop();
                if (currentSpan != null) {
                    currentSpan.close();
                }
            }
        });
    }
```
If onEnter & onExit will be called in pairs, we don't need to do null check.(and it should, right?)

kojilin:
cc @trustin & @anuraaga

anuraaga:
Hmm - I'm curious how you can end up with a double `makeContextAware` it seems like where a thread-switch was expected, there wasn't one.

Either way it seems reasonable to make `RequestContext` reentrant. I don't think we need the overhead of the queue, can just use a counter to track the enters and exits and only close on the last exit. It probably makes sense to handle this within `RequestContext` itself by adding `RequestContext.onFirstEnter / onLastExit`.

Another option would be to actually not allow `RequestContext` to be reentrant and make it an `IllegalStateException` to try to do so. We could add a reference to the parent to any child contexts, and traverse up during the context push to see if it's already been pushed. I can't think of a situation where nested `makeContextAware` would actually be necessary and think it's always a programming bug, where the user doesn't understand what thread some code is running on, so making it easy to find these by throwing seems ok to me too, but either option would work.

WDYT @trustin @kojilin ?

kojilin:
you are right, let me check where I make it have onEnter twice.

kojilin:
asked as slack, but add it here too.
If we submit runnable to non armeria executor with makeContextAware, `ctx.attr` will be overwritten?

anuraaga:
[...]

PR #812, opened 2017-11-01
Fix zipkin span not correctly closed problem

kojilin:
Apply this change will fix concurrent problem of SpanScope.
It's possible that RequestContext.push be called concurrently. So context's attr may be overwrite.
Change to use thread local to make scope bind on thread, then we can retrieve it back.

Fix issue https://github.com/line/armeria/issues/811

- [x] Add test with this case.

### DIFF
--- a/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
+++ b/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
@@ -21,25 +21,25 @@ import com.linecorp.armeria.common.RequestContext;
 import brave.Span;
 import brave.Tracer;
 import brave.Tracer.SpanInScope;
-import io.netty.util.AttributeKey;
+import io.netty.util.concurrent.FastThreadLocal;
 
 /**
  * Utility for pushing and popping a {@link Span} via a {@link RequestContext}.
  */
 public final class SpanContextUtil {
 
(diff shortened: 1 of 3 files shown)


zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
```diff
@@ -21,25 +21,24 @@
...
      */
     public static void setupContext(RequestContext ctx, Span span, Tracer tracer) {
-        ctx.onEnter(unused -> ctx.attr(SPAN_IN_SCOPE_KEY).set(tracer.withSpanInScope(span)));
+        ctx.onEnter(unused -> SPAN_IN_THREAD.set(tracer.withSpanInScope(span)));
         ctx.onExit(unused -> {
-            SpanInScope currentSpan = ctx.attr(SPAN_IN_SCOPE_KEY).getAndSet(null);
-            if (currentSpan != null) {
-                currentSpan.close();
+            SpanInScope spanInScope = SPAN_IN_THREAD.get();
+            if (spanInScope != null) {
```
anuraaga: Adding `SPAN_IN_THREAD.remove()` for safety seems good.

kojilin:
Updated, when old util  with `CurrentTraceContext.Default.create()` will failed the test because there is unclosed span.

And to prevent unexpected span in thread, change to `currentTraceContext(CurrentTraceContext.Default.create())` is good in application I think.

trustin:
Thanks, @kojilin !

PR #813, opened 2017-11-01
Add link header to HttpHeaderNames

minwoox:
Used to express a typed relationship with another resource, where the relation type is defined by RFC 5988


trustin:
Thanks, @minwoox !

Issue #799, opened 2017-10-26
Armeria gRPC performance analysis


anuraaga:
I noticed some reasonably large differences in performance between armeria-grpc and upstream grpc on the same business logic and did some profiling - want to capture them and maybe start an investigation. My initial suspicion is flushing on every write?

I've attached VisualVm profiles, and here's a screenshot of hot spots in the armeria version. I don't believe `writevAddress`, `readAddress`, `invokeFlush0` are supposed to be so high here.

![image](https://user-images.githubusercontent.com/198344/32044127-a1e035be-ba77-11e7-957c-c8b92d00c56f.png)

[visualvm-profiles.tar.gz](https://github.com/line/armeria/files/1417584/visualvm-profiles.tar.gz)

anuraaga:
perfasm profiles (presumably I need to understand its flags to get better information but for posterity)

armeria
```
....[Hottest Regions]...............................................................................
  2.75%    1.29%                                                  libc-2.24.so  malloc_consolidate (303 bytes) 
  2.26%    3.08%                                                  libc-2.24.so  _int_free (251 bytes) 
  1.81%    1.30%                                                     libjvm.so  PhaseIdealLoop::is_dominator (50 bytes) 
  1.60%    2.97%                                                  libc-2.24.so  malloc (110 bytes) 
  1.18%    1.75%                                                  libc-2.24.so  malloc (106 bytes) 
  1.15%    1.45%                                              libfcml.so.0.0.0  fcml_fn_dt_dts_tree_element_free (102 bytes) 
  1.09%    1.41%                                                  libc-2.24.so  _int_malloc (557 bytes) 
  0.99%    0.15%                                             [kernel.kallsyms]  entry_SYSCALL_64 (200 bytes) 
  0.82%    0.54%                                             [kernel.kallsyms]  _raw_spin_lock_irqsave (45 bytes) 
  0.79%    2.35%                                                   C2, level 4  java.util.regex.Pattern$Start::match, version 2533 (469 bytes) 
  0.71%    0.49%                                             [kernel.kallsyms]  copy_user_enhanced_fast_string (12 bytes) 
  0.67%    0.12%                                             [kernel.kallsyms]  mutex_unlock (24 bytes) 
  0.65%    0.27%                                             [kernel.kallsyms]  mutex_lock (48 bytes) 
  0.59%    0.40%                                             [kernel.kallsyms]  __fget (92 bytes) 
[...]

PR #815, opened 2017-11-03
Add okhttp client and non-blocking gRPC benchmarks.

anuraaga:
Also refactors more common logic into a base class.

~~While the asynchronous version of the upstream benchmark seemed to be showing awesome performance (2-4x downstream), it doesn't complete since there seems to be a memory leak...~~
After modifying the benchmark to wait for pending requests before moving on, it works. Guess the QPS was high enough to overload the server through multiple iterations.

For #799

### DIFF
--- a/grpc/src/jmh/java/com/linecorp/armeria/grpc/downstream/DownstreamSimpleBenchmark.java
+++ b/grpc/src/jmh/java/com/linecorp/armeria/grpc/downstream/DownstreamSimpleBenchmark.java
@@ -17,58 +17,68 @@
 package com.linecorp.armeria.grpc.downstream;
 
 import static com.linecorp.armeria.common.SessionProtocol.HTTP;
-import static com.linecorp.armeria.grpc.shared.GithubApiService.SEARCH_RESPONSE;
 
-import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.Scope;
-import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
-import org.openjdk.jmh.annotations.TearDown;
-import org.openjdk.jmh.infra.Blackhole;
-
-import com.google.protobuf.Empty;
 
 import com.linecorp.armeria.benchmarks.GithubServiceGrpc.GithubServiceBlockingStub;
+import com.linecorp.armeria.benchmarks.GithubServiceGrpc.GithubServiceFutureStub;
 import com.linecorp.armeria.client.Clients;
 import com.linecorp.armeria.grpc.shared.GithubApiService;
+import com.linecorp.armeria.grpc.shared.SimpleBenchmarkBase;
 import com.linecorp.armeria.server.Server;
 import com.linecorp.armeria.server.ServerBuilder;
 import com.linecorp.armeria.server.ServerPort;
 import com.linecorp.armeria.server.grpc.GrpcServiceBuilder;
 
 @State(Scope.Benchmark)
-public class DownstreamSimpleBenchmark {
+public class DownstreamSimpleBenchmark extends SimpleBenchmarkBase {
 
     private Server server;
     private GithubServiceBlockingStub githubApiClient;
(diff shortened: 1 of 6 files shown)


grpc/src/jmh/java/com/linecorp/armeria/grpc/shared/SimpleBenchmarkBase.java
```diff
@@ -0,0 +1,195 @@
...
+     * Benchmark teardown logic.
+     */
+    protected abstract void tearDown() throws Exception;
+
+    private ManagedChannel okhttpChannel;
+    private GithubServiceBlockingStub githubApiOkhttpClient;
+    private GithubServiceFutureStub githubApiOkhttpFutureClient;
+
+    @Param
+    public ClientType clientType = ClientType.NORMAL;
```
trustin: - Can be private
- No need to set the initial value

grpc/src/jmh/java/com/linecorp/armeria/grpc/shared/SimpleBenchmarkBase.java
```diff
@@ -0,0 +1,195 @@
...
+            waiting = false;
+            numRequests.set(0);
+            numFailures.set(0);
+            currentRequests.set(0);
+        }
+
+        @TearDown(Level.Iteration)
+        public void waitForCurrentRequests() {
+            waiting = true;
+            await().forever().until(() -> currentRequests.get() == 0);
```
trustin: Question: does JMH take the changes during iteration teardown into account?

settings/checkstyle/checkstyle-suppressions.xml
```diff
@@ -17,4 +17,6 @@
   <suppress checks=".*" files="[\\/]gen-java[\\/]" />
   <!-- Suppress checks in large forks to make diffing against upstream easier -->
   <suppress checks=".*" files="[\\/]grpc[\\/]ServerCallImpl.java$" />
+  <!-- Suppress visibility checks for JMH which requires public fields, etc -->
+  <suppress checks="VisibilityModifier" files="[\\/](jmh)[\\/]" />
```
trustin: I found annotating `private` field with `@Param` just works and thus this suppression is unnecessary.

grpc/src/jmh/java/com/linecorp/armeria/grpc/shared/SimpleBenchmarkBase.java
anuraaga: Indeed it does - it seems to complain about private on everything except for fields, that's surprising...

grpc/src/jmh/java/com/linecorp/armeria/grpc/shared/SimpleBenchmarkBase.java
anuraaga: JMH returns the value of the counters at the end of teardown, so while it doesn't add the time of the teardown to the run, the counters would be affected. It's why I added the `waiting` guard.

grpc/src/jmh/java/com/linecorp/armeria/grpc/downstream/DownstreamSimpleBenchmark.java
```diff
@@ -17,58 +17,67 @@
...
+        server.stop().join();
     }
+
+    /*public static void main(String[] args) throws Exception {
+        DownstreamSimpleBenchmark benchmark = new DownstreamSimpleBenchmark();
+        benchmark.start();
+        System.out.println(benchmark.simple());
+        System.out.println(benchmark.empty());
+        benchmark.stop();
+    }*/
```
trustin: Which would be better? `/* .. */` or `//`?

anuraaga: Switched to line comment

trustin:
Thanks, @anuraaga!

PR #819, opened 2017-11-04
Make endpoint selection asynchronous

imasahiro:


imasahiro:
I want to think about changing endpoint selection API to make asynchronous because want to select endpoint using in comming http request body. But not sure it is useful to other endpoint selection strategies. WDYT?  @huydx

core/src/test/java/com/linecorp/armeria/client/endpoint/RoundRobinStrategyTest.java
```diff
@@ -49,20 +49,20 @@ public void setup() {
     }
 
     @Test
-    public void select() {
-        assertThat(EndpointGroupRegistry.selectNode(ctx, "endpoint"))
+    public void select() throws Exception {
+        assertThat(EndpointGroupRegistry.selectNode(ctx, "endpoint").get())
```
trustin: https://github.com/joel-costigliola/assertj-examples/blob/master/assertions-examples/src/test/java/org/assertj/examples/CompletableFutureAssertionsExamples.java

core/src/test/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategyTest.java
```diff
@@ -55,14 +56,14 @@ public void setup() {
...
+        assertThat(selectNode(ctx, "endpoint")).isNotNull();
 
-        assertThat(catchThrowable(() -> EndpointGroupRegistry.selectNode(ctx, "empty")))
+        assertThat(catchThrowable(() -> selectNode(ctx, "empty")))
                 .isInstanceOf(EndpointGroupException.class);
     }
 
     @Test
-    public void testRoundRobinSelect() {
+    public void testRoundRobinSelect() throws Exception {
```
trustin: You don't need to add `throws Exception` by using `.join()` instead of `.get()`

imasahiro: Fixed

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -123,11 +125,11 @@ public boolean isGroup() {
      * @return the {@link Endpoint} resolved by {@link EndpointGroupRegistry}.
      *         {@code this} if this endpoint is already a host endpoint.
```
minwoox: Does this need to be updated?

huydx:
LGTM, thanks for the effort @imasahiro !

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
trustin: What do you think, @imasahiro ?

trustin: Nevermind, it seems like the body of the Javadoc has been updated already.

trustin:
@hyangtack Gentle ping

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
imasahiro: Ah, forgot to add @return section. Updated.

trustin:
Thanks, @imasahiro !

trustin:
Reverted by #845 
Related discussion #833

PR #820, opened 2017-11-05
Add a first pass at matching gRPC content types without parsing, whic…

anuraaga:
…h should almost always be enough.

See #799


trustin:
Seems like Travis CI failure is unrelated with this PR.

trustin:
Thanks, @anuraaga !

PR #824, opened 2017-11-08
Add a decorator for annotated service methods

hyangtack:
Motivation:
Currently there is no way to decorate a single annotated service method when an annotated service object is added by ServerBuilder#annotatedService method.

Modifications:
- Add `@Decorate` annotation to specify decorators for an annotated service method
- Use DecoratingServiceFunction and FunctionalDecoratingService to decorate the method

Result:
A user can use a decorator easily by specifying annotations over annotated service method implementations.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
@@ -22,6 +22,7 @@ import static com.linecorp.armeria.internal.ArmeriaHttpUtil.concatPaths;
 import static com.linecorp.armeria.server.AbstractPathMapping.ensureAbsolutePath;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
@@ -31,6 +32,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
@@ -42,12 +46,15 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 
 import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.MediaType;
 import com.linecorp.armeria.internal.DefaultValues;
 import com.linecorp.armeria.server.annotation.ConsumeType;
 import com.linecorp.armeria.server.annotation.ConsumeTypes;
 import com.linecorp.armeria.server.annotation.Converter;
 import com.linecorp.armeria.server.annotation.Converter.Unspecified;
+import com.linecorp.armeria.server.annotation.Decorate;
 import com.linecorp.armeria.server.annotation.Delete;
 import com.linecorp.armeria.server.annotation.Get;
 import com.linecorp.armeria.server.annotation.Head;
@@ -67,6 +74,14 @@ import com.linecorp.armeria.server.annotation.Trace;
  */
 final class AnnotatedHttpServices {
     private static final Logger logger = LoggerFactory.getLogger(AnnotatedHttpServices.class);
+
+    /**
+     * A {@link DecoratingServiceFunction} map for reusing.
+     */
+    private static final ConcurrentMap<
+            Class<? extends DecoratingServiceFunction<HttpRequest, HttpResponse>>,
+            DecoratingServiceFunction<HttpRequest, HttpResponse>>
+            decoratingServiceFunctions = new ConcurrentHashMap<>();
 
     /**
      * Mapping from HTTP method annotation to {@link HttpMethod}, like following.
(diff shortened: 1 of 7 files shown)


hyangtack:
@trustin @minwoox @kojilin 
This is WIP and I need to add more tests, but I want to get your comments for proof of concept. Please feel free to leave comments. 😃

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -69,6 +75,14 @@
     private static final Logger logger = LoggerFactory.getLogger(AnnotatedHttpServices.class);
 
     /**
+     * A {@link DecoratingServiceFunction} map for reusing.
+     */
+    private static final ConcurrentMap<
+            Class<? extends DecoratingServiceFunction<HttpRequest, HttpResponse>>,
+            DecoratingServiceFunction<HttpRequest, HttpResponse>>
+            DECORATING_SERVICE_FUNCTIONS = new ConcurrentHashMap<>();
```
trustin: I guess we don't need use a uppercase name here because it's not a constant but mutable state.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -326,6 +340,48 @@ private static ResponseConverter converter(Method method) {
...
+     * Returns a new decorator which decorates a {@link Service} by {@link FunctionalDecoratingService}
+     * and the specified {@link DecoratingServiceFunction}.
+     */
+    private static Function<Service<HttpRequest, HttpResponse>,
+            ? extends Service<HttpRequest, HttpResponse>> newDecorator(
+            Class<? extends DecoratingServiceFunction<HttpRequest, HttpResponse>> clazz) {
+        return service -> new FunctionalDecoratingService<>(
+                service, DECORATING_SERVICE_FUNCTIONS.computeIfAbsent(clazz, newClazz -> {
+            try {
+                return newClazz.newInstance();
```
trustin: - Perhaps we could try to call `setAccessible(true)` on the constructor?
- How about also trying to call `setAccessible(true)` on all annotated service methods, so a user can use an anonymous class like this:

  ```java
  sb.annotatedService(new Object() {
      // Fails even if hello() is public.
      @Get("/")
      public AggregatedHttpMessage hello() {
          return AggregatedHttpMessage.of(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "Hello!");
      }
  });
  ```

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceDecorationTest.java
```diff
@@ -0,0 +1,137 @@
...
+import com.linecorp.armeria.server.TestConverters.UnformattedStringConverter;
+import com.linecorp.armeria.server.annotation.Converter;
+import com.linecorp.armeria.server.annotation.Decorate;
+import com.linecorp.armeria.server.annotation.Get;
+import com.linecorp.armeria.server.logging.LoggingService;
+import com.linecorp.armeria.testing.server.ServerRule;
+
+public class AnnotatedHttpServiceDecorationTest {
+
+    private static final ClientFactory CLIENT_FACTORY = ClientFactory.DEFAULT;
```
trustin: Seems unused?

kojilin:
LGTM, just curious any idea of decorator that needs instantiate by user?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -97,7 +112,7 @@
      * Returns the list of {@link Path} annotated methods.
      */
     private static List<Method> requestMappingMethods(Object object) {
-        return Arrays.stream(object.getClass().getMethods())
+        return Arrays.stream(object.getClass().getDeclaredMethods())
```
trustin: Need to climb up the class hierarchy because `getDeclaredMethods()` excludes the inherited methods.

core/src/main/java/com/linecorp/armeria/server/annotation/Decorate.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Specifies {@link DecoratingServiceFunction} classes which handle a {@link HttpRequest} before invoking
+ * an annotated service method.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Decorate {
+
+    /**
+     * An array of the {@link DecoratingServiceFunction} classes.
```
trustin: Could you specify that the class will be instantiated using its default constructor?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -307,25 +318,69 @@ private static ResponseConverter converter(Method method) {
...
+    /**
+     * Returns a new instance of the specified {@link Class}.
+     */
+    private static <T> T newInstance(Class<? extends T> clazz) {
+        try {
+            final Constructor<? extends T> constructor = clazz.getDeclaredConstructor();
+            constructor.setAccessible(true);
+            return constructor.newInstance();
+        } catch (Exception e) {
+            throw new IllegalStateException(e);
```
trustin: How about adding a message to the `IllegalStateException`, e.g.

    throw new IllegalStateException("A decorator function class specified in @" + 
                                    Decorate.class.getSimpleName() + 
                                    " annotation must have an accessible default constructor:" +
                                    clazz.getName(), e);

..?

hyangtack:
> LGTM, just curious any idea of decorator that needs instantiate by user?

@kojilin I'm not sure how to do that for now. I'll think about that. Please let me know if you have an idea. 😄

trustin:
Thanks, @hyangtack !

PR #825, opened 2017-11-09
Provide a way to inject HTTP request headers to annotated services

### LINKED ISSUE #708: Annotated service: request header injection
... like spring `@RequestHeader` annotation.

minwoox:
Motivation:
There's no way to get the value of the specific HTTP request header using annotated services.

Modifications:
- Add the `Header` annotation to represent a HTTP request header in annotated services.

Result:
- Can close #708
- Can implement the watcher in Central Dogma!!!

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
@@ -50,12 +50,14 @@ import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.armeria.internal.Types;
 import com.linecorp.armeria.server.annotation.Default;
+import com.linecorp.armeria.server.annotation.Header;
 import com.linecorp.armeria.server.annotation.Param;
 import com.linecorp.armeria.server.annotation.ResponseConverter;
 
 import io.netty.handler.codec.http.HttpConstants;
 import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.QueryStringDecoder;
+import io.netty.util.AsciiString;
 
 /**
  * Invokes an individual method of an annotated service. An annotated service method whose return type is not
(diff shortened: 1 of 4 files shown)


minwoox:
Plz do not review yet.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -321,12 +340,33 @@ private static String httpParameterValue(HttpParameters httpParameters, Paramete
...
             // May return null if no default value is specified.
             return entry.defaultValue();
+        } else {
+            throw new IllegalArgumentException(
+                    "Required parameter '" + entry.name() + "' is missing.");
         }
+    }
+
+    private static Object convertWithOptional(String value, Parameter entry) {
+        Object converted = value != null ? convertParameter(value, entry.type()) : null;
```
hyangtack: nit: final?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -321,12 +340,33 @@ private static String httpParameterValue(HttpParameters httpParameters, Paramete
...
+        return entry.isOptionalWrapped() ? Optional.ofNullable(converted) : converted;
+    }
 
-        throw new IllegalArgumentException("Required parameter '" + entry.name() + "' is missing.");
+    private static String httpHeaderValue(Parameter entry, HttpRequest req) {
+        final String name = entry.name();
+        assert name != null;
+        String value = req.headers().get(AsciiString.of(name));
```
hyangtack: nit: final?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -242,10 +259,12 @@ private Object invoke(ServiceRequestContext ctx, HttpRequest req, @Nullable Aggr
                         httpParameters = httpParametersOf(ctx, req, message);
                     }
                     value = httpParameterValue(httpParameters, entry);
-                    final Object converted = value != null ? convertParameter(value, entry.type())
-                                                           : null;
-                    values[i] = entry.isOptionalWrapped() ? Optional.ofNullable(converted)
-                                                          : converted;
+                    values[i] = convertWithOptional(value, entry);
+
```
hyangtack: nit: Could we remove this blank line?

minwoox:
It is ready.!

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -161,65 +163,79 @@ private Object invoke(ServiceRequestContext ctx, HttpRequest req, @Nullable Aggr
...
-                        try {
-                            type = (Class<?>) ((ParameterizedType) p.getParameterizedType())
-                                    .getActualTypeArguments()[0];
-                        } catch (Exception e) {
-                            throw new IllegalArgumentException("Invalid optional parameter: " + p.getName(), e);
-                        }
-                    } else {
-                        type = p.getType();
-                    }
+                entries.add(createParamOrHeaderParam(parameterInfo, ParameterType.PARAM,
```
hyangtack: nit: How about moving out the code for making a path variable `Parameter` from `createParamOrHeaderParam` method so that we can let the method make an optional supported `Parameter` only? There may be some code duplication, but it would be more understandable. WDYT?

```
if (pathParams.contains(param.value())) {
    if (parameterInfo.getAnnotation(Default.class) != null ||
        parameterInfo.getType().isAssignableFrom(Optional.class)) {
        throw new IllegalArgumentException(
                "Path variable '" + param.value() + "' should not be an optional.");
    }
    entries.add(Parameter.ofPathParam(
            validateAndNormalizeSupportedType(parameterInfo.getType()), param.value()));
} else {
    entries.add(createParameter(parameterInfo, ParameterType.PARAM, param.value()));
}
continue;
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -161,65 +163,79 @@ private Object invoke(ServiceRequestContext ctx, HttpRequest req, @Nullable Aggr
...
             }
 
-            throw new IllegalArgumentException("Unsupported object type: " + p.getType());
+            throw new IllegalArgumentException("Unsupported object type: " + parameterInfo.getType());
         }
         return entries.build();
     }
 
+    private static Parameter createParamOrHeaderParam(java.lang.reflect.Parameter parameterInfo,
+                                                      ParameterType paramType, final Set<String> pathParams,
```
hyangtack: nit: As far as I know, we don't usually put `final` keyword on a method parameter.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
minwoox: That is a good idea~! Changed it.

trustin:
@minwoox Could you resolve the conflict?
@kojilin Care to review?

minwoox:
The confliction is resolved.!

trustin:
Thanks, @minwoox !

PR #826, opened 2017-11-11
Move all benchmarks into separate benchmarks project for simpler build…

anuraaga:
… configuration.

I was having a weird error when running `:core:jmh`, related to `testing-internal`. I noticed most projects consolidate their benchmarks into one project anyways, and it seems like a good pattern so I've done it here. It will allow, e.g. sharing `AuxCounters` among benchmarks.


grpc/build.gradle
```diff
@@ -23,77 +10,4 @@ managedDependencies {
...
-    }
-    test {
-        output.dir("${generatedFilesDir}/test/resources", builtBy: 'generateTestProto')
-    }
-    jmh {
-        output.dir("${generatedFilesDir}/jmh/resources", builtBy: 'generateJmhProto')
-    }
-}
-
-protobuf {
```
imasahiro: Guess we still need this protobuf task for generating `grpc/src/test/proto/com/linecorp/armeria/grpc/testing/`

anuraaga: Oops - had forgotten to add the source sets, fixed

benchmarks/build.gradle
```diff
@@ -0,0 +1,72 @@
+
+/*
+ * Copyright 2017 LINE Corporation
```
trustin: Probably OK to remove the license header from a build file.

build.gradle
```diff
@@ -431,39 +402,95 @@ configure(javaProjects) {
...
-                    }.each { sourceFile ->
-                        project.mkdir(outputDir)
-                        project.exec {
-                            commandLine project.ext.thriftPath,
-                                    '-gen', 'java',
-                                    '-out', outputDir,
-                                    '-I', "${sourceFile.parentFile.absolutePath}",
-                                    sourceFile.absolutePath
+    // TODO(anuraaga): Consider replacing with https://github.com/yodle/griddle
+    afterEvaluate {
```
trustin: Could you explain why this had to be moved into `afterEvaluate { .. }`?

anuraaga: Gives a chance for other plugins (ike JMH) to add source sets, which we use here. I guess in general, top-level build logic often needs to go in afterevaluate to let the individual projects do their thing first.

benchmarks/src/jmh/java/com/linecorp/armeria/benchmarks/thrift/PooledResponseBufferBenchmark.java
```diff
@@ -5,7 +5,7 @@
  * version 2.0 (the "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at:
  *
- *   https://www.apache.org/licenses/LICENSE-2.0
+ *   http://www.apache.org/licenses/LICENSE-2.0
```
imasahiro: Revert this line?

anuraaga: Weird, my copyright style is old or wrong for some reason

trustin:
@hyangtack @minwoox Gentle ping

PR #828, opened 2017-11-12
Add CentralDogma base EndpointGroup

huydx:
## Motivation
Use single file, or any part of file which served by CentralDogma as EndpointGroup.
I tried to add ServerListener to auto add/remove Endpoint as well, but CentralDogma seems not support atomic operation, so it leads to race condition when multiple armeria servers are woken up at the same time. This conditon leads to very weird state and many unexpected error, so this patch is for read-only operation.

I copied some code from ZooKeeperEndpoint (parser logic) because I can't find any place good for put those logic for reuse.

One more small things that I don't know my patch is enough for publish `armeria-centraldogma` artifacts, if I miss something please leave comments. 

Thanks

### DIFF
--- a/centraldogma/build.gradle
+++ b/centraldogma/build.gradle
@@ -1,0 +1,4 @@
+managedDependencies {
+    compile 'com.linecorp.centraldogma:centraldogma-client'
+    testCompile 'com.linecorp.centraldogma:centraldogma-testing'
+}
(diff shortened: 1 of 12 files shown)


imasahiro:
Thanks for the PR. But it will introduce cyclic dependency between armeria and CentralDogma. IMHO, it is better to put these things to CentralDogma repo than armeria.

Since I didn't open sourced our CentralDogmaEndpointGroup that is heavily used in our production, yet. So let me send a PR to CentralDogma to open-source that with your CentralDogmaCodec idea.

huydx:
I wondered about how depedency should be resolved too, put this module to CentralDogma repo is also a good idea, HDYT @trustin

trustin:
This PR itself does not introduce a cyclic dependency strictly speaking, but I agree with @imasahiro. How about moving this into `centraldogma-client`? Also, please make sure your `EndpointGroup` implementation is generic enough that it covers the requirements of @imasahiro. It'd be a good idea to compare yours to his implementation to get some hints. What do you think?

trustin:
Thanks a lot for your understanding, @huydx!

Issue #774, opened 2017-10-11
RedirectService


trustin:
Related: https://github.com/line/centraldogma/pull/33

It would be useful when a user wants to send a redirect response on a certain request:

```java
ServerBuilder sb = ...;
sb.service("/old-location", new RedirectService("/new-location"));
sb.service("/old-location/{name}", new RedirectService(ctx -> "/new-location/" + ctx.pathParam("name")));
sb.service("/old-location/{name}", new RedirectService("/new-location/{name}"));
sb.service("/old-location", new RedirectService(RedirectType.PERMANENT, "/new-location"));
```

/cc @hyangtack

trustin:
@abdulazizali77 Yes, of course. Thanks for your interest!

> should we only implement doGet and doPost? Im completely unsure whether OPTIONS, PUT, HEAD, DELETE should have a redirect, please advise

We could probably allow a user specify the methods to redirect? By default we could redirect for all methods.

> should we only support 301, 307, 308?

I think so. Please feel free to add support for more cases though if there are any.

> how should we handle the case /new-path that a RedirectService points to doesnt have a configured Service before we call build?

That's a good question. Ideally, we could add a listener to the server in `serviceAdded()` to get notified when the server if fully built and check if there's a matching path mapping actually for validation purpose. However, I think it's just fine to allow a user to specify the target path which leads to 404, because he or she might have done it intentionally.

> am i correct to believe that this will have no impact on GrpcServiceBuilder or THttpService . I am unsure if grpc or thrift has mechanism for redirect of its own.

I think `GrpcServiceBuilder` or `THttpService` is not relevant.

trustin:
Fixed via #945

Issue #773, opened 2017-10-11
HttpResponseWriter.redirect()


trustin:
Related: https://github.com/line/centraldogma/pull/33

It would be useful when a user wants to send a redirect response:

```java
res.redirect("/new-location");
res.redirect("/new-location/%s", someVar);
res.redirect(RedirectType.PERMANENT, "/new-location");
```

/cc @hyangtack

trustin:
Closing in favor of #850 and #914

PR #834, opened 2017-11-14
Add convenience methods to HttpResponseWriter for redirect (#773)

abdulazizali77:
Hi @trustin please have a look
Some things im unsure of (hence didnt do yet)
-Should we add `redirect` to `HttpResponseDecoder` `HttpResponseWrapper`, `HttpResponse` ?
-Should we have a `redirect(AggregatedHttpMessage)`  interface?
-Unsure whether we should check for the `HttpStatus` inside `redirect` to restrict only `30*`
  -i do wonder if we should just expose `redirectMultiple`, `redirectPermanent`, `redirectTemporary` to respectively map to 300, 301, 307
-Didnt find any preexisting unit tests for `DefaultHttpResponse` or `HttpResponseWriter`, should they be added?

-Most importantly, should we have one PR for both #773 and #774 or should we keep them separate? (i was thinking the latter)

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpResponseWriter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpResponseWriter.java
@@ -199,4 +199,113 @@ public interface HttpResponseWriter extends StreamWriter<HttpObject> {
 
         close();
     }
+
+    /**
+     * Convenience method for redirect.
+     *
+     * @param format {@linkplain Formatter the format string} of the response content
+     * @param args the arguments referenced by the format specifiers in the format string
+     */
+    default void redirect(String format, Object... args) {
+        requireNonNull(format, "format");
+        requireNonNull(args, "args");
+        redirect(HttpStatus.TEMPORARY_REDIRECT, format, args);
+    }
+
+    /**
+     * Convenience method for redirect.
+     *
+     * @param newLocation the new location
+     */
+    default void redirect(String newLocation) {
+        requireNonNull(newLocation, "newLocation");
+        redirect(HttpStatus.TEMPORARY_REDIRECT, newLocation);
+    }
+
+    /**
+     * Convenience method for redirect.
+     *
+     * @param status send the {@link HttpStatus} code
+     * @param format {@linkplain Formatter the format string} of the response content
+     * @param args the arguments referenced by the format specifiers in the format string
+     */
+    default void redirect(HttpStatus status, String format, Object... args) {
+        requireNonNull(status, "status");
+        requireNonNull(format, "format");
+        requireNonNull(args, "args");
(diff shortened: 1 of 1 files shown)


trustin:
Great! Thanks a lot for your contribution.

> Should we add `redirect` to `HttpResponseDecoder` `HttpResponseWrapper`, `HttpResponse` ?

Do you mean `HttpResponseDecoder.HttpResponseWrapper`? Yes. Please delegate all invocations.

`HttpResponse` is consumption-only type, so there's no point of adding `redirect()` to it.

> Should we have a `redirect(AggregatedHttpMessage)`  interface?

I'm not sure. A user could always use `HttpResponseWriter.respond(AggregatedHttpMessage)` instead in such a case? Do you have a specific use case in mind?

> Unsure whether we should check for the `HttpStatus` inside `redirect` to restrict only `30*`

Yes, please. If a user specified non-30x status, please throw an `IllegalArgumentException`, using `checkArguments()` from Guava.

> i do wonder if we should just expose `redirectMultiple`, `redirectPermanent`, `redirectTemporary` to respectively map to 300, 301, 307

I'm not sure to be honest. How about just removing `redirect()` without `HttpStatus` parameter and letting a user always be specific? What's your opinion on this?

> Didnt find any preexisting unit tests for `DefaultHttpResponse` or `HttpResponseWriter`, should they be added?

Yes, please.

> Most importantly, should we have one PR for both #773 and #774 or should we keep them separate? (i was thinking the latter)

It's up to you. Either ways are fine to us.

By the way, please sign the ICLA if you did not yet. :-)

core/src/main/java/com/linecorp/armeria/common/HttpResponseWriter.java
```diff
@@ -199,4 +199,113 @@ default void respond(AggregatedHttpMessage res) {
...
+    }
+
+    /**
+     * Convenience method for redirect.
+     *
+     * @param newLocation the new location
+     */
+    default void redirect(String newLocation) {
+        requireNonNull(newLocation, "newLocation");
+        redirect(HttpStatus.TEMPORARY_REDIRECT, newLocation);
```
huydx: assume that default redirect use 302 seems not a very nature assumption, how about remove this method and method at line 212 and remain the explicit (whith HttpStatus parameter) only

huydx: Ah, looks line @trustin already commented the same thing.

trustin:
ICLA confirmed.

trustin:
@abdulazizali77 This PR may be affected by #850.

trustin:
Closing in favor of #850. It's a shame that your work has been superceded, but I really appreciate your effort!

PR #835, opened 2017-11-14
Make tracing client has remote endpoint information.

kojilin:
User can decide to add custom service name or using RequestLog#host. Setting remote endpoint make ui looks better(It will show remote name rather than local service name if exists)

--

Should we just using Tracing.current rather than pass in Tracing? I think usually 1 vm should have only 1 Tracing. Because some library like brave.mysql using Tracing#current, so if we use multiple tracing, the service name will be weird.

### DIFF
--- a/zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
+++ b/zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
@@ -16,7 +16,12 @@
 
 package com.linecorp.armeria.client.tracing;
 
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.util.function.Function;
+
+import javax.annotation.Nullable;
 
 import com.linecorp.armeria.client.Client;
 import com.linecorp.armeria.client.ClientRequestContext;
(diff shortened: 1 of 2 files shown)


zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -88,7 +107,23 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
...
+        if (remoteAddress instanceof InetSocketAddress) {
+            endpointBuilder.ip(((InetSocketAddress) remoteAddress).getAddress());
+        }
+
+        if (remoteServiceName != null) {
+            endpointBuilder.serviceName(remoteServiceName);
+        } else if (log.host() != null) {
+            endpointBuilder.serviceName(log.host());
+        }
+        span.remoteEndpoint(endpointBuilder.build());
```
anuraaga: Should we make sure this only happens if one of the three if's happened?

anuraaga:
I think having a default constructor that uses `Tracing.current` would be fine, but we should still give the options for passing one in

kojilin:
Then I think keep as is is ok. Change as suggest, I didn't trace too much, so just avoid setting like current.

trustin:
Thanks, @kojilin !

PR #836, opened 2017-11-15
Execute stream callbacks directly when executor is the current event …

anuraaga:
…loop.

TBH, haven't been able to reason through whether this is a correct change, but tests seem fine, and the benchmarks show that we need to do something along these lines to optimize `StreamMessage` performance for the standard case where everything happens on a server event loop thread. Curious what you think.

Benchmarks (due to synchronization overhead, only high values of `num` are relevant). Probably about 1.5-2x better stream performance in practice.

After:
```
Benchmark                                                    (num)   Mode  Cnt       Score       Error  Units
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop      3  thrpt   20  238282.293 ± 11059.030  ops/s
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop      5  thrpt   20  223524.752 ±  5595.936  ops/s
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop     20  thrpt   20  170606.288 ±  4851.481  ops/s
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop    100  thrpt   20   78263.884 ±  1800.286  ops/s
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop   1000  thrpt   20   10792.808 ±   216.650  ops/s
```

Before:
```
Benchmark                                                    (num)   Mode  Cnt       Score      Error  Units
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop      3  thrpt   20  222284.597 ± 6456.982  ops/s
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop      5  thrpt   20  215316.254 ± 6570.347  ops/s
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop     20  thrpt   20  173076.370 ± 5739.022  ops/s
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop    100  thrpt   20   60886.142 ± 1018.291  ops/s
StreamMessageBenchmark.defaultStreamMessage_notJmhEventLoop   1000  thrpt   20    6617.666 ±  118.178  ops/s
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
@@ -40,6 +40,8 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.EventLoop;
+import io.netty.channel.SingleThreadEventLoop;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
 
@@ -193,7 +195,7 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
             return;
         }
 
-        if (executor != null) {
+        if (shouldRunInExecutor(executor)) {
             // NB: We did not use subscriber.onSubscribe() because that will increase the memory footprint
             //     of the anonymous class by referring to 2 variables in the lambda expression.
             executor.execute(() -> subscription.subscriber().onSubscribe(subscription));
@@ -211,7 +213,7 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
             cause = new IllegalStateException("subscribed by other subscriber already");
         }
 
-        if (executor != null) {
+        if (shouldRunInExecutor(executor)) {
             executor.execute(() -> {
                 lateSubscriber.onSubscribe(NoopSubscription.INSTANCE);
                 lateSubscriber.onError(cause);
@@ -296,7 +298,7 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
         }
 
         final Executor executor = subscription.executor();
-        if (executor != null) {
+        if (shouldRunInExecutor(executor)) {
             executor.execute(() -> notifySubscriber(subscription, queue));
         } else {
             notifySubscriber(subscription, queue);
@@ -492,6 +494,13 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
         }
     }
 
+    // We directly run callbacks for event loops if we're already on the loop, which applies to the vast
+    // majority of cases.
+    private static boolean shouldRunInExecutor(@Nullable Executor executor) {
+        return executor != null &&
+               (!(executor instanceof SingleThreadEventLoop) || !((EventLoop) executor).inEventLoop());
+    }
+
     private static final class SubscriptionImpl implements Subscription {
 
         private final DefaultStreamMessage<?> publisher;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -471,6 +473,13 @@ private void cleanup() {
         }
     }
 
+    // We directly run callbacks for event loops if we're already on the loop, which applies to the vast
+    // majority of cases.
+    private static boolean shouldRunInExecutor(@Nullable Executor executor) {
+        return executor != null &&
+               (!(executor instanceof SingleThreadEventLoop) || !((EventLoop) executor).inEventLoop());
```
trustin: The reason we always used `Executor.execute()` was to avoid the event ordering/overlap problem due to reentrance, e.g. a `Subscriber` calls `Subscription.request()` at `Subscriber.onNext()`. Could we use a thread-local variable to make sure we are not re-entering?

anuraaga: Can you describe the issue a bit more? My impression was that since events are all pushed through the queue, the ordering would be controlled by the queue, not executor, even when reentrant.

trustin: Sure. Even if all events are pushed into a queue, `Subscriber.onNext()` can overlap:

1. `DefaultStreamMessage` calls `Subscriber.onNext()`
2. `Subscriber.onNext()` calls `Subscription.request()`
3. `Subscription.request()` may trigger `DefaultStreamMessage.notifySubscriber()`
4. `DefaultStreamMessage.notifySubscriber()` calls `Subscriber.onNext()` <-- Reentrance

anuraaga: Is this a problem? Since each round through, notify will pull from the queue, it seems like order is preserved and it won't deadlock since the recursion ends after the queue is empty.

trustin: Consider this `onNext` implementation:

```java
public void onNext(Object e) {
    subscription.request(1);
    ... Do something with 'e' ...
}
```

The later element can be handled earlier due to recursion.

anuraaga: Thanks - I get it, sorry for taking a while to understand.

Do we have any determined semantics for the case where there are multi threaded writes on a stream with no executor? If not (it seems like a very bad usage of the stream), I think we can use a normal field instead of theeadlocal since we would only need to worry about single-thread event loop.

anuraaga: I have added a guard to prevent reentrance - for now went with just a field, it should handle most cases except where multiple threads write simultaneously with no executor - in that case, only one thread will be used for notification whereas without the guard, multiple threads would; not clear to me whether this is a terrible behavior or not.

trustin: Your fix won't work because of a race condition:

- Thread A sets `notifying` to `1` and enters the notification loop.
- Thread A finishes the notification loop, but it did not set `notifying` to `0` yet.
- Thread B adds some elements to the queue and calls `notifySubscriber()`, but it does nothing because `notifying` is `1` yet.
- Thread A sets `notifying` back to `0`, but it did not consume the elements produced by thread B.

anuraaga: I moved the guard so it's only in the synchronous case, and also changed it from volatile to normal boolean since I found that subscription methods can only be called by a single thread at a time

https://github.com/reactive-streams/reactive-streams-jvm#2-subscriber-code  - 2.7

Also, the spec does seem to indicate that there needs to be a recursion limit, which was only true when we had the executor, so solving the non-executor case with this or similar seems required. 

https://github.com/reactive-streams/reactive-streams-jvm#3-subscription-code 3.3

But getting strange errors with the compliance tests...

anuraaga:
I have split off a PR to first solve the reentrancy problem in the non-executor case which seems required by the spec.

trustin:
@anuraaga Time to rebase :-)

anuraaga:
@trustin Rebased :)

trustin:
Thanks a lot, @anuraaga !

PR #839, opened 2017-11-16
Fix KeeperException.NoNodeException in ZooKeeperConnector

trustin:
Motivation:

KeeperException.NoNodeException may be thrown when iterating over the
children of a node if they were removed by other party during iteration.

Modifications:

- Handle NoNodeException in ZooKeeperConnector.notifyChange()

Result:

- Robustness


trustin:
Thanks for reviewing!

PR #841, opened 2017-11-17
Prevent reentrant synchronous callback execution.

anuraaga:
I have split this off of #836 to focus on the re-entrancy problem. Even without trying to optimize event loop access of the stream message, we should be handling it correctly for the non-executor stream case.

https://github.com/reactive-streams/reactive-streams-jvm#2-subscriber-code - 2.7

The test fails without the logic change and passes after it. Unfortunately, I'm trying to wrap my head around failures with the `testNg`. Let me know if you have any ideas what could be up.

```
PublisherBasedStreamMessageVerification.required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue: FAILURE

Gradle suite > Gradle test > com.linecorp.armeria.common.stream.PublisherBasedStreamMessageVerification.required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue FAILED
    java.lang.OutOfMemoryError: GC overhead limit exceeded
DefaultStreamMessageVerification.required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue: FAILURE

Gradle suite > Gradle test > com.linecorp.armeria.common.stream.DefaultStreamMessageVerification.required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue FAILED
    java.lang.OutOfMemoryError: GC overhead limit exceeded
DeferredStreamMessageVerification.required_abortMustNotifySubscriber: FAILURE

Gradle suite > Gradle test > com.linecorp.armeria.common.stream.DeferredStreamMessageVerification.required_abortMustNotifySubscriber FAILED
    java.lang.AssertionError: Expected onError(com.linecorp.armeria.common.stream.AbortedStreamException) within 100 ms

        Caused by:
        java.lang.AssertionError: Expected onError(com.linecorp.armeria.common.stream.AbortedStreamException) within 100 ms
DeferredStreamMessageVerification.required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue: FAILURE

Gradle suite > Gradle test > com.linecorp.armeria.common.stream.DeferredStreamMessageVerification.required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue FAILED
    java.lang.OutOfMemoryError: GC overhead limit exceeded

172 tests completed, 4 failed, 49 skipped

```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
@@ -217,6 +217,9 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
         }
 
         private void pushSignal(Object obj) {
+            if (state == State.CLOSED) {
+                return;
+            }
             if (!(obj instanceof CloseEvent)) {
                 final int dataLength = signalLengthGetter.length(obj);
                 if (dataLength > 0) {
@@ -784,10 +787,6 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
 
         int addAndRemoveIfRequested(Object o) {
             requireNonNull(o);
-            if (elements == null) {
-                // Stream already closed.
-                return 0;
-            }
             int removedLength = 0;
             if (headOffset < lastRemovalRequestedOffset) {
                 removedLength = removeElements();
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
@@ -784,6 +784,10 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
 
         int addAndRemoveIfRequested(Object o) {
             requireNonNull(o);
+            if (elements == null) {
+                // Stream already closed.
+                return 0;
+            }
             int removedLength = 0;
             if (headOffset < lastRemovalRequestedOffset) {
                 removedLength = removeElements();
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
@@ -132,6 +132,8 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
 
     private volatile boolean wroteAny;
 
+    private boolean inOnNext;
+
     /**
      * Creates a new instance with a new {@link ConcurrentLinkedQueue}.
      */
@@ -302,12 +304,29 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
     }
 
     private void notifySubscriber(SubscriptionImpl subscription, Queue<Object> queue) {
-        if (state == State.CLEANUP) {
-            cleanup();
+        if (inOnNext) {
+            // Do not let Subscriber.onNext() reenter, because it can lead to weird-looking event ordering
+            // for a Subscriber implemented like the following:
+            //
+            //   public void onNext(Object e) {
+            //       subscription.request(1);
+            //       ... Handle 'e' ...
+            //   }
+            //
+            // Note that we do not call this method again, because we are already in the notification loop
+            // and it will consume the element we've just added in pushObject() from the queue as expected.
+            //
+            // We do not need to worry about synchronizing the access to 'inOnNext' because the subscriber
+            // methods must be on the same thread, or synchronized, according to Reactive Streams spec.
             return;
         }
 
         for (;;) {
+            if (state == State.CLEANUP) {
+                cleanup();
+                return;
+            }
+
             final Object o = queue.peek();
             if (o == null) {
                 break;
@@ -356,7 +375,9 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
                     }
                 }
 
+                inOnNext = true;
                 subscriber.onNext(o);
+                inOnNext = false;
                 return true;
             }
         }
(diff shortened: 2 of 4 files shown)


trustin:
Made the following change and problem seems gone:

```diff
diff --git a/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java b/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
index ffa09cee..82da570c 100644
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
@@ -132,7 +132,7 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
 
     private volatile boolean wroteAny;
 
-    private boolean notifyingSynchronously;
+    private boolean inOnNext;
 
     /**
      * Creates a new instance with a new {@link ConcurrentLinkedQueue}.
@@ -299,20 +299,29 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
         if (executor != null) {
             executor.execute(() -> notifySubscriber(subscription, queue));
         } else {
-            if (notifyingSynchronously) {
-                // We don't notify again if we're already notifying, as it may cause incorrect order of handling
-                // objects in the Subscriber - the current notifier will loop through queue correctly and we
-                // don't do anything here. Because subscriber methods must be on the same thread, or
-                // synchronized, we do not have to worry about synchronizing this field access.
-                return;
-            }
-            notifyingSynchronously = true;
             notifySubscriber(subscription, queue);
-            notifyingSynchronously = false;
         }
     }
 
     private void notifySubscriber(SubscriptionImpl subscription, Queue<Object> queue) {
+        if (inOnNext) {
+            // Do not let Subscriber.onNext() reenter, because it can lead to weird-looking event ordering
+            // issue for a Subscriber implemented like the following:
+            //
+            //   public void onNext(Object e) {
+            //       subscription.request(1);
+            //       ... Handle 'e' ...
+            //   }
+            //
+            // Note that we do not call this method again, because we are already in the notification loop
+            // and it will consume the element we've just added in pushObject() from the queue as expected.
+            //
+            // We do not need to worry about synchronizing the access to the 'inOnNext' because the subscriber
+            // methods must be on the same thread, or synchronized, according to Reactive Streams spec.
+            return;
+        }
+
+
         for (;;) {
             if (state == State.CLEANUP) {
                 cleanup();
@@ -367,7 +376,9 @@ public class DefaultStreamMessage<T> implements StreamMessage<T>, StreamWriter<T
                     }
                 }
 
+                inOnNext = true;
                 subscriber.onNext(o);
+                inOnNext = false;
                 return true;
             }
         }
```

trustin:
- I think we don't need to protect reentrance for `onComplete()` and `onError()` because it's terminal condition, so I used the flag only for `onNext()`.
- I prevented the reentrance even when an executor is specified because we will optimize this later in #836.
- The reason why the TestNG tests failed was it gave no chance to the `Subscriber` to consume anything, because your patch set `notifyingSynchronously` too soon, even before invoking `Subscriber`.

anuraaga:
Oh! This might also be related to the TCK tests being implemented using `onDemand` - as I found, this was the reason that the implementation passed required_spec303_mustNotAllowUnboundedRecursion despite the code being wrong. Do you think we need to rethink the TCK test harness? And in that case, I think it won't matter too much which location the booleans are set.

trustin:
> the implementation passed required_spec303_mustNotAllowUnboundedRecursion despite the code being wrong. Do you think we need to rethink the TCK test harness? 

That's a good question. I think we can't do much about `DefaultStreamMessageVerification` because `onDemand()` is just how a user is expected to use to stream a large number of streams. We could improve `PublishedBasedStreamMessageVerification` to source the stream from a simpler `Publisher` implementation which does not extend `DefaultStreamMessage`, and use `PublisherBasedStreamMessage` for `DeferredStreamMessageVerification`.

> And in that case, I think it won't matter too much which location the booleans are set.

I'm not sure. Looking from the debugger, the flag was set even before `onNext()` was called first time, so I'd say it has to be set only when it is apparent that `onNext()` will be called.

anuraaga:
Yeah it was set by awaitdemandfuture - to be honest I wasn't sure what its semantics are supposed to be, but I agree that just wrapping onNext is probably the right call. Will incorporate the suggestion and add comments to the test case!

trustin:
@anuraaga Yes, please! Thanks again :-)

anuraaga:
Updated - please also take a good look at `AbstractStreamMessageDuplicator`, I found with this change it was logging a NPE that didn't affect test results.

I think it's because while I was debugging TCK code for this, I noticed we weren't following 1.08 - `If a Subscription is cancelled its Subscriber MUST eventually stop being signaled.` Without checking for `CLEANUP` state in the loop, we ended up signaling all objects to the subscriber even if they cancelled. Note, 1.08 is not verified by the TCK.

After fixing this, it seems the duplicator would try to push a `CloseEvent` onto the queue after already having been `clear`ed. Not too sure why this would happen so please check my fix makes any sense at all.

minwoox:
Will check the duplicator soon. Thank you~!!

trustin:
@minwoox Is #843 also related with the NPE we observed here?

minwoox:
@trustin I could not take a look at this carefully because I had to make weaning foods -_-;. I will look into it.

minwoox:
@anuraaga Thank you for the fix~!! I think it's reasonable fix. 👍  But I think it's a little bit better to check if the `signals` is cleared before `signals.addAndRemoveIfRequested(obj)` is called.

I will dig into more about why `CloseEvent` is pushed after it is cleared. By the way, at which test case did you find this situation?

anuraaga:
Thanks @minwoox - I tried applying the fix from #843 and removing mine and indeed the NPE still happens, so they seem to be two separate issues.

Most of the tests in `RetryingHttpClientTest`, for example `respectRetryAfter`, reproduce the NPE.

anuraaga:
By the way @minwook I have tried to follow your suggestion by checking `state == CLOSED` before pushing objects in the duplicator. I believe the double-close happens when downstream is closed, for example by `HttpRequestSubscriber.setDone` and upstream is also closed before getting the cancellation from the duplicator. I guess this is pretty standard - WDYT?

minwoox:
@anuraaga yeah, they are two different issues.
>  I guess this is pretty standard - WDYT?

You're right. I should've prepared for that situation.
Thank you for fix that~! 💯

trustin:
Thanks!

PR #843, opened 2017-11-18
Change isEmpty() in ChildStreamMessage

### LINKED ISSUE #842: Retrying http request does not contains http content
When a RetryingHttpClient resend a request, it only contains http headers and http content is not sent to server.

* Test code
```java
    @Rule
    public final ServerRule server = new ServerRule() {
        @Override
            sb.service("/post-ping-pong", new AbstractHttpService() {
                final AtomicInteger reqPostCount = new AtomicInteger();

                @Override
                protected void doPost(ServiceRequestContext ctx, HttpRequest req, HttpResponseWriter res)
                        throws Exception {
                    req.aggregate().handle(voidFunction((message, thrown) -> {
                        if (reqPostCount.getAndIncrement() < 1) {
                            res.respond(HttpStatus.SERVICE_UNAVAILABLE);
                        } else {
                            res.respond(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, message.content());
                        }
                    }));
                }
            });
        }
    };

    @Test
    public void sendHttpBodyCorrectly() {
        final Backoff backoffOnServerError = Backoff.fixed(10);
        final RetryStrategy<HttpRequest, HttpResponse> strategy =
                RetryStrategy.onServerErrorStatus(backoffOnServerError);
        final HttpClient client =
                new HttpClientBuilder(server.uri("/"))
                        .factory(clientFactory)
                        .decorator(RetryingHttpClient.newDecorator(strategy))
                        .build();

        final AggregatedHttpMessage res = client.post("/post-ping-pong", "bar").aggregate().join();
        assertThat(res.content().toStringUtf8()).isEqualTo("bar");
    }
```

* Test result:
```
org.junit.ComparisonFailure: 
Expected :"bar"
Actual   :""
```

minwoox:
Motivation:
Currently, `isEmpty()` and `isOpen()` in the `streamMessge` from a `duplicator` using different layer to check their return value.
This is the problem because, for example, if the `streamMessage` has 3 signals and the streamMessage’s `close()` is called, it’s status will be closed and not empty.
However, a `ChildStreamMessage`’s status will be closed(upstream’s status) and empty(downstream’s status) because `subscription.request(long)` is not called.

There are two options to solve this problems. First one is using downstream’s(or processor in the `duplicator`) status.
If we using this, the child stream will be open until it receives `onComplete()` or `onError()` even thought the upstream is closed already.

Second one is using upstream’s status with dealing the situation when ChildStream’s `abort()` is called.

I prefer second one because the `HttpRequestSubscriber` will do its work right away when it writes a http request without the body.(If I use the first one, it needs one more subscription.request(1)).

Modifications:
- Change the inside of `isEmpty()`

Result:
- Fix #842

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
@@ -385,8 +385,7 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
                 return false;
             }
 
-            final DownstreamSubscription<T> subscription = this.subscription;
-            return subscription != null && !subscription.publishedAny;
+            return processor.upstream().isEmpty();
         }
 
         @Override
@@ -518,7 +517,6 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
 
         private volatile int offset;
         private long cumulativeDemand;
-        private boolean publishedAny;
 
         DownstreamSubscription(ChildStreamMessage<T> streamMessage,
                                Subscriber<? super T> subscriber, StreamMessageProcessor<T> processor,
@@ -680,7 +678,6 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
                         }
                     }
 
-                    publishedAny = true;
                     subscriber.onNext(obj);
                     return true;
                 }
--- a/core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.client.retry;
 
+import static com.linecorp.armeria.common.util.Functions.voidFunction;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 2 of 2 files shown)


minwoox:
Also, I fount this from the java-doc of `StreamMessage` which is:
A StreamMessage is open since its instantiation until:
- the Subscriber consumes all elements and Subscriber.onComplete() is invoked,
- an error occurred and Subscriber.onError(Throwable) is invoked,
- the Subscription has been cancelled or
- abort() has been requested.

Currently, IIUC, the `StreamMessage` is closed when the corresponding `StreamWriter`'s `close()` is called regardless of the subscriber subscribes or not.
I'm not sure if we need to add those statement to the java-doc.

trustin:
> Currently, IIUC, the `StreamMessage` is closed when the corresponding `StreamWriter`'s `close()` is called regardless of the subscriber subscribes or not.

No, `StreamMessage` is should not be closed even if `StreamWriter.close()` was called. Whether a stream is closed or open entirely depends on the consumption of the stream.

minwoox:
> StreamMessage is should not be closed even if StreamWriter.close() was called.

I think so. But currently, When calling the `isOpen()` in the `DefaultStreamMessage`, it returns `false` after `StreamWriter.close()` is called.
It is because `StreamMessage` and `StreamWriter` both have `isOpen()` method and it changes the  `StreamMessge`'s state as well in current implementation.

imasahiro:
Thanks for quick fix!

PR #845, opened 2017-11-20
Revert "Make endpoint selection asynchronous (#819)"

imasahiro:
This reverts commit 429b658cf057be454691b2489270cc4ca4f55fb6.

Because we realized that it does not provide any merits for now, make endpoint selection synchronous again.


trustin:
Thanks, @imasahiro !

PR #847, opened 2017-11-21
Add exception handler for annotated http service

hyangtack:
Motivation:
Currently, there is no way to send a customized response when an exception is raised from an annotated http service method. We only have an HttpResponseException which sends an http status code without http body.

Modifications:
- Add `@ExceptionHandler` annotation in order to specify exception handler on annotated method.
- Add ExceptionHandlerFunction to handle exception.
- Rename HttpResponseException to HttpStatusException.
- Add HttpResponseException which holds an HttpResponse as its member.
- Rename `@Decorate` to `@Decorator` and make it repeatable.
- Remove ResourceNotFoundException and ServiceUnavailableException. Use HttpStatusException instead.

Result:
We can specify exception handlers on annotated method.


hyangtack:
I will add more tests but please review the concept of exception handler first. :-)

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -136,8 +139,8 @@ public void onNext(T t) {
 
         @Override
         public void onError(Throwable t) {
-            beforeError(delegate, t);
-            delegate.onError(t);
+            final Throwable filteredCause = beforeError(delegate, t);
+            delegate.onError(filteredCause != null ? filteredCause : t);
```
trustin: How about translating `null` into `onComplete()` instead of falling back to the original cause?

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -64,9 +64,12 @@ protected void beforeComplete(Subscriber<? super U> subscriber) {}
     /**
      * A callback executed just before calling {@link Subscriber#onError(Throwable)} on {@code subscriber}.
      * Override this method to execute any cleanup logic that may be needed before failing the
-     * subscription.
+     * subscription. This method may rewrite the {@code cause} and then return a new one so that the new
+     * {@link Throwable} would be passed to {@link Subscriber#onError(Throwable)}.
      */
-    protected void beforeError(Subscriber<? super U> subscriber, Throwable cause) {}
+    protected Throwable beforeError(Subscriber<? super U> subscriber, Throwable cause) {
```
trustin: Add `@Nullable` to the method.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -71,11 +77,15 @@
     private final List<Parameter> parameters;
     private final boolean isAsynchronous;
     private final AggregationStrategy aggregationStrategy;
+    private final Map<Class<? extends Throwable>, ExceptionHandlingFunction> exceptionHandlerMap;
 
-    AnnotatedHttpServiceMethod(Object object, Method method, PathMapping pathMapping) {
+    AnnotatedHttpServiceMethod(Object object, Method method, PathMapping pathMapping,
+                               Map<Class<? extends Throwable>,
+                                       ExceptionHandlingFunction> exceptionHandlerMap) {
```
trustin: Indentation

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -140,24 +150,54 @@ public Object apply(ServiceRequestContext ctx, HttpRequest req) {
...
     private Object invoke(ServiceRequestContext ctx, HttpRequest req, @Nullable AggregatedHttpMessage message) {
         try (SafeCloseable ignored = RequestContext.push(ctx, false)) {
             return method.invoke(object, parameterValues(ctx, req, parameters, message));
         } catch (Exception e) {
-            if (e instanceof InvocationTargetException) {
-                final Throwable cause = e.getCause();
-                if (cause != null) {
-                    return Exceptions.throwUnsafely(cause);
-                }
+            final Throwable cause = e instanceof InvocationTargetException ? firstNonNull(e.getCause(), e)
```
trustin: This seems unnecessary because `getResponseForCause()` will call `Throwables.getRootCause()`.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -364,20 +390,63 @@ private static ResponseConverter converter(Method method) {
...
+            ExceptionHandlingFunction> exceptionHandlerMap(ExceptionHandler[] handlers) {
+        final ImmutableMap.Builder<Class<? extends Throwable>,
+                ExceptionHandlingFunction> builder = new Builder<>();
+        Arrays.stream(handlers)
+              .map(ExceptionHandler::value)
+              .forEach(clazz -> {
+                  final ExceptionHandling types = clazz.getAnnotation(ExceptionHandling.class);
+                  if (types == null) {
+                      throw new IllegalArgumentException(
+                              "Exception handler must specify an exception to handle as @" +
```
trustin: as -> at?

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -216,12 +222,22 @@ public void onNext(HttpObject o) {
     @Override
     public void onError(Throwable cause) {
         if (cause instanceof HttpResponseException) {
-            failAndRespond(cause, ((HttpResponseException) cause).httpStatus(), Http2Error.CANCEL);
+            // Timeout may be occurred because calling failAndRespond is postponed.
```
trustin: - `Timeout may occur when the aggregation of the error response take long.`?
- How about adding this comment as well? - `If timeout occurs, respond with 500 Internal Server Error.`
  - How about reusing `HttpResponse.status()` instead?

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandling.java
```diff
@@ -0,0 +1,32 @@
...
+package com.linecorp.armeria.server.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface ExceptionHandling {
```
trustin: How about just adding a method that matches an exception to `ExceptionHandlingFunction`?

```java
public interface ExceptionHandlingFunction {
    default boolean accept(Throwable cause) {
        return true;
    }
    ...
}
```

Than we wouldn't this need annotation.

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandlingFunction.java
```diff
@@ -0,0 +1,34 @@
...
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.RequestContext;
+
+/**
+ * An interface for exception handler.
+ */
+@FunctionalInterface
+public interface ExceptionHandlingFunction {
```
trustin: Which is better? `ExceptionHandlingFunction` vs. `ExceptionHandlerFunction` /cc @anuraaga :-)

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandling.java
hyangtack: Do you mean that we use ExceptionHandlingFunction list instead of map when finding a suitable exception handler?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
hyangtack: Ah, you're right. I added `Throwables.getRootCause()` at the last, so I missed removing this.

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandling.java
trustin: Yes. Even if we specified `@ExceptionHandling(SomeException.class)` we will have to do some iteration of class hierarchy and interface traversal anyway. Given that the number of exception handler is going to be small, I think it may be fine. What do you think?

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
hyangtack: It makes sense. I will check it again.

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandling.java
hyangtack: Actually this was another option that I thought. So I think it's also okay. :-)

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
hyangtack: As we talked about this offline, there is no way to make a response if returning `null`. So it would be better not fixing this.

imasahiro:
FWIW, recently I wrote similer ExceptionHandler using SimpleDecoratingService.
https://gist.github.com/imasahiro/53dc8747cb6594bbbc9a5935123805c3

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
minwoox: I think we should apply the same logic to `beforeSubscribe` as well.

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
```diff
@@ -558,19 +558,19 @@ public String headerDefault(RequestContext ctx,
 
         @Get("/headerWithParam")
         public String headerWithParam(RequestContext ctx,
-                                    @Header("username") @Default("hello") String username,
-                                    @Header("password") @Default("world") Optional<String> password,
-                                    @Param("extra") Optional<String> extra,
-                                    @Param("number") int number) {
+                                      @Header("username") @Default("hello") String username,
```
minwoox: Thank you!

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
```diff
@@ -558,19 +558,19 @@ public String headerDefault(RequestContext ctx,
...
             return username + "/" + password.get() + "/" + extra.orElse("(null)") + "/" + number;
         }
 
         @Get
         @Path("/headerWithoutValue")
         public String headerWithoutValue(RequestContext ctx,
-                                    @Param("username") @Default("hello") String username,
-                                    @Param("password") String password) {
+                                         @Param("username") @Default("hello") String username,
+                                         @Param("password") String password) {
```
minwoox: Thank you~! could you replace above two `@Param` with `@Header` as well? That was my first intention. :-)

hyangtack:
@imasahiro Oh, your idea is very similar to mine! 😃  Please review again once I make changes about @trustin 's comment.

hyangtack:
@minwoox 
> I think we should apply the same logic to beforeSubscribe as well.

Is there no problem when replacing Subscription with new one? If so, I can make a change. Would you check it? 😃

minwoox:
> Is there no problem when replacing Subscription with new one?

I think it's ok as long as `request(long n)` and `cancel()` are propagated properly. However, I thought about again, and there's not so much usage except just counting `n`.
So, I think we better leave it as it is, and change it when we need it.

hyangtack:
@minwoox Thank you for checking! 👍

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandlingFunction.java
hyangtack: I will rename it as `ExceptionHandlerFunction` because I removed `ExceptionHandling` annotation while changing exception map to list.

core/src/main/java/com/linecorp/armeria/common/HttpMessageAggregator.java
```diff
@@ -116,8 +116,11 @@ public void accept(Void unused, Throwable cause) {
             content = HttpData.of(merged);
         }
 
-        final AggregatedHttpMessage aggregated = onSuccess(content);
-        future.complete(aggregated);
+        try {
+            future.complete(onSuccess(content));
+        } catch (Exception e) {
```
trustin: Maybe we should catch a `Throwable` here because otherwise the future will never be notified?

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -136,8 +143,13 @@ public void onNext(T t) {
 
         @Override
         public void onError(Throwable t) {
-            beforeError(delegate, t);
-            delegate.onError(t);
+            final Throwable filteredCause = beforeError(delegate, t);
+            if (filteredCause != null) {
+                delegate.onError(filteredCause);
+            } else {
+                logger.warn("FilteredStreamMessage#beforeError returns null for exception: ", t);
```
trustin: How about:

```java
logger.warn("{}#beforeError() returned null. Using the original exception:",
            FilteredStreamMessage.this.getClass().getName(), t);
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -364,20 +377,38 @@ private static ResponseConverter converter(Method method) {
...
+        if (handlers.length == 0) {
+            handlers = clazz.getAnnotationsByType(ExceptionHandler.class);
+        }
+        if (handlers.length == 0) {
+            return defaultExceptionHandlers;
+        }
+        return Arrays.stream(handlers)
+                     .map(h -> exceptionHandlingFunctions.computeIfAbsent(h.value(), type ->
+                             newInstance(type, ExceptionHandler.class)))
+                     .collect(Collectors.toList());
```
trustin: `toImmutableList()`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -71,11 +75,14 @@
...
     private final AggregationStrategy aggregationStrategy;
+    private final List<ExceptionHandlerFunction> exceptionHandlers;
 
-    AnnotatedHttpServiceMethod(Object object, Method method, PathMapping pathMapping) {
+    AnnotatedHttpServiceMethod(Object object, Method method, PathMapping pathMapping,
+                               List<ExceptionHandlerFunction> exceptionHandlers) {
         this.object = requireNonNull(object, "object");
         this.method = requireNonNull(method, "method");
         requireNonNull(pathMapping, "pathMapping");
+        this.exceptionHandlers = requireNonNull(exceptionHandlers, "exceptionHandlers");
```
trustin: Wrap `exceptionHandlers` with `ImmutableList.copyOf()`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -140,22 +147,54 @@ public Object apply(ServiceRequestContext ctx, HttpRequest req) {
...
+     */
+    private HttpResponse getResponseForCause(ServiceRequestContext ctx, HttpRequest req,
+                                             Throwable cause) {
+        final Throwable rootCause = Throwables.getRootCause(cause);
+        for (ExceptionHandlerFunction func : exceptionHandlers) {
+            if (func.accept(rootCause)) {
+                try {
+                    return func.handle(ctx, req, rootCause);
+                } catch (Exception e) {
+                    logger.warn("An exception is raised from exception handler: ", e);
```
trustin: How about:

```java
logger.warn("Unexpected exception from an exception handler {}:", func.getClass().getName(), e);
```

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -0,0 +1,89 @@
...
+                    Exceptions.clearTrace(new HttpStatusException(code)));
+        }
+    }
+
+    private static final long serialVersionUID = 3341744805097308847L;
+
+    private final HttpStatus httpStatus;
+
+    /**
+     * Creates a new instance with HTTP status code.
```
trustin: HTTP -> an HTTP

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * A {@link RuntimeException} that is raised when an armeria internal http exception has occurred.
+ * This class is the general class of exceptions produced by a failed request or a reset stream.
+ */
+public class HttpStatusException extends RuntimeException {
+
+    private static final Map<Integer, HttpStatusException> EXCEPTIONS = new ConcurrentHashMap<>();
+
+    /**
+     * Returns a new {@link HttpStatusException} instance with the HTTP status code.
```
trustin: with the -> with the specified

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public static HttpStatusException of(int statusCode) {
+        return of(HttpStatus.valueOf(statusCode));
+    }
+
+    /**
+     * Returns a new {@link HttpStatusException} instance with the {@link HttpStatus}.
```
trustin: with the -> with the specified

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.linecorp.armeria.common.Flags;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.util.Exceptions;
+
+/**
+ * A {@link RuntimeException} that is raised when an armeria internal http exception has occurred.
+ * This class is the general class of exceptions produced by a failed request or a reset stream.
```
trustin: How about: It is a general exception raised by a failed request or a reset stream.

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -0,0 +1,89 @@
...
+     * Creates a new instance with HTTP status code.
+     */
+    public HttpStatusException(int statusCode) {
+        this(HttpStatus.valueOf(statusCode));
+    }
+
+    /**
+     * Creates a new instance.
+     */
+    public HttpStatusException(HttpStatus httpStatus) {
```
trustin: Probably OK to make the public constructors private, because a user can use the static factory methods?

core/src/main/java/com/linecorp/armeria/server/annotation/Decorator.java
```diff
@@ -29,13 +30,14 @@
...
  */
+@Repeatable(Decorators.class)
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
-public @interface Decorate {
+public @interface Decorator {
 
     /**
      * An array of the {@link DecoratingServiceFunction} classes. Each class specified in the {@code value}
      * must have an accessible default constructor.
```
trustin: Update Javadoc?

core/src/main/java/com/linecorp/armeria/server/annotation/Decorator.java
```diff
@@ -29,13 +30,14 @@
  * Specifies {@link DecoratingServiceFunction} classes which handle a {@link HttpRequest} before invoking
  * an annotated service method.
```
trustin: ... classes -> a ... class

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandlerFunction.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.linecorp.armeria.server.annotation;
+
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.RequestContext;
+
+/**
+ * An interface for exception handler.
```
trustin: How about adding `@see ExceptionHandler`?

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceExceptionHandlerTest.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+    }
+
+    static class BadExceptionHandler implements ExceptionHandlerFunction {
+        @Override
+        public HttpResponse handle(RequestContext ctx, HttpRequest req, Throwable cause) {
+            final DefaultHttpResponse response = new DefaultHttpResponse();
+            response.write(HttpHeaders.of(HttpStatus.OK));
+            Executors.newSingleThreadScheduledExecutor().schedule(
+                    (Runnable) response::close, 1, TimeUnit.SECONDS);
```
trustin: - `ctx.eventLoop().schedule(...)`?
- Could use longer delay so that the response is not closed before the assertion in a very slow machine?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -76,13 +80,26 @@
...
             Class<? extends DecoratingServiceFunction<HttpRequest, HttpResponse>>,
             DecoratingServiceFunction<HttpRequest, HttpResponse>>
             decoratingServiceFunctions = new ConcurrentHashMap<>();
 
+    /**
+     * An {@link ExceptionHandlerFunction} map for reusing exception handlers.
+     */
+    private static final ConcurrentMap<
+            Class<? extends ExceptionHandlerFunction>, ExceptionHandlerFunction>
+            exceptionHandlingFunctions = new ConcurrentHashMap<>();
```
minwoox: exceptionHandlerFuntions?

trustin:
@kojilin @imasahiro Care to review? Would be great to get your feed back about whether the API meets your requirements before releasing 0.55.0.

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandlerFunction.java
hyangtack: Done.

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -136,8 +143,14 @@ public void onNext(T t) {
...
         @Override
         public void onError(Throwable t) {
-            beforeError(delegate, t);
-            delegate.onError(t);
+            final Throwable filteredCause = beforeError(delegate, t);
+            if (filteredCause != null) {
+                delegate.onError(filteredCause);
+            } else {
+                logger.warn("{}#beforeError() returned null. Using the original exception:",
+                            FilteredStreamMessage.this.getClass().getName(), t);
```
trustin: I wonder if we should log the full trace here. We delegate the exception to `delegate` anyway, so it is likely that it will be handled there. It's probably OK to log `t.toString()` only?

core/src/main/java/com/linecorp/armeria/server/HttpResponseException.java
```diff
@@ -17,21 +17,17 @@
...
+import com.linecorp.armeria.common.AggregatedHttpMessage;
 import com.linecorp.armeria.common.Flags;
+import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
-import com.linecorp.armeria.common.util.Exceptions;
 
 /**
  * A {@link RuntimeException} that is raised when an armeria internal http exception has occurred.
- * This class is the general class of exceptions produced by a failed request or a reset stream.
+ * This class holds an {@link HttpResponse} which would be sent to the client sent the corresponding request.
```
trustin: How about: sent back to the client who sent

core/src/main/java/com/linecorp/armeria/server/HttpResponseException.java
```diff
@@ -44,44 +40,39 @@ public static HttpResponseException of(int statusCode) {
...
-                    Exceptions.clearTrace(new HttpResponseException(code)));
-        }
+        return new HttpResponseException(HttpResponse.of(httpStatus));
     }
 
-    private static final long serialVersionUID = 3487991462085151316L;
-
-    private final HttpStatus httpStatus;
+    /**
+     * Returns a new {@link HttpResponseException} instance with the {@link AggregatedHttpMessage}.
```
trustin: the -> the specified

core/src/main/java/com/linecorp/armeria/server/HttpResponseException.java
```diff
@@ -44,44 +40,39 @@ public static HttpResponseException of(int statusCode) {
...
+     */
+    public static HttpResponseException of(AggregatedHttpMessage httpMessage) {
+        return of(requireNonNull(httpMessage, "httpMessage").toHttpResponse());
+    }
 
     /**
-     * Creates a new instance with HTTP status code.
+     * Returns a new {@link HttpResponseException} instance with the {@link HttpResponse}.
```
trustin: the -> the specified

core/src/main/java/com/linecorp/armeria/server/HttpResponseException.java
```diff
@@ -44,44 +40,39 @@ public static HttpResponseException of(int statusCode) {
...
+        return new HttpResponseException(httpResponse);
     }
 
+    private static final long serialVersionUID = 3487991462085151316L;
+
+    private final HttpResponse httpResponse;
+
     /**
-     * Creates a new instance.
+     * Creates a new instance with {@link HttpResponse}.
```
trustin: with -> with the specified

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -308,8 +324,8 @@ private void failAndRespond(Throwable cause, HttpStatus status, Http2Error error
 
         if (wroteNothing(state)) {
             // Did not write anything yet; we can send an error response instead of resetting the stream.
-            responseEncoder.writeHeaders(ctx, id, streamId, headers, false);
-            responseEncoder.writeData(ctx, id, streamId, content, true);
+            responseEncoder.writeHeaders(ctx, id, streamId, message.headers(), false);
+            responseEncoder.writeData(ctx, id, streamId, message.content(), true);
```
trustin: Not related with this PR directly, but how about not writing the data frame if data is empty? i.e.

```java
final HttpHeaders headers = message.headers();
final HttpData content = message.content();
if (content.isEmpty()) {
    responseEncoder.writeHeaders(ctx, id, streamId, headers, true);
} else {
    responseEncoder.writeHeaders(ctx, id, streamId, headers, false);
    responseEncoder.writeData(ctx, id, streamId, content, true);
}
```

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -261,7 +261,8 @@ private void handleRequest(ChannelHandlerContext ctx, DecodedHttpRequest req) th
         final PathMapped<ServiceConfig> mapped;
         try {
             mapped = host.findServiceConfig(mappingCtx);
-        } catch (HttpResponseException cause) {
+        } catch (HttpStatusException cause) {
+            // We do not use HttpResponseException internally.
```
trustin: How about:

    // We do not need to handle HttpResponseException here because we do not use it internally.

for clarification?

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.linecorp.armeria.common.Flags;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.util.Exceptions;
+
+/**
+ * A {@link RuntimeException} that is raised when an armeria internal http exception has occurred.
```
trustin: - IIUC, a user can also throw this exception, so it's not an internal exception. How about:

  ```
  A {@link RuntimeException} that is raised to send a simplistic HTTP response with minimal content by a {@link Service}.
  ```

- How about adding `@see HttpResponseException`?

core/src/main/java/com/linecorp/armeria/server/annotation/Decorator.java
```diff
@@ -26,16 +27,17 @@
...
+@Repeatable(Decorators.class)
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
-public @interface Decorate {
+public @interface Decorator {
 
     /**
-     * An array of the {@link DecoratingServiceFunction} classes. Each class specified in the {@code value}
+     * {@link DecoratingServiceFunction} class specified in the {@code value}
      * must have an accessible default constructor.
```
trustin: How about:

```
{@link DecoratingServiceFunction} implementation type. The specified class must have an accessible default constructor.
```

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+ * annotated service method.
+ */
+@Repeatable(ExceptionHandlers.class)
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+public @interface ExceptionHandler {
+
+    /**
+     * {@link ExceptionHandlerFunction} class specified in the {@code value}
+     * must have an accessible default constructor.
```
trustin: Please see the comment for `@Decorator` annotation.

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandlerFunction.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @see ExceptionHandler
+ */
+@FunctionalInterface
+public interface ExceptionHandlerFunction {
```
trustin: This class has missing field and method Javadoc.

core/src/main/java/com/linecorp/armeria/server/annotation/Decorator.java
```diff
@@ -26,16 +27,17 @@
 import com.linecorp.armeria.server.DecoratingServiceFunction;
 
 /**
- * Specifies {@link DecoratingServiceFunction} classes which handle a {@link HttpRequest} before invoking
+ * Specifies a {@link DecoratingServiceFunction} class which handle a {@link HttpRequest} before invoking
```
trustin: handle a -> handles an

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandlerFunction.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * A default exception handler function. It returns {@link HttpResponse} with
```
trustin: returns -> returns an

trustin:
AppVeyor failure seems unrelated.

trustin:
@imasahiro OK to merge? :-)

trustin:
Thank you all!

PR #848, opened 2017-11-21
Ensure Subscriber.onSubscribe() is invoked before other methods

trustin:
Motivation:

Commit 4f0cee147101db3911b36c6c1a8525be69199cfe introduced a regression
where Subscriber.onSubscribe() can be invoked after onComplete().

Modifications:

- Reschedule the notification when onSubscribe() is not invoked yet.

Result:

- Correct event order

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
@@ -22,11 +22,10 @@ import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -494,19 +501,13 @@ private void cleanup() {
...
-               (!(executor instanceof SingleThreadEventLoop) || !((EventLoop) executor).inEventLoop());
-    }
-
     private static final class SubscriptionImpl implements Subscription {
 
         private final DefaultStreamMessage<?> publisher;
         private Subscriber<Object> subscriber;
         private final Executor executor;
         private final boolean withPooledObjects;
+        private boolean invokedOnSubscribe; // Always updated and read from an event loop thread
```
anuraaga: Is this true for the case of executor == null?

trustin: No, if `executor == null`. Removed the comment. It doesn't have to be volatile though due to the same reason we did not for `inOnNext`.

trustin:
AppVeyor build completed successfully but it seems like build status notification is not working at the moment.

trustin:
Thank you for reviewing!

PR #849, opened 2017-11-21
Rename StreamMessage.closeFuture() to completionFuture() and revise Javadoc

trustin:
Motivation:

StreamMessage.closeFuture() has a confusing name because it conflicts
with the semantic of isOpen(). It is obviously possible that isOpen() is
false when closeFuture() is not complete.

Modifications:

- Add StreamMessage.completionFuture()
- Add StreamMessage.isComplete()
- Deprecate StreamMessage.closeFuture()
- Update the Javadoc of StreamMessage

Result:

- Less confusing API

/cc @anuraaga

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
@@ -34,13 +34,13 @@ import io.netty.util.ReferenceCounted;
  * <ul>
  *   <li>{@link #isOpen()}</li>
  *   <li>{@link #isEmpty()}</li>
- *   <li>{@link #closeFuture()}</li>
+ *   <li>{@link #completionFuture()}</li>
  *   <li>{@link #abort()}</li>
  * </ul>
  *
- * <h3>When is a {@link StreamMessage} open?</h3>
+ * <h3>When is a {@link StreamMessage} fully consumed?</h3>
  *
- * <p>A {@link StreamMessage} is open since its instantiation until:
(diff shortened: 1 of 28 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -88,10 +87,29 @@
     boolean isEmpty();
 
     /**
-     * Returns a {@link CompletableFuture} that completes when this publisher is complete,
+     * Return {@code true} if this stream is complete, either successfully or exceptionally,
```
anuraaga: Thanks, the naming was driving me a bit crazy.

Possibly here or elsewhere, it'd be good to define `complete`, e.g., publisher has indicated it has finished writing by calling `close()` or `abort()`, and subscriber has either consumed all writes or cancelled.

trustin: Updated Javadoc. A little bit of duplication but better being user-friendly :-)

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -88,10 +87,46 @@
...
```
minwoox: nit: Returns?

trustin: Fixed :-)

trustin:
Build failure will be fixed by #848

trustin:
Thanks for reviewing!

PR #851, opened 2017-11-22
Update Gradle to 4.3.1

trustin:


site/build.gradle
```diff
@@ -50,7 +50,7 @@ task xref(group: 'Documentation',
         return srcDirs
     }
 
-    inputs.dir sourceDirs
+    inputs.files sourceDirs
```
trustin: Specifying multiple files with `dir()` has been deprecated in Gradle 4.3.

PR #846, opened 2017-11-20
Add a FixedStreamMessage which can be used to publish objects from a …

anuraaga:
…stream fixed at construction time.

It is very common for an `HttpRequest` or `HttpResponse` to be constructed with all its objects known, e.g. using the `respond` family of methods. When all the objects are known, the synchronization overhead of `DefaultStreamMessage` is very wasteful and we can optimize this common use case.

This adds `FixedStreamMessage`, which serves objects straight from an indexed array. Since per reactive streams spec, only one thread can be used when notifying subscribers, we don't need synchronization when accessing the array and can iterate through these very quickly.

In the future, I would like to

- Add extra-optimized `FixedStreamMessage` implementations for one or two object cases. Headers-only or Headers+content only are extremely common uses of `FixedStreamMessage` and it's worth removing the array for them.

Benchmarks

```
Benchmark                            (flowControl)  (num)                   (streamType)   Mode  Cnt        Score        Error  Units
StreamMessageBenchmark.jmhEventLoop          false      3         DEFAULT_STREAM_MESSAGE  thrpt   20  2447995.498 ±  76571.326  ops/s
StreamMessageBenchmark.jmhEventLoop          false      3             EVENT_LOOP_MESSAGE  thrpt   20  3998523.391 ±  65931.260  ops/s
StreamMessageBenchmark.jmhEventLoop          false      3           FIXED_STREAM_MESSAGE  thrpt   20  5244650.672 ± 118638.870  ops/s
StreamMessageBenchmark.jmhEventLoop          false      3  DEFERRED_FIXED_STREAM_MESSAGE  thrpt   20  3388117.744 ±  50573.722  ops/s
StreamMessageBenchmark.jmhEventLoop          false      5         DEFAULT_STREAM_MESSAGE  thrpt   20  1735894.712 ±  51988.264  ops/s
StreamMessageBenchmark.jmhEventLoop          false      5             EVENT_LOOP_MESSAGE  thrpt   20  2791565.534 ± 105005.742  ops/s
StreamMessageBenchmark.jmhEventLoop          false      5           FIXED_STREAM_MESSAGE  thrpt   20  3808046.875 ±  92248.136  ops/s
StreamMessageBenchmark.jmhEventLoop          false      5  DEFERRED_FIXED_STREAM_MESSAGE  thrpt   20  2702439.823 ±  76267.413  ops/s
[...]

PR #855, opened 2017-11-23
Add EventLoopStreamMessage, which serializes stream operations throug…

anuraaga:
…h an event loop.

As there was some back-and-forth, I have restricted this PR to only add `EventLoopStreamMessage` without using it anywhere. We can do the `DefaultHttp*` juggling in the next PR.

`EventLoopStreamMessage` is designed to have as few volatile reads as possible, going so far as to duplicate some fields to allow volatile reads in cold paths and non-volatile reads in hot paths (`state` and `isOpen` / `subscription` and `subscribed`). Notably, any `write` only has a single volatile read and store, while looping through objects has none at all.

The class is designed to be used when both the writer and reader thread are the same as the stream's. In such a case, there is no thread-switch overhead and we take advantage of it all being on the same thread to have almost no synchronization. For any other case, there will be heavy overhead from thread-switching - but Armeria is built on event loops, and it is extremely unlikely that the appropriate event loop cannot be used from writers and readers. Any case where it isn't possible probably isn't a performance sensitive application anyways.

This PR heavily refactors `DefaultStreamMessage` to allow logic-sharing between the two implementations, and any future ones (e.g., FixedStreamMessage #846 which I will rebase onto this one). One big design change is moving more logic out of `SubscriptionImpl` and into the publishers - by keeping `SubscriptionImpl` as a state-holder and method forwarder, it can be shared by any `StreamMessage` implementation.

An `AbstractStreamMessageTest` is also added to provide test cases for both implementations easily.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageAndWriter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageAndWriter.java
@@ -5,7 +5,7 @@
  * version 2.0 (the "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at:
  *
- *   http://www.apache.org/licenses/LICENSE-2.0
+ *   https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageAndWriter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageAndWriter.java
@@ -90,12 +90,12 @@ abstract class AbstractStreamMessageAndWriter<T> extends AbstractStreamMessage<T
     }
 
     /**
-     * Add an object to publish to the stream.
+     * Adds an object to publish to the stream.
      */
     abstract void addObject(T obj);
 
     /**
-     * Add an object to publish (of type {@code T} or an event (e.g., {@link CloseEvent},
+     * Adds an object to publish (of type {@code T} or an event (e.g., {@link CloseEvent},
      * {@link AwaitDemandFuture}) to the stream.
      */
     abstract void addObjectOrEvent(Object obj);
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageAndWriter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageAndWriter.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2017 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.common.stream;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.concurrent.CompletableFuture;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.ReferenceCounted;
+
+abstract class AbstractStreamMessageAndWriter<T> extends AbstractStreamMessage<T>
+        implements StreamMessageAndWriter<T> {
+
+    static final CloseEvent SUCCESSFUL_CLOSE = new CloseEvent(null);
+
+    enum State {
+        /**
+         * The initial state. Will enter {@link #CLOSED} or {@link #CLEANUP}.
+         */
+        OPEN,
+        /**
+         * {@link #close()} or {@link #close(Throwable)} has been called. Will enter {@link #CLEANUP} after
+         * {@link org.reactivestreams.Subscriber#onComplete()} or
+         * {@link org.reactivestreams.Subscriber#onError(Throwable)} is invoked.
+         */
+        CLOSED,
+        /**
+         * Anything in the queue must be cleaned up.
+         * Enters this state when there's no chance of consumption by subscriber.
+         * i.e. when any of the following methods are invoked:
+         * <ul>
+         *   <li>{@link org.reactivestreams.Subscription#cancel()}</li>
+         *   <li>{@link #abort()} (via {@link AbortingSubscriber})</li>
+         *   <li>{@link org.reactivestreams.Subscriber#onComplete()}</li>
+         *   <li>{@link org.reactivestreams.Subscriber#onError(Throwable)}</li>
+         * </ul>
+         */
+        CLEANUP
+    }
+
+    @Override
+    public boolean write(T obj) {
+        requireNonNull(obj, "obj");
+        if (obj instanceof ReferenceCounted) {
+            ((ReferenceCounted) obj).touch();
+            if (!(obj instanceof ByteBufHolder) && !(obj instanceof ByteBuf)) {
+                throw new IllegalArgumentException(
+                        "can't publish a ReferenceCounted that's not a ByteBuf or a ByteBufHolder: " + obj);
+            }
+        }
+
+        if (!isOpen()) {
+            ReferenceCountUtil.safeRelease(obj);
+            return false;
+        }
+
+        addObject(obj);
+        return true;
+    }
+
+    @Override
+    public CompletableFuture<Void> onDemand(Runnable task) {
+        requireNonNull(task, "task");
+
+        final AwaitDemandFuture f = new AwaitDemandFuture();
+        if (!isOpen()) {
+            f.completeExceptionally(ClosedPublisherException.get());
+            return f;
+        }
+
+        addObjectOrEvent(f);
+        return f.thenRun(task);
+    }
+
+    /**
+     * Add an object to publish to the stream.
+     */
+    abstract void addObject(T obj);
+
+    /**
+     * Add an object to publish (of type {@code T} or an event (e.g., {@link CloseEvent},
+     * {@link AwaitDemandFuture}) to the stream.
+     */
+    abstract void addObjectOrEvent(Object obj);
+
+    static final class AwaitDemandFuture extends CompletableFuture<Void> {}
+}
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
@@ -25,23 +25,10 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
-import java.util.function.Supplier;
 
 import org.reactivestreams.Subscriber;
-import org.reactivestreams.Subscription;
-
-import com.google.common.base.MoreObjects;
 
 import com.linecorp.armeria.common.Flags;
-import com.linecorp.armeria.common.util.Exceptions;
-
-import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufHolder;
-import io.netty.buffer.ByteBufUtil;
-import io.netty.buffer.Unpooled;
-import io.netty.channel.EventLoop;
-import io.netty.util.ReferenceCountUtil;
-import io.netty.util.ReferenceCounted;
 
 /**
  * A {@link StreamMessage} which buffers the elements to be signaled into a {@link Queue}.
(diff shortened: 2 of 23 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,402 @@
...
+
+/**
+ * A {@link StreamMessage} optimized for when writes and reads all happen on the provided {@link EventLoop},
+ * removing the need for most synchronization from the hot path. This should satisfy standard cases when using
+ * Armeria. It is not required for writes or reads to use the provided {@link EventLoop} but in such a case, it
+ * will be significantly faster to use {@link DefaultStreamMessage}.
+ *
+ * <p>Note that when {@link Subscription#cancel()} or {@link #abort()} are called from a different thread, the
+ * stream will continue to signal objects until demand is satisfied, rather than stopping in the middle. If this
+ * is an issue, it is recommended to use {@link DefaultStreamMessage}.
```
trustin: Is there any reason we can't stop in the middle?

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,402 @@
...
+ *
+ * <p>Methods in this class prefixed with 'do' must be run on the stream's event loop.
```
trustin: This is an internal comment and thus should not be part of Javadoc.

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,402 @@
...
+    public EventLoopStreamMessage() {
+        this(CommonPools.workerGroup().next());
+        logger.debug("Creating EventLoopStreamMessage without specifying EventLoop. This will be very slow.");
+    }
+
+    /**
+     * Creates a new {@link EventLoopStreamMessage} which executes all writes on the provided {@link EventLoop}.
+     */
+    public EventLoopStreamMessage(EventLoop eventLoop) {
+        this.eventLoop = requireNonNull(eventLoop, "eventLoop");
```
trustin: Could we determine the event loop rather at subscription time? e.g.

```java
public void subscribe(Subscriber ...) { // no executor specified
    EventLoop eventLoop = RequestContext.mapCurrent(
            RequestContext::eventLoop, () -> CommonPools.workerGroup().next());
}
```

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,402 @@
...
+                AwaitDemandFuture f = (AwaitDemandFuture) queue.remove();
+                f.complete(null);
+                continue;
+            }
+
+            demand--;
+
+            @SuppressWarnings("unchecked")
+            T obj = (T) queue.remove();
+            notifySubscriberOfObject(subscription, obj);
```
trustin: `doNotifySubscriberOfObject()`? (no need to check if direct invocation is necessary?)

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageAndWriter.java
```diff
@@ -0,0 +1,24 @@
...
+ * under the License.
+ */
+
+package com.linecorp.armeria.common.stream;
+
+/**
+ * A type which is both a {@link StreamMessage} and a {@link StreamWriter}. This type is mainly used by tests
+ * which need to exercise both functionality.
+ */
+public interface StreamMessageAndWriter<T> extends StreamMessage<T>, StreamWriter<T> {
```
trustin: Probably OK to make this package-local?

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
anuraaga: I renamed the method since indeed it does need to be called in the event loop to set `inOnNext`. But either way, we do need to check direct invocation since the stream's event loop may be different from the subscriber's. Note that we are not executing any code outside the subscriber callbacks on the subscriber's event loop - it's all the stream's.

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageAndWriter.java
trustin: SGTM

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
trustin: What if we use an MPSC queue from jctools? Not as fast as `ArrayDequeue` but may be good enough for low-contention?

trustin:
> One thing to keep in mind is that `EventLoopStreamMessage` is tied closer to the writer rather than the reader (e.g., all code runs on its event loop, which is presumed to be the writer's). This is different from `DefaultStreamMessage` where all code runs on the subscriber's executor.

Ah, I see. Then it sounds natural to specify an `EventLoop` at construction time.  Although it'd not be good in terms of encapsulation, I can't think of a better way than getting the current event loop from `RequestContext.mapCurrent()`. Otherwise it would be very easy to create a suboptimal stream.

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
trustin: Read your comment after writing this comment. Now I understand why it's a good idea to specify an `EventLoop` at construction time. See [my comment](https://github.com/line/armeria/pull/855#issuecomment-346567012).

anuraaga:
I thought there might not be much use case of instantiating a stream message directly, as opposed to httprequest / response, but I guess there might be - went ahead and added the RequestContext check here.

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
trustin: OK. :-)

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,428 @@
...
+
+    /**
+     * Creates a new {@link EventLoopStreamMessage} which executes all writes on an arbitrary {@link EventLoop}.
+     * It is highly recommended to use {@link #EventLoopStreamMessage(EventLoop)} instead to allow writes to
+     * happen on the same {@link EventLoop} as this stream's.
+     */
+    public EventLoopStreamMessage() {
+        this(RequestContext.mapCurrent(RequestContext::eventLoop, () -> {
+            logger.debug(
+                    "Creating EventLoopStreamMessage without specifying EventLoop. This will be very slow.");
```
trustin: - will be very slow if subscriber runs in a different EventLoop?
- This can produce a lot of log messages. Could you log it for the first-time only? We could keep the stack trace of problematic calls and log only when the warning message was not logged before at the same call stack. (Just like Netty's leak detector suppresses duplicate log messages for leaks)

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -224,62 +140,97 @@ private static void failLateSubscriber(SubscriptionImpl subscription, Subscriber
...
-        pushObject(f);
-        return f.thenRun(task);
+    @Override
+    void cancel() {
+        cancelOrAbort(true);
     }
 
-    private void pushObject(Object obj) {
+    private void cancelOrAbort(boolean cancel) {
+        if (setState(State.OPEN, State.CLEANUP)) {
```
anuraaga: By the way, this code is the major change for `DefaultStreamMessage` in this PR. I noticed that `cancelRequested` was not set when entering this block - I couldn't think of a reason for that behavior, so I set it always when `Subscription.cancel` is called now. Also, `cancelRequested` got overwritten by `abort`, which also didn't seem correct to me. Hopefully this new flow makes sense but let's look at it carefully.

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
anuraaga: Hmm - not very experienced with recognizing stack traces, let me know if my approach works :)

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,440 @@
...
+    private volatile boolean wroteAny;
+
+    /**
+     * Creates a new {@link EventLoopStreamMessage} which executes all writes on an arbitrary {@link EventLoop}.
+     * It is highly recommended to use {@link #EventLoopStreamMessage(EventLoop)} instead to allow writes to
+     * happen on the same {@link EventLoop} as this stream's.
+     */
+    public EventLoopStreamMessage() {
+        this(RequestContext.mapCurrent(RequestContext::eventLoop, () -> {
+            Throwable t = new Throwable();
```
trustin: Could we define a dedicated exception?

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
trustin: Seems OK to me. To be honest, I'm not sure why such a field is necessary at all. I don't remember why it was needed either. It just looks cleaner now.

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
anuraaga: Yep done

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,446 @@
...
+    private volatile boolean wroteAny;
+
+    /**
+     * Creates a new {@link EventLoopStreamMessage} which executes all writes on an arbitrary {@link EventLoop}.
+     * It is highly recommended to use {@link #EventLoopStreamMessage(EventLoop)} instead to allow writes to
+     * happen on the same {@link EventLoop} as this stream's.
+     */
+    public EventLoopStreamMessage() {
+        this(RequestContext.mapCurrent(RequestContext::eventLoop, () -> {
+            UnexpectedEventLoopException e = new UnexpectedEventLoopException();
```
hyangtack: nit: `final`

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,446 @@
...
+                doNotifySubscriberWithCloseEvent(subscription, (CloseEvent) o);
+                break;
+            }
+
+            if (demand == 0) {
+                break;
+            }
+
+            if (o instanceof AwaitDemandFuture) {
+                AwaitDemandFuture f = (AwaitDemandFuture) queue.remove();
```
hyangtack: nit: `final`

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,446 @@
...
+        }
+        demand = newDemand;
+
+        if (oldDemand == 0) {
+            doNotifySubscriberIfNotEmpty();
+        }
+    }
+
+    private void doAddObject(T obj) {
+        SubscriptionImpl subscription = this.subscription;
```
hyangtack: nit: could we move this line to the inside of the following if-clause?

trustin:
Build failure:

```
EventLoopStreamMessageVerification.required_completionFutureMustCompleteOnTermination0: FAILURE
Gradle suite > Gradle test > com.linecorp.armeria.common.stream.EventLoopStreamMessageVerification.required_completionFutureMustCompleteOnTermination0 FAILED
    java.lang.AssertionError: 
    Expecting
      <CompletableFuture[Incomplete]>
    to be completed
```

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,446 @@
...
+        } else {
+            subscription.executor().execute(() -> {
+                subscription.subscriber().onSubscribe(subscription);
+                eventLoop.execute(() -> invokedOnSubscribe = true);
+            });
+        }
+    }
+
+    private void doRequest(long n) {
+        long oldDemand = demand;
```
hyangtack: nit: `final`

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -0,0 +1,446 @@
...
+            eventLoop.execute(() -> doAddObject(obj));
+        }
+    }
+
+    @Override
+    void addObjectOrEvent(Object obj) {
+        if (eventLoop.inEventLoop()) {
+            doAddObjectOrEvent(obj);
+        } else {
+            eventLoop.execute(() -> doAddObjectOrEvent(obj));
```
hyangtack: Just curious. Is there any unexpected chance to make added objects mis-ordered? I imagined the scenario as follows:

1. Thread A(not a eventloop) calls addObjectOrEvent with obj 1. -> into the eventloop queue.
2. IO Event occurs on that eventloop. 
3. While the eventloop handles the IO event, addObjectOrEvent is called with obj 2. -> add obj 2 to the queue.
4. After handling IO event, eventloop handles pending tasks. -> add obj 1 to the queue.

I'm not sure if this kind of scenario is possible. Also I guess it's very rare even if possible.

anuraaga: I believe you mean obj2 is added from an event loop thread so gets added immediately? It means there is Thread A and Event Loop (e.g., Thread B) writing simultaneously to the stream. I don't think there's any way to guarantee ordering when multiple threads can write simultaneously to a queue. Admittedly, with an event loop, there is a wider window for where such a race could happen, but since any multithreaded writes still have the problem to some degree, I wouldn't worry about it.

hyangtack: Ok, it sounds reasonable. Thank you for answer. :-)

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageAndWriter.java
```diff
@@ -0,0 +1,104 @@
...
+            f.completeExceptionally(ClosedPublisherException.get());
+            return f;
+        }
+
+        addObjectOrEvent(f);
+        return f.thenRun(task);
+    }
+
+    /**
+     * Add an object to publish to the stream.
```
minwoox: nit: Adds

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageAndWriter.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    abstract void addObject(T obj);
+
+    /**
+     * Add an object to publish (of type {@code T} or an event (e.g., {@link CloseEvent},
```
minwoox: nit: Adds

core/src/test/java/com/linecorp/armeria/common/stream/EventLoopStreamMessageVerification.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public StreamMessage<Long> createAbortedPublisher(long elements) {
+        return createStreamMessage(elements, true);
+    }
+
+    // createStreamMessage creates a publisher that relies on onDemand for triggering writes - unfortunately
+    // means that it is not possible to have reads and writes to the stream happening synchronously in the same
+    // call tree, so this test passes regardless of whether DefaultStreamMessage actually correctly handles
```
minwoox: nit: EventLoopStreamMessage?

anuraaga:
Thanks - updated with the cleanups

anuraaga:
Rerunning the builds (different flakiness on each of travis and appveyor...)

trustin:
Thanks a lot, @anuraaga !

PR #856, opened 2017-11-25
Add a cache to parsed path+query / validation.

anuraaga:
Currently, we do relatively expensive processing, mainly sanitizing, of a HTTP path on every single request. However, it is common for servers to only ever handle a fixed set of urls - this is the case for POST-based APIs (including thrift, gRPC, and some JSON APIs) and `HttpFileService`. For such cases, it is very wasteful executing the exact same processing on every single request.

This PR adds a caching layer to path parsing. When it is determined that a path does not have unfixed entropy (here we check the presence of a query and `mapping.paramNames`), the parse result can be stored in cache, boosting performance for subsequent requests.

The memory impact should be pretty low since the strings will be shared with any other caches, such as the routing cache.

For #799 

Benchmarks. Cached parses are 20x faster (I was originally sad thinking it was only 2x faster but then noticed the extra decimal place :) ).
```
Benchmark                                               Mode  Cnt        Score        Error  Units
PathParsingBenchmark.cachedAndNotCached                thrpt   20   521946.368 ±   9936.017  ops/s
PathParsingBenchmark.cachedAndNotCached_cacheDisabled  thrpt   20   346507.550 ±   7853.314  ops/s
PathParsingBenchmark.normal                            thrpt   20  7675720.727 ± 108896.234  ops/s
PathParsingBenchmark.normal_cacheDisabled              thrpt   20   576549.419 ±   8253.508  ops/s
```


core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -0,0 +1,204 @@
...
+     *
+     * <p>Should allow the asterisk character in the path, query, or fragment components of a URL(RFC2396).
+     * @see <a href="https://tools.ietf.org/html/rfc3986#section-3.3">RFC 3986, section 3.3</a>
+     */
+    private static final Pattern PROHIBITED_PATH_PATTERN =
+            Pattern.compile("^/[^/]*:[^/]*/|[|<>\\\\]|/\\.\\.|\\.\\.$|\\.\\./");
+
+    private static final Pattern CONSECUTIVE_SLASHES_PATTERN = Pattern.compile("/{2,}");
+
+    private static final Cache<String, PathAndQuery> CACHE =
```
imasahiro: Not sure useful or not but how about exposing this cache stats to e.g. prometheus?

anuraaga:
@imasahiro Thanks, will let the stats optionally be exported if the user specifies recording in the cache spec.

I noticed that recording reduces performance of the cache by about 10% so don't want to enable it by default since many people will never even look at the numbers - what do you think? If we default recording to off I would want to make the same change to RouteCache

imasahiro:
@anuraaga it sounds reasonable. Will you have a chance to send a PR?

anuraaga:
Yeah I'm tweaking it right now - ran into https://github.com/ben-manes/caffeine/issues/207

anuraaga:
Updated to export metrics if they're enabled by the flag, will send a PR for the router cache separately.  Admittedly, wasn't sure of a good pattern for server-wide metrics, currently there is a similar static `CACHE` for virtualhost service lookup, but that's exported per virtualhost so I guess the same cache may get exported several times. I'm planning on moving that next to the line I added here.

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -0,0 +1,214 @@
...
+            Flags.parsedPathCacheSpec().map(PathAndQuery::buildCache).orElse(null);
+
+    private static Cache<String, PathAndQuery> buildCache(String spec) {
+        return Caffeine.from(spec).build();
+    }
+
+    public static void registerMetrics(MeterRegistry registry, MeterIdPrefix idPrefix) {
+        if (CACHE != null && CACHE.policy().isRecordingStats()) {
```
trustin: How about moving this check into `CaffeineMetricSupport.setup()` so that all Caffeine `Cache`s in Armeria are affected?

anuraaga: Great point - moved

anuraaga:
Oops updating tests now - felt like a method that wouldn't be affected but was wrong

anuraaga:
Updated, ready for another look

trustin:
Great job!

Issue #850, opened 2017-11-21
[Proposal] Change API to encourage using HttpResponse instead of HttpResponseWriter


anuraaga:
Currently we have a convenience abstract class, `AbstractHttpService` which takes care of HTTP request method dispatch to user service methods. The user methods have a signature like

```java
protected void doGet(ServiceRequestContext ctx, HttpRequest req, HttpResponseWriter res);
```

where `res` is always an instance of `DefaultHttpResponse`.

I propose we consider changing these to not pass in a `HttpResponse` implementation and leave it to the user, i.e.,

```java
protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req);
```

Reasons for this change are

#### Having `HttpResponseWriter` passed in isn't actually all that convenient. 

Consider a common pattern

```java
if (!checkContentType()) {
  res.respond(BAD_CONTENT_TYPE);
  return;
}
if (!checkArgs()) {
  res.respond(BAD_REQUEST);
  return;
}
res.respond(aggregatedHttpMessage);
```

is actually more lines than
```java
if (!checkContentType()) {
  return HttpResponse.of(BAD_CONTENT_TYPE);
}
if (!checkArgs()) {
  return HttpResponse.of(BAD_REQUEST);
}
return HttpResponse.of(aggregatedHttpMessage);
```

and if the user actually does need a streaming response, it's just two lines to construct a streaming-type response and return it. I think in practice, passing in the writer rarely wins and often loses in terms of code complexity.

#### Provides more flexibility

As mentioned on #846, it is good to be able to select response implementation based on what the response actually is, often for performance reasons. For example, in the above examples all `HttpResponse.of` could be implemented using a non-streaming, optimized version of `HttpResponse`, without the user having to know the details. This allows us to experiment with implementations in the background without being tied to user code. All we needed was an `HttpResponse`, but restricting this interface to `HttpResponseWriter` cuts down on options, and as described above doesn't seem to have much benefit.

#### Encourages good practices

One option would be to not make this change and expect users that want to be able to select `HttpResponse` implementation to implement `HttpService`. But as method dispatch logic in `AbstractHttpService` is quite useful in most cases, many users will probably continue to use it and miss out on any work we do on `HttpResponse`. In particular, I suspect that most user code will be optimized if they switch from `HttpResponseWriter` to `DeferredHttpResponse` + `FixedHttpResponse`.

#### Removes some duplication

Currently, we have this comment, and annoying duplication of APIs among many objects

```java
[...]

Issue #818, opened 2017-11-03
HTTP/2 connection pool can create extra connections unnecessarily


trustin:
When an HTTP/2 connection is acquired from a `KeyedChannelPool`, it is removed from the pool:

- https://github.com/line/armeria/blob/armeria-0.54.1/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java#L128

and then added back after the initial request headers are written:

- https://github.com/line/armeria/blob/armeria-0.54.1/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java#L168

When another client makes a request at the same time, a new HTTP/2 connection may be created unnecessarily, because the connection acquired for the first request does not exist in the pool.

Instead of removing the connection completely from the pool's data structure, we could count the number of active requests for each connection and return the one with the least number of active requests, which is similar to what `EventLoopScheduler` does.

For even more efficiency, we could merge `EventLoopScheduler` and `DefaultKeyedChannelPool` into one and defer the selection of `EventLoop` as late as possible. Then, we won't need to count per-`EventLoop` active requests at all too choose the most idle `EventLoop`; we can just find the connection with the least active requests and that will get us the right `EventLoop`.

Related discussion: https://github.com/line/armeria/issues/816

anuraaga:
I haven't ever used `ClientRequestContext.eventLoop` ever. Since my client is always used from a server handler, I would always schedule a client's callbacks onto the server event loops. Not worrying about breaking the API SGTM.

By the way, did the event loops scheduler change remove the propagation of server event loop to client context? I can't find where the server loop is propagated to the client anymore.

trustin:
@anuraaga It's in `HttpClientFactory.eventLoopSupplier`.

anuraaga:
Right but I couldn't find it being used by the client. Since client (`UserClient` or `ArmeriaChannel`) use `acquireEventLoop`, don't they ignore `eventLoopSupplier` / `RequestContext.current()` entirely?

trustin:
@anuraaga Indeed. We lost that optimization when we introduced `EventLoopScheduler`. Let me try to get it back while resolving this issue.

trustin:
Related: #800

imasahiro:
The breaking API sounds good to me too. 
(And actually, I faced extra connection when I write a batch that calls http2+thrift (and it causes "Too many open files" error..). So it would be happy reducing connection as much as possible)

anuraaga:
[...]

PR #857, opened 2017-11-25
[WIP] Switch the default StreamMessage to EventLoopStreamMessage

trustin:
Findings so far:

- There are test failures that need investigation.
- Sometimes we just don't know from which `EventLoop` we will publish the events, especially for client-side. i.e. It may be hard to specify an `EventLoop` at construction time.
  - Loggging when a user does not specify an `EventLoop` may not be a good idea.
  - Choosing from `CommonPools.workerGroup()` may not work either because the client will use a different `EventLoop`.

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/common/stream/StreamMessageBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/common/stream/StreamMessageBenchmark.java
@@ -54,8 +54,8 @@ public class StreamMessageBenchmark {
     public static class StreamObjects {
 
         public enum StreamType {
-            DEFAULT_STREAM_MESSAGE,
-            EVENT_LOOP_MESSAGE,
+            CONCURRENT,
+            EVENT_LOOP,
         }
 
         @Param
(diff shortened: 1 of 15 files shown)


anuraaga:
Would https://github.com/line/armeria/issues/818 help with the client side, by restoring the client's reuse of the server event loop?

trustin:
I'm not sure. #818 rather delays the determination of `EventLoop`. `HttpRequest` is constructed even before the invocation starts.

trustin:
> by restoring the client's reuse of the server event loop?

We can't make a client always reuse the server event loop unfortunately, as explained [here](https://github.com/line/armeria/issues/818#issuecomment-346937485).

anuraaga:
Ah - yeah with a few dozen endpoints and a dozen threads, it does cause the connection count to balloon a lot. I guess it might hinder us from using `EventLoopStreamMessage` for clients without some other innovation.

One insight from a quick scan of the logs, I guess we'll want `FixedHttpResponse` as well, which will eliminate all the `AggregatedHttpMessage.toHttpRequest` usages that are missing the event loop.

I also feel, as a part of #850, we should try to deprecate and make private `DefaultHttpResponse` (or `ConcurrentHttpResponse` as it might be called later) rather than trying to just slot the event loop version in. So something like

```
interface HttpResponse {
  // If you have a RequestContext, you probably want to use streaming(RequestContext) to make sure it's used!
  HttpResponseWriter streaming() {
    return RequestContext.mapCurrent(HttpResponse::streaming, ConcurrentHttpResponse::new);
  }
  
  HttpResponseWriter streaming(RequestContext ctx) {
    return new EventLoopHttpResponse(ctx);
  }
}
```

And update `AbstractHttpService` to first replace `new DefaultHttpResponse()` with `HttpResponse.streaming(ctx);`. Eventually, we'll hopefully change the signature and not do either of these though.

Similar refactoring for the request-side. Hiding the implementation will allow us to do whatever-we-want with the stream whenever we want, hopefully eventually finding the true optimal solution :D

How does that sound?

trustin:
Closing for now. Will revisit when time allows.

PR #861, opened 2017-11-27
Allow null value as a return value of annotated http service methods

hyangtack:
Motivation:
When an annotated method wants to return null, now it should return an empty object instead because returning null would make NPE.

Modifications:
- Find a converter which handles Object.class for the return value of null.
- Avoid NPE.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
@@ -371,7 +371,7 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
      */
     @Nullable
     private static String httpParameterValue(HttpParameters httpParameters, Parameter entry) {
-        String value = httpParameters.get(entry.name());
+        final String value = httpParameters.get(entry.name());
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -524,20 +525,22 @@ private static ResponseConverter findResponseConverter(
...
 
         // Search for the converter mapped to itself or one of its superclasses, except Object.class.
-        Class<?> current = type;
-        while (current != Object.class) {
-            ResponseConverter converter = converters.get(current);
-            if (converter != null) {
-                return converter;
+        if (type != Object.class) {
+            Class<?> current = type;
+            while (current != Object.class) {
```
trustin: Can do `do { ... } while ((current = current.getSuperclass()) != Object.class);` instead

hyangtack: Fixed!

trustin:
AppVeyor failure irrelevant

trustin:
Thanks, @hyangtack !

PR #862, opened 2017-11-27
Treat RpcResponse.cause as log's cause if not explicitly set.

anuraaga:
In `GrpcService`, I call `responseContent` but not `endResponse`, and as a result `LoggingService` never sees any exception. While I could call `endResponse` as well, it was very unintuitive to me that setting a failure response wasn't enough to log the error, and feels repetitive to have to call twice with the same exception. WDYT?


trustin:
@anuraaga Suggestion sounds good to me. Would you mind adding a test case for this?

trustin:
Thanks, @anuraaga !

PR #863, opened 2017-11-27
Miscellaneous StreamMessage improvements

trustin:
- Change the return type of SubscriptionImpl.executor() to EventExecutor
- Make DefaultStreamMessage.demand non-volatile
- Make sure 'inOnNext' is reset to false even if Subscriber.onNext()
  raises an exception
- Simplify doRequest()
- No noticeable performance regression or improvement


trustin:
Thanks!

PR #866, opened 2017-11-28
Add request converter to annotated http service

hyangtack:
Motivation:
When we are handling HTTP POST request, we usually convert its body to a Java object by some converting logic like JSON message converter. It is a general use-case for handling a request because a Java object is more convenient than a raw HTTP request itself. So it would be better if we have a request object mapping feature like Spring message converter.

Modifications:
- Add `@RequestConverter` annotation to specify a request converter of a method or a class. It is also repeatable.
- Add RequestConverterFunction interface to implement a request converter class.
- Add `@RequestObject` annotation to specify which parameter should be converted by a request converter.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
@@ -56,6 +56,8 @@ import com.linecorp.armeria.server.annotation.Default;
 import com.linecorp.armeria.server.annotation.ExceptionHandlerFunction;
 import com.linecorp.armeria.server.annotation.Header;
 import com.linecorp.armeria.server.annotation.Param;
+import com.linecorp.armeria.server.annotation.RequestConverterFunction;
+import com.linecorp.armeria.server.annotation.RequestObject;
 import com.linecorp.armeria.server.annotation.ResponseConverter;
 
 import io.netty.handler.codec.http.HttpConstants;
@@ -76,16 +78,20 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
     private final boolean isAsynchronous;
     private final AggregationStrategy aggregationStrategy;
     private final List<ExceptionHandlerFunction> exceptionHandlers;
+    private final List<RequestConverterFunction> requestConverters;
 
     AnnotatedHttpServiceMethod(Object object, Method method, PathMapping pathMapping,
-                               List<ExceptionHandlerFunction> exceptionHandlers) {
+                               List<ExceptionHandlerFunction> exceptionHandlers,
+                               List<RequestConverterFunction> requestConverters) {
         this.object = requireNonNull(object, "object");
         this.method = requireNonNull(method, "method");
         requireNonNull(pathMapping, "pathMapping");
         this.exceptionHandlers = ImmutableList.copyOf(
                 requireNonNull(exceptionHandlers, "exceptionHandlers"));
+        this.requestConverters = ImmutableList.copyOf(
+                requireNonNull(requestConverters, "requestConverters"));
 
-        parameters = parameters(method, pathMapping.paramNames());
+        parameters = parameters(method, pathMapping.paramNames(), !requestConverters.isEmpty());
         final Class<?> returnType = method.getReturnType();
         isAsynchronous = HttpResponse.class.isAssignableFrom(returnType) ||
                          CompletionStage.class.isAssignableFrom(returnType);
@@ -170,7 +176,7 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
 
     private Object invoke(ServiceRequestContext ctx, HttpRequest req, @Nullable AggregatedHttpMessage message) {
         try (SafeCloseable ignored = RequestContext.push(ctx, false)) {
-            return method.invoke(object, parameterValues(ctx, req, parameters, message));
+            return method.invoke(object, parameterValues(ctx, req, parameters, message, requestConverters));
         } catch (Throwable cause) {
             final HttpResponse response = getResponseForCause(ctx, req, cause);
             if (response != null) {
@@ -202,7 +208,8 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
     /**
      * Returns the array of {@link Parameter}, which holds the type and {@link Param} value.
      */
-    private static List<Parameter> parameters(Method method, final Set<String> pathParams) {
+    private static List<Parameter> parameters(Method method, Set<String> pathParams,
+                                              boolean hasRequestConverters) {
         requireNonNull(pathParams, "pathParams");
         boolean hasRequestMessage = false;
         final ImmutableList.Builder<Parameter> entries = ImmutableList.builder();
@@ -227,6 +234,14 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
             final Header header = parameterInfo.getAnnotation(Header.class);
             if (header != null) {
                 entries.add(createOptionalSupportedParam(parameterInfo, ParameterType.HEADER, header.value()));
+                continue;
+            }
+
+            if (parameterInfo.isAnnotationPresent(RequestObject.class)) {
+                if (!hasRequestConverters) {
+                    throw new IllegalArgumentException("No request converter for method: " + method.getName());
+                }
+                entries.add(Parameter.ofRequestObject(parameterInfo.getType()));
                 continue;
             }
 
@@ -279,7 +294,8 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
      */
     private static Object[] parameterValues(ServiceRequestContext ctx, HttpRequest req,
                                             List<Parameter> parameters,
-                                            @Nullable AggregatedHttpMessage message) {
+                                            @Nullable AggregatedHttpMessage message,
+                                            List<RequestConverterFunction> requestConverters) {
         HttpParameters httpParameters = null;
         Object[] values = new Object[parameters.size()];
         for (int i = 0; i < parameters.size(); ++i) {
@@ -314,6 +330,25 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
                             httpParameters = httpParametersOf(ctx, req, message);
                         }
                         values[i] = httpParameters;
+                    }
+                    break;
+                case REQUEST_OBJECT:
+                    for (final RequestConverterFunction func : requestConverters) {
+                        if (func.accept(message, entry.type())) {
+                            try {
+                                values[i] = func.convert(message, entry.type());
+                                break;
+                            } catch (Exception e) {
+                                // Go to the next request converter if the conversion is failed.
+                                logger.warn("Request converter {} cannot convert a request to {}",
+                                            func.getClass().getName(), entry.type().getName(), e);
+                            }
+                        }
+                    }
+                    if (values[i] == null) {
+                        throw new IllegalArgumentException(
+                                "No suitable request converter found for a @" +
+                                RequestObject.class.getSimpleName() + " '" + entry.name() + '\'');
                     }
                     break;
             }
@@ -595,6 +630,10 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
             return new Parameter(ParameterType.PREDEFINED_TYPE, true, false, type, null, null);
         }
 
+        static Parameter ofRequestObject(Class<?> type) {
+            return new Parameter(ParameterType.REQUEST_OBJECT, true, false, type, null, null);
+        }
+
         private final ParameterType parameterType;
         private final boolean isRequired;
         private final boolean isOptionalWrapped;
@@ -645,7 +684,7 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
     }
 
     private enum ParameterType {
-        PATH_PARAM, PARAM, HEADER, PREDEFINED_TYPE
+        PATH_PARAM, PARAM, HEADER, PREDEFINED_TYPE, REQUEST_OBJECT
     }
 
     private enum AggregationStrategy {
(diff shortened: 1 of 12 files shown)


hyangtack:
Please review the concept of the request converter first. :-)
If we use `@RequestConverter` annotation, we need to think about renaming the existing `@Converter` annotation which means a response converter. (Leave it as it is or rename)

hyangtack:
Also I'm thinking of changing `@Converter`'s behaviors as follows:

### AS-IS
- Able to specify multiple annotation on class:
```
@Converter(target = String.class, value = UnformattedStringConverter.class)
@Converter(DefaultConverter.class)
```
- Able to specify only **one** converter on a method:
```
@Converter(ConverterForSpecificMethod.class)
```
- Manages the converters as a `map`.

### TO-BE
- Changing its usage the same as `@RequestConverter` and`@ExceptionHandler`.
- Still repeatable.
- Do not specify target class as its `target` value.
- Manages the converters as a `list`. Use `accept()` method to check whether it is able to handle the result or not:
```
@FunctionalInterface
public interface ResponseConverter {

    /**
     * Returns whether this converter is able to convert the specified {@code result} to
     * {@link HttpResponse}.
     */
    default boolean accept(Object result) {
        return true;
    }

    /**
     * Returns {@link HttpResponse} instance corresponds to the given {@code resObj}.
     */
    HttpResponse convert(Object resObj) throws Exception;
}
```

Please feel free to leave any comment. :-)

trustin:
@hyangtack SGTM

hyangtack:
FYI, https://github.com/hyangtack/armeria/pull/1 includes the changes of response converter. After merging this PR, I will make another PR for that.

core/src/main/java/com/linecorp/armeria/server/annotation/RequestConverter.java
```diff
@@ -0,0 +1,44 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.linecorp.armeria.common.AggregatedHttpMessage;
+
+/**
+ * Specifies an {@link RequestConverterFunction} class which converts {@link AggregatedHttpMessage} to
```
minwoox: nit: an -> a?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -100,6 +102,13 @@
     private static final List<ExceptionHandlerFunction> defaultExceptionHandlers =
             ImmutableList.of(new DefaultExceptionHandler());
 
+    /**
+     * An {@link RequestConverterFunction} map for reusing request converters.
```
minwoox: nit: An -> A

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -316,6 +332,20 @@ private static Parameter createOptionalSupportedParam(java.lang.reflect.Paramete
...
                     break;
+                case REQUEST_OBJECT:
+                    for (final RequestConverterFunction<?> func : requestConverters) {
+                        if (func.accept(message, entry.type())) {
+                            try {
+                                values[i] = func.convert(message);
+                            } catch (Exception e) {
+                                throw new IllegalArgumentException(
+                                        "Request converter " + func.getClass().getName() +
+                                        " cannot convert a request to " + entry.type().getName() + ':', e);
```
trustin: Drop `+ ':'` or append the cause to the message?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -286,19 +295,15 @@ private static String findPattern(Method method, Set<Annotation> methodAnnotatio
...
-                pattern = p;
-            } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
-                logger.debug("Invoking annotation 'value' method failed: {}", e);
+            final String p = (String) invokeValueMethod(a);
+            if (DefaultValues.isUnspecified(p)) {
+                continue;
+            }
+            if (pattern != null) {
+                throw new IllegalArgumentException(
+                        "Only one path can be specified. (" + pattern + ", " + p + ')');
```
trustin: Could use `checkArgument()`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -384,20 +389,60 @@ private static ResponseConverter converter(Method method) {
...
+                                                                       Class<?> targetClass) {
+        return objectsOf(RequestConverter.class, targetMethod, targetClass,
+                         requestConverterFunctions, ImmutableList.of());
+    }
+
+    /**
+     * Returns a list of objects which are specified as the {@code value} of annotation {@code T}.
+     */
+    @SuppressWarnings("unchecked")
+    private static <T extends Annotation, R> List<R> objectsOf(
```
trustin: How about `annotationValues()`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -384,20 +389,60 @@ private static ResponseConverter converter(Method method) {
...
 
+    /**
+     * Returns an object which is returned by {@code value()} method of the specified object {@code a}.
+     */
+    private static Object invokeValueMethod(Annotation a) {
+        try {
+            return a.getClass().getMethod("value").invoke(a);
+        } catch (Exception e) {
+            throw new IllegalStateException("An annotation @" + a.getClass().getSimpleName() +
+                                            " must have an 'value' method:", e);
```
trustin: - an -> a
- Drop ':' or append the cause

core/src/main/java/com/linecorp/armeria/server/annotation/RequestConverter.java
```diff
@@ -0,0 +1,44 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.linecorp.armeria.common.AggregatedHttpMessage;
+
+/**
+ * Specifies a {@link RequestConverterFunction} class which converts {@link AggregatedHttpMessage} to
```
trustin: converts -> converts an

core/src/main/java/com/linecorp/armeria/server/annotation/RequestConverterFunction.java
```diff
@@ -0,0 +1,43 @@
...
+ * under the License.
+ */
+
+package com.linecorp.armeria.server.annotation;
+
+import com.linecorp.armeria.common.AggregatedHttpMessage;
+
+/**
+ * Converts an {@link AggregatedHttpMessage} to an object. The class implementing this interface would
+ * be specified as {@link RequestConverter} annotation.
```
trustin: as -> as a value of a

trustin:
@kojilin Any comments?

kojilin:
Can convert contains type? 1 thing want to know is can I write generic converter like
```
public static class TestRequestConverter3 implements RequestConverterFunction<Object> {
        private static final ObjectMapper mapper = new ObjectMapper();

        @Override
        public boolean accept(AggregatedHttpMessage request, Class<?> expectedResultType) {
            return true;
        }

        @Override
        public Object convert(AggregatedHttpMessage request, Class<?> expectedResultType) throws Exception {
            return mapper.readValue(request.content().toStringUtf8(), expectedResultType);
        }
    }

```
So I can design just 1 converter for almost use case. Or is there any other way to achieve it?

core/src/main/java/com/linecorp/armeria/server/annotation/RequestConverterFunction.java
```diff
@@ -0,0 +1,43 @@
...
+     * {@code expectedResultType}.
+     */
+    default boolean accept(AggregatedHttpMessage request, Class<?> expectedResultType) {
+        return true;
+    }
+
+    /**
+     * Converts the specified {@code request} to an object.
+     */
+    T convert(AggregatedHttpMessage request) throws Exception;
```
trustin: Idea by @kojilin - could we add `expectedResultType` to `convert()`?

```java
T convert(AggregatedHttpMessage request, Class<T> expectedResultType) throws Exception;
```

core/src/main/java/com/linecorp/armeria/server/annotation/RequestConverterFunction.java
```diff
@@ -0,0 +1,43 @@
...
+ * @see RequestObject
+ */
+@FunctionalInterface
+public interface RequestConverterFunction<T> {
+
+    /**
+     * Returns whether this converter is able to convert the specified {@code request} to
+     * {@code expectedResultType}.
+     */
+    default boolean accept(AggregatedHttpMessage request, Class<?> expectedResultType) {
```
trustin: `Class<?>` -> `Class<T>` ?

hyangtack:
@kojilin Sorry for late response. I fixed the following things:

- Remove `<T>` from `RequestConverterFunction`.
- Fix `convert` method signature.
- Add some default request converters such as: `JacksonRequestConverterFunction` and `ByteArrayRequestConverterFunction`.

Would you please do a test again with the latest codes? :-)

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    @Override
+    public Object convert(AggregatedHttpMessage request, Class<?> expectedResultType) throws Exception {
+        final ObjectReader reader = readers.get(expectedResultType);
+        assert reader != null;
+        final String contentType = request.headers().get(HttpHeaderNames.CONTENT_TYPE);
+        assert contentType != null;
+
+        final Charset charset = MediaType.parse(contentType).charset()
+                                         .orElse(StandardCharsets.ISO_8859_1);
```
kojilin: should be `UTF-8` ?

hyangtack: I found the following sentence from [here](https://www.w3.org/International/articles/http-charset/index.en). So I used ISO-8859-1 as the default charset. 
> HTTP 1.1 says that the default charset is ISO-8859-1.

hyangtack: I think you are right. UTF-8 is a default charset for JSON. Thanks!

ISO-8859-1 is a default charset for "text" type.
https://tools.ietf.org/html/rfc2616#section-3.7.1
> The "charset" parameter is used with some media types to define the
   character set (section 3.4) of the data. When no explicit charset
   parameter is provided by the sender, media subtypes of the "text"
   type are defined to have a default charset value of "ISO-8859-1" when
   received via HTTP. Data in character sets other than "ISO-8859-1" or
   its subsets MUST be labeled with an appropriate charset value. See
   section 3.4.1 for compatibility problems.

hyangtack: Fixed!

hyangtack:
@trustin @kojilin @minwoox Would you have a look again especially for the default `RequestConverterFunction`s?

Name | Acceptable MIME types | Injected Type
------ | ----------------------- | --------------
`JacksonRequestConverterFunction` | `application/json` | `Object`
`StringRequestConverterFunction` | `text/*` | `String`
`ByteArrayRequestConverterFunction` | `application/octet-stream`, `application/binary` or no `Content-Type` header | `HttpData` or `byte[]`

Please let me know if there is not reasonable specification. :-)

core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
```diff
@@ -0,0 +1,65 @@
...
+            return true;
+        }
+
+        final MediaType mediaType = MediaType.parse(contentType);
+        return mediaType.is(MediaType.OCTET_STREAM) ||
+               mediaType.is(MediaType.APPLICATION_BINARY);
+    }
+
+    /**
+     * Converts the specified {@link AggregatedHttpMessage} to an object of {@code expectedResultType} type.
```
minwoox: nit: just {@code expectedResultType}. ?? can remove the type at the very end.

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -0,0 +1,95 @@
...
+                logger.warn(JacksonRequestConverterFunction.class.getName() +
+                            " cannot read a JSON of '" + request.content().toStringUtf8() +
+                            "' as a type '" + expectedResultType.getName() + '\'', cause);
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Converts the specified {@link AggregatedHttpMessage} to an object of {@code expectedResultType} type.
```
minwoox: ditto?

core/src/main/java/com/linecorp/armeria/server/annotation/RequestConverterFunction.java
```diff
@@ -0,0 +1,43 @@
...
+        return true;
+    }
+
+    /**
+     * Converts the specified {@code request} to an object of {@code expectedResultType} type.
```
minwoox: ditto?

kojilin:
just curious when to use `ByteArrayRequestConverterFunction` or `StringRequestConverterFunction `?

hyangtack:
> just curious when to use ByteArrayRequestConverterFunction or StringRequestConverterFunction?
When a user wants to get binary data as a variable? It's just to provide default converters like Spring.. not having a special purpose. :-)

trustin:
Great job, @hyangtack !

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -384,20 +399,60 @@ private static ResponseConverter converter(Method method) {
...
+                                                              type -> newInstance(type, annotationClass)))
                      .collect(toImmutableList());
     }
 
+    /**
+     * Returns an object which is returned by {@code value()} method of the specified object {@code a}.
+     */
+    private static Object invokeValueMethod(Annotation a) {
+        try {
+            return a.getClass().getMethod("value").invoke(a);
```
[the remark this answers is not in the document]
hyangtack: Actually it is used for several annotations in `com.linecorp.armeria.server.annotation` package. So about 10+ annotations would be `a`. :-) And it is called only when building annotated http service instances, so it'd not be a big problem. 
I will send a next PR about `Converter` annotation, so I can try the way you suggested in that PR. Thanks! :-)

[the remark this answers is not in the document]
anuraaga: Ah makes sense

PR #871, opened 2017-11-28
Fix buffer leak in Http(Encoded|Decoded)Response

trustin:
Handle the case where HttpData is a ByteBufHttpData.


trustin:
Closing in favor ot #872

PR #872, opened 2017-11-28
Make FilteredStreamMessage and FilteredHttpResponse understand buffer pooling

trustin:
Motivation:

FilteredStreamMessage currently passes pooled objects such as
ByteBufHttpData, ByteBufHolder and ByteBuf as they are, causing buffer
leaks when a user publishes them.

Modifications:

- Add a constructor parameter 'withPooledObjects' to
  FilteredStreamMessage and FilteredHttpResponse
- Extract the common code that converts a pooled object into an unpooled
  one into a new utility class 'PooledObjects'
- Add test cases for HttpEncodedResponse and ZlibStreamDecoder to verify
  the change

Result:

- Less leaks

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/encoding/ZlibStreamDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/encoding/ZlibStreamDecoder.java
@@ -19,6 +19,7 @@ package com.linecorp.armeria.client.encoding;
 import com.linecorp.armeria.common.HttpData;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.CompositeByteBuf;
 import io.netty.buffer.Unpooled;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/linecorp/armeria/client/encoding/ZlibStreamDecoder.java
```diff
@@ -40,6 +41,7 @@
 
     @Override
     public HttpData decode(HttpData obj) {
+        obj = PooledObjects.toUnpooled(obj);
```
anuraaga: How about writing directly if it's bytebuf?

trustin: That's a good idea. Fixed!

trustin:
> Can you add `FilteredStreamMessageTest extends AbstractStreamMessageTest` or maybe after the split I did in the fixed stream message test

OK. Let me add one once #846 is merged.

trustin:
Thanks for reviewing!

PR #874, opened 2017-11-28
Fix a race condition in PublisherBasedStreamMessage

trustin:
PublisherBasedStreamMessage may not invoke Subscriber.onSubscribe() on a
given situation:

1. Publisher invokes AbortableSubscriber.onSubscribe().
2. AbortableSubscriber.onSubscribe() sets 'subscription'.
3. AbortableSubscriber.onSubscribe() schedules onSubscribe0() that
   invokes user's Subscriber.onSubscribe() to run on an EventLoop.
4. abort() is requested from the EventLoop in step 3 before user's
   Subscriber.onSubscribe() is invoked by onSubscribe0(). The abortion
   (cancelOrAbort0(boolean)) is performed before onSubscribe0() of
   the step 3 is executed because abort() will be invoked directly.
5. The Subscriber is not notified because cancelOrAbort0() replaced
   the 'subscriber' field before onSubscribe0() is run.

This commit also contains the following changes:

- Fix a bug where required_abortMustNotifySubscriber() did not handle
  the case where abortion occurs before the first element is signaled
- Adjust the TestEnvironment timeout to the saner values.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/PublisherBasedStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/PublisherBasedStreamMessage.java
@@ -172,12 +172,17 @@ public class PublisherBasedStreamMessage<T> implements StreamMessage<T> {
 
         @Override
         public void request(long n) {
+            final Subscription subscription = this.subscription;
             assert subscription != null;
             subscription.request(n);
         }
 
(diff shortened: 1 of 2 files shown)


trustin:
Still failing. :-D

trustin:
Thanks for reviewing!

PR #878, opened 2017-11-29
Wait for refcnt assertions.

anuraaga:
Since `isOpen` is usually updated immediately, it's not a sufficient guard for item processing, so add await to the refcnt assertions too.


trustin:
Great! Thanks. :-)

PR #880, opened 2017-11-29
Fix flaky HttpHealthCheckedEndpointGroupTest

trustin:


trustin:
> I want a bytecode rewriter to just redirect all `assert` calls to awaitility ;)

😆

PR #889, opened 2017-12-05
Add specialized implementations of fixed stream message to reduce all…

anuraaga:
…ocations for common cases.

It's very common to allocate stream messages with 0-2 objects, so it's nice to have specializations to avoid the extra garbage of an object array. 

Unfortunately, the duplicate implementations introduce a lot of complexity. Let me know if you think it's worth it. `FixedHttpRequest` turned out much worse than I had originally expected.

The total allocation of each is
- Empty: 16 bytes (3 fields in base class)
- One element: 16 bytes (3 fields in base, one in sub)
- Two elements: 24 bytes (3 fields in base, 3 in sub)
- Three elements: 24 + 36 (3 fields in base, 3 in sub, 24 bytes array overhead, 12 bytes in array)

Also unrelated, removed `invokedOnSubscribe` as it's unnecessary for fixed streams which have no writes.

Benchmarks. Not totally apple-to-apple due to the `switch` overhead in `After` from what I can tell - I think the overhead is shown by `num=3`, where the implementation hasn't really changed, so presumably `After` should be treated a little higher than they look. But still pretty minor, let me know what you think.

```
After

Benchmark                            (flowControl)  (num)          (streamType)   Mode  Cnt         Score        Error  Units
StreamMessageBenchmark.jmhEventLoop          false      0  FIXED_STREAM_MESSAGE  thrpt   20  11644168.631 ± 152677.634  ops/s
StreamMessageBenchmark.jmhEventLoop          false      1  FIXED_STREAM_MESSAGE  thrpt   20   8573895.575 ±  34659.073  ops/s
StreamMessageBenchmark.jmhEventLoop          false      2  FIXED_STREAM_MESSAGE  thrpt   20   6602223.767 ±  24179.698  ops/s
StreamMessageBenchmark.jmhEventLoop          false      3  FIXED_STREAM_MESSAGE  thrpt   20   5326205.531 ±  17208.362  ops/s

Before

Benchmark                            (flowControl)  (num)          (streamType)   Mode  Cnt         Score       Error  Units
StreamMessageBenchmark.jmhEventLoop          false      0  FIXED_STREAM_MESSAGE  thrpt   20  11338855.960 ± 56332.436  ops/s
StreamMessageBenchmark.jmhEventLoop          false      1  FIXED_STREAM_MESSAGE  thrpt   20   8165172.831 ± 22656.208  ops/s
StreamMessageBenchmark.jmhEventLoop          false      2  FIXED_STREAM_MESSAGE  thrpt   20   6681994.146 ± 32692.861  ops/s
StreamMessageBenchmark.jmhEventLoop          false      3  FIXED_STREAM_MESSAGE  thrpt   20   5441368.221 ± 20874.088  ops/s

```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
@@ -429,74 +429,20 @@ public interface AggregatedHttpMessage {
     /**
      * Converts this message into a new complete {@link HttpRequest}.
      *
-     * @return the converted {@link HttpRequest} whose stream is complete
-     * @throws IllegalStateException if this message is not a request
+     * @deprecated use {@link HttpRequest#of(AggregatedHttpMessage)}.
      */
+    @Deprecated
     default HttpRequest toHttpRequest() {
-        final HttpHeaders headers = headers();
-
-        // From the section 8.1.2.3 of RFC 7540:
-        //// All HTTP/2 requests MUST include exactly one valid value for the :method, :scheme, and :path
-        //// pseudo-header fields, unless it is a CONNECT request (Section 8.3)
-        // NB: ':scheme' will be filled when a request is written.
-        if (headers.method() == null) {
-            throw new IllegalStateException("not a request (missing :method)");
-        }
-        if (headers.path() == null) {
-            throw new IllegalStateException("not a request (missing :path)");
-        }
-
-        final HttpData content = content();
-        final HttpHeaders trailingHeaders = trailingHeaders();
-        final int numObjects = (!content.isEmpty() ? 1 : 0) +
-                               (!trailingHeaders.isEmpty() ? 1 : 0);
-        final HttpObject[] objs = new HttpObject[numObjects];
-        int writerIndex = 0;
-        if (!content.isEmpty()) {
-            objs[writerIndex++] = content;
-        }
-        if (!trailingHeaders.isEmpty()) {
-            objs[writerIndex] = trailingHeaders;
-        }
-        return FixedHttpRequest.ofWrittenHeaders(headers, objs);
+        return HttpRequest.of(this);
     }
 
     /**
      * Converts this message into a new complete {@link HttpResponse}.
      *
-     * @return the converted {@link HttpResponse} whose stream is complete
-     * @throws IllegalStateException if this message is not a response.
+     * @deprecated use {@link HttpResponse#of(AggregatedHttpMessage)}.
      */
+    @Deprecated
     default HttpResponse toHttpResponse() {
-        final List<HttpHeaders> informationals = informationals();
-        final HttpHeaders headers = headers();
-
-        // From the section 8.1.2.4 of RFC 7540:
-        //// For HTTP/2 responses, a single :status pseudo-header field is defined that carries the HTTP status
-        //// code field (see [RFC7231], Section 6). This pseudo-header field MUST be included in all responses;
-        //// otherwise, the response is malformed (Section 8.1.2.6).
-        if (headers.status() == null) {
-            throw new IllegalStateException("not a response (missing :status)");
-        }
-
-        final HttpData content = content();
-        final HttpHeaders trailingHeaders = trailingHeaders();
-        final int numObjects = informationals.size() +
-                               1 /* headers */ +
-                               (!content.isEmpty() ? 1 : 0) +
-                               (!trailingHeaders.isEmpty() ? 1 : 0);
-        final HttpObject[] objs = new HttpObject[numObjects];
-        int writerIndex = 0;
-        for (HttpHeaders informational : informationals()) {
-            objs[writerIndex++] = informational;
-        }
-        objs[writerIndex++] = headers;
-        if (!content.isEmpty()) {
-            objs[writerIndex++] = content;
-        }
-        if (!trailingHeaders.isEmpty()) {
-            objs[writerIndex] = trailingHeaders;
-        }
-        return FixedHttpResponse.of(objs);
+        return HttpResponse.of(this);
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
@@ -458,7 +458,7 @@ public interface AggregatedHttpMessage {
         if (!trailingHeaders.isEmpty()) {
             objs[writerIndex] = trailingHeaders;
         }
-        return FixedHttpRequest.ofHeadersWritten(headers, objs);
+        return FixedHttpRequest.ofWrittenHeaders(headers, objs);
     }
 
     /**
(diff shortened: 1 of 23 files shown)


core/src/main/java/com/linecorp/armeria/common/FixedHttpRequest.java
```diff
@@ -19,64 +19,191 @@
...
+                return of(keepalive, headers, objs[0]);
+            default:
+                return new RegularFixedHttpRequest(headers, objs, keepalive);
+        }
     }
 
     /**
-     * Creates a new {@link FixedHttpRequest} that will publish the given {@code objs}. {@code objs} is not
+     * Creates a new {@link HttpRequest} that will publish the given {@code objs}. {@code objs} is not
      * copied so must not be mutated after this method call (it is generally meant to be used with a varargs
```
trustin: I guess this may be premature optimization. Given this is not true for the case there `objs.length` <= 2, could we just make `RefularFixedHttpRequest` make a copy of `objs`?

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -33,277 +35,467 @@
...
+    public static <T> StreamMessage<T> of(T obj1, T obj2) {
+        requireNonNull(obj1, "obj1");
+        requireNonNull(obj2, "obj2");
+        return new TwoElementFixedStreamMessage<>(obj1, obj2);
+    }
 
     /**
-     * Creates a new {@link FixedStreamMessage} that will publish the given {@code objs}. {@code objs} is not
+     * Creates a new {@link StreamMessage} that will publish the given {@code objs}. {@code objs} is not
      * copied so must not be mutated after this method call (it is generally meant to be used with a varargs
```
trustin: Ditto

anuraaga:
Thanks a lot for the suggestion! Indeed, there was significant duplication between those factories and the one in `HttpRequest`, and there's no good point for it. I made `FixedHttp*` package-private, and also fixed my misunderstanding that it's possible to have a `HttpRequest` that publishes headers (and made an explicit check to prevent this).

I also went ahead and deprecated `AggregatedHttpMessage.to*` in the interest of making `HttpRequest` and `HttpResponse` the only entrypoints for creating them (will make the API clearer after #850 ). In the process, I moved some error-checking logic from there to `HttpRequest/Response`, presumably it was supposed to be there either way.

Also, what do you think about moving `FixedStreamMessage` to `armeria.internal.stream`? I would only do it if we're also ok with moving `Default/EventLoopStreamMessage` there to later.

trustin:
> Also, what do you think about moving `FixedStreamMessage` to `armeria.internal.stream`? I would only do it if we're also ok with moving `Default/EventLoopStreamMessage` to there later.

What's the benefit of moving them there? I usually prefer keeping classes package-local if possible since the classes in the internal package are visible to users anyway.

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
```diff
@@ -429,74 +429,20 @@ default HttpStatus status() {
     /**
      * Converts this message into a new complete {@link HttpRequest}.
      *
-     * @return the converted {@link HttpRequest} whose stream is complete
-     * @throws IllegalStateException if this message is not a request
+     * @deprecated use {@link HttpRequest#of(AggregatedHttpMessage)}.
```
trustin: use -> Use

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
```diff
@@ -429,74 +429,20 @@ default HttpStatus status() {
...
-        return FixedHttpRequest.ofHeadersWritten(headers, objs);
+        return HttpRequest.of(this);
     }
 
     /**
      * Converts this message into a new complete {@link HttpResponse}.
      *
-     * @return the converted {@link HttpResponse} whose stream is complete
-     * @throws IllegalStateException if this message is not a response.
+     * @deprecated use {@link HttpResponse#of(AggregatedHttpMessage)}.
```
trustin: use -> Use

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -33,277 +36,465 @@
...
+    public static class RegularFixedStreamMessage<T> extends AbstractFixedStreamMessage<T> {
+
+        private final T[] objs;
+
+        private int fulfilled;
+
+        private boolean inOnNext;
+
+        protected RegularFixedStreamMessage(T[] objs) {
+            this.objs = Arrays.copyOf(objs, objs.length);
```
trustin: `objs.clone()`?

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -33,277 +36,465 @@
...
-        return new FixedStreamMessage<>(objs);
+        switch (objs.length) {
+            case 0:
+                return of();
+            case 1:
+                return of(objs[0]);
+            case 2:
+                return of(objs[0], objs[1]);
+            default:
+                return new RegularFixedStreamMessage<>(objs);
```
trustin: Check all elements are non-null?

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -33,277 +36,465 @@
  * Reduced synchronization and allocation allow for much higher performance than {@link DefaultStreamMessage},
  * so this class should generally be used when the objects are known.
  */
-public class FixedStreamMessage<T> extends AbstractStreamMessage<T> {
+public final class FixedStreamMessage {
```
trustin: How about:

- moving all the factory methods to `StreamMessage`
- removing `FixedStreamMessage` and moving all its nested classes top-level
- renaming `AbstractFixedStreamMessage` to `FixedStreamMessage`?

anuraaga: Thanks - reorganized. Unfortunately I can't make the `FixedStreamMessage` implementations package-private since we inherit from outside the package. It's why I was thinking of moving them to `internal` but I guess this way is ok for now.

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -16,24 +16,21 @@
...
 import com.linecorp.armeria.common.Flags;
 
-import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.ImmediateEventExecutor;
 
 /**
- * A {@link StreamMessage} used when all the objects that will be published are known at construction time.
- * Reduced synchronization and allocation allow for much higher performance than {@link DefaultStreamMessage},
- * so this class should generally be used when the objects are known.
+ * A {@link AbstractStreamMessage} which only publishes a fixed number of objects known at construction time.
```
minwoox: nit: A {@link Fixed~~ } or An {@link Abstract~~~}

anuraaga: Thanks

trustin:
Thanks, @anuraaga !

PR #891, opened 2017-12-05
[WIP] Change to record the first request and the last response when …

### LINKED ISSUE #869: RetryingHttpClient has to create a new context on each retry
`RetryingHttpClient` currently reuses the initial `ClientRequestContext` for retry requests. This has the following issues:

- `RequestLog` of the retries are not recorded.
- The `RequestLog` of the original request is recorded as 'failed' prematurely. We could record it as 'succeeded' if a retry succeeded.
- If there is a stateful `DecoratingClient`, which keeps its state in the context attribute map, between `RetryingHttpClient` and `HttpClientDelegate`, the `DecoratingClient` may not function correctly because the state of the retries may confuse it.

/cc @minwoox

minwoox:
…retry

Motivation:
`RetryingHttpClient` currently records only the first request and response.
If the reqeust is succeeded after couple of retrying, it is recored as failed because the first reponse is failed.

Modifications:
- Use `RetryingRequestLog` when retrying
- Reset the `ClientRequestContext` when retrying

Result:
- Close #869

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
@@ -53,6 +53,12 @@ import io.netty.channel.Channel;
 public interface RequestLog {
 
     /**
+     * Returns the unified integer value of flags that represents {@link RequestLogAvailability}s.
+     * You can restore the {@link RequestLogAvailability}s using {@link RequestLogAvailabilitySet#of(int)}.
+     */
+    int flags();
+
+    /**
      * Returns the set of satisfied {@link RequestLogAvailability}s.
      */
     Set<RequestLogAvailability> availabilities();
--- a/core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAvailability.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAvailability.java
@@ -16,6 +16,10 @@
 
 package com.linecorp.armeria.common.logging;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
+import java.util.stream.Stream;
+
 /**
  * Tells which properties are available in a {@link RequestLog}.
  */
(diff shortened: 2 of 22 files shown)


core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
```diff
@@ -52,6 +52,11 @@
  */
 public interface RequestLog {
 
+    /**
+     * Returns the unified integer value of flags that represents {@link RequestLogAvailability}s.
+     */
+    int flags();
```
trustin: This should not be exposed in the public API since it's implementation detail.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAvailability.java
```diff
@@ -94,7 +98,32 @@ int getterFlags() {
         return getterFlags;
     }
 
+    /**
+     * Returns the unified integer value of getter flags from the specified {@code availabilities}.
+     */
+    public static int getterFlags(RequestLogAvailability... availabilities) {
```
trustin: This should not be exposed in the public API since it's implementation detail.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAvailability.java
```diff
@@ -94,7 +98,32 @@ int getterFlags() {
...
+        return getterFlags(Stream.of(availabilities).collect(toImmutableList()));
+    }
+
+    /**
+     * Returns the unified integer value of getter flags from the specified {@code availabilities}.
+     */
+    public static int getterFlags(Iterable<RequestLogAvailability> availabilities) {
```
trustin: This should not be exposed in the public API since it's implementation detail.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAvailability.java
```diff
@@ -94,7 +98,32 @@ int getterFlags() {
...
+    }
+
     int setterFlags() {
         return setterFlags;
     }
+
+    /**
+     * Returns whether the specified {@code flags} contains only the availabilities related with the request.
+     */
+    public static boolean isRequestAvailabilityOnly(int flags) {
```
trustin: This should not be exposed in the public API since it's implementation detail.

minwoox:
Thanks @trustin for the suggestion. I think I need to change a lot so I just close this PR and send the new PR again.

Issue #897, opened 2017-12-06
Performance problem when use armeria as proxy server


huydx:
# Problem
I used armeria to create very simple proxy server and notice some performance problem when upstream is very slow response server. The proxy server itself is just few lines of code:

```java
    public static void main(String[] args) {
        int workerNumber = 16;
        ClientFactory clientFactory = new ClientFactoryBuilder()
                .workerGroup(EventLoopGroups.newEventLoopGroup(workerNumber), true)
                .build();
        ClientBuilder clientBuilder = new ClientBuilder("none+http://127.0.0.1:9998").factory(clientFactory);
        final HttpClient httpClient = clientBuilder.build(HttpClient.class);

        ServerBuilder sb =  new ServerBuilder().port(20080, SessionProtocol.HTTP).defaultRequestTimeoutMillis(0);
        sb.service("/", (HttpService) (ctx, req) ->
                HttpResponse.from(req.aggregate()
                                     .thenComposeAsync(message -> {
                                         try (SafeCloseable ignored = Clients.withContextCustomizer(clientCtx -> {})) {
                                             return httpClient
                                                     .execute(message.toHttpRequest())
                                                     .aggregate()
                                                     .thenApply(AggregatedHttpMessage::toHttpResponse);
                                         }
                                     })));
        sb.build().start().join();
    }
```

Assume that upstream server has very low response, which I simulate using below golang code:

```golang
package main

import (
	"fmt"
	"net/http"
	"time"
)

func handler(w http.ResponseWriter, r *http.Request) {
	time.Sleep(time.Duration(2) * time.Second)
	fmt.Fprintf(w, "ok")
}

func main() {
	http.HandleFunc("/", handler)
	http.ListenAndServe(":9998", nil)
}
```

# Some numbers using `ab` tool
 ab -n 1000 -c 100 http://localhost:20080/

Raw result without using proxy

```
Percentage of the requests served within a certain time (ms)
  50%   2016
  66%   2017
  75%   2018
  80%   2019
  90%   2023
  95%   2027
  98%   2031
  99%   2031
 100%   2032 (longest request)
```

### Using 16 thread (*2 core)

```
Percentage of the requests served within a certain time (ms)
  50%   8272
  66%  12015
  75%  12017
  80%  12019
  90%  12022
  95%  14022
  98%  14026
  99%  14028
 100%  14029 (longest request)
```

### Using 100 threads (== number of concurrent request)
Had better result, but 80% still have 2 times compare to upstream

```
[...]

PR #899, opened 2017-12-07
Change AbstractStreamMessageDuplicator to have an EventExecutor

minwoox:
Motivation:
`AbstractStreamMessageDuplicator` was desgined initially to be used in the situation where
a lot of threads are competing for the resourses.
However, the `Duplicator` is created and used during retrying mostly, which is conducted by
an event loop.
So, I thought it would be better if we remove the volatile fields as possible and use the event loop to
syncronize the access using `doSomeMethod` as we do in `EventLoopStreamMessage`.

Modifications:
- `StreamMessageProcessor` in `AbstractStreamMessageDuplicator` has an `EventExecutor`
- Use the `EventExecutor` where syncronization is needed

Result:
- Better dupplicator less bugs


huydx:
this will fix NPE which I found at https://github.com/line/armeria/issues/897 right?

trustin:
Thanks, @minwoox !

minwoox:
> this will fix NPE which I found at #897 right?

You are right~! Thank you @huydx

PR #900, opened 2017-12-08
Add addChild method to RequestLogBuilder

### LINKED ISSUE #869: RetryingHttpClient has to create a new context on each retry

minwoox:
Motivation:
`RetryingClient` could not record the logs properly.

Modifications:
- Add `addChild` to `RequestLogBuilder` to propagate the logs from child

Result:
- Close #869

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
@@ -82,9 +82,17 @@ public abstract class RetryingClient<I extends Request, O extends Response>
      */
     protected final O executeDelegate(ClientRequestContext ctx, I req) throws Exception {
         final ClientRequestContext derivedContext = ctx.newDerivedContext();
+        ctx.logBuilder().addChild(derivedContext.log());
         try (SafeCloseable ignore = RequestContext.push(derivedContext, false)) {
             return delegate().execute(derivedContext, req);
         }
+    }
+
+    /**
+     * This should be called when retrying is finished.
+     */
+    protected static void onRetryingComplete(ClientRequestContext ctx) {
+        ctx.logBuilder().endResponseWithLastChild();
     }
 
     protected RetryStrategy<I, O> retryStrategy() {
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
@@ -120,16 +120,15 @@ public final class RetryingHttpClient extends RetryingClient<HttpRequest, HttpRe
     private void doExecute0(ClientRequestContext ctx, HttpRequestDuplicator rootReqDuplicator,
                             DeferredHttpResponse deferredRes) {
         if (!setResponseTimeout(ctx)) {
-            closeOnException(deferredRes, rootReqDuplicator, ResponseTimeoutException.get());
+            closeOnException(ctx, deferredRes, rootReqDuplicator, ResponseTimeoutException.get());
             return;
         }
 
-        final HttpResponse response;
+        HttpResponse response;
         try {
             response = executeDelegate(ctx, rootReqDuplicator.duplicateStream());
         } catch (Exception e) {
-            closeOnException(deferredRes, rootReqDuplicator, e);
-            return;
+            response = HttpResponse.ofFailure(e);
         }
 
         final HttpResponseDuplicator resDuplicator =
@@ -148,7 +147,7 @@ public final class RetryingHttpClient extends RetryingClient<HttpRequest, HttpRe
                                try {
                                    nextDelay = getNextDelay(ctx, backoffOpt.get(), millisAfter);
                                } catch (Exception e) {
-                                   closeOnException(deferredRes, rootReqDuplicator, e);
+                                   closeOnException(ctx, deferredRes, rootReqDuplicator, e);
                                    return;
                                }
 
@@ -161,6 +160,7 @@ public final class RetryingHttpClient extends RetryingClient<HttpRequest, HttpRe
                                            nextDelay, TimeUnit.MILLISECONDS);
                                }
                            } else {
+                               onRetryingComplete(ctx);
                                deferredRes.delegate(resDuplicator.duplicateStream(true));
                                rootReqDuplicator.close();
                            }
@@ -205,8 +205,10 @@ public final class RetryingHttpClient extends RetryingClient<HttpRequest, HttpRe
         return future.join();
     }
 
-    private static void closeOnException(DeferredHttpResponse deferredRes,
+    private static void closeOnException(ClientRequestContext ctx,
+                                         DeferredHttpResponse deferredRes,
                                          HttpRequestDuplicator rootReqDuplicator, Throwable cause) {
+        onRetryingComplete(ctx);
         deferredRes.close(cause);
         rootReqDuplicator.close();
     }
--- a/core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
@@ -127,16 +127,7 @@ public class DefaultRequestLog implements RequestLog, RequestLogBuilder {
 
     private void propagateRequestSideLog(RequestLog child) {
         child.addListener(log -> {
-            final long requestStartTimeNanos;
-            final long requestStartTimeMillis;
-            if (log instanceof DefaultRequestLog) {
-                requestStartTimeNanos = ((DefaultRequestLog) log).requestStartTimeNanos;
-                requestStartTimeMillis = log.requestStartTimeMillis();
-            } else {
-                requestStartTimeNanos = System.nanoTime();
-                requestStartTimeMillis = System.currentTimeMillis();
-            }
-            startRequest0(requestStartTimeNanos, requestStartTimeMillis, log.channel(),
+            startRequest0(log.requestStartTimeNanos(), log.requestStartTimeMillis(), log.channel(),
                           log.sessionProtocol(), log.host(), true);
         }, REQUEST_START);
         child.addListener(log -> serializationFormat(log.serializationFormat()), SCHEME);
@@ -144,17 +135,10 @@ public class DefaultRequestLog implements RequestLog, RequestLogBuilder {
         child.addListener(log -> requestContent(log.requestContent(), log.rawRequestContent()),
                           REQUEST_CONTENT);
         child.addListener(log -> {
-            final long requestEndTimeNanos;
-            if (log instanceof DefaultRequestLog) {
-                requestEndTimeNanos = ((DefaultRequestLog) log).requestEndTimeNanos;
-            } else {
-                requestEndTimeNanos = System.nanoTime();
-            }
-
             if (log.requestCause() != null) {
-                endRequest0(requestEndTimeNanos, log.requestCause());
+                endRequest0(log.requestEndTimeNanos(), log.requestCause());
             } else {
-                endRequest0(requestEndTimeNanos, null);
+                endRequest0(log.requestEndTimeNanos(), null);
             }
         }, REQUEST_END);
     }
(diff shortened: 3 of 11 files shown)


core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -109,6 +113,67 @@ public DefaultRequestLog(RequestContext ctx) {
...
+            }
+            if (log.responseContent() != null) {
+                responseContent(log.responseContent(), log.rawResponseContent());
+            }
+            if (log.responseCause() != null) {
+                endResponse(log.responseCause());
+            } else {
+                endResponse();
+            }
+        }, COMPLETE);
```
trustin: Doesn't need to use `COMPLETE` but just `RESPONSE_END`.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -612,13 +682,28 @@ private void notifyListeners(RequestLogListener[] listeners) {
...
+           .append(res)
+           .append('}');      // 1 char
+        if (children != null && children.size() > 1) {
+            buf.append("{totalAttempts=");
+            buf.append(children.size());
+            buf.append("}[");
+            for (int i = 0; i < children.size(); i++) {
+                buf.append(children.get(i));
+                if (i != children.size() - 1) {
+                    buf.append(',');
```
trustin: `", "`?

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogBuilder.java
```diff
@@ -36,6 +36,23 @@
...
+     * @see RequestLogBuilder#setRecentAsLastChild
+     */
+    void addChild(RequestLog child, boolean lastChild);
+
+    /**
+     * Sets the recent child as the last child log. The response-side logs of last child will be propagated.
+     * Note that the response-side logs are propagated when the {@link RequestLogAvailability} fulfills
+     * {@link RequestLogAvailability#COMPLETE}.
+     */
+    void setRecentAsLastChild();
```
trustin: `endResponseWithLastChild`?

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogBuilder.java
```diff
@@ -36,6 +36,23 @@
...
+     *                  request-side logs will be propagated.
+     *
+     * @see RequestLogBuilder#setRecentAsLastChild
+     */
+    void addChild(RequestLog child, boolean lastChild);
+
+    /**
+     * Sets the recent child as the last child log. The response-side logs of last child will be propagated.
+     * Note that the response-side logs are propagated when the {@link RequestLogAvailability} fulfills
+     * {@link RequestLogAvailability#COMPLETE}.
```
trustin: `#RESPONSE_END`?

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithLoggingTest.java
```diff
@@ -0,0 +1,156 @@
...
+                ctx.log().addListener(listener, RequestLogAvailability.REQUEST_END);
+                ctx.log().addListener(listener, RequestLogAvailability.COMPLETE);
+                return delegate().execute(ctx, req);
+            }
+        };
+    }
+
+    private static void assertRequestSideLog(RequestLog log) {
+        assertThat(log.requestHeaders()).isNotNull();
+        assertThat(log.requestHeaders().path()).isEqualToIgnoringCase("/hello");
```
trustin: `isEqualTo()`?

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithLoggingTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        assertThat(log.toStringResponseOnly()).isEqualToIgnoringCase("{}"); // empty response
```
trustin: `isEqualTo()`?

core/src/test/java/com/linecorp/armeria/common/logging/DefaultRequestLogTest.java
```diff
@@ -73,4 +86,60 @@ public void rpcFailure_endResponseDifferentCause() {
...
+        child.serializationFormat(SerializationFormat.NONE);
+        assertThat(log.serializationFormat()).isSameAs(SerializationFormat.NONE);
+
+        final HttpHeaders foo = HttpHeaders.of(AsciiString.of("foo"), "foo");
+        child.requestHeaders(foo);
+        assertThat(log.requestHeaders()).isSameAs(foo);
+
+        child.requestContent("baz", "bax");
+        assertThat(log.requestContent()).isSameAs("baz");
+        assertThat(log.rawRequestContent()).isSameAs("bax");
```
trustin: Could you extract the constants into variables so that it's less confusing?

core/src/test/java/com/linecorp/armeria/common/logging/DefaultRequestLogTest.java
```diff
@@ -73,4 +86,60 @@ public void rpcFailure_endResponseDifferentCause() {
...
+                .isExactlyInstanceOf(RequestLogAvailabilityException.class);
+        assertThatThrownBy(() -> log.rawResponseContent())
+                .isExactlyInstanceOf(RequestLogAvailabilityException.class);
+
+        child.endResponse(new AnticipatedException("Oops!"));
+        // all response-side logs are propagated
+        assertThat(log.responseStartTimeMillis()).isEqualTo(child.responseStartTimeMillis());
+        assertThat(log.responseHeaders()).isSameAs(bar);
+        assertThat(log.responseContent()).isSameAs("baz1");
+        assertThat(log.rawResponseContent()).isSameAs("bax1");
```
trustin: Could you extract the constants into variables so that it's less confusing?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
minwoox: All of the codes are fixed. Thank you~!

core/src/test/java/com/linecorp/armeria/common/logging/DefaultRequestLogTest.java
```diff
@@ -73,4 +86,58 @@ public void rpcFailure_endResponseDifferentCause() {
...
+        final String responseContent = "baz1";
+        final String rawResponseContent = "bax1";
+        child.responseContent(responseContent, rawResponseContent);
+        assertThat(log.responseContent()).isSameAs(responseContent);
+        assertThat(log.rawResponseContent()).isSameAs(rawResponseContent);
+
+        child.endResponse(new AnticipatedException("Oops!"));
+        final long childTotalDurationMillis = TimeUnit.NANOSECONDS.toMillis(child.totalDurationNanos());
+        assertThat(TimeUnit.NANOSECONDS.toMillis(log.totalDurationNanos()))
+                .isBetween(Math.max(0, childTotalDurationMillis - 300), childTotalDurationMillis); // 300 buffer
```
trustin: - 300 buffer -> Could you rephrase?
- Could you extract 300 into a variable? It's used twice.
- How about increasing to 1000 for less flakiness?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -109,6 +113,95 @@ public DefaultRequestLog(RequestContext ctx) {
...
+        if (children == null) {
+            // first child's all request-side logging events are propagated immediately to the parent
+            children = new ArrayList<>();
+            propagateRequestSideLog(child);
+        }
+        children.add(child);
+    }
+
+    private void propagateRequestSideLog(RequestLog child) {
+        child.addListener(log -> startRequest(log.channel(), log.sessionProtocol(), log.host()), REQUEST_START);
```
trustin: Could we set the start time explicitly so that the start time of the parent request log is same with that of the first child? You could add a private variant of `startRequest()` which accepts the start time as a parameter.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -109,6 +113,95 @@ public DefaultRequestLog(RequestContext ctx) {
...
+        child.addListener(log -> serializationFormat(log.serializationFormat()), SCHEME);
+        child.addListener(log -> requestHeaders(log.requestHeaders()), REQUEST_HEADERS);
+        child.addListener(log -> requestContent(log.requestContent(), log.rawRequestContent()),
+                          REQUEST_CONTENT);
+        child.addListener(log -> {
+            if (log.requestCause() != null) {
+                endRequest(log.requestCause());
+            } else {
+                endRequest();
```
trustin: Ditto - Could we set the end time explicitly so that the end time of the parent request log is same with that of the first child? You could add a private variant of `endRequest()` which accepts the start time as a parameter.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -109,6 +113,95 @@ public DefaultRequestLog(RequestContext ctx) {
...
+            endResponseFrom(lastChild);
+        }
+    }
+
+    private void startResponseFrom(RequestLog log) {
+        if (log instanceof DefaultRequestLog) {
+            startResponse0(true, ((DefaultRequestLog) log).responseStartTimeNanos,
+                           log.responseStartTimeMillis());
+        } else {
+            startResponse();
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -109,6 +113,95 @@ public DefaultRequestLog(RequestContext ctx) {
...
+        if (lastChild.responseContent() != null) {
+            responseContent(lastChild.responseContent(), lastChild.rawResponseContent());
+        }
+    }
+
+    private void endResponseFrom(RequestLog log) {
+        if (log.responseCause() != null) {
+            endResponse(log.responseCause());
+        } else {
+            endResponse();
```

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -82,11 +82,19 @@ public O execute(ClientRequestContext ctx, I req) throws Exception {
...
+        ctx.logBuilder().addChild(derivedContext.log());
         try (SafeCloseable ignore = RequestContext.push(derivedContext, false)) {
             return delegate().execute(derivedContext, req);
         }
     }
 
+    /**
+     * This should be called when retrying is finished.
+     */
+    protected static void onRetryingComplete(ClientRequestContext ctx) {
```
hyangtack: Is there any reason making this method `static` unlike the other methods such as `setResponseTimeout`, `getNextDelay`, etc?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -120,16 +120,15 @@ protected HttpResponse doExecute(ClientRequestContext ctx, HttpRequest req) thro
...
         }
 
-        final HttpResponse response;
+        HttpResponse response;
         try {
             response = executeDelegate(ctx, rootReqDuplicator.duplicateStream());
         } catch (Exception e) {
-            closeOnException(deferredRes, rootReqDuplicator, e);
-            return;
+            response = HttpResponse.ofFailure(e);
```
hyangtack: Just curious. Does it aim to apply the retry strategy when catching an exception from the delegate?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -109,6 +113,121 @@ public DefaultRequestLog(RequestContext ctx) {
...
+            propagateRequestSideLog(child);
+        }
+        children.add(child);
+    }
+
+    private void propagateRequestSideLog(RequestLog child) {
+        child.addListener(log -> {
+            final long requestStartTimeNanos;
+            final long requestStartTimeMillis;
+            if (log instanceof DefaultRequestLog) {
```
hyangtack: nit: How about adding `requestStartTimeNanos()` and `requestEndTimeNanos()` to `RequestLog` interface? @trustin WDYT?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -109,6 +113,121 @@ public DefaultRequestLog(RequestContext ctx) {
...
+        checkState(!hasLastChild, "last child is already added");
+        checkState(!children.isEmpty(), "at least one child should be already added");
+        hasLastChild = true;
+        final RequestLog lastChild = children.get(children.size() - 1);
+        propagateResponseSideLog(lastChild);
+    }
+
+    private void propagateResponseSideLog(RequestLog lastChild) {
+        // update the available logs if the lastChild already has them
+        if (lastChild.isAvailable(RequestLogAvailability.RESPONSE_START)) {
```
hyangtack: RequestLogAvailability.RESPONSE_START -> RESPONSE_START? Because we have static imports already.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
minwoox: You are right. That is the reason.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
minwoox: Oops~ fixed

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
minwoox: That was my first thought. But I was not sure if it's ok to expose the value to the outside since it is relative value.
I'm not sticking to that though :-).

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -612,13 +738,28 @@ private void notifyListeners(RequestLogListener[] listeners) {
...
-                  .append(req)
-                  .append(", res=") // 6 chars
-                  .append(res)
-                  .append('}')      // 1 char
-                  .toString();
+        buf.append("{req=")  // 5 chars
+           .append(req)
+           .append(", res=") // 6 chars
+           .append(res)
+           .append('}');      // 1 char
```
trustin: `//` is misaligned?

trustin:
Thanks, @minwoox !

PR #902, opened 2017-12-11
Disables http1 pipelining by default

imasahiro:
Most http clients including modern http browser does not enable
http1 pipelining by default because many http servers are still
do not works well with it. For this reason, disabling http1 pipelining
by default.

Related to #897

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -132,7 +132,7 @@ public final class Flags {
                    value -> value >= 0);
 
     private static final boolean DEFAULT_USE_HTTP2_PREFACE = getBoolean("defaultUseHttp2Preface", false);
-    private static final boolean DEFAULT_USE_HTTP1_PIPELINING = getBoolean("defaultUseHttp1Pipelining", true);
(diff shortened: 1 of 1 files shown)


imasahiro:
You can see similer topic from https://developer.mozilla.org/en-US/docs/Web/HTTP/Connection_management_in_HTTP_1.x#HTTP_Pipelining

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -373,7 +373,7 @@ public static boolean defaultUseHttp2Preface() {
      * Returns the default value of the {@link ClientFactoryBuilder#useHttp1Pipelining(boolean)} option.
      * Note that this value has effect only if a user did not specify it.
      *
-     * <p>This flag is enabled by default. Specify the
+     * <p>This flag is disabled by default. Specify the
      * {@code -Dcom.linecorp.armeria.defaultUseHttp1Pipelining=false} JVM option to disable it.
```
hyangtack: -Dcom.linecorp.armeria.defaultUseHttp1Pipelining=true ... enable it?

huydx:
LGTM except hyangtack comment

trustin:
Thanks, @imasahiro !

PR #903, opened 2017-12-11
Have DeferredStreamMessage manage the stream instead of just forwarding.

anuraaga:
Currently request timeout doesn't work correctly with deferred responses - the timeout is set after `onSubscribe` is called, which only happens after `delegate` at some point possibly well in the future. And in general, stream lifecycle events should happen before `delegate` has been called. This means it's not enough for a `DefferedStreamMessage` to simply forward all stream events to the delegate - it needs to handle them and only start forwarding after `delegate` has been called.

This change adds keeping track of demand and cancellation to `DeferredStreamMessage` - even before `delegate` has been called, `onSubscribe` is called on the subscriber to make sure logic that should run as soon as subscription (e.g., setting a request timeout) can happen, and since this will immediately trigger demand, the demand is kept safe until delegation starts. If the stream is aborted, the subscriber is notified immediately as it should. If the Subscriber cancels the stream, the delegate will also be cancelled after delegation happens.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
@@ -18,15 +18,17 @@ package com.linecorp.armeria.common.stream;
 
 import static java.util.Objects.requireNonNull;
 
-import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
 
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.util.CompletionActions;
 
 import io.netty.util.concurrent.EventExecutor;
+import io.netty.util.concurrent.ImmediateEventExecutor;
 
 /**
  * A {@link StreamMessage} whose stream is published later by another {@link StreamMessage}. It is useful when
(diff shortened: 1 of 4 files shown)


anuraaga:
Found another failing case for deferred, will poke at it more.

trustin:
Basically LGTM. Please remove WIP once you are done with it.

minwoox:
Sounds good too.
Do we need to make the same change in the client-side which is in `HttpRequestSubscriber` as well?

anuraaga:
@trustin After digging in, I realized it's not possible to handle all cases with pure forwarding of events to the delegate, so I have added some state tracking to `DeferredStreamMessage`. This allows the stream's lifecycle to run independently of when `delegate` is called. This is probably required for correct behavior.

@minwoox I ended up changing the implementation to something more general, so request should be covered too now.

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -134,125 +152,129 @@ public boolean isEmpty() {
...
     }
 
     @Override
-    public void subscribe(Subscriber<? super T> subscriber) {
-        requireNonNull(subscriber, "subscriber");
-        subscribe0(subscriber, null, false);
+    void request(long n) {
+        Subscription subscription = delegateSubscription;
+        if (subscription != null) {
+            subscription.request(n);
```
hyangtack: nit: Could we add before here `assert pendingDemand == 0;` just to confirm that there is no pending demand?

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -18,45 +18,63 @@
...
+    // Only accessed from subscription's executor.
+    private long pendingDemand;
+
     private volatile boolean abortPending;
 
+    // Only accessed from subscription's executor.
+    private boolean cancelPending;
+
     /**
      * Sets the delegate {@link HttpResponse} which will publish the stream actually.
```
minwoox: This is not relevant but we need to change the `HttpResponse` here :-)

anuraaga:
Thanks - updated

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
anuraaga: BTW, I ended up having to make the ForwardingSubscriber's onSubscribe slightly more complex to satisfy the assertion. Not a big deal, but before the pendingDemand usage was still guaranteed to be correct with a bit less complexity. Let me know if you think it's better this way.

hyangtack: Oh.. I think removing the assertion is okay for less complexity.

anuraaga: Thanks :)

trustin:
Thanks, @anuraaga !

PR #904, opened 2017-12-12
Added contentType() to HttpHeaders

jmostella:
When making HttpHeaders for testing it would be more convenient to
have a method for content-type

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
+++ b/core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
@@ -38,6 +38,7 @@ public final class DefaultHttpHeaders
 
     private HttpMethod method;
     private HttpStatus status;
+    private MediaType contentType;
 
     /**
      * Creates a new instance.
@@ -172,6 +173,33 @@ public final class DefaultHttpHeaders
     public HttpHeaders status(HttpStatus status) {
         requireNonNull(status, "status");
         return status(status.code());
+    }
+
+    @Override
+    public MediaType contentType() {
+        final MediaType contentType = this.contentType;
+        if (contentType != null) {
+            return contentType;
+        }
+
+        final String contentTypeString = get(HttpHeaderNames.CONTENT_TYPE);
+        if (contentTypeString == null) {
+            return null;
+        }
+
+        try {
+            this.contentType = MediaType.parse(contentTypeString);
+            return this.contentType;
+        } catch (IllegalArgumentException unused) {
+            return null;
+        }
+    }
+
+    @Override
+    public HttpHeaders contentType(MediaType contentType) {
+        requireNonNull(contentType, "contentType");
+        this.contentType = contentType;
+        return set(HttpHeaderNames.CONTENT_TYPE, contentType.toString());
     }
 
     @Override
--- a/core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
@@ -175,6 +175,18 @@ public interface HttpHeaders extends HttpObject, Headers<AsciiString, String, Ht
     HttpHeaders status(HttpStatus status);
 
     /**
+     * Returns the value of the {@code 'content-type'} header.
+     * @return the valid header value if present. {@code null} otherwise.
+     */
+    @Nullable
+    MediaType contentType();
+
+    /**
+     * Sets the {@link HttpHeaderNames#CONTENT_TYPE} header.
+     */
+    HttpHeaders contentType(MediaType mediaType);
+
+    /**
      * Returns the immutable view of this headers.
      */
     default HttpHeaders asImmutable() {
--- a/core/src/main/java/com/linecorp/armeria/common/ImmutableHttpHeaders.java
+++ b/core/src/main/java/com/linecorp/armeria/common/ImmutableHttpHeaders.java
@@ -90,6 +90,16 @@ final class ImmutableHttpHeaders implements HttpHeaders {
     }
 
     @Override
+    public MediaType contentType() {
+        return delegate.contentType();
+    }
+
+    @Override
+    public HttpHeaders contentType(MediaType mediaType) {
+        return delegate.contentType(mediaType);
+    }
+
+    @Override
     public boolean isEndOfStream() {
         return delegate.isEndOfStream();
     }
--- a/core/src/main/java/com/linecorp/armeria/common/ImmutableHttpHeaders.java
+++ b/core/src/main/java/com/linecorp/armeria/common/ImmutableHttpHeaders.java
@@ -96,7 +96,7 @@ final class ImmutableHttpHeaders implements HttpHeaders {
 
     @Override
     public HttpHeaders contentType(MediaType mediaType) {
-        return delegate.contentType(mediaType);
+        return unsupported();
     }
 
     @Override
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
@@ -42,7 +42,6 @@ import com.google.common.collect.ImmutableList;
 
 import com.linecorp.armeria.common.AggregatedHttpMessage;
 import com.linecorp.armeria.common.FilteredHttpResponse;
-import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpObject;
 import com.linecorp.armeria.common.HttpParameters;
 import com.linecorp.armeria.common.HttpRequest;
@@ -61,7 +60,6 @@ import com.linecorp.armeria.server.annotation.RequestObject;
 import com.linecorp.armeria.server.annotation.ResponseConverter;
 
 import io.netty.handler.codec.http.HttpConstants;
-import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.QueryStringDecoder;
 import io.netty.util.AsciiString;
 
@@ -708,10 +706,9 @@ final class AnnotatedHttpServiceMethod implements BiFunction<ServiceRequestConte
          * Whether the request is available to be aggregated.
          */
         static boolean aggregationAvailable(HttpRequest req) {
-            final String contentType = req.headers().get(HttpHeaderNames.CONTENT_TYPE);
+            final MediaType contentType = req.headers().contentType();
             // We aggregate request stream messages for the media type of form data currently.
-            return contentType != null &&
-                   MediaType.FORM_DATA.toString().equals(HttpUtil.getMimeType(contentType));
+            return contentType != null && MediaType.FORM_DATA.type().equals(contentType.type());
         }
 
         /**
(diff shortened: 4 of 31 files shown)


core/src/main/java/com/linecorp/armeria/common/ImmutableHttpHeaders.java
```diff
@@ -89,6 +89,16 @@ public HttpHeaders status(HttpStatus status) {
         return unsupported();
     }
 
+    @Override
+    public MediaType contentType() {
+        return unsupported();
```
minwoox: delegate.contentType()?

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -174,6 +174,21 @@ public HttpHeaders status(HttpStatus status) {
...
+        String contentType = get(HttpHeaderNames.CONTENT_TYPE);
+        if (contentType == null) {
+            return null;
+        }
+        return MediaType.parse(contentType);
+    }
+
+    @Override
+    public HttpHeaders contentType(MediaType mediaType) {
+        String contentType = mediaType.toString();
```
minwoox: requireNonNull? final String?

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
```diff
@@ -830,7 +830,7 @@ static void checkResult(org.apache.http.HttpResponse res,
                             int statusCode,
                             @Nullable String body,
                             @Nullable Charset encoding,
-                            @Nullable String contentType) throws IOException {
+                            @Nullable MediaType contentType) throws IOException {
```
hyangtack: How about changing this type to `String` back and parsing the given `contentType` inside the `checkResult` method?

core/src/main/java/com/linecorp/armeria/server/DefaultPathMappingContext.java
```diff
@@ -155,13 +155,9 @@ static PathMappingContext of(VirtualHost virtualHost, String hostname,
 
     @VisibleForTesting
     static MediaType resolveConsumeType(HttpHeaders headers) {
-        final String contentType = headers.get(HttpHeaderNames.CONTENT_TYPE);
+        final MediaType contentType = headers.contentType();
```
hyangtack: Could we need to catch an exception from `headers.contentType()` not to be raised it to the caller?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -669,10 +667,9 @@ static boolean aggregationRequired(AggregationStrategy strategy, HttpRequest req
          * Whether the request is available to be aggregated.
          */
         static boolean aggregationAvailable(HttpRequest req) {
-            final String contentType = req.headers().get(HttpHeaderNames.CONTENT_TYPE);
+            final MediaType contentType = req.headers().contentType();
```
hyangtack: Could we need to catch an exception from here not to be raised it to the caller?

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -174,6 +174,21 @@ public HttpHeaders status(HttpStatus status) {
         return status(status.code());
     }
 
+    @Override
+    public MediaType contentType() {
+        String contentType = get(HttpHeaderNames.CONTENT_TYPE);
+        if (contentType == null) {
+            return null;
+        }
+        return MediaType.parse(contentType);
```
hyangtack: Do we need to cache the result not to parse the string again?

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -174,6 +174,21 @@ public HttpHeaders status(HttpStatus status) {
         return status(status.code());
     }
 
+    @Override
+    public MediaType contentType() {
+        String contentType = get(HttpHeaderNames.CONTENT_TYPE);
```
hyangtack: nit: `final`

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
jmostella: Ah I see it https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/MediaTypeJsonDeserializer.java#L54
or 
https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/encoding/HttpEncodedResponse.java#L167

Just curious is why do we want to hide this exception?

hyangtack: From the annotated http service point of view, I think the user's method should be executed regardless of the media type, especially for the cases which are not using content type negotiation. So ignoring invalid content type is simple and reasonable way to do it.

core/src/main/java/com/linecorp/armeria/server/DefaultPathMappingContext.java
jmostella: if ``DefaultHttpHeaders.contentType`` returns null is that acceptable?

hyangtack: Yes. See https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/DefaultPathMappingContext.java#L152
`consumeType` passing into the constructor is `@Nullable`.

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
```diff
@@ -174,6 +174,17 @@ static HttpHeaders copyOf(HttpHeaders headers) {
      */
     HttpHeaders status(HttpStatus status);
 
+    /**
+     * Gets the {@link HttpHeaderNames#CONTENT_TYPE} header or {@code null} if there is no such header.
```
trustin: How about:

```
Returns the value of the {@code 'content-type'} header.

@return the valid header value if present. {@code null} otherwise.
``

File: core/src/test/java/com/linecorp/armeria/server/HttpServerTest.java [hunk 9]
```diff
@@ -207,8 +207,8 @@ protected void doGet(ServiceRequestContext ctx, HttpRequest req, HttpResponseWri
                                     final HttpData data;
                                     if (pooled) {
                                         final ByteBuf content = PooledByteBufAllocator.DEFAULT
-                                                                                      .buffer(1)
-                                                                                      .writeByte('0' + finalI);
+                                                .buffer(1)
+                                                .writeByte('0' + finalI);
```
trustin: nit: revert?

grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceTest.java
```diff
@@ -123,14 +124,14 @@ public void missingMethod() throws Exception {
...
-                    PathMapping.ofExact("/armeria.grpc.testing.TestService/EmptyCall"),
-                    PathMapping.ofExact("/armeria.grpc.testing.TestService/UnaryCall"),
-                    PathMapping.ofExact("/armeria.grpc.testing.TestService/StreamingOutputCall"),
-                    PathMapping.ofExact("/armeria.grpc.testing.TestService/StreamingInputCall"),
-                    PathMapping.ofExact("/armeria.grpc.testing.TestService/FullDuplexCall"),
-                    PathMapping.ofExact("/armeria.grpc.testing.TestService/HalfDuplexCall"),
-                    PathMapping.ofExact("/armeria.grpc.testing.TestService/UnimplementedCall"));
+        assertThat(grpcService.pathMappings())
+                .containsExactlyInAnyOrder(
+                        PathMapping.ofExact("/armeria.grpc.testing.TestService/EmptyCall"),
```
trustin: nit: revert?

jmostella:
Thanks, updated!

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -174,6 +175,32 @@ public HttpHeaders status(HttpStatus status) {
...
+            return contentType;
+        }
+
+        final String contentTypeString = get(HttpHeaderNames.CONTENT_TYPE);
+        if (contentTypeString == null) {
+            return null;
+        }
+
+        try {
+            return MediaType.parse(contentTypeString);
```
anuraaga: Cache the media type for multiple calls?

this.contentType = parse
return this.contentType

hyangtack:
@jmostella I recently added some `RequestConverter` implementations such as:

-
 https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java#L66
- https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java#L43
- https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/annotation/StringRequestConverterFunction.java#L41
 
Would you please rebase your code with upstream, and then replace them with your new method? 😃

jmostella:
@hyangtack OK no problem!

core/src/main/java/com/linecorp/armeria/common/ImmutableHttpHeaders.java
```diff
@@ -89,6 +89,16 @@ public HttpHeaders status(HttpStatus status) {
...
     }
 
+    @Override
+    public MediaType contentType() {
+        return delegate.contentType();
+    }
+
+    @Override
+    public HttpHeaders contentType(MediaType mediaType) {
+        return delegate.contentType(mediaType);
```
minwoox: this should be unsupported()?

jmostella: Oh I see, sorry my mistake. Its supposed to be Immutable

trustin:
@kojilin Any comments?

jmostella:
I also made some new changes to:

```
THttpService.findSerializationFormat
THttpClientDelegate
GrpcService
```
Can someone take a second look it?

Also I avoided some of the Grpc test stuff since there weren't any constants defined so I didn't see much value changing, without adding constants for the new media types

grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
```diff
@@ -221,33 +220,13 @@ public void serviceAdded(ServiceConfig cfg) throws Exception {
...
-        try {
-            mediaType = MediaType.parse(contentType);
-        } catch (IllegalArgumentException e) {
-            logger.debug("Failed to parse the 'content-type' header: {}", contentType, e);
-            return null;
-        }
-
-        for (SerializationFormat format : supportedSerializationFormats) {
-            if (format.isAccepted(mediaType)) {
+            if (format.isAccepted(contentType)) {
```
anuraaga: Since media type is parsed once in `DefaultPathMappingContext` and not parsed again here, this LGTM.

trustin:
Will merge once the build passes.

trustin:
Thanks, @jmostella !

PR #905, opened 2017-12-12
Cancel subscription after writing the object is complete for better completion signaling

anuraaga:
Currently, the subscription is cancelled as soon as the response subscriber receives an end-of-stream object, before it is actually written. This makes sense from a reactive streams perspective, but we also have the concept of a completion future which is tied into the stream's lifecycle. When the subscription is canceled, the completion future is set to a cancelled state as well - depending on the stream's implementation, this could be synchronous, meaning the completion future is notified before the last write can even happen. If someone is listening for the completion future to do a cleanup type of thing, it might happen even though things were successful. This is the result of a slightly confusing correlation between the stream lifecycle and completion future.

This PR changes the response/request subscriber to only cancel after finishing their write. This allows the completion future to be completed successfully on stream close before the subscriber calls cancel, for cleaner semantics.

Admittedly, I'm not sure why this only seems to repro when the request is HTTP/1.0 (not 1.1), but it seems to be better behavior either way.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
@@ -247,6 +247,10 @@ final class HttpRequestSubscriber implements Subscriber<HttpObject>, ChannelFutu
         if (flush) {
             ctx.flush();
         }
+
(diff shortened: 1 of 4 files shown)


trustin:
Thanks, @anuraaga !

PR #907, opened 2017-12-13
Change specifications of the response converter of the annotated serv…

hyangtack:
…ice.

Motivation:
Recently we added `@RequestConverter` and `@ExceptionHandler` to annotated service. They have basically similar behavior such as:
- Iterating a converter (or handler) list when applying it.
- Using `accept()` method in order to check whether the converter (or handler) is able to consume the input.

`@Converter`, however, has its own behavior unlike the others. It maps a type of a result object to a converter so that it is only able to find a converter with the type. It is not customizable and is not useful in the real world.
So, it would be better to have them the same behavior and design as the `@RequestConverter` to make the response converter flexible.

Modifications:
- Rename `@Converter` to `@ResponseConverter` and `ResponseConverter` to `ResponseConverterFunction` for consistency.
- Rename method names of `*Function` interfaces so that a user can write a single class which inherits all interfaces.
- Change converter map to converter list.
- Add more `annotatedService()` methods to `ServerBuilder` and `AbstractVirtualHostBuilder`.
- Use default exception handlers and default request converters regardless of specifying them by a user.


core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -283,75 +284,193 @@ B service(T serviceWithPathMappings, Function<T, R> decorator) {
...
      */
-    public B annotatedService(
-            Object service, Map<Class<?>, ResponseConverter> converters,
-            Function<Service<HttpRequest, HttpResponse>,
-                     ? extends Service<HttpRequest, HttpResponse>> decorator) {
-        return annotatedService("/", service, converters, decorator);
+    public B annotatedService(Object service,
+                              List<ExceptionHandlerFunction> exceptionHandlers,
+                              List<RequestConverterFunction> requestConverters,
+                              List<ResponseConverterFunction> responseConverters) {
```
trustin: I feel like the overloaded `annotatedService()` methods are now quite confusing because they look pretty close to each other. How about just accepting `Iterable<?>` and `Object...`, and then doing `instanceof` and downcasts? i.e.

```java
public B annotatedService(Object service, Object... exceptionHandlersAndConverters) {
    return annotatedService(service, ImmutableList.copyOf(requireNonNull(exceptionHandlersAndConverters)));
}

public B annotatedService(Object service, Iterable<?> exceptionHandlersAndConverters) {
    for (Object o : exceptionHandlersAndConverters) {
        boolean added = false;
        if (o instanceof ExceptionHandler) {
            exceptionHandlers.add((ExceptionHandlerFunction) o);
        }
        if (o instanceof RequestConverterFunction) {
            requestConverters.add((RequestConverterFunction) o);
        }
        if (o instanceof ResponseConverterFunction) {
            responseConverters.add((ResponseConverterFunction) o);
        }
        if (!added) {
            throw new IllegalArgumentException(...);
        }
    }
    ...
}
```

Please note that I did not use `else if` when checking the type of an object. This allows a user pass an object that implements more than one interface, e.g.

```java
public class MyAnnotatedServiceExtension implements ExceptionHandlerFunction, RequestConverterFunction, ResponseConverterFunction { ... }
```

However, as you know, the 3 interfaces all have the method `accept` which will be confusing when a class implements all of them. Therefore, we may need to rename `accept()` to `canHandle(Throwable)` and `canConvert(...)`. What do you think? Do you think it is desirable for a user to implement the 3 interfaces in a single class for his/her convenience?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -132,65 +162,53 @@ public Object apply(ServiceRequestContext ctx, HttpRequest req) {
...
+        }
+
+        for (final ResponseConverterFunction func : responseConverters) {
+            if (func.accept(result)) {
+                try (SafeCloseable ignored = RequestContext.push(ctx, false)) {
+                    return func.convert(result);
+                } catch (Exception e) {
+                    throw new IllegalStateException(
+                            "Response converter " + func.getClass().getName() +
+                            " cannot convert a result to HttpResponse:", e);
```
trustin: How about: `" to HttpResponse: " + result, e` ?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -399,51 +365,64 @@ private static ResponseConverter converter(Method method) {
...
     }
 
     /**
      * Returns a request converter list which is specified by {@link RequestConverter} annotations.
      */
-    private static List<RequestConverterFunction> requestConverters(Method targetMethod,
-                                                                    Class<?> targetClass) {
+    private static ImmutableList.Builder<RequestConverterFunction> requestConverters(Method
+                                                                                             targetMethod,
+                                                                                     Class<?> targetClass) {
```
trustin: This doesn't look nice :-)

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -399,51 +365,64 @@ private static ResponseConverter converter(Method method) {
...
+                                requestConverterFunctions);
     }
 
     /**
-     * Returns a list of objects which are specified as the {@code value} of annotation {@code T}.
+     * Returns a response converter list which is specified by {@link ResponseConverter} annotations.
+     */
+    private static ImmutableList.Builder<ResponseConverterFunction> responseConverters(Method
+                                                                                               targetMethod,
+                                                                                       Class<?> targetClass) {
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
hyangtack: Oops! Fixed!

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
hyangtack: Oops! Fixed!

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -283,75 +281,75 @@ B service(T serviceWithPathMappings, Function<T, R> decorator) {
      * Binds the specified annotated service object under the path prefix {@code "/"}.
      */
     public B annotatedService(Object service) {
-        return annotatedService("/", service);
+        return annotatedService("/", service, ImmutableList.of(), Function.identity());
     }
 
     /**
      * Binds the specified annotated service object under the path prefix {@code "/"}.
```
trustin: Could you add `@param exceptionHandlersAndConverters` to explain what it has to contain?

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -283,75 +281,75 @@ B service(T serviceWithPathMappings, Function<T, R> decorator) {
...
      */
-    public B annotatedService(
-            Object service,
-            Function<Service<HttpRequest, HttpResponse>,
-                     ? extends Service<HttpRequest, HttpResponse>> decorator) {
-        return annotatedService("/", service, decorator);
+    public B annotatedService(Object service,
+                              Iterable<?> exceptionHandlersAndConverters) {
```
trustin: I wonder what happens if we add a version that accepts `Object...` instead of `Iterable<?>`. Compilation errors?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -388,84 +330,96 @@ private static ResponseConverter converter(Method method) {
...
+                } catch (Exception e) {
+                    throw new IllegalStateException(
+                            "A class specified in @" + annotation.getClass().getSimpleName() +
+                            " annotation must have an accessible default constructor: " + clazz.getName(), e);
+                }
+            }));
+        } catch (ClassCastException e) {
+            throw new IllegalArgumentException(
+                    "A class specified in @" + annotation.getClass().getSimpleName() +
+                    " annotation cannot be casted to " + expectedType, e);
```
trustin: casted -> cast

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -388,84 +330,96 @@ private static ResponseConverter converter(Method method) {
...
         }
     }
 
     /**
-     * Returns a new instance of the specified {@link Class}.
+     * Returns an object which is returned by {@code value()} method of the specified object {@code a}.
```
trustin: `object {@code a}` -> `annotation {@code a}`?

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandlerFunction.java
```diff
@@ -39,12 +39,12 @@
     /**
      * Returns whether the specified {@code cause} is acceptable.
      */
-    default boolean accept(Throwable cause) {
+    default boolean acceptableException(Throwable cause) {
```
trustin: How about just `canHandleException`?

core/src/main/java/com/linecorp/armeria/server/annotation/RequestConverterFunction.java
```diff
@@ -32,12 +32,12 @@
      * Returns whether this converter is able to convert the specified {@code request} to
      * {@code expectedResultType}.
      */
-    default boolean accept(AggregatedHttpMessage request, Class<?> expectedResultType) {
+    default boolean acceptableRequest(AggregatedHttpMessage request, Class<?> expectedResultType) {
```
trustin: How about just `canConvertRequest?

core/src/main/java/com/linecorp/armeria/server/annotation/ResponseConverterFunction.java
```diff
@@ -0,0 +1,42 @@
...
+ * @see ResponseConverter
+ */
+@FunctionalInterface
+public interface ResponseConverterFunction {
+
+    /**
+     * Returns whether this converter is able to convert the specified {@code result} to
+     * {@link HttpResponse}.
+     */
+    default boolean acceptableResponse(Object result) {
```
trustin: How about just `canConvertResponse`?

core/src/main/java/com/linecorp/armeria/server/annotation/ExceptionHandlerFunction.java
hyangtack: Ok, `can*` would be better for consistency. :-)

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
hyangtack: I will try to do it.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -130,86 +160,76 @@ public Object apply(ServiceRequestContext ctx, HttpRequest req) {
...
-                    return func.handle(ctx, req, rootCause);
+                    return func.handleException(ctx, req, rootCause);
                 } catch (Exception e) {
                     logger.warn("Unexpected exception from an exception handler {}:",
                                 func.getClass().getName(), e);
                 }
             }
         }
-        return ExceptionHandlerFunction.DEFAULT.handle(ctx, req, rootCause);
+        return ExceptionHandlerFunction.DEFAULT.handleException(ctx, req, rootCause);
```
minwoox: How about logging the exception here so that user could know the root cause when none of the exceptionHandlers could not catch the exception???????

hyangtack: Ok, good idea. I will write warning log on ExceptionHandler.DEFAULT instance.

hyangtack:
@minwoox Thanks! Updated!

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -283,75 +284,89 @@ B service(T serviceWithPathMappings, Function<T, R> decorator) {
...
-                     ? extends Service<HttpRequest, HttpResponse>> decorator) {
-        return annotatedService("/", service, decorator);
+        return annotatedService("/", service, Function.identity(), ImmutableList.of());
     }
 
     /**
      * Binds the specified annotated service object under the path prefix {@code "/"}.
+     *
+     * @param exceptionHandlersAndConverters instances of {@link ExceptionHandlerFunction},
+     *                                       {@link RequestConverterFunction} and
```
trustin: and -> and/or ?

hyangtack:
Rebased

hyangtack:
Thanks, reviewers!

borag:
Does this add my instance to all lists ? I guess there is a problem here

        for (final Object o : exceptionHandlersAndConverters) {
            if (o instanceof RequestConverterFunction) {
                requestConverters.add((RequestConverterFunction) o);
            } else if (o instanceof ResponseConverterFunction) {
                responseConverters.add((ResponseConverterFunction) o);
            } else if (o instanceof ExceptionHandlerFunction) {
                exceptionHandlers.add((ExceptionHandlerFunction) o);
            } else {
                throw new IllegalArgumentException(o.getClass().getName() +
                                                   " is neither an exception handler nor a converter.");
            }
        }

trustin:
Hi @borag,

This is also a very old PR, so could you open a new issue and elaborate on your problem there instead?

PR #908, opened 2017-12-13
[WIP] Test if micrometer snapshot does not fail anything

trustin:
Just a temporary PR for testing. Will close once I have confidence of no regression.


trustin:
No regression found.

PR #909, opened 2017-12-13
Change AbstractHttpService API to always return HttpResponse instead …

anuraaga:
…of passing in DefaultHttpResponse.

Discussion in #850 - this change allows service implementations to choose an `HttpResponse` that is appropriate for a given codepath, with little to no loss in writability.

`HttpRequest` and `HttpResponse` are the entrypoints for creating request and response objects now - `DefaultHttpRequest` and `DefaultHttpResponse` and `DeferredHttpResponse` are also planned to be made private (`HttpResponse.from(CompletableFuture)` seems like a more easy-to-use and natural API for users than `DeferredHttpResponse.delegate` so I think we can consolidate on it).

Most `respond` methods from `HttpResponseWriter` have been deprecated as it should now be more common to use `HttpResponse.of` or `HttpResponse.from` for its use cases. One easy to maintain one is left for advanced use cases.

Result is less exposure of implementation classes for more flexibility, less API duplication, reducing of "two ways of doing something" (I noticed that all of `HttpResponse.of` methods are accessible as `DefaultHttpResponse.of` as well, I guess interfaces with static methods should not have public implementations to prevent this sort of confusion), less object allocations / memory usage for many cases like `HttpFileService`, and probably thrift too.

This is a highly deprecating change, but no explicit breaks. Templates for helping in conversion towards when we actually do remove the APIs below.

Some IntelliJ templates
##### Converting service implementations

Search template:
```java
protected void doGet(ServiceRequestContext $CtxVar$, HttpRequest $ReqVar$, HttpResponseWriter $ResVar$) {
  $MethodBody$;
}
```

Replacement template:
```java
protected HttpResponse doGet(ServiceRequestContext $CtxVar$, HttpRequest $ReqVar$) {
  HttpResponseWriter $ResVar$ = HttpResponse.streaming();
  $MethodBody$;
  return $ResVar$;
}
```

##### Converting manual initializations

Search:
```java
DefaultHttpResponse $Var$ = new DefaultHttpResponse();
```

Replace:
```java
HttpResponseWriter $Var$ = HttpResponse.streaming();
```

Search:
```java
DefaultHttpRequest $Var$ = new DefaultHttpRequest();
```

Replace:
```java
HttpRequestWriter $Var$ = HttpRequest.streaming();
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
@@ -328,11 +328,4 @@ public interface HttpResponse extends Response, StreamMessage<HttpObject> {
         subscribe(aggregator, executor);
         return future;
     }
-
-    /**
-     * Whether this is a deferred {@link HttpResponse}, created by {@link #from(CompletionStage)}.
-     */
-    default boolean isDeferred() {
-        return false;
-    }
 }
--- a/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
@@ -43,8 +43,36 @@ import io.netty.util.concurrent.EventExecutor;
  */
 public interface HttpResponse extends Response, StreamMessage<HttpObject> {
 
-    // Note: Ensure we provide the same set of `of()` methods with the `of()` and `respond()` methods of
-    //       HttpResponseWriter and AggregatedHttpMessage for consistency.
+    // Note: Ensure we provide the same set of `of()` methods with the `of()` methods of
+    //       AggregatedHttpMessage for consistency.
+
+    /**
+     * Creates a new HTTP response that can stream an arbitrary number of {@link HttpObject} to the client.
+     * The first object written must be of type {@link HttpHeaders}.
+     */
+    static HttpResponseWriter streaming() {
+        return new DefaultHttpResponse();
+    }
+
+    /**
+     * Creates a new HTTP response that delegates to the {@link HttpResponse} produced by the specified
+     * {@link CompletionStage}. If the specified {@link CompletionStage} fails, the returned response will be
+     * closed with the same cause as well.
+     */
+    static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
+        requireNonNull(stage, "stage");
+        final DeferredHttpResponse res = new DeferredHttpResponse();
+        stage.whenComplete((delegate, thrown) -> {
+            if (thrown != null) {
+                res.close(Throwables.getRootCause(thrown));
+            } else if (delegate == null) {
+                res.close(new NullPointerException("delegate stage produced a null response: " + stage));
+            } else {
+                res.delegate(delegate);
+            }
+        });
+        return res;
+    }
 
     /**
      * Creates a new HTTP response of the specified {@code statusCode} and closes the stream if the
@@ -61,7 +89,7 @@ public interface HttpResponse extends Response, StreamMessage<HttpObject> {
     static HttpResponse of(HttpStatus status) {
         requireNonNull(status, "status");
         if (status.codeClass() == HttpStatusClass.INFORMATIONAL) {
-            DefaultHttpResponse res = new DefaultHttpResponse();
+            HttpResponseWriter res = streaming();
             res.write(HttpHeaders.of(status));
             return res;
         } else if (isContentAlwaysEmpty(status)) {
@@ -254,7 +282,7 @@ public interface HttpResponse extends Response, StreamMessage<HttpObject> {
      * Creates a new failed HTTP response.
      */
     static HttpResponse ofFailure(Throwable cause) {
-        final DefaultHttpResponse res = new DefaultHttpResponse();
+        final HttpResponseWriter res = streaming();
         res.close(cause);
         return res;
     }
@@ -267,26 +295,6 @@ public interface HttpResponse extends Response, StreamMessage<HttpObject> {
     @Deprecated
     static HttpResponse ofFailed(Throwable cause) {
         return ofFailure(cause);
-    }
-
-    /**
-     * Creates a new HTTP response that delegates to the {@link HttpResponse} produced by the specified
-     * {@link CompletionStage}. If the specified {@link CompletionStage} fails, the returned response will be
-     * closed with the same cause as well.
-     */
-    static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
-        requireNonNull(stage, "stage");
-        final DeferredHttpResponse res = new DeferredHttpResponse();
-        stage.whenComplete((delegate, thrown) -> {
-            if (thrown != null) {
-                res.close(Throwables.getRootCause(thrown));
-            } else if (delegate == null) {
-                res.close(new NullPointerException("delegate stage produced a null response: " + stage));
-            } else {
-                res.delegate(delegate);
-            }
-        });
-        return res;
     }
 
     @Override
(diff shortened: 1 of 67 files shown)


trustin:
I wonder if we can:

- keep the old methods in `AbstractHttpService` and deprecate them
- add the new methods whose default implementations delegate to the old methods

.. for even smoother migration path?

anuraaga:
Good idea! Since the METHOD_NOT_ALLOWED versions are not commonly used, not using `FixedHttpRseponse` isn't a big deal at all. Updated PR to have no hard breaks.

trustin:
Created a new label 'deprecation' so that a PR with deprecation can be found more easily. :-)

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -202,64 +198,63 @@ private void doGetOrPost(HttpRequest req, HttpResponseWriter res) {
...
-                                "METHOD: %s|ACCEPT: %s|BODY: %s",
-                                req.method().name(), accept,
-                                aReq.content().toString(StandardCharsets.UTF_8))));
-                        res.close();
-                    })).exceptionally(CompletionActions::log);
+                        return HttpResponse.of(
+                                HttpHeaders.of(HttpStatus.OK)
+                                           .set(HttpHeaderNames.CACHE_CONTROL, "alwayscache"),
+                                HttpData.ofUtf8(String.format(
+                                        Locale.ENGLISH,
```
trustin: nit: You can use `HttpData.ofUtf8(format, ...args)` instead.

core/src/test/java/com/linecorp/armeria/client/endpoint/StickyEndpointSelectionStrategyTest.java
```diff
@@ -99,9 +100,8 @@ public void select() {
 
     private static ClientRequestContext contextWithHeader(String k, String v) {
         ClientRequestContext ctx = mock(ClientRequestContext.class);
-        HttpRequest testReq = new DefaultHttpRequest();
-        testReq.headers().set(AsciiString.of(k), v);
-        when(ctx.request()).thenReturn(testReq);
+        when(ctx.request()).thenReturn(HttpRequest.of(HttpHeaders.of(HttpMethod.GET, "/")
+                                                                 .set(AsciiString.of(k), v)));
```
trustin: I'm curious if it is a good idea to provide a method like `HttpHeaders.toHttpRequest()` and `HttpHeaders.toHttpResponse()`. This will give a user two ways to create an `HttpRequest` or `HttpResponse`, which may be good or bad. What do you think? (We recently deprecated `AggregatedHttpMessage.toHttp*()` in favor of `Http*.of(..)` by the way, but wanted to know which a user would prefer.)

anuraaga: It's a difficult question, especially since the number of characters is exactly the same ;) I like the reduced cognitive load of having only one entrypoint to HttpRequest/Response. Since the buck stops at the `toHttpRequest` in general, I don't think there's particular value to being fluent either. But it's a personal preference so wonder what others think.

anuraaga:
Rebased

core/src/test/java/com/linecorp/armeria/client/endpoint/StickyEndpointSelectionStrategyTest.java
trustin: Probably not the scope of this PR. Let us keep this in mind though. :-)

core/src/main/java/com/linecorp/armeria/common/HttpResponseWriter.java
```diff
@@ -173,7 +192,10 @@ default void respond(HttpStatus status, MediaType mediaType, HttpData content,
 
     /**
      * Writes the specified HTTP response and closes the stream.
+     *
+     * @deprecated Use {@link HttpResponse#of(AggregatedHttpMessage)}.
      */
+    @Deprecated
```
trustin: Maybe some users will find this method useful when he or she initially created a streaming response and then realized that the stream is actually very small and thus decided to write a fixed response?

core/src/main/java/com/linecorp/armeria/common/HttpResponseWriter.java
```diff
@@ -199,4 +221,12 @@ default void respond(AggregatedHttpMessage res) {
 
         close();
     }
+
+    /**
+     * Write the given {@link HttpObject} and close the stream.
+     */
+    default void respond(HttpObject obj) {
```
trustin: - `respond` -> `close` ?
- Could we move this method to `StreamWriter`? i.e. `default void close(T obj)` If we are worried about the confusion with ` close(Throwable)` then we can copy this method to `HttpRequestWriter` instead.

anuraaga:
Good ideas - updated with both

core/src/main/java/com/linecorp/armeria/common/HttpResponseWriter.java
```diff
@@ -192,10 +192,7 @@ default void respond(HttpStatus status, MediaType mediaType, HttpData content,
 
     /**
      * Writes the specified HTTP response and closes the stream.
-     *
-     * @deprecated Use {@link HttpResponse#of(AggregatedHttpMessage)}.
      */
-    @Deprecated
     default void respond(AggregatedHttpMessage res) {
```
trustin: I'm curious if we should rename this method to `close` as well while having a deprecated `respond(AggregatedHttpMessage)`. What do you think?

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -320,4 +328,11 @@ static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
         subscribe(aggregator, executor);
         return future;
     }
+
+    /**
+     * Whether this is a deferred {@link HttpResponse}, created by {@link #from(CompletionStage)}.
+     */
+    default boolean isDeferred() {
```
hyangtack: Just curious. When is this used? I saw its usage only in test cases, so I don't know when to use.

core/src/main/java/com/linecorp/armeria/common/HttpResponseWriter.java
anuraaga: Makes sense, it's more consistent that way

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
anuraaga: Yeah it's only in the test case - maybe I can just remove those assertions, @minwoox what do you think?

minwoox: I cannot figure out when to use neither. :-)

core/src/main/java/com/linecorp/armeria/common/HttpResponseWriter.java
```diff
@@ -140,7 +156,10 @@ default void respond(HttpStatus status, MediaType mediaType, HttpData content) {
      * @param mediaType the {@link MediaType} of the response content
      * @param content the content of the response
      * @param trailingHeaders the trailing HTTP headers
+     *
+     * @deprecated Use {@link HttpResponse#of(HttpStatus, MediaType, HttpData, HttpHeaders)}.
      */
+    @Deprecated
     default void respond(HttpStatus status, MediaType mediaType, HttpData content,
```
minwoox: Don't we need to deprecate the `response(AggregatedHttpMessage)` method as well?

minwoox: Oops~ I misread forget about this. :-)

anuraaga:
I thought I had replied but don't see it - apologies if double-reply. Went ahead and removed isDeferred and its test usage to keep the API cleaner :)

trustin:
Thanks a lot, @anuraaga !

PR #911, opened 2017-12-13
Fix flaky test: CaffeineMetricSupportTest.aggregationAfterGC()

trustin:
Motivation:

The offending test case keeps the MockLoadingCache in a WeakReference
from its beginning of life. As a result, CaffeineMetricSupport.setup()
sometimes throws a NullPointerException because `WeakReference<MockLoadingCache>.get()`
can return null even before our System.gc() call.

Modifications:

- Keep the strong reference until System.gc()
- More friendly NPE message

Result:

- Build stability


trustin:
Related stack trace:

```
com.linecorp.armeria.internal.metric.CaffeineMetricSupportTest > aggregationAfterGC FAILED
        at com.linecorp.armeria.internal.metric.CaffeineMetricSupport.setup(CaffeineMetricSupport.java:61)
        at com.linecorp.armeria.internal.metric.CaffeineMetricSupportTest.aggregationAfterGC(CaffeineMetricSupportTest.java:165)
```

PR #913, opened 2017-12-18
Make JacksonRequestConverterFunction support +json media type as well

minwoox:
Motivation:
Not only `application/json` has JSON content, but also many variants whose subtype
contains `+json` has JSON contents. It is reasonable that `JacksonRequestConveterFunction`
supports converting the requests who have those headers.
I found a similar issue: https://github.com/NativeScript/NativeScript/issues/2593

Modifications:
- make `JacksonRequestConverterFunction` convert requests who have `+json` header
- Add `json-patch+json` media type

Result:
- Robust `JacksonRequestConveterFunction`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
@@ -41,6 +41,8 @@ public class JacksonRequestConverterFunction implements RequestConverterFunction
 
     private static final Logger logger = LoggerFactory.getLogger(JacksonRequestConverterFunction.class);
 
+    private static final ObjectMapper DEFAULT_MAPPER = new ObjectMapper();
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -64,7 +66,8 @@ public JacksonRequestConverterFunction(ObjectMapper mapper) {
     @Override
     public boolean canConvertRequest(AggregatedHttpMessage request, Class<?> expectedResultType) {
         final MediaType contentType = request.headers().contentType();
-        if (contentType != null && contentType.is(MediaType.JSON)) {
+        if (contentType != null && (contentType.is(MediaType.JSON) ||
+                                    contentType.subtype().contains("+json"))) {
```
trustin: `endsWith`?

PR #915, opened 2017-12-18
Throw an exception raised while converting a request to a `@RequestObject` instance

hyangtack:
Modifications:
- Do not ignore an exception from RequestConverterFunction#convert.
- Do not allow returning `null` from RequestConverterFunction#convert.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.server;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -366,21 +367,18 @@ private static Parameter createOptionalSupportedParam(java.lang.reflect.Paramete
...
-                                values[i] = func.convertRequest(message, entry.type());
-                                break;
-                            } catch (Exception e) {
-                                // Go to the next request converter if the conversion is failed.
-                                logger.warn("Request converter {} cannot convert a request to {}",
-                                            func.getClass().getName(), entry.type().getName(), e);
-                            }
+                            values[i] = func.convertRequest(message, entry.type());
+                            checkArgument(values[i] != null,
+                                          "Returned 'null' from the converter '" +
```
minwoox: nit: how about just 'null' is returned?

trustin:
Thanks, @hyangtack !

PR #917, opened 2017-12-19
Write pooled objects to the request so pooled-aware subscribers can p…

anuraaga:
…rocess ByteBuf directly.

Since our stream messages are aware of pooling, it should be ok to write pooled objects to the request for less copying with pooling-aware subscribers.

After
```
Benchmark                         (clientType)   Mode  Cnt    Score    Error  Units
DownstreamSimpleBenchmark.simple        OKHTTP  thrpt   20  490.181 ± 33.306  ops/s
```

Before
```
Benchmark                         (clientType)   Mode  Cnt    Score    Error  Units
DownstreamSimpleBenchmark.simple        OKHTTP  thrpt   20  482.963 ± 30.219  ops/s
```


trustin:
Thanks, @anuraaga !

PR #918, opened 2017-12-19
Precompute mapping of string content types to well known types to imp…

anuraaga:
…rove mediatype parsing when content types are known.

`MediaType.parse` is relatively slow as it supports all the variability of a content type. As servers only deal with a small number of media types in practice, it is wasteful to do this parse for every single request. Conveniently, servers also have a good idea of what media types they will support, so we can just precompute a mapping from `String` to `MediaType`.

A map like this should save memory / minimize garbage as all hits will result in sharing the same `MediaType` instance among requests. `MediaType.parse` itself has an interning cache, but it only recognizes its well known types, while this one also includes our `SerializationFormat`s.

A `LoadingCache` could also be used to be more general, but I feel like it would help a more limited use case (we'd probably want to add it as a layer after the precomputed map either way since it would have some synchronization overhead).

Benchmark
```
Benchmark                           Mode  Cnt         Score         Error  Units
HttpHeadersBenchmark.parseKnown    thrpt   20  13516374.953 ±  758315.535  ops/s
HttpHeadersBenchmark.parseUnknown  thrpt   20    700837.841 ±   38172.512  ops/s
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/MediaType.java
+++ b/core/src/main/java/com/linecorp/armeria/common/MediaType.java
@@ -795,7 +795,7 @@ public final class MediaType {
      */
     public static MediaType parse(String input) {
         checkNotNull(input);
-        final MediaType wellKnown = KNOWN_TYPES_BY_STRING.get(input);
+        final MediaType wellKnown = KnownTypesByString.get(input);
         if (wellKnown != null) {
             return wellKnown;
         }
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/common/MediaType.java
```diff
@@ -479,6 +480,10 @@ private static MediaType addKnownType(MediaType mediaType) {
     public static final MediaType XRD_UTF_8 = createConstantUtf8(APPLICATION_TYPE, "xrd+xml");
     public static final MediaType ZIP = createConstant(APPLICATION_TYPE, "zip");
 
+    public static Set<MediaType> knownTypes() {
+        return KNOWN_TYPES.keySet();
```
trustin: and I'm not sure if it's a good idea to add this to the public API. What's the definition of 'known' here? It seems confusing to me.

core/src/main/java/com/linecorp/armeria/common/SerializationFormat.java
```diff
@@ -188,7 +188,7 @@ private static void checkMediaType(Multimap<MediaType, SerializationFormat> simp
     }
 
     /**
-     * Returns all available {@link SessionProtocol}s.
+     * Returns all available {@link SerializationFormat}s.
```
trustin: Oops! Thank you :-)

anuraaga:
Good point - moved

trustin:
@anuraaga Seems like there's some cyclic dependencies involved during class initialization. Could you check the build log?

trustin:
Great job, @anuraaga ! :-)

Issue #887, opened 2017-12-01
Investigate GrpcClientTest.serverStreamingShouldBeFlowControlled


hyangtack:
https://ci.appveyor.com/project/trustin/armeria/build/master.1680#L1814

```
com.linecorp.armeria.client.grpc.GrpcClientTest > serverStreamingShouldBeFlowControlled FAILED
    org.junit.ComparisonFailure: expected:<[payload {
      body: "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\00
...(edited)...
0\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000"
    }
    ]> but was:<[Status{code=UNKNOWN, description=null, cause=com.linecorp.armeria.common.ClosedSessionException
    	at com.linecorp.armeria.common.ClosedSessionException.get(ClosedSessionException.java:34)
    	at com.linecorp.armeria.client.Http2ResponseDecoder.onStreamClosed(Http2ResponseDecoder.java:102)
    	at io.netty.handler.codec.http2.DefaultHttp2Connection.notifyClosed(DefaultHttp2Connection.java:349)
    	at io.netty.handler.codec.http2.DefaultHttp2Connection$ActiveStreams.removeFromActiveStreams(DefaultHttp2Connection.java:985)
    	at io.netty.handler.codec.http2.DefaultHttp2Connection$ActiveStreams$2.process(DefaultHttp2Connection.java:946)
    	at io.netty.handler.codec.http2.DefaultHttp2Connection$ActiveStreams.decrementPendingIterations(DefaultHttp2Connection.java:1007)
[...]

PR #919, opened 2017-12-19
Only log when grpc failure happens on CI, where socket seems to be my…

anuraaga:
…steriously closed during tests.

Even when I loop through the test thousands of time, can't repro a failure locally, so suspecting it is slow-machine specific. Though if anyone wants to help me troubleshoot what else might cause `Stream XX does not exist` (log is in the issue), it would be nice to be sure there isn't something wrong with the netty stream handling.

Works around #887


trustin:
Thanks, @anuraaga !

PR #920, opened 2017-12-19
Enhance request converting feature

hyangtack:
Modifications:
- Add `ServiceRequestContext` parameter to `convert*` methods of the converters and the `handleException` method of the exception handler.
- Remove `can*` method from the converters and exception handlers.
- Fall through to the next request converter if the converter raises `FallthroughException`. It was handled by `can*` method of the converter.
- Can specify a class of `RequestConverterFunction` as the value of @RequestObject` so that a user can specify a request converter explicitly for a class of the annotated parameter.


core/src/main/java/com/linecorp/armeria/server/annotation/FallthroughException.java
```diff
@@ -0,0 +1,51 @@
...
+package com.linecorp.armeria.server.annotation;
+
+import com.linecorp.armeria.common.Flags;
+import com.linecorp.armeria.common.util.Exceptions;
+
+/**
+ * A {@link RuntimeException} raised for falling through to the next something. It would be raised from
+ * {@link ExceptionHandlerFunction}, {@link RequestConverterFunction} and/or {@link ResponseConverterFunction}.
+ */
+public final class FallthroughException extends RuntimeException {
```
trustin: Probably OK to hide this from the public API?

hyangtack: Ok! Changed to package-local. Also, removed public `get()` method and renamed `singleton()` to `get()`.

minwoox:
Thanks @hyangtack~!!

PR #923, opened 2017-12-21
Remove HttpRequest.isKeepAlive()

trustin:
Motivation:

HttpRequest.isKeepAlive() was originally meant to be server-side only
transport-level detail, which means:

- It has no use on the client side. Creating a request with the false
  keepAlive property has no effect at all.
- There's no notion of keep-alive in HTTP/2.

Modification:

- Move HttpRequest.keepAlive down to DecodedHttpRequest so that it's
  kept only for the server-side decoder-level
- Remove HttpRequest.streaming() which creates an invalid request.

Result:

- Less noise in the public API
- A user does not have a way anymore to know if the current request will
  keep the connection alive or not. It was transport-layer detail anyway.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/internal/DefaultHttpRequestTest.java
+++ b/core/src/test/java/com/linecorp/armeria/internal/DefaultHttpRequestTest.java
@@ -65,7 +65,7 @@ public class DefaultHttpRequestTest {
     @Test
     public void abortedAggregation() {
         final Thread mainThread = Thread.currentThread();
-        final DefaultHttpRequest req = new DefaultHttpRequest(HttpHeaders.of(HttpMethod.GET, "/foo"), true);
+        final DefaultHttpRequest req = new DefaultHttpRequest(HttpHeaders.of(HttpMethod.GET, "/foo"));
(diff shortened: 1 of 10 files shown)


core/src/test/java/com/linecorp/armeria/internal/DefaultHttpRequestTest.java
```diff
@@ -65,7 +66,7 @@ public DefaultHttpRequestTest(boolean executorSpecified) {
     @Test
     public void abortedAggregation() {
         final Thread mainThread = Thread.currentThread();
-        final DefaultHttpRequest req = new DefaultHttpRequest(HttpHeaders.of(HttpMethod.GET, "/foo"), true);
+        final HttpRequestWriter req = new DefaultHttpRequest(HttpHeaders.of(HttpMethod.GET, "/foo"));
```
anuraaga: Since this is `DefaultHttpRequestTest`, it might make sense to use the concrete class here.

trustin: Agreed. Fixed :-)

trustin:
Thanks, reviewers.

PR #924, opened 2017-12-21
Don't release when message parse fails since parser takes ownership o…

anuraaga:
…f the buffer. Do release when an unexpected error happens before parsing though.


grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
```diff
@@ -277,34 +277,35 @@ public void setCompression(String compressorName) {
 
     @Override
     public void messageRead(ByteBufOrStream message) {
-        boolean success = false;
+
         final I request;
 
+        boolean success = true;
```
trustin: I guess you can still start with `success = false`, since the ownership transfer after this try block?

anuraaga: Updated

PR #928, opened 2017-12-22
Update to Brave 4.13.1

anuraaga:
Added an explicit dependency on zipkin-java as the transitive dependency seems to have been removed. 

/cc @adriancole


imasahiro:
Thanks! @anuraaga

PR #929, opened 2017-12-22
Port the recent upstream changes of the forked code

trustin:
- Add the information about the revision of the original code and what
  modifications were made, so that it's easier to know what changes
  needs to be ported
- Remove the classes that do not need a fork anymore:
  - FlushConsolidationHandler
  - HeaderDateTimeFormat
- Fixes #906

PR #931, opened 2017-12-29
Fix incorrect cookie split

### LINKED ISSUE #930: ArmeriaHttpUtil.toArmeria breakes cookie into linked list entry which will lead only the first cookie entry can be obtained
Netty's `DefaultHeaders` will only get the value of first entry after head node:
https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java#L117

But in armeria's header conversion, value of cookie header is broken into linked list entries.
https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java#L329

This will lead armera's cookie can only obtain the first cookie entry's value.

Example:
```
cookie: a=1; b=2;

But only a=1 will be obtained from HttpRequest#headers().get("cookie").
```

trustin:
Motivation:

The pull request #929 introduced the violation of RFC 7540 section
8.1.2.5:

> If there are multiple Cookie header fields after decompression, these
> MUST be concatenated into a single octet string ..

Modifications:

- Merge and split cookies in a correct and robust way using StringJoiner
  and Splitter
- Inline Http2ToHttpHeaderTranslator which is used only in a single place

Result:

- Fixes #930

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
@@ -44,8 +44,10 @@ import java.net.URISyntaxException;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map.Entry;
+import java.util.StringJoiner;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Splitter;
 
 import com.linecorp.armeria.common.DefaultHttpHeaders;
 import com.linecorp.armeria.common.HttpData;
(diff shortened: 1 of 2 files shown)


trustin:
@southernkasaist Please review as well.

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -319,34 +339,22 @@ public static void toArmeria(io.netty.handler.codec.http.HttpHeaders inHeaders,
...
-
-                        if (start < value.length()) {
-                            out.add(HttpHeaderNames.COOKIE, value.substring(start));
-                        }
-                    } else {
-                        out.add(HttpHeaderNames.COOKIE, value);
+                    if (cookieJoiner == null) {
+                        cookieJoiner = new StringJoiner(COOKIE_SEPARATOR);
                     }
+                    COOKIE_SPLITTER.split(value).forEach(cookieJoiner::add);
```
anuraaga: Shouldn't this conversion be in `toAmeria(Http2Headers` and not here? AFAIU, the spec is only related to HTTP/2 while a similar conversion is not needed for HTTP/1.

core/src/test/java/com/linecorp/armeria/internal/ArmeriaHttpUtilTest.java
```diff
@@ -48,7 +53,38 @@ public void testConcatPaths() throws Exception {
...
+        in.add(HttpHeaderNames.COOKIE, "i=j");
+        in.add(HttpHeaderNames.COOKIE, "k=l;");
+
+        final Http2Headers out = toNettyHttp2(in);
+        assertThat(out.getAll(HttpHeaderNames.COOKIE))
+                .containsExactly("a=b", "c=d", "e=f", "g=h", "i=j", "k=l");
+    }
+
+    @Test
+    public void inboundCookiesMustBeMerged() {
```
anuraaga: Neet test for http2 input as well.

Also I think you take special care to handle Cookie, Cookie, Content-Type, Cookie etc (mixed up headers) would be good to add that to the tests

trustin: Indeed. Fixed as suggested.

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
trustin: Right. I added it just to make it more robust so it handles a poorly written peer. Should we just accept only the first one?

trustin: .. and, yes, the conversion should also be in `toArmeria(Http2Headers)`. Added there as well.

anuraaga: As a user I have some expectation that mutiple cookie headers would not be handled specially for HTTP/1 but think either way works. I'd just follow whatever tomcat etc does not sure what that is though :)

southernkasaist:
(awthanks)

PR #927, opened 2017-12-22
Update more dependencies

trustin:
- shadow plugin 2.0.1 -> 2.0.2
- snakeyaml 1.18 -> 1.19
- protoc 3.4.0 -> 3.5.1
- tomcat 8.0.47 -> 8.0.48

PR #932, opened 2017-12-29
Upgrade Caffeine to 2.6.1

kojilin:


trustin:
@imasahiro Gentle ping

trustin:
Thanks!

ben-manes:
Not sure if it helps, but I wrote the first [gradle plugin](https://github.com/ben-manes/gradle-versions-plugin) to automatically discover updates (when migrating from Maven and missing its feature). The plugin just leverages Gradle's own dependency management capabilities rather than do any heavy lifting itself.

I personally prefer having a [dependencies.gradle](https://github.com/ben-manes/caffeine/blob/master/gradle/dependencies.gradle) file vs your yaml, but that's stylistic.

anuraaga:
Yeah your plugin came up once :) https://github.com/line/armeria/pull/927

Think @trustin is looking into it and / or nebula plugin.

ben-manes:
nebula looks good! Always been too lazy to try it since I had my own method for a few years =)

trustin:
Tried to apply the Nebula plugins but it had some issues:

- The generated global lock file contained many duplicate entries, which are hard to read.
- Their publish setup does not work when a project publishes more than one artifact.

The motivation of adopting Nebula plugins was to remove the boilerplates related with dependency management and publication in `build.gradle` but none of them was solved as expected. 😭

ben-manes:
In the `dependencies.gradle` style, you write the usages like.,

```gradle
compile libraries.caffeine
testCompile testLibraries.testng
```

so the coordinate is fully encapsulated. That's nice for readability and if you need to bundle a few artifacts together into one, e.g. Jackson and its extensions.

anuraaga:
I didn't look at their publication stuff but I wonder if the experimental lock plugin would fit better (it seems much less intrusive to boot)

https://github.com/nebula-plugins/lock-experimental/blob/master/README.md

PR #934, opened 2018-01-02
Catch the Exception while converting RequestObject using JacksonRequestConverterFunction

minwoox:
Motivation:
If the client request with invalid JSON, `JacksonRequestConvertFunction` currently, throw
a converting excpeiton and there's no place to catch the exception. Therefore, the client
get `500` error which is not right.

Modification:
- Catch the exception and throw new `IllegalArgumentException` so that `DefaultExceptinHandler` handles it

Result:
- Respond with proper status to clients

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
@@ -22,6 +22,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 
@@ -66,8 +67,13 @@ public class JacksonRequestConverterFunction implements RequestConverterFunction
                                     contentType.subtype().endsWith("+json"))) {
             final ObjectReader reader = readers.computeIfAbsent(expectedResultType, mapper::readerFor);
             if (reader != null) {
-                return reader.readValue(request.content().toString(
-                        contentType.charset().orElse(StandardCharsets.UTF_8)));
+                final String content = request.content().toString(
+                        contentType.charset().orElse(StandardCharsets.UTF_8));
+                try {
+                    return reader.readValue(content);
+                } catch (JsonProcessingException e) {
+                    throw new IllegalArgumentException("failed to parse a JSON document: " + e, e);
+                }
             }
         }
         return RequestConverterFunction.fallthrough();
--- a/core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceRequestConverterTest.java
+++ b/core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceRequestConverterTest.java
@@ -28,6 +28,7 @@ import org.junit.Test;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import com.linecorp.armeria.client.HttpClient;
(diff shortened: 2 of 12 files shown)


core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceRequestConverterTest.java
```diff
@@ -113,6 +114,12 @@ public String defaultText(@RequestObject String obj1) {
             assertThat(obj1).isNotNull();
             return obj1;
         }
+
+        @Post("/default/invalid")
+        public String invalidJson(@RequestObject JsonNode node) {
+            // should never reach hear because we are sending invalid JSON
```
trustin: - hear -> here
- should -> Should

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -66,8 +67,16 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
...
-                        contentType.charset().orElse(StandardCharsets.UTF_8)));
+                try {
+                    return reader.readValue(request.content().toString(
+                            contentType.charset().orElse(StandardCharsets.UTF_8)));
+                } catch (JsonProcessingException e) {
+                    // Do not create the Exception with cause and throw IllegalArgumentException
+                    // so that DefaultExceptionHandler handles this
+                    throw new IllegalArgumentException(
+                            e.getClass().getSimpleName() + " is raised while converting " +
+                            expectedResultType.getSimpleName());
```
trustin: `"failed to parse a JSON document: " + e, e` ?

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -66,8 +67,16 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
                                     contentType.subtype().endsWith("+json"))) {
             final ObjectReader reader = readers.computeIfAbsent(expectedResultType, mapper::readerFor);
             if (reader != null) {
-                return reader.readValue(request.content().toString(
-                        contentType.charset().orElse(StandardCharsets.UTF_8)));
+                try {
+                    return reader.readValue(request.content().toString(
+                            contentType.charset().orElse(StandardCharsets.UTF_8)));
+                } catch (JsonProcessingException e) {
+                    // Do not create the Exception with cause and throw IllegalArgumentException
```
anuraaga: Out of curiosity, why not add the cause? It might provide useful information when using `LoggingService`.

minwoox: The reason is that If I add the cause, the `DefaultExceptionHandler` cannot handle this exception because the `handleException(ctx, req, rootCause)` is called with the `rootCause`.
However, I talk to @trustin and we have reached the conclusion that we change the part that passing the `rootCause`.
I will update PR and please review again after that. :-)

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -66,8 +67,13 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
...
             final ObjectReader reader = readers.computeIfAbsent(expectedResultType, mapper::readerFor);
             if (reader != null) {
-                return reader.readValue(request.content().toString(
-                        contentType.charset().orElse(StandardCharsets.UTF_8)));
+                final String content = request.content().toString(
+                        contentType.charset().orElse(StandardCharsets.UTF_8));
+                try {
+                    return reader.readValue(content);
+                } catch (JsonProcessingException e) {
+                    throw new IllegalArgumentException("failed to parse a JSON document: " + content, e);
```
trustin: I'm somewhat concerned about stringifying `content` because it can be very long.

anuraaga: May want to cap the length of content, or just hope the exception cause has enough context.

minwoox: Alright let me just remove the `content`. Thank you~!!

minwoox: Alright let me just remove the `content`. Thank you~!!

trustin:
Thanks, @minwoox !

PR #937, opened 2018-01-02
Add newDerivedContext(Request) to RequestContext

minwoox:
So, a user can derive a context with the specified `Request`.
Also, I changed to store the `request` field as `Request` instead of `Object`
inside `NonWrappingRequestContext`.


core/src/main/java/com/linecorp/armeria/common/RequestContextWrapper.java
```diff
@@ -97,7 +97,7 @@ public String query() {
     }
 
     @Override
-    public <T> T request() {
+    public <E extends Request> E request() {
```
trustin: E -> T ?

trustin:
Thanks, @minwoox !

PR #939, opened 2018-01-03
Change to apply service-level annotation after method-level annotation

minwoox:
Motivation:
Currently, if there are method-level `RequestConverter`s, `ResponseConveter`s or
`ExceptionHandler`s in `AnnotatedHttpService`, the class-level `Annotation`s are not applied.
We want to make sure all of the `Annotation`s are applied, so that the order will be
parameter-level -> method-level -> class-level -> service-level.

Modification:
- Add class-level `Annotation`s no matter what, to the `annotationValues`

Result:
- A user does not surprise


trustin:
Thanks, @minwoox !

PR #944, opened 2018-01-04
Update documentation

trustin:
- README.md
  - Update the project summary
  - Wrap at 112 column
- 'Calling a gRPC service'
  - Do not use the deprecated Futures.addCallback()
  - Make all examples wait until the call is finished so that the JVM
    does not exit too early, because the default event loop threads are
    daemon.
  - responseStream -> replyStream
- 'Running a gRPC service'
  - Add a note about how to configure the build
- 'Retrofit integration'
  - Remove the benefit item that's not true anymore
- Update the list of shaded dependencies

### DIFF
--- a/site/src/sphinx/setup-common.rst
+++ b/site/src/sphinx/setup-common.rst
@@ -1,6 +1,8 @@
+.. _`Caffeine`: https://github.com/ben-manes/caffeine
 .. _`completable-futures`: https://github.com/spotify/completable-futures
 .. _`fastutil`: http://fastutil.di.unimi.it/
 .. _`Guava`: https://github.com/google/guava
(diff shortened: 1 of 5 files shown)


site/src/sphinx/setup-common.rst
```diff
@@ -1,16 +1,20 @@
+.. _`Caffeine`: https://github.com/ben-manes/caffeine
 .. _`completable-futures`: https://github.com/spotify/completable-futures
 .. _`fastutil`: http://fastutil.di.unimi.it/
 .. _`Guava`: https://github.com/google/guava
+.. _`JCTools`: http://jctools.github.io/JCTools/
```
minwoox: https?

trustin: Fixed :-)

Issue #935, opened 2018-01-02
Support enum as a type of @Param annotated parameter


hyangtack:

AmosDoan:
Can I try this?

hyangtack:
No problem! :-)

trustin:
Fixed via #949

PR #949, opened 2018-01-06
Support enum as a type of @Param annotated parameter(#935)

AmosDoan:
Motivation:

Users may want to use enumeration type for requesting parameter

Modifications:

- Add converting logic from the enumeration type to string in ‘AnnotatedHttpServiceMethod'
- Add unit test for supporting enumeration type

Results:

- Users are now able to use requesting parameter

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
@@ -17,6 +17,7 @@
 package com.linecorp.armeria.server;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.linecorp.armeria.common.HttpParameters.EMPTY_PARAMETERS;
 import static com.linecorp.armeria.internal.DefaultValues.getSpecifiedValue;
@@ -24,12 +25,14 @@ import static java.util.Objects.requireNonNull;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
 
 import javax.annotation.Nullable;
 
@@ -37,6 +40,7 @@ import org.reactivestreams.Subscriber;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Ascii;
 import com.google.common.collect.ImmutableList;
 
 import com.linecorp.armeria.common.AggregatedHttpMessage;
@@ -346,7 +350,7 @@ final class AnnotatedHttpServiceMethod {
                 case PATH_PARAM:
                     value = ctx.pathParam(entry.name());
                     assert value != null;
-                    values[i] = stringToType(value, entry.type());
+                    values[i] = convertParameter(value, entry);
                     break;
                 case PARAM:
                     if (httpParameters == null) {
@@ -484,7 +488,12 @@ final class AnnotatedHttpServiceMethod {
     }
 
     private static Object convertParameter(String value, Parameter entry) {
-        final Object converted = value != null ? stringToType(value, entry.type()) : null;
+        Object converted = null;
+        if (value != null) {
+            converted = entry.isEnum() ? entry.getEnumElement(value)
+                                       : stringToType(value, entry.type());
+        }
+
         return entry.isOptionalWrapped() ? Optional.ofNullable(converted) : converted;
     }
 
@@ -561,6 +570,9 @@ final class AnnotatedHttpServiceMethod {
         if (clazz == Double.TYPE || clazz == Double.class) {
             return Double.TYPE;
         }
+        if (clazz.isEnum()) {
+            return clazz;
+        }
         if (clazz == String.class) {
             return String.class;
         }
@@ -626,6 +638,8 @@ final class AnnotatedHttpServiceMethod {
         private final String name;
         private final String defaultValue;
         private final RequestConverterFunction requestConverterFunction;
+        private final boolean isCaseSensitiveEnum;
+        private final Map<String, ? extends Enum<?>> enumMap;
 
         Parameter(ParameterType parameterType,
                   boolean isRequired, boolean isOptionalWrapped, Class<?> type,
@@ -638,6 +652,39 @@ final class AnnotatedHttpServiceMethod {
             this.name = name;
             this.defaultValue = defaultValue;
             this.requestConverterFunction = requestConverterFunction;
+
+            if (type.isEnum()) {
+                final Set<? extends Enum<?>> enumInstances = EnumSet.allOf(type.asSubclass(Enum.class));
+                Map<String, ? extends Enum<?>> lowerCaseEnumMap = enumInstances.stream().collect(
+                        toImmutableMap(e -> Ascii.toLowerCase(e.name()), Function.identity(), (e1, e2) -> e1));
+                if (enumInstances.size() != lowerCaseEnumMap.size()) {
+                    enumMap = enumInstances.stream().collect(toImmutableMap(Enum::name, Function.identity()));
+                    isCaseSensitiveEnum = true;
+                } else {
+                    enumMap = lowerCaseEnumMap;
+                    isCaseSensitiveEnum = false;
+                }
+            } else {
+                enumMap = null;
+                isCaseSensitiveEnum = false;
+            }
+        }
+
+        <T> T getEnumElement(String str) {
+            final Object o = enumMap.get(isCaseSensitiveEnum ? str : Ascii.toLowerCase(str));
+
+            if (o == null) {
+                throw new IllegalArgumentException("unknown enum value: " + str + " (expected: " +
+                                                   enumMap.values() + ')');
+            }
+
+            @SuppressWarnings("unchecked")
+            final T result = (T) o;
+            return result;
+        }
+
+        boolean isEnum() {
+            return type.isEnum();
         }
 
         ParameterType parameterType() {
--- a/core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
+++ b/core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
@@ -409,6 +409,22 @@ public class AnnotatedHttpServiceTest {
             return parameters.get("username") + "/" + parameters.get("password");
         }
 
+        @Get("/param/enum")
+        public String paramEnum(RequestContext ctx,
+                                @Param("username") String username,
+                                @Param("level") UserLevel level) {
+            validateContext(ctx);
+            return username + "/" + level;
+        }
+
+        @Get("/param/enum2")
+        public String paramEnum2(RequestContext ctx,
+                                @Param("type") UserType type,
+                                @Param("level") UserLevel level) {
+            validateContext(ctx);
+            return type + "/" + level;
+        }
+
         @Get
         @Path("/param/default1")
         public String paramDefault1(RequestContext ctx,
@@ -680,6 +696,17 @@ public class AnnotatedHttpServiceTest {
             testBody(hc, form("/7/map/post", null,
                               "username", "line4", "password", "armeria4"), "line4/armeria4");
 
+            testBody(hc, get("/7/param/enum?username=line5&level=LV1"), "line5/LV1");
+            // Case insensitive test for enum element
+            testBody(hc, get("/7/param/enum?username=line5&level=lv1"), "line5/LV1");
+            testBody(hc, get("/7/param/enum?username=line6&level=Lv2"), "line6/LV2");
+            testStatusCode(hc, get("/7/param/enum?username=line6&level=TEST3"), 400);
+
+            // Case sensitive test enum
+            testBody(hc, get("/7/param/enum2?type=normal&level=LV1"), "normal/LV1");
+            testBody(hc, get("/7/param/enum2?type=NORMAL&level=LV1"), "NORMAL/LV1");
+            testStatusCode(hc, get("/7/param/enum2?type=MINOOX&level=LV1"), 400);
+
             testBody(hc, get("/7/param/default1"), "hello/world/(null)");
             testBody(hc, get("/7/param/default1?extra=people&number=1"), "hello/world/people/1");
 
@@ -790,6 +817,16 @@ public class AnnotatedHttpServiceTest {
 
             testStatusCode(hc, get("/11/headerWithoutValue"), 400);
         }
+    }
+
+    private enum UserLevel {
+        LV1,
+        LV2
+    }
+
+    private enum UserType {
+        normal,
+        NORMAL
     }
 
     static void testBodyAndContentType(CloseableHttpClient hc, HttpRequestBase req,


core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -521,6 +521,8 @@ private static String httpHeaderValue(Parameter entry, HttpRequest req) {
                 return (T) Float.valueOf(str);
             } else if (clazz == Double.TYPE) {
                 return (T) Double.valueOf(str);
+            } else if (clazz.isEnum()) {
+                return (T) Enum.valueOf(clazz.asSubclass(Enum.class), str);
```
trustin: Could you replace the `catch (NumberFormatException e)` below with `catch (IllegalArgumentException e)` so we do not wrap an IAE with an IAE?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -561,6 +563,9 @@ private static String httpHeaderValue(Parameter entry, HttpRequest req) {
         if (clazz == Double.TYPE || clazz == Double.class) {
             return Double.TYPE;
         }
+        if (clazz.isEnum()) {
+            return clazz.asSubclass(Enum.class);
```
trustin: Do we need `asSubclass()` here? An enum class is always an `Enum` so there seems no need for it.

minwoox:
I'm wondering if we need to restrict the `enum` in the `pathParam` since this will allow it. What do you think @hyangtack?

hyangtack:
> I'm wondering if we need to restrict the enum in the pathParam since this will allow it. What do you think @hyangtack?

I think there's no problem to convert `path variable` into an enum. Also we use `@Param` annotation on both of `path variable` and `parameter`, so I think it's more general to support both of them. :-)

hyangtack:
> I'm curious if it would be useful to allow the enum values to be case-insensitive. We could keep the Map<String, Enum<?>> whose key is an upper-cased enum name. The reason why I'm asking this is most RESTful APIs I have seen in the wild do not seem to use upper-cased parameters.

@trustin 
I'm wondering that a user may think armeria converts a string to enum in a case-insensitive way due to this optimization. How about leaving it as it is? If a user want to make their URL lowercase, they may make their enum lowercase or they may use `@RequestObject` annotation with a converter. What do you think?

trustin:
> I'm wondering that a user may think armeria converts a string to enum in a case-insensitive way due to this optimization. How about leaving it as it is? If a user wants to make their URL lowercase, they may make their enum lowercase or they may use `@RequestObject` annotation with a converter. What do you think?

I was actually suggesting to make string-to-enum conversion case-insensitive if it is sure that there will be no ambiguity for the target enum type, for user convenience.

I think it will be useful to make this behavior default because most HTTP APIs use lower-cased constants unlike Java's naming convention. If a user wants case-sensitivity, he or she could specify an alternative converter, which is similar to your suggestion.

hyangtack:
@trustin Ok! I will care about this behavior when I make a document about annotated http service. :-)

anuraaga:
FWIW, spring was case-sensitive IIRC - it lead to many gross(?) enums with lowercase values but it might make sense to follow their example. No real opinion either way though just FYI

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -499,6 +499,17 @@ private static String httpHeaderValue(Parameter entry, HttpRequest req) {
         return entryDefaultValue(entry);
     }
 
+    @SuppressWarnings("unchecked")
```
AmosDoan: @trustin, @minwoox, @hyangtack, @anuraaga Thanks you very much for many comments :) BTW, I'm curious about that if the user wants to use case-sensitive enum parameter, then should we need to keep the map for the enumeration type? (Map<String, Enum<?>>) I considered about adding the map inside the Parameter class, but how about this implementation just checking when the convert to string? (maybe this is more simpler?)

AmosDoan: Hmm.. I also thought the time complexity would be O(n), the performance degradation is so high in the case of a lot of the enum element Let me change to the map as @trustin suggested.

trustin: @AmosDoan Yes, please. We may want to introduce some conversion option or value converters to support case-sensitive enums, but for now, let's keep it case-insensitive.

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
```diff
@@ -680,6 +688,12 @@ public void testParam() throws Exception {
             testBody(hc, form("/7/map/post", null,
                               "username", "line4", "password", "armeria4"), "line4/armeria4");
 
+            testBody(hc, get("/7/param/enum?username=line5&level=LV1"), "line5/LV1");
+            // Case insensitive test for enum element
+            testBody(hc, get("/7/param/enum?username=line5&level=lv1"), "line5/LV1");
+            testBody(hc, get("/7/param/enum?username=line6&level=LV2"), "line6/LV2");
```
trustin: How about `Lv2`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -499,6 +511,17 @@ private static String httpHeaderValue(Parameter entry, HttpRequest req) {
...
+    private static <T> T stringToEnum(String str, Class<T> clazz, Parameter parameter) {
+        final Map<String, Enum<?>> map = parameter.getUpperCaseEnum();
+        if (map.containsKey(str.toUpperCase())) {
```
minwoox: `Ascii.toUpperCase()?` (more efficient)

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -524,7 +547,7 @@ private static String httpHeaderValue(Parameter entry, HttpRequest req) {
             } else if (clazz == String.class) {
                 return (T) str;
             }
-        } catch (NumberFormatException e) {
+        } catch (IllegalArgumentException e) {
```
minwoox: Since you take out the enum check here, you can change it back.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -626,6 +652,8 @@ static Parameter ofRequestObject(Class<?> type,
         private final String name;
         private final String defaultValue;
         private final RequestConverterFunction requestConverterFunction;
+        private final Map<String, Enum<?>> upperCaseEnum;
```
minwoox: How about `Map<String, ? extends Enum<?>>` ?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -638,6 +666,24 @@ static Parameter ofRequestObject(Class<?> type,
...
+                isEnum = false;
+            }
+        }
+
+        Map<String, Enum<?>> getUpperCaseEnum() {
+            return upperCaseEnum;
+        }
+
+        boolean isEnum() {
+            return isEnum;
```
minwoox: How about removing the member and just returning `type.isEnum()`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -638,6 +666,24 @@ static Parameter ofRequestObject(Class<?> type,
...
+                upperCaseEnum = enumInstances.stream().collect(toImmutableMap(e -> e.name().toUpperCase(),
+                                                                              Function.identity()));
+                isEnum = true;
+            } else {
+                upperCaseEnum = null;
+                isEnum = false;
+            }
+        }
+
+        Map<String, Enum<?>> getUpperCaseEnum() {
```
minwoox: How about removing this method and instead:
```
<T extends Enum<T>> T get(String key) {
    return (T) upperCaseEnum.get(Ascii.toUpperCase(key));
}
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -499,6 +511,17 @@ private static String httpHeaderValue(Parameter entry, HttpRequest req) {
...
 
+    @SuppressWarnings("unchecked")
+    private static <T> T stringToEnum(String str, Class<T> clazz, Parameter parameter) {
+        final Map<String, Enum<?>> map = parameter.getUpperCaseEnum();
+        if (map.containsKey(str.toUpperCase())) {
+            return (T) map.get(str.toUpperCase());
+        } else {
+            throw new IllegalArgumentException(
+                    "Can't convert '" + str + "' to type '" + clazz.getSimpleName() + "'.");
+        }
```
minwoox: How about
```
        private static <T> T stringToEnum(String str, Class<T> clazz, Parameter entry) {
        final T result = entry.get(str);
        if (result == null) {
            throw new IllegalArgumentException(
                    "Can't convert '" + str + "' to type '" + clazz.getSimpleName() + "'.");
        }
        return result;
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -499,6 +511,17 @@ private static String httpHeaderValue(Parameter entry, HttpRequest req) {
         return entryDefaultValue(entry);
     }
 
+    @SuppressWarnings("unchecked")
+    private static <T> T stringToEnum(String str, Class<T> clazz, Parameter parameter) {
+        final Map<String, Enum<?>> map = parameter.getUpperCaseEnum();
+        if (map.containsKey(str.toUpperCase())) {
+            return (T) map.get(str.toUpperCase());
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -638,6 +666,24 @@ static Parameter ofRequestObject(Class<?> type,
             this.name = name;
             this.defaultValue = defaultValue;
             this.requestConverterFunction = requestConverterFunction;
+
+            if (type.isEnum()) {
+                final Set<? extends Enum<?>> enumInstances = EnumSet.allOf(type.asSubclass(Enum.class));
+                upperCaseEnum = enumInstances.stream().collect(toImmutableMap(e -> e.name().toUpperCase(),
```
minwoox: `Ascii.toUpperCase()`?

hyangtack:
I think we can compare enum value in a case-insensitive way only if there is no conflict. (No both of "Hello" and "HELLO" in the enum) So we need to check it when making a `upperCaseEnum` map.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -626,6 +653,8 @@ static Parameter ofRequestObject(Class<?> type,
         private final String name;
         private final String defaultValue;
         private final RequestConverterFunction requestConverterFunction;
+        private final Map<String, ? extends Enum<?>> upperCaseEnum;
+        private final boolean isCaseSensitive;
```
AmosDoan: @hyangtack  I added this flag for the case when the element of the enum has same spell but different case.

hyangtack: Could we use `upperCaseEnum` variable as a flag instead of `isCaseSensitive`? We may use `upperCaseEnum` only if it's not `null`. Also, it'd be better to make `upperCaseEnum` as `final`. :-)

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -638,6 +665,34 @@ static Parameter ofRequestObject(Class<?> type,
...
+                isCaseSensitiveEnum = false;
+            }
+        }
+
+        <T extends Enum<T>> T get(String str) {
+            if (isCaseSensitiveEnum) {
+                return Enum.valueOf(type.asSubclass(Enum.class), str);
+            }
+            @SuppressWarnings("unchecked")
+            final T result = (T) upperCaseEnum.get(Ascii.toUpperCase(str));
```
hyangtack: nit: If this method is called for the type which is not an enum, NPE may be thrown here. In that case, it'd be better to throw an exception like `IllegalStateException`. Also, how about changing the method name? I think `get` is too general.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -499,6 +512,15 @@ private static String httpHeaderValue(Parameter entry, HttpRequest req) {
         return entryDefaultValue(entry);
     }
 
+    private static <T> T stringToEnum(String str, Class<T> clazz, Parameter entry) {
+        final T result = (T)entry.get(str);
```
hyangtack: nit: Do we need `(T)` here?

AmosDoan: Checked & Updated Thanks

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -638,6 +652,43 @@ static Parameter ofRequestObject(Class<?> type,
             this.name = name;
             this.defaultValue = defaultValue;
             this.requestConverterFunction = requestConverterFunction;
+
+            if (type.isEnum()) {
+                final Set<? extends Enum<?>> enumInstances = EnumSet.allOf(type.asSubclass(Enum.class));
+                Map<String, ? extends Enum<?>> upperCaseEnumMap = enumInstances.stream().collect(
+                        toImmutableMap(e -> Ascii.toUpperCase(e.name()), Function.identity(), (e1, e2) -> e1));
+                if (enumInstances.size() != upperCaseEnumMap.size()) {
+                    upperCaseEnum = null;
```
trustin: - How about building a lower-case map given most HTTP API uses lower-case values?
- Instead of setting the map to null, how about building a map with case-sensitive enum values, so that `getEnumElement()` becomes simpler?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -638,6 +652,43 @@ static Parameter ofRequestObject(Class<?> type,
...
+                isCaseSensitiveEnum = false;
+            }
+        }
+
+        <T> T getEnumElement(String str) {
+            final Object o;
+            if (isCaseSensitiveEnum) {
+                o = Enum.valueOf(type.asSubclass(Enum.class), str);
+            } else {
+                o = upperCaseEnum.get(Ascii.toUpperCase(str));
```
trustin: With the comment above addressed, this could be simplified:

    o = enumMap.get(isCaseSensitiveEnum ? str : Ascii.toLowerCase(str));

Also, please note you did not handle the `IllegalArgumentException` raised by `Enum.valueOf()`, although `valueOf()` call should be gone after addressing the review comments.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -638,6 +652,43 @@ static Parameter ofRequestObject(Class<?> type,
...
+            }
+
+            if (o == null) {
+                throw new IllegalArgumentException("failed to convert an enum: " + str);
```
trustin: How about:

    "unknown enum value: " + str + " (expected: " + enumMap.values() + ')'

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks!!

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks!!

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
AmosDoan: Checked & Updated Thanks!!

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -638,6 +652,39 @@ static Parameter ofRequestObject(Class<?> type,
             this.name = name;
             this.defaultValue = defaultValue;
             this.requestConverterFunction = requestConverterFunction;
+
+            if (type.isEnum()) {
+                final Set<? extends Enum<?>> enumInstances = EnumSet.allOf(type.asSubclass(Enum.class));
+                Map<String, ? extends Enum<?>> upperCaseEnumMap = enumInstances.stream().collect(
```
trustin: `lowerCaseEnumMap`?

AmosDoan:
@trustin @minwoox I updated the PR. I squashed all the commits and arrange the commit message.

trustin:
Thanks for your patience, @AmosDoan!

PR #954, opened 2018-01-12
Add Resource file backed endpoint group

imasahiro:


trustin:
Please let us know if this PR is ready for review.

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
```diff
@@ -0,0 +1,95 @@
...
+            return newEndpoints;
+        } catch (IOException e) {
+            throw new IllegalArgumentException("failed to load: " + resourceName, e);
+        }
+    }
+
+    private final List<Endpoint> endpoints;
+
+    private ResourceFileEndpointGroup(List<Endpoint> endpoints) {
+        this.endpoints = endpoints;
```
hyangtack: nit: How about making this list immutable?

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public static ResourceFileEndpointGroup of(ClassLoader classLoader, String resourceName,
+                                               String endpointKeyPrefix, int defaultPort) {
+        return new ResourceFileEndpointGroup(loadEndpoints(classLoader, resourceName,
+                                                           endpointKeyPrefix, defaultPort));
+    }
+
+    private static List<Endpoint> loadEndpoints(ClassLoader classLoader, String resourceName,
+                                                String endpointKeyPrefix, int defaultPort) {
+        URL resourceUrl = classLoader.getResource(resourceName);
```
hyangtack: nit: `final`?

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
```diff
@@ -0,0 +1,95 @@
...
+     * @throws IllegalArgumentException if failed to load any hosts from the specified resource file
+     */
+    public static ResourceFileEndpointGroup of(ClassLoader classLoader, String resourceName,
+                                               String endpointKeyPrefix, int defaultPort) {
+        return new ResourceFileEndpointGroup(loadEndpoints(classLoader, resourceName,
+                                                           endpointKeyPrefix, defaultPort));
+    }
+
+    private static List<Endpoint> loadEndpoints(ClassLoader classLoader, String resourceName,
+                                                String endpointKeyPrefix, int defaultPort) {
```
hyangtack: How about checking arguments using `requireNonNull`?

imasahiro:
Ready for review now.

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
imasahiro: Changed `loadEndpoints` to return immutable list instead.

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Map.Entry;
+import java.util.Properties;
+
+import com.google.common.collect.ImmutableList;
+
+import com.linecorp.armeria.client.Endpoint;
+
+/**
+ * A resource file backed {@link EndpointGroup}. The list of {@link Endpoint}s are loaded from the resource file
+ * in a current class loader.
```
trustin: current class loader -> `given {@link ClassLoader}`

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        try (InputStream in = resourceUrl.openStream()) {
+            final List<Endpoint> newEndpoints = new ArrayList<>();
+            final Properties props = new Properties();
+            props.load(in);
+            for (Entry<Object, Object> e : props.entrySet()) {
+                final String key = (String) e.getKey();
+                final String value = (String) e.getValue();
+
+                if (key.startsWith(endpointKeyPrefix)) {
```
trustin: How about appending `.` to `endpointKeyPrefix` (if it does not end with a dot)?

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
```diff
@@ -0,0 +1,102 @@
...
+     * example.hosts.1=example2.com:36462
+     * example.hosts.2=example3.com:36462
+     * }</pre>
+     *
+     * @param resourceName the resource file name that the list of {@link Endpoint}s loaded from
+     * @throws IllegalArgumentException if failed to load any hosts from the specified resource file
+     */
+    public static ResourceFileEndpointGroup of(ClassLoader classLoader, String resourceName,
+                                               String endpointKeyPrefix, int defaultPort) {
+        checkArgument(defaultPort > 0, "defaultPort: %s (expected: > 0)", defaultPort);
```
minwoox: nit: How about adding the upper bound like we do in `DnsAddressEndpointGroup` or `Endpoint`?

imasahiro:
>  Perhaps we need to reload the endpoint list if the file is updated? Some resources may be a real file in a filesystem. You may just want to add a TODO and resolve it later though.


Leave todo to head of the class.

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
```diff
@@ -0,0 +1,107 @@
...
+     * example.hosts.1=example2.com:36462
+     * example.hosts.2=example3.com:36462
+     * }</pre>
+     *
+     * @param resourceName the resource file name that the list of {@link Endpoint}s loaded from
+     * @throws IllegalArgumentException if failed to load any hosts from the specified resource file
+     */
+    public static ResourceFileEndpointGroup of(ClassLoader classLoader, String resourceName,
+                                               String endpointKeyPrefix, int defaultPort) {
+        checkArgument(defaultPort >= 0 && defaultPort <= 65535,
```
minwoox: I think the port should be `> 0`. `DnsAddressEndpointGroup` should be changed as well. WDYT?

core/src/main/java/com/linecorp/armeria/client/endpoint/ResourceFileEndpointGroup.java
```diff
@@ -0,0 +1,107 @@
...
+     * example.hosts.2=example3.com:36462
+     * }</pre>
+     *
+     * @param resourceName the resource file name that the list of {@link Endpoint}s loaded from
+     * @throws IllegalArgumentException if failed to load any hosts from the specified resource file
+     */
+    public static ResourceFileEndpointGroup of(ClassLoader classLoader, String resourceName,
+                                               String endpointKeyPrefix, int defaultPort) {
+        checkArgument(defaultPort > 0 && defaultPort <= 65535,
+                      "defaultPort(%s) must be between 0 and 65535", defaultPort);
```
minwoox: nit: between 1 and ?

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -116,7 +116,7 @@ public static DnsAddressEndpointGroup of(String hostname, int defaultPort, Event
     @VisibleForTesting
     DnsAddressEndpointGroup(String hostname, int defaultPort, NameResolver<InetAddress> resolver,
                             EventLoop eventLoop, Duration queryInterval) {
-        checkArgument(defaultPort >= 0 && defaultPort <= 65535, "defaultPort must be between 0 and 65535");
+        checkArgument(defaultPort > 0 && defaultPort <= 65535, "defaultPort must be between 0 and 65535");
```
minwoox: nit: between 1 and ?

minwoox:
@imasahiro  I'm sorry that I confused you. The `defaultPort` can be `0`. If it is 0, the port is configured using `withDefaultPort()`. I'm sorry about that. Could you change it back to the original plz?

huydx:
@imasahiro By named it ResourceFileEndpointGroup, it seems that the current implement is for testing purpose only? AFAIK get file path from ClassLoader is not usable when you pack it as jar file.
So how about make the class more generic by:
- Name is as "StaticFileEndpointGroup"
- Add constructor which receive explicit file path in addition to constructor that receive ClassLoad.

HDYT @trustin @minwoox ?

minwoox:
> Name is as "StaticFileEndpointGroup"

Either is fine to me. :-)

> Add constructor which receive explicit file path in addition to constructor that receive ClassLoad.

I think it's a good idea to have that constructor :)

trustin:
@huydx wrote:

> By named it ResourceFileEndpointGroup, it seems that the current implement is for testing purpose only? AFAIK get file path from ClassLoader is not usable when you pack it as jar file.

`ResourceFileEndpointGroup` is reading the content of the file using `ClassLoader.getResource()` so it can read the file in a JAR. I think it can also be used for production depending on use case. i.e. not solely for testing purpose.

> Add constructor which receive explicit file path in addition to constructor that receive ClassLoader.

Sounds like a good idea. How about also adding a constructor that accepts a `Properties` and renaming the class to `PropertiesEndpointGroup, @imasahiro?

imasahiro:
Renamed to PropertiesEndpointGroup. But want to treat *FileEndpointGroup on future PR because better to care about dynamic changes, symbolic link, etc.

trustin:
> want to treat *FileEndpointGroup on future PR because better to care about dynamic changes, symbolic link, etc.

OK!

PR #955, opened 2018-01-12
Fix logback setting for TrafficLoggingHandler

kojilin:
https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/TrafficLoggingHandler.java#L34


trustin:
Thanks, @kojilin !

PR #957, opened 2018-01-15
Micrometer 1.0.0-rc.7 / Increase metrics integration test timeout

trustin:
- Update Micrometer to 1.0.0-rc.7
- Increase metrics integration test timeout to 30 seconds so that the
  build doesn't fail on slow CI machines


trustin:
Thanks, reviewers.

PR #959, opened 2018-01-17
Prioritize rpc response error over end response error.

anuraaga:
I think I had originally written it this way, but thought I could simplify the logic. I realized though, that the response cause pretty much always ends up being `AbortedStreamException` since we call `req.abort()` when finishing a request in `HttpServerHandler`. I guess if an RPC failure has been specified, it will always be more relevant than an `endResponse` exception, so it makes sense to prioritize it, but let me know what you think.


anuraaga:
Oops, almost reset this branch since I thought it was done - can we merge? :)

trustin:
Sorry that I'm late!

PR #960, opened 2018-01-17
[TRIVIAL] ZooKeeperRegistrationTest#{startServers, stopServers} does not throw Exception anymore

dongjinleekr:
Found while I was working on `ZooKeeperUpdatingListener`.


dongjinleekr:
... Found one more.

Issue #961, opened 2018-01-18
Make default MetricIdPrefixFunction  aware of response as well


huydx:
Original request is to adding status code tag to metrics. To do that we need to make MetricIdPrefixFunction aware of response as well.
Current implement at MetricIdPrefixFunction#ofDefault is called only once at `RequestMetricSupport#onRequestStart`, therefore at the creation moment, there is no information about response yet.

PR #962, opened 2018-01-18
Split ActiveRequestMetrics from RequestMetric to support better metric life cycle 

huydx:
To solve https://github.com/line/armeria/issues/961

not test, or format code yet, just want to make sure the approach make sense.

PTAL @trustin @minwoox

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
+++ b/grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
@@ -175,10 +175,13 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
 
     private boolean compressedFlag;
     private boolean endOfStream;
+
     @Nullable
-    private CompositeByteBuf nextFrame;
+    private ByteBuf firstFrame;
+
     @Nullable
     private CompositeByteBuf unprocessed;
+
     private long pendingDeliveries;
     private boolean deliveryStalled = true;
     private boolean inDelivery;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -130,13 +134,17 @@ private RequestMetricSupport() {}
     private interface RequestMetrics {
         AtomicInteger active();
 
-        Counter success();
-
         Counter failure();
```
huydx: we will have separate request_failure and response_failure, which may be a bit hard to understand from user view, but from system view, it's apparently different thing. HDYT about this?

huydx:
gentle ping @trustin

trustin:
How about separating the number of active requests from others instead? Then we can keep both request/response metrics under the same meter ID.

trustin:
Sorry I'm late!

> However I found that this approach will make we miss some request metrics if the request does not make it to COMPLETE stage, so split into RequestMetrics , ResponseMetrics and ActiveRequestMetrics seems make it more clear what need to be listened, and when. HDYT?

All requests are going to enter the COMPLETE stage eventually, so it is OK to keep request and response metrics in a single class.

The reason I asked for splitting active request count to a separate class is because the meter ID of the active request count must be determined *before* `RequestLogAvailability` enters `COMPLETE`. This means:

- We have to use two different `MeterIdPrefixFunction`s:

  - One for active request count which must be called when a request starts
  - The other for all other metrics which must be called when `RequestLogAvailability.COMPLETE` is achieved so that more sophiscated meter ID can be generated.

> Btw, I have a newbie question: I couldn't make RequestMetricSupportTest to passed, because
calling ctx.logBuilder().endResponse() only seems do not trigger RequestLogAvailability.COMPLETE, which make the test failed.

Both `endRequest()` and `endResponse()` have to be called to trigger `COMPLETE` state.

huydx:
Realize that I'm using master to PR, so will make another PR for feature branch

PR #964, opened 2018-01-22
Write serialized gRPC message as ByteBufHttpData to avoid unnecessary…

anuraaga:
… copy.

Request side was already skipping the copy but not response side.

- Cleans up `ArmeriaClientCall` to reflect that `writePayload` takes ownership.
- Write to a normal buffer instead of composite buffer when framing since it is what's written out to the wire. During profiling / debugging exercises, I found that the buffer written to wire has to be consolidated so there's no reason to use a composite buffer at the final stage.


trustin:
Thanks, @anuraaga !

PR #968, opened 2018-01-23
Fix zipkin can't trace problem when there is multiple armeria http re…

kojilin:
…quests in service method.

If we trigger armeria client in server, it will reuse the same event loop for Client. Due to using thread local to store the span scope, if we trigger multiple client at same time, the first client's context push/pop will remove the span scope from thread local.

Separate client/server thread local may be a good solution(But I'm not confident)


zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -98,7 +100,14 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
         final String method = ctx.method().name();
         span.kind(Kind.CLIENT).name(method).start();
 
-        SpanContextUtil.setupContext(ctx, span, tracer);
+        ctx.onEnter(unused -> SPAN_IN_THREAD.set(tracer.withSpanInScope(span)));
```
anuraaga: Looks like the logic can still be shared if the threadlocal is passed in.

trustin:
@imasahiro Comments?

trustin:
Thanks, @kojilin !

PR #973, opened 2018-01-25
Closes the request stream message when the HTTP/2 stream is removed f…

### LINKED ISSUE #971: Get client disconnect notifications in bidirectional async gRPC
We don't get any notification when the client closes the connection when using asynchronous bidirectional stream calls.

https://github.com/SourceForgery/tachikoma/blob/ddfb03ba5d25f3de61e8fe1e1b2cb9be2aba0eb4/tachikoma-grpc/src/main/kotlin/com/sourceforgery/tachikoma/mta/MTAEmailQueueService.kt#L31

We get neither `onComplete()`, nor `onError()` in the ~ResponseObserver~`StreamObserver` when the client closes the socket.

Just to clarify the code, the `getEmails()` method will return immediately, but will hook up a listener to the RabbitMQ which will then deliver messages via `responseObserver.onNext(response)`. The RabbitMQ will stop delivering messages when the future is finished (which will only happen in response to exceptions or close).

The lifetime of this call is counted in months, as it's more of a subscription than anything else.

anuraaga:
…rom the connection.

In the case of an abrupt HTTP/2 connection close, HTTP/2 streams will be removed from the connection before the client has sent an END_STREAM frame. Without handling this and closing the request stream message, there is no way for server logic to handle this error for e.g., cleanup actions, especially if they disable response timeout because the stream is intended to have infinite length.

Because clients are expected to send END_STREAM in normal circumstances, this case is handled as a ClosedSessionException.

Logic change originally proposed by @tobias-.

Fixes #971

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Http2RequestDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Http2RequestDecoder.java
@@ -23,6 +23,7 @@ import static io.netty.handler.codec.http2.Http2Exception.streamError;
 
 import java.nio.charset.StandardCharsets;
 
+import com.linecorp.armeria.common.ClosedSessionException;
 import com.linecorp.armeria.common.ContentTooLargeException;
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpMethod;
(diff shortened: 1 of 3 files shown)


trustin:
Thanks, @anuraaga !

PR #974, opened 2018-01-25
Remove STATIC_INIT token from EmptyLineSeparator checkstyle rule

### LINKED ISSUE #700: Discuss STATIC_INIT checkstyle rule
Just happened to run into this checkstyle rule and thought a small discussion would be good. It requires an empty line around a static initialization block. I feel like this can be left up to the programmer, since many static initialization blocks are directly tied to a field - removing the empty line makes this association clearer to me.

Rule: 'STATIC_INIT' should be separated from previous statement. [EmptyLineSeparator]
Example code:
```java
class Foo {
    private static final List<InetAddress> ADDRESSES;

    static {
        try {
            ADDRESSES = ImmutableList.of(
                    InetAddress.getByName("1.2.3.4"),
                    InetAddress.getByName("2.3.4.5"));
        } catch (UnknownHostException e) {
            throw new Error(e);
        }
    }
}
```

trustin:
Motivation:

Many static initialization blocks are directly tied to a field. Removing
the empty line between the field declaration and its static
initialization block makes the association clearer:

    private static final String X;
    static {
        X = ...;
    }

Modifications:

- Remove STATIC_INIT token from the EmptyLineSeparator checkstyle rule

Result:

- Fixes #700

### DIFF
--- a/settings/checkstyle/checkstyle.xml
+++ b/settings/checkstyle/checkstyle.xml
@@ -233,7 +233,7 @@
     <module name="EmptyLineSeparator">
       <property name="allowNoEmptyLineBetweenFields" value="true"/>
       <property name="tokens"
-                value="IMPORT, CLASS_DEF, INTERFACE_DEF, ENUM_DEF, STATIC_INIT, INSTANCE_INIT, METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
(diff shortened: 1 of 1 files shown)


trustin:
We already agreed to make this change at #700. Will merge once build passes.

trustin:
Thanks again for initiating the discussion, @anuraaga. Very helpful. :+1:

trustin:
Thanks for your patience, @anuraaga. I hope contributing to Armeria gives you less hassle :-)

PR #975, opened 2018-01-25
Keep track of created channels to close them if the connection pool i…

### LINKED ISSUE #972: ClientFactory.close doesn't actually close in-progress connections
I was trying to simulate a scenario where a client opens a connection to the server (e.g., starts a gRPC stream) but then forcibly closes the socket. I would expect `ClientFactory.close` to close the socket, but it doesn't actually seem to. It calls `DefaultKeyedChannelPool.close`, but that only keeps track of available connections, not offered connections, so the actual stream doesn't seem to be affected. Should the pool also keep track of offered connections to allow closing them when needed?

anuraaga:
…s closed.

Also changes the behavior to wait for channel close, as it seems like it's intended to block (similar to how the factory blocks on worker group shutdown when closed).

Fixes #972

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
+++ b/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
@@ -17,10 +17,12 @@ package com.linecorp.armeria.client.pool;
 
 import static java.util.Objects.requireNonNull;
 
+import java.util.ArrayList;
 import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -34,6 +36,7 @@ import com.linecorp.armeria.common.ClosedSessionException;
 import com.linecorp.armeria.common.util.Exceptions;
 
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelFuture;
 import io.netty.channel.EventLoop;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.Promise;
@@ -303,20 +306,29 @@ public class DefaultKeyedChannelPool<K> implements KeyedChannelPool<K> {
     @Override
     public void close() {
         if (eventLoop.inEventLoop()) {
-            doClose();
+            // While we'd prefer to block until the pool is actually closed, we cannot block for the channels to
+            // close if it was called from the event loop or we would deadlock. In practice, it's rare to call
+            // close from an event loop thread, and not a main thread.
+            doClose(false);
         } else {
-            eventLoop.execute(this::doClose);
+            eventLoop.submit(() -> doClose(true)).syncUninterruptibly();
         }
     }
 
-    private void doClose() {
+    private void doClose(boolean blocking) {
         closed = true;
+
+        List<ChannelFuture> closeFutures = new ArrayList<>();
 
         for (Channel ch : allChannels) {
             if (ch.isOpen()) {
-                ch.close();
+                closeFutures.add(ch.close());
             }
         }
         allChannels.clear();
+
+        if (blocking) {
+            closeFutures.forEach(ChannelFuture::syncUninterruptibly);
+        }
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
+++ b/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
@@ -292,6 +292,7 @@ public class DefaultKeyedChannelPool<K> implements KeyedChannelPool<K> {
                 ch.close().syncUninterruptibly();
             }
         }
+        createdChannels.clear();
         for (Iterator<Deque<Channel>> i = pool.values().iterator(); i.hasNext();) {
             final Deque<Channel> queue = i.next();
             i.remove();
--- a/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
+++ b/core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
@@ -19,8 +19,10 @@ import static java.util.Objects.requireNonNull;
 
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.function.Function;
@@ -58,6 +60,8 @@ public class DefaultKeyedChannelPool<K> implements KeyedChannelPool<K> {
     private final Map<K, Deque<Channel>> pool;
     private final Map<K, Future<Channel>> pendingConnections;
 
+    private final Set<Channel> createdChannels;
+
     /**
      * Creates a new instance.
      */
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
```diff
@@ -74,6 +78,7 @@ public DefaultKeyedChannelPool(EventLoop eventLoop, Function<K, Future<Channel>>
 
         pool = new ConcurrentHashMap<>();
         pendingConnections = new HashMap<>();
+        createdChannels = new HashSet<>();
```
trustin: How about `allChannels`?

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
```diff
@@ -280,6 +287,12 @@ protected boolean offerChannel(K key, Channel channel) {
 
     @Override
     public void close() {
+        for (Channel ch : createdChannels) {
+            if (ch.isOpen()) {
+                ch.close().syncUninterruptibly();
```
trustin: Actually, this could be done asynchronously and you can wait for the future after calling close() for all channels.

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
```diff
@@ -291,7 +304,7 @@ public void close() {
                 }
 
                 if (ch.isOpen()) {
-                    ch.close();
+                    ch.close().syncUninterruptibly();
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
```diff
@@ -280,6 +287,12 @@ protected boolean offerChannel(K key, Channel channel) {
 
     @Override
     public void close() {
+        for (Channel ch : createdChannels) {
```
trustin: `creativeChannels` is not thread safe.

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
```diff
@@ -280,6 +287,12 @@ protected boolean offerChannel(K key, Channel channel) {
...
+            if (ch.isOpen()) {
+                ch.close().syncUninterruptibly();
+            }
+        }
+        createdChannels.clear();
         for (Iterator<Deque<Channel>> i = pool.values().iterator(); i.hasNext();) {
```
trustin: It seems like we do not need to iterate over `pool` anymore because we have all the channels in `pool` also in `createdChannels`.

Instead, we need to iterate over `pendingConnections` and close them as soon as they are connected.

This introduces the risk of `close() being taking too long depending on network condition, so we may want to allow closing the channels in `pendingConnections` asynchronously as an exception.

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
anuraaga: I thought about this more and realized even with the current code, I think it would deadlock if close is called from the pool's event loop thread since the callbacks would need the event loop to be free to run. I can't think of a way to fix this and keep the synchronous nature of `close()` - I could make `close()` return a promise, but then it wouldn't implement `AutoCloseable` anymore. 

I'm thinking of just leaving these as `sync` since I guess it's not a huge deal if `close()` doesn't block - what do you think?

trustin: I'm fine with closing synchronously, but we can make the process faster like this:

```java
Queue<ChannelFuture> closeFutures = new ArrayDeque<>();
for (Channel ch : allChannels) {
    closeFutures.add(ch.close());
}

for (;;) {
    ChannelFuture f = closeFutures.poll();
    if (f == null) break;
    f.syncUninterruptibly();
}
```

anuraaga: Yeah but now I'm worried that that will deadlock if `close` is called from the event loop. Do you think it will work? Technically, it looks like it wouldn't lock and throw an exception instead

```java
protected void checkDeadLock() {
        EventExecutor e = executor();
        if (e != null && e.inEventLoop()) {
            throw new BlockingOperationException(toString());
        }
    }
```

I could make it synchronous when off event loop and async otherwise, but seems a little weird.

trustin: > I could make it synchronous when off event loop and async otherwise, but seems a little weird.

Yeah, a little bit weird, but that's probably the best I can.

anuraaga:
Updated to block when not on the event loop thread, and to not sync sequentially.

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
```diff
@@ -74,6 +84,7 @@ public DefaultKeyedChannelPool(EventLoop eventLoop, Function<K, Future<Channel>>
 
         pool = new ConcurrentHashMap<>();
```
trustin: We can make this a `HashMap` now that `pool` is only accessed from `eventLoop`. `close()` was the only place that accesses `pool` out of `eventLoop`. I may have misread the code, so would you mind double-checking?

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
```diff
@@ -280,20 +305,30 @@ protected boolean offerChannel(K key, Channel channel) {
...
+        for (Channel ch : allChannels) {
+            if (ch.isOpen()) {
+                closeFutures.add(ch.close());
             }
         }
+        allChannels.clear();
+
+        if (blocking) {
+            closeFutures.forEach(ChannelFuture::syncUninterruptibly);
+        }
```
trustin: Shouldn't we clear `closeFutures` as well?

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
```diff
@@ -280,20 +305,30 @@ protected boolean offerChannel(K key, Channel channel) {
...
-                if (ch == null) {
-                    break;
-                }
+    private void doClose(boolean blocking) {
+        closed = true;
 
-                if (ch.isOpen()) {
-                    ch.close();
-                }
+        List<ChannelFuture> closeFutures = new ArrayList<>();
```
trustin: Could specify initial capacity

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
anuraaga: Sweet, thanks for noticing!

core/src/main/java/com/linecorp/armeria/client/pool/DefaultKeyedChannelPool.java
anuraaga: Since it's just a local variable that gets GC-able right away, figured it's better not to explicitly clear, but let me know if you'd still prefer it.

trustin: Ah, yeah. No worries. Just missed that part. :-)

PR #979, opened 2018-01-29
Annotate all public packages with NonNullByDefault

### LINKED ISSUE #943: Make use of @Nonnull annotation
We currently assume all return values and parameters are non-null unless it's annotated with `@Nullable` explicitly. However, some languages or validation frameworks seem to expect the explicit use of `@Nonnull` annotations.

It would be great if we could apply some compiler plugin to automatically annotate all return types, parameters and fields with `@Nonnull`.

trustin:
Motivation:

Some languages such as Kotlin assumes all Java API are nullable unless
annotated with `@Nonnull`. Our API assumes non-null unless annotated
with `@Nullable`.  As a result, Kotlin complains about Armeria usage.

Modifications:

- Add a new annotation `@NonNullByDefault` to tell Kotlin that all
  return values, parameters and fields are non-null by default
- Annotate all packages with `@NonNullByDefault`
- Include `jsr305` as a dependency so that Kotlin compiler enables
  JSR-305 support

Result:

- Fixes #943

/cc @tobias- @anuraaga @kojilin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,8 +59,7 @@ configure(projectsWithFlags('java')) {
         compile 'com.google.guava:guava'
 
         // JSR305
-        compileOnly 'com.google.code.findbugs:jsr305'
-        testCompile 'com.google.code.findbugs:jsr305'
+        compile 'com.google.code.findbugs:jsr305'
 
         // JCTools
         compile 'org.jctools:jctools-core'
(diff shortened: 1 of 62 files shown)


trustin:
@tobias- Please let me know if this work for your Kotlin project.

trustin:
According to @tobias-, I have to introduce a new annotation because `@ParametersAreNonnullByDefault` does not cover methods. We need something like this:

```java
@Documented
@Nonnull
// @ParametersAreNonnullByDefault does not have ElementType.METHOD.
@TypeQualifierDefault({ElementType.METHOD, ElementType.PARAMETER})
public @interface NonNullApi {
}
```

trustin:
Please review again, @minwoox @hyangtack @kojilin

PR #981, opened 2018-01-29
Only cache non-exact path mappings when opted-in by the service.

### LINKED ISSUE #967: Disable path mapping cache for prefix mapping
We currently cache a path mapping result as long as the response status is successful and the mapping does not have any path parameters: https://github.com/line/armeria/blob/8e6cbc36a92de7ae3e03a41531e7ab486cd0121a/core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java#L334-L338

This does not work well with prefix mapping which can map to infinit number of paths, because the mapping result will always have zero path parameters.

To address this issue, we could:

1. Disable caching for prefix mapping by default
2. Provide a way to enable path mapping caching for some special cases, e.g. `HttpFileService` would use prefix mapping but still could use cache because it will return successful response status for a limited number of paths.
   - Annotation? e.g. `@EnablePathMappingCache`
   - Interface?

     ```java
     public interface SomeInterfaceThatServiceWillImplement {
         boolean canCache(PathMapping, PathAndQuery);
     }
     ```

Thoughts?

/cc @hyangtack @huydx @anuraaga

anuraaga:
Currently, prefix mappings used with an infinite number of URLs (common for a frontend proxy) will cause cache-thrashing. While it can be disabled for these situations, defaulting to exact paths + our common services should be good for most cases either way.

Fixes #967

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
@@ -23,6 +23,7 @@ import javax.annotation.Nullable;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
 
@@ -90,6 +91,11 @@ public final class PathAndQuery {
         if (CACHE != null) {
             CACHE.put(rawPath, this);
         }
+    }
+
+    @VisibleForTesting
+    public static Cache<String, PathAndQuery> pathCache() {
+        return CACHE;
     }
 
     private final String path;
--- a/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
@@ -16,7 +16,10 @@
 
 package com.linecorp.armeria.internal;
 
+import static java.util.Objects.requireNonNull;
+
 import java.util.Objects;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -92,6 +93,11 @@ public void storeInCache(String rawPath) {
         }
     }
 
+    @VisibleForTesting
+    public static Cache<String, PathAndQuery> pathCache() {
```
trustin: Javadoc?

core/src/test/java/com/linecorp/armeria/server/HttpServerTest.java
```diff
@@ -894,6 +913,33 @@ public void testTrailers() throws Exception {
         assertThat(res.trailingHeaders().get(AsciiString.of("foo")), is("bar"));
     }
 
+    @Test(timeout = 10000)
+    public void testExactPathCached() throws Exception {
+        org.assertj.core.api.Assertions.assertThat(client().get("/cached-exact-path")
```
trustin: Would you mind just replacing hamcrest with assertj? 😉 I can send a follow-up PR if you do, though.

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
trustin: Ah, never mind. `PathAndQuery` is already an internal class.

core/src/main/java/com/linecorp/armeria/server/DecoratingService.java
```diff
@@ -65,6 +66,11 @@ public void serviceAdded(ServiceConfig cfg) throws Exception {
         return result.isPresent() ? result : delegate.as(serviceType);
     }
 
+    @Override
+    public boolean shouldCachePath(PathAndQuery pathAndQuery, PathMapping pathMapping) {
```
trustin: `PathAndQuery` is an internal class. Could you replace it with two `String` parameters?

core/src/test/java/com/linecorp/armeria/server/HttpServerTest.java
anuraaga: Yep I was thinking of doing it myself in a follow up PR, will be much clearer

trustin: Thanks!

core/src/main/java/com/linecorp/armeria/server/Service.java
```diff
@@ -130,4 +132,12 @@ R decorate(Function<T, R> decorator) {
     default Service<I, O> decorate(DecoratingServiceFunction<I, O> function) {
         return new FunctionalDecoratingService<>(this, function);
     }
+
+    /**
+     * Returns whether the given {@code path} and {@code query }should be cached if the service's result is
```
imasahiro: nit. Move whitespace before `should`
`{@code query }should`
->
`{@code query} should`

anuraaga: Oops, thanks

trustin:
Thanks, @minwoox

trustin:
Thanks, @anuraaga !

PR #987, opened 2018-02-01
Fix the release task

trustin:
Motivation:

`release` task did not handle an SSH remote URL whose path starts with a
colon (:) correctly, e.g. github.com:line/armeria.git

Modifications:

- Match host and path correctly for both forms:
  - `scheme://host/path`
  - `user@host:path`

Result:

- We can release.


trustin:
Thanks, reviewers!

PR #993, opened 2018-02-06
Allow specifying no default port when building PropertiesEndpointGroup

trustin:
Motivation:

A user does not always want to specify the default port number when
building a PropertiesEndpointGroup, especially for HTTP and HTTPS
endpoints that run on their default ports.

Modifications:

- Add two overloaded factory methods that do not require 'defaultPort'
- Fix off-by-one validation bug with 'defaultPort'
- Overall clean-up on Javadoc and exception messages

Result:

- A user can build PropertiesEndpointGroup without defaultPort
- Specifying 0 as defaultPort will trigger an exception at earlier point

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
@@ -19,26 +19,92 @@ package com.linecorp.armeria.client.endpoint;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+import java.util.Properties;
+
 import org.junit.Test;
 
 import com.linecorp.armeria.client.Endpoint;
 
 public class PropertiesEndpointGroupTest {
+
+    private static final Properties PROPS = new Properties();
+
+    static {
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -25,20 +25,42 @@
 
 public class PropertiesEndpointGroupTest {
     @Test
-    public void test() {
+    public void testWithoutDefaultPort() {
+        PropertiesEndpointGroup endpointGroup = PropertiesEndpointGroup.of(
+                getClass().getClassLoader(), "server-list.properties", "serverA.hosts");
+
+        assertThat(endpointGroup.endpoints()).containsOnly(Endpoint.of("127.0.0.1:8080"),
+                                                            Endpoint.of("127.0.0.1:8081"),
```
hyangtack: nit: alignment :-)

trustin: Oops. Fixed :-)

trustin:
Removed the 'defect' label because the exception message is actually correct. I made the exception is raised at earlier point for less confusion and correctness.

trustin:
Updated the test case to increase the test coverage.

trustin:
Thanks, reviewers!

PR #995, opened 2018-02-07
Expose prefix-mapped THttpServices in DocService

trustin:
Motivation:

Although it's strongly recommended to use exact path mapping when a user
binds a THttpService, a user may want to or mistakenly bind his or her
service with a prefix path mapping.

When a THttpService is bound with a prefix path mapping, a user cannot
use a debug form because ThriftDocServicePlugin expects an exact path
mapping.

Modifications:

- Make ThriftDocServicePlugin also accept a prefix path mapping.

Result:

- Less surprising


trustin:
Thanks, reviewers!

Issue #996, opened 2018-02-08
Throttling decorator for rpc can't works well


kojilin:
due to https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingService.java#L70

it may create RpcResponse of RpcResponse.

I think 1 way is make `RpcResponse#from` supports `CompletableFuture<? extends RpcResponse>` like
```
static RpcResponse from(CompletionStage<?> stage) {
        requireNonNull(stage, "stage");
        final DefaultRpcResponse res = new DefaultRpcResponse();
        stage.whenComplete((value, cause) -> {
            if (cause != null) {
                res.completeExceptionally(cause);
                return;
            }
            if (value instanceof RpcResponse) {
                ((RpcResponse) value).whenComplete((rpcResponseResult, rpcResponseCause) -> {
                    if (rpcResponseCause != null) {
                        res.completeExceptionally(Exceptions.peel(rpcResponseCause));
                        return;
                    }
                    res.complete(rpcResponseResult);
                });
            } else {
                res.complete(value);
            }
        });
        return res;
    }
```

But this make me think how about `CompletableFuture<CompletableFuture<RpcResponse>>` or nested `RpcResponse`. or we should separate RPC's implementation of `ThrottlingRpcService` from `ThrottlingService`?

jmostella:
How about changing `CompletableFuture<Boolean> ThrottlingStrategy.accept -> boolean ThrottlingStrategy.accept`?

Practically speaking, in `RateLimitingThrottlingStrategy` its calling `RateLimiter.tryAcquire` which internally has a mutex lock here: https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/RateLimiter.java#L348

I can't imagine similar strategies won't have the same type of penalty. So I don't see the value in having this op be asynchronous.

kojilin:
I'm not sure if there is a needs about it. And maybe not all user use RateLimitingThrottlingStrategy?

anuraaga:
I think a method that returns a completed future and a synchronous method are conceptually the same, so switching to synchronous can never be a real solution, it's more for situations where the extra allocations / volatile reads of the future would be prohibitive. Here that doesn't seem to be the case and I think adding a sync version seems weird.

I think @kojilin's proposal seems correct, currently `RpcResponse.from` does not work the same as `HttpResponse.from`, which seems like a bug in general, not just this decorator.

[...]

PR #997, opened 2018-02-08
Fix ThrottlingService for rpc can works correctly

kojilin:
Due to `CompletableFuture<RpcResponse>`, current throttling rpc may create `RpcResponse<RpcResponse>`.

https://github.com/line/armeria/issues/996

### DIFF
--- a/thrift/src/test/java/com/linecorp/armeria/it/thrift/ThrottlingRpcServiceTest.java
+++ b/thrift/src/test/java/com/linecorp/armeria/it/thrift/ThrottlingRpcServiceTest.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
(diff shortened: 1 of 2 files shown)


thrift/src/test/java/com/linecorp/armeria/it/thrift/ThrottlingRpcServiceTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.linecorp.armeria.client.ClientBuilder;
+import com.linecorp.armeria.server.ServerBuilder;
+import com.linecorp.armeria.server.thrift.THttpService;
+import com.linecorp.armeria.server.thrift.ThriftCallService;
+import com.linecorp.armeria.server.throttling.ThrottlingRpcService;
+import com.linecorp.armeria.service.test.thrift.main.HelloService;
+import com.linecorp.armeria.testing.server.ServerRule;
+
+public class ThrottlingRpcServiceTest {
```
kojilin: Any good place to put this test? Because I want to just RPC service in `ServerRule`.

trustin:
It seems like #997 and #998 are trying to solve the same problem although #998 adds some additional factory method. How shall we proceed about this? @imasahiro @kojilin Could you guys do some discussion so you can merge your effort?

PR #998, opened 2018-02-08
Unwrap Future of RpcResponse to encode service response correctly

imasahiro:


core/src/main/java/com/linecorp/armeria/common/RpcResponse.java
```diff
@@ -61,6 +63,29 @@ static RpcResponse from(CompletionStage<?> stage) {
...
+        stage.whenComplete((delegate, thrown) -> {
+            if (thrown != null) {
+                res.completeExceptionally(Exceptions.peel(thrown));
+            } else if (delegate == null) {
+                res.completeExceptionally(
+                        new NullPointerException("delegate stage produced a null response: " + stage));
+            } else if (delegate.isCompletedExceptionally()) {
+                res.completeExceptionally(delegate.cause());
+            } else {
+                res.complete(delegate.join());
```
kojilin: This line is dangerous, need to wait delegate completes, so we should do nested whenComplete.

thrift/src/test/java/com/linecorp/armeria/it/thrift/ThrottlingRpcServiceTest.java
```diff
@@ -0,0 +1,75 @@
...
+                                                         .decorate(ThrottlingRpcService.newDecorator(never()))
+                                                         .decorate(THttpService.newDecorator()));
+
+            sb.service("/thrift-always", ThriftCallService.of(serviceHandler)
+                                                          .decorate(ThrottlingRpcService.newDecorator(always()))
+                                                          .decorate(THttpService.newDecorator()));
+        }
+    };
+
+    private final HelloService.Iface serviceHandler = mock(HelloService.Iface.class);
```
trustin: How about using `@Mock` annotation and `MockitoJUnit.rule()` so that a new mock is created for each test method?

imasahiro: Tried to use `MockitoRule` but due to ordering issue, failed to apply `MockitoRule` and `ServerRule` at once. (And RuleChain does not support `MethodRule`, cannot apply it)
Let me keep it as is.

trustin: Then how about moving the mock creation into each method?

imasahiro: Realized that if `autoStart=false`, we can use `@Mock` or create mocks into each methods.
How do you think about this commit? https://github.com/line/armeria/pull/998/commits/7db24c0be99f623d6064b2ecf6206103dd31988c

trustin:
Thanks, @imasahiro and @kojilin !

PR #1002, opened 2018-02-09
Changes ThrottlingStrategy.accept to boolean

jmostella:
Simplifies logic for ThrottlingService

Should solve: https://github.com/line/armeria/issues/996

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/CircuitBreakerThrottlingStrategy.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/CircuitBreakerThrottlingStrategy.java
@@ -17,8 +17,6 @@ package com.linecorp.armeria.server.throttling;
 
 import static java.util.Objects.requireNonNull;
 
-import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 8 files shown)


kojilin:
I replied at https://github.com/line/armeria/issues/996 . not sure if this is a good idea to just remove future for solving this issue.

trustin:
A user may want to decide after consuming the request body?

jmostella:
@trustin Understood, but in that case couldn't they create their own `DeferedRpcThrottlingService` (just as example).

Unless this is the common use-case.

trustin:
Yeah, I'm fine with providing two variants. How about updating this PR that way?

trustin:
@jmostella Discussion continuing at #996

jmostella:
@trustin per discussion should I will close this PR? Since it doesn't directly solve the issue.

trustin:
> @trustin per discussion should I will close this PR? Since it doesn't directly solve the issue.

I'm leaning toward @anuraaga 's suggestion. +1 for closing this PR.

PR #1004, opened 2018-02-12
Implement ServerListenerBuilder

### LINKED ISSUE #1003: Add ServerListenerBuilder
Until now, `ServerListener` instance should be written by the user one by one. This approach is enough when there are only a few `ServerListener`s, but if lots of `ServerListener`s are required, this approach becomes not only cumbersome but also time-consuming and error-prone. Writing Armeria application that interacts with Zookeeper using Curator is a good example of this (#882) - you have to start and stop some Curator instances when the Server is started or stopping. Even worse, most of those routines are repetitive.

If Armeria supports an easy way to build `ServerListener` instance by composing `Runnable` instances, there are following advantages:

- Easy to write: the listener can be written with only a few lambda syntax.
- Better Readability: it focuses what it does.
- Reusable: we can define reusable capsulations of the repetitive routines.

dongjinleekr:
This commit resolves #1003.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
@@ -21,6 +21,8 @@ import static java.util.Objects.requireNonNull;
 import java.util.ArrayList;
 import java.util.List;
 
+import com.google.common.collect.ImmutableList;
+
 /**
  * Builds a new {@link ServerListener}.
  * <h2>Example</h2>
@@ -89,38 +91,30 @@ public class ServerListenerBuilder {
                                List<Runnable> serverStartedCallbacks,
                                List<Runnable> serverStoppingCallbacks,
                                List<Runnable> serverStoppedCallbacks) {
-            this.serverStartingCallbacks = serverStartingCallbacks;
-            this.serverStartedCallbacks = serverStartedCallbacks;
-            this.serverStoppingCallbacks = serverStoppingCallbacks;
-            this.serverStoppedCallbacks = serverStoppedCallbacks;
+            this.serverStartingCallbacks = ImmutableList.copyOf(serverStartingCallbacks);
+            this.serverStartedCallbacks = ImmutableList.copyOf(serverStartedCallbacks);
+            this.serverStoppingCallbacks = ImmutableList.copyOf(serverStoppingCallbacks);
+            this.serverStoppedCallbacks = ImmutableList.copyOf(serverStoppedCallbacks);
         }
 
         @Override
         public void serverStarting(Server server) {
-            for (Runnable serverStartingCallback : serverStartingCallbacks) {
-                serverStartingCallback.run();
-            }
+            serverStartingCallbacks.forEach(Runnable::run);
         }
 
         @Override
         public void serverStarted(Server server) {
-            for (Runnable serverStartedCallback : serverStartedCallbacks) {
-                serverStartedCallback.run();
-            }
+            serverStartedCallbacks.forEach(Runnable::run);
         }
 
         @Override
         public void serverStopping(Server server) {
-            for (Runnable serverStoppingCallback : serverStoppingCallbacks) {
-                serverStoppingCallback.run();
-            }
+            serverStoppingCallbacks.forEach(Runnable::run);
         }
 
         @Override
         public void serverStopped(Server server) {
-            for (Runnable serverStoppedCallback : serverStoppedCallbacks) {
-                serverStoppedCallback.run();
-            }
+            serverStoppedCallbacks.forEach(Runnable::run);
         }
     }
 
@@ -129,7 +123,7 @@ public class ServerListenerBuilder {
      * (see: {@link ServerListener#serverStarting(Server)})
      */
     public ServerListenerBuilder addStartingCallback(Runnable runnable) {
-        serverStartingCallbacks.add(requireNonNull(runnable));
+        serverStartingCallbacks.add(requireNonNull(runnable, "runnable"));
         return this;
     }
 
@@ -138,6 +132,7 @@ public class ServerListenerBuilder {
      * (see: {@link ServerListener#serverStarting(Server)})
      */
     public ServerListenerBuilder addStartingCallbacks(Runnable... runnables) {
+        requireNonNull(runnables, "runnables");
         for (Runnable runnable : runnables) {
             serverStartingCallbacks.add(requireNonNull(runnable));
         }
@@ -149,6 +144,7 @@ public class ServerListenerBuilder {
      * (see: {@link ServerListener#serverStarting(Server)})
      */
     public ServerListenerBuilder addStartingCallbacks(Iterable<Runnable> runnables) {
+        requireNonNull(runnables, "runnables");
         for (Runnable runnable : runnables) {
             serverStartingCallbacks.add(requireNonNull(runnable));
         }
@@ -160,7 +156,7 @@ public class ServerListenerBuilder {
      * (see: {@link ServerListener#serverStarted(Server)})
      */
     public ServerListenerBuilder addStartedCallback(Runnable runnable) {
-        serverStartedCallbacks.add(requireNonNull(runnable));
+        serverStartedCallbacks.add(requireNonNull(runnable, "runnable"));
         return this;
     }
 
@@ -169,6 +165,7 @@ public class ServerListenerBuilder {
      * (see: {@link ServerListener#serverStarted(Server)})
      */
     public ServerListenerBuilder addStartedCallbacks(Runnable... runnables) {
+        requireNonNull(runnables, "runnables");
         for (Runnable runnable : runnables) {
             serverStartedCallbacks.add(requireNonNull(runnable));
         }
@@ -180,6 +177,7 @@ public class ServerListenerBuilder {
      * (see: {@link ServerListener#serverStarted(Server)})
      */
     public ServerListenerBuilder addStartedCallbacks(Iterable<Runnable> runnables) {
+        requireNonNull(runnables, "runnables");
         for (Runnable runnable : runnables) {
             serverStartedCallbacks.add(requireNonNull(runnable));
         }
@@ -191,7 +189,7 @@ public class ServerListenerBuilder {
      * (see: {@link ServerListener#serverStopping(Server)})
      */
     public ServerListenerBuilder addStoppingCallback(Runnable runnable) {
-        serverStoppingCallbacks.add(requireNonNull(runnable));
+        serverStoppingCallbacks.add(requireNonNull(runnable, "runnable"));
         return this;
     }
 
@@ -200,6 +198,7 @@ public class ServerListenerBuilder {
      * (see: {@link ServerListener#serverStopping(Server)})
      */
     public ServerListenerBuilder addStoppingCallbacks(Runnable... runnables) {
+        requireNonNull(runnables, "runnables");
         for (Runnable runnable : runnables) {
             serverStoppingCallbacks.add(requireNonNull(runnable));
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Builds a new {@link ServerListener}.
+ * */
+public class ServerListenerBuilder {
+
+    /**
+     * {@link Runnable}s invoked on when the {@link Server} is starting.
```
anuraaga: Don't need `on`

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+     * */
+    private List<Runnable> onStartings = new ArrayList<>();
```
anuraaga: How about watching the names of server listener?

`serverStartingCallbacks` etc

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * {@link Runnable}s invoked on when the {@link Server} is stopped.
+     * */
+    private List<Runnable> onStoppeds = new ArrayList<>();
+
+    /**
+     * Set {@link Runnable}s invoked on when the {@link Server} is starting.
+     * (see: {@link ServerListener#serverStarting(Server)})
+     */
+    public ServerListenerBuilder onStarting(Runnable... runnables) {
```
anuraaga: Let's provide `Iterable` versions too.

Also, just wondering, should these set or add, or should we provide both versions?

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+    public ServerListenerBuilder onStopped(Runnable... runnables) {
+        onStoppeds = Arrays.asList(runnables);
+        return this;
+    }
+
+    /**
+     * Returns a newly-created {@link ServerListener} based on the {@link Runnable}s added to this builder.
+     */
+    public ServerListener build() {
+        return new ServerListener() {
```
anuraaga: Let's definite a static class and pass the callbacks in to store as fields, we don't want a built object to reference a mutable builder. Also use immutablelist for the built objects.

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * {@link Runnable}s invoked on when the {@link Server} is starting.
+     * */
+    private List<Runnable> onStartings = new ArrayList<>();
+
+    /**
+     * {@link Runnable}s invoked on when the {@link Server} is started.
+     * */
+    private List<Runnable> onStarteds = new ArrayList<>();
```
trustin: All these fields could be final.

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
trustin: Also the one that accepts `Consumer<? super Server>`

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+
+package com.linecorp.armeria.server;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Builds a new {@link ServerListener}.
+ * */
```
trustin: Please add an example.

core/src/test/java/com/linecorp/armeria/server/ServerListenerTest.java
```diff
@@ -0,0 +1,112 @@
...
+    private static long STOPPING_AT = 0L;
+    private static long STOPPED_AT = 0L;
+
+    @ClassRule
+    public static final ServerRule server = new ServerRule() {
+        @Override
+        protected void configure(ServerBuilder sb) {
+            sb.meterRegistry(PrometheusMeterRegistries.newRegistry());
+
+            final Service<HttpRequest, HttpResponse> immediateResponseOnIoThread =
```
trustin: - Could be inlined
- Doesn't need a decorator. We don't use this service anyway.

core/src/test/java/com/linecorp/armeria/server/ServerListenerTest.java
```diff
@@ -0,0 +1,112 @@
...
+        assertThat(ServerListenerTest.STOPPING_AT).isGreaterThan(0L);
+        assertThat(ServerListenerTest.STOPPED_AT).isGreaterThan(ServerListenerTest.STOPPING_AT);
+    }
+
+    private static class EchoService extends AbstractHttpService {
+        @Override
+        protected final HttpResponse doPost(ServiceRequestContext ctx, HttpRequest req) {
+            return HttpResponse.from(req.aggregate()
+                                        .thenApply(this::echo)
+                                        .exceptionally(CompletionActions::log));
```
trustin: `exceptionally` will trigger NPE from `HttpResponse.from()` because `CompletionActions::log` will translate an exception into a `null` value.

Given this service will never be invoked, how about just replacing this with a hello world service for even more simplicity?

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
dongjinleekr: @anuraaga it provides set operation only.

dongjinleekr:
Here is the update:

1. Fix typo: invoked on -> invoked.
2. Rename & make immutable: `serverStartingCallbacks`, `serverStartedCallbacks`, `serverStoppingCallbacks`, `serverStoppedCallbacks`.
3. Add `CallbackServerListener`: Now returns a concrete class which stores the callbacks as fields.
4. Moved from Set operation to Add operation. Support `Iterable<Runnable>` API also.
5. Add example to Javadoc.
6. Refine `ServerListenerTest`.

dongjinleekr:
While working on this task, I found that `Consumer<? super Server>` can't be supported; to be converted into a Runnable form, it requires Server instance which can be used like `() -> consumer.accept(server)`. However, `ServerListenerBuilder` has no information of `Server` instance so it can't be supported.

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,258 @@
...
+        /**
+         * {@link Runnable}s invoked when the {@link Server} is stopped.
+         * */
+        private final List<Runnable> serverStoppedCallbacks;
+
+        CallbackServerListener(List<Runnable> serverStartingCallbacks,
+                               List<Runnable> serverStartedCallbacks,
+                               List<Runnable> serverStoppingCallbacks,
+                               List<Runnable> serverStoppedCallbacks) {
+            this.serverStartingCallbacks = serverStartingCallbacks;
```
anuraaga: ImmutableList.copyOf for all

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,258 @@
...
+            this.serverStartedCallbacks = serverStartedCallbacks;
+            this.serverStoppingCallbacks = serverStoppingCallbacks;
+            this.serverStoppedCallbacks = serverStoppedCallbacks;
+        }
+
+        @Override
+        public void serverStarting(Server server) {
+            for (Runnable serverStartingCallback : serverStartingCallbacks) {
```
anuraaga: Can do `callbacks.forEach(Runnable::run)`

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,258 @@
...
+            }
+        }
+    }
+
+    /**
+     * Add {@link Runnable} invoked when the {@link Server} is starting.
+     * (see: {@link ServerListener#serverStarting(Server)})
+     */
+    public ServerListenerBuilder addStartingCallback(Runnable runnable) {
+        serverStartingCallbacks.add(requireNonNull(runnable));
```
anuraaga: requireNonNull(runnable, "runnable")

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,258 @@
...
+        return this;
+    }
+
+    /**
+     * Add {@link Runnable}s invoked when the {@link Server} is starting.
+     * (see: {@link ServerListener#serverStarting(Server)})
+     */
+    public ServerListenerBuilder addStartingCallbacks(Runnable... runnables) {
```
anuraaga: requireNonNull(runnables...

dongjinleekr:
@anuraaga OMG; Thanks!

dongjinleekr:
OMG, I misapprehended; Since the ServerListener methods should be able to get the `Server` parameter, the callback type should be `Consumer<? extends Server>`, not `Runnable`. Because of the type erasion, we can not support both. Wait, I will update the PR.

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+    }
+
+    /**
+     * Add {@link Runnable} invoked when the {@link Server} is starting.
+     * (see: {@link ServerListener#serverStarting(Server)})
+     */
+    public ServerListenerBuilder addStartingCallback(Runnable runnable) {
+        requireNonNull(runnable, "runnable");
+        serverStartingCallbacks.add((Server server) -> runnable.run());
```
anuraaga: Do you need the type parameter? (not sure, maybe you do since it's a wildcard).

Also consider naming the variable `unused`

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /**
+     * Add {@link Consumer}s invoked when the {@link Server} is starting.
+     * (see: {@link ServerListener#serverStarting(Server)})
+     */
+    public ServerListenerBuilder addStartingCallbacks(Consumer<? super Server>... consumers) {
+        requireNonNull(consumers, "consumers");
+        for (Consumer<? super Server> consumer : consumers) {
+            serverStartingCallbacks.add(requireNonNull(consumer));
```
anuraaga: `requireNonNull(consumer, "consumer")`

dongjinleekr:
Applied:

1. Remove explicit consumer paramater in Add(...)Callback(Runnable) Methods: also, rename the parameter name to 'unused'.
2. `requireNonNull(consumer)` -> `requireNonNull(consumer, "consumer")`

core/src/main/java/com/linecorp/armeria/server/ServerListenerBuilder.java
```diff
@@ -0,0 +1,309 @@
...
+    public ServerListenerBuilder addStartingCallback(Consumer<? super Server> consumer) {
+        serverStartingCallbacks.add(requireNonNull(consumer, "consumer"));
+        return this;
+    }
+
+    /**
+     * Add {@link Consumer}s invoked when the {@link Server} is starting.
+     * (see: {@link ServerListener#serverStarting(Server)})
+     */
+    public ServerListenerBuilder addStartingCallbacks(Consumer<? super Server>... consumers) {
```
anuraaga: Delegate all array versions to iterable versions, e.g.,

`addStartingCallbacks(Arrays.asList(consumers))`

dongjinleekr: Sure.

dongjinleekr:
@anuraaga Thanks for your kind review. Let's continue to #1007 and #1035, which can be solved with `ServerListenerBuilder`. 😄

minwoox:
Thanks @dongjinleekr !

PR #1011, opened 2018-02-19
Add docs for annotated HTTP service

hyangtack:
Modifications:
- `@Decorator` is able to annotate a class.
- Add server-annotated-service.rst.


core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -309,19 +311,37 @@ private static String findPattern(Method method, Set<Annotation> methodAnnotatio
      * Returns a decorator chain which is specified by {@link Decorator} annotations.
      */
     private static Function<Service<HttpRequest, HttpResponse>,
-            ? extends Service<HttpRequest, HttpResponse>> decorator(Method method) {
+            ? extends Service<HttpRequest, HttpResponse>> decorator(Method method, Class<?> clazz) {
+
+        // Decorators which are specified on the class will handle a request first.
```
trustin: How about: `Method-level decorators are applied before class-level decorators.`

site/src/sphinx/server.rst
```diff
@@ -15,3 +15,4 @@ Writing a server
     server-http-file
     server-servlet
     server-access-log
+    server-annotated-service
```
trustin: We could 1) remove `server-rest` and 2) put `server-annotated-service` after `server-docservice`.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+.. _ServerBuilder: apidocs/index.html?com/linecorp/armeria/server/ServerBuilder.html
+.. _Service: apidocs/index.html?com/linecorp/armeria/server/Service.html
+.. _StringRequestConverterFunction: apidocs/index.html?com/linecorp/armeria/server/annotation/StringRequestConverterFunction.html
+
+.. _server-annotated-service:
+
+Annotated HTTP Service
+======================
+
+*Armeria* provides a way to write an HTTP service using annotations. It helps a user to make his or her codes
```
trustin: - `*Armeria*` -> `Armeria`? `*Armeria*` was used in the front page since I was defining a new term that appeared first time.
- help a user to make -> help a user make ('to' is not necessary for 'help')
- codes -> code (code is uncountable.)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+simple and easy to understand. A user is able to run an HTTP service by couple of lines codes using
```
trustin: couple of lines codes -> fewer lines of code

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+            return HttpResponse.of(HttpStatus.OK,
+                                   MediaType.PLAIN_TEXT_UTF_8,
+                                   "Hello, %s!", name);
+        }
+    });
+
+Mapping HTTP service methods
+----------------------------
+
+To map each method in an annotated HTTP service class to HTTP path, both `@Path`_ annotation and HTTP method
```
trustin: - each method -> a Java method
- both .. annotation ... are used -> it has to be annotated with both ...
- Could we start from the simpler case where no `@Path` annotation is required instead? We could even explain our path mapping syntax without introducing `@Path` annotation at all and then explain how to handle multiple HTTP method.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+- `@Get`_
+- `@Head`_
+- `@Post`_
+- `@Put`_
+- `@Delete`_
+- `@Options`_
+- `@Patch`_
+- `@Trace`_
+
+If you want to handle several types of HTTP methods in a single service method, you can annotate your
```
trustin: - several types of HTTP methods -> more than one HTTP method type

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
+        @Get("/hello")
+        public HttpResponse hello() {
+            ...
+        }
+    }
+
+There are 5 types of PathMapping_ s provided for describing a path with `@Path`_ annotation.
```
trustin: 5 PathMapping types ?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+
+- Exact mapping
```
trustin: How about moving the examples after the mapping name, so that the examples are more visible?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+
+- Mapping with regular expression
+
+  - a service method will handle the path matched with the specified regular expression. If a named capturing
+    group exists in the regular expression, it may be mapped to a parameter of the service method.
+    e.g, ``regex:^/hello/(?<name>.*)$``
+
+- Mapping with glob pattern
+
+  - a service method will handle the path matched with the specified glob pattern.
```
trustin: Could you mention that each component will be mapped to the parameter whose name is an integer? (with example, please)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        }
+
+        @Get("regex:^/hello/(?<name>.*)$")
+        public HttpResponse regex(@Param("name") String name) {
+            ...
+        }
+    }
+
+When the value of `@Param`_ annotation is not shown in the path pattern, it will be handled as a parameter
+name of the query string of the request. If you have a service class like the below and a user sends an
```
trustin: - like the below -> like the example below?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+HTTP GET request with URI of ``/hello1?name=armeria``, the service method will get ``armeria`` as the value
+of parameter ``name``. If there is no parameter named ``name`` in the query string, the parameter ``name``
+of the method would be ``null``. If you want to avoid ``null`` in this case, you can use `@Default`_
+annotation or ``Optional<?>`` class, e.g, ``hello2`` and ``hello3`` method in the example, respectively.
```
trustin: e.g, -> e.g. (dot)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
+
+        @Get("/hello1")
+        public HttpResponse hello1(@Param("name") String name) {
+            ...
+        }
```
trustin: How about compacting into a single line? `{ ... }`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        public HttpResponse hello3(@Param("name") Optional<String> name) {
+            String clientName = name.orElse("armeria");
+            ...
+        }
+    }
+
+``Enum`` type is also automatically converted if you annotate a parameter of your service method with
+`@Param`_ annotation. If your ``Enum`` type is able to be handled in a case-insensitive way, *Armeria*
+automatically converts the string value of a parameter to a value of ``Enum`` in a case-insensitive way.
+Otherwise, *Armeria* tries to find the value of ``Enum`` which is exactly matched to the value of
```
trustin: How about: Otherwise, case-sensitive exact match will be performed.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+            // So 'ALPHA' and 'alpha' are only acceptable.
+        }
+    }
+
+Handling exceptions
+-------------------
+
+An exception may be raised from a service method, of course. *Armeria* provides `@ExceptionHandler`_
+annotation to transform an exception into a response. You can write your own exception handler
+by implementing ExceptionHandlerFunction_ interface. Then, annotate your service object or method
```
trustin: interface. Then, annotate -> interface and annotate

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+with `@ExceptionHandler`_ annotation. Here is an example of an exception handler. If your exception
+handler is not able to handle an exception given, you can call ``ExceptionHandlerFunction.fallThrough()``
```
trustin: - a given exception?
- fallThrough -> fallthrough

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+to pass the exception to the next exception handlers.
```
trustin: - handlers -> handler

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        }
+    }
+
+If there is no exception handler which is able to handle an exception, the exception would be passed to
+the default exception handler. It handles ``IllegalArgumentException``, HttpStatusException_ and
+HttpResponseException_ by default. ``IllegalArgumentException`` would be converted into ``400 Bad Request``
+response, and the other two exceptions would be converted into a response with the status code which
+they are holding. For another exceptions, ``500 Internal Server Error`` would be sent to the client.
+
+Converting between an HTTP message and a Java object
```
trustin: Conversion

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+----------------------------------------------------
+
+Converting an HTTP request to a Java object
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In some cases like receiving a JSON document from a client, it may be useful to convert the document to
+a Java object automatically. So *Armeria* supports that using `@RequestConverter`_ and `@RequestObject`_
```
trustin: So Armeria Supports that using -> Armeria provides ... annotations so that such conversion can be done conveniently.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        @Post("/hello")
+        public HttpResponse hello(@RequestObject MyObject myObject) {
+            // MyRequestConverter will be used to convert a request.
+            ...
+        }
+
+        @Post("/hola")
+        @RequestConverter(MySpanishRequestConverter.class)
+        public HttpResponse hola(@RequestObject MySpanishObject myObject) {
+            // MySpanishRequestConverter and MyRequestConverter will be used to convert a request.
```
trustin:
```java
// MySpanishRequestConverter will be tried to convert a request first.
// MyRequestConverter will be used if MySpanishRequestConverter fell through.
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+
+*Armeria* also provides built-in request converters such as, JacksonRequestConverterFunction_ for JSON
+documents, StringRequestConverterFunction_ for text contents and ByteArrayRequestConverterFunction_
+for binary contents. They will be applied after your request converters by default, so you can use these
+built-in converters by just putting `@RequestObject`_ annotation on the parameters which you want to convert.
+
+In some cases, a parameter of a method may be annotated with `@RequestConverter`_. Assume that you have
+a Java class named ``MyRequest`` that it is usually able to be converted by ``MyDefaultRequestConverter``.
+But what if there is only one method which has a parameter of ``MyRequest`` that you have to convert
+it differently. In this case, you may annotate a variable with `@RequestConverter`_ as follows.
```
trustin: a variable -> a parameter instead of a method

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+In the example, ``MySpecialRequestConverter`` will be used first for converting ``MyRequest``.
+
+.. code-block:: java
+
+    @RequestConverter(MyDefaultRequestConverter.class)
+    class MyAnnotatedServiceClass {
+
+        @Post("/hello")
+        public HttpResponse hello(
+            @RequestConverter(MySpecialRequestConverter.class) MyRequest myRequest) {
```
trustin: Don't we need to annotate with `@RequestObject` as well? If not, we could mention that a user doesn't have to annotate with `@RequestObject` when a parameter is annotated with `@RequestConverter`.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        }
+    }
+
+    ...
+
+    sb.annotatedService(new MyAnnotatedServiceClass(),
+                        new MyGlobalExceptionHandler5(),        // order 5
+                        new MyGlobalExceptionHandler6());       // order 6
+
+Decorating a service
```
trustin: Decorating an annotated service?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        @Override
+        public HttpResponse serve(Service<HttpRequest, HttpResponse> delegate,
+                                  ServiceRequestContext ctx,
+                                  HttpRequest req) throws Exception {
+            ... Do something ...
+            return delegate.serve(ctx, req);
+        }
+    }
+
+Then, annotate your class or method with `@Decorator`_ annotation. In the following example, ``MyDecorator``
```
trustin: a `@Decorator` annotation (a)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        @Decorator(AnotherDecorator.class)
+        @Get("/hello")
+        public HttpResponse hello() {
+            ...
+        }
+    }
+
+.. note::
+
+    The applying order of the decorators is slightly different to that of the converters and exception handlers.
```
trustin: Could you expand a little bit more on this, perhaps with an example?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        }
+
+        @Get("/hello")
+        public HttpResponse hello2() {
+            // Return a JSON object to the client.
+        }
+    }
+
+If no media type is specified on the method, the first method is usually mapped to a request for the path
+``/hello``. So you would always receive a text document from a server. What should you do if you want to
```
trustin: - `/hello` and the second method will be ignored.
- So you -> so that a client

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+
+A request like the following would be get a JSON object::
+
+    GET /hello HTTP/1.1
+    Accept: application/json
+
+.. note::
+
+    Note that ``Content-Type`` header of a response is not automatically set. You are able to get the
+    negotiated `@ProduceType`_ from ``ServiceRequestContext.negotiatedProduceType()`` method.
```
trustin: That being said, we never showed an example that injects `ServiceRequestContext`. Could you add a section about the types of automatically injected params? e.g. `ServiceRequestContext`, `HttpRequest`, `AggregatedHttpMessage`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+        }
+    }
+
+A request like the following would be handled by ``helloText()`` method::
+
+    POST /hello HTTP/1.1
+    Content-Type: text/plain
+    Content-Length: 7
+
+    bla bla
```
trustin: `Armeria`?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,627 @@
...
+
+A request like the following would be handled by ``helloJson()`` method::
+
+    POST /hello HTTP/1.1
+    Content-Type: application/json
+    Content-Length: 21
+
+    { "name": "armeria" }
```
trustin: armeria -> Armeria (A)

site/src/sphinx/server-annotated-service.rst
hyangtack: Thanks. It was my mistake, it had to be `@RequestObject`, not `@RequestConverter`. I fixed it. :-)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+            return HttpResponse.of(HttpStatus.OK,
+                                   MediaType.PLAIN_TEXT_UTF_8,
+                                   "Hello, %s!", name);
+        }
+    });
+
+Mapping HTTP service methods
+----------------------------
+
+To map a Java method in an annotated HTTP service class to HTTP path, it has to be annotated with one of
```
trustin: - HTTP path -> an HTTP path
- a Java method -> a service method (sorry!)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+HTTP method annotations. Here is the list of HTTP method annotations that Armeria supports. Each of them
```
trustin: HTTP method annotations ... -> the following HTTP method annotations where each of them is ...:

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+- `@Get`_
+- `@Head`_
+- `@Post`_
+- `@Put`_
+- `@Delete`_
+- `@Options`_
+- `@Patch`_
+- `@Trace`_
+
+To handle a single HTTP method in a single service method, you can annotate your service method simply,
```
trustin: an HTTP request with a service method, you ..

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        public HttpResponse pathvar(@Param("name") String name) { ... }
+
+        @Get("regex:^/hello/(?<name>.*)$")
+        public HttpResponse regex(@Param("name") String name) { ... }
+
+        @Get("glob:/*/hello/**")
+        public HttpResponse glob(@Param("0") String prefix, @Param("1") String name) { ... }
+    }
+
+There are Java methods which are mapped to one HTTP method in the example above. What should you do if you want
```
trustin: Every service method in the examples so far had a single HTTP method annotation with it. What if you want ..

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+to map more than one HTTP method to your Java method? You can use `@Path`_ annotation to specify a path
```
trustin: Java method -> service method

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+Armeria tries to convert a value to the specified type and ``IllegalArgumentException`` will be raised
+if the conversion failed.
+
+- ``boolean`` (or ``Boolean``)
+- ``byte`` (or ``Byte``)
+- ``short`` (or ``Short``)
+- ``integer`` (or ``Integer``)
+- ``long`` (or ``Long``)
+- ``float`` (or ``Float``)
+- ``double`` (or ``Double``)
```
trustin: `(` and `)` are unnecessary. i.e. `double or Double`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        }
+
+        @Get("/hello2/{there}")
+        public HttpResponse hello2(@Param("there") CaseSensitive there) {
+            // 'there' must be converted in a case-sensitive way.
+            // So 'ALPHA' and 'alpha' are only acceptable.
+        }
+    }
+
+When the value of `@Param`_ annotation is not shown in the path pattern, it will be handled as a parameter
```
trustin: How about adding a subsection heading right before this paragraph?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+- ``byte`` (or ``Byte``)
+- ``short`` (or ``Short``)
+- ``integer`` (or ``Integer``)
+- ``long`` (or ``Long``)
+- ``float`` (or ``Float``)
+- ``double`` (or ``Double``)
+- ``String``
+- ``Enum``
+
+``Enum`` type is also automatically converted if you annotate a parameter of your service method with
```
trustin: How about adding a subsection heading right before this paragraph?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        public HttpResponse hello2(@Param("name") @Default("armeria") String name) { ... }
+
+        @Get("/hello3")
+        public HttpResponse hello3(@Param("name") Optional<String> name) {
+            String clientName = name.orElse("armeria");
+            ...
+        }
+    }
+
+Armeria also provides `@Header`_ annotation to inject an HTTP header value into a parameter. The parameter
```
trustin: How about adding a subsection heading right before this paragraph?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+    class MyAnnotatedServiceClass {
+
+        @Get("/hello1")
+        public HttpResponse hello1(@Header("Authorization") String auth) { ... }
+
+        @Post("/hello2")
+        public HttpResponse hello2(@Param("Content-Length") long contentLength) { ... }
+    }
+
+Moreover, the following classes are automatically injected into your parameter when you specify it on the
```
trustin: - Moreover -> Also ?
- 'into your parameter' -> ''

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+- RequestContext_
+- ServiceRequestContext_
+- Request_
+- HttpRequest_
+- AggregatedHttpMessage_
+- HttpParameters_
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+        @Get("/hello1")
+        public HttpResponse hello1(ServiceRequestContext ctx, HttpRequest req) {
+            // Use the context and request inside a method.
+        }
+
+        @Post("/hello2")
+        public HttpResponse hello2(AggregatedHttpMessage aggregatedMessage) {
+            // Armeria aggregates every message received, then call this method with an aggregated message.
```
trustin: every message received, then call -> the received HttpRequest and calls .. with the aggregated request.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        }
+
+        @Get("/hello3")
+        public HttpResponse hello3(HttpParameters httpParameters) {
+            // 'httpParameters' holds parameters which are parsed from a query string of a request.
```
trustin: parameters -> the parameters parsed from ..

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        @Post("/hello4")
+        public HttpResponse hello4(HttpParameters httpParameters) {
+            // If a request has a url-encoded form as its body, it can be accessed via 'httpParameters'.
+        }
+    }
+
+Handling exceptions
+-------------------
+
+An exception may be raised from a service method, of course. Armeria provides `@ExceptionHandler`_
```

An exception may ... of course. -> It is often useful to extract exception handling logic from service methods into a separate common class.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+In some cases like receiving a JSON document from a client, it may be useful to convert the document to
+a Java object automatically. Armeria provides `@RequestConverter`_ and `@RequestObject`_ annotations
+so that such conversion can be done conveniently.
+You can write your own request converter by implementing RequestConverterFunction_ as follows.
+Similar to the exception handler, you can call ``RequestConverterFunction.fallthrough()`` when your request
+converter is not able to convert the request.
+
+.. code-block:: java
+
+    class MyRequestConverter implements RequestConverterFunction {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+        @Get("/hello")
+        public HttpResponse hello() { ... }
+    }
+
+To catch an exception from a single method in your service object, annotate your method with the annotation.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+            return ExceptionHandlerFunction.fallthrough();
+        }
+    }
+
+To catch an exception from every method in your service object, annotate your class with the annotation.
+
+.. code-block:: java
+
+    @ExceptionHandler(MyExceptionHandler.class)
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
```
trustin: Probably OK to remove this empty line because this class has only a single method?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+An exception may be raised from a service method, of course. Armeria provides `@ExceptionHandler`_
+annotation to transform an exception into a response. You can write your own exception handler
+by implementing ExceptionHandlerFunction_ interface and annotate your service object or method
+with `@ExceptionHandler`_ annotation. Here is an example of an exception handler. If your exception
+handler is not able to handle a given exception, you can call ``ExceptionHandlerFunction.fallthrough()``
+to pass the exception to the next exception handler.
+
+.. code-block:: java
+
+    class MyExceptionHandler implements ExceptionHandlerFunction {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        }
+    }
+
+Armeria also provides `@Header`_ annotation to inject an HTTP header value into a parameter. The parameter
+annotated with `@Header`_ is also specified as one of the built-in types as follows. `@Default`_ and
+``Optional<?>`` are also supported.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+When the value of `@Param`_ annotation is not shown in the path pattern, it will be handled as a parameter
+name of the query string of the request. If you have a service class like the example below and a user sends an
+HTTP GET request with URI of ``/hello1?name=armeria``, the service method will get ``armeria`` as the value
+of parameter ``name``. If there is no parameter named ``name`` in the query string, the parameter ``name``
+of the method would be ``null``. If you want to avoid ``null`` in this case, you can use `@Default`_
+annotation or ``Optional<?>`` class, e.g. ``hello2`` and ``hello3`` method in the example, respectively.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+.. code-block:: java
+
+    enum CaseSensitive {
+        ALPHA, alpha
+    }
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+.. code-block:: java
+
+    enum CaseInsensitive {
+        ALPHA, BRAVO, CHARLIE
+    }
+
+.. code-block:: java
+
+    enum CaseSensitive {
```
trustin: `public enum`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+- ``Enum``
+
+``Enum`` type is also automatically converted if you annotate a parameter of your service method with
+`@Param`_ annotation. If your ``Enum`` type is able to be handled in a case-insensitive way, Armeria
+automatically converts the string value of a parameter to a value of ``Enum`` in a case-insensitive way.
+Otherwise, case-sensitive exact match will be performed.
+
+.. code-block:: java
+
+    enum CaseInsensitive {
```
trustin: `public enum`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+.. _parameter-injection:
+
+Parameter injection
+-------------------
+
+Let's see the example in the above section again.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        public HttpResponse glob(@Param("0") String prefix, @Param("1") String name) { ... }
+    }
+
+There are Java methods which are mapped to one HTTP method in the example above. What should you do if you want
+to map more than one HTTP method to your Java method? You can use `@Path`_ annotation to specify a path
+and use the HTTP method annotations without a path just to specify the HTTP methods to be mapped as follows.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+  - a service method will handle the path matched with the specified glob pattern. A wildcard is mapped to
+    an index which starts with ``0``, so it may be mapped to a parameter of the service method.
+
+You can get the value of a path variable, a named capturing group of the regular expression or wildcards of
+the glob pattern in your service method by annotating a parameter with `@Param`_ as follows.
+Please refer to :ref:`parameter-injection` for more information about `@Param`_.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+- `@Options`_
+- `@Patch`_
+- `@Trace`_
+
+To handle a single HTTP method in a single service method, you can annotate your service method simply,
+as follows.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
```
trustin: Could remove this empty line?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        @Post("/hello")
+        public HttpResponse hello(
+            @RequestObject(MySpecialRequestConverter.class) MyRequest myRequest) { ... }
+    }
+
+Converting a Java object to an HTTP response
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Every object returning by the annotated service method is able to convert to an HTTP response message by
```
trustin: - returning by .. -> returned by an annotated ...
- is able to convert to -> can be converted to

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+response converters, except for HttpResponse_ and AggregatedHttpMessage_ which are already in a
+form of response message. You can also write your own response converter by implementing
+ResponseConverterFunction_ as follows. Also similar to the request converter, you can call
+``ResponseConverterFunction.fallthrough()`` when your response converter is not able to convert the result
+to the HttpResponse_.
```
trustin: the -> an

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+            @RequestObject(MySpecialRequestConverter.class) MyRequest myRequest) { ... }
+    }
+
+Converting a Java object to an HTTP response
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Every object returning by the annotated service method is able to convert to an HTTP response message by
+response converters, except for HttpResponse_ and AggregatedHttpMessage_ which are already in a
+form of response message. You can also write your own response converter by implementing
+ResponseConverterFunction_ as follows. Also similar to the request converter, you can call
```
trustin: the request converter -> `RequestConverterFunction`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+``ResponseConverterFunction.fallthrough()`` when your response converter is not able to convert the result
+to the HttpResponse_.
+
+.. code-block:: java
+
+    class MyResponseConverter implements ResponseConverterFunction {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        @Override
+        public HttpResponse convertResponse(ServiceRequestContext ctx,
+                                            Object result) throws Exception {
```
trustin: Probably OK to drop `throws Exception` and make method declaration part into a single line?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+            return ResponseConverterFunction.fallthrough();
+        }
+    }
+
+You can annotate your service method and class as follows.
+
+.. code-block:: java
+
+    @ResponseConverter(MyResponseConverter.class)
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        @Post("/hello")
+        public MyObject hello() {
+            // MyResponseConverter will be used to make a response.
+            ...
+        }
+
+        @Post("/hola")
+        @ResponseConverter(MySpanishResponseConverter.class)
+        public MyObject hola() {
+            // MySpanishResponseConverter and MyResponseConverter will be used to make a response.
```
trustin:
```
            // MySpanishResponseConverter will be tried to convert a request first.
            // MyResponseConverter will be used if MySpanishResponseConverter fell through.
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+.. _more-about-converters-exception-handlers:
+
+More about converters and exception handlers
+--------------------------------------------
+
+There are 3 ways to specify converters and exception handlers, except that `@RequestConverter`_ annotates
+a variable. The first one is annotating a service method and the second one is annotating a service class,
+as you read the above sections. The last one is using ServerBuilder_ as follows.
```
trustin: How about just renaming this section to:

```rst
Using ServerBuilder_ to configure converters and exception handlers
-------------------------------------------------------------------
```

and replacing this paragraph with:

```rst
You can specify converters and exception handlers using ServerBuilder_, without using the annotations explained in the previous sections::
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+.. code-block:: java
+
+    sb.annotatedService(new MyAnnotatedServiceClass(),
+                        new MyExceptionHandler(), new MyRequestConverter(), new MyResponseConverter());
+
+When you specify exception handlers in every way like below, they are working in the following order commented.
```
trustin: - in every way -> in a mixed manner
- they are working -> they will be evaluated

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+Decorating an annotated service
+-------------------------------
+
+Every Service_ is able to be wrapped by another Service_ in Armeria (Refer to :ref:`server-decorator`
+for more information). Simply, you are able to write your own decorator by implementing
+DecoratingServiceFunction_ interface as follows.
+
+.. code-block:: java
+
+    class MyDecorator implements DecoratingServiceFunction<HttpRequest, HttpResponse> {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        @Override
+        public HttpResponse serve(Service<HttpRequest, HttpResponse> delegate,
+                                  ServiceRequestContext ctx,
+                                  HttpRequest req) throws Exception {
+            ... Do something ...
```
trustin: `// ... Do something ... `

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+for more information). Simply, you are able to write your own decorator by implementing
+DecoratingServiceFunction_ interface as follows.
+
+.. code-block:: java
+
+    class MyDecorator implements DecoratingServiceFunction<HttpRequest, HttpResponse> {
+        @Override
+        public HttpResponse serve(Service<HttpRequest, HttpResponse> delegate,
+                                  ServiceRequestContext ctx,
+                                  HttpRequest req) throws Exception {
```
trustin: We could remove `throws Exception` whereever possible to reduce the cognitive load of a reader.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+Then, annotate your class or method with a `@Decorator`_ annotation. In the following example, ``MyDecorator``
+will handle a request first, then ``AnotherDecorator`` will handle the request next, and finally ``hello()``
+method will handle the request.
+
+.. code-block:: java
+
+    @Decorator(MyDecorator.class)
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+        @Decorator(AnotherDecorator.class)
+        @Get("/hello")
+        public HttpResponse hello() { ... }
+    }
+
+Not that the applying order of the decorators is slightly different from that of the converters and exception
```
trustin: - Not -> Note
- evaluation order?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+handlers. As you read in :ref:`more-about-converters-exception-handlers`, both the converters and exception
+handlers are applied in the order of method-level ones, class-level ones and global ones. Unlike them,
+decorators are applied in the opposite order as follows.
```
trustin: Could you add a paragraph about why?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+.. code-block:: java
+
+    @Decorator(MyClassDecorator2.class)                 // order 2
+    @Decorator(MyClassDecorator3.class)                 // order 3
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+Media type negotiation
+----------------------
+
+Armeria provides `@ProduceType`_ and `@ConsumeType`_ annotations to support media type negotiation. It is not
+necessary if you have only one service method for a path and an HTTP method. Assume that you have multiple
+service methods for the same path and the same HTTP method as follows.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+            // Return a text document to the client.
+        }
+
+        @Get("/hello")
+        public HttpResponse hello2() {
+            // Return a JSON object to the client.
+        }
+    }
+
+If no media type is specified on the method, the first method is usually mapped to a request for the path
```

If media type is not specified on any methods bound to the same path pattern, the first method declared will be used and the other methods will be ignored. In this example, `hello1()` will be chosen and the client will always receive a text document. What should you do ...

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+``/hello`` and the second method will be ignored so that a client would always receive a text document
+from a server. What should you do if you want to get a JSON object from the path ``/hello``? You can just
+specify the type of the content what your method produces as follows and add ``Accept`` header to your request.
```
trustin: - what your -> your
- to your client request (to clarify that this is a client-side modification)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        }
+
+        @Get("/hello")
+        @ProduceType("application/json")
+        public HttpResponse helloJson() {
+            // Return a JSON object to the client.
+        }
+    }
+
+A request like the following would be get a text document::
```
trustin: be get -> get

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+    GET /hello HTTP/1.1
+    Accept: text/plain
+
+A request like the following would be get a JSON object::
```
trustin: be get -> get

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+    GET /hello HTTP/1.1
+    Accept: application/json
+
+.. note::
+
+    Note that ``Content-Type`` header of a response is not automatically set. You are able to get the
+    negotiated `@ProduceType`_ from ``ServiceRequestContext.negotiatedProduceType()`` method.
+
+If a client sends a request without ``Accept`` header(or sending ``Accept`` header with a unsupported content
```
trustin: `header(or` -> `header (or` (space)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+        }
+    }
+
+Next, let's learn how to handle ``Content-Type`` header of a request. Assume that there are two service
+methods that expect a text document and a JSON object as a content of a request, respectively.
+You can annotate them with `@ConsumeType`_ annotation.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+However, if a client sends a request with ``Content-Type: application/octet-stream`` header which is not
+specified with `@ConsumeType`_, the client would get an HTTP status code of 415 which means
+``Unsupported Media Type``. If you want to make one of the methods catch-all, you can remove the annotation
+as follows. ``helloCatchAll()`` method would accept every request except for a request with
+``Content-Type: application/json`` header.
+
+.. code-block:: java
+
+    class MyAnnotatedServiceClass {
```
trustin: `public class`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+        @Post("/hello")
+        @ConsumeType("text/plain")
+        public HttpResponse helloText() {
+            // Handle a content of a request as a text document.
+        }
```
trustin: How about putting the actual implementation?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,718 @@
...
+
+        @Post("/hello")
+        @ConsumeType("application/json")
+        public HttpResponse helloJson() {
+            // Handle a content of a request as a JSON object.
+        }
```
trustin: How about putting the actual implementation?

```java
return HttpResponse.of(OK, JSON_UTF8, "{ \"name\": \"Armeria\" }");
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+                                   "Hello, %s!", name);
+        }
+    });
+
+Mapping HTTP service methods
+----------------------------
+
+To map a service method in an annotated HTTP service class to an HTTP path, it has to be annotated with one of
+HTTP method annotations. Here is the list of the following HTTP method annotations where each of them
+is mapped to the HTTP method which has the same name.
```
trustin: Here is... -> The following is the list of HTTP method annotations where each of them is mapped to an HTTP method?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+        @Get("regex:^/hello/(?<name>.*)$")
+        public HttpResponse regex(@Param("name") String name) { ... }
+
+        @Get("glob:/*/hello/**")
+        public HttpResponse glob(@Param("0") String prefix, @Param("1") String name) { ... }
+    }
+
+Every service method in the examples so far had a single HTTP method annotation with it. What if you want
+to map more than one HTTP method to your service method? You can use `@Path`_ annotation to specify a path
+and use the HTTP method annotations without a path just to specify the HTTP methods to be mapped as follows.
```
trustin: just to specify ... -> to map multiple HTTP methods, e.g.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+    public class MyAnnotatedServiceClass {
+
+        @Get("/hello/{name}")
+        public HttpResponse pathvar(@Param("name") String name) { ... }
+
+        @Get("regex:^/hello/(?<name>.*)$")
+        public HttpResponse regex(@Param("name") String name) { ... }
+    }
+
+A value of a parameter ``name`` is automatically injected as a ``String`` type by Armeria.
```
trustin: String type -> String?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+Which types are available for automatic injection? First, the following built-in types are available.
+Armeria tries to convert a value to the specified type and ``IllegalArgumentException`` will be raised
```
trustin: > Which types are ...? First .... are available. Armeria tries to ...

How about:

Armeria will try to convert the value appropriately if the parameter type is not `String`. `IllegalArgumentException` will be raised if the conversion fails or the parameter type is not one of the following supported types:

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+- ``float`` or ``Float``
+- ``double`` or ``Double``
+- ``String``
+- ``Enum``
+
+Injecting a parameter as an ``Enum`` type
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``Enum`` type is also automatically converted if you annotate a parameter of your service method with
+`@Param`_ annotation. If your ``Enum`` type is able to be handled in a case-insensitive way, Armeria
```
trustin: is able to be -> can be

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+
+Armeria also provides built-in request converters such as, JacksonRequestConverterFunction_ for JSON
+documents, StringRequestConverterFunction_ for text contents and ByteArrayRequestConverterFunction_
+for binary contents. They will be applied after your request converters by default, so you can use these
+built-in converters by just putting `@RequestObject`_ annotation on the parameters which you want to convert.
+
+In some cases, `@RequestObject`_ annotation may have a request converter as its value.
+Assume that you have a Java class named ``MyRequest`` that it is usually able to be converted by
+``MyDefaultRequestConverter``. But what if there is only one method which has a parameter of ``MyRequest``
+that you have to convert it differently. In this case, you may specify a request converter with
```
trustin: differently? (question mark)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+-------------------------------------------------------------------
+
+You can specify converters and exception handlers using ServerBuilder_, without using the annotations
+explained in the previous sections::
+
+    sb.annotatedService(new MyAnnotatedServiceClass(),
+                        new MyExceptionHandler(), new MyRequestConverter(), new MyResponseConverter());
+
+When you specify exception handlers in a mixed manner like below, they will be evaluated in the following
+order commented. It is also the same as the evaluation order of the converters.
```
trustin: It is also ... -> Please note that the same ordering rule applies for converters as well.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+    // ...
+
+    sb.annotatedService(new MyAnnotatedServiceClass(),
+                        new MyGlobalExceptionHandler5(),        // order 5
+                        new MyGlobalExceptionHandler6());       // order 6
+
+Decorating an annotated service
+-------------------------------
+
+Every Service_ is able to be wrapped by another Service_ in Armeria (Refer to :ref:`server-decorator`
```
trustin: is able to be -> can be

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+                        new MyGlobalDecorator1());      // order 1
+
+
+.. _media_type_nego:
+
+Media type negotiation
+----------------------
+
+Armeria provides `@ProduceType`_ and `@ConsumeType`_ annotations to support media type negotiation. It is not
+necessary if you have only one service method for a path and an HTTP method. Assume that you have multiple
```
trustin: Assume -> However, assume

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+        @Get("/hello")
+        public HttpResponse hello2() {
+            // Return a JSON object to the client.
+            return HttpResponse.of(HttpStatus.OK, MediaType.JSON_UTF8, "{ \"name\": \"Armeria\" }");
+        }
+    }
+
+If media type is not specified on any methods bound to the same path pattern, the first method declared will
+be used and the other methods will be ignored. In this example, ``hello1()`` will be chosen and the client
+will always receive a text document. What should you do if you want to get a JSON object from the path
```
trustin: What should you do if -> What if

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+
+A request like the following would get a JSON object::
+
+    GET /hello HTTP/1.1
+    Accept: application/json
+
+.. note::
+
+    Note that ``Content-Type`` header of a response is not automatically set. You are able to get the
+    negotiated `@ProduceType`_ from ``ServiceRequestContext.negotiatedProduceType()`` method.
```
trustin: You are able to get .... method.  -> You may want to get .... method and set it in the `Content-Type` header of your response.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,725 @@
...
+
+If a client sends a request without ``Accept`` header (or sending ``Accept`` header with a unsupported content
+type), it would be usually mapped to ``helloJson()`` method because the methods are sorted by the name of
+the type in an alphabetical order.
+In this case, you can adjust the order of the methods with `@Order`_ annotation. The default value of
```
trustin: How about inserting an empty line after `an alphabetical order.`? so that we get a paragraph break?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -239,11 +242,25 @@ annotation or ``Optional<?>`` class, e.g. ``hello2`` and ``hello3`` method in th
         }
     }
 
+If an HTTP POST request with a ``Content-Type: x-www-form-urlencoded`` is received and any `@Param`_ value
+is not shown in the path pattern, Armeria will aggregate the received request and decode its body as an
```
trustin: - any `@Param` value is now shown -> no `@Param` value appears
- an url-encoded -> a URL-encoded

site/src/sphinx/server-annotated-service.rst
```diff
@@ -239,11 +242,25 @@ annotation or ``Optional<?>`` class, e.g. ``hello2`` and ``hello3`` method in th
...
+            // a '400 Bad Request' because there is no way to inject a mandatory parameter 'name'.
+        }
+    }
+
 Getting an HTTP header
 ^^^^^^^^^^^^^^^^^^^^^^
 
 Armeria also provides `@Header`_ annotation to inject an HTTP header value into a parameter. The parameter
-annotated with `@Header`_ is also specified as one of the built-in types as follows. `@Default`_ and
+annotated with `@Header`_ can be also specified as one of the built-in types as follows. `@Default`_ and
```
trustin: can also be

site/src/sphinx/server-annotated-service.rst
```diff
@@ -318,7 +335,7 @@ If your exception handler is not able to handle a given exception, you can call
         }
     }
 
-To catch an exception from every method in your service object, annotate your class with the annotation.
+To catch an exception from every method in your service class, annotate your class with the annotation.
```

    You can annotate at class level to catch an exception from every method in your service class.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -328,7 +345,7 @@ To catch an exception from every method in your service object, annotate your cl
         public HttpResponse hello() { ... }
     }
 
-To catch an exception from a single method in your service object, annotate your method with the annotation.
+To catch an exception from a single method in your service class, annotate your method with the annotation.
```

    You can also annotate at method level to catch an exception from a single method in your service class.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -494,6 +510,30 @@ explained in the previous sections::
     sb.annotatedService(new MyAnnotatedServiceClass(),
                         new MyExceptionHandler(), new MyRequestConverter(), new MyResponseConverter());
 
+Also, they have a different method signature for conversion and exception handling so even you can write them
```
trustin: even you can write -> you can even write

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+.. _StringRequestConverterFunction: apidocs/index.html?com/linecorp/armeria/server/annotation/StringRequestConverterFunction.html
+
+.. _server-annotated-service:
+
+Annotated HTTP Service
+======================
+
+Armeria provides a way to write an HTTP service using annotations. It helps a user make his or her code
+simple and easy to understand. A user is able to run an HTTP service by fewer lines of code using
+annotations as follows. ``getHello()`` method in the example would be mapped to the path of ``/hello``
```
minwoox: getHello is hello?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+
+Getting an HTTP parameter
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When the value of `@Param`_ annotation is not shown in the path pattern, it will be handled as a parameter
+name of the query string of the request. If you have a service class like the example below and a user sends an
+HTTP GET request with URI of ``/hello1?name=armeria``, the service method will get ``armeria`` as the value
+of parameter ``name``. If there is no parameter named ``name`` in the query string, the parameter ``name``
+of the method would be ``null``. If you want to avoid ``null`` in this case, you can use `@Default`_
+annotation or ``Optional<?>`` class, e.g. ``hello2`` and ``hello3`` method in the example, respectively.
```
minwoox: how about 'methods below' since we use example twice here?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+        public HttpResponse hello1(@Header("Authorization") String auth) { ... }
+
+        @Post("/hello2")
+        public HttpResponse hello2(@Param("Content-Length") long contentLength) { ... }
+    }
+
+Other classes automatically injected
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The following classes are automatically injected when you specify it on the parameter list of your method.
```
minwoox: you specify them?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+
+.. code-block:: java
+
+    public class MyResponseConverter implements ResponseConverterFunction {
+        @Override
+        public HttpResponse convertResponse(ServiceRequestContext ctx, Object result) {
+            if (result == MyObject.class) {
+                return HttpResponse.of(HttpStatus.OK,
+                                       MediaType.PLAIN_TEXT_UTF_8,
+                                       "Hello, %s!", result.processedName());
```
minwoox: Don't we need a code to convert the `result` to `MyObject`?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+        @Get("/hello")
+        public HttpResponse hello1() {
+            // Return a text document to the client.
+            return HttpResponse.of(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "Armeria");
+        }
+
+        @Get("/hello")
+        public HttpResponse hello2() {
+            // Return a JSON object to the client.
+            return HttpResponse.of(HttpStatus.OK, MediaType.JSON_UTF8, "{ \"name\": \"Armeria\" }");
```
minwoox: JSON_UTF_8?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+        }
+    }
+
+If media type is not specified on any methods bound to the same path pattern, the first method declared will
```
minwoox: the media type?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+be used and the other methods will be ignored. In this example, ``hello1()`` will be chosen and the client
+will always receive a text document. What if you want to get a JSON object from the path ``/hello``?
+You can just specify the type of the content your method produces as follows and add an ``Accept`` header
```
minwoox: content which your method produces?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+        public HttpResponse helloText() {
+            // Return a text document to the client.
+            return HttpResponse.of(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "Armeria");
+        }
+
+        @Get("/hello")
+        @ProduceType("application/json")
+        public HttpResponse helloJson() {
+            // Return a JSON object to the client.
+            return HttpResponse.of(HttpStatus.OK, MediaType.JSON_UTF8, "{ \"name\": \"Armeria\" }");
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+    GET /hello HTTP/1.1
+    Accept: application/json
+
+.. note::
+
+    Note that a ``Content-Type`` header of a response is not automatically set. You may want to get the
+    negotiated `@ProduceType`_ from ``ServiceRequestContext.negotiatedProduceType()`` method and set it
+    as the value of the ``Content-Type`` header of your response.
+
+If a client sends a request without an ``Accept`` header (or sending an ``Accept`` header with a unsupported
```
minwoox: an unsupported?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+        public HttpResponse helloText() {
+            // Return a text document to the client.
+            return HttpResponse.of(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "Armeria");
+        }
+
+        @Get("/hello")
+        @ProduceType("application/json")
+        public HttpResponse helloJson() {
+            // Return a JSON object to the client.
+            return HttpResponse.of(HttpStatus.OK, MediaType.JSON_UTF8, "{ \"name\": \"Armeria\" }");
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -0,0 +1,767 @@
...
+    POST /hello HTTP/1.1
+    Content-Type: application/json
+    Content-Length: 21
+
+    { "name": "Armeria" }
+
+However, if a client sends a request with a ``Content-Type: application/octet-stream`` header which is not
+specified with `@ConsumeType`_, the client would get an HTTP status code of 415 which means
+``Unsupported Media Type``. If you want to make one of the methods catch-all, you can remove the annotation
+as follows. ``helloCatchAll()`` method would accept every request except for a request with a
```
minwoox: except for the request?

site/src/sphinx/server-annotated-service.rst
hyangtack: Copy&paste mistake. `result instanceof MyObject` is correct. Let me fix it.

trustin:
Thanks, @hyangtack !

PR #1012, opened 2018-02-20
Subscribe to pooled objects for client requests.

anuraaga:
I had originally done this for responses, but not for requests, probably assuming that requests tend to be small compared to responses. Now I'm writing an object storage client where requests are orders of magnitude larger than responses...


hyangtack:
Is it OK if we don't release `o` when `write()` did nothing? e.g. channel inactive, illegal state..

anuraaga:
Thanks for noticing, yeah missed those branches (also added to `HttpResponseSubscriber` for consistency)

anuraaga:
Actually I went ahead and removed some of the checks, it seems impossible to execute that code

trustin:
Thanks, @anuraaga !

PR #1016, opened 2018-02-21
Fix NPE handling headers if they are split between windows

daveconde:
This fixes an issue where headers can arrive over 2 windows thus causing a NPE


trustin:
I never knew headers can be split into two HEADERS frame. Would you mind sharing all `HttpHeaders` objects you have received from Linkerd so I understand the problem better?

trustin:
Also, could you sign the ICLA? https://feedback.line.me/enquete/public/799-Z6ygVz3U Thanks!

daveconde:
Unfortunately I dont have a TCP dump handy to check the packets but the behaviour I'm seeing is that a first call comes in to onNext with the first group of headers including status etc, this is then followed by a second call that contains a single header called l5d-success-class. This then causes the NPE. 

Thanks I'll get that signed for you.

trustin:
Closing in favor of #1024 Thanks for the inspiration!

Issue #1014, opened 2018-02-20
H2 with Linkerd


daveconde:
I am trying to get retrofit to proxy via Linkerd using h2c. I need to be able to override the authority header as mentioned [here](https://linkerd.io/getting-started/k8s/) while still keeping the URL pointing at the linkerd host and port. How do I set this using the armeria client?

Thanks

trustin:
You can override the headers sent by an Armeria HTTP client as explained here: https://line.github.io/armeria/client-custom-http-headers.html

You could override the ':authority' header using one of the 3 ways explained? Please let me know if it does not work so we can fix it.

I'm also curious about what you think about the current way of overriding headers. Can we do better in terms of simplicity and user-friendliness? Any suggestions for better API?

trustin:
More specifically, you could use 'ArmeriaRetrofitBuilder.withClientOptions()' to set the ClientOption.HTTP_HEADERS option:

withClientOptions((uri, builder) -> builder.setHttpHeader(HttpHeaderNames.AUTHORITY, "..."))

daveconde:
I attempted the below with a decorator but it seems the authority is being set from the URL instead of the header override.

`retrofit = new ArmeriaRetrofitBuilder()
                    .baseUrl("h2c://" + host)
                    .addConverterFactory(JacksonConverterFactory.create())
                    .withClientOptions((url, optionsBuilder) -> {
                        optionsBuilder.decorator(HttpRequest.class, HttpResponse.class, (delegate, ctx, req) -> {
                            req.headers().set(AUTHORITY, adStoreApiHost + "." + namespace);
                            return delegate.execute(ctx, req);
                        });
                        return optionsBuilder;
                    })
                    .build();`

I also attempted the below but then got an exception around authority being blacklisted.

` optionsBuilder.httpHeader(HEADER_NAME, value);`

Is there another method you would recommend?

Thanks for the help :)

trustin:
I guess :authority header cannot be overridden in current implementation but it's an unnecessary restriction we enforced. Let me whitelist :authority in the next release, which should be out this week.

Thanks!

daveconde:
As a workaround I have configured linkerd to look for another header but as far as I can see any extra headers are removed before the call goes over the wire.

trustin:
That's weird. Could you share your construction code so I can reproduce?

daveconde:
[...]

PR #1017, opened 2018-02-22
Allow overriding ':authority' header

trustin:
Motivation:

Some L7 reverse proxies such as Linkerd requires a user to send a
different ':authority' header than the actual host name.

Modification:

- Do not blacklist ':authority' header from ClientOptions.HTTP_HEADERS.

Result:

- Partial fix for #1014

/cc @daveconde

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
@@ -68,11 +68,10 @@ import com.linecorp.armeria.unsafe.ByteBufHttpData;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
+import io.netty.util.AsciiString;
 import io.netty.util.ReferenceCountUtil;
 
 public class HttpClientIntegrationTest {
-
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -372,33 +379,56 @@ private static void testDoubleSlashSuppression(String path, String normalizedPat
...
+        assertEquals(headerValue, response.content().toStringUtf8());
     }
 
-    @Test
-    public void testUserAgentOverridableByRequestHeader() throws Exception {
-
-        HttpHeaders headers = HttpHeaders.of(HttpHeaderNames.USER_AGENT, TEST_USER_AGENT_NAME);
+    private static void testHeaderOverridableByRequestHeader(String path, AsciiString headerName,
+                                                             String headerValue) throws Exception {
+        HttpHeaders headers = HttpHeaders.of(HttpHeaderNames.USER_AGENT, headerValue);
```
minwoox: `headerName` instead of `HttpHeaderNames.USER_AGENT`?

trustin:
Thanks, reviewers!

rukkus:
Sure I’ll give it a try earlier next week, thanks for the fast turnaround

PR #1018, opened 2018-02-22
Allow session protocol to be specified for HttpHealthCheckedEndpointG…

anuraaga:
…roup.

As there are many parameters now, added a builder instead of extending the static factories.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
@@ -35,8 +35,8 @@ public class HttpHealthCheckedEndpointGroupBuilder {
     private final EndpointGroup delegate;
     private final String healthCheckPath;
 
-    private SessionProtocol healthCheckProtocol = SessionProtocol.HTTP;
-    private Duration healthCheckRetryInterval = DEFAULT_HEALTHCHECK_RETRY_INTERVAL;
+    private SessionProtocol protocol = SessionProtocol.HTTP;
+    private Duration retryInterval = DEFAULT_HEALTHCHECK_RETRY_INTERVAL;
     private ClientFactory clientFactory = ClientFactory.DEFAULT;
 
     /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public HttpHealthCheckedEndpointGroupBuilder(EndpointGroup delegate, String healthCheckPath) {
+        this.delegate = requireNonNull(delegate, "delegate");
+        this.healthCheckPath = requireNonNull(healthCheckPath, "healthCheckPath");
+    }
+
+    /**
+     * Sets the {@link SessionProtocol} to be used when making health check requests.
+     */
+    public HttpHealthCheckedEndpointGroupBuilder healthCheckProtocol(SessionProtocol healthCheckProtocol) {
```
trustin: Could remove `healthCheck` prefix?

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+        this.healthCheckProtocol = requireNonNull(healthCheckProtocol, "healthCheckProtocol");
+        return this;
+    }
+
+    /**
+     * Sets the interval between health check requests. Must be positive.
+     */
+    public HttpHealthCheckedEndpointGroupBuilder healthCheckRetryInterval(Duration healthCheckRetryInterval) {
```
trustin: Could remove `healthCheck` prefix?

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+        requireNonNull(healthCheckRetryInterval, "healthCheckRetryInterval");
+        checkArgument(!healthCheckRetryInterval.isNegative() && !healthCheckRetryInterval.isZero(),
+                      "healthCheckRetryInterval must be positive.");
```
trustin: `"retryInterval: %s (expected > 0)", retryInterval`?

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
anuraaga: Good idea, updated

testing/src/main/java/com/linecorp/armeria/testing/server/ServerRule.java
```diff
@@ -156,10 +153,20 @@ public int httpPort() {
...
      */
     public int httpsPort() {
+        return port(HTTPS);
+    }
+
+    /**
+     * Returns the port number of the {@link Server} for the specified {@link SessionProtocol}.
+     *
+     * @throws IllegalStateException if the {@link Server} is not started or it did not open a port of the
+     *     specified protocol.
```
minwoox: indentation?

anuraaga: This is the indentation checkstyle told me to use.

trustin: Could look better if indentet at the same column with `if the..`

anuraaga: Ah didn't realize the checkstyle was at least 4, not exactly 4 - fixed

trustin:
Thanks, @anuraaga !

PR #1019, opened 2018-02-22
Set previously health-checked endpoints that don't exist anymore to f…

anuraaga:
…alse instead of remaining true.

While it might be possible to encode an enum, like `-1` indicates removed, I don't know if it's worth it. But showing a removed endpoint as healthy seems like a confusing default.


trustin:
Thanks, @anuraaga

PR #1021, opened 2018-02-22
Add Nullable to RequestContext.mapCurrent

anuraaga:
Not using kotlin, but IntelliJ still provides good warnings now that we have the package annotation :)


trustin:
Thanks, @anuraaga !

PR #1024, opened 2018-02-23
Handle split headers in ArmeriaCallSubscriber

### LINKED ISSUE #1014: H2 with Linkerd

trustin:
Motivation:

Some HTTP/2 implementations can send additional headers after sending a
non-informational HEADERS frame.

ArmeriaCallSubscriber will fail with NPE because it assumes HttpHeaders
always have a `:status` header, but it's not the case of split headers.

Modifications:

- Append the HTTP headers received after the first non-informational
  HTTP headers
- Defer the call to `Response.Builder.body()` so that we can handle
  `content-type` or `content-length` header even if it is not in the
  first HEADERS frame.

Result:

- Fixes #1014 

/cc @daveconde

### DIFF
--- a/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallSubscriber.java
+++ b/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallSubscriber.java
@@ -17,6 +17,8 @@ package com.linecorp.armeria.client.retrofit2;
 
 import java.io.IOException;
 
+import javax.annotation.Nullable;
+
 import org.reactivestreams.Subscriber;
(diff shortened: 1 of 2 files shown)


trustin:
Thanks, reviewers. Let me send a PR for `HttpResponseAggregator` soon.

trustin:
Follow-up PR for HttpRequestAggregator and HttpResponseAggregator: https://github.com/line/armeria/pull/1027

Issue #1009, opened 2018-02-13
Expose netty channel option to end-user


huydx:
For some high performance network applications, users may need to tune those options
SO_REUSEADDR for example

imasahiro:
We may want to introduce server/client netty option like lettuce's [NettyCustomizer](https://github.com/lettuce-io/lettuce-core/blob/9b7b318c134ed7ae7993f9d1030b38f4653dd24a/src/main/java/com/lambdaworks/redis/resource/NettyCustomizer.java)?

trustin:
I guess that will increase the future difficulty of shading Netty. It'd be fine if choose not to shade Netty, though :-)

imasahiro:
Realized that we already exposing API that customize socket options at `ClientFactoryBuilder` but didn't expose at `ServerBuilder`. How about just exposing `ChannelOption` instead of adding general netty customizer?

rkapsi:
+1 I'm very interested in this feature and happy to PR it

trustin:
We must decide we should shade Netty or not first. If we do, we must hide the references to the Netty classes from our public API. Otherwise, we can expose Netty classes in our public API. How we should tackle this issue changes vastly depending on this. Let me create a poll.

trustin:
Created a poll - https://github.com/line/armeria/issues/1168

PR #1029, opened 2018-02-23
Expose netty configuration using NettyServerCustomizer and NettyClien…

huydx:
…tCustomizer

Solve https://github.com/line/armeria/issues/1009
just POC first to discuss. Wil add NettyClientCustomizer later.

Some concern:
- As @trustin said, it may make it hard to shade netty later, I'm not sure but is that user still use shaded Netty's ChannelOption to config, and the only disadvantage is that they could not share common config only?
- Added method to expose `ServerBootstrap` for testing , seems not beautiful but I could not find any other way.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/NettyServerCustomizer.java
+++ b/core/src/main/java/com/linecorp/armeria/common/NettyServerCustomizer.java
@@ -15,19 +15,15 @@
  */
 package com.linecorp.armeria.common;
 
+import java.util.function.Consumer;
+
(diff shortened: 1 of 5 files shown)


trustin:
I believe this will increase the amount of our future effort to shade Netty. Perhaps we should provide dedicated API for this?

huydx:
@trustin we decided to not shade netty, so I guess I could continue this issue. Let me close this PR once (it has many conflicts) and recreate it later.

PR #1030, opened 2018-02-24
Close in-progress responses when channel is closed.

### LINKED ISSUE #1028: Zombie gRPC connections
When using constructs of the type:
```
rpc StreamClientCancels2(SimpleRequest) returns (stream SimpleResponse);
```
there is no way to know if the client has disconnected. OnCancelHandler is not called, nor is there any exception thrown when sending more data. It will just silently discard all data sent.

I've added a test case (CC0 on my changes, claim ownership if you want)
https://github.com/tobias-/armeria/tree/zombie_connections

anuraaga:
This ended up being more complicated than I expected. Please check that my understanding makes sense :)

Currently, there is no notification to response streams if the client disconnects. If the client disconnects before finishing a request, server logic will be notified in the request stream, but if it disconnects after finshing a request but before finishing a response, the server logic has no way of knowing to abort processing.

- Now, any in-progress response streams are aborted when channel is closed
- Default behavior of `StreamMessage.write` changed to throw an exception when stream is closed, which is more natural. `tryWrite` can be used when this is not desired.
- Also fix that a client close on HTTP/1 request did not notify the request stream of closure
- Fixes a test which kept a stream in a static variable and writing consecutively to it

Fixes #1028

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
@@ -37,12 +37,37 @@ public interface StreamWriter<T> {
      * Writes the specified object to the {@link StreamMessage}. The written object will be transferred to the
      * {@link Subscriber}.
      *
+     * @throws AbortedStreamException if the stream was already closed
+     * @throws IllegalArgumentException if the publication of the specified object has been rejected
+     */
+    default void write(T o) {
+        if (!tryWrite(o)) {
+            throw AbortedStreamException.get();
+        }
+    }
+
+    /**
+     * Writes the specified object {@link Supplier} to the {@link StreamMessage}. The object provided by the
+     * {@link Supplier} will be transferred to the {@link Subscriber}.
+     *
+     * @throws AbortedStreamException if the stream was already closed.
+     */
+    default void write(Supplier<? extends T> o) {
+        if (!tryWrite(o)) {
+            throw AbortedStreamException.get();
+        }
+    }
+
+    /**
+     * Writes the specified object to the {@link StreamMessage}. The written object will be transferred to the
+     * {@link Subscriber}.
+     *
      * @return {@code true} if the specified object has been scheduled for publication. {@code false} if the
      *         stream has been closed already.
      *
      * @throws IllegalArgumentException if the publication of the specified object has been rejected
      */
-    boolean write(T o);
+    boolean tryWrite(T o);
 
     /**
      * Writes the specified object {@link Supplier} to the {@link StreamMessage}. The object provided by the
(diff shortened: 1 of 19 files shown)


tobias-:
Just to confirm, yes, I've tested with this patch, and it solves the main issue with messages sent over gRPC being silently dropped.

core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
```diff
@@ -33,6 +35,31 @@
...
+    /**
+     * Writes the specified object to the {@link StreamMessage}. The written object will be transferred to the
+     * {@link Subscriber}.
+     *
+     * @throws AbortedStreamException if the stream was already closed
+     * @throws IllegalArgumentException if the publication of the specified object has been rejected
+     */
+    default void write(T o) {
+        if (!tryWrite(o)) {
+            throw AbortedStreamException.get();
```
trustin: How about using `new IllegalStateException("stream closed")`? `AbortedStreamException` is specifically for the notification of `abort()`.

anuraaga: Sure - done

core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
```diff
@@ -33,6 +35,31 @@
...
+        if (!tryWrite(o)) {
+            throw new IllegalStateException("stream closed");
+        }
+    }
+
+    /**
+     * Writes the specified object {@link Supplier} to the {@link StreamMessage}. The object provided by the
+     * {@link Supplier} will be transferred to the {@link Subscriber}.
+     *
+     * @throws AbortedStreamException if the stream was already closed.
```
trustin: You need to update Javadoc as well. :-)

anuraaga: Oops

core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
```diff
@@ -33,6 +35,32 @@
      */
     boolean isOpen();
 
+    /**
+     * Writes the specified object to the {@link StreamMessage}. The written object will be transferred to the
+     * {@link Subscriber}.
+     *
+     * @throws IllegalArgumentException if the stream was already closed or the publication of the specified
+     *         object has been rejected
```
trustin: `IllegalStateException` is raised when stream was closed.

anuraaga: Urgh - too much coffee can't read

trustin: covfefe

trustin:
Oops. Build failure?

trustin:
Another failure:

```
com.linecorp.armeria.server.grpc.GrpcServiceServerTest > clientSocketClosedAfterHalfCloseBeforeCloseHttp1 FAILED
    java.lang.IllegalStateException: stream closed
com.linecorp.armeria.server.grpc.GrpcServiceServerTest > clientSocketClosedAfterHalfCloseBeforeCloseHttp2 FAILED
    java.lang.IllegalStateException: stream closed
```

trustin:
Thanks, @anuraaga!

PR #1031, opened 2018-02-24
Make sure RPC information is set to metrics for unframed requests.

### LINKED ISSUE #1000: `MeterIdPrefixFunction.ofDefault` doesn't collect service level metrics for unframed request
Please refer to https://gist.github.com/openaphid/63ee584a4b7bb4c646137ff5eea92dea for a reproducible case.

A custom `MeterIdPrefixFunction` could extract method name from `ServiceRequestContext` correctly, but I think there are some execution order issues in `UnframedGrpcService` and `GrpcService`.

For unframed requests, `requestContent` is not attached to context log:
https://github.com/line/armeria/blob/d3334ce5c25e9ecb45b0c94c54b9b99f2fb0e9ea/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java#L143

anuraaga:
This seems a bit hacky, but couldn't think of a cleaner approach.

Fixes #1000


trustin:
Thanks, @anuraaga !

PR #1032, opened 2018-02-24
Migrate remaining usage of hamcrest to assertj

anuraaga:


trustin:
Thank you so much, @anuraaga !

PR #1036, opened 2018-02-27
Make closeSpan method static

huydx:
Trivial fix, it will save a very small memory usage per HttpTracingService

### DIFF
--- a/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
+++ b/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
@@ -17,6 +17,7 @@
 package com.linecorp.armeria.internal.tracing;
 
 import com.linecorp.armeria.common.RequestContext;
+import com.linecorp.armeria.common.logging.RequestLog;
 
 import brave.Span;
 import brave.Tracer;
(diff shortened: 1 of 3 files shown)


anuraaga:
While it's ok to change to static, there should be no difference in runtime behavior since it's a private method, and not a field - methods are resolved through a virtual function table on the class, not instance, and JVM recognizes private methods and optimizes their call to be equivalent to static where possible. 

Why do you think there'll be a memory benefit?

huydx:
@anuraaga I see, thanks for the information. Did not know that JVM already optimize for private method. Do you think this method is better being utility method? Just a trivial fix so if you think it's not worth it, feel free to close it.

anuraaga:
I think it's fine to move to the util, unrelated to performance. Do you mind replacing the two lines in `HttpTracingClient` with this utility as well? :D

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
```diff
@@ -43,5 +44,13 @@ public static void setupContext(FastThreadLocal<SpanInScope> threadLocalSpan, Re
         });
     }
 
+    /**
+     * Closes a {@link Span} which related to a {@link RequestLog}.
```
anuraaga: Adds logging tags to the provided {@link Span} and closes it. The span cannot be used further after this method has been called.

huydx:
@anuraaga PTAL~

trustin:
Thanks, @huydx !

PR #1038, opened 2018-03-01
Better document / test reference counting of stream duplication.

anuraaga:
While the queue releases references fine, it doesn't retain the first one, so it will not be able to do signal the object a second time as it has likely already been released. The unit test forgot about the reference the test itself holds to the buffers.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
@@ -821,6 +821,7 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
             elements = new Object[16];
         }
 
+        // Steals a reference to o.
         int addAndRemoveIfRequested(Object o) {
             requireNonNull(o);
             int removedLength = 0;
@@ -836,6 +837,7 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
             return removedLength;
         }
 
+        // Releases references to removed elements.
         private int removeElements() {
             final int removalRequestedOffset = lastRemovalRequestedOffset;
             final int numElementsToBeRemoved = removalRequestedOffset - headOffset;
@@ -883,6 +885,7 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
             this.elements = a;
         }
 
+        // Steals a reference to the object.
         Object get(int offset) {
             final int head = this.head;
             final int tail = this.tail;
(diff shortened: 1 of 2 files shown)


minwoox:
@anuraaga Thank you for checking this :)

But I think we don't need to retain the object there.
In the unit test, we used `duplicateStream(true)` which `true` indicates that this is the last stream who subscribes the elements.
We do know that this is the last stream, so that we don't have to keep the reference of subscribed objects.
To release the elements, we count the number of subscribed elements to each subscriber. When the sum reaches 50, we release the elements. e.g) If there are 5 subscribers, every 10 elements(5 * 10 = 50) are released when the elements are subscribed ideally. 
Sorry for making you confused about it.

To retain the reference of the elements which are subscribed by the subscribers who called `subscribe(withPooledObject /* true */)`, we just retain it before we hand over.
```java
if (withPooledObjects) {
    if (obj instanceof ByteBufHolder) {
        obj = retainedDuplicate((ByteBufHolder) obj);
    } else if (obj instanceof ByteBuf) {
        obj = retainedDuplicate((ByteBuf) obj);
    }
} else {
    if (obj instanceof ByteBufHolder) {
        obj = copy((ByteBufHolder) obj);
    } else if (obj instanceof ByteBuf) {
        obj = copy((ByteBuf) obj);
    }
}
```

anuraaga:
I don't think it matters what is the last stream or not - the signal queue itself holds a reference to the objects and thus must retain it. It releases whenever it loses a reference, so not retaining is clearly wrong. When trying to use in production, I see exceptions due to the signal queue releasing an object with refcnt.

I would just look at the references and nothing else. Anytime you do x = pooled, you need to retain a reference and before doing x = null, you release it. In the unit tests, but = pooled has one reference by default, if the stream's signal queue has another than 2 is clearly the correct count.

minwoox:
> I don't think it matters what is the last stream or not

The reason we introduced the last stream is that when the number of elements are growing bigger, we worried about all the buffers that the elements occupy. Because the duplicator has to keep all the elements in it by its nature, if we do not release it, it can be too much burden for the system.
So we decided to bring the concept of the last stream here. If we know that the last stream is already appended to the duplicator, we can release the elements which are already subscribed.(we are sure that nobody is going the need it.)

We thought that after the moment the `StreamMeassgeDuplicator` subscribes the original publisher with `subscribe(withPooledObject /* true */)`, the ownership of the reference in the original publisher is moved to the `StreamMeassgeDuplicator`. The original publisher cannot be subscribed anymore, So the reference count is 1 which `StreamMeassgeDuplicator` holds.

anuraaga:
Ah yeah you're right, the stream always steals the reference to pooled objects. I'll have to dig more into why I saw the refcnt exception that I posted to the other issue, for now I reworked this to just clarify comments / the test behavior (stream releases the references instead of buf, since stream owns them).

minwoox:
@anuraaga Thank you for clarifying this~!
Please let me know when you see those sort of weird exceptions. I might be a little help  :)

trustin:
Thanks, @anuraaga !

Issue #1041, opened 2018-03-01
Not to use `Deplicated` methods in the docments


kishida:
😭 
![image](https://user-images.githubusercontent.com/1178746/36855658-6a2e65f2-1db7-11e8-92f7-8e1df9a795ce.png)

PR #999, opened 2018-02-08
Typo fix to 'server basics' doc.

krisjey:
Removed non-existing selection annotations.
Fixing code errors (parentheses and braces).
Adding a missing import statement.

PR #1043, opened 2018-03-01
Do not use deprecated methods in documentation

kishida:
Not to use `respond` method.
Use lambda if it can.
~~Use `:name` style instead of `{name}` style.~~
Use `:name` style to indicate that Armeria allows it.
Fix the lack of braces.
https://github.com/line/armeria/issues/1041

Including some of https://github.com/line/armeria/pull/999
Should I include `import`?

### DIFF
--- a/site/src/sphinx/server-basics.rst
+++ b/site/src/sphinx/server-basics.rst
@@ -45,53 +45,54 @@ Even if we opened a port, it's of no use if we didn't bind any services to them.
 
 .. code-block:: java
 
+    import com.linecorp.armeria.common.HttpParameters;
     import com.linecorp.armeria.common.HttpRequest;
-    import com.linecorp.armeria.common.HttpResponseWriter;
+    import com.linecorp.armeria.common.HttpResponse;
     import com.linecorp.armeria.common.HttpStatus;
     import com.linecorp.armeria.common.MediaType;
+    import com.linecorp.armeria.common.SessionProtocol;
 
-    import com.linecorp.armeria.server.ServiceRequestContext;
     import com.linecorp.armeria.server.AbstractHttpService;
+    import com.linecorp.armeria.server.Server;
+    import com.linecorp.armeria.server.ServerBuilder;
+    import com.linecorp.armeria.server.ServiceRequestContext;
+    import com.linecorp.armeria.server.annotation.ConsumeType;
+    import com.linecorp.armeria.server.annotation.Default;
+    import com.linecorp.armeria.server.annotation.Get;
+    import com.linecorp.armeria.server.annotation.Param;
+    import com.linecorp.armeria.server.annotation.Post;
+    import com.linecorp.armeria.server.annotation.ProduceType;
+    import com.linecorp.armeria.server.logging.LoggingService;
 
     ServerBuilder sb = new ServerBuilder();
     sb.port(8080, HTTP);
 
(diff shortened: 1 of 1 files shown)


site/src/sphinx/server-basics.rst
```diff
@@ -106,7 +96,7 @@ Even if we opened a port, it's of no use if we didn't bind any services to them.
     sb.annotatedService(new Object() {
         @Get("/greet5")
         public HttpResponse greet(@Param("name") String name,
-                                  @Param("title") @Optional("Mr.") String title) {
+                                  @Param("title") /*@Optional("Mr.")*/ String title) {
```
kishida: Where is the `@Optional`?

trustin: It is now `@Default`.

trustin:
> Should I include import?

Only when a class appears first time in the page. No need to import JDK classes.

trustin:
Thanks a lot for fixing the documentation, @kishida!

kishida:
Update～

site/src/sphinx/server-basics.rst
```diff
@@ -45,53 +45,56 @@ Even if we opened a port, it's of no use if we didn't bind any services to them.
 
 .. code-block:: java
 
+    import com.linecorp.armeria.server.ServiceRequestContext;
+    import com.linecorp.armeria.server.AbstractHttpService;
```
trustin: Could you sort the imports?

kishida: Fixed 😃

hyangtack:
@kishida Oops! I recently added [server-annotated-service.rst](https://github.com/line/armeria/blob/master/site/src/sphinx/server-annotated-service.rst) and it has been merged. I used `{name}` style path variables In that document. Do we better change them to `:name` style? WDYT?

kishida:
@hyangtack I've changed to introduce the both style. 😄

kishida:
Talking with Trustin-san on Slack, `{name}` style may be better since JAX-RS (and Spring MVC) are using it.

trustin:
@kishida Thanks a lot!

site/src/sphinx/server-basics.rst
```diff
@@ -45,53 +45,54 @@ Even if we opened a port, it's of no use if we didn't bind any services to them.
...
-            res.respond(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "Hello, %s!", name);
+            return HttpResponse.of(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "Hello, %s!", name);
         }
-    }.decorate(LoggingService.newDecorator()); // Enable logging
+    }.decorate(LoggingService.newDecorator())); // Enable logging
 
     // Using an annotated service object:
     sb.annotatedService(new Object() {
-        @Get("/greet2/{name}")
+        @Get("/greet2/:name") // `:name` style is also available
```
minwoox: Didn't change it back to `{name}` :)

kishida: Only here, I want to indicate that Armeria allows `:name` style.

trustin:
Thanks a lot, @kishida !

kishida:
I'm very happy to be able to commit Armeria!

PR #1047, opened 2018-03-05
Cancel gRPC stream when client cancels.

anuraaga:
Currently, a server would know the client cancelled when they attempt to write to the closed stream. Now, they can listen for the cancellation itself.

### DIFF
--- a/grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
+++ b/grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
@@ -242,10 +242,17 @@ public class GrpcServiceServerTest {
                 await().until(CLIENT_CLOSED::get);
                 try {
                     responseObserver.onNext(SimpleResponse.getDefaultInstance());
-                } catch (IllegalStateException e) {
+                } catch (Throwable t) {
                     COMPLETED.set(true);
                 }
             });
+        }
+
+        @Override
+        public void streamClientCancelsBeforeResponseClosedCancels(
+                SimpleRequest request, StreamObserver<SimpleResponse> responseObserver) {
+            ((ServerCallStreamObserver<?>) responseObserver).setOnCancelHandler(() -> COMPLETED.set(true));
+            responseObserver.onNext(SimpleResponse.getDefaultInstance());
         }
     }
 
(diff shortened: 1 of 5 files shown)


grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
```diff
@@ -506,6 +513,45 @@ public void onCompleted() {
...
+
+    @Test
+    public void clientSocketClosedAfterHalfCloseBeforeCloseCancelsHttp1() throws Exception {
+        clientSocketClosedAfterHalfCloseBeforeCloseCancels("h1c");
+    }
+
+    private void clientSocketClosedAfterHalfCloseBeforeCloseCancels(String protocol) {
+        ClientFactory factory = new ClientFactoryBuilder().build();
+        UnitTestServiceStub stub =
+                new ClientBuilder("gproto+" + protocol + "://127.0.0.1:" + server.httpPort() + "/")
```
trustin: Could use `server.httpUri()`?

trustin:
@anuraaga Would this a bug or an improvement?

anuraaga:
I guess bug, even though there's a workaround - the cancel should reach the stream observer correctly. Also added a method on `ServerRule` and used it.

testing/src/main/java/com/linecorp/armeria/testing/server/ServerRule.java
```diff
@@ -211,6 +211,28 @@ public String uri(String path) {
...
 
+    /**
+     * Returns the URI for the {@link Server} of the specified protocol and format.
+     *
+     * @throws IllegalStateException if the {@link Server} is not started or
+     *                               it did not open a port of the protocol.
+     */
+    public String uri(SessionProtocol protocol, SerializationFormat format, String path) {
+        // This will ensure that the server has started.
+        server();
```
trustin: Could you add `requireNonNull(...)` on `protocol`, `format` and `path`?

trustin:
@anuraaga I didn't realize that a new method had to be added to `ServerRule`. Thanks!

testing/src/main/java/com/linecorp/armeria/testing/server/ServerRule.java
trustin: Thanks!

trustin:
Thanks, @anuraaga !

PR #1049, opened 2018-03-05
Restructure spring-boot package to support both spring-boot 1 and 2

imasahiro:


imasahiro:
This PR is still in progress because I don't test with spring-boot2 yet (And it does not pass `spring:boot:autoconfigure` because it fails to load thrift generated code somehow). But I think it is enough to discuss about direction of the changes.

trustin:
The new structure looks great to me. Please proceed!

dependencies.yml
```diff
@@ -184,7 +184,7 @@ net.javacrumbs.json-unit:
   json-unit-fluent: { version: *JSON_UNIT_VERSION }
 
 net.sf.proguard:
-  proguard-gradle: { version: '5.3.3' }
+  proguard-gradle: { version: '6.0.1' }
```
imasahiro: Without upgrade, failed to execute trimShadedJar with following error.
```
Execution failed for task ':core:trimShadedJar'.
> java.io.IOException: Can't read [/home/travis/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-api/2.10.0/fec5797a55b786184a537abd39c3fa1449d752d6/log4j-api-2.10.0.jar] (Can't process class [META-INF/versions/9/org/apache/logging/log4j/util/ProcessIdUtil.class] (Unsupported class version number [53.0] (maximum 52.0, Java 1.8)))
```

settings.gradle
```diff
@@ -6,21 +6,23 @@ apply from: "${rootDir}/gradle/scripts/settings-flags.gradle"
...
-includeWithFlags ':tomcat8.0',                 'java', 'publish', 'relocate', 'no_aggregation'
-includeWithFlags ':zipkin',                    'java', 'publish', 'relocate'
-includeWithFlags ':zookeeper',                 'java', 'publish', 'relocate'
+includeWithFlags ':core',                       'java', 'publish', 'shade', 'trim'
+includeWithFlags ':grpc',                       'java', 'publish', 'relocate'
+includeWithFlags ':jetty',                      'java', 'publish', 'relocate'
+includeWithFlags ':kafka',                      'java', 'publish', 'relocate'
+includeWithFlags ':logback',                    'java', 'publish', 'relocate'
+includeWithFlags ':retrofit2',                  'java', 'publish', 'relocate'
+includeWithFlags ':spring:boot-autoconfigure',  'java', 'publish', 'relocate'
```
imasahiro: I cannot find/remember the reference but because gradle cannot handle the projects that have same project name, need to rename a project using e.g. `project(':spring:boot:autoconfigure').name = 'boot-autoconfigure'`. To make less gradle configuration, changed directory structure to flatten spring-boot and spring-boot2 projects to make each project name unique.
(hope @anuraaga remember about such kind of gradle issue)

anuraaga: I feel as if that issue should have been resolved by now, though I have seen plugins that don't use a project path and therefore can fail (gradle-protobuf-plugin) . Just wondering, did you try it and it didn't work? I guess it might work with Gradle, but not IntelliJ, which is why this sort of hack could be used instead

https://github.com/curioswitch/curiostack/blob/master/tools/gradle-plugins/gradle-curiostack-plugin/src/main/java/org/curioswitch/gradle/plugins/curiostack/CuriostackPlugin.java#L374

Flattening seems fine to me too though

imasahiro: I see. Thanks for explanation. Actually, I tried but looks our thrift gradle didn't generate correctly with Gradle.

trustin:
Thanks, @imasahiro !

spring/boot-autoconfigure/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+
+    testCompile 'org.springframework.boot:spring-boot-starter-test:1.5.10.RELEASE'
+    testCompile 'org.hibernate.validator:hibernate-validator'
+}
+
+// Use the sources from 'armeria-spring-boot2-autoconfigure'.
+// NB: We should not add these in the 'sourceSets' directive because that will make the directories mentioned
+//     below are added to two projects and having a source directory with two output directories will confuse
+//     the IDE.
+def autoconfigureProjectDir = "${rootProject.projectDir}/spring/boot2/autoconfigure"
```
imasahiro: oops, this should be `"${rootProject.projectDir}/spring/boot2-autoconfigure"`

PR #1051, opened 2018-03-06
More concise port and TLS configuration

### LINKED ISSUE #1050: ServerBuilder.port() raises an exception when H1/H1C/H2/H2C is specified
From Slack:

> yashinovsky: I have a question about Armeria and HTTP2. I tried passing H2 in the ServerBuilder set port, but it throws an exception since it only accepts HTTP or HTTPS
> trustin: You can just specify HTTPS then it will support both H1 and H2. (and HTTP for both H1C and H2C) There's no way currently to disable only HTTP/1 or HTTP/2. Do you need to do that?

It seems like some users are confused about this behavior. For disambiguation, we could replace `port()` with `http()` and `https()`, with some deprecation phase. What do you think?

trustin:
Motivation:

- A user currently has to specify SessionProtocol when adding a port,
  which increases the chance of specifying incorrect SessionProtocol.
- sslContext() builder methods require SessionProtocol which is
  practically unncessary.
- It would be confusing if Server serves plaintext HTTP when a user
  configured TLS even if he or she did not specify a port. More sensible
  behavior would be serving HTTPS when no port is specified in this case.

Modifications:

- Add http() and https() which supercede most port() methods.
- Add tls() which supercede sslContext() methods.
- Serve HTTPS by default when a user called tls() and no port was
  specified.
- Add log messages that indicates the protocol and local address.
- Deprecate the old methods.
- Update documentation.
- Miscellaneous:
  - Remove redundant `port(0, HTTP)` calls
  - Add `@Nullable` where necessary

Result:

- Fixes #1050
- More user-friendliness

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Server.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Server.java
@@ -64,7 +64,6 @@ import io.netty.channel.EventLoop;
 import io.netty.channel.EventLoopGroup;
 import io.netty.handler.ssl.SslContext;
 import io.netty.util.DomainNameMapping;
-import io.netty.util.DomainNameMappingBuilder;
 import io.netty.util.concurrent.FastThreadLocalThread;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GlobalEventExecutor;
@@ -80,6 +79,7 @@ public final class Server implements AutoCloseable {
     private static final Logger logger = LoggerFactory.getLogger(Server.class);
 
     private final ServerConfig config;
+    @Nullable
     private final DomainNameMapping<SslContext> sslContexts;
 
     private final StateManager stateManager = new StateManager();
@@ -92,6 +92,7 @@ public final class Server implements AutoCloseable {
 
     private final ConnectionLimitingHandler connectionLimitingHandler;
 
+    @Nullable
     private volatile ServerPort primaryActivePort;
 
     /**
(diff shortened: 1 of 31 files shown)


core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -369,19 +440,65 @@ public ServerBuilder accessLogWriter(Consumer<? super RequestLog> accessLogWrite
...
+
+    /**
+     * Sets the {@link SslContext} of the default {@link VirtualHost} from the specified
+     * {@code keyCertChainFile}, {@code keyFile} and {@code keyPassword}.
+     *
+     * @throws IllegalStateException if the default {@link VirtualHost} has been set via
+     *                               {@link #defaultVirtualHost(VirtualHost)} already
+     */
+    public ServerBuilder tls(
+            File keyCertChainFile, File keyFile, String keyPassword) throws SSLException {
```
hyangtack: `@Nullable` on `keyPassword`?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -119,13 +123,71 @@
...
+     * @see #http(int)
+     */
+    public ServerBuilder http(InetSocketAddress localAddress) {
+        ports.add(new ServerPort(requireNonNull(localAddress, "localAddress"), HTTP));
+        return this;
+    }
+
+    /**
+     * Adds an HTTPS port that listens on all available network interfaces. If no HTTPS port is added,
+     * (i.e. no {@code https()} method is called), HTTPS support is disabled.
```
hyangtack: Just curious. What if a user called `tls()` method? I'm not sure but what's better behavior? Listening on randomly-assigned HTTPS port or disabling HTTPS or raising an exception? (Actually the current behavior is also OK to me)

trustin: That's a good question. Refactored a little bit to default to HTTPS when tls() was called and added documentation about the behavior.

trustin:
@anuraaga Now defaulting to HTTPS when tls() was called. Please review again.

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -653,6 +631,14 @@ public void operationComplete(ChannelFuture f) throws Exception {
                     primaryActivePort = actualPort;
                 }
 
+                if (localAddress.getAddress().isAnyLocalAddress() ||
+                    localAddress.getAddress().isLoopbackAddress()) {
+                    logger.info("Serving {} at {} - {}://127.0.0.1:{}/",
```
anuraaga: This is a good change :) FWIW, it might be considered a breaking change as it will result in double-logging for production users of armeria who almost certainly are doing the logging themselves.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -74,15 +78,30 @@
...
+ *           .service(...)
+ *           .serviceUnder(...)
+ *       .and().build(); // Build a server.
  * }</pre>
+ *
+ *
+ * <h2 id="no_port_specified">What happens if no HTTP(S) port is specified?</h2>
+ *
+ * <p>When no TCP/IP port number or local address is specified, {@link ServerBuilder} will automatically bind
+ * to a random TCP/IP port assigned by O/S. It will serve HTTPS if you configured TLS (or HTTP otherwise), e.g.
```
anuraaga: `by the O/S` (or maybe OS, I'm not used to seeing a slash there)

trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/Server.java
trustin: I'm curious if we should log at DEBUG level.

trustin: Let me just leave this as it is for the sake of beginners' convenience. :-)

anuraaga: Yeah I think it's good to use `INFO` but should probably mention in the release notes.

trustin:
Thanks, reviewers!

PR #1052, opened 2018-03-06
Fix typo in `logback.xml` example

hyangtack:
In the `logback.xml` example, `SizeAndTimeBasedRollingPolicy` must be used because:
- `maxFileSize` element is specified.
- `%i` is used in the `fileNamePattern`.


trustin:
Thanks, @hyangtack !

PR #1054, opened 2018-03-07
Update the introductory part of the web site

trustin:
- Replace the slides with the latest one
- List some notable RPC protocol combinations we support
- Mention that we support DNS-based service discovery

Please let me know if you have a suggestion for making the index page more attractive. :-)


site/src/sphinx/index.rst
```diff
@@ -45,15 +45,20 @@ Integration with gRPC and Thrift
 - Your gRPC or Thrift service implementation runs on top of Armeria without any modification.
 - Works with the official gRPC or Thrift-over-HTTP client
 - Works with the code generated by the official Protobuf or Thrift IDL compiler
-- Automatically upgrades to HTTP/2 when used with the Armeria client
+- Supports various protocol combinations, such as:
+
+  - gRPC-Web
```
anuraaga: I guess gRPC-web is less commonly used so could go below in order. While a great, unique feature, I hope users of normal gRPC also feel like like switching to armeria :)

trustin: Reordered. Thanks for a feed back!

trustin:
Thanks, reviewers! @imasahiro Please feel free to let me know if you have a suggestion later. :-)

PR #1056, opened 2018-03-07
Fix to retry for HttpStatusBasedRetryStrategy when it gets exceptions…

minwoox:
… while processing response

Motivation:
Currently, `HttpStatusBasedRetryStrategy` ignores exceptions in the result.
It is common to have exceptions like `ClosedSessionException`, which should be retried
since they can be random socket closures, so it seems like a bad default to ignore
the exceptions.

Modifications:
- Change to retry on any `Exception`
- Change not to look up the response contents to retry by default
- Return `Backoff` instead of `Optional<Backoff>` in`RetryStrategy.shouldRetry()` 

Result:
- Retry on any exception

To-do:
- Provide knobs for the RetryStrategy

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
@@ -18,9 +18,8 @@ package com.linecorp.armeria.client.retry;
 
 import static java.util.Objects.requireNonNull;
 
-import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.Function;
+import java.util.function.BiFunction;
 
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpRequest;
@@ -30,35 +29,27 @@ import com.linecorp.armeria.internal.HttpHeaderSubscriber;
 
 /**
  * Provides a {@link RetryStrategy} that decides to retry the request based on the {@link HttpStatus} of
- * its response.
+ * its response or the {@link Exception} raised while processing the {@link HttpResponse}.
  */
 final class HttpStatusBasedRetryStrategy implements RetryStrategy<HttpRequest, HttpResponse> {
 
-    private final Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction;
+    private final BiFunction<HttpStatus, Throwable, Backoff> backoffFunction;
 
     /**
      * Creates a new instance.
      */
-    HttpStatusBasedRetryStrategy(Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction) {
-        this.statusBasedBackoffFunction = requireNonNull(
-                statusBasedBackoffFunction, "statusBasedBackoffFunction");
+    HttpStatusBasedRetryStrategy(BiFunction<HttpStatus, Throwable, Backoff> backoffFunction) {
+        this.backoffFunction = requireNonNull(backoffFunction, "backoffFunction");
     }
 
     @Override
-    public CompletableFuture<Optional<Backoff>> shouldRetry(HttpRequest request, HttpResponse response) {
+    public CompletableFuture<Backoff> shouldRetry(HttpRequest request, HttpResponse response) {
         final CompletableFuture<HttpHeaders> future = new CompletableFuture<>();
         final HttpHeaderSubscriber subscriber = new HttpHeaderSubscriber(future);
         response.completionFuture().whenComplete(subscriber);
         response.subscribe(subscriber);
 
-        return future.handle((headers, unused) -> {
-            if (headers != null) {
-                final HttpStatus resStatus = headers.status();
-                if (resStatus != null) {
-                    return statusBasedBackoffFunction.apply(resStatus);
-                }
-            }
-            return Optional.empty();
-        });
+        return future.handle((headers, thrown) -> backoffFunction
+                .apply(headers != null ? headers.status() : null, thrown));
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
@@ -17,9 +17,10 @@ package com.linecorp.armeria.client.retry;
 
 import static java.util.Objects.requireNonNull;
 
-import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.Function;
+import java.util.function.BiFunction;
+
+import javax.annotation.Nullable;
 
 import com.linecorp.armeria.client.ResponseTimeoutException;
 import com.linecorp.armeria.common.Flags;
@@ -44,17 +45,12 @@ public interface RetryStrategy<I extends Request, O extends Response> {
      * A {@link RetryStrategy} that defines a retry should not be performed.
      */
     static <I extends Request, O extends Response> RetryStrategy<I, O> never() {
-        return new RetryStrategy<I, O>() {
-            @Override
-            public CompletableFuture<Optional<Backoff>> shouldRetry(I request, O response) {
-                return CompletableFuture.completedFuture(Optional.empty());
-            }
-        };
+        return (request, response) -> CompletableFuture.completedFuture(null);
     }
 
     /**
      * Returns the {@link RetryStrategy} that retries the request with the {@link #defaultBackoff}
-     * when the response matches {@link HttpStatusClass#SERVER_ERROR}.
+     * when the response matches {@link HttpStatusClass#SERVER_ERROR} or an {@link Exception} is raised.
      */
     static RetryStrategy<HttpRequest, HttpResponse> onServerErrorStatus() {
         return onServerErrorStatus(defaultBackoff);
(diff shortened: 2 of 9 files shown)


core/src/main/java/com/linecorp/armeria/internal/HttpHeaderSubscriber.java
```diff
@@ -63,7 +68,6 @@ public void onNext(HttpObject o) {
             final HttpStatus status = headers.status();
             if (status != null && status.codeClass() != HttpStatusClass.INFORMATIONAL && this.headers == null) {
                 this.headers = headers;
-                subscription.cancel();
```
hyangtack: Just curious. Is it okay to remove `subscription.cancel()` here? 
I'm not sure, but is there any chance that `HttpData` is passed here due to the missing of cancellation?
If so, don't we need to release `HttpData` here if it is `ByteBufHttpData`?

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -36,12 +36,16 @@
...
     private final Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction;
 
+    private Backoff backoffOnException;
+
     /**
      * Creates a new instance.
      */
-    HttpStatusBasedRetryStrategy(Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction) {
+    HttpStatusBasedRetryStrategy(Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction,
+                                 Backoff backoffOnException) {
```
trustin: Shouldn't this be `Function<? super Exception, Optional<Backoff>> exceptionBasedBackoffFunction`?

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -66,8 +66,9 @@
      */
     static RetryStrategy<HttpRequest, HttpResponse> onServerErrorStatus(Backoff backoff) {
```
trustin: Please update the Javadoc.

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -79,7 +80,22 @@
      */
     static RetryStrategy<HttpRequest, HttpResponse> onStatus(
             Function<HttpStatus, Optional<Backoff>> statusBasedBackoffFunction) {
```
trustin: Please update the Javadoc.

core/src/main/java/com/linecorp/armeria/internal/HttpHeaderSubscriber.java
minwoox: I designed to use this class like `HttpMessageAggregator` which does `subscribe(false /* withPooledObject */)`, so do not have to worry about releasing the reference. However. this is a `public` class(even though it is in the `internal` package), I think I need to explain about it in the javadoc.

I was not sure about the `subscription.cancel()`, but it might be better to put it back.

hyangtack: Got it. Thank you for explanation. :-)

trustin:
I'm wonder if it's a good idea to replace `Function<HttpStatus, Optional<Backoff>>` to `Function<? super Exception, Optional<Backoff>>` and convert an `HttpStatus` to an `HttpStatusException`. This way, a user doesn't need to specify two functions, i.e. easier to compose.

minwoox:
I think it is a really good idea to have just one `Function`. But I have a doubt about converting the `HttpStatus` to an `HttpStatusException`. Because I think the `HttpStatusException` has to represent an `Exception` but it may not.

How about taking a BiFunction which is `BiFunction<HttpStatus, ? super Throwable, Optional<Backoff>>` or just `Function<Object, Optional<Backoff>>` ??

trustin:
I think `Function<? super Exception, Optional<Backoff>>` is fine:

```java
Function<? super Exception, Optional<Backoff>> fn = cause -> {
    if (cause instanceof HttpStatusException) {
        ...
    } else if ... {
    }
    return Optional.empty();
};
```

However, it would be pretty painful to write this from user's standpoint, so I think `BiFunction` approach may be better: `BiFunction<HttpStatus, Throwable, Optional<Backoff>>` and yeah, it's probably OK to omit `? super` for `Throwable`.

trustin:
Another question is: should we return `Optional<Backoff>` or just nullable `Backoff`?

minwoox:
I forgot why I brought the `Optional` here instead of nullable `Backoff`. Nullable `Backoff` looks just fine here and I actually prefer it.
But It is a breaking change, so I'm not sure if it's worth it. What do you think??

trustin:
Using `BiFunction` already guarantees a breaking change, so I see no issue with breaking more. ;-)

minwoox:
Alright then I will change it~! Thank you~!

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
minwoox: Updated!

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -44,51 +45,53 @@
...
     /**
-     * Returns the {@link RetryStrategy} that decides to retry the request using
-     * {@code statusBasedBackoffFunction}.
+     * Returns the {@link RetryStrategy} that decides to retry the request using the specified
+     * {@code backoffFunction}.
      *
-     * @param statusBasedBackoffFunction the {@link Function} that returns {@code Optional<Backoff>} according
-     *                                   to the {@link HttpStatus}
+     * @param backoffFunction the {@link BiFunction} that returns the {@link Backoff} according to the
+     *                        {@link HttpStatus} or {@link Throwable}
```
trustin: Could you mention that these input values can be `null`?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClientBuilder.java
```diff
@@ -78,6 +78,9 @@ public T maxTotalAttempts(int maxTotalAttempts) {
      * dose not specify.
      *
      * @return {@link T} to support method chaining.
+     *
+     * @see <a href=https://line.github.io/armeria/advanced-retry.html#per-attempt-timeout>
+     *     Per-attempt-timeout</a>
```
trustin: - There should be no whitespace between `>` and `Per`, because otherwise it will look like "[ this](https://line.github.io/armeria/)"
- `href` attribute value should be quoted.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClientBuilder.java
```diff
@@ -92,6 +95,9 @@ public T responseTimeoutMillisForEachAttempt(long responseTimeoutMillisForEachAt
      * corresponding responses are timed out by this value. {@code 0} disables the timeout.
      *
      * @return {@link T} to support method chaining.
+     *
+     * @see <a href=https://line.github.io/armeria/advanced-retry.html#per-attempt-timeout>
+     *     Per-attempt-timeout</a>
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -44,51 +45,53 @@
...
-     * to stop retry attempt. Note that {@link ResponseTimeoutException} is not retriable for
-     * the whole retry, but each attempt.
+     * Returns a {@link CompletableFuture} that contains {@link Backoff} which will be used for retry.
+     * If the condition does not match, this will return {@code null} to stop retry attempt.
+     * Note that {@link ResponseTimeoutException} is not retriable for the whole retry, but each attempt.
      *
      * @see RetryingClientBuilder#responseTimeoutMillisForEachAttempt(long)
+     *
+     * @see <a href=https://line.github.io/armeria/advanced-retry.html#per-attempt-timeout>
+     *     Per-attempt-timeout</a>
```
trustin: - There should be no whitespace between `>` and `Per`, because otherwise it will look like "[ this](https://line.github.io/armeria/)"
- `href` attribute value should be quoted.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -135,19 +135,17 @@ private void doExecute0(ClientRequestContext ctx, HttpRequestDuplicator rootReqD
...
-                       .handle(voidFunction((backoffOpt, unused) -> {
-                           if (backoffOpt.isPresent()) {
+        retryStrategy().shouldRetry(rootReqDuplicator.duplicateStream(), contentPreviewResponse(resDuplicator))
+                       .handle(voidFunction((backoff, unused) -> {
+                           if (backoff != null) {
                                final long millisAfter =
-                                       useRetryAfter ? getRetryAfterMillis(resDuplicator.duplicateStream())
-                                                     : -1;
+                                       useRetryAfter ?
+                                       getRetryAfterMillis(contentPreviewResponse(resDuplicator)) : -1;
```
trustin: It seems like you are struggling with indentation rules. :-) How about extracting this into a class so we don't have to worry about indentation? Alternatively, you could use `if`.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -30,7 +30,7 @@
 public class RetryingHttpClientBuilder extends RetryingClientBuilder<
         RetryingHttpClientBuilder, RetryingHttpClient, HttpRequest, HttpResponse> {
 
-    private static final int DEFAULT_CONTENT_PREVIEW_LENGTH = 1024 * 1024; // 1 MiB
+    private static final int DEFAULT_CONTENT_PREVIEW_LENGTH = 0;
```
trustin: How about adding `// No preview (0 byte)`?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -61,7 +61,9 @@ public RetryingHttpClientBuilder useRetryAfter(boolean useRetryAfter) {
     /**
      * Sets the length of content to look up whether retry or not. If the total length of content exceeds
      * this and there's no retry condition matched, it will hand over the stream to the client.
-     * @param contentPreviewLength the content length to preview. {@code 0} disables the length limit
+     * @param contentPreviewLength the content length to preview. Unlike other length related argument,
```
trustin: argument -> arguments

trustin:
Thanks, @minwoox !

PR #1057, opened 2018-03-07
[TRIVIAL] Remove unthrown Exceptions from HttpHealthCheckService

dongjinleekr:


trustin:
Thanks, @dongjinleekr !

PR #1058, opened 2018-03-07
Add server side connections, pending responses metrics

imasahiro:


imasahiro:
Because currently our service does not shutdown gracefully, want to know about pendings. Do we have any other metrics that is useful to investigate this?

core/src/main/java/com/linecorp/armeria/server/GracefulShutdownSupport.java
```diff
@@ -56,6 +56,11 @@ static GracefulShutdownSupport disabled() {
      */
     public abstract void dec();
 
+    /**
+     * Returns the number of pending requests.
```
hyangtack: nit: `requests` -> `responses` for consistency?

imasahiro: Oops, fixed

trustin:
> Because currently our service does not shutdown gracefully, want to know about pendings. Do we have any other metrics that is useful to investigate this?

A gauge for 'time taken since graceful shutdown started'?

imasahiro:
> A gauge for 'time taken since graceful shutdown started'?

Ah, sounds useful to me. Let me keep this PR as is and will open another PR to add the gauge.

trustin:
Thanks, @imasahiro !

PR #1059, opened 2018-03-07
Fix IntelliJ inspection warnings about nullness

trustin:
- Add missing `@Nullable` annotations. See the following IntelliJ
  inspections for more info:
  - Constant Conditions & Exceptions
  - `@NotNull`/`@Nullable` problems
- Fixed some obvious yet trivial bugs such as:
  - Redundant null check
  - Incorrectly annotated methods
- Also:
  - Add missing `final` to some local variables
  - Add missing `static` to some methods
  - Remove redundant `this.`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -18,6 +18,8 @@ package com.linecorp.armeria.client;
 
 import java.time.Duration;
 
+import javax.annotation.Nullable;
+
 import com.linecorp.armeria.common.ContentTooLargeException;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpRequest;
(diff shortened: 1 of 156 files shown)


trustin:
Ran the inspection a few more rounds to add more `@Nullable`s.

trustin:
I find:

- Some nullable methods can be pretty annoying - e.g. the access methods in `HttpHeaders` such as `path()`, `method()` and `status()`.
- Nullable fields, which will be initialized by the contract of Netty event model or Reactive Streams specification, can be also annoying.

Not sure this is a big proble, though. Just taking a note.

anuraaga:
There are even more fun annotations for dealing with lazy fields

https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/nullness/qual/MonotonicNonNull.java
https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/nullness/qual/EnsuresNonNull.java
https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/nullness/qual/EnsuresNonNullIf.java

Definitely annoying, but complete usage of these can make the codebase completely null-safe, even without the `assert` statements.

PR #1064, opened 2018-03-08
Implement flow control of gRPC stream writes.

### LINKED ISSUE #1037: Check gRPC backpressure code.
Since the gRPC stream reader requests an HTTP frame as soon as it gets another one, I don't think it is correctly implementing backpressure, it should probably only request as many frames as needed to satisfy the current request of messages.

anuraaga:
Currently, reads from the stream are correctly handled by propagating `request` through the stub, but writes are not since `isReady` is not properly implemented. When I first implemented the streams, I didn't understand `StreamWriter` properly, but now I do and realize we can use `onDemand` for this :)

With the change, the stream is only considered ready for writes when there is a request for them. I sort of want to optimize the non-backpressure case by only doing this if `isReady` is called, as that's a heuristic that should generally show the caller is implementing backpressure, but it's a bit weird and for unary calls the one call to `onDemand` doesn't seem like a big deal.

Fixes #1037

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
+++ b/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
@@ -22,6 +22,7 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import javax.annotation.Nullable;
 
@@ -72,6 +73,10 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     private static final Logger logger = LoggerFactory.getLogger(ArmeriaClientCall.class);
 
+    @SuppressWarnings("rawtypes")
+    private static final AtomicIntegerFieldUpdater<ArmeriaClientCall> pendingMessagesUpdater =
+            AtomicIntegerFieldUpdater.newUpdater(ArmeriaClientCall.class, "pendingMessages");
+
     private final ClientRequestContext ctx;
     private final Client<HttpRequest, HttpResponse> httpClient;
     private final HttpRequestWriter req;
@@ -90,7 +95,7 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     private boolean cancelCalled;
 
-    private volatile boolean ready = true;
+    private volatile int pendingMessages;
 
     ArmeriaClientCall(
             ClientRequestContext ctx,
@@ -201,17 +206,7 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     @Override
     public void sendMessage(I message) {
-        if (ready) {
-            ready = false;
-            req.onDemand(() -> {
-                ready = true;
-                try {
-                    listener.onReady();
-                } catch (Throwable t) {
-                    close(Status.fromThrowable(t));
-                }
-            });
-        }
+        pendingMessagesUpdater.incrementAndGet(this);
         if (ctx.eventLoop().inEventLoop()) {
             doSendMessage(message);
         } else {
@@ -221,13 +216,22 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     @Override
     public boolean isReady() {
-        return ready;
+        return pendingMessages == 0;
     }
 
     private void doSendMessage(I message) {
         try {
             final ByteBuf serialized = marshaller.serializeRequest(message);
             req.write(messageFramer.writePayload(serialized));
+            req.onDemand(() -> {
+                if (pendingMessagesUpdater.decrementAndGet(this) == 0) {
+                    try {
+                        listener.onReady();
+                    } catch (Throwable t) {
+                        close(Status.fromThrowable(t));
+                    }
+                }
+            });
         } catch (Throwable t) {
             cancel(null, t);
         }
--- a/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
+++ b/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
@@ -90,6 +90,8 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     private boolean cancelCalled;
 
+    private volatile boolean ready = true;
+
     ArmeriaClientCall(
             ClientRequestContext ctx,
             Client<HttpRequest, HttpResponse> httpClient,
@@ -199,11 +201,27 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     @Override
     public void sendMessage(I message) {
+        if (ready) {
+            ready = false;
+            req.onDemand(() -> {
+                ready = true;
+                try {
+                    listener.onReady();
+                } catch (Throwable t) {
+                    close(Status.fromThrowable(t));
+                }
+            });
+        }
         if (ctx.eventLoop().inEventLoop()) {
             doSendMessage(message);
         } else {
             ctx.eventLoop().submit(() -> doSendMessage(message));
         }
+    }
+
+    @Override
+    public boolean isReady() {
+        return ready;
     }
 
     private void doSendMessage(I message) {
--- a/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
+++ b/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import javax.annotation.Nullable;
 
@@ -86,6 +87,10 @@ class ArmeriaServerCall<I, O> extends ServerCall<I, O>
 
     private static final Logger logger = LoggerFactory.getLogger(ArmeriaServerCall.class);
 
+    @SuppressWarnings("rawtypes")
+    private static final AtomicIntegerFieldUpdater<ArmeriaServerCall> pendingMessagesUpdater =
+            AtomicIntegerFieldUpdater.newUpdater(ArmeriaServerCall.class, "pendingMessages");
+
     // Only most significant bit of a byte is set.
     @VisibleForTesting
     static final byte TRAILERS_FRAME_HEADER = (byte) (1 << 7);
@@ -126,7 +131,7 @@ class ArmeriaServerCall<I, O> extends ServerCall<I, O>
     private boolean sendHeadersCalled;
     private boolean closeCalled;
 
-    private volatile boolean ready = true;
+    private volatile int pendingMessages;
 
     ArmeriaServerCall(HttpHeaders clientHeaders,
                       MethodDescriptor<I, O> method,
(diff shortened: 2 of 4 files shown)


trustin:
Test failure:
```
com.linecorp.armeria.it.grpc.GrpcFlowControlTest > serverBackPressure FAILED
    java.lang.AssertionError: 
    Expecting:
     <11>
    to be between:
     [3, 5]
        at com.linecorp.armeria.it.grpc.GrpcFlowControlTest.serverBackPressure(GrpcFlowControlTest.java:273)
```

anuraaga:
Disabled the test by default, since I don't think it's possible to verify flow control reliably on the CI. The test relies on the rate of bytes transferred over the connection.

trustin:
I can't make `serverBackPressure()` pass on my laptop which is relatively idle, although the change itself looks fine to me.

anuraaga:
It's hard for me to be confident in the change, since the underlying flow control is a bit opaque :/ Let me know if you have any ideas how I might be able to improve the test, or of course the logic too

trustin:
I think we can't test backpressure with precision because OS will always buffer the data before actually sending it to the wire. What I did in `HttpServerTest` is probably more reliable way to test backpressure although it takes longer time.

hyangtack:
The change itself looks good to me, too.

trustin:
Boolean flag doesn't seem to work. Switched to a counter and the test passes reliably.

```diff
diff --git a/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java b/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
index e6da194d..64ccdaa2 100644
--- a/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
+++ b/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
@@ -22,6 +22,7 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import javax.annotation.Nullable;
 
@@ -72,6 +73,10 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     private static final Logger logger = LoggerFactory.getLogger(ArmeriaClientCall.class);
 
+    @SuppressWarnings("rawtypes")
+    private static final AtomicIntegerFieldUpdater<ArmeriaClientCall> pendingMessagesUpdater =
+            AtomicIntegerFieldUpdater.newUpdater(ArmeriaClientCall.class, "pendingMessages");
+
     private final ClientRequestContext ctx;
     private final Client<HttpRequest, HttpResponse> httpClient;
     private final HttpRequestWriter req;
@@ -90,7 +95,7 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     private boolean cancelCalled;
 
-    private volatile boolean ready = true;
+    private volatile int pendingMessages;
 
     ArmeriaClientCall(
             ClientRequestContext ctx,
@@ -201,17 +206,7 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     @Override
     public void sendMessage(I message) {
-        if (ready) {
-            ready = false;
-            req.onDemand(() -> {
-                ready = true;
-                try {
-                    listener.onReady();
-                } catch (Throwable t) {
-                    close(Status.fromThrowable(t));
-                }
-            });
-        }
+        pendingMessagesUpdater.incrementAndGet(this);
         if (ctx.eventLoop().inEventLoop()) {
             doSendMessage(message);
         } else {
@@ -221,13 +216,22 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
 
     @Override
     public boolean isReady() {
-        return ready;
+        return pendingMessages == 0;
     }
 
     private void doSendMessage(I message) {
         try {
             final ByteBuf serialized = marshaller.serializeRequest(message);
             req.write(messageFramer.writePayload(serialized));
+            req.onDemand(() -> {
+                if (pendingMessagesUpdater.decrementAndGet(this) == 0) {
+                    try {
+                        listener.onReady();
+                    } catch (Throwable t) {
+                        close(Status.fromThrowable(t));
+                    }
+                }
+            });
         } catch (Throwable t) {
             cancel(null, t);
         }
diff --git a/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java b/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
index bda7c82a..81cb5865 100644
--- a/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
+++ b/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import javax.annotation.Nullable;
 
@@ -86,6 +87,10 @@ class ArmeriaServerCall<I, O> extends ServerCall<I, O>
 
     private static final Logger logger = LoggerFactory.getLogger(ArmeriaServerCall.class);
 
+    @SuppressWarnings("rawtypes")
+    private static final AtomicIntegerFieldUpdater<ArmeriaServerCall> pendingMessagesUpdater =
+            AtomicIntegerFieldUpdater.newUpdater(ArmeriaServerCall.class, "pendingMessages");
+
     // Only most significant bit of a byte is set.
     @VisibleForTesting
     static final byte TRAILERS_FRAME_HEADER = (byte) (1 << 7);
@@ -126,7 +131,7 @@ class ArmeriaServerCall<I, O> extends ServerCall<I, O>
     private boolean sendHeadersCalled;
     private boolean closeCalled;
 
-    private volatile boolean ready = true;
+    private volatile int pendingMessages;
 
     ArmeriaServerCall(HttpHeaders clientHeaders,
                       MethodDescriptor<I, O> method,
@@ -223,17 +228,7 @@ class ArmeriaServerCall<I, O> extends ServerCall<I, O>
 
     @Override
     public void sendMessage(O message) {
-        if (ready) {
-            ready = false;
-            res.onDemand(() -> {
-                ready = true;
-                try {
-                    listener.onReady();
-                } catch (Throwable t) {
-                    close(Status.fromThrowable(t), EMPTY_METADATA);
-                }
-            });
-        }
+        pendingMessagesUpdater.incrementAndGet(this);
         if (ctx.eventLoop().inEventLoop()) {
             doSendMessage(message);
         } else {
@@ -247,6 +242,15 @@ class ArmeriaServerCall<I, O> extends ServerCall<I, O>
 
         try {
             res.write(messageFramer.writePayload(marshaller.serializeResponse(message)));
+            res.onDemand(() -> {
+                if (pendingMessagesUpdater.decrementAndGet(this) == 0) {
+                    try {
+                        listener.onReady();
+                    } catch (Throwable t) {
+                        close(Status.fromThrowable(t), EMPTY_METADATA);
+                    }
+                }
+            });
         } catch (RuntimeException e) {
             close(Status.fromThrowable(e), EMPTY_METADATA);
             throw e;
@@ -258,7 +262,7 @@ class ArmeriaServerCall<I, O> extends ServerCall<I, O>
 
     @Override
     public boolean isReady() {
-        return !closeCalled && ready;
+        return !closeCalled && pendingMessages == 0;
     }
 
     @Override
```

anuraaga:
Thanks for debugging this and the patch! Applied it (feel free to push straight into the branch if you ever want to though :) )

Too be honest, I only half-grok why this works and the boolean didn't, and it makes the `onDemand` API seem very unintuitive. What do you think? Anyways, glad it's working though.

trustin:
`StreamWriter.onDemand()` writes a `Future` into `StreamWriter`'s message queue, and the `Future` is completed when all messages before itself are consumed by the subscriber. Because we always call `onDemand()` after each `write()`, we know exactly one message was consumed when the callback is invoked.

The problem with using a `boolean` flag is that the `onReadyHandler` we set in `ClientResponseObserver.beforeStart()` can reenter. If I add some printlns around the `onReadyHandler` and where `ready` changes, the output looks like this:

When test passes:

```
0 ready = false
0 ready = true, calling listener.onReady()
0 ready = false
0 ready = true, calling listener.onReady()
0 ready = false
0 ready = true, calling listener.onReady()
0 ready = false
ENTER
onNext(REQUEST)
1 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
2 ready = false
EXIT
EXIT
0 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
1 ready = false
EXIT
0 ready = false
0 ready = true, calling listener.onReady()
```

When test fails:

```
0 ready = false
0 ready = true, calling listener.onReady()
0 ready = false
0 ready = true, calling listener.onReady()
0 ready = false
0 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
1 ready = false
1 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
2 ready = false
2 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
3 ready = false
3 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
4 ready = false
4 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
5 ready = false
5 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
6 ready = false
6 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
7 ready = false
7 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
8 ready = false
8 ready = true, calling listener.onReady()
ENTER
onNext(REQUEST)
9 ready = false
9 ready = true, calling listener.onReady()
ENTER
EXIT
EXIT
EXIT
EXIT
EXIT
EXIT
EXIT
EXIT
EXIT
EXIT
0 ready = false
```

The integer before `ready =` is the depth of recursion calculated using an integer thread-local variable. Because `ready` becomes `true` immediately after it's set to `false`, gRPC never sees the `false` flag and thus keeps calling the `onReadyHandler`.

trustin:
Actually, this makes me think we can actually solve this problem using a `boolean`. :-) Gimme a sec.

trustin:
Uh, actually `boolean` will not work if `sendMessage()` is called simultaneously from multiple threads. Here's the half-fix with `boolean` anyway:

```java
    @Override
    public void sendMessage(I message) {
        ready = false;
        if (ctx.eventLoop().inEventLoop()) {
            doSendMessage(message);
        } else {
            ctx.eventLoop().submit(() -> doSendMessage(message));
        }
    }

    private void doSendMessage(I message) {
        try {
            final ByteBuf serialized = marshaller.serializeRequest(message);
            req.write(messageFramer.writePayload(serialized));
            req.onDemand(() -> {
                ready = true;
                try {
                    listener.onReady();
                } catch (Throwable t) {
                    close(Status.fromThrowable(t));
                }
            });
        } catch (Throwable t) {
            cancel(null, t);
        }
    }
```

Note that we called `onDemand()` *after* `write()` so that the callback is invoked only after the message is consumed. This would work only if there is only a single writer, but I guess we can't guarantee that.

trustin:
So, the core of this problem is two fold:

- `onDemand()` must be invoked after `write()` to have the callback invoked after the message is consumed.
- When there are more than one writers simultaneously, the `boolean` flag will be set to `true` prematurely, when there are still messages left in the queue.

anuraaga:
Thanks for the explanation! That makes sense. Personally, the `isReady` and `setReadyHandler` style of gRPC was easier to use (required no state on caller end and ready is an easier concept to grok than demand for me) so we may be able to find some inspiration there to make the API cleaner.

For project reactor, I found a method that returns the number of pending objects (basically equivalent to isReady) but didn't find a equivalent of either onDemand or setReadyHandler, wonder how they implement the concept (they have tons of methods so I'm probably just missing it...)

https://projectreactor.io/docs/core/release/api/reactor/core/publisher/EmitterProcessor.html

PR #1065, opened 2018-03-08
Fix to propegate the Exception in the StreamMessageDuplicator to the …

minwoox:
…Subsriber

Motivation:
`StreamMessageDuplicator` just logs refcnt exception and does not propegate to
the `Subscriber`. So the `Subsriber` hangs forever.

Modification:
- Handle the `Exception` while processing the `ByteBuf`

Result:
- Client does not hang forever

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
@@ -607,7 +607,7 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
             if (n <= 0) {
                 final Throwable cause = new IllegalArgumentException(
                         "n: " + n + " (expected: > 0, see Reactive Streams specification rule 3.9)");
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -704,18 +704,26 @@ private boolean doSignalSingle(SignalQueue signals) {
...
+                            obj = retainedDuplicate((ByteBuf) obj);
+                        }
+                    } else {
+                        if (obj instanceof ByteBufHolder) {
+                            obj = copy((ByteBufHolder) obj);
+                        } else if (obj instanceof ByteBuf) {
+                            obj = copy((ByteBuf) obj);
+                        }
                     }
+                } catch (Exception e) {
```
trustin: Should we catch a `Throwable` here?

minwoox: Oops let me change this PR~

trustin:
Thanks, @minwoox

Issue #882, opened 2017-11-29
Use Curator Framework instead of vanilla ZooKeeper for armeria-zookeeper


trustin:
.. because it simplifies our code a lot and frees us from maintaining ZooKeeper-handling code. i.e. less bugs and test failures.

trustin:
Note about dependencies:

The following would be the only dependencies involved:

- `curator-framework`
- `curator-client`
- `zookeeper`
- `slf4j-api`

We can exclude the following transitive dependencies:

- `guava` (not really a dependency since it's shaded)
- `log4j` (not really a required dependency)
- `slf4j-log4j12` (not really a required dependency)
- `netty` (not really a required dependency)
- `audience-annotations` (documentation-purpose only)
- `jline` (only used for command line tools?)

dongjinleekr:
Hello. I am currently working on this issue. Before progress more, I would like to ask some questions.

As all of you know, apache curator provides its core package (curator-client) with its high-level module, curator-x-discovery, which serves the Service Discovery feature.

My question is: **Should we adapt curator-x-discovery for armeria's service discovery feature? There are pros and cons for each of strategy.**

- Use curator-client only: We can maintain the current API & implementations, but can not make use of safe, verified module.
- Use curator-x-discovery: We can make use of the verified module, but there is some potential to change the current API, to be fitted into the curator-x-discovery module.

The one thing that should be noted is that **We have to support the other Service Discovery frameworks in the future (issue https://github.com/line/armeria/issues/194).** That is, avoiding curator-x-discovery does not eliminate the potential to make some changes to current Service Discovery API. If we support other Service Discovery Service (for both of registration/resolving), defining a new API would be required at some point.

dongjinleekr:
Please note: https://curator.apache.org/apidocs/org/apache/curator/x/discovery/UriSpec.html

trustin:
`curator-x-discovery` looks great, but it doesn't look compatible with how we store the server list in ZooKeeper. How about implementing both - one that's backward-compatible with the current implementation and the other that uses `curator-x-discovert`? It's OK to break the API compatibility as long as the data layout is compatible. If you are not interested in implemeting two, then you can just implement one you prefer.

dongjinleekr:
It's almost done. I expect that I can open the pull request in next week.

However, I have a question: **In the current work, I implemented `ServerListenerBuilder` which creates `ServerListener` instance by composing `Runnable` instances. Is this okay?**

[...]

PR #1071, opened 2018-03-10
[TRIVIAL] Remove unused JettyService#PATH_SPLITTER

dongjinleekr:
Found while working on #882.


trustin:
Thanks, @dongjinleekr! By the way, you don't need to prefix your PR subject with `[TRIVIAL]`.

PR #1085, opened 2018-03-12
Add ServerBuilder.tlsSelfSigned()

trustin:
Motivation:

It is a common pattern to create a self-signed certificate to test a
server and we could simplify that.

Modifications:

- Add ServerBuilder.tlsSelfSigned() which generates a self-signed
  certificate automatically.
- Remove some code that uses `SelfSignedCertificateGenerator` directly.

Result:

- Brevity

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
@@ -200,7 +200,7 @@ abstract class AbstractVirtualHostBuilder<B extends AbstractVirtualHostBuilder> 
 
     /**
      * Configures SSL or TLS of this {@link VirtualHost} with an auto-generated self-signed certificate.
-     * Note: You should never use this in production but only for a testing purpose.
+     * <strong>Note:</strong> You should never use this in production but only for a testing purpose.
      *
      * @throws CertificateException if failed to generate a self-signed certificate
      */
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -462,8 +463,8 @@ public ServerBuilder tls(SslContext sslContext) {
     }
 
     /**
-     * Sets the {@link SslContext} of the default {@link VirtualHost} from the specified
-     * {@code keyCertChainFile} and cleartext {@code keyFile}.
+     * Configures SSL or TLS of  the default {@link VirtualHost} from the specified {@code keyCertChainFile}
```
imasahiro: Remove extra whitespace between `of` and `the`
* `Configures SSL or TLS of  the` -> `Configures SSL or TLS of the`

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -196,6 +198,17 @@ public B tls(File keyCertChainFile, File keyFile, @Nullable String keyPassword)
         return self();
     }
 
+    /**
+     * Configures SSL or TLS of this {@link VirtualHost} with an auto-generated self-signed certificate.
+     * Note: You should never use this in production but only for a testing purpose.
```
anuraaga: Maybe worth putting in all, or many, caps for emphasis

trustin: Used `<strong>` :-)

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: Fixed. Thanks!

trustin:
Thanks, reviewers!

PR #1089, opened 2018-03-14
Add Endpoint.ipAddr

### LINKED ISSUE #1087: Add Endpoint.ipAddr()
Currently, `Endpoint.host()` is used as a hostname or an IP address of an endpoint. This does not work well for the `Endpoint`s whose IP addresses are known already. For example, let's say we want to use `DnsServiceEndpointGroup` with the following SRV records:

```
_https._tcp.foo.com. 3600 IN SRV 10 100 443 one.foo.com.
_https._tcp.foo.com. 3600 IN SRV 10 100 443 two.foo.com.
_https._tcp.foo.com. 3600 IN SRV 10 100 443 three.foo.com.
```

Because the client will have to create a TLS connection with proper SNI settings, the `Endpoint` should provide `foo.com` as the host name and provide `<the IP address of {one,two,three}.foo.com>` as the IP address of `foo.com`, which is not possible with our current API.

To make this possible, we would also improve `DnsServiceEndpointGroup` to send A/AAAA queries for the host names in the `SRV` records.

What do you think, @anuraaga?

trustin:
Motivation:

Currently, `Endpoint.host()` is used as a hostname or an IP address of
an endpoint. This does not work well for the `Endpoint`s whose IP
addresses are known already. For example, let's say we want to use
`DnsServiceEndpointGroup` with the following SRV records:

    _https._tcp.foo.com. 3600 IN SRV 10 100 443 one.foo.com.
    _https._tcp.foo.com. 3600 IN SRV 10 100 443 two.foo.com.
    _https._tcp.foo.com. 3600 IN SRV 10 100 443 three.foo.com.

Because the client will have to create a TLS connection with proper SNI
settings, the `Endpoint` should provide `foo.com` as the host name and
provide `<the IP address of {one,two,three}.foo.com>` as the IP address
of `foo.com`, which is not possible with our current API.

Modifications:

- Add a new property `ipAddr` to `Endpoint`
- Add `Endpoint.withIpAddr()`
- Handle the case where a host name in an `Endpoint` is an IP address
- Add test cases
- Compare authority, IP address and port only
  - Take `port` and `ipAddr` into account when comparing.
  - Do *not* take `weight` into account when comparing.
- Add host name validation using Guava InternetDomainName
- Fix incorrect uses of `Endpoint.of()` in some test cases
- Deprecate `Endpoint.of(host, port, weight)` in favor of
  `Endpoint.of(host, port).withWeight(weight)`

Result:

- Fixes #1087
- A user can create an `Endpoint` with an IP address.
- `Endpoint.equals()` makes more sense.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Endpoint.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Endpoint.java
@@ -19,23 +19,33 @@ package com.linecorp.armeria.client;
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
+import java.util.Objects;
+
 import javax.annotation.Nullable;
 
+import com.google.common.base.MoreObjects;
+import com.google.common.base.MoreObjects.ToStringHelper;
 import com.google.common.net.HostAndPort;
+import com.google.common.net.InternetDomainName;
 
 import com.linecorp.armeria.client.endpoint.EndpointGroupRegistry;
+
+import io.netty.util.NetUtil;
 
 /**
  * A remote endpoint that refers to a single host or a group of multiple hosts.
  *
- * <p>A host endpoint has {@link #host()} and optional {@link #port()} and it can be represented as
- * {@code "<host>"} or {@code "<host>:<port>"} in the authority part of a URI.
+ * <p>A host endpoint has {@link #host()}, optional {@link #ipAddr()} and optional {@link #port()}. It can be
+ * represented as {@code "<host>"} or {@code "<host>:<port>"} in the authority part of a URI. It can have
+ * an IP address if the host name has been resolved and thus there's no need to query a DNS server.
  *
  * <p>A group endpoint has {@link #groupName()} and it can be represented as {@code "group:<groupName>"}
  * in the authority part of a URI. It can be resolved into a host endpoint with
  * {@link #resolve(ClientRequestContext)}.
  */
 public final class Endpoint {
+
+    private static final int DEFAULT_WEIGHT = 1000;
 
     /**
      * Parse the authority part of a URI. The authority part may have one of the following formats:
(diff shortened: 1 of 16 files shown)


core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -83,36 +93,71 @@ public static Endpoint of(String host) {
...
 
     @Nullable
     private final String groupName;
     @Nullable
     private final String host;
+    @Nullable
+    private final String ipAddr;
     private final int port;
     private final int weight;
+    private final boolean hostIsIpAddr;
```
anuraaga: You can also consider a `@Nullable` enum so it's just one field.

trustin: Good idea. Now using a nullable enum.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -283,21 +409,36 @@ public boolean equals(Object obj) {
...
-        return authority().hashCode();
+        return (authority().hashCode() * 31 + Objects.hashCode(ipAddr)) * 31 + port;
+    }
+
+    @Override
+    public String toString() {
+        final ToStringHelper helper = MoreObjects.toStringHelper(this).omitNullValues();
+        helper.addValue(authority());
+        if (!isGroup()) {
+            if (hostIpAddrType == null) {
```
hyangtack: I'm not sure but `hostIpAddrType != null`? Or did you intend writing non-IP address?

trustin: if `hostIpAddrType` is not `null`, it means `host` is an IP address, and we already did `addValue()` with it, so we do not add it again.

trustin:
Thanks, reviewers!

PR #1090, opened 2018-03-15
Add MoreNamingConventions.configure()

trustin:
Motivation:

It would be useful to provide a utility method that configure an
existing MeterRegistry with our NamingConvention implementations in
MoreNamingConventions.

Modifications:

- Add MoreNamingConventions.configure() and its tests.

Result:

- Less boilerplates in a user application

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/common/metric/MoreNamingConventionsTest.java
+++ b/core/src/test/java/com/linecorp/armeria/common/metric/MoreNamingConventionsTest.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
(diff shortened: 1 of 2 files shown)


trustin:
@kojilin wrote:
> LGTM, just curious should we call it in armeria-line rather than let user decide?

I think so.

core/src/test/java/com/linecorp/armeria/common/metric/MoreNamingConventionsTest.java
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2017 LINE Corporation
```
minwoox: 2018 :)

trustin:
Build failure seems unrelated. Running the build again.

trustin:
Thanks, reviewers!

PR #1091, opened 2018-03-15
Fix potential buffer leaks due to empty ByteBufHttpData

trustin:
Motivation:

We have many places where drops an empty HttpData silently. This can
lead to the leak of empty ByteBufHttpData.

Modifications:

- Use Unpooled.EMPTY_BUFFER for an empty buffer when creating a new
  ByteBufHttpData
- Release ByteBufHttpData explicitly when we drop it even if it's not
  empty
- Fix the incorrect reference count tests in AbstractStreamMessageAndWriterTest
- Update the Javadoc of StreamWriter about the life cycle of
  reference-counted objects

Result:

- Less buffer leak

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
@@ -35,6 +35,7 @@ import com.linecorp.armeria.common.FixedHttpResponse.TwoElementFixedHttpResponse
 import com.linecorp.armeria.common.stream.StreamMessage;
 import com.linecorp.armeria.common.util.Exceptions;
 
+import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.EventExecutor;
 
 /**
@@ -209,6 +210,7 @@ public interface HttpResponse extends Response, StreamMessage<HttpObject> {
         }
 
         if (isContentAlwaysEmptyWithValidation(status, content, trailingHeaders)) {
+            ReferenceCountUtil.safeRelease(content);
             return new OneElementFixedHttpResponse(headers);
         } else if (!content.isEmpty()) {
             if (trailingHeaders.isEmpty()) {
--- a/core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
@@ -23,8 +23,31 @@ import javax.annotation.CheckReturnValue;
 
 import org.reactivestreams.Subscriber;
 
+import com.linecorp.armeria.unsafe.ByteBufHttpData;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.util.ReferenceCounted;
+
 /**
  * Produces the objects to be published by a {@link StreamMessage}.
+ *
+ * <h3 id="reference-counted">Life cycle of reference-counted objects</h3>
+ *
+ * <p>When the following methods are given with a {@link ReferenceCounted} object, such as {@link ByteBuf} and
+ * {@link ByteBufHttpData}, or the {@link Supplier} that provides such an object:
+ *
+ * <ul>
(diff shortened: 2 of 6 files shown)


core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -209,6 +210,7 @@ static HttpResponse of(HttpHeaders headers, HttpData content, HttpHeaders traili
         }
 
         if (isContentAlwaysEmptyWithValidation(status, content, trailingHeaders)) {
+            ReferenceCountUtil.safeRelease(content);
```
anuraaga: Do you still want to keep these checks with the other change? I agree that it is pretty hairy (we'll want to add below too I think for the branches triggered by `content.isEmpty()`

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -40,13 +41,20 @@
     private final int length;
 
     /**
-     * Construct a new {@link ByteBufHttpData}. Ownership of {@code buf} is taken by this
+     * Constructs a new {@link ByteBufHttpData}. Ownership of {@code buf} is taken by this
      * {@link ByteBufHttpData}, which must not be mutated anymore.
      */
     public ByteBufHttpData(ByteBuf buf, boolean endOfStream) {
-        this.buf = buf;
+        final int length = requireNonNull(buf, "buf").readableBytes();
```
anuraaga: Don't see a big reason to not just use the field without the local variable.

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
anuraaga: Oh nevermind, these are cases where the buf might not be empty, got it

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
trustin: Like this?

    length = buf.readableBytes();
    if (length != 0) { ... }

anuraaga: Yup - the field is final too, so seems to be the same

trustin: Fixed

trustin:
I've just updated the Javadoc of StreamWriter.write() and tryWrite(). Please review again.

core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
```diff
@@ -74,7 +98,14 @@ default void write(Supplier<? extends T> o) {
 
     /**
      * Writes the specified object {@link Supplier} to the {@link StreamMessage}. The object provided by the
-     * {@link Supplier} will be transferred to the {@link Subscriber}.
+     * {@link Supplier} will be transferred to the {@link Subscriber}. Note: If the object provided by the
+     * {@link Supplier} is a {@link ReferenceCounted} such as {@link ByteBuf} and {@link ByteBufHttpData},
+     * it will be released on any of the following events:
```
anuraaga: Is there any event that doesn't release it? I think it might be simpler to just say ownership is transferred since there doesn't seem to be a situation it'd be safe for the user not to retain if they want to keep using the object.

trustin: Ownership is not transferred completely if a user retains it, so I wasn't sure if it's enough to just say so. Let me try to change this a little bit so that it's clear that the object will always be released.

trustin:
Updated Javadoc again. Please review again. :-)

trustin:
Thanks, reviewers!

PR #1093, opened 2018-03-15
Reduce the amount of static memory required for testing :grpc

trustin:
Motivation:

:grpc:test sometimes fails due to OutOfMemoryError.

Modifications:

- Dereference multi-megabyte strings when the relevant tests are
  finished

Result:

- Less chance of OOME in an environment with memory constraint


trustin:
Build failure looks like a timing issue. Re-building.

trustin:
Thanks!

Issue #1066, opened 2018-03-08
Default "RequestConverterFunction" for Java beans


LiYing2010:
Currently, there are only 3 default implements for `RequestConverterFunction`:
(1)`ByteArrayRequestConverterFunction`
(2)`StringRequestConverterFunction`
(3)`JacksonRequestConverterFunction`

If I want to parse request parameters from `path mapping result`, `query string`, `request headers`, and wrap all these parameters into a Java Bean, I have to implement my own `RequestConverterFunction` class.
It is possible, but a little in-convenience.

It will be helpful if `Armeria` can provide a new default implement for `RequestConverterFunction` which can handle all these boilerplate code for us.

The basic idea is:
(1)user only need to declare a request bean, e.g.
```
public class MyRequestBean {
  @Param("userName")    <== this param should retrieve from path variable or query string
  private String userName;

  @Param("age")
  private int age;

  @Header("x-user-permission")    <== this param should retrieve from HTTP Request Header
  private String permission;
}
```

(2)and the new `RequestConverterFunction` implement should parse request parameters from `path mapping result`, `query string`, `request headers` etc, and set them into the bean instance

I think this new feature will make `request conversion` pattern much more usable.

trustin:
Sounds like a good idea. Please feel free to send us a PR if you are interested. I'm curious how you would inject the parameters into the member fields or the setter method invocations. Do you know any light-weight bean property injection library we can use as a shaded dependency?

LiYing2010:
Hi, I created a PR for this:  https://github.com/line/armeria/pull/1096
Could anyone review it?

trustin:
Thank you! We will review it soon. :+1:

PR #1096, opened 2018-03-15
New Feature: Default "RequestConverterFunction" for Java beans

LiYing2010:
related issue: https://github.com/line/armeria/issues/1066

Change List:
1. make `@Header` and `@Param` can be used for `fields`/`methods`/`constructors`/`method params`/`constructor params` of request bean
2. implemented `BeanRequestConverterFunction`
3. refactor: moved some param processing related code from `AnnotatedHttpServiceMethod` to a new class `AnnotatedHttpServiceParamUtil`, so these code can be reused by `BeanRequestConverterFunction`
4. added `BeanRequestConverterFunction` as a default Request Converters in `AnnotatedHttpServices`
5. added unit tests for `BeanRequestConverterFunction`


core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        final BeanParamInfo beanParamInfo =
+                beanParamsMap.computeIfAbsent(expectedResultType,
+                                              JavaBeanRequestConverterFunction::buildBeanParamInfo);
+
+        final Object requestBean = beanParamInfo.createInstance();
+
+        HttpParameters httpParams = null;
+        for (final FieldParamInfo fieldParamInfo : beanParamInfo.fieldParamList) {
```
hyangtack: How about splitting `FieldParamInfo` into something like `ParamField` and `HeaderField`? Then, we can move the code in this `for` loop into them.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        return requestBean;
+    }
+
+    private static boolean shouldHandle(final AggregatedHttpMessage request) {
+        final MediaType contentType = request.headers().contentType();
+        if (contentType != null && (contentType.is(MediaType.JSON) ||
+                                    contentType.subtype().endsWith("+json"))) {
+            // request in JSON format, should be handled by {@link JacksonRequestConverterFunction}
```
hyangtack: nit: You don't need to use `{@link}` here because it's not a part of Javadoc. :-)

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,181 @@
...
+    public Object convertRequest(final ServiceRequestContext ctx,
+                                 final AggregatedHttpMessage request,
+                                 final Class<?> expectedResultType) throws Exception {
+        if (!shouldHandle(request)) {
+            return RequestConverterFunction.fallthrough();
+        }
+
+        final BeanParamInfo beanParamInfo =
+                beanParamsMap.computeIfAbsent(expectedResultType,
+                                              JavaBeanRequestConverterFunction::buildBeanParamInfo);
```
hyangtack: Is there any way to build `BeanParamInfo` when the server starts up? I'm a little worried about the performance when many requests are coming at first time.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
```diff
@@ -104,7 +105,8 @@
     private static final List<RequestConverterFunction> defaultRequestConverters =
             ImmutableList.of(new JacksonRequestConverterFunction(),
                              new StringRequestConverterFunction(),
-                             new ByteArrayRequestConverterFunction());
+                             new ByteArrayRequestConverterFunction(),
+                             new JavaBeanRequestConverterFunction());
```
hyangtack: I guess it would be used more than `StringRequestConverterFunction` and `ByteArrayRequestConverterFunction`.  😃 How about moving this to up?

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: It can be done, but need to add a new method like `prepare` or `init` to the `RequestConverterFunction` interface, and call this new method from `AnnotatedHttpServiceMethod` at the server starts up phase.
Does this strategy sounds good?

If you agree with it, I will do it.

hyangtack: I think that it's enough to add a new method like `prepare` or `init` to the `JavaBeanRequestConverterFunction` only. 
If you create a singleton instance of that class in `AnnotatedHttpServices` class, we may build `BeanParamInfo` by calling `prepare` method of the singleton instance [here (maybe)](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java#L449). Also, we may add the converter only if necessary. 
(It means that we can remove it from the default converter list)
What do you think?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
LiYing2010: Actually, I add `JavaBeanRequestConverterFunction` as the last item in this `defaultRequestConverters` list **on purpose**, to make it has the **lowest** priority in all the converters, so `ByteArray`/`String`/`Jackson` requests will be handled by corresponding converters first, and then pass all the other requests to `JavaBeanRequestConverterFunction`, which will treat request object as Java Bean.
By this way, I don't need to implement a complicated `shouldHandle this request or should pass it to other converts?` check  for `JavaBeanRequestConverterFunction`

But to be honest, this trick is a little too hacky, and hard to understand.

So, I will change it, to move `JavaBeanRequestConverterFunction` to the head of `defaultRequestConverters` list, and implement a better `shouldHandle...?` check for it.

hyangtack: I see. I'm not sure but what if we instantiate a request converter which is dedicated to a class of Java bean. In [this comment](https://github.com/line/armeria/pull/1096#discussion_r175024295), I mentioned a singleton instance, but it may be better to create a new converter instance on demand.

```
public class JavaBeanRequestConverterFunction implements RequestConverterFunction {
    
    private final Class<?> expectedResultType;

    public JavaBeanRequestConverterFunction(Class<?> expectedResultType) {
        this.expectedResultType = expectedResultType;
    }

    @Override
    public Object convertRequest(final ServiceRequestContext ctx,
                                 final AggregatedHttpMessage request,
                                 final Class<?> expectedResultType) throws Exception {
        if (this.expectedResultType != expectedResultType) {
            return RequestConverterFunction.fallthrough();
        }
        ...
    }
```

LiYing2010: I think I can do this:
1. create a singleton instance of `JavaBeanRequestConverterFunction`
2. call `checkResultType` method of this singleton instance, for **every** `@RequestObject` annotated method params, at the server starts up phase.
3. in this `checkResultType` method, `JavaBeanRequestConverterFunction` can check if `@Param/@Header` annotated fields exist? if so, build a `BeanParamInfo` for this ResultType, and add this  ResultType in to a `tagetResultTypeList`
4. `shouldHandle this request?` check can simply check if `expectedResultType` exists in `tagetResultTypeList`
5. if `tagetResultTypeList` is empty, which means there is no JavaBean request param exists, so we can remove `JavaBeanRequestConverterFunction` from our `defaultRequestConverters` list 

How do you think about this strategy?

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServices.java
LiYing2010: fixed, but a little different from the plan I mentioned above.
1. not using `singleton instance` pattern, but change `beanParamsMap` to `static`
2. created a `static void initFor(final Class<?> expectedResultType)`  method, to check if every result type should be processed, and if so, build a `BeanParamInfo` for it, and save to `beanParamsMap`. This method is invoked at the server starts up phase, for every `@RequestObject` annotated method params
3. `shouldHandle this request?` check simply check if `expectedResultType` is in the `beanParamsMap`
4. abandoned the plan which remove the `JavaBeanRequestConverterFunction` instance from  `defaultRequestConverters` list. because there is a Immutable copy of this list in every `AnnotatedHttpServiceMethod`, so the remove logic will be a little complicated. And, `shouldHandle this request?` is very quick, so I think there will be no performance issue even we don't remove `JavaBeanRequestConverterFunction` which is no need to exist

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceRequestConverterTest.java
```diff
@@ -115,20 +123,51 @@ public String defaultText(@RequestObject String obj1) {
             return obj1;
         }
 
-        @Post("/default/invalid")
+        @Post("/default/javabean/{userName}/{seqNum}")
+        public String defaultJavaBeanForPost(@RequestObject RequestJavaBean1 bean1) throws JsonProcessingException {
+            assertThat(bean1).isNotNull();
+            assertThat(bean1.userName).isNotNull();
```
huydx: should we confirm value of bean as well? (not just nullable check)

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+        if (beanParamInfo == null) {
+            ignoreResultTypeList.add(expectedResultType);
+        } else {
+            beanParamsMap.put(expectedResultType, beanParamInfo);
+        }
+    }
+
+    private static boolean shouldHandle(final AggregatedHttpMessage request) {
+        final MediaType contentType = request.headers().contentType();
+        if (contentType != null && (contentType.is(MediaType.JSON) ||
```
huydx: could be simplified into single return

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+import com.linecorp.armeria.common.HttpParameters;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.common.util.Holder;
+import com.linecorp.armeria.server.AnnotatedHttpServiceParamUtil.EnumConverter;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+import io.netty.util.AsciiString;
+
+public class JavaBeanRequestConverterFunction implements RequestConverterFunction {
+    private static final Map<Class<?>, BeanParamInfo> beanParamsMap = new HashMap<>();
```
huydx: is it need to be concurrent? See that same logic at JsonConverter need ConcurrentHashmap, so there is a chance that convertRequest will be called concurrently right?

core/src/main/java/com/linecorp/armeria/common/util/Holder.java
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.common.util;
+
+import java.util.function.Supplier;
+
+public class Holder<T> {
```
huydx: should be final

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,223 @@
...
+     * <p>Usually one of a query string of a URI or URL-encoded form data is specified in the request.
+     * If both of them exist though, they would be decoded and merged into a parameter map.</p>
+     *
+     * <p>Names and values of the parameters would be decoded as UTF-8 character set.</p>
+     *
+     * @see QueryStringDecoder#QueryStringDecoder(String, boolean)
+     * @see HttpConstants#DEFAULT_CHARSET
+     */
+    public static HttpParameters httpParametersOf(ServiceRequestContext ctx,
+                                                  HttpHeaders reqHeaders,
```
huydx: is `HttpHeaders` needed while we could get header from `AggregatedHttpMessage`?

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+                                 final AggregatedHttpMessage request,
+                                 final Class<?> expectedResultType) throws Exception {
+        final BeanParamInfo beanParamInfo = beanParamsMap.get(expectedResultType);
+
+        if (beanParamInfo == null) {
+            // no BeanParamInfo for this result type, we should not handled it
+            return RequestConverterFunction.fallthrough();
+        }
+
+        if (!shouldHandle(request)) {
```
hyangtack: I'm not sure what if we remove `shouldHandle()`? We collected `expectedResultType`s which we can handle, so I think we should handle the request if `beanParamInfo` is non-null.

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceRequestConverterTest.java
LiYing2010: Can not check value here, because value will be different for every request.

So I convert the whole request bean into a JSON string, and return it as a response, then check if the response is same as expected JSON string. like this:
`assertThat(response.content().toStringUtf8()).isEqualTo(expectedResponseContent)`

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
LiYing2010: This code is extracted from `AnnotatedHttpServiceMethod.java`, and in `AnnotatedHttpServiceMethod`, `HttpHeaders` is retrieved from `HttpRequest.headers()` instead of `AggregatedHttpMessage.headers()`.
I am not sure if I can change the original logic of `AnnotatedHttpServiceMethod`

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: The current implementation is, `shouldHandle` also checks if contentType of request is in JSON format, if it is, this request should be handled by `JacksonRequestConverterFunction` .

In this strategy, we can re-use same Java Bean for JSON request(`JacksonRequestConverterFunction`) and normal request`JavaBeanRequestConverterFunction`


If we remove this `shouldHandle` check, then there will be not this `re-use same Java Bean` feature.

So, what do you think, which is better?

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.
Changed to `ConcurrentMap` type, and `beanParamInfoCache` field name

core/src/main/java/com/linecorp/armeria/common/util/Holder.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+        private final Field field;
+
+        private final EnumConverter<?> enumConverter;
+
+        private FieldParamInfo(final Field field) {
+            this.field = field;
+            this.field.setAccessible(true);
+
+            if (field.getType().isEnum()) {
+                enumConverter = new EnumConverter(field.getType().asSubclass(Enum.class));
```
hyangtack: `new EnumConverter<>(...)`?

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
hyangtack: I can see what you intend. Is that one of your usage? 

I'm worried that your strategy might make a user be confused. Let's assume that a user writes `@Param` and `@Header` annotation in their class for extracting HTTP request information. In this case, it would not be converted if the request has 'json' as its body. There's no way to do it except that they change their request body.

Actually, the other request converts are originally designed to handle the body of the request, but I think this request converter is different to the others. If I understand correctly, its goal is to provide convenience to a user to get a parameter or a header from the request easily. 

So, I think that the case you mentioned might be an advanced use case. How about:
- making `shouldHandle` `protected` and the default implementation always return `true`
  (or, adding a `Function` type member as a decision maker?)
- removing `JavaBeanRequestConverterFunction` from the default converters
- adding it by default only if a user doesn't specify it as their request converter

It makes our implementation more complicated, but it makes a user be able to customize the converter. 

Please let me know what you think. Also, let's wait what our folks think. :-)

LiYing2010: hmm....
That sound like too complicated, and hard to understand.
Let's just remove the `shouldHandle` check, and keep it simple for now.
If this kind of feature request exits in the future, let's deal with it then

LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+
+        beanParamInfo.fieldParamList
+                .forEach(
+                        fieldParamInfo -> fieldParamInfo.setValue(ctx, request, httpParams, requestBean)
+                );
+
+        return requestBean;
+    }
+
+    public static void initFor(final Class<?> expectedResultType) {
```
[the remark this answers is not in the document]
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,223 @@
...
+        } catch (Exception e) {
+            throw new IllegalArgumentException(
+                    "Can't convert '" + str + "' to type '" + clazz.getSimpleName() + "'.", e);
+        }
+
+        throw new IllegalArgumentException(
+                "Type '" + clazz.getSimpleName() + "' can't be converted.");
+    }
+
+    public static class EnumConverter<E extends Enum> {
```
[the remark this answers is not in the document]
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
hyangtack: I thought you need to reuse your Java Bean. :-)
Thank you for the change.

hyangtack:
BTW, would you address `checkstyle` errors? You can refer [here](https://github.com/line/armeria/blob/master/CONTRIBUTING.md#intellij-idea) to configure checkstyle on your IDE.

core/src/main/java/com/linecorp/armeria/common/util/Holder.java
```diff
@@ -0,0 +1,63 @@
...
+package com.linecorp.armeria.common.util;
+
+import java.util.function.Supplier;
+
+/**
+ * Mutable object value holder.
+ *
+ * @param <T> Type of value.
+ */
+public final class Holder<T> {
```
trustin: Could you:

- move this class to `server.annotation`
- make this class package-local
- rename this class to `MutableHolder` to emphasize its mutability?

core/src/main/java/com/linecorp/armeria/common/util/Holder.java
```diff
@@ -0,0 +1,63 @@
...
+            value = valueSupplier.get();
+        }
+        return value;
+    }
+
+    public static <T> Holder<T> of(final T value) {
+        return new Holder<>(value);
+    }
+
+    public static <T> Holder<T> ofEmpty() {
```
trustin: Could you move the `static` methods before all non-static methods/members?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,246 @@
...
+import com.linecorp.armeria.common.AggregatedHttpMessage;
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpParameters;
+import com.linecorp.armeria.common.MediaType;
+
+import io.netty.handler.codec.http.HttpConstants;
+import io.netty.handler.codec.http.QueryStringDecoder;
+
+/**
+ * Utility class for AnnotatedHttpService parameters processing.
```
trustin: `{@link AnnotatedHttpService}`

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,246 @@
...
+ */
+public final class AnnotatedHttpServiceParamUtil {
```
trustin: Could we move this class to the `internal` package so it's hidden from the public API?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,246 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(AnnotatedHttpServiceParamUtil.class);
+
+    private AnnotatedHttpServiceParamUtil() { }
```
trustin: Could we move this to the bottom, after all static methods?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,246 @@
...
+        if (value != null) {
+            converted = type.isEnum() ? enumConverter.getEnumElement(value)
+                                      : stringToType(value, type);
+        }
+
+        return optionalWrapped ? Optional.ofNullable(converted) : converted;
+    }
+
+    /**
+     * Validates whether the specified {@link Class} is supported. Throws {@link IllegalArgumentException} if
```
trustin: Could use a `@throws` tag.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * A default implementation of a {@link RequestConverterFunction} which converts the following parameters
+ * of the {@link AggregatedHttpMessage} to an object.
+ *
+ * <p>Path parameters to fields annotated by {@link Param}
+ *
+ * <p>HTTP parameters to fields annotated by {@link Param}
+ *
+ * <p>Headers to fields annotated by {@link Header}
```
trustin: Could you use `<ul><li/></ul>`? e.g.

```html
<ul>
  <li>...</li>
  <li>...</li>
  <li>...</li>
</ul>
```

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+import com.linecorp.armeria.common.HttpParameters;
+import com.linecorp.armeria.common.util.Holder;
+import com.linecorp.armeria.server.AnnotatedHttpServiceParamUtil.EnumConverter;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+import io.netty.util.AsciiString;
+
+/**
+ * A default implementation of a {@link RequestConverterFunction} which converts the following parameters
+ * of the {@link AggregatedHttpMessage} to an object.
```
trustin: When you browse the javadoc, the list of parameters you mentioned using 'the following parameters' will not appear in the summary, making the javadoc look confusing, i.e. The first Javadoc sentence must be self-contained.

How about:

```
A default {@link RequestConverterFunction} which converts a request into a Java bean.
The following parameters are eligible for automatic injection into bean fields or properties:
```

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+ * <p>Path parameters to fields annotated by {@link Param}
+ *
+ * <p>HTTP parameters to fields annotated by {@link Param}
+ *
+ * <p>Headers to fields annotated by {@link Header}
+ */
+public class JavaBeanRequestConverterFunction implements RequestConverterFunction {
+    private static final ConcurrentMap<Class<?>, BeanParamInfo> beanParamInfoCache = new ConcurrentHashMap<>();
+
+    private static final List<Class<?>> ignoredResultTypeList = new ArrayList<>();
```
trustin: Both `beanParamInfoCache` and `ignoredResultTypeList` must have weak or phantom reference key, because otherwise they can lead to class loader leak in a container environment. You could use Guava's `MapMaker` for `beanParamInfoCache` for instance.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Initialization process for a expectedResultType.
+     *
+     * @param expectedResultType a candidate data type, which will converted by JavaBeanRequestConverterFunction
+     */
+    public static synchronized void initFor(final Class<?> expectedResultType) {
```
trustin: I'm curious if this method could be invoked automatically in `convertRequest()` because we won't even need to expose this method in the public API then.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    @Override
+    public Object convertRequest(final ServiceRequestContext ctx,
+                                 final AggregatedHttpMessage request,
+                                 final Class<?> expectedResultType) throws Exception {
+        final BeanParamInfo beanParamInfo = beanParamInfoCache.get(expectedResultType);
+
+        if (beanParamInfo == null) {
+            // no BeanParamInfo for this result type, we should not handled it
```
trustin: handled -> handle

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+        } else {
+            beanParamInfoCache.put(expectedResultType, beanParamInfo);
+        }
+    }
+
+    private static BeanParamInfo buildBeanParamInfo(final Class<?> expectedResultType) {
+        final List<FieldParamInfo> fieldParamList = new ArrayList<>();
+        getAllFields(expectedResultType).forEach(
+                field -> {
+                    final Param param = field.getAnnotation(Param.class);
```
trustin: Do we need to call `field.setAccessible(true)`?

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,241 @@
...
+        if (beanParamInfo == null) {
+            ignoredResultTypeList.add(expectedResultType);
+        } else {
+            beanParamInfoCache.put(expectedResultType, beanParamInfo);
+        }
+    }
+
+    private static BeanParamInfo buildBeanParamInfo(final Class<?> expectedResultType) {
+        final List<FieldParamInfo> fieldParamList = new ArrayList<>();
+        getAllFields(expectedResultType).forEach(
```
trustin: Seems like we do not support the setter methods at the moment. How about adding one in this PR?

trustin:
@LiYing2010 Looks promising. Left some comments. :bowing_man:

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
LiYing2010: according to the code style setting, constructors are arranged before static methods, so, even if I move this to the bottom, it will go back when I reformat the code.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: We need expose it, and call it from `AnnotatedHttpServiceMethod`, to build up `BeanParamInfo` map, at the server starts up phase
see https://github.com/line/armeria/pull/1096#discussion_r174990361

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: Yes, it is done in constructor of class `FieldParamInfo`

core/src/main/java/com/linecorp/armeria/common/util/Holder.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/common/util/Holder.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
LiYing2010: fixed.
But we can not import `AnnotatedHttpService` here, because it is package-private, so I have to write `{@link com.linecorp.armeria.server.AnnotatedHttpService}` in JavaDoc

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.

But looks like there is not something like `WeakReferenceList`, so I renamed `ignoredResultTypeList ` to `ignoredResultTypeCache`, and use `new MapMaker().weakKeys().makeMap()` for it too

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
trustin: I see. How about renaming it to `register` then?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceParamUtil.java
trustin: Ah, I see.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: added support for setter methods

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,323 @@
...
+ * A default {@link RequestConverterFunction} which converts a request into a Java bean.
+ *
+ * <p>The following parameters are eligible for automatic injection into bean fields or properties:
+ * <ul>
+ *   <li>Path parameters to fields annotated by {@link Param}</li>
+ *   <li>HTTP parameters to fields annotated by {@link Param}</li>
+ *   <li>Headers to fields annotated by {@link Header}</li>
+ * </ul>
+ */
+public class JavaBeanRequestConverterFunction implements RequestConverterFunction {
```
trustin: How about renaming to `BeanRequestConverterFunction`, because all we know it's Java? :-)

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,323 @@
...
+    private static final ConcurrentMap<Class<?>, BeanParamInfo> beanParamInfoCache = new MapMaker().weakKeys()
+                                                                                                   .makeMap();
+
+    private static final ConcurrentMap<Class<?>, Boolean> ignoredResultTypeCache = new MapMaker().weakKeys()
+                                                                                                 .makeMap();
+
+    private static final String PREFIX_SETTER_METHOD_NAME = "set";
+
+    /**
+     * register and build up parameters information for {@code expectedResultType}.
```
trustin: Builds and registers the parameter information for ...

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,323 @@
...
+     *
+     * @param expectedResultType a candidate data type, which will converted by JavaBeanRequestConverterFunction
```
trustin: `{@link BeanRequestConverterFunction}`

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,323 @@
...
+                            }
+
+                            final Header header = field.getAnnotation(Header.class);
+                            if (header != null) {
+                                paramList.add(new HeaderFieldParamInfo(field, header));
+                            }
+                        }
+                );
+
+        getAllMethods(expectedResultType, withPrefix(PREFIX_SETTER_METHOD_NAME), withParametersCount(1))
```
trustin: Perhaps we don't need to filter out the methods that do not start with `set`, because we only process the methods annotated with `@Param` or `@Header`?

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
```diff
@@ -0,0 +1,323 @@
...
+        private final Constructor<?> constructor;
+
+        private final List<ParamInfo> paramList;
+
+        private BeanParamInfo(final Class<?> beanClass, final List<ParamInfo> paramList) {
+            this.beanClass = beanClass;
+            this.paramList = paramList;
+
+            try {
+                constructor = beanClass.getDeclaredConstructor();
```
trustin: I guess users may want to avoid using the default constructor but inject the constructor parameter. Could you add a `TODO` item like this:

```java
// TODO: Support injecting constructor parameters.
```

Or, if you feel like, you can implement that in this PR as well.

core/src/main/java/com/linecorp/armeria/server/annotation/Param.java
```diff
@@ -22,10 +22,16 @@
...
+ *
+ * <p>a method parameter
+ *
+ * <p>or, a field of a request bean (See: {@link RequestConverter} and {@link RequestConverterFunction})
+ *
+ * <p>or, a setter method of a request bean (See: {@link RequestConverter} and {@link RequestConverterFunction})
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target(ElementType.PARAMETER)
+@Target({ ElementType.PARAMETER, ElementType.FIELD, ElementType.METHOD })
```
trustin: I wonder if we should use a parameter-scope annotation for setter methods:

```java
@Param("foo")
public void setFoo(String foo) { ... }
// vs.
public void setFoo(@Param("foo") String foo) { ... }
```

I'm not sure which is better, but we could potentially inject more than one params in a single call:

```java
public void setAll(@Param("foo") String foo, @Param("bar") String bar) { .. }
```

What do you think? Which sounds better or do you think we should implement both? (not necessarily all in this PR - no pressure)

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: I will implement it later.

core/src/main/java/com/linecorp/armeria/server/annotation/Param.java
LiYing2010: I will implement it later.

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,320 @@
...
+    private static final ConcurrentMap<Class<?>, BeanParamInfo> beanParamInfoCache = new MapMaker().weakKeys()
+                                                                                                   .makeMap();
+
+    private static final ConcurrentMap<Class<?>, Boolean> ignoredResultTypeCache = new MapMaker().weakKeys()
+                                                                                                 .makeMap();
+
+    /**
+     * Builds and registers the parameter information for {@code expectedResultType}.
+     *
+     * @param expectedResultType a candidate data type, which will be converted by ${@link BeanRequestConverterFunction}
```
minwoox: The length of this line exceeds the limit which is 112.
Don't think we need `$`

core/src/main/java/com/linecorp/armeria/server/annotation/MutableValueHolder.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * If the value is not {@code null}, return it.
+     * If the value is {@code null}, attempts to compute its value using the given supplier function
+     * and set it into this Holder.
+     *
+     * @param valueSupplier the function to compute a value
+     *
+     * @return the current (existing or computed) value, or null if the computed value is null
```
minwoox: nit: `{@code null}` as well, like the above ones?

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        @Override
+        protected void setParamValue(final Object targetBean,
+                                     final Object paramValue) {
+            try {
+                field.set(targetBean, paramValue);
+            } catch (final Exception e) {
+                throw new IllegalArgumentException(
+                        "Can't set '" + paramValue + "' " + "to field '" +
```
minwoox: `paramValue + "' " + "to field '"`  -> `paramValue + "' to field '"`

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        @Override
+        protected void setParamValue(final Object targetBean,
+                                     final Object paramValue) {
+            try {
+                method.invoke(targetBean, paramValue);
+            } catch (final Exception e) {
+                throw new IllegalArgumentException(
+                        "Can't use '" + paramValue + "' " + " to invoke method '" +
```
minwoox: `paramValue + "' " + " to invoke method '"` -> `paramValue + "' to invoke method '"`

core/src/main/java/com/linecorp/armeria/server/annotation/Param.java
LiYing2010: Done.
now `@Param`/`@Header` can be used to bean method and method params

core/src/main/java/com/linecorp/armeria/server/annotation/JavaBeanRequestConverterFunction.java
LiYing2010: Done.
now `@Param`/`@Header` can be used to bean constructor and constructor params

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,253 @@
...
+     * Whether the request is available to be aggregated.
+     */
+    public static boolean aggregationAvailable(HttpHeaders reqHeaders) {
+        final MediaType contentType = reqHeaders.contentType();
+        // We aggregate request stream messages for the media type of form data currently.
+        return contentType != null && MediaType.FORM_DATA.type().equals(contentType.type());
+    }
+
+    /**
+     * Convert a HTTP Parameter String to proper data type.
```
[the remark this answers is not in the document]
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,253 @@
...
+                "Type '" + clazz.getSimpleName() + "' can't be converted.");
+    }
+
+    public static class EnumConverter<T extends Enum<T>> {
+        private final boolean isCaseSensitiveEnum;
+
+        private final Map<String, T> enumMap;
+
+        /**
+         * Create an instance for the given Enum class.
```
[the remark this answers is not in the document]
LiYing2010: fixed.

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
LiYing2010: these code was changed to `String.format('message template', msgParams)`

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
LiYing2010: these code was changed to `String.format('message template', msgParams)`

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
LiYing2010: Fixed the line length problem.

About the `${@link BeanRequestConverterFunction}`, please see this comment: https://github.com/line/armeria/pull/1096#discussion_r176300804

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,546 @@
...
+ *   <li>Path parameters to fields annotated by {@link Param}</li>
+ *   <li>HTTP parameters to fields annotated by {@link Param}</li>
+ *   <li>Headers to fields annotated by {@link Header}</li>
+ * </ul>
+ */
+public class BeanRequestConverterFunction implements RequestConverterFunction {
+    private static final String PACKAGE_NAME_PREFIX_JAVA = "java.";
+
+    private static final ConcurrentMap<Class<?>, BeanParamInfo> beanParamInfoCache = new MapMaker().weakKeys()
+                                                                                                   .makeMap();
```
trustin: After thinking a little bit about this, I realized that the entries in this map will never be garbage-collected because `BeamParamInfo` will have strong references to `Field`, `Method` and `Constructor`, which end up with strong references to its parent `Class`.

I think it's not worth fix this properly by making all reflective fields `WeakReference`s given their performance and complexity problem. How about just using `ConcurrentHashMap` for `beamParamInfoCache` and adding some comment about this?

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,546 @@
...
+    /**
+     * Builds and registers the parameter information for {@code expectedResultType}.
+     *
+     * @param expectedResultType a candidate data type, which will be converted
+     *      by ${@link BeanRequestConverterFunction}
+     */
+    public static synchronized void register(final Class<?> expectedResultType) {
+        final Package pkg = expectedResultType.getPackage();
+        if (pkg != null && pkg.getName().startsWith(PACKAGE_NAME_PREFIX_JAVA)) {
+            // don't handle classes under "java.*" packages
```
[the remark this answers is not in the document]
LiYing2010: because:
1. `@RequestObject` could be used on `java.lang.String`, which should be handled by `StringRequestConverterFunction`
2. it also be used on `java.util.Optional<XXX>`, in unit test: https://github.com/line/armeria/blob/2793fc14ddf9b6ba9a6afef07acbe6b9fecfc61e/core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceRequestConverterTest.java#L83-L84
which should be handled by customized request converter `TestRequestConverterOptional1 `
3. on the other hand, in general, our request java bean should never under the "java.*" packages

So, by skip classes under the "java.*" packages, we can make our server start a little faster.

But no big deal, we can totally remove this skip logic, because there are not too many `@RequestObject` annotated classes.

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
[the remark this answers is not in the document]
LiYing2010: removed

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,546 @@
...
+ * A default {@link RequestConverterFunction} which converts a request into a Java bean.
+ *
+ * <p>The following parameters are eligible for automatic injection into bean fields or properties:
+ * <ul>
+ *   <li>Path parameters to fields annotated by {@link Param}</li>
+ *   <li>HTTP parameters to fields annotated by {@link Param}</li>
+ *   <li>Headers to fields annotated by {@link Header}</li>
+ * </ul>
+ */
+public class BeanRequestConverterFunction implements RequestConverterFunction {
```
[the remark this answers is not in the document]
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/MutableValueHolder.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static <T> MutableValueHolder<T> of(final T value) {
+        return new MutableValueHolder<>(value);
+    }
+
+    public static <T> MutableValueHolder<T> ofEmpty() {
+        return of(null);
+    }
+
+    public T getValue() {
```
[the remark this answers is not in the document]
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/MutableValueHolder.java
```diff
@@ -0,0 +1,63 @@
...
+        return value;
+    }
+
+    public void setValue(final T value) {
```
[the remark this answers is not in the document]
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/MutableValueHolder.java
```diff
@@ -0,0 +1,63 @@
...
+ * @param <T> Type of value.
+ */
+final class MutableValueHolder<T> {
+    private T value;
+
+    private MutableValueHolder(final T value) {
+        this.value = value;
+    }
+
+    public static <T> MutableValueHolder<T> of(final T value) {
```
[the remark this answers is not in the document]
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/MutableValueHolder.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * If the value is not {@code null}, return it.
+     * If the value is {@code null}, attempts to compute its value using the given supplier function
+     * and set it into this Holder.
+     *
+     * @param valueSupplier the function to compute a value
+     *
+     * @return the current (existing or computed) value, or {@code null} if the computed value is {@code null}
+     */
+    public T computeIfAbsent(final Supplier<T> valueSupplier) {
```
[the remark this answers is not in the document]
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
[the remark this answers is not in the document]
trustin: Thanks for explanation! :heart:

trustin:
@LiYing2010 Thank you very much for your patience and high quality contribution! I wonder if you are interested in updating the documentation regarding this new feature. It's fine if you are not. We will create an issue and follow up later.

LiYing2010:
@trustin 
OK, I think I can update documentation related to this.
But I am kind of little busy these days, So, can we create another issue for documentation, and take care of it later?

trustin:
@LiYing2010 Sure. No problem!

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,537 @@
...
+ *   <li>HTTP parameters to fields annotated by {@link Param}</li>
+ *   <li>Headers to fields annotated by {@link Header}</li>
+ * </ul>
+ */
+public final class BeanRequestConverterFunction implements RequestConverterFunction {
+    // In BeanParamInfo, there are strong references to Field/Method/Constructor, which end up with strong
+    // references to their parent Class. So it don't make sense if we use map with weak reference key here
+    private static final ConcurrentMap<Class<?>, BeanParamInfo> beanParamInfoCache = new ConcurrentHashMap<>();
+
+    private static final ConcurrentMap<Class<?>, Boolean> ignoredResultTypeCache = new ConcurrentHashMap<>();
```
minwoox: Could you change this `Map` back to using `MapMaker` because this does not have any strong references?

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,537 @@
...
+
+    /**
+     * Builds and registers the parameter information for {@code expectedResultType}.
+     *
+     * @param expectedResultType a candidate data type, which will be converted
+     *      by ${@link BeanRequestConverterFunction}
```
minwoox: You don't need the `$` in front of `{@link ...`.

minwoox: Could you align `by {@link BeanRequestConverterFunction}` with a candidate?
So it will be:
```
     * @param expectedResultType a candidate data type, which will be converted
     *                           by {@link BeanRequestConverterFunction}
```

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -0,0 +1,253 @@
...
+     * Whether the request is available to be aggregated.
+     */
+    public static boolean aggregationAvailable(HttpHeaders reqHeaders) {
+        final MediaType contentType = reqHeaders.contentType();
+        // We aggregate request stream messages for the media type of form data currently.
+        return contentType != null && MediaType.FORM_DATA.type().equals(contentType.type());
+    }
+
+    /**
+     * Converts a HTTP Parameter String to proper data type.
```
minwoox: a HTTP -> an HTTP (HTTP's pronunciation starts with a vowel.)

core/src/main/java/com/linecorp/armeria/server/annotation/Param.java
```diff
@@ -22,10 +22,26 @@
 import java.lang.annotation.Target;
 
 /**
- * Annotation for mapping a parameter of a request onto a method parameter.
+ * Annotation for mapping a parameter of a request onto the following elements.
+ *
+ * <p>a parameter of an annotated service method
+ *
+ * <p>or, a field of a request bean (See: {@link RequestConverter} and {@link RequestConverterFunction})
```
minwoox: I think we can just leave one `(See: {@link RequestConverter} and {@link RequestConverterFunction})` at the end and remove the others.

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -0,0 +1,537 @@
...
+                // there should be only one parameter
+                if (elementParams.length != 1) {
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "There should be only 1 parameter for '%1$s'.",
+                                    getMemberFullName(element))
+                    );
+                }
+
+                // parameter should has no annotation
```
minwoox: should have?

core/src/main/java/com/linecorp/armeria/server/annotation/Header.java
```diff
@@ -22,10 +22,26 @@
 import java.lang.annotation.Target;
 
 /**
- * Annotation for mapping an HTTP request header onto a method parameter.
+ * Annotation for mapping an HTTP request header onto the following elements.
+ *
+ * <p>a parameter of an annotated service method
+ *
+ * <p>or, a field of a request bean (See: {@link RequestConverter} and {@link RequestConverterFunction})
```
minwoox: I think we can just leave one `(See: {@link RequestConverter} and {@link RequestConverterFunction})` at the end and remove the others.

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
LiYing2010: fixed

LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/Param.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
LiYing2010: fixed

core/src/main/java/com/linecorp/armeria/server/annotation/Header.java
LiYing2010: fixed

trustin:
BOOM! Much appreciated, @LiYing2010 !

PR #1100, opened 2018-03-22
Allow to serve HTTP and HTTPS on a single port

### LINKED ISSUE #1099: Port unification
In some use-case, an HTTPS server needs to handle HTTP messages, but Armeria doesn't allow this case.

hyangtack:
Motivation:
In some environment, a user may need to serve HTTP and HTTPS on a single port, but Armeria has not allowed it.

Modifications:
- Make `ServerPort` have a list of `SessionProtocol` enum values in order to allow to serve HTTP and HTTPS on a single port.
- Add an option for enabling PROXY protocol in order to interoperate with HAProxy well.
- Expose source and destination addresses delivered from PROXY protocol as `ProxiedAddresses` class.

Result:
Closes #1099

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
@@ -22,6 +22,9 @@ import java.util.List;
 
 import javax.annotation.Nullable;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.internal.Http2GoAwayListener;
 import com.linecorp.armeria.internal.ReadSuppressingHandler;
@@ -34,6 +37,9 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelPipeline;
 import io.netty.handler.codec.ByteToMessageDecoder;
+import io.netty.handler.codec.MessageToMessageDecoder;
+import io.netty.handler.codec.haproxy.HAProxyMessage;
+import io.netty.handler.codec.haproxy.HAProxyMessageDecoder;
 import io.netty.handler.codec.http.HttpServerCodec;
 import io.netty.handler.codec.http.HttpServerUpgradeHandler;
 import io.netty.handler.codec.http2.DefaultHttp2Connection;
@@ -55,6 +61,7 @@ import io.netty.handler.ssl.ApplicationProtocolNames;
 import io.netty.handler.ssl.ApplicationProtocolNegotiationHandler;
 import io.netty.handler.ssl.SniHandler;
 import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslHandler;
 import io.netty.util.AsciiString;
 import io.netty.util.DomainNameMapping;
 
@@ -63,16 +70,49 @@ import io.netty.util.DomainNameMapping;
  */
 final class HttpServerPipelineConfigurator extends ChannelInitializer<Channel> {
 
+    private static final Logger logger = LoggerFactory.getLogger(HttpServerPipelineConfigurator.class);
+
+    private static final int SSL_RECORD_HEADER_LENGTH = 5;
+
     private static final AsciiString SCHEME_HTTP = AsciiString.of("http");
     private static final AsciiString SCHEME_HTTPS = AsciiString.of("https");
 
     private static final int UPGRADE_REQUEST_MAX_LENGTH = 16384;
+
+    private static final byte[] PROXY_V1_MAGIC_BYTES = {
+            (byte) 'P',
+            (byte) 'R',
+            (byte) 'O',
+            (byte) 'X',
+            (byte) 'Y',
+            };
+
+    private static final byte[] PROXY_V2_MAGIC_BYTES = {
+            (byte) 0x0D,
+            (byte) 0x0A,
+            (byte) 0x0D,
+            (byte) 0x0A,
+            (byte) 0x00,
+            (byte) 0x0D,
+            (byte) 0x0A,
+            (byte) 0x51,
+            (byte) 0x55,
+            (byte) 0x49,
+            (byte) 0x54,
+            (byte) 0x0A
+    };
+
+    enum PortType {
+        TLS, CLEARTEXT, UNCERTAIN
+    }
 
     private final ServerConfig config;
     private final ServerPort port;
     @Nullable
     private final DomainNameMapping<SslContext> sslContexts;
     private final GracefulShutdownSupport gracefulShutdownSupport;
+
+    private final PortType portType;
 
     /**
      * Creates a new instance.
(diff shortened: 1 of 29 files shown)


core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -251,7 +250,19 @@ public ServerBuilder port(InetSocketAddress localAddress, SessionProtocol protoc
      * @see <a href="#no_port_specified">What happens if no HTTP(S) port is specified?</a>
      */
     public ServerBuilder port(ServerPort port) {
-        ports.add(requireNonNull(port, "port"));
+        requireNonNull(port, "port");
+        for (int i = 0; i < ports.size(); i++) {
+            final ServerPort p = ports.get(i);
+            if (p.equals(port)) {
```
trustin: Could we compare only the port number here and make `ServerPort.equals()` returns when both port number and protocols match?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -251,7 +250,19 @@ public ServerBuilder port(InetSocketAddress localAddress, SessionProtocol protoc
...
+                final Set<SessionProtocol> protocols =
+                        new ImmutableSet.Builder<SessionProtocol>().addAll(p.protocols())
+                                                                   .addAll(port.protocols())
+                                                                   .build();
```
trustin: This will raise an exception when a user attempts this:

```java
sb.http(8080);
sb.http(8080);
```

Could we quietly deduplicate?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -450,6 +461,29 @@ public ServerBuilder accessLogWriter(Consumer<? super RequestLog> accessLogWrite
         return this;
     }
 
+    /**
+     * Enables PROXY protocol support.
+     */
+    public ServerBuilder acceptProxyProtocol() {
```
trustin: How about just `proxyProtocol(boolean enableProxyProtocol)`?

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -430,6 +435,21 @@ public String serviceLoggerPrefix() {
         return accessLogWriter;
     }
 
+    /**
+     * Returns whether PROXY protocol is enabled.
+     */
+    public boolean acceptProxyProtocol() {
```
trustin: `proxyProtocolEnabled()`?

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -430,6 +435,21 @@ public String serviceLoggerPrefix() {
...
+        return acceptProxyProtocol;
+    }
+
+    /**
+     * Returns the maximum size of additional data (TLV, Tag-Length-Value). It is only used when
+     * {@link #acceptProxyProtocol()} is {@code true}.
+     */
+    public int proxyProtocolMaxTlvSize() {
```
trustin: Could return `null` or some invalid value (0? -1?) if PROXY protocol is disabled.

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -36,25 +39,41 @@
...
+        this(new InetSocketAddress(port), ImmutableSet.copyOf(protocols));
     }
 
     /**
      * Creates a new {@link ServerPort} that listens to the specified {@code localAddress} using the specified
-     * {@link SessionProtocol}.
+     * {@link SessionProtocol}s.
+     */
+    public ServerPort(InetSocketAddress localAddress, SessionProtocol... protocols) {
+        this(localAddress, ImmutableSet.copyOf(protocols));
```
trustin: `requireNonNull` on `protocols`

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -36,25 +39,41 @@
...
 
     /**
      * Creates a new {@link ServerPort} that listens to the specified {@code port} of all available network
-     * interfaces using the specified {@link SessionProtocol}.
+     * interfaces using the specified {@link SessionProtocol}s.
      */
-    public ServerPort(int port, SessionProtocol protocol) {
-        this(new InetSocketAddress(port), protocol);
+    public ServerPort(int port, SessionProtocol... protocols) {
+        this(new InetSocketAddress(port), ImmutableSet.copyOf(protocols));
```
trustin: `requireNonNull` on `protocols`

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -85,9 +103,41 @@ public InetSocketAddress localAddress() {
...
+     * Returns the {@link SessionProtocol}s this {@link ServerPort} uses.
+     */
+    public Set<SessionProtocol> protocols() {
+        return protocols;
+    }
+
+    /**
+     * Returns whether there is a {@link SessionProtocol} which is over TLS.
+     */
+    public boolean anyProtocolOverTls() {
```
trustin: `hasTls()`?

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -85,9 +103,41 @@ public InetSocketAddress localAddress() {
...
+        return protocols.stream().anyMatch(SessionProtocol::isTls);
+    }
+
+    /**
+     * Returns whether all {@link SessionProtocol}s are over TLS.
+     */
+    public boolean allProtocolOverTls() {
+        return protocols.stream().allMatch(SessionProtocol::isTls);
```
trustin: `isTlsOnly()`?

core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
```diff
@@ -94,21 +105,30 @@ protected void initChannel(Channel ch) throws Exception {
...
         p.addLast(ReadSuppressingHandler.INSTANCE);
 
-        if (port.protocol().isTls()) {
-            assert sslContexts != null;
-            p.addLast(new SniHandler(sslContexts));
-            p.addLast(TrafficLoggingHandler.SERVER);
-            configureHttps(p);
+        if (config.acceptProxyProtocol()) {
+            p.addLast(new HAProxyMessageDecoder(config.proxyProtocolMaxTlvSize()));
+            p.addLast(new HAProxyMessageConsumer());
```
trustin: Could we auto-detect between HTTPS, PROXY and HTTP instead of accepting only PROXY protocol?

trustin: Perhaps we could add `SessionProtocol.PROXY`? e.g.

- If `ServerPort` contains `PROXY` and `HTTP`, support HTTP and proxied HTTP.
- If `ServerPort` contains `PROXY` and `HTTPS`, support HTTPS and proxied HTTPS.
- If `ServerPort` contains `PROXY`, `HTTP` and `HTTPS`, support HTTP, HTTPS, proxied HTTP and proxied HTTPS.
- If `ServerPort` contains `PROXY` only, `ServerBuilder.build()` fails.

Then we could replace `allowProxyProtocol` with port configuration methods such as:

```java
sb.http(443).https(443).haproxy(443);
```

hyangtack: Good suggestion! I will change my code with `SessionProtocol.PROXY`. :-)

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
hyangtack: It doesn't raise an exception when adding duplicate elements as Javadoc describes. :-)
```
    /**
     * Adds each element of {@code elements} to the {@code ImmutableSet}, ignoring duplicate
     * elements (only the first duplicate element is added).
     *
     * @param elements the {@code Iterable} to add to the {@code ImmutableSet}
     * @return this {@code Builder} object
     * @throws NullPointerException if {@code elements} is null or contains a null element
     */
```

trustin: Oops! Thanks for education. It was `ImmutableMap` builder which disallowed a duplicate.

trustin:
Cannot merge due to a conflict. Will send a new PR based on this PR, because he is on vacation and will be back the week after next week.

PR #1101, opened 2018-03-23
Allow specifying multiple protocols on a single port / Add PROXY protocol support

### LINKED ISSUE #1099: Port unification

trustin:
Originally written by @hyangtack in #1100 

Motivation:

In some environment, a user may need to serve HTTP and HTTPS on a single port, but Armeria does not allow it.

Modifications:
- Make `ServerPort` have a list of `SessionProtocol`s in order to allow to serve HTTP and HTTPS on a single port.
- Add support for PROXY protocol in order to interoperate with HAProxy well.
- Expose source and destination addresses delivered from PROXY protocol as `ProxiedAddresses`.

Result:
- Closes #1099


core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
```diff
@@ -93,22 +120,36 @@ protected void initChannel(Channel ch) throws Exception {
...
-            p.addLast(new SniHandler(sslContexts));
-            p.addLast(TrafficLoggingHandler.SERVER);
-            configureHttps(p);
-        } else {
-            p.addLast(TrafficLoggingHandler.SERVER);
-            configureHttp(p);
+    private void configurePipeline(ChannelPipeline p, Set<SessionProtocol> protocols,
+                                   @Nullable ProxiedAddresses proxiedAddresses) {
+        if (protocols.size() == 1) {
+            switch (protocols.iterator().next()) {
```
anuraaga: Can use `Iterables.getFirst` which is a little more semantic.

core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
```diff
@@ -146,10 +190,152 @@ private Http2ConnectionHandler newHttp2ConnectionHandler(ChannelPipeline pipelin
...
+        private final EnumSet<SessionProtocol> candidates;
+        @Nullable
+        private final EnumSet<SessionProtocol> nextCandidates;
+        @Nullable
+        private final ProxiedAddresses proxiedAddresses;
+
+        ProtocolDetectionHandler(Set<SessionProtocol> protocols, @Nullable ProxiedAddresses proxiedAddresses) {
+            candidates = EnumSet.copyOf(protocols);
+            if (protocols.contains(PROXY)) {
+                // NB: We copy from 'candidates' rather than from 'protocols'
```
anuraaga: This seems somewhat obvious can consider leaving out.

core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
```diff
@@ -146,10 +190,152 @@ private Http2ConnectionHandler newHttp2ConnectionHandler(ChannelPipeline pipelin
         return handler;
     }
 
+    private final class ProtocolDetectionHandler extends ByteToMessageDecoder {
+
+        private final EnumSet<SessionProtocol> candidates;
+        @Nullable
+        private final EnumSet<SessionProtocol> nextCandidates;
```
anuraaga: Since they're only used with PROXY, how about `proxiedCandidates`?

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -640,13 +642,15 @@ public void operationComplete(ChannelFuture f) throws Exception {
...
-                } else {
-                    logger.info("Serving {} at {}", port.protocol().name(), localAddress);
+                if (logger.isInfoEnabled()) {
+                    if (localAddress.getAddress().isAnyLocalAddress() ||
+                        localAddress.getAddress().isLoopbackAddress()) {
+                        port.protocols().forEach(p -> logger.info(
+                                "Serving {} at {} - {}://127.0.0.1:{}/",
+                                p.name(), localAddress, p.uriText(), localAddress.getPort()));
+                    } else {
+                        logger.info("Serving {} at {}", Joiner.on('+').join(port.protocols()), localAddress);
```
anuraaga: `String.join`

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -665,7 +669,10 @@ private static String bossThreadName(ServerPort port) {
 
         // e.g. 'armeria-boss-http-*:8080'
         //      'armeria-boss-http-127.0.0.1:8443'
-        return "armeria-boss-" + port.protocol().uriText() + '-' + localHostName + ':' + localAddr.getPort();
+        //      'armeria-boss-proxy+http+https-127.0.0.1:8443'
+        final StringJoiner joiner = new StringJoiner("+");
+        port.protocols().forEach(p -> joiner.add(p.uriText()));
```
anuraaga: `port.protocols().stream().map(SessionProtocl::uriText).collect(Collectors.joining("+"))`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -191,8 +194,33 @@ public ServerBuilder https(int port) {
...
      */
     public ServerBuilder https(InetSocketAddress localAddress) {
-        ports.add(new ServerPort(requireNonNull(localAddress, "localAddress"), HTTPS));
-        return this;
+        return port(new ServerPort(requireNonNull(localAddress, "localAddress"), HTTPS));
+    }
+
+    /**
+     * Adds a <a href="https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt">PROXY protocol</a> port
+     * that listens on all available network interfaces. It should be specified with HTTP(S) port.
```
anuraaga: `specified along with`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -251,7 +277,19 @@ public ServerBuilder port(InetSocketAddress localAddress, SessionProtocol protoc
      * @see <a href="#no_port_specified">What happens if no HTTP(S) port is specified?</a>
      */
     public ServerBuilder port(ServerPort port) {
-        ports.add(requireNonNull(port, "port"));
+        requireNonNull(port, "port");
+        for (int i = 0; i < ports.size(); i++) {
+            final ServerPort p = ports.get(i);
+            if (p.localAddress().equals(port.localAddress())) {
+                final Set<SessionProtocol> protocols =
+                        new ImmutableSet.Builder<SessionProtocol>().addAll(p.protocols())
```
anuraaga: Consider `Sets.immutableEnumSet`

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -66,14 +92,17 @@ public ServerPort(InetSocketAddress localAddress, SessionProtocol protocol) {
...
-        requireNonNull(protocol, "protocol");
-        checkArgument(protocol == HTTP || protocol == HTTPS,
-                      "protocol: %s (expected: %s or %s)", protocol, HTTP, HTTPS);
-
+        requireNonNull(protocols, "protocols");
+        checkArgument(protocols.stream().allMatch(p -> p == HTTP || p == HTTPS || p == PROXY),
+                      "protocol: %s (expected: %s, %s or %s)", protocols, HTTP, HTTPS, PROXY);
         this.localAddress = localAddress;
-        this.protocol = protocol;
+        this.protocols = ImmutableSortedSet.copyOf(protocols);
```
anuraaga: If you use `immutableEnumSet`, I don't think you need to use `ImmutableSortedSet`

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -66,14 +92,17 @@ public ServerPort(InetSocketAddress localAddress, SessionProtocol protocol) {
...
+
+        final StringJoiner protocolNameJoiner = new StringJoiner("+");
+        this.protocols.forEach(p -> protocolNameJoiner.add(p.uriText()));
```
anuraaga: Can use `.stream()` (may want a utility method for this, a static method on `SessionProtocol` seems reasonable)

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -85,9 +114,56 @@ public InetSocketAddress localAddress() {
 
     /**
      * Returns the {@link SessionProtocol} this {@link ServerPort} uses.
+     *
+     * @deprecated Use {@link #protocols()}.
      */
+    @Deprecated
     public SessionProtocol protocol() {
-        return protocol;
+        return protocols.iterator().next();
```
anuraaga: `Iterables.getFirst`

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -85,9 +114,56 @@ public InetSocketAddress localAddress() {
...
+     */
+    public boolean hasProxyProtocol() {
+        return hasProtocol(PROXY);
+    }
+
+    /**
+     * Returns whether the specified {@code protocol} is in the list of {@link SessionProtocol}s.
+     */
+    public boolean hasProtocol(SessionProtocol protocol) {
+        return protocols.stream().anyMatch(p -> p == protocol);
```
anuraaga: `protocols.contains(protocol)`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -251,7 +277,19 @@ public ServerBuilder port(InetSocketAddress localAddress, SessionProtocol protoc
      * @see <a href="#no_port_specified">What happens if no HTTP(S) port is specified?</a>
      */
     public ServerBuilder port(ServerPort port) {
```
anuraaga: How common do you think it is to have `HTTP` and `HTTPS` on the same port? I sort of want to log something when both are enabled since doing it accidentally could be a major security hole.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -450,6 +488,19 @@ public ServerBuilder accessLogWriter(Consumer<? super RequestLog> accessLogWrite
...
+    /**
+     * Sets the maximum size of additional data for PROXY protocol. The default value of this property is
+     * {@value #PROXY_PROTOCOL_DEFAULT_MAX_TLV_SIZE}.
+     *
+     * <p><b>Note:</b> limiting TLV size only affects processing of v2, binary headers. Also, as allowed by
+     * the 1.5 spec, TLV data is currently ignored. For maximum performance, it would be best to configure
+     * your upstream proxy host to <b>NOT</b> send TLV data and set this property to {@code 0}.
+     */
+    public ServerBuilder proxyProtocolMaxTlvSize(int proxyProtocolMaxTlvSize) {
+        this.proxyProtocolMaxTlvSize = proxyProtocolMaxTlvSize;
```
minwoox: No need to use `validateNonNegative`?

core/src/main/java/com/linecorp/armeria/server/Server.java
trustin: `SessionProtocol` is not a `CharSequence`.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: Can't find a simple way to create an immutable enum set from 1 enum and multiple enums. Ended up converting in `ServerPort` constructor.

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
trustin: Used `Sets.immutableEnumSet` here.

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
trustin: Didn't need to use string joiner at all in this case. Will just use `Set.toString()`.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: Added `checkArgument`

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/Server.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
trustin: Renamed

core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
trustin: Removed :-)

core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
trustin: Fixed. Could have been better if there's a version that throws an exception when the iterable is empty.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: What do you think about adding a method `usePortUnification()` and allowing only when it's called in prior?

trustin: Added `allowPortUnification()`:

```java
sb.allowPortUnification(PROXY, HTTPS);
sb.proxyProtocol(443); // OK
sb.https(443); // OK
sb.http(443); // Not OK
```

@minwoox @anuraaga WDYT?

trustin: Actually PROXY has to be allowed even without allowPortUnification. Let me update the PR.

trustin: Updated:

```java
sb.proxyProtocol(443); // OK
sb.https(443); // OK
sb.http(443); // Not OK

ServerBuilder sb2 = new ServerBuilder();
sb2.usePortUnification();
sb2.proxyProtocol(443); // OK
sb2.https(443); // OK
sb2.http(443); // OK
```

trustin:
Added Javadoc and updated the web site to mention PROXY protocol support.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: The current API doesn't fit very well when it comes to ephemeral port (0):

```java
sb.proxyProtocol(0);
sb.http(0);
sb.https(0);
``

The code above gives us various questions:

- Should `http(0)` and `https(0)` be bound to the same port?
- Where should `proxyProtocol(0)` be applied, to HTTP, HTTPS or both?

Let me:

- Remove `usePortUnification()`,
- Remove `proxyProtocol()` and
- Add `port(int, SessionProtocol...)` and its variants which would be the only way to enable PROXY protocol and/or port unification.

File: core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java [hunk 14]
```diff
@@ -58,6 +58,9 @@
     @Nullable
     private final SSLSession sslSession;
 
+    @Nullable
+    private final ProxiedAddresses proxiedAddresses;
```
anuraaga: Just wondering, do you think we can use this mechanism to handle `X-Forwarded-For` as well or would that be completely at the `Logger` layer?

trustin: We wanna evolve this into something more generic so that it provides the access to the `X-Forwarded-For` headers. Just didn't expand that far in this PR. :-)

anuraaga: Cool - IIUC, we don't have headers available yet at this point so thought it might be difficult, but we'll get to that later

trustin: `HttpRequest` is given when `DefaultServiceRequestContext` is created, so I think it's trivial to get any request header.

trustin:
@minwoox @anuraaga PR updated

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -209,23 +210,51 @@ public ServerBuilder port(int port, String protocol) {
...
-     * interfaces using the specified {@link SessionProtocol}.
+     * interfaces using the specified {@link SessionProtocol}s. Specify multiple protocols to serve more than
+     * one protocol on the same port:
      *
-     * @deprecated Use {@link #http(int)} or {@link #https(int)}.
-     * @see <a href="#no_port_specified">What happens if no HTTP(S) port is specified?</a>
+     * <pre>{@code
+     * ServerBuilder sb = new ServerBuilder();
+     * // Serve both HTTP and HTTPS at port 8080.
+     * sb.port(new InetSocketAddress(8080),
```
minwoox: `sb.port(new InetSocketAddress(8080),` -> `sb.port(8080,`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -209,23 +210,51 @@ public ServerBuilder port(int port, String protocol) {
...
+     *         SessionProtocol.HTTP,
+     *         SessionProtocol.HTTPS);
+     * // Enable HTTPS with PROXY protocol support at port 8443.
+     * sb.port(new InetSocketAddress(8443),
```
minwoox: `sb.port(new InetSocketAddress(8443),` -> `sb.port(8443,`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -209,23 +210,51 @@ public ServerBuilder port(int port, String protocol) {
...
+     * Adds a new {@link ServerPort} that listens to the specified {@code port} of all available network
+     * interfaces using the specified {@link SessionProtocol}s. Specify multiple protocols to serve more than
+     * one protocol on the same port:
+     *
+     * <pre>{@code
+     * ServerBuilder sb = new ServerBuilder();
+     * // Serve both HTTP and HTTPS at port 8080.
+     * sb.port(new InetSocketAddress(8080),
+     *         SessionProtocol.HTTP,
+     *         SessionProtocol.HTTPS);
```
minwoox: This method takes a `port` and `Iterable<SessionProtocol> protocols`, but the example is little different.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -234,15 +263,42 @@ public ServerBuilder port(InetSocketAddress localAddress, String protocol) {
...
+
+    /**
+     * Adds a new {@link ServerPort} that listens to the specified {@code localAddress} using the specified
+     * {@link SessionProtocol}s. Specify multiple protocols to serve more than one protocol on the same port:
+     *
+     * <pre>{@code
+     * ServerBuilder sb = new ServerBuilder();
+     * // Serve both HTTP and HTTPS at port 8080.
+     * sb.port(new InetSocketAddress(8080),
+     *         SessionProtocol.HTTP,
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/server/ServerPort.java
```diff
@@ -66,14 +92,20 @@ public ServerPort(InetSocketAddress localAddress, SessionProtocol protocol) {
...
-
-        localAddressString = localAddress.getAddress().getHostAddress() + ':' + localAddress.getPort();
+        this.protocols = Sets.immutableEnumSet(protocols);
+
+        checkArgument(!this.protocols.isEmpty(),
+                      "protocols: %s (must not be empty)", this.protocols);
+        checkArgument(this.protocols.contains(HTTP) || this.protocols.contains(HTTPS),
+                      "protocols: %s (must contain HTTP or HTTPS)", this.protocols);
+        checkArgument(this.protocols.stream().allMatch(p -> p == HTTP || p == HTTPS || p == PROXY),
+                      "protocols: %s (must not contains other than %s, %s or %s)",
```
minwoox: contains -> contain

site/src/sphinx/server-basics.rst
```diff
@@ -162,6 +162,26 @@ You can also add an HTTPS port with your certificate and its private key files:
       .tls(new File("certificate.crt"), new File("private.key"), "myPassphrase");
     ...
 
+
+PROXY protocol
+--------------
+
+Armeria supports both text (v1) and binary (v2) variant of `PROXY protocol <https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt/>`_.
```
minwoox: variants?

trustin:
@anuraaga @minwoox Fixed all. Please review again.

trustin:
Thanks, reviewers!

PR #1107, opened 2018-03-28
Update Tomcat to 9.0.6 / Replace TomcatHandler with MethodHandles

### LINKED ISSUE #790: Upgrade Tomcat to 9.0
Tomcat 9.0 has been released.

trustin:
- Update Tomcat to 9.0.6
  - Make JarSubsetResourceSet work for both Tomcat 8.5 and 9.0 by using
    MethodHandles
  - Make TomcatUtil.getEngine() work for both Tomcat 8.5 and 9.0 by
    creating a new Engine when a Service has no Engine
  - Add a test-only project 'tomcat8.5' which tests ABI compatibility
  - Rename Tomcat85* to Tomcat90*
  - Fixes #790

- Replace TomcatHandler with MethodHandles
  - This will fix the compilation errors when building from an IDE.


imasahiro:
I didn't check but if I remember correctly, it has a memory leak issue that is not back ported to JDK8, JDK9.
https://bugs.java.com/view_bug.do?bug_id=JDK-8174749
It is small amount of memory. So we can ignore about that but how do you think?

Related commit: https://github.com/ben-manes/caffeine/commit/20d9e0cfab5a8dde64e9e7f96a344f4c77a6d410

trustin:
@imasahiro I'm curious 'when' a leak occurs. I guess it's only when creating a `MethodHandle` and not when making an invocation, right? If so, I'd stick to `MethodHandle`.

trustin:
@imasahiro Anything else?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/Tomcat90OutputBuffer.java
```diff
@@ -25,15 +25,14 @@
...
+class Tomcat90OutputBuffer implements OutputBuffer {
     private final Queue<HttpData> data;
     private long bytesWritten;
 
-    Tomcat85OutputBuffer(Queue<HttpData> data) {
+    Tomcat90OutputBuffer(Queue<HttpData> data) {
         this.data = data;
     }
 
-    @Override
```
minwoox: // Required for 8.5. ?

imasahiro:
Checked the original issue and IIUC, leaks are happened only when finding a handler. Then LGTM now.

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/Tomcat90OutputBuffer.java
trustin: Good point. Fixed

trustin:
Thanks, reviewers!

PR #1113, opened 2018-04-01
Remove unthrown Exceptions from GrpcClientTest

dongjinleekr:


trustin:
Thanks, @dongjinleekr

Issue #1112, opened 2018-03-30
ArmeriaHttpUtil.toArmeria allocates tons of garbage


anuraaga:
Currently, armeria converts netty headers to armeria headers. In some profiling, I noticed tons of `byte[]` garbage is created by `AsciiString.toString()`, because the value type of our headers object is `String`. Would it make sense to change that to `CharSequence` since many use cases are satisfied by it? Alternatively, I'm thinking of putting a cache of `AsciiString` to `String` there to reduce garbage. Out of around 20GB of garbage in a profiling run, 4GB is these headers I think, which is huge for such boring strings like `http`, `https`.

JFR:
[flight_recording_downstream.zip](https://github.com/line/armeria/files/1863483/flight_recording_downstream.zip)

anuraaga:
Sent #1114 as a strawman's proposal - changing the interface of `HttpHeaders` seems pretty intense and not clearly a win.

I'm also wondering whether we should have `DefaultHttpHeaders` contain an `Http2Headers` field instead of creating a new `HttpHeaders` - I noticed a fair bit of allocation of new `HeaderEntry` objects when we copy from the netty object to armeria one. I think it'd be overall better for memory efficiency to wrap instead of copy.

trustin:
Solved partially by #1114, but needs a fundamental fix. Need to think a little bit..

trustin:
gRPC seems to use an alternative `Http2Headers` implementation which serves their use case better: `io.grpc.netty.GrpcHttp2HeadersUtils`, `io.grpc.netty.GrpcHttp2OutboundHeaders`. I think we should follow the same path.

I tried to modify `DefaultHttpHeaders` wrap `DefaultHttp2Headers` to reduce some garbage and was able to gain 2-3% throughput. A better implementation would yield even better numbers.

anuraaga:
Ah I wonder if our branches are similar, I've been slowly working on wrapping too

https://github.com/line/armeria/compare/master...anuraaga:netty-http-headers?expand=1

I wondered if it's worth it without us changing the definition of `HttpHeaders` so the value is `CharSequence` to reduce string conversion and took a break before trying such a large change ;)

I also happened to find https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java but didn't have good results reproducing the generality of our header conversion logic using it yet.

trustin:
[...]

PR #1114, opened 2018-04-02
Allow caching header values to avoid creating new Strings for the sam…

anuraaga:
…e value.

Currently, header values are copied from `AsciiString` to `String` all the time, which results in a copy of the underlying storage. As header values tend to be fairly constant, this is quite wasteful as it results in a lot of garbage on the heap with similar content. With a cache of `String` instances for commonly low-variance cache values, we can avoid this extra allocation.

For #1112


trustin:
Thanks, @anuraaga !

PR #1115, opened 2018-04-02
Allow empty lambdas in the check style

minwoox:
Related to https://github.com/line/centraldogma/pull/176


trustin:
Thanks, @minwoox !

PR #1118, opened 2018-04-04
Allow aggregating pooled objects.

anuraaga:
For advanced use cases, including inside armeria itself, it is nice to have the option to aggregate without copying into heap buffers. 

Also removed the thread-local thrift protocol factories since it seemed like saving a couple of objects isn't worth the thread-local overhead of the read and very complex code - let me know if you'd like me to restore it though.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpMessageAggregator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpMessageAggregator.java
@@ -26,18 +26,28 @@ import javax.annotation.Nullable;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
 
+import com.linecorp.armeria.unsafe.ByteBufHttpData;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.ByteBufHolder;
 import io.netty.util.ReferenceCountUtil;
 
 abstract class HttpMessageAggregator implements Subscriber<HttpObject>, BiConsumer<Void, Throwable> {
 
     private final CompletableFuture<AggregatedHttpMessage> future;
     private final List<HttpData> contentList = new ArrayList<>();
+    // ByteBufAllocator is only provided when pooled objects are requested.
+    @Nullable
+    private final ByteBufAllocator alloc;
     private int contentLength;
     @Nullable
     private Subscription subscription;
 
-    protected HttpMessageAggregator(CompletableFuture<AggregatedHttpMessage> future) {
+    HttpMessageAggregator(CompletableFuture<AggregatedHttpMessage> future,
+                          @Nullable ByteBufAllocator alloc) {
         this.future = future;
+        this.alloc = alloc;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
```diff
@@ -355,10 +356,54 @@ default HttpRequest authority(String authority) {
...
+        return future;
+    }
+
+    /**
+     * Aggregates this request. The returned {@link CompletableFuture} will be notified when the content and
+     * the trailing headers of the request is received fully. {@link AggregatedHttpMessage#content()} will
+     * return a pooled object, and the caller must ensure to release it. If you don't know what this means,
+     * use {@link HttpResponse#aggregate()}.
+     */
+    default CompletableFuture<AggregatedHttpMessage> aggregateWithPooledObjects(RequestContext ctx) {
```
trustin: How about removing this method? Although it may be convenient, it introduces an unnecessary dependency to the context API.

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
```diff
@@ -355,10 +356,54 @@ default HttpRequest authority(String authority) {
...
         return future;
     }
+
+    /**
+     * Aggregates this request. The returned {@link CompletableFuture} will be notified when the content and
+     * the trailing headers of the request is received fully. {@link AggregatedHttpMessage#content()} will
+     * return a pooled object, and the caller must ensure to release it. If you don't know what this means,
+     * use {@link HttpResponse#aggregate()}.
+     */
+    default CompletableFuture<AggregatedHttpMessage> aggregateWithPooledObjects(ByteBufAllocator alloc) {
```
trustin: I wonder if it's a good (or bad) idea to rename this method to `aggregate()`. Thoughts? Should `StreamMessage.subscribe(..., boolean)` be replaced with `subscribeWithPooledObjects()`?

thrift/src/test/java/com/linecorp/armeria/server/thrift/ThriftServiceTest.java
```diff
@@ -126,6 +133,16 @@ public ThriftServiceTest(SerializationFormat defaultSerializationFormat) {
...
     }
 
+    @BeforeClass
+    public static void setupLoop() {
+        eventLoop = new DefaultEventLoop(Executors.newSingleThreadExecutor());
+    }
+
+    @AfterClass
+    public static void closeLoop() {
+        eventLoop.shutdownGracefully().syncUninterruptibly();
```
trustin: Probably OK not to call `syncUninterruptably()`.

core/src/main/java/com/linecorp/armeria/common/HttpMessageAggregator.java
```diff
@@ -109,14 +119,32 @@ public void accept(Void unused, Throwable cause) {
...
-                System.arraycopy(data.array(), data.offset(), merged, offset, dataLength);
-                offset += dataLength;
+            if (alloc != null) {
+                final ByteBuf merged = alloc.buffer(contentLength);
+                for (int i = 0; i < contentList.size(); i++) {
+                    final HttpData data = contentList.set(i, null);
+                    if (data instanceof ByteBufHolder) {
+                        ByteBufHolder byteBufData = (ByteBufHolder) data;
+                        try {
+                            merged.writeBytes(byteBufData.content());
```
hyangtack: Just curious. Is [CompositeByteBuf](https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java) not suitable for this case?

anuraaga: I have gone back and forth on usage of `CompositeByteBuf` - I feel like it's great when the number of buffers is unbounded since the number of copies is unknown, but when we know the length, trading off exactly one copy to avoid the overhead of `CompositeByteBuf` (ArrayList, managing indexes over buffer boundaries) is probably often worth it. Perfect might be adding a heuristic that uses it based on the number of elements / contentLength here, but getting quite deep at that point.

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
anuraaga: I went back and forth on this many times actually, and had the same question about `subscribe`. Here, the `alloc` is needed for efficient aggregation, and using a `@Nullable` alloc to control pooled or not seems quite unclear and possibly dangerous (when it's a parameter passing in an alloc is almost always the right thing to do since it normally indicates a safe optimization, but not here).

`subscribe` doesn't have the same semantic limitation, though renaming it might make sense for consistency

PR #1119, opened 2018-04-04
Improve the cleanup of streaming responses

trustin:
Motivation:

A successfully sent response can be logged with AbortedStreamException
or ClosedChannelException due to race conditions.

Modifications:

- Make HttpResponseSubscriber handle AbortedStreamException and
  ClosedChannelException better so that they are not logged
  unnecessarily
- Make HttpServerHandler clean up unfinished HTTP/1 responses a little
  bit later so that they are not aborted too early.
- Miscellaneous:
  - Fix a bug in DeferredStreamMessage where NeverInvokedSubscriber is
    invoked, which happens when a DeferredStreamMessage is delegated to
    another DeferredStreamMessage.

Result:

- Much less chance of redundant AbortedStreamException or
  ClosedChannelException

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.server;
 
+import java.nio.channels.ClosedChannelException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
@@ -36,6 +37,8 @@ import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.HttpStatusClass;
 import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.common.logging.RequestLogBuilder;
+import com.linecorp.armeria.common.stream.AbortedStreamException;
+import com.linecorp.armeria.internal.Http1ObjectEncoder;
 import com.linecorp.armeria.internal.HttpObjectEncoder;
 
 import io.netty.channel.Channel;
@@ -209,7 +212,7 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                 return;
         }
 
-        write(o, endOfStream, true);
+        write(o, endOfStream);
     }
 
     @Override
@@ -232,6 +235,11 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
             failAndRespond(cause,
                            AggregatedHttpMessage.of(((HttpStatusException) cause).httpStatus()),
                            Http2Error.CANCEL);
+        } else if (cause instanceof AbortedStreamException) {
+            // One of the two cases:
+            // - Client closed the connection too early.
+            // - Response publisher aborted the stream.
+            failAndReset((AbortedStreamException) cause);
         } else {
             logger.warn("{} Unexpected exception from a service or a response publisher: {}",
                         ctx.channel(), service(), cause);
@@ -254,26 +262,29 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
         }
 
         if (state != State.DONE) {
-            write(HttpData.EMPTY_DATA, true, true);
+            write(HttpData.EMPTY_DATA, true);
         }
     }
 
-    private void write(HttpObject o, boolean endOfStream, boolean flush) {
+    private void write(HttpObject o, boolean endOfStream) {
         final Channel ch = ctx.channel();
         if (endOfStream) {
             setDone();
         }
 
-        ch.eventLoop().execute(() -> write0(o, endOfStream, flush));
+        ch.eventLoop().execute(() -> write0(o, endOfStream));
     }
 
-    private void write0(HttpObject o, boolean endOfStream, boolean flush) {
+    private void write0(HttpObject o, boolean endOfStream) {
         final ChannelFuture future;
+        final boolean wroteEmptyData;
         if (o instanceof HttpData) {
             final HttpData data = (HttpData) o;
+            wroteEmptyData = data.isEmpty();
             future = responseEncoder.writeData(ctx, req.id(), req.streamId(), data, endOfStream);
             logBuilder().increaseResponseLength(data.length());
         } else if (o instanceof HttpHeaders) {
+            wroteEmptyData = false;
             future = responseEncoder.writeHeaders(ctx, req.id(), req.streamId(), (HttpHeaders) o, endOfStream);
         } else {
             // Should never reach here because we did validation in onNext().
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -232,6 +235,11 @@ public void onError(Throwable cause) {
             failAndRespond(cause,
                            AggregatedHttpMessage.of(((HttpStatusException) cause).httpStatus()),
                            Http2Error.CANCEL);
+        } else if (cause instanceof AbortedStreamException) {
+            // One of the two cases:
+            // - Client closed the connection too early.
+            // - Response publisher aborted the stream.
+            failAndReset((AbortedStreamException) cause);
```
trustin: /cc @kojilin You may want to send a canary to make sure this PR fixes your problem.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -197,16 +197,41 @@ public int unfinishedRequests() {
     }
 
     @Override
-    public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
-        super.channelUnregistered(ctx);
-        unfinishedResponses.keySet().forEach(StreamMessage::abort);
```
trustin: No idea why `channelUnregistered` is used instead of `channelInactive` but they don't make any difference, so I moved the cleanup logic to `channelInactive`.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -373,17 +398,15 @@ private void handleRequest(ChannelHandlerContext ctx, DecodedHttpRequest req) th
                 }
             })).exceptionally(CompletionActions::log);
 
-            res.completionFuture().handle(voidFunction((ret, cause) -> {
+            res.completionFuture().handleAsync(voidFunction((ret, cause) -> {
```
trustin: Not really related with this PR, but thought this is cleaner.

jetty/src/main/java/com/linecorp/armeria/server/jetty/JettyService.java
```diff
@@ -282,11 +282,12 @@ private void invoke(ServiceRequestContext ctx, HttpResponseWriter res,
             }
 
             final HttpHeaders headers = toResponseHeaders(transport);
-            res.write(headers);
-            for (;;) {
-                final HttpData data = out.poll();
-                if (data == null || !res.tryWrite(data)) {
-                    break;
+            if (res.tryWrite(headers)) {
+                for (;;) {
```
trustin: /cc @kojilin This should fix the `IllegalStateException` issue you reported.

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -406,11 +406,12 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
                     try {
                         coyoteAdapter.service(coyoteReq, coyoteRes);
                         final HttpHeaders headers = convertResponse(coyoteRes);
-                        res.write(headers);
-                        for (;;) {
-                            final HttpData d = data.poll();
-                            if (d == null || !res.tryWrite(d)) {
-                                break;
+                        if (res.tryWrite(headers)) {
+                            for (;;) {
```
trustin: /cc @kojilin This should fix the `IllegalStateException` issue you reported.

tomcat/src/test/java/com/linecorp/armeria/server/tomcat/TomcatServiceTest.java
```diff
@@ -110,6 +112,7 @@ public void jarBasedWebAppWithAlternativeRoot() throws Exception {
                         .startsWith("application/java");
                 assertThat(res.getFirstHeader(HttpHeaderNames.CONTENT_LENGTH.toString()).getValue())
                         .isEqualTo("1361");
+                EntityUtils.consume(res.getEntity());
```
trustin: Without this, Apache HC will close the connection before reading the response fully.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -197,16 +197,41 @@ public int unfinishedRequests() {
...
+        // 3) The response stream is aborted once the server detects the disconnection.
+        // 4) The publisher calls StreamWriter.close() but it's aborted already.
+        //
+        // To reduce the chance of such situation, we wait a little bit before aborting unfinished responses.
+
+        switch (protocol) {
+            case H1C:
+            case H1:
+                // XXX(trustin): How much time is 'a little bit'?
+                ctx.channel().eventLoop().schedule(this::cleanup, 1, TimeUnit.SECONDS);
```
trustin: I'm not sure if we should make this delay configurable.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
anuraaga: The netty documentation on these callbacks wasn't very clear, so I implemented every method and added breakpoints to try to figure it out. This was called where I expected and unregistering from armeria when unregistering from netty made intuitive sense, but if they're similar semantics seems good.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
anuraaga: Configuring just this knob is probably overkill - setting it based on idle timeout handler might make sense, but might also be overkill.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -197,16 +197,41 @@ public int unfinishedRequests() {
...
 
-    @Override
-    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+    private void cleanup() {
         if (responseEncoder != null) {
             responseEncoder.close();
         }
+
+        unfinishedResponses.keySet().forEach(res -> {
+            if (!res.isComplete()) {
```
anuraaga: Might be mistaken but I thought our semantics allowed aborting already completed streams (since completion is asynchronous I thought it's considered best effort vs other completion events)

trustin: Yeah, it's probably redundant. Let me run the tests again without this guard.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
trustin: unregistration will always follow `channelInactive`. Netty allows deregistration without closing for a very special case, but we don't use it.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
trustin: Everything's fine without the guard. Removed.

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -304,18 +329,14 @@ private void write0(HttpObject o, boolean endOfStream, boolean flush) {
...
-            subscription.cancel();
-        }
+        ctx.flush();
     }
 
-    private void setDone() {
+    private State setDone() {
         cancelTimeout();
-        state = State.DONE;
+        final State state = this.state;
```
minwoox: `oldState` to make it clear?

trustin: Fixed

trustin:
Thanks for reviewing. Will merge once CI passes.

trustin:
Thank you!

PR #1121, opened 2018-04-04
Throw ClosedPublisherException when StreamWriter.write() fails

trustin:
Motivation:

StreamWriter.write() throws IllegalStateException. However, we have a
dedicated exception for that - ClosedPublisherException.

Modifications:

- Throw ClosedPublisherException instead of IllegalStateException

Result:

- Easier to catch

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
@@ -62,13 +62,13 @@ public interface StreamWriter<T> {
      * Writes the specified object to the {@link StreamMessage}. The written object will be transferred to the
      * {@link Subscriber}.
      *
-     * @throws IllegalStateException if the stream was already closed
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
```diff
@@ -62,13 +62,13 @@
      * Writes the specified object to the {@link StreamMessage}. The written object will be transferred to the
      * {@link Subscriber}.
      *
-     * @throws IllegalStateException if the stream was already closed
+     * @throws ClosedPublisherException if the stream was already closed
```
minwoox: Could you change the line 79 as well?

trustin: Oops! Fixed :-)

trustin:
Thanks, reviewers!

PR #1123, opened 2018-04-05
Remove unthrown Exceptions from anonymous ServerListenerAdapter classes

dongjinleekr:


trustin:
Thanks, @dongjinleekr !

PR #1125, opened 2018-04-05
Fix a bug where contentType is not reset

minwoox:
Motivation:
If a user set a `contentType` once, and then set it again using `setObject(...)`,
the `contentType()` returns the previous one.

Modification:
- Check the contentType again when `contentType()` is called`

Result:
- No surprise

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
+++ b/core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
@@ -22,6 +22,8 @@ import java.util.List;
 
 import javax.annotation.Nullable;
 
+import com.google.common.base.Ascii;
+
 import com.linecorp.armeria.internal.ArmeriaHttpUtil;
 
 import io.netty.handler.codec.DefaultHeaders;
@@ -186,12 +188,13 @@ public final class DefaultHttpHeaders
     @Nullable
     @Override
     public MediaType contentType() {
-        final MediaType contentType = this.contentType;
         final String contentTypeString = get(HttpHeaderNames.CONTENT_TYPE);
         if (contentTypeString == null) {
             return null;
         }
-        if (contentType != null && contentType.toString().equals(contentTypeString)) {
+
+        final MediaType contentType = this.contentType;
+        if (contentType != null && Ascii.equalsIgnoreCase(contentType.toString(), contentTypeString.trim())) {
             return contentType;
         }
 
--- a/core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
+++ b/core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
@@ -187,13 +187,12 @@ public final class DefaultHttpHeaders
     @Override
     public MediaType contentType() {
         final MediaType contentType = this.contentType;
-        if (contentType != null) {
-            return contentType;
-        }
-
         final String contentTypeString = get(HttpHeaderNames.CONTENT_TYPE);
         if (contentTypeString == null) {
             return null;
+        }
+        if (contentType != null && contentType.toString().equals(contentTypeString)) {
+            return contentType;
         }
 
         try {
--- a/core/src/test/java/com/linecorp/armeria/common/HttpHeadersTest.java
+++ b/core/src/test/java/com/linecorp/armeria/common/HttpHeadersTest.java
@@ -37,7 +37,7 @@ public class HttpHeadersTest {
         assertThat(headers.get(of("HEADER3"))).isEqualTo("VALUE3");
 
         assertThat(headers.names())
-                  .containsExactlyInAnyOrder(of("header1"), of("header2"), of("header3"));
+                .containsExactlyInAnyOrder(of("header1"), of("header2"), of("header3"));
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -186,32 +186,41 @@ public HttpHeaders status(HttpStatus status) {
...
     public HttpHeaders contentType(MediaType contentType) {
         requireNonNull(contentType, "contentType");
-        this.contentType = contentType;
         return set(HttpHeaderNames.CONTENT_TYPE, contentType.toString());
     }
 
+    @Override
+    public HttpHeaders set(AsciiString name, String value) {
+        super.set(name, value);
+        setIfContentType(name, value);
```
trustin: Every `set()` call will involve string comparison even if the header name is not `content-type`, which is an overkill. We could just do this:

```java
public MediaType contentType() {
    final MediaType contentType = this.contentType;
    final String contentTypeString = get(HttpHeaderNames.CONTENT_TYPE);
    if (contentTypeString == null) {
        return null;
    }
    if (contentType != null && contentType.toString().equals(contentTypeString)) {
        return contentType;
    }
    ...
}
```

core/src/test/java/com/linecorp/armeria/common/HttpHeadersTest.java
```diff
@@ -48,4 +48,26 @@ public void testInvalidHeaderName() throws Exception {
         assertThatThrownBy(() -> HttpHeaders.of(AsciiString.of(""), "value1"))
                 .isInstanceOf(IllegalArgumentException.class);
     }
+
+    @Test
+    public void contentType() {
+        final HttpHeaders headers = HttpHeaders.of();
+        headers.contentType(MediaType.ANY_TYPE);
+        assertThat(headers.contentType()).isEqualTo(MediaType.ANY_TYPE);
```
trustin: `isSameAs`?

core/src/test/java/com/linecorp/armeria/common/HttpHeadersTest.java
```diff
@@ -48,4 +48,26 @@ public void testInvalidHeaderName() throws Exception {
...
+
+        headers.contentType(MediaType.ANY_APPLICATION_TYPE);
+        assertThat(headers.contentType()).isEqualTo(MediaType.ANY_APPLICATION_TYPE);
```
trustin: `isSameAs`?

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
minwoox: Fixed Thank you~!

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -187,14 +187,13 @@ public HttpHeaders status(HttpStatus status) {
     @Override
     public MediaType contentType() {
         final MediaType contentType = this.contentType;
```
hyangtack: We may move this line to L194?

minwoox: Oops~ Thank you~!

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -186,16 +186,16 @@ public HttpHeaders status(HttpStatus status) {
...
-            return contentType;
-        }
-
         final String contentTypeString = get(HttpHeaderNames.CONTENT_TYPE);
         if (contentTypeString == null) {
             return null;
         }
 
+        final MediaType contentType = this.contentType;
+        if (contentType != null && contentType.toString().equals(contentTypeString)) {
```
anuraaga: I believe this will always be `false` if the content type header was formatted weird, with extra spaces for example since `toString` is always normalized. It's probably rare enough to not worry about, but what do you think? I guess we could fix this without the efficiency loss of the map lookup if we overrode the logic of the setter instead of the getter

trustin: If we override the setter, we would need to check if the header name is equal to 'content-type' for all headers. I guess this would be more inefficient?

minwoox: There are three ways to set a content type which are `contentType(mediaType)`, `set*()` and `add*()`. They are all ended up calling `set(AsciiString name, String value)`. If we override the `set` method, we should check the name equals to `HttpHeaderNames.CONTENT_TYPE` every time add or set header.

minwoox: There are three ways to set a content type which are `contentType(mediaType)`, `set*()` and `add*()`. They are all ended up calling `set(AsciiString name, String value)`. If we override the `set` method, we should check the name equals to `HttpHeaderNames.CONTENT_TYPE` every time add or set header.

anuraaga:
Ah yeah if it's not just setObject then that seems worse. Anyways since RPC
will always have well defined media type think it's fine as is.

minwoox:
@anuraaga, @trustin, @hyangtack  please take a look and let me know silently if there's something wrong.

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -186,19 +189,23 @@ public HttpHeaders status(HttpStatus status) {
...
+            if (contentTypeString.equals(this.contentTypeString) ||
+                contentTypeString.equals(contentType.toString())) {
+                return this.contentType;
+            }
+        }
+
         try {
-            this.contentType = MediaType.parse(contentTypeString);
-            return this.contentType;
+            contentType = MediaType.parse(contentTypeString);
```
hyangtack: nit: How about setting `contentType` and `this. contentTypeString` to `null` before parsing?

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
```diff
@@ -186,16 +188,16 @@ public HttpHeaders status(HttpStatus status) {
...
-            return contentType;
-        }
-
         final String contentTypeString = get(HttpHeaderNames.CONTENT_TYPE);
         if (contentTypeString == null) {
             return null;
         }
 
+        final MediaType contentType = this.contentType;
+        if (contentType != null && Ascii.equalsIgnoreCase(contentType.toString(), contentTypeString.trim())) {
```
anuraaga: This is slightly safer but still misses e.g. `application/json;    charset=utf-8` - notice spaces within the content type. The safest way would be to store `parsedContentTypeString` as a field to compare with the raw input instead of normalized, but maybe overkill - I feel as if it'd be find to leave out the `trim` though since it's not a complete normalization anyways.

trustin:
Thanks, @minwoox !

core/src/main/java/com/linecorp/armeria/common/DefaultHttpHeaders.java
minwoox: Yeah it might be better just remove the `trim` because it's not a normalization. :-)
 Thank you~!

PR #1126, opened 2018-04-06
Rename spring-boot-* to spring-boot1-* and spring-boot2-* to spring-b…

trustin:
…oot-*

Motivation:

We traditionally appended the version number to our artifact names only
when it's not the latest version we support.

Modifications:

- Rename spring-boot-* to spring-boot1-*
- Rename spring-boot2-* to spring-boot-*

Result:

Consistency


trustin:
Thanks!

PR #1130, opened 2018-04-08
Make periodical wait of monitoring graceful shutdown configurable

imasahiro:
Motivations
- Current graceful shutdown takes 100ms at minimum if not completed
  requests are exists when start to shutdown.

Changes
- Make periodical wait configurable to make a chance to stop server
  with shorter period


trustin:
Is there any noticeable effect on total time taken for shutdown? I guess the difference would be less than 100ms, no?

imasahiro:
> Is there any noticeable effect on total time taken for shutdown because of this change? I guess the difference would be less than 100ms, no?

No, and correct, less than 100ms. Let me add more metrics and will come back here if needed.

PR #1132, opened 2018-04-09
Add more flexibility in adding token-based authentication to HttpAuth…

anuraaga:
…ServiceBuilder.

Currently if a user wants to use a bearer token on a header other than `Authorization` (perhaps they need to be able to assert two identities, a user and a server), it would require a lot of code duplication. This change allows users to specify a header name when adding token authorization. It also exposes a generic `addTokenAuthorizer` method which can be used for different token types too, there doesn't seem to be a great reason to keep the method private.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/auth/HttpAuthServiceBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/auth/HttpAuthServiceBuilder.java
@@ -83,7 +83,7 @@ public final class HttpAuthServiceBuilder {
      * Adds an OAuth1a {@link Authorizer} for the given {@code header}.
      */
     public HttpAuthServiceBuilder addOAuth1a(Authorizer<? super OAuth1aToken> authorizer, AsciiString header) {
-        return addTokenAuthorizer(new OAuth1aTokenExtractor(header),
+        return addTokenAuthorizer(new OAuth1aTokenExtractor(requireNonNull(header, "header")),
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/server/auth/HttpAuthServiceBuilder.java
```diff
@@ -57,26 +59,64 @@ public HttpAuthServiceBuilder add(Iterable<? extends Authorizer<HttpRequest>> au
...
-        return this;
+        return addTokenAuthorizer(AuthTokenExtractors.OAUTH1A,
+                                  requireNonNull(authorizer, "authorizer"));
+    }
+
+    /**
+     * Adds an OAuth1a {@link Authorizer} for the given {@code header}.
+     */
+    public HttpAuthServiceBuilder addOAuth1a(Authorizer<? super OAuth1aToken> authorizer, AsciiString header) {
+        return addTokenAuthorizer(new OAuth1aTokenExtractor(header),
```
hyangtack: `requireNonNull(header)`? or adding `requireNonNull` to the constructor of each extractor?

anuraaga: Oops somehow missed this one - thanks. I think we usually add `requireNonNull` to public APIs and not so much to internal, especially package-local, ones so will leave them here.

trustin:
Thanks, @anuraaga !

PR #1135, opened 2018-04-09
Use ServerBuilder.service() instead of serviceUnder() in gRPC server …

trustin:
…tutorial

Motivation:

We do not use ServerBuilder.serviceUnder() anymore for binding gRPC
services. We use service(ServiceWithPathMappings).

Modifications:

Fix the gRPC server tutorial to use service()

Result:

Less user confusion


trustin:
Need to update a little bit more.

trustin:
Updated. Replaced the paragraph after the example:

```rst
.. note::

    We bound the GrpcService_ without specifying any path mappings. It is because GrpcService_ implements
    ServiceWithPathMappings_, which dynamically provides path mappings by itself.
```

trustin:
Thanks, @hyangtack !

PR #1138, opened 2018-04-10
Fix a bug where a GrpcDocServicePlugin does not handle exact path map…

### LINKED ISSUE #1136: GrpcDocServicePlugin does not handle the services with exact mapping
When a user binds a `GrpcService` using `ServerBuilder.service(ServiceWithPathMappings)`, its debug form does not show up in `DocService`, because `GrpcDocServicePlugin` only accepts the `ServiceConfig`s with a prefix-mapping. See `GrpcDocServicePlugin.java:149`.

trustin:
…pings

Motivation:

When a user adds a GrpcService with ServerBuilder.service(ServiceWithPathMappings),
GrpcDocServicePlugin fails to add the service endpoints to ServiceSpecification.

To work around this issue, a user has to fall back to the old usage:

    ServerBuilder sb = ...;
    sb.serviceUnder("/", grpcService); // Works
    sb.service(grpcService); // Does not work

Modifications:

- Update the endpoint detection logic in GrpcDocServicePlugin

Result:

- Fixes #1136

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcDocServicePlugin.java
+++ b/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcDocServicePlugin.java
@@ -146,23 +146,28 @@ public class GrpcDocServicePlugin implements DocServicePlugin {
                             return new ServiceEntryBuilder(serviceDescriptor);
                         });
             }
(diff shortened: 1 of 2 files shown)


trustin:
Thanks, reviewers!

PR #1142, opened 2018-04-10
Add more concise variants of HttpResponse.of()

kishida:
Motivations
- There are many boiler plate code in the document.
- Make it more simple and friendly.
- Light code will be more light.
- It will make more user. I hope.

Changes
- Introduce `of(String)` and `of(String, Object...)` for `HttpResponse` with `HttpStatus.OK` and `MediaType.PLAIN_TEXT_UTF_8` as a default.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/common/HttpResponseTest.java
+++ b/core/src/test/java/com/linecorp/armeria/common/HttpResponseTest.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.common;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class HttpResponseTest {
+
+    @Test
+    public void ofWithPlainContent() {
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/linecorp/armeria/common/HttpResponseTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class HttpResponseTest {
+
+    @Test
+    public void ofWithPlainContent() {
+        // Using non-ascii to test UTF-8 conversion
+        HttpResponse res = HttpResponse.of("Armeriaはいろんな使い方がアルメリア");
```
kishida: Once I had used escaped unicode but CheckStyle compalins.
Can I use a multibyte code in the source?

kishida:
I have added omit `HttpStatus` version!

core/src/test/java/com/linecorp/armeria/common/HttpResponseTest.java
trustin: Sure, you can, because we always assume UTF-8 for source code. If you really wish to use escaped unicode, You can use `@SuppressWarnings` annotation, e.g.

```java
@SuppressWarnings("checkstyle:AvoidEscapedUnicodeCharacters")
@Test
public void ...
```

kishida: I want to leave it since `Armeriaはいろんな使い方がアルメリア` seems funny. 😄

kishida:
I've changed the commit comment along with the PR title.

trustin:
@kishida wrote:
> I've changed the commit comment along with the PR title.

Thanks a lot!

core/src/test/java/com/linecorp/armeria/common/HttpResponseTest.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2017 LINE Corporation
```
hyangtack: 2018 :-)

kishida: ah...
Fixed! I'm now in 2018!

trustin:
Thanks, @kishida !

PR #1144, opened 2018-04-11
Support SSLeay PKCS#5 private key format

trustin:
Motivation:

Server configuration fails when a TLS private key is in PKCS#5 format.

    Caused by: java.security.InvalidKeyException: IOException : algid parse error, not a sequence
        at sun.security.pkcs.PKCS8Key.decode(PKCS8Key.java:352)
        at sun.security.pkcs.PKCS8Key.decode(PKCS8Key.java:357)
        at sun.security.ec.ECPrivateKeyImpl.<init>(ECPrivateKeyImpl.java:73)
        at sun.security.ec.ECKeyFactory.implGeneratePrivate(ECKeyFactory.java:237)
        at sun.security.ec.ECKeyFactory.engineGeneratePrivate(ECKeyFactory.java:165)
        ... 29 more

Modifications:

- Use Bouncy Castle to load PKCS#5 private keys
  - Add BouncyCastleKeyFactoryProvider, which is a downsized version of
    BouncyCastleProvider that registers absolutely necessary
    KeyFactorySpis only, so that a user doesn't have to worry about
    pulling unwanted ciphers unexpectedly

Result:

- Less surprises


trustin:
Thanks, @hyangtack !

PR #1147, opened 2018-04-12
Get `@Param` or `@Header` annotation value from a variable if the val…

### LINKED ISSUE #570: Parameter mapping without @PathParam annotation
https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html

It's possible to get the parameter names via the reflection API for the classes compiled with the `-parameters` javac option.

/cc @hyangtack

hyangtack:
…ue is not specified

Modifications:
- Add `findParamName` and `findHeaderName` to `AnnotatedHttpServiceParamUtil` in order to
  obtain the variable name if a `@Param` or `@Header` annotation does not have a value.
- Refactor `BeanRequestConverterFunction` for cleanup.
- Documentation.

Result:
- Closes #570

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
@@ -15,9 +15,14 @@
  */
 package com.linecorp.armeria.internal;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.linecorp.armeria.common.HttpParameters.EMPTY_PARAMETERS;
+import static java.util.Objects.requireNonNull;
 
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Field;
+import java.lang.reflect.Parameter;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
@@ -30,13 +35,17 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Ascii;
+import com.google.common.base.CaseFormat;
 
 import com.linecorp.armeria.common.AggregatedHttpMessage;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpParameters;
 import com.linecorp.armeria.common.MediaType;
 import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.annotation.Header;
+import com.linecorp.armeria.server.annotation.Param;
 
 import io.netty.handler.codec.http.HttpConstants;
 import io.netty.handler.codec.http.QueryStringDecoder;
@@ -119,8 +128,7 @@ public final class AnnotatedHttpServiceParamUtil {
     public static Object convertParameter(@Nullable String value,
                                           Class<?> type,
                                           EnumConverter<?> enumConverter,
-                                          boolean optionalWrapped
-    ) {
+                                          boolean optionalWrapped) {
         Object converted = null;
         if (value != null) {
             converted = type.isEnum() ? enumConverter.getEnumElement(value)
@@ -206,6 +214,131 @@ public final class AnnotatedHttpServiceParamUtil {
 
         throw new IllegalArgumentException(
                 "Type '" + clazz.getSimpleName() + "' can't be converted.");
+    }
+
+    /**
+     * Returns the value of the {@link Param} annotation which is specified on the {@code element} if
+     * the value is not blank. If the value is blank, it returns the name of the specified
+     * {@code element} object which is an instance of {@link Parameter} or {@link Field}.
+     */
+    @Nullable
+    public static String findParamName(AnnotatedElement element) {
+        return findParamName(element, element);
+    }
+
+    /**
+     * Returns the value of the {@link Param} annotation which is specified on the {@code element} if
+     * the value is not blank. If the value is blank, it returns the name of the specified
+     * {@code nameRetrievalTarget} object which is an instance of {@link Parameter} or {@link Field}.
+     */
+    @Nullable
+    public static String findParamName(AnnotatedElement element, Object nameRetrievalTarget) {
+        requireNonNull(element, "element");
+        requireNonNull(nameRetrievalTarget, "nameRetrievalTarget");
+        final Param param = element.getAnnotation(Param.class);
+        if (param == null) {
+            return null;
+        }
+
+        final String value = param.value();
+        if (DefaultValues.isSpecified(value)) {
+            checkArgument(!value.isEmpty(), "value is empty");
+            return value;
+        }
+        return getName(nameRetrievalTarget);
+    }
+
+    /**
+     * Returns the value of the {@link Header} annotation which is specified on the {@code element} if
+     * the value is not blank. If the value is blank, it returns the name of the specified
+     * {@code element} object which is an instance of {@link Parameter} or {@link Field}.
+     *
+     * <p>Note that the name of the specified {@code element} will be converted as
+     * {@link CaseFormat#LOWER_HYPHEN} that the string elements are separated with one hyphen({@code -})
+     * character. The value of the {@link Header} annotation will not be converted because it is clearly
+     * specified by a user.
+     */
+    @Nullable
+    public static String findHeaderName(AnnotatedElement element) {
+        return findHeaderName(element, element);
+    }
+
+    /**
+     * Returns the value of the {@link Header} annotation which is specified on the {@code element} if
+     * the value is not blank. If the value is blank, it returns the name of the specified
+     * {@code nameRetrievalTarget} object which is an instance of {@link Parameter} or {@link Field}.
+     *
+     * <p>Note that the name of the specified {@code nameRetrievalTarget} will be converted as
+     * {@link CaseFormat#LOWER_HYPHEN} that the string elements are separated with one hyphen({@code -})
+     * character. The value of the {@link Header} annotation will not be converted because it is clearly
+     * specified by a user.
+     */
+    @Nullable
+    public static String findHeaderName(AnnotatedElement element, Object nameRetrievalTarget) {
+        requireNonNull(element, "element");
+        requireNonNull(nameRetrievalTarget, "nameRetrievalTarget");
+        final Header header = element.getAnnotation(Header.class);
+        if (header == null) {
+            return null;
+        }
+
+        final String value = header.value();
+        if (DefaultValues.isSpecified(value)) {
+            checkArgument(!value.isEmpty(), "value is empty");
+            return value;
+        }
+        return toHeaderName(getName(nameRetrievalTarget));
+    }
+
+    /**
+     * Returns whether the specified {@code element} is annotated with a {@link Param} or {@link Header}.
+     */
+    public static boolean hasParamOrHeader(AnnotatedElement element) {
+        return element.isAnnotationPresent(Param.class) ||
+               element.isAnnotationPresent(Header.class);
+    }
+
+    private static String getName(Object element) {
+        if (element instanceof Parameter) {
+            final Parameter parameter = (Parameter) element;
+            if (!parameter.isNamePresent()) {
+                throw new IllegalArgumentException(
+                        "cannot obtain the name of the parameter or field automatically. " +
+                        "Please make sure you compiled your code with '-parameters' option. " +
+                        "If not, you need to specify parameter and header names with @" +
+                        Param.class.getName() + " and @" + Header.class.getName() + '.');
+            }
+            return parameter.getName();
+        }
+        if (element instanceof Field) {
+            return ((Field) element).getName();
+        }
+        throw new IllegalArgumentException("cannot find the name: " + element.getClass().getName());
+    }
+
+    @VisibleForTesting
+    static String toHeaderName(String name) {
+        requireNonNull(name, "name");
+        checkArgument(!name.isEmpty(), "name is empty");
+
+        final String upperCased = Ascii.toUpperCase(name);
+        if (name.equals(upperCased)) {
+            return CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.LOWER_HYPHEN, name);
+        }
+        final String lowerCased = Ascii.toLowerCase(name);
+        if (name.equals(lowerCased)) {
+            return CaseFormat.LOWER_UNDERSCORE.to(CaseFormat.LOWER_HYPHEN, name);
+        }
+        // Ensure that the name does not contain '_'.
+        // If it contains '_', we give up to make it lower hyphen case. Just converting it to lower case.
+        if (name.indexOf('_') >= 0) {
+            return lowerCased;
+        }
+        if (Ascii.isUpperCase(name.charAt(0))) {
+            return CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_HYPHEN, name);
+        } else {
+            return CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_HYPHEN, name);
+        }
     }
 
     public static class EnumConverter<T extends Enum<T>> {
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
@@ -16,12 +16,15 @@
 package com.linecorp.armeria.server.annotation;
 
 import static com.linecorp.armeria.internal.AnnotatedHttpServiceParamUtil.convertParameter;
+import static com.linecorp.armeria.internal.AnnotatedHttpServiceParamUtil.findHeaderName;
+import static com.linecorp.armeria.internal.AnnotatedHttpServiceParamUtil.findParamName;
+import static com.linecorp.armeria.internal.AnnotatedHttpServiceParamUtil.hasParamOrHeader;
 import static com.linecorp.armeria.internal.AnnotatedHttpServiceParamUtil.httpParametersOf;
 import static com.linecorp.armeria.internal.AnnotatedHttpServiceParamUtil.validateAndNormalizeSupportedType;
+import static java.util.Collections.singletonList;
 import static org.reflections.ReflectionUtils.getAllFields;
 import static org.reflections.ReflectionUtils.getAllMethods;
 
-import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Executable;
@@ -32,17 +35,21 @@ import java.lang.reflect.Parameter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
+import java.util.StringJoiner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.stream.Collectors;
 
+import javax.annotation.Nullable;
+
 import org.reflections.ReflectionUtils;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.MapMaker;
 
 import com.linecorp.armeria.common.AggregatedHttpMessage;
 import com.linecorp.armeria.common.HttpParameters;
+import com.linecorp.armeria.internal.AnnotatedHttpServiceParamUtil;
 import com.linecorp.armeria.internal.AnnotatedHttpServiceParamUtil.EnumConverter;
 import com.linecorp.armeria.server.ServiceRequestContext;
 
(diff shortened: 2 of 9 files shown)


LiYing2010:
`@Default` annotation can be used only for method parameter now,
can we just remove this annotation, and add a new `defaultValue` attribute to `@Param` and `@Header` annotation?

e.g:
change `@Header("username") @Default("hello")`
to `@Header(value="username", defaultValue="hello")`

and change `@Param("age") @Default("20")`
to `@Param(value="age", defaultValue="20")`

so, we can specify default value for any methods/constructors/parameters/fields annotated by `@Param` and/or `@Header`

LiYing2010:
> @Header(value="username", defaultValue="hello")
> @Param(value="age", defaultValue="20")

the annotation attribute name `value` here is a little ambiguous, can we change it to `name`?
e.g:
@Header(name="username", defaultValue="hello")
@Param(name="age", defaultValue="20")

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -208,6 +217,133 @@ public static Object convertParameter(@Nullable String value,
...
+     * Returns {@code true} if the specified {@code element} is not annotated with a {@link Param} nor
+     * {@link Header}.
+     */
+    public static boolean hasNeitherParamNorHeader(AnnotatedElement element) {
+        return !element.isAnnotationPresent(Param.class) &&
+               !element.isAnnotationPresent(Header.class);
+    }
+
+    private static String getName(Object element) {
+        if (element instanceof Parameter) {
```
LiYing2010: it will be better if we support `element instanceof Method` case here,
and looks like you are using this pattern on the method `AbstractRequestBean#setGender` in the unit test code

trustin:
> the annotation attribute name value here is a little ambiguous, can we change it to name?

Is it still possible to use `@Param("name")` with that change?

LiYing2010:
> Is it still possible to use @Param("name") with that change?

You are right, this will be a little problem.
I will try if I can find an elegant solution.

But do you agree the basic idea of merging `@Default` into  `@Param` and `@Header`?
If that is OK, I will create a PR for it.

trustin:
@LiYing2010 wrote:
> `@Default` annotation can be used only for method parameter now,
can we just remove this annotation, and add a new `defaultValue` attribute to `@Param` and `@Header` annotation?
> ...
> do you agree the basic idea of merging `@Default` into  `@Param` and `@Header`?
If that is OK, I will create a PR for it.

Perhaps we could just allow specifying `@Default` on a method and a field? e.g.

```java
public class MyBean {
   @Param("id") @Default("something")
   private String id;

   @Param("foo") @Default("bar")
   public void setFoo(String foo) { ... }
}
```

.. or did I miss something?

hyangtack:
@LiYing2010 Basically, I also agree your idea about specifying a default value of a field or method. But I think that @trustin 's suggestion slightly looks better because specifying `name = "value"` is a little inconvenient to me.

LiYing2010:
> Basically, I also agree your idea about specifying a default value of a field or method. But I think that @trustin 's suggestion slightly looks better because specifying name = "value" is a little inconvenient to me.


I think `@Default` means a default value of parameter, but parameter can be retrieved only from `@Param` or `@Header`.
So, when we use `@Default`,  we must use it as a pair with `@Param` or `@Header`.

It will make more sense if we change `default value` as a attribute of `@Param` or `@Header`.

If we keep `@Default` separated from `@Param` and `@Header`
- it will be a little harder to understand, so user may use it improperly
- we have to take care of `@Default` everywhere `@Param` or `@Header` appears, our code will be more complex



> specifying name = "value" is a little inconvenient

That's true.
how about we define 3 attributes for `@Param` and `@Header`: `value`/`name`/`defaultValue`
and implement something like [Spring AliasFor](https://docs.spring.io/spring/docs/4.3.17.BUILD-SNAPSHOT/javadoc-api/org/springframework/core/annotation/AliasFor.html)
then we can support both: ` @Param("foo")` and ` @Param(name="foo", defaultValue="bar")`

trustin:
I'm not sure I like the idea of introducing an additional annotation property, mainly because it makes annotations look inconsistent:

```java
public void someMethod(@Param("foo") String foo,
                       @Param(vaule="bar", defaultValue="def") String bar,
                       @Param("baz") String baz) { ... }

// vs.
public void someMethod(@Param("foo") String foo,
                       @Param("bar") @Default("def") String bar,
                       @Param("baz") String baz) { ... }
```

`@Default` is shorter and it gives us better aligned code.

trustin:
No, it's rather about sometimes having to specify attribute name and sometimes not:

```
@Param("without_name")
       ^
@Param(value="with_name")
             ^ -> Inconsistent column alignment
@Param("without_name_2")
       ^
```

I'm not sure if it's a good idea to rename `value` to `name` since it will only increase the number of keystrokes a user has to make even when a user doesn't want to specify a default value:

```
@Param(name="why should a specify attribute name when there's only a single attribute?")
@Param(name="...", defaultValue="...")
```

trustin:
> If we keep @Default separated from @Param and @Header
>
> - it will be a little harder to understand, so user may use it improperly
> - we have to take care of @Default everywhere @Param or @Header appears, our code will be more complex

Could you show me an example code that shows its increased complexity and improper use?

LiYing2010:
> I'm not sure if it's a good idea to rename value to name since it will only increase the number of keystrokes a user has to make even when a user doesn't want to specify a default value:

I have explained this here:
>how about we define 3 attributes for `@Param` and `@Header`: `value`/`name`/`defaultValue`
and implement something like Spring `@AliasFor`
then we can support both: `@Param("foo")` and `@Param(name="foo", defaultValue="bar")`

trustin:
Ah, I see. That sounds not that bad, then. However, I'm still curious why the current API leads to increased complexity and improper use. Any examples?

LiYing2010:
>  Could you show me an example code that shows its increased complexity and improper use?

I think `defaultValue` and `name` are a pair of attributes which describe detail information of Query Parameters or HTTP Headers.

In concept, `defaultValue` is a part of  `@Param` and `@Header`, but it is implemented as a separated annotation,
so, everytime we process anything about `@Param` or `@Header`, we have to check if there is `@Default`, and take care of it, instead of using the `defaultValue` attribute attached to `@Param` or `@Header`.

By "user may use it improperly", I mean, user could declare a `@Default` without `@Param` or `@Header`, which make no sense.


And BTW, [@RequestParam](https://docs.spring.io/spring/docs/4.2.0.RC2_to_4.2.0.RC3/Spring%20Framework%204.2.0.RC3/index.html?org/springframework/core/annotation/AliasFor.html) and [@RequestHeader](https://docs.spring.io/spring/docs/4.2.0.RC2_to_4.2.0.RC3/Spring%20Framework%204.2.0.RC3/index.html?org/springframework/core/annotation/AliasFor.html) in Spring MVC also uses same strategy I proposed:
1. attribute `name()` and `defaultValue()`
2. attribute `value()` as a alias for `name()`

LiYing2010:
> implement something like Spring [`@AliasFor`](https://docs.spring.io/spring/docs/4.2.0.RC2_to_4.2.0.RC3/Spring%20Framework%204.2.0.RC3/org/springframework/core/annotation/AliasFor.html)

About this part, I have read source code and document of Spring, what they do is:
>Spring will synthesize an annotation by wrapping it in a dynamic proxy that transparently enforces attribute alias semantics for annotation attributes that are annotated with `@AliasFor`

I think we can implement the same thing.
Of course, not all the features of Spring [`@AliasFor`](https://docs.spring.io/spring/docs/4.2.0.RC2_to_4.2.0.RC3/Spring%20Framework%204.2.0.RC3/org/springframework/core/annotation/AliasFor.html), but only what we need: `Aliases within an Annotation`

LiYing2010:
> I understand that some people may prefer the proposed API, but I still fail to see that its benefit outweighs the removal of `@default` .


OK, understood.

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -208,6 +217,135 @@ public static Object convertParameter(@Nullable String value,
...
+        requireNonNull(element, "element");
+        requireNonNull(nameRetrievalTarget, "nameRetrievalTarget");
+        final Param param = element.getAnnotation(Param.class);
+        if (param == null) {
+            return null;
+        }
+
+        final String value = param.value();
+        if (DefaultValues.isSpecified(value)) {
+            checkArgument(!value.trim().isEmpty(), "value: %s (expected: not blank)");
```
minwoox: `"value: %s` there's no argument matching %s.

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -208,6 +217,135 @@ public static Object convertParameter(@Nullable String value,
...
+        requireNonNull(element, "element");
+        requireNonNull(nameRetrievalTarget, "nameRetrievalTarget");
+        final Header header = element.getAnnotation(Header.class);
+        if (header == null) {
+            return null;
+        }
+
+        final String value = header.value();
+        if (DefaultValues.isSpecified(value)) {
+            checkArgument(!value.trim().isEmpty(), "value: %s (expected: not blank)");
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -208,6 +217,135 @@ public static Object convertParameter(@Nullable String value,
...
+    public static boolean hasParamOrHeader(AnnotatedElement element) {
+        return element.isAnnotationPresent(Param.class) ||
+               element.isAnnotationPresent(Header.class);
+    }
+
+    /**
+     * Returns {@code true} if the specified {@code element} is not annotated with a {@link Param} nor
+     * {@link Header}.
+     */
+    public static boolean hasNeitherParamNorHeader(AnnotatedElement element) {
```
minwoox: It seems like this method is not used anywhere and the name is sort of confusing. How about removing this method?

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -208,6 +217,135 @@ public static Object convertParameter(@Nullable String value,
...
+        }
+        if (element instanceof Field) {
+            return ((Field) element).getName();
+        }
+        throw new IllegalArgumentException("cannot find the name: " + element.getClass().getName());
+    }
+
+    @VisibleForTesting
+    static String toHeaderName(String name) {
+        checkArgument(!Strings.isNullOrEmpty(name), "name");
```
minwoox: Did you mean `requireNonNull`? otherwise we need to change the second parameter which is more specific string.

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -223,30 +218,28 @@ private static ConstructorInfo buildFrom(final Class<?> beanType) {
...
+                // Only 1 annotated constructor.
                 return constructorList.get(0);
-            } else {
-                // there are more than 1 annotated constructors, we don't know to use which one
-                throw new IllegalArgumentException(
-                        String.format("There are more than 1 annotated constructors in class '%1$s'.",
-                                      beanType.getCanonicalName())
-                );
             }
+            // More than 1 annotated constructors. We don't know to use which one.
```
minwoox: We don't know which one to use.? more natural?

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
hyangtack: Oops! Thanks!

site/src/sphinx/server-annotated-service.rst
```diff
@@ -472,6 +525,13 @@ or HTTP headers:
         @Header("age") // This field will be injected by value of HTTP header "age".
         private int age;
 
+        // You can omit the value of @Param or @Header if you compiled your code with ``-parameters`` javac option.
+        @Param         // This field will be injected by value of parameter "gender".
```
minwoox: by the value?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -472,6 +525,13 @@ or HTTP headers:
...
+        private String gender;
+
+        @Header        // This field will be injected by value of HTTP header "accept-language".
```

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
hyangtack: Replaced  `allMatch(AnnotatedHttpServiceParamUtil::hasNeitherParamNorHeader)` with `.noneMatch(AnnotatedHttpServiceParamUtil::hasParamOrHeader)` and removed this. Thanks!

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -310,10 +312,10 @@ private static String getName(Object element) {
         if (element instanceof Parameter) {
             final Parameter parameter = (Parameter) element;
             checkArgument(parameter.isNamePresent(),
-                          "Cannot obtain the name of the parameter or field automatically. " +
+                          "cannot obtain the name of the parameter or field automatically. " +
                           "Please check whether your compile option includes '-parameters'. " +
-                          "If not, you need to specify parameter and header names with @" +
-                          Param.class.getName() + " and @" + Header.class.getName() + '.');
+                          "If not, you need to specify parameter and header names with @%s and @%s.",
+                          Param.class.getName(), Header.class.getName());
```
trustin: Could we use traditional `if` block instead of `checkArgument()` here, because otherwise `Class.getName()` will be invoked even for successful path.

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -333,9 +335,9 @@ static String toHeaderName(String name) {
             return CaseFormat.LOWER_UNDERSCORE.to(CaseFormat.LOWER_HYPHEN, name);
         }
         // Ensure that the name does not contain '_'.
-        // If it contains '_', we give up to make it lower hyphen case.
+        // If it contains '_', we give up to make it lower hyphen case. Just converting it to lower case.
         if (name.indexOf('_') >= 0) {
-            return name;
+            return name.toLowerCase();
```
trustin: `Ascii.toLowerCase()`?

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -422,7 +429,7 @@ private static ParamArrayValueRetriever fromAnnotatedMethodOrConstructor(Executa
                           getMemberFullName(element), elementParams.length);
             checkArgument(Arrays.stream(elementParams)
                                 .allMatch(AnnotatedHttpServiceParamUtil::hasNeitherParamNorHeader),
-                          "no @%s and @%s annotations are allowed to the parameter of %s",
+                          "no @%s and @%s annotations are allowed in the parameter of %s",
                           Param.class.getSimpleName(), Header.class.getSimpleName(),
```
trustin: Ditto (use traditional `if` block)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -190,6 +190,36 @@ In this case the variable name is used as the value.
         public HttpResponse hello1(@Param String name) { ... }
     }
 
+.. note::
+
+    You can configure ``-parameters`` javac option with your build framework as follows.
```

    You can configure your build tool to add ``-parameters`` javac option as follows.

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -208,6 +216,127 @@ public static Object convertParameter(@Nullable String value,
...
+        return element.isAnnotationPresent(Param.class) ||
+               element.isAnnotationPresent(Header.class);
+    }
+
+    private static String getName(Object element) {
+        if (element instanceof Parameter) {
+            final Parameter parameter = (Parameter) element;
+            checkArgument(parameter.isNamePresent(),
+                          "cannot obtain the name of the parameter or field automatically. " +
+                          "Please check whether your compile option includes '-parameters'. " +
```
trustin: check whether ... -> make sure you compiled your code with '-parameters' option. ?

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -458,77 +405,56 @@ protected String getParamStrValue(final RequestConvertContext context) {
...
-            if (retrieverList.isEmpty()) {
-                // there is no annotation on params of this method/constructor
+            checkArgument(elementParams.length == 1,
+                          "the number of parameters of %s: %s (expected: 1)",
+                          getMemberFullName(element), elementParams.length);
+            checkArgument(Arrays.stream(elementParams)
+                                .noneMatch(AnnotatedHttpServiceParamUtil::hasParamOrHeader),
+                          "no @%s and @%s annotations are allowed in the parameter of %s",
+                          Param.class.getSimpleName(), Header.class.getSimpleName(),
+                          getMemberFullName(element));
```
trustin: Better using traditional `if` blocks because otherwise `Class.getSimpleName()` and `getMemberFullName()` will be evaluated even on successful path.

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -458,77 +405,56 @@ protected String getParamStrValue(final RequestConvertContext context) {
...
-                throw new IllegalArgumentException(
-                        String.format(
-                                "There are %1$d parameter(s) for '%2$s', but only %3$d of them are annotated.",
-                                elementParams.length, getMemberFullName(element), retrieverList.size())
-                );
-            }
+            checkArgument(annotatedParameterCount == elementParams.length,
+                          "every parameter of %s should be annotated with one of @%s or @%s",
+                          getMemberFullName(element), Param.class.getSimpleName(),
+                          Header.class.getSimpleName());
```
trustin: Ditto (use traditional `if` block)

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -208,6 +216,127 @@ public static Object convertParameter(@Nullable String value,
...
+        if (name.equals(name.toUpperCase())) {
+            return CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.LOWER_HYPHEN, name);
+        }
+        if (name.equals(name.toLowerCase())) {
+            return CaseFormat.LOWER_UNDERSCORE.to(CaseFormat.LOWER_HYPHEN, name);
+        }
+        // Ensure that the name does not contain '_'.
+        // If it contains '_', we give up to make it lower hyphen case. Just converting it to lower case.
+        if (name.indexOf('_') >= 0) {
+            return name.toLowerCase();
```
trustin: - Please use `Ascii.toLowerCase()` for all `String.toLowerCase()` calls.
- The lowercased `name` could be extracted into a local variable because it appears twice.

core/src/main/java/com/linecorp/armeria/internal/AnnotatedHttpServiceParamUtil.java
```diff
@@ -208,6 +216,127 @@ public static Object convertParameter(@Nullable String value,
...
+        }
+        throw new IllegalArgumentException("cannot find the name: " + element.getClass().getName());
+    }
+
+    @VisibleForTesting
+    static String toHeaderName(String name) {
+        requireNonNull(name, "name");
+        checkArgument(!name.trim().isEmpty(), "name: %s (expected: not blank)", name);
+
+        if (name.equals(name.toUpperCase())) {
```
trustin: `Ascii.toUpperCase()`?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -180,6 +180,46 @@ one of the following supported types:
...
+.. code-block:: java
+
+    public class MyAnnotatedServiceClass {
+        @Get("/hello/{name}")
+        public HttpResponse hello1(@Param String name) { ... }
+    }
+
+.. note::
+
+    You can configure ``-parameters`` javac option with your build framework as follows.
```

    You can configure your build tool to add ``-parameters`` javac option as follows.

core/src/main/java/com/linecorp/armeria/server/annotation/BeanRequestConverterFunction.java
```diff
@@ -458,77 +405,58 @@ protected String getParamStrValue(final RequestConvertContext context) {
...
-                // parameter should have no annotation
-                checkNoParamAnnotations(element, elementParams);
-
-                final Class<?> type = elementParams[0].getType();
-                final ParamValueRetriever retriever = ParamValueRetriever.buildFrom(type, annotation);
-
-                return of(retriever);
+            checkArgument(elementParams.length == 1,
+                          "the number of parameters of %s: %s (expected: 1)",
+                          getMemberFullName(element), elementParams.length);
```
trustin: Could use traditional `if` block :-)

trustin:
Thanks a lot, @hyangtack and reviewers!

PR #1148, opened 2018-04-13
Revamp DNS-based service discovery

### LINKED ISSUE #1088: Generalize endpoint decoding from DNS records
We currently provide `DnsServiceEndpointGroup` that decodes the SRV records compatible with k8s. In some case, a user may want to:

- Use a different format for SRV records or
- Use other record type than SRV, such as TXT to decode the endpoint information

We could generalize `DnsServiceEndpointGroup` so we support such use cases.

trustin:
Motivation:

Current DNS-based service discovery implementations simply send a query
every second, not respecting TTL or the query result (failure).

Recent addition of Netty DnsNameResolver.resolveAll(DnsQuestion) in
4.1.23 enables us to improve this behavior.

Modifications:

- Add abstract DnsEndpointGroup which makes use of DnsNameResolver.resolveAll()
  - Rewrite DnsAddressEndpointGroup and DnsServiceEndpointGroup by
    extending DnsEndpointGroup
  - Add DnsTextEndpointGroup
- Add builders for Dns*EndpointGroup
  - Reduce the number of factory methods in Dns*EndpointGroup
- Rewrite the test cases
  - Add TestDnsServer to test the actual network traffic
- Miscellaneous:
  - Make Endpoint implement Comparable
  - Allow 0-weight in Endpoint, because it's allowed in a DNS SRV record
  - Add DynamicEndpoint.initialEndpointsFuture

Result:

- DNS servers are not hammered by Dns*EndpointGroup
- Customizable backoff policy
- Customizable DNS server addresses
- Customizable TTL range
- I had solid fun doing this.
- Fixes #1088


core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -429,6 +434,27 @@ public int hashCode() {
...
+            if (that.isGroup()) {
+                return groupName().compareTo(that.groupName());
+            } else {
+                return -1;
+            }
+        } else {
+            if (that.isGroup()) {
+                return 1;
+            } else {
+                return ComparisonChain.start()
```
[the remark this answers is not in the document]
trustin: Ah, completely forgot about functional composition. :-)

core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
```diff
@@ -32,53 +33,89 @@
...
+     */
+    public CompletableFuture<List<Endpoint>> initialEndpointsFuture() {
+        return initialEndpointsFuture;
+    }
+
+    /**
+     * Waits until the initial {@link Endpoint}s are ready.
+     */
+    public List<Endpoint> awaitInitialEndpoints() {
+        return initialEndpointsFuture.join();
```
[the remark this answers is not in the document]
trustin: Good point. Added a variant that requires timeout. I also switched from `join()` to `get()` so that the caller deals with interruption.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -16,150 +16,190 @@
...
-     * @param defaultPort the port to use when the DNS answer does not contain one. {@code 0} indicates an
-     *     unspecified port, meaning the port will use a protocol-specific well-defined port number
-     *     (e.g., 80, 443).
+     * @param hostname the hostname to query DNS queries for
+     * @param port     the port of the {@link Endpoint}s
      */
-    public static DnsAddressEndpointGroup of(String hostname, int defaultPort) {
-        return of(hostname, defaultPort, CommonPools.workerGroup().next());
+    public static DnsAddressEndpointGroup of(String hostname, int port) {
+        return new DnsAddressEndpointGroupBuilder(hostname, port).build();
```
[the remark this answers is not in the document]
trustin: Agreed. Removed `port` from the constructor and added `.port(int)`.

trustin:
Just added `DnsTextEndpointGroup` which allows you to source TXT records.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsEndpointGroup.java
```diff
@@ -0,0 +1,239 @@
+/*
+ * Copyright 2017 LINE Corporation
```
imasahiro: 2018?

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsTextEndpointGroup.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright 2017 LINE Corporation
```
imasahiro: 2018?

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/DnsTextEndpointGroupTest.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright 2017 LINE Corporation
```
minwoox: 2018?

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/TestDnsServer.java
```diff
@@ -0,0 +1,124 @@
...
+import io.netty.handler.codec.dns.DatagramDnsResponse;
+import io.netty.handler.codec.dns.DatagramDnsResponseEncoder;
+import io.netty.handler.codec.dns.DnsQuestion;
+import io.netty.handler.codec.dns.DnsResponse;
+import io.netty.handler.codec.dns.DnsResponseCode;
+import io.netty.handler.codec.dns.DnsSection;
+import io.netty.util.NetUtil;
+import io.netty.util.ReferenceCountUtil;
+
+final class TestDnsServer implements AutoCloseable {
```
minwoox: Boom!

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -16,150 +16,183 @@
...
+import io.netty.handler.codec.dns.DnsRecordType;
+import io.netty.resolver.ResolvedAddressTypes;
+import io.netty.resolver.dns.DnsServerAddressStreamProvider;
+import io.netty.util.NetUtil;
 
 /**
- * {@link DynamicEndpointGroup} which resolves targets using DNS address queries (A and AAAA). This is useful
- * for environments where service discovery is handled using DNS - for example, Kubernetes uses SkyDNS for
- * service discovery.
+ * {@link DynamicEndpointGroup} which resolves targets using DNS address queries ({@code A} and {@code AAAA}).
```
anuraaga: Consider changing to `{@link DnsEndpointGroup} which...`

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -16,150 +16,183 @@
...
+                continue;
+            }
+
+            // Convert the content into an IP address into an endpoint.
+            final String ipAddr;
+            final byte[] addrBytes = new byte[contentLen];
+            content.getBytes(content.readerIndex(), addrBytes);
+
+            if (contentLen == 16) {
+                // 'result' contains both A and AAAA records.
```
anuraaga: There isn't any variable named `result` here - do you mean `records may contain both A and AAAA records.`?

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -16,150 +16,183 @@
...
+                            addrBytes[14] == 0x00 && addrBytes[15] == 0x01) {
+                            // Loopback address (::1)
+                            if (hasLoopbackARecords) {
+                                // Contains an IPv4 loopback address already; skip.
+                                continue;
+                            } else {
+                                ipAddr = "::1";
+                            }
+                        } else {
+                            // IPv4-compatible address.
```
anuraaga: Out of curiosity, do you think Netty should handle this conversion? I found this document about compatibility in Oracle docs and wonder if the JDK does this transparently.

https://docs.oracle.com/cd/E19683-01/817-0573/transition-4/index.html

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroupBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+     * and IPv4 addresses will be included in {@link Endpoint}s. IPv4 addresses are enabled by default.
+     */
+    public DnsAddressEndpointGroupBuilder ipV4Enabled(boolean ipV4Enabled) {
+        this.ipV4Enabled = ipV4Enabled;
+        return this;
+    }
+
+    /**
+     * Sets whether IPv6 addresses are enabled. If enabled, {@code AAAA} records are retrieved from DNS servers
+     * and IPv6 address will be included in {@link Endpoint}s. IPv6 addresses are enabled by default unless
```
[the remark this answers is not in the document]
anuraaga: It's annoying that there is both the concept of `enabled` and `preferred`. The JDK default behavior is to only return IPv4, unless there are no IPv4 in which case only return IPv6, unless there are no IPv6 in which case resolution fails. We should try to reproduce that here.

What do you think about just leaving these knobs out and let users use the standard JDK JVM flags since reproducing the flags in yet another place seems tedious? I believe the Netty behavior (after the recent fix) should work without us needing to tweak it in code.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsEndpointGroup.java
```diff
@@ -0,0 +1,239 @@
...
+                private final List<DnsRecord> records = new ArrayList<>();
+                private int remaining = numQuestions;
+
+                @Override
+                public void operationComplete(Future<List<DnsRecord>> future) throws Exception {
+                    if (future.isSuccess()) {
+                        final List<DnsRecord> records = future.getNow();
+                        if (aggregatedPromise.isDone()) {
+                            // parentFuture has been notified already, which means the previous query has
+                            // failed; discard the records since it's not useful anymore.
```
anuraaga: Shouldn't we only fail if all queries fail?

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/TestDnsServer.java
anuraaga: Replace the apache one with this in netty? :)

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/DnsNameEncoder.java
```diff
@@ -0,0 +1,44 @@
...
+
+final class DnsNameEncoder {
+
+    static void encodeName(String name, ByteBuf out) {
+        DefaultDnsRecordEncoderTrampoline.INSTANCE.encodeName(name, out);
+    }
+
+    private DnsNameEncoder() {}
+
+    // Hacky trampoline class to be able to access encodeName
```
anuraaga: I forget if I suggested making this accesible from netty code, but now that we've been making changes there wonder if we can modify the API to avoid this hack

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/DnsEndpointGroupBuilderTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Test
+    public void ttl() {
+        assertThat(builder().minTtl()).isOne();
+        assertThat(builder().maxTtl()).isEqualTo(Integer.MAX_VALUE);
+        final Builder builderWithCustomTtl = builder().ttl(10, 20);
+        assertThat(builderWithCustomTtl.minTtl()).isEqualTo(10);
+        assertThat(builderWithCustomTtl.maxTtl()).isEqualTo(20);
+        assertThatThrownBy(() -> builder().ttl(0, 10)).isInstanceOf(IllegalArgumentException.class);
+        assertThatThrownBy(() -> builder().ttl(20, 10)).isInstanceOf(IllegalArgumentException.class);
+        builder().ttl(1, 1);
```
anuraaga: Assertions?

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -16,150 +16,183 @@
...
+            } else if (type == DnsRecordType.AAAA) {
+                if (contentLen != 16) {
+                    warnInvalidRecord(DnsRecordType.AAAA, content);
+                    continue;
+                }
+            } else {
+                continue;
+            }
+
+            // Convert the content into an IP address into an endpoint.
```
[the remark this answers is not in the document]
trustin: The conversion is two-step process actually. content -> IP addr + IP addr -> endpoint. Let me change to `... content into an IP address and then into an endpoint`.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsEndpointGroup.java
```diff
@@ -0,0 +1,239 @@
...
+        final long serverTtl = records.stream().mapToLong(DnsRecord::timeToLive).min().orElse(minTtl);
+        final int effectiveTtl = (int) Math.max(Math.min(serverTtl, maxTtl), minTtl);
+
+        try {
+            setEndpoints(onDnsRecords(records, effectiveTtl));
+            scheduledFuture = eventLoop.schedule(this::sendQueries, effectiveTtl, TimeUnit.SECONDS);
+        } catch (Throwable t) {
+            logger.warn("{} Failed to process the DNS query result: {}", logPrefix, records, t);
+            records.forEach(ReferenceCountUtil::safeRelease);
+            scheduledFuture = eventLoop.schedule(this::sendQueries, effectiveTtl, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
trustin: There's no point of retrying sooner than TTL in this case because the content of the DNS records will not change until then. If no content has changes, it is likely the same exception will be raised again.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
trustin: `DnsEndpointGroup` is a package-private class so we can't refer to it here.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
trustin: Oops, just a cruft I forgot to remove during refactoring. :-)

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroupBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Builds a new {@link DnsAddressEndpointGroup} that sources its {@link Endpoint} list from the {@code A} or
+ * {@code AAAA} DNS records of a certain hostname.
+ */
+public final class DnsAddressEndpointGroupBuilder
+        extends DnsEndpointGroupBuilder<DnsAddressEndpointGroupBuilder> {
+
+    private int port;
+    private boolean ipV4Enabled = true;
+    private boolean ipV6Enabled = NetUtil.isIpV4StackPreferred();
```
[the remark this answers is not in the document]
trustin: Thanks. Changed to `!NetUtil.isIpV4StackPreferred()`

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroupBuilder.java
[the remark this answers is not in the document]
trustin: These flags are actually only useful for testing. Let me make then package-private.

trustin:
@anuraaga @minwoox @hyangtack @imasahiro Ready for another round. Hoping to make it into today's release if this passes the review today.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsEndpointGroup.java
anuraaga: Hmm - I wonder if there's a case where one query will consistently fail for an extended period of time, in which case I guess the working records should still be used (it's why Java queries for both IPv4 and IPv6 even though it almost always only uses IPv4, but it could use IPv6 if that was failing for some reason)

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -16,150 +16,188 @@
...
     }
 
-    /**
-     * Starts polling for service updates.
-     */
-    public void start() {
-        checkState(scheduledFuture == null, "already started");
-        scheduledFuture = eventLoop.scheduleAtFixedRate(this::query, 0, queryInterval.getSeconds(),
-                                                        TimeUnit.SECONDS);
+    private static List<DnsQuestion> newQuestions(String hostname, ResolvedAddressTypes resolvedAddressTypes) {
```
hyangtack: Just curious. It seems that the returned list is empty if `resolvedAddressTypes` is `null`, and the constructor specifies `@Nullable ResolvedAddressTypes resolvedAddressTypes`. Is this intended?

anuraaga: I think `resolvedAddressTypes` is `@Nullable`, though I think that would result in empty questions which seems off.

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroupTest.java
```diff
@@ -13,161 +13,273 @@
...
-        assertThat(endpointGroup.endpoints()).containsExactlyElementsOf(ENDPOINTS_NO_PORT);
+    public void backoffWithPartialResponse() throws Exception {
+        try (TestDnsServer server = new TestDnsServer(ImmutableMap.of(
+                // Respond A record only.
+                new DefaultDnsQuestion("backoff.com.", A),
+                new DefaultDnsResponse(0).addRecord(ANSWER, newAddressRecord("backoff.com", "1.1.1.1"))
+        ))) {
+            try (DnsAddressEndpointGroup group = new DnsAddressEndpointGroupBuilder("backoff.com")
+                    .serverAddresses(server.addr())
+                    .resolvedAddressTypes(ResolvedAddressTypes.IPV4_PREFERRED)
```
anuraaga: I could be mistaken, but I can't find a test that doesn't call this - from my reading such a case would result in an assertion `!questions.isEmpty()` to fail.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
trustin: Oops. Fixed. Now checks if `resolvedAddressTypes` is `null` and use `IPv4_ONLY` ov `IPv4_PREFERRED` based on the return value of `NetUtil.isIpV4StackPreferred()`.

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroupTest.java
trustin: Adde a new test case `platformDefault`. Could you review again?

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsEndpointGroup.java
trustin: I see. Let me update this.

core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -16,150 +16,198 @@
...
-        this.hostname = requireNonNull(hostname, "hostname");
-        this.defaultPort = defaultPort;
-        this.resolver = requireNonNull(resolver, "resolver");
-        this.eventLoop = requireNonNull(eventLoop, "eventLoop");
-        this.queryInterval = requireNonNull(queryInterval, "queryInterval");
-    }
+        if (resolvedAddressTypes == null) {
+            if (NetUtil.isIpV4StackPreferred()) {
+                resolvedAddressTypes = ResolvedAddressTypes.IPV4_ONLY;
+            } else {
```
anuraaga: How about:

```
else if (NetUtil.isIpV6AddressesPreferred()) resolvedAddressTypes = IPV6_PREFERRED
else resolvedAddressTypes = IPV4_PREFERRED
```

trustin: It doesn't really matter, because for both cases we just send both A and AAAA.

trustin: To be honest, I'm not sure, either. Let's see if people files an issue :-)

trustin:
Thanks a lot, reviewers!

PR #1150, opened 2018-04-15
Update sphinx-gradle-plugin to 2.0.2

trustin:
- Sphinx 1.7.2
- Much faster build time
- Remove the forked sphinx_rtd_theme

I didn't try this on OS X. Please give it a try and let me know if it doesn't break anything and it shortens build time. :-)


trustin:
Thanks, reviewers :-)

PR #1151, opened 2018-04-16
(WIP)Support "defaultValue" attribute for @Param and @Header

LiYing2010:
Change List:
(1)Remove `@Default` annotation
(2)Support `defaultValue` attribute for `@Param` and `@Header` annotations
(3)Support `@AliasFor` annotation like Spring Framework (but much more simpler)

TODO:
(1)implement `@AliasFor` mechanism
(2)update official document


LiYing2010:
this plan is discarded

trustin:
I really appreciate your understanding. However, if you are really strong about `@Param(name="name", defaultValue="default")` pattern, please let us know. We could discuss again about introducing it without removing `@Default`, although we will have to ask our users as well.

LiYing2010:
Maybe it will be good if we create a poll about this, and ask users which pattern they like more

PR #1152, opened 2018-04-17
Handle non-OK HTTP status in gRPC responses.

anuraaga:
Currently, if a gRPC client gets a non-OK HTTP status, it fails with a cryptic error when trying to deframe the non-gRPC response body. Now, it gets the correct status instead.

Just copied the function as-iss from upstream.

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/internal/grpc/HttpStreamReader.java
+++ b/grpc/src/main/java/com/linecorp/armeria/internal/grpc/HttpStreamReader.java
@@ -30,6 +30,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpObject;
+import com.linecorp.armeria.common.HttpStatus;
 
 import io.grpc.Decompressor;
 import io.grpc.DecompressorRegistry;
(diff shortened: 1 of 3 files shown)


grpc/src/main/java/com/linecorp/armeria/internal/grpc/HttpStreamReader.java
```diff
@@ -95,6 +98,22 @@ public void onNext(HttpObject obj) {
             // Only clients will see headers from a stream. It doesn't hurt to share this logic between server
             // and client though as everything else is identical.
             HttpHeaders headers = (HttpHeaders) obj;
+
+            if (!sawLeadingHeaders) {
+                sawLeadingHeaders = true;
+                if (headers.status() == null) {
```
trustin: Shouldn't we do something similar to what we did in: https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/HttpResponseAggregator.java#L54 ?

trustin:
On the other hand, all LINE devs are at workshop until Friday, so our availability may be scarce. :-)

grpc/src/main/java/com/linecorp/armeria/internal/grpc/HttpStreamReader.java
anuraaga: Thanks - added skipping of informationals and made error message for missing status code same as upstream

PR #1153, opened 2018-04-20
Remove unthrown Exceptions from EndpointGroupRegistryTest

dongjinleekr:


trustin:
I was OK-ish accept these 'remove redundant `throws Exception` clauses for test classes' PRs, mainly because it doesn't do any harm, but I'm not really sure if this ever increases our code quality or improves readability. Removal of `throws Exception` from a test method is even sometimes controversial because a developer may have to add a `throws` clause again when he or she updates the test code.

Unless there's a good reason for this change, I'm not going to accept 'remove redundant `throws Exception` clasues for test classes` PRs from now on. I'd be happy to accept 'remove redundant `throws MoreSpecificException` clauses', though.

Let me leave this PR for a few more days just in case there's really good justification.

PR #1157, opened 2018-04-24
Write rpc method name to access log

hyangtack:
Motivation:
Currently there is no way to get rpc method name from access log.

Modifications:
- Write rpc method name to request line log.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
+++ b/core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
@@ -34,6 +34,7 @@ import javax.annotation.Nullable;
 import com.google.common.annotations.VisibleForTesting;
 
 import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.RpcRequest;
 import com.linecorp.armeria.common.logging.RequestLog;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -181,8 +182,14 @@ public Object getMessage0(RequestLog log) {
                             Instant.ofEpochMilli(log.requestStartTimeMillis()), defaultZoneId));
                 case REQUEST_LINE:
                     final HttpHeaders headers = log.requestHeaders();
-                    return headers.method().name() + ' ' +
-                           headers.path() + ' ' +
+                    final Object requestContent = log.requestContent();
+                    final String path;
+                    if (requestContent instanceof RpcRequest) {
+                        path = headers.path() + '#' + ((RpcRequest) requestContent).method();
```
trustin: Could we use a `StringBuilder`?

hyangtack: Thanks. Fixed!

trustin:
Thanks, @hyangtack !

PR #1158, opened 2018-04-24
Avoid RequestLogAvailabilityException while writing access logs

hyangtack:
Modifications:
- Fix a bug introduced by #1157


trustin:
A better fix would be use `RequestLog.addListener()` so that the `Consumer<RequestLog>` is invoked after all `RequestLogAvailability` is fulfilled:

    reqCtx.log().addListener(accessLogWriter::accept, RequestLogAvailability.COMPLETE);

.. instead of modifying the logger implementation.

There are two classes that call `Consumer.accept(RequestLog)`:

- `HttpResponseSubscriber`
- `HttpServerHandler`

trustin:
Closing in favor of #1159

PR #1164, opened 2018-04-26
Support `List<String>` type for `@Header` annotation

hyangtack:
Motivation:
Several HTTP headers can be sent with the same name. We missed that kind of situation before. Need to enhance annotated HTTP service to support it.

Modifications:
- Support `List<String>` type for `@Header` annotation.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
@@ -28,6 +28,8 @@ import static java.util.Objects.requireNonNull;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
@@ -264,7 +266,7 @@ final class AnnotatedHttpServiceMethod {
                     entries.add(Parameter.ofPathParam(
                             validateAndNormalizeSupportedType(parameterInfo.getType()), param));
                 } else {
-                    entries.add(createOptionalSupportedParam(
+                    entries.add(createHttpComponentParameter(
                             parameterInfo, ParameterType.PARAM, param));
                 }
                 continue;
@@ -272,7 +274,7 @@ final class AnnotatedHttpServiceMethod {
 
             final String header = findHeaderName(parameterInfo);
             if (header != null) {
-                entries.add(createOptionalSupportedParam(parameterInfo, ParameterType.HEADER, header));
+                entries.add(createHttpComponentParameter(parameterInfo, ParameterType.HEADER, header));
                 continue;
             }
 
@@ -318,27 +320,65 @@ final class AnnotatedHttpServiceMethod {
      * Creates a {@link Parameter} instance which describes a parameter of the annotated method. If the
      * parameter type is {@link Optional}, its actual argument's type is used.
      */
-    private static Parameter createOptionalSupportedParam(java.lang.reflect.Parameter parameterInfo,
+    private static Parameter createHttpComponentParameter(java.lang.reflect.Parameter parameterInfo,
                                                           ParameterType paramType, String paramValue) {
-        final Default aDefault = parameterInfo.getAnnotation(Default.class);
-        final boolean isOptionalType = parameterInfo.getType() == Optional.class;
+        assert paramType == ParameterType.PARAM ||
+               paramType == ParameterType.HEADER;
 
+        final Default aDefault = parameterInfo.getAnnotation(Default.class);
         // Set the default value to null if it was not specified.
         final String defaultValue = aDefault != null ? getSpecifiedValue(aDefault.value()).get() : null;
+
         final Class<?> type;
-        if (isOptionalType) {
+        final Class<?> wrapperType;
+
+        final Type parameterizedType = parameterInfo.getParameterizedType();
+        if (parameterizedType instanceof ParameterizedType) {
             try {
-                type = (Class<?>) ((ParameterizedType) parameterInfo.getParameterizedType())
-                        .getActualTypeArguments()[0];
-            } catch (Exception e) {
-                throw new IllegalArgumentException("Invalid optional parameter: " + parameterInfo.getName(), e);
+                type = (Class<?>) ((ParameterizedType) parameterizedType).getActualTypeArguments()[0];
+                wrapperType = validateWrapperAndElementType(paramType, parameterInfo.getType(), type);
+            } catch (Throwable cause) {
+                throw new IllegalArgumentException("Invalid optional parameter: " + parameterInfo.getName(),
+                                                   cause);
             }
         } else {
             type = parameterInfo.getType();
+            wrapperType = null;
         }
 
-        return new Parameter(paramType, !isOptionalType && aDefault == null, isOptionalType,
-                             validateAndNormalizeSupportedType(type), paramValue, defaultValue, null);
+        final boolean isRequired = wrapperType != Optional.class && aDefault == null;
+        return new Parameter(paramType, isRequired,
+                             validateAndNormalizeSupportedType(type), wrapperType,
+                             paramValue, defaultValue, null);
+    }
+
+    private static Class<?> validateWrapperAndElementType(ParameterType paramType,
+                                                          Class<?> clazz, Class<?> elementClazz) {
+        assert paramType == ParameterType.PARAM ||
+               paramType == ParameterType.HEADER;
+
+        if (clazz == Optional.class) {
+            return Optional.class;
+        }
+
+        // A list of string is supported only for HTTP headers.
+        if (paramType == ParameterType.HEADER && elementClazz == String.class) {
+            if (clazz == List.class) {
+                return ArrayList.class;
+            }
+            if (List.class.isAssignableFrom(clazz)) {
+                try {
+                    // Only if there is a default constructor.
+                    clazz.getConstructor();
+                    return clazz;
+                } catch (Throwable cause) {
+                    throw new IllegalArgumentException("Unsupported type: " + clazz.getName() +
+                                                       '<' + String.class.getName() + '>', cause);
+                }
+            }
+        }
+        throw new IllegalArgumentException("Unsupported type: " + clazz.getName() +
+                                           '<' + elementClazz.getName() + '>');
     }
 
     /**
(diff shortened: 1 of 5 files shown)


hyangtack:
This is 1st PR for resolving #1164 . I will send the next PR that `BeanRequestConverterFunction` also supports `List<String>` for `@Header` annotation.

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
```diff
@@ -801,6 +802,11 @@ public void testRequestHeaderInjection() throws Exception {
             request.setHeader("a-name", "minwoox");
             testBody(hc, request, "minwoox is awesome");
 
+            request = post("/11/customHeader");
+            request.addHeader("a-name", "minwoox");
+            request.addHeader("a-name", "giraffe");
+            testBody(hc, request, "minwoox:giraffe is awesome");
```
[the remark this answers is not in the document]
inch772: image

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -313,27 +315,65 @@ private HttpResponse convertException(ServiceRequestContext ctx, HttpRequest req
      * Creates a {@link Parameter} instance which describes a parameter of the annotated method. If the
      * parameter type is {@link Optional}, its actual argument's type is used.
      */
-    private static Parameter createOptionalSupportedParam(java.lang.reflect.Parameter parameterInfo,
+    private static Parameter createHttpComponentParameter(java.lang.reflect.Parameter parameterInfo,
                                                           ParameterType paramType, String paramValue) {
-        final Default aDefault = parameterInfo.getAnnotation(Default.class);
-        final boolean isOptionalType = parameterInfo.getType() == Optional.class;
+        assert paramType == ParameterType.PARAM ||
+               paramType == ParameterType.HEADER;
```
trustin: How about appending ` : String.valueOf(paramType)`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -313,27 +315,65 @@ private HttpResponse convertException(ServiceRequestContext ctx, HttpRequest req
...
+        final boolean isRequired = wrapperType != Optional.class && aDefault == null;
+        return new Parameter(paramType, isRequired,
+                             validateAndNormalizeSupportedType(type), wrapperType,
+                             paramValue, defaultValue, null);
+    }
+
+    private static Class<?> validateWrapperAndElementType(ParameterType paramType,
+                                                          Class<?> clazz, Class<?> elementClazz) {
+        assert paramType == ParameterType.PARAM ||
+               paramType == ParameterType.HEADER;
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -313,27 +315,65 @@ private HttpResponse convertException(ServiceRequestContext ctx, HttpRequest req
...
+
+        if (clazz == Optional.class) {
+            return Optional.class;
+        }
+
+        // A list of string is supported only for HTTP headers.
+        if (paramType == ParameterType.HEADER && elementClazz == String.class) {
+            if (clazz == List.class) {
+                return ArrayList.class;
```
trustin: Could we support:

- `Iterable` and `List` -> `ArrayList`
- `Set` -> `LinkedHashSet`

..?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -449,21 +485,42 @@ private static String entryDefaultValue(Parameter entry) {
...
+    private static Object httpHeaderValue(Parameter entry, HttpRequest req) {
         final String name = entry.name();
         assert name != null;
-        final String value = req.headers().get(AsciiString.of(name));
-        if (value != null) {
-            return value;
+
+        final Class<?> wrapperType = entry.wrapperType();
+        if (wrapperType != null && List.class.isAssignableFrom(wrapperType)) {
+            assert entry.type() == String.class;
```
trustin: Could we append ` : entry.type().getName()`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -449,21 +485,42 @@ private static String entryDefaultValue(Parameter entry) {
...
+            try {
+                @SuppressWarnings("unchecked")
+                final List<String> list = (List<String>) wrapperType.newInstance();
+
+                // We do not call convertParameter() here because the element type is String.
+                final List<String> values = req.headers().getAll(AsciiString.of(name));
+                if (values != null || !values.isEmpty()) {
```
trustin: `&&`?

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceBuilderTest.java
```diff
@@ -118,6 +121,21 @@ public Object root(@Param("a") byte a) {
                                                  }
                                              },
                                              LoggingService.newDecorator());
+
+        new ServerBuilder().annotatedService(new Object() {
+            @Get("/")
+            public Object root(@Header("a") List<String> a) {
+                return null;
```
trustin: Could use `void` now?

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceBuilderTest.java
```diff
@@ -118,6 +121,21 @@ public Object root(@Param("a") byte a) {
...
+            }
+        });
+
+        new ServerBuilder().annotatedService(new Object() {
+            @Get("/")
+            public Object root(@Header("a") ArrayList<String> a,
+                               @Header("a") LinkedList<String> b) {
+                return null;
```
trustin: Could use `void` now?

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceBuilderTest.java
```diff
@@ -193,6 +211,28 @@ public Object root(@Param("name") List<String> name) {
                 return null;
             }
         })).isInstanceOf(IllegalArgumentException.class);
+
+        assertThatThrownBy(() -> new ServerBuilder().annotatedService(new Object() {
+            @Get("/test")
+            public Object root(@Header("name") List<Object> name) {
+                return null;
```
trustin: Could use `void` now?

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceBuilderTest.java
```diff
@@ -193,6 +211,28 @@ public Object root(@Param("name") List<String> name) {
...
+            }
+        })).isInstanceOf(IllegalArgumentException.class);
+
+        assertThatThrownBy(() -> new ServerBuilder().annotatedService(new Object() {
+            @Get("/test")
+            public Object root(@Header("name") NoDefaultConstructorList<String> name) {
+                return null;
```
trustin: Could use `void` now?

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
```diff
@@ -571,8 +572,8 @@ public String aHeader(@Header String ifMatch) {
         }
 
         @Post("/customHeader")
-        public String customHeader(@Header String aName) {
-            return aName + " is awesome";
+        public String customHeader(@Header List<String> aName) {
+            return Strings.join(aName).with(":") + " is awesome";
```
trustin: Could use `String.join()` in JDK

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
hyangtack: Sorry, I can't see what you mean. :-( Would you explain why you want to append that?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
hyangtack: Ok, let me make a fix to support `Set<String>` as well.

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
hyangtack: Do you mean `assert entry.type().getName().equals(String.class.getName())`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
hyangtack: Oops! Thanks!

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceTest.java
hyangtack: Oops! Thanks!

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -318,27 +322,68 @@ private HttpResponse convertException(ServiceRequestContext ctx, HttpRequest req
...
+
+        if (clazz == Optional.class) {
+            return Optional.class;
         }
 
-        return new Parameter(paramType, !isOptionalType && aDefault == null, isOptionalType,
-                             validateAndNormalizeSupportedType(type), paramValue, defaultValue, null);
+        // A list of string is supported only for HTTP headers.
+        if (paramType == ParameterType.HEADER && elementClazz == String.class) {
+            if (clazz == Iterable.class || clazz == List.class) {
```
trustin: I missed the case of `Collection.class`. :sweat_smile:

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -318,27 +322,68 @@ private HttpResponse convertException(ServiceRequestContext ctx, HttpRequest req
...
+        final Type parameterizedType = parameterInfo.getParameterizedType();
+        if (parameterizedType instanceof ParameterizedType) {
             try {
-                type = (Class<?>) ((ParameterizedType) parameterInfo.getParameterizedType())
-                        .getActualTypeArguments()[0];
-            } catch (Exception e) {
-                throw new IllegalArgumentException("Invalid optional parameter: " + parameterInfo.getName(), e);
+                type = (Class<?>) ((ParameterizedType) parameterizedType).getActualTypeArguments()[0];
+                wrapperType = validateWrapperAndElementType(paramType, parameterInfo.getType(), type);
+            } catch (Throwable cause) {
```
minwoox: Do we need to catch here? It seems like `validateWrapperAndElementType` throws same exception with the enough information

hyangtack: You're right. Let me move the line to the out of the try-catch block. Thanks!

trustin:
Thanks, @hyangtack !

PR #1167, opened 2018-04-27
Only create composite bytebufs if there is more than one incoming gRPC frame.

anuraaga:
Currently, we use a `CompositeByteBuf` to collect incoming frames, and we read required bytes into another `CompositeByteBuf` as data is available. This causes a lot of overhead for the common case where the message fits in one frame. There also doesn't seem to be a good reason to read into a `nextFrame` as we get bytes, instead of when we have enough bytes (I had copied the pattern in this code pretty much verbatim from upstream without thinking too much about it).

Now, a `CompositeByteBuf` for incoming frames is only created on the second one. If the stream is finished with a single frame, no composites will ever be allocated. Also, incoming bytes are only read when we have enough bytes to deliver a message - this means we don't need to allocate a composite to keep track of read bytes pre-delivery.

After
```
# Run complete. Total time: 00:01:13

DownstreamSimpleBenchmark.empty        NORMAL  thrpt   20  13314.558 ± 223.515  ops/s

Benchmark result is saved to /home/anuraag/git/armeria/benchmarks/build/reports/jmh/results.txt
```

Before
```
# Run complete. Total time: 00:01:13

DownstreamSimpleBenchmark.empty        NORMAL  thrpt   20  12508.147 ± 180.084  ops/s

Benchmark result is saved to /home/anuraag/git/armeria/benchmarks/build/reports/jmh/results.txt
`

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
+++ b/grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
@@ -175,10 +175,13 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
 
     private boolean compressedFlag;
     private boolean endOfStream;
+
     @Nullable
-    private CompositeByteBuf nextFrame;
+    private ByteBuf firstFrame;
+
     @Nullable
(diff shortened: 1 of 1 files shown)


trustin:
Thanks, @anuraaga!

PR #1172, opened 2018-05-01
Split ActiveRequestMetrics from RequestMetric to support better metri…

huydx:
…c life cycle

Related https://github.com/line/armeria/pull/962 to solve https://github.com/line/armeria/issues/961.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
@@ -18,6 +18,8 @@ package com.linecorp.armeria.common.metric;
 
 import static java.util.Objects.requireNonNull;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.function.BiFunction;
 
 import com.google.common.base.MoreObjects;
@@ -28,6 +30,7 @@ import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.RequestContext;
 import com.linecorp.armeria.common.RpcRequest;
 import com.linecorp.armeria.common.logging.RequestLog;
+import com.linecorp.armeria.common.logging.RequestLogAvailability;
 import com.linecorp.armeria.server.PathMapping;
 import com.linecorp.armeria.server.ServiceRequestContext;
 import com.linecorp.armeria.server.metric.MetricCollectingService;
@@ -73,15 +76,21 @@ public interface MeterIdPrefixFunction {
                 methodName = MoreObjects.firstNonNull(log.method().name(), "__UNKNOWN_METHOD__");
             }
 
+            final List<Tag> tags = new ArrayList<>(4); // method, hostNamePattern, pathMapping, status
+            tags.add(Tag.of("method", methodName));
+
             if (ctx instanceof ServiceRequestContext) {
                 final ServiceRequestContext sCtx = (ServiceRequestContext) ctx;
-                return new MeterIdPrefix(name,
-                                         "hostnamePattern", sCtx.virtualHost().hostnamePattern(),
-                                         "pathMapping", sCtx.pathMapping().meterTag(),
-                                         "method", methodName);
-            } else {
-                return new MeterIdPrefix(name, "method", methodName);
+                tags.add(Tag.of("hostnamePattern", sCtx.virtualHost().hostnamePattern()));
+                tags.add(Tag.of("pathMapping", sCtx.pathMapping().meterTag()));
             }
+
+            if (log.isAvailable(RequestLogAvailability.RESPONSE_HEADERS) &&
+                log.status() != null) {
+                tags.add(Tag.of("status", String.valueOf(log.statusCode())));
+            }
+
+            return new MeterIdPrefix(name, tags);
         };
     }
 
--- a/core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
@@ -20,7 +20,7 @@ import static com.linecorp.armeria.common.metric.MoreMeters.summaryWithDefaultQu
 import static com.linecorp.armeria.common.metric.MoreMeters.timerWithDefaultQuantiles;
 
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.LongAdder;
 
 import com.linecorp.armeria.common.RequestContext;
 import com.linecorp.armeria.common.RpcResponse;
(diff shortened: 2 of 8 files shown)


core/src/test/java/com/linecorp/armeria/internal/metric/RequestMetricSupportTest.java
```diff
@@ -60,31 +60,50 @@ public void http() {
...
         ctx.logBuilder().endRequest();
+        ctx.logBuilder().endResponse();
+
         measurements = measureAll(registry);
-        assertThat(measurements).containsEntry("foo.activeRequests#value{method=POST}", 1.0)
-                                .containsEntry("foo.requestDuration#count{method=POST}", 1.0)
-                                .containsEntry("foo.requestLength#count{method=POST}", 1.0)
-                                .containsEntry("foo.requestLength#total{method=POST}", 123.0);
+        assertThat(measurements).containsEntry("foo.activeRequests#value{method=POST}", 0.0)
+                                .containsEntry("foo.requests#count{method=POST,result=success,status=200}", 1.0)
```
huydx: find that `request` metrics which contain status code is a bit akward, but to split it, we will need to split `RequestMetrics` and `ResponseMetrics` which @trustin not really interested it. Could you give me your comment?

huydx:
@trustin sorry for supperrr long delay. Finally find some time to finish the PR, could you take a look.

huydx:
gonna fix some other failed test later, so please check the code first (test for exact match metric name is a bit painful to maintain).

core/src/test/java/com/linecorp/armeria/internal/metric/RequestMetricSupportTest.java
trustin: I think it's not awkward at all because because a user would want to know the stats like 'the request length of the requests that ende up with non-200 statuses'. i.e. no need to split at all in my opinion.

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -40,56 +40,64 @@
...
     }
 
     private static void onRequestEnd(RequestLog log) {
-        final RequestMetrics metrics = requestMetrics(log);
-        metrics.requestDuration().record(log.requestDurationNanos(), TimeUnit.NANOSECONDS);
-        metrics.requestLength().record(log.requestLength());
         if (log.requestCause() != null) {
-            metrics.failure().increment();
-            metrics.active().decrementAndGet();
+            activeRequestMetrics(log).active().decrementAndGet();
```
trustin: I'm inconfident if we really should decrement here given that `onResponse()` is expected to be always invoked.

A quick search shows that `endRequest(Throwable)` is always called with `endResponse(Throwable)` on the server side, which means we don't need to decrement here.

However, on the client side, `HttpRequestSubscriber.fail(Throwable)` only calls `endRequest(Throwable)`  without `endResponse(Throwable)`.

So, how about:

- Removing the decrement here and
- Adding a call to `endResponse(Throwable)` at `HttpRequestSubscriber.fail(Throwable)`

..?

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -145,9 +157,22 @@ private RequestMetricSupport() {}
         Timer totalDuration();
     }
 
+    private static final class DefaultActiveRequestMetrics implements ActiveRequestMetrics {
+        private final AtomicInteger active;
```
trustin: - How about using `LongAdder` which is known to have less contention?
- How about extending `LongAdder` instead of using it as a field, for reduced footprint?

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -40,56 +40,64 @@
...
-                registry, idPrefix, RequestMetrics.class, DefaultRequestMetrics::new);
+        final MeterIdPrefix idPrefixActive = new MeterIdPrefix(idPrefix.name("activeRequests"),
+                                                               idPrefix.tags());
+
+        final ActiveRequestMetrics activeRequestMetrics =  MicrometerUtil.register(
+                registry, idPrefixActive, ActiveRequestMetrics.class, DefaultActiveRequestMetrics::new);
 
-        ctx.attr(ATTR_REQUEST_METRICS).set(requestMetrics);
-        requestMetrics.active().incrementAndGet();
+        ctx.attr(ATTR_ACTIVE_REQUEST_METRICS).set(activeRequestMetrics);
```
trustin: I'm curious if we really need to store `ActiveRequestMetrics` as a context attribute. We could just do this:

```java
// At setup():
.. Remove ctx.log().addListener(log -> onResponse(...)) ..

// At onRequestStart():
final ActiveRequestMetrics activeRequestMetrics = MicrometerUtil.register(...);
log.addListener(log -> onResponse(log, meterIdPrefixFunction, activeRequestMetrics), COMPLETE);

// At onResponse():
.. Do not access the context attribute at all, because we have direct refernece to `activeRequestMetrics` ..
```

As a result, the attribute value type of `ATTR_REQUEST_METRICS` could be just `Boolean` to determine if `setup()` was called before.

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -216,12 +234,14 @@ public Timer totalDuration() {
...
 
-    private static final class RequestMetricsPlaceholder implements RequestMetrics {
+    private static final class ActiveRequestMetricsPlaceHolder implements ActiveRequestMetrics {
         @Override
         public AtomicInteger active() {
             throw new IllegalStateException();
         }
+    }
 
+    private static final class RequestMetricsPlaceholder implements RequestMetrics {
```
trustin: This seems unused?

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -216,12 +234,14 @@ public Timer totalDuration() {
         }
     }
 
-    private static final class RequestMetricsPlaceholder implements RequestMetrics {
+    private static final class ActiveRequestMetricsPlaceHolder implements ActiveRequestMetrics {
```
trustin: With the changes I proposed above, this placeholder won't be necessary at all.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -392,6 +392,7 @@ private void handleRequest(ChannelHandlerContext ctx, DecodedHttpRequest req) th
                     logBuilder.endRequest();
                 } else {
                     logBuilder.endRequest(cause);
+                    logBuilder.endResponse(cause);
```
huydx: @trustin I found one more place that should add endResponse beside HttpRequestSubscriber#fail(cause) , could you confirm it?

trustin:
@huydx I guess it's just a flaky test.

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -33,63 +33,50 @@
...
-    private static final RequestMetrics PLACEHOLDER = new RequestMetricsPlaceholder();
-
     public static void setup(RequestContext ctx, MeterIdPrefixFunction meterIdPrefixFunction) {
-        if (ctx.hasAttr(ATTR_REQUEST_METRICS)) {
-            return;
-        }
-
-        // Set the attribute to the placeholder so that calling this method again has no effect.
-        // The attribute will be set to the real one later in onRequestStart()
-        ctx.attr(ATTR_REQUEST_METRICS).set(PLACEHOLDER);
```
trustin: We need at least a boolean attribute so that calling `setup()` multiple times does not add the log listener multiple times:

```java
private static final AttributeKey<Boolean> ATTR_REQUEST_METRICS_SET = ...;

public static void setup(...) {
    if (ctx.hasAttr(ATTR_REQUEST_METRICS_SET) { return; }
    ctx.attr(ATTR_REQUEST_METRICS_SET).set(true);
    ...
}
```

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -33,63 +33,50 @@
...
         ctx.log().addListener(log -> onRequestStart(log, meterIdPrefixFunction),
                               RequestLogAvailability.REQUEST_HEADERS,
                               RequestLogAvailability.REQUEST_CONTENT);
-        ctx.log().addListener(RequestMetricSupport::onRequestEnd,
-                              RequestLogAvailability.REQUEST_END);
-        ctx.log().addListener(RequestMetricSupport::onResponse,
-                              RequestLogAvailability.COMPLETE);
     }
 
     private static void onRequestStart(RequestLog log, MeterIdPrefixFunction meterIdPrefixFunction) {
```
trustin: Could be renamed to `onRequest()`?

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -33,63 +33,50 @@
...
+                                                               idPrefix.tags());
+
+        ActiveRequestMetrics activeRequestMetrics = MicrometerUtil.register(
+                registry, idPrefixActive, ActiveRequestMetrics.class,
+                (reg, prefix) ->
+                        reg.gauge(prefix.name(), prefix.tags(),
+                                  new ActiveRequestMetrics(), ActiveRequestMetrics::doubleValue));
+        ctx.log().addListener(requestLog -> onResponse(requestLog, meterIdPrefixFunction, activeRequestMetrics),
+                              RequestLogAvailability.COMPLETE);
+        activeRequestMetrics.increment();
```
trustin: Could be incremented before calling `addListener()` just in case `RequestLogAvailability` is fulfilled early?

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -33,63 +33,50 @@
...
-        ctx.attr(ATTR_REQUEST_METRICS).set(requestMetrics);
-        requestMetrics.active().incrementAndGet();
+        final MeterIdPrefix idPrefixActive = new MeterIdPrefix(idPrefix.name("activeRequests"),
+                                                               idPrefix.tags());
+
+        ActiveRequestMetrics activeRequestMetrics = MicrometerUtil.register(
+                registry, idPrefixActive, ActiveRequestMetrics.class,
+                (reg, prefix) ->
+                        reg.gauge(prefix.name(), prefix.tags(),
+                                  new ActiveRequestMetrics(), ActiveRequestMetrics::doubleValue));
```
trustin: We could then just replace `ActiveRequestMetrics` with `LongAdder` completely? e.g. `new LongAdder(), LongAdder::doubleValue`?

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
trustin: I think we don't need to call `endResponse(Throwable)` here because `HttpResponseSubscriber` will handle it. How about adding some comment instead?

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
huydx: It seems that MicrometerUtil will use class type as identification for mapping with MeterIdPrefix, so I thought it's better to not use std type like LongAdder, or it will prevent us to register further LongAdder. HDYT?

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -80,6 +81,12 @@ static MeterIdPrefixFunction ofDefault(String name) {
                                          "pathMapping", sCtx.pathMapping().meterTag(),
                                          "method", methodName);
             } else {
+                if (log.isAvailable(RequestLogAvailability.RESPONSE_HEADERS) &&
```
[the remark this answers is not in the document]
huydx: sounds reasonable, added too

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -73,15 +76,21 @@ static MeterIdPrefixFunction ofDefault(String name) {
                 methodName = MoreObjects.firstNonNull(log.method().name(), "__UNKNOWN_METHOD__");
             }
 
+            List<Tag> tags = new ArrayList<>();
```
huydx: this will increase memory footprint a bit, but without the List, I can not think about any another way without a lot of if/else. Could anybody give me some advice?

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
trustin: SGTM

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
trustin: - I think it's fine. If escape analysis works as expected, it should be stack-allocated. You could specify the initial capacity correctly though. How about 4 (method, hostnamePattern, pathMapping, status)?
- Please make `tags` final.

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -73,15 +76,21 @@ static MeterIdPrefixFunction ofDefault(String name) {
...
-                                         "pathMapping", sCtx.pathMapping().meterTag(),
-                                         "method", methodName);
-            } else {
-                return new MeterIdPrefix(name, "method", methodName);
+                tags.add(Tag.of("hostnamePattern", sCtx.virtualHost().hostnamePattern()));
+                tags.add(Tag.of("pathMapping", sCtx.pathMapping().meterTag()));
+            }
+
+            if (log.isAvailable(RequestLogAvailability.RESPONSE_HEADERS) &&
+                    log.status() != null) {
```
trustin: Indentation

```java
if (log.isAvailable(...) &&
    log.status() != null) {
```

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
huydx: not sure List<Tag> will be stack-allocated until it will be escaped to return method, anyway specify initial cap SGTM, and added

trustin: `MeterIdPrefix` will make a copy of it so it can stay at stack if the constructor of `MeterIdPrefix` is inlined.

trustin:
@anuraaga Any comments?

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -392,6 +392,7 @@ private void handleRequest(ChannelHandlerContext ctx, DecodedHttpRequest req) th
                     logBuilder.endRequest();
                 } else {
                     logBuilder.endRequest(cause);
+                    // NB: logBuidler.endResponse(cause) will be called by HttpResponseSubscriber below
```
anuraaga: `logBuilder`

trustin:
@huydx Could you fix the type?

huydx:
@trustin fixed!

trustin:
Thanks, @huydx !

Issue #1141, opened 2018-04-10
Missing properties in AnnotatedServiceRegitrationBean


trustin:
Reported by @taicki 

There's no way to specify exception handlers and converters when registering an annotated service via `AnnotatedServiceRegistrationBean`. A user may be able to use `@ExceptionHandler`, `@RequestConverter` or `@ResponseConverter`, but it may not be possible if they do not have default constructor.

PR #1173, opened 2018-05-02
Provides a way to specify exception/request/response handlers in AnnotatedServiceRegitrationBean

imasahiro:
Fixes https://github.com/line/armeria/issues/1141

### DIFF
--- a/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
+++ b/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
@@ -21,6 +21,8 @@ import java.util.function.Function;
 
 import javax.validation.constraints.NotNull;
 
+import com.google.common.collect.ImmutableList;
+
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.server.Service;
@@ -168,9 +170,17 @@ public class AnnotatedServiceRegistrationBean {
      * Sets the exception handlers of the annotated service object.
      */
     public AnnotatedServiceRegistrationBean setExceptionHandlers(
-            @NotNull Collection<? extends ExceptionHandlerFunction> exceptionHandlers) {
+            Collection<? extends ExceptionHandlerFunction> exceptionHandlers) {
         this.exceptionHandlers = exceptionHandlers;
         return this;
+    }
+
+    /**
+     * Sets the exception handlers of the annotated service object.
+     */
+    public AnnotatedServiceRegistrationBean setExceptionHandlers(
+            ExceptionHandlerFunction... exceptionHandlers) {
+        return setExceptionHandlers(ImmutableList.copyOf(exceptionHandlers));
     }
 
     /**
@@ -190,6 +200,14 @@ public class AnnotatedServiceRegistrationBean {
     }
 
     /**
+     * Sets the request converters of the annotated service object.
+     */
+    public AnnotatedServiceRegistrationBean setRequestConverters(
+            RequestConverterFunction... requestConverters) {
+        return setRequestConverters(ImmutableList.copyOf(requestConverters));
+    }
+
+    /**
      * Returns the response converters of the annotated service object.
      */
     public Collection<? extends ResponseConverterFunction> getResponseConverters() {
@@ -203,5 +221,13 @@ public class AnnotatedServiceRegistrationBean {
             Collection<? extends ResponseConverterFunction> responseConverters) {
         this.responseConverters = responseConverters;
         return this;
+    }
+
+    /**
+     * Sets the response converters of the annotated service object.
+     */
+    public AnnotatedServiceRegistrationBean setResponseConverters(
+            ResponseConverterFunction... responseConverters) {
+        return setResponseConverters(ImmutableList.copyOf(responseConverters));
     }
 }
--- a/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
+++ b/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
@@ -15,6 +15,8 @@
  */
 package com.linecorp.armeria.spring;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.function.Function;
 
 import javax.validation.constraints.NotNull;
@@ -22,6 +24,7 @@ import javax.validation.constraints.NotNull;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.server.Service;
+import com.linecorp.armeria.server.annotation.ExceptionHandlerFunction;
 
 /**
  * A bean with information for registering an annotated service object. It enables dropwizard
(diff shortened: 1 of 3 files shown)


spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
```diff
@@ -130,4 +156,52 @@ public AnnotatedServiceRegistrationBean setDecorator(
...
+     */
+    public Collection<? extends ExceptionHandlerFunction> getExceptionHandlers() {
+        return exceptionHandlers;
+    }
+
+    /**
+     * Sets the exception handlers of the annotated service object.
+     */
+    public AnnotatedServiceRegistrationBean setExceptionHandlers(
+            @NotNull Collection<? extends ExceptionHandlerFunction> exceptionHandlers) {
```
hyangtack: nit: Is `@NotNull` not necessary? Or, do we need to put `@NotNull` on the parameter of the other setter methods?

trustin: Not necessary because all our classes are non-null by default, as annotated with `NonNullByDefault` at package level.

trustin: Actually, this annotation is different from JSR-305. It seems to be used by Spring for validation. i.e. maybe necessary?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
```diff
@@ -33,7 +38,10 @@
  * >             .setServiceName("myAnnotatedService")
  * >             .setPathPrefix("/my_service")
  * >             .setService(new MyAnnotatedService())
- * >             .setDecorator(LoggingService.newDecorator());
+ * >             .setDecorator(LoggingService.newDecorator())
+ * >             .setExceptionHandlers(ImmutableList.of(new MyExceptionHandler()))
+ * >             .setRequestConverters(ImmutableList.of(new MyRequestConverter()))
+ * >             .setResponseConverters(ImmutableList.of(new MyResponseConverter()));
```
trustin: - Could we also have the versions of these methods that accept varargs?
- We could define some common supertype of all these registration beans to reduce code duplication.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
```diff
@@ -130,4 +156,52 @@ public AnnotatedServiceRegistrationBean setDecorator(
...
+     */
+    public Collection<? extends RequestConverterFunction> getRequestConverters() {
+        return requestConverters;
+    }
+
+    /**
+     * Sets the request converters of the annotated service object.
+     */
+    public AnnotatedServiceRegistrationBean setRequestConverters(
+            Collection<? extends RequestConverterFunction> requestConverters) {
```
minwoox: I don't know if `@NotNull` is necessary or not. But if it needs, we should add it here as well.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
```diff
@@ -130,4 +156,52 @@ public AnnotatedServiceRegistrationBean setDecorator(
...
+     */
+    public Collection<? extends ResponseConverterFunction> getResponseConverters() {
+        return responseConverters;
+    }
+
+    /**
+     * Sets the response converters of the annotated service object.
+     */
+    public AnnotatedServiceRegistrationBean setResponseConverters(
+            Collection<? extends ResponseConverterFunction> responseConverters) {
```
minwoox: ditto

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
imasahiro: Removed.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
imasahiro: Added varargs versions.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
imasahiro: No need to add here.

imasahiro:
Updated

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
trustin: Then perhaps we could update this example for less verbosity?

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaAutoConfigurationTest.java
```diff
@@ -105,11 +116,41 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req) throws
         }
     }
 
+    public static class IllegalArgumentExceptionHandler implements ExceptionHandlerFunction {
+
+        @Override
+        public HttpResponse handleException(RequestContext ctx, HttpRequest req, Throwable cause) {
+            if (cause instanceof IllegalArgumentException) {
+                return HttpResponse.of(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "exception");
```
trustin: Could be shorten to `HttpResponse.of("exception")`

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaAutoConfigurationTest.java
```diff
@@ -105,11 +116,41 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req) throws
...
+    }
+
+    public static class StringResponseConverter implements ResponseConverterFunction {
+        @Override
+        public HttpResponse convertResponse(ServiceRequestContext ctx, @Nullable Object result)
+                throws Exception {
+            if (result instanceof String) {
+                return HttpResponse.of(HttpStatus.OK,
+                                       MediaType.ANY_TEXT_TYPE,
+                                       HttpData.ofUtf8(result.toString()));
```
trustin: Could be shorten to `HttpResponse.of(MediaType.ANY_TEXT_TYPE, result.toString())`

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
minwoox: Then, don't we have to `requireNonNull()` here?

trustin:
Thanks, @imasahiro!

PR #1179, opened 2018-05-03
Add 'apiplural' directive for referring to a type in plural form

trustin:
Motivation:

While documenting, it is often necessary to refer to a type in plural
form:

    This class retrieves the list of :api:`Endpoint`\s from ZooKeeper.

Although `\` does its job here, the rendered output isn't very pretty,
mainly because:

- There's spacing between `Endpoint` and `s`. The spacing between them
  should be moved after the plural suffix `s`.
- There's gray background at `Endpoint` but not at `s`. Both `Endpoint`
  and `s` should have the same background color.

Modifications:

- Introduce a new directive `apiplural` which pluralizes a type
  reference automatically

Result:

- Aesthetics and convenience

Before:
screenshot from 2018-05-03 12-09-31

After:
screenshot from 2018-05-03 12-08-26

### DIFF
--- a/site/src/sphinx/_extensions/api.py
+++ b/site/src/sphinx/_extensions/api.py
@@ -9,6 +9,14 @@ import re
 
 
 def api_role(role, rawtext, text, lineno, inliner, options={}, content=[]):
+    return api_role_internal(False, role, rawtext, text, lineno, inliner, options, content)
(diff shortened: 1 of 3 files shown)


trustin:
Thanks!

Issue #650, opened 2017-06-30
Spring integration fails with NoClassDefFoundError without armeria-thrift


trustin:
/cc @imasahiro @yuinacor 

We recently removed armeria-thrift(-shaded) from the dependency list of Spring integration. However, Spring integration won't start if a user doesn't add it to the dependency list even if he or she doesn't use Thrift registration beans at all. Spring integration should work without armeria-thrift if a user doesn't have a Thrift service.

delegacy:
I have the same experience. Is there a reason to exclude it from starter, instead of letting users choose the dependency by excluding in their pom?

trustin:
Just a hindsight? :-)

imasahiro:
Here is full stacktrace.
```
2018-05-04 14:56:54.945 ERROR 32626 --- [main] o.s.boot.SpringApplication               : Application run failed

java.lang.IllegalStateException: Error processing condition on org.springframework.boot.actuate.autoconfigure.audit.AuditAutoConfiguration$AuditEventRepositoryConfiguration
	at org.springframework.boot.autoconfigure.condition.SpringBootCondition.matches(SpringBootCondition.java:64)
	at org.springframework.context.annotation.ConditionEvaluator.shouldSkip(ConditionEvaluator.java:109)
	at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader$TrackedConditionEvaluator.shouldSkip(ConfigurationClassBeanDefinitionReader.java:443)
	at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader.loadBeanDefinitionsForConfigurationClass(ConfigurationClassBeanDefinitionReader.java:128)
	at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader.loadBeanDefinitions(ConfigurationClassBeanDefinitionReader.java:117)
	at org.springframework.context.annotation.ConfigurationClassPostProcessor.processConfigBeanDefinitions(ConfigurationClassPostProcessor.java:328)
	at org.springframework.context.annotation.ConfigurationClassPostProcessor.postProcessBeanDefinitionRegistry(ConfigurationClassPostProcessor.java:233)
	at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanDefinitionRegistryPostProcessors(PostProcessorRegistrationDelegate.java:273)
[...]

PR #1182, opened 2018-05-04
Fetches thrift service names using reflection not to access armeria-thrift package

imasahiro:
Motivations:
- armeria-thrift-* dependencies are removed from armeria-spring-boot*-autoconfigure to choose thrift
  implementations by a user but it causes NoClassDefFoundError if armeria-thrift-* dependency is not
  populated in dependency list.

Changes:
- Add a reflection based implementation to retrieve names of thrift service
- Fixes https://github.com/line/armeria/issues/650


trustin:

PR #1187, opened 2018-05-09
Use Highlight.js instead of Pygments for syntax highlighting

trustin:
Motivation:

I can't find a nice theme for Pygments.

Modifications:

- Use Highlight.js instead of Pygments for syntax highlighting

Result:

- Code snippets look better.

screenshot-2018-5-9 running a grpc service armeria 0 63 2-snapshot documentation


trustin:
Thanks!

PR #1190, opened 2018-05-10
Do not export '*.percentile' and '*.histogram' gauges to Dropwizard MetricRegistry

trustin:
Related: https://github.com/micrometer-metrics/micrometer/blob/v1.0.0/micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardMeterRegistry.java#L77

Motivation:

Since 1.0.0, Micrometer registers additional gauges for percentile and
histogram values of Timers and DistributionSummarys for some monitoring
systems such as Dropwizard and Statsd.

I think it's not a sensible default to export additional gauges to
Dropwizard Timers and DistributionSummarys because Dropwizard already
provides its own percentile values via Timers and Histograms. Also,
because Dropwizard does not provide any aggregatation, such additional
gauges with tags will make things worse.

Modifications:

- Make the DropwizardMeterRegistry created by DropwizardMeterRegistries
  filter out the extra gauges registered when a Timer or a
  DistributionSummary is created

Result:

- Less surprising default

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistriesTest.java
+++ b/core/src/test/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistriesTest.java
@@ -15,6 +15,8 @@
  */
 package com.linecorp.armeria.common.metric;
 
+import static com.linecorp.armeria.common.metric.DropwizardMeterRegistries.DEFAULT_DROPWIZARD_CONFIG;
+import static com.linecorp.armeria.common.metric.DropwizardMeterRegistries.DEFAULT_NAME_MAPPER;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistriesTest.java
```diff
@@ -0,0 +1,108 @@
...
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
+import io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry;
+import io.micrometer.prometheus.PrometheusMeterRegistry;
+import io.prometheus.client.Collector.MetricFamilySamples;
+import io.prometheus.client.Collector.MetricFamilySamples.Sample;
+
+public class DropwizardMeterRegistriesTest {
+    @Test
+    public void unwantedGaugesAreFilteredOut() {
```
huydx: nits: how about add one more test with default DropwizardMeterRegistry, to make sure that they have added gauges with ". percentile" and ".histogram" naming, to prevent case they change the naming rule and our filter without our awareness?

trustin: Sounds like a good idea. Will do.

trustin: @huydx Adde the test case. Please review again.

trustin:
Thanks, reviewers!

PR #1191, opened 2018-05-10
Refactor annotated HTTP service

### LINKED ISSUE #1175: Refactor annotated HTTP service
The first goal is to extract a common code from `AnnotatedHttpServiceMethod` and `BeanRequestConverterFunction` to support `@Param`, `@Header` and `@Default` annotations in a same way.

It might be to support `@RequestObject` annotation for a bean so that a bean is able to hold another bean. I'm not sure yet if this feature is good or not.

hyangtack:
Motivation:
The first goal is to extract a common code from AnnotatedHttpServiceMethod and BeanRequestConverterFunction to support `@Param`, `@Header` and `@Default` annotations in a same way.
Next, converting a bean in a recursive way would be useful. e.g. a bean inside another bean.

Modifications:
- Add `AnnotatedValueResolver` in order to resolve values in a same way between annotated service method and bean request converter.
- Change `AnnotatedHttpServices` class to `AnnotatedHttpServiceFactory` and refactor it.
- Add `AnnotatedBeanFactory` which replaces `BeanRequestConverterFunction`.
- Now `@Default` and `@RequestObject` can be also put on a field, method and constructor.
- Now `List<?>`, `Set<?>` and `Optional<List<?>>` can be specified with `@Param` and `@Header`.
  - `@Param` is only allowed to HTTP parameters. A path variable cannot be specified as a collection.
  - The element type of a collection is not limited to `String` anymore. The elements would be automatically converted to the desired type.
    e.g. `List<Integer>`, `Set<UserEnumType>`, etc.

Result:
- Closes #1175


hyangtack:
Let me add negative test cases more.

core/src/main/java/com/linecorp/armeria/server/AbstractVirtualHostBuilder.java
```diff
@@ -446,9 +447,9 @@ public B annotatedService(String pathPrefix, Object service,
         requireNonNull(decorator, "decorator");
         requireNonNull(exceptionHandlersAndConverters, "exceptionHandlersAndConverters");
 
-        final List<AnnotatedHttpService> entries =
-                AnnotatedHttpServices.build(pathPrefix, service, exceptionHandlersAndConverters);
-        entries.forEach(e -> service(e.pathMapping(), decorator.apply(e.decorator().apply(e))));
+        final List<AnnotatedHttpServiceElement> elements =
+                AnnotatedHttpServiceFactory.find(pathPrefix, service, exceptionHandlersAndConverters);
+        elements.forEach(el -> service(el.pathMapping(), decorator.apply(el.decorator().apply(el.service()))));
```
trustin: `e` or `elem`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedBeanFactory.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+        return ret;
+    }
+
+    private AnnotatedBeanFactory() {}
+
+    /**
+     * An identifier of the registered bean factory.
+     */
+    static final class BeanId {
```
trustin: `BeanFactoryId`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedBeanFactory.java
```diff
@@ -0,0 +1,257 @@
...
+        if (!factories.containsKey(beanId)) {
+            factories.put(beanId, firstNonNull(createFactory(beanId, objectResolvers), unsupportedBeanFactory));
+        }
+        return beanId;
+    }
+
+    /**
+     * Returns a factory of the specified {@link BeanId}.
+     */
+    @Nullable
```
trustin: Not a nullable method? It returns an `Optional`.

core/src/main/java/com/linecorp/armeria/server/AnnotatedBeanFactory.java
```diff
@@ -0,0 +1,257 @@
...
+            // A default constructor exists but there is no annotated field or method.
+            return null;
+        }
+
+        // Suppress Java language access checking.
+        constructor.getKey().setAccessible(true);
+        fields.forEach(field -> field.getKey().setAccessible(true));
+        methods.forEach(method -> method.getKey().setAccessible(true));
+
+        logger.info("Register bean factory: {}", beanId);
```
trustin: Probably better using `debug()`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpService.java
```diff
@@ -16,85 +16,221 @@
...
+        this.exceptionHandlers = ImmutableList.copyOf(
+                requireNonNull(exceptionHandlers, "exceptionHandlers"));
+        this.responseConverters = ImmutableList.copyOf(
+                requireNonNull(responseConverters, "responseConverters"));
+
+        aggregationStrategy = AggregationStrategy.resolve(resolvers);
+
+        final Class<?> returnType = method.getReturnType();
+        returnHttpResponseType = HttpResponse.class.isAssignableFrom(returnType);
+        returnCompletionStageType = CompletionStage.class.isAssignableFrom(returnType);
```
trustin: Perhaps we could use an `enum` and use `switch-case` in `serve0()`?

core/src/test/java/com/linecorp/armeria/server/AnnotatedValueResolverTest.java
```diff
@@ -0,0 +1,959 @@
...
+                arg -> existingHttpHeaders.contains(arg.getArguments()[0]) ? headerValues
+                                                                           : ImmutableList.of());
+    }
+
+    static boolean shouldExistHttpHeader(AnnotatedValueResolver element) {
+        return element.isRequired() ||
+               existingHttpHeaders.contains(AsciiString.of(element.name()));
+    }
+
+    static boolean shouldExistHttpParameter(AnnotatedValueResolver element) {
```
trustin: `shouldHttpParameterExist`?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -263,7 +263,9 @@ Getting an HTTP header
 
 Armeria also provides :api:`@Header` annotation to inject an HTTP header value into a parameter.
 The parameter annotated with :api:`@Header` can also be specified as one of the built-in types as follows.
-:api:`@Default` and ``Optional<?>`` are also supported.
+:api:`@Default` and ``Optional<?>`` are also supported. Unlike :api:`@Param` annotation, :api:`@Header`
+annotation supports ``List`` or ``Set`` as its type because HTTP headers can be added several times with the
```
trustin: I'm curious if we need to support `List` or `Set` for query strings, e.g. `?foo=1&foo=2` could be translated into `@Param("foo") List<String>`.

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -0,0 +1,764 @@
...
+            final String name = findName(param, typeElement);
+            if (pathParams.contains(name)) {
+                final AnnotatedValueResolver element =
+                        ofHttpComponent(param, name, typeElement, type,
+                                        annotatedElement.getAnnotation(Default.class),
+                                        httpPathOrParamValueResolver(ctx -> ctx.context().pathParam(name)),
+                                        false, AggregationStrategy.NONE);
+                if (!element.isRequired()) {
+                    throw new IllegalArgumentException(
+                            "Path variable '" + name + "' should not be an optional.");
```
trustin: - `Path variable 'foo' should not be an optional.` -> `Path variable 'foo' annotated with @Default` cannot be an Optional.`
- Using `@Default` on an `Optional` parameter is not optimal, but it's probably OK to support such a use case? We may want to log some warning though.

hyangtack:
Will ping @line/cdtf if it's ready for review

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -0,0 +1,1001 @@
...
+                          .add("resolver", resolver)
+                          .add("enumConverter", enumConverter)
+                          .toString();
+    }
+
+    static Builder builder(AnnotatedElement annotatedElement, Type type) {
+        return new Builder(annotatedElement, type);
+    }
+
+    static class Builder {
```
hyangtack: I added a builder for `AnnotatedValueResolver` to make each annotation's specification clear like https://github.com/line/armeria/pull/1191/files#diff-9df84bbcae0217f0942af4e4d6a3f8c0R244.

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -0,0 +1,1001 @@
...
+            return request;
+        }
+
+        @Nullable
+        AggregatedHttpMessage message() {
+            return message;
+        }
+
+        HttpParameters httpParameters() {
+            // Use a temporary variable to reduce the number of reads of the volatile field.
```
hyangtack: Borrowed from https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/concurrent/LazyInitializer.java#L97

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceRequestConverterTest.java
```diff
@@ -755,54 +753,6 @@ public void testDefaultRequestConverter_bean3() throws Exception {
         assertThat(response.content().toStringUtf8()).isEqualTo(expectedResponseContent);
     }
 
-    @Test
-    public void testDefaultRequestConverter_beanDefineError() throws Exception {
```
hyangtack: Need to add these cases tomorrow

site/src/sphinx/server-annotated-service.rst
```diff
@@ -263,7 +280,9 @@ Getting an HTTP header
 
 Armeria also provides :api:`@Header` annotation to inject an HTTP header value into a parameter.
 The parameter annotated with :api:`@Header` can also be specified as one of the built-in types as follows.
-:api:`@Default` and ``Optional<?>`` are also supported.
+:api:`@Default` and ``Optional<?>`` are also supported. :api:`@Header` annotation also supports
+``List<String>`` or ``Set<String>`` as its type because HTTP headers can be added several times
```
hyangtack: Need to update tomorrow.

core/src/test/java/com/linecorp/armeria/server/AnnotatedHttpServiceRequestConverterTest.java
hyangtack: Moved to https://github.com/line/armeria/pull/1191/files#diff-acfda6b218745abeaece7a779cd740b7R35

hyangtack:
@line/cdtf Review please. :-)

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceMethod.java
```diff
@@ -1,712 +0,0 @@
...
-        this.responseConverters = ImmutableList.copyOf(
-                requireNonNull(responseConverters, "responseConverters"));
-
-        parameters = parameters(method, pathMapping.paramNames(), !requestConverters.isEmpty());
-        final Class<?> returnType = method.getReturnType();
-        isAsynchronous = HttpResponse.class.isAssignableFrom(returnType) ||
-                         CompletionStage.class.isAssignableFrom(returnType);
-        aggregationStrategy = AggregationStrategy.resolve(parameters);
-
-        this.method.setAccessible(true);
```
kojilin: I found the problem of CI, so we need to call this in new code.

kojilin: Maybe need to add test for `IllegalAccess` case.

hyangtack: Thanks, @kojilin ! I missed that code. 😱

core/src/main/java/com/linecorp/armeria/server/AnnotatedBeanFactory.java
```diff
@@ -0,0 +1,261 @@
...
+            // A default constructor exists but there is no annotated field or method.
+            return null;
+        }
+
+        // Suppress Java language access checking.
+        constructor.getKey().setAccessible(true);
+        fields.forEach(field -> field.getKey().setAccessible(true));
+        methods.forEach(method -> method.getKey().setAccessible(true));
+
+        logger.debug("Register bean factory: {}", beanFactoryId);
```
trustin: `Registered a bean factory` ?

core/src/main/java/com/linecorp/armeria/server/AnnotatedBeanFactory.java
```diff
@@ -0,0 +1,261 @@
...
+        };
+    }
+
+    @Nullable
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    private static <T> Entry<Constructor<T>, List<AnnotatedValueResolver>> findConstructor(
+            BeanFactoryId beanFactoryId, List<RequestObjectResolver> objectResolvers) {
+
+        Constructor<T> defaultConstructor = null;
+        final List<Entry<Constructor<T>, List<AnnotatedValueResolver>>> candidates = new ArrayList<>();
```
trustin: I think we don't really use a `List` here but just a single value holder:

```java
for (constructor : constructors) {
    ...
    if (!resolvers.isEmpty()) {
        if (candidate == null) {
            candidate = new SimpleImmutableEntry<>(...);
        } else {
            throw new IllegalARgumentException("too many annotated constructors in ...");
        }
    }
}

if (candidate == null) {
    return defaultConstructor != null ? ...;
}

return candidate;
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceFactory.java
```diff
@@ -136,14 +126,139 @@
...
+                                                      List<ResponseConverterFunction> baseResponseConverters) {
+
+        final Set<Annotation> methodAnnotations = httpMethodAnnotations(method);
+        if (methodAnnotations.isEmpty()) {
+            throw new IllegalArgumentException("HTTP Method specification is missing: " + method.getName());
+        }
+
+        final Set<HttpMethod> methods = toHttpMethods(methodAnnotations);
+        if (methods.isEmpty()) {
+            throw new IllegalArgumentException("HTTP Method specification is missing: " + method.getName());
```
trustin: How about:

    method.getDeclaringClass().getName() + '#' + method.getName() +
    " must have an HTTP method annotation."

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpServiceFactory.java
```diff
@@ -136,14 +126,139 @@
...
+        final Set<String> expectedParamNames = pathMapping.paramNames();
+        final Set<String> requiredParamNames =
+                resolvers.stream()
+                         .filter(AnnotatedValueResolver::isPathVariable)
+                         .map(AnnotatedValueResolver::httpElementName)
+                         .collect(Collectors.toSet());
+
+        if (!expectedParamNames.containsAll(requiredParamNames)) {
+            final Set<String> missing = Sets.difference(requiredParamNames, expectedParamNames);
+            throw new IllegalArgumentException("Missing @Param exists: " + missing);
```
trustin: How about:

    "cannot find path variables: " + missing

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -0,0 +1,1080 @@
...
+            if (value != null) {
+                return value;
+            }
+            throw new IllegalArgumentException("No suitable request converter found for a @" +
+                                               RequestObject.class.getSimpleName() + " '" +
+                                               resolver.elementType().getSimpleName() + '\'');
+        };
+    }
+
+    // Test-purpose.
```
trustin: How about adding `@VisibleForTesting` to `annotataion()` as well?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
trustin: I think it's too trivial and well known pattern (double-checked locking) and thus there's no need for an attribution.

core/src/test/java/com/linecorp/armeria/server/AnnotatedValueResolverTest.java
```diff
@@ -0,0 +1,1023 @@
...
+    @BeforeClass
+    public static void setUp() throws Exception {
+        when(resolverContext.httpParameters()).thenReturn(parameters);
+        when(resolverContext.context()).thenReturn(context);
+        when(resolverContext.request()).thenReturn(request);
+        when(resolverContext.message()).thenReturn(message);
+
+        when(context.request()).thenReturn(request);
+        when(context.query())
+                .thenReturn(new QueryStringEncoder(
```
trustin: Seems like `new QueryStringEncoder()` is not necessary at all or you need to use `QueryStringEncoder.addParam()` for each parameters?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -242,6 +242,23 @@ as the value of parameter ``name``. If there is no parameter named ``name`` in t
         }
     }
 
+If multiple parameters exist with the same name in a query string, they can be injected as a ``List<?>``
+or ``Set<?>`` type, e.g. ``/hello1?number=1&number=2&number=3``. You can use :api:`@Default` annotation
```
trustin: How about just `a List<?> or a Set<?>`? (no 'type')

site/src/sphinx/server-annotated-service.rst
```diff
@@ -263,7 +280,8 @@ Getting an HTTP header
 
 Armeria also provides :api:`@Header` annotation to inject an HTTP header value into a parameter.
 The parameter annotated with :api:`@Header` can also be specified as one of the built-in types as follows.
-:api:`@Default` and ``Optional<?>`` are also supported.
+:api:`@Default` and ``Optional<?>`` are also supported. :api:`@Header` annotation also supports
+``List<?>`` or ``Set<?>`` as its type because HTTP headers can be added several times with the same name.
```
trustin: Ditto

site/src/sphinx/server-annotated-service.rst
```diff
@@ -455,7 +479,7 @@ converting ``MyRequest``.
 Injecting value of parameters and HTTP headers into a Java object
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
-:api:`BeanRequestConverterFunction` is a built-in request converter for Java object. You can use it by
+There is a built-in request converter for Java object. You can use it by
```
trustin: - There is a built-in -> Armeria provides a generic built-in
- Java object. You can use it by -> Java objects which is activated by

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -0,0 +1,1079 @@
...
+    }
+
+    boolean isAnnotationType(Class<? extends Annotation> target) {
+        return annotation != null &&
+               annotation.annotationType() == target;
+    }
+
+    @Nullable
+    String httpElementName() {
+        // Currently, this is non-null only if the element one of the HTTP path variable, parameter or header.
```
minwoox: element is?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -0,0 +1,1079 @@
...
+                                                 : null;
+    }
+
+    @Nullable
+    private static EnumConverter<?> enumConverter(Class<?> elementType) {
+        if (!elementType.isEnum()) {
+            return null;
+        }
+        return enumConverters.computeIfAbsent(elementType, newElementType -> {
+            logger.info("Register Enum {}", newElementType);
```
minwoox: Registered an Enum?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -0,0 +1,1079 @@
...
+        /**
+         * Sets whether the element can be annotated with {@link Default} annotation.
+         */
+        private Builder supportDefault(boolean supportDefault) {
+            this.supportDefault = supportDefault;
+            return this;
+        }
+
+        /**
+         * Sets an {@link AnnotatedElement} which is to infer its type.
```
minwoox: is used?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -242,6 +242,23 @@ as the value of parameter ``name``. If there is no parameter named ``name`` in t
...
+or ``Optional<?>`` class here, too.
+
+.. code-block:: java
+
+    public class MyAnnotatedService {
+        @Get("/hello1")
+        public HttpResponse hello1(@Param("number") List<Integer> numbers) { ... }
+
+        @Get("/hello2")
+        public HttpResponse hello2(@Param("number") @Default("1") List<Integer> numbers) { ... }
```
minwoox: Could you add a comment how the default value gets into the `numbers`?

PR #1192, opened 2018-05-10
Add CircuitBreakerStrategy

minwoox:
Motivation:
Currently, `CircuitBreaker.onFailure()` is only called when an `Exception` is raised.
When a user gets `503` response, it is counted as a success althought we must not.

Modifications:
- Add `CircuitBreakerStrategy` which determines a `Response` is successful or not
- Make `CircuitBreakerClient` as abstract
- Add `CircuitBreakerHttpClient` and `CircuitBreakerRpcClient` which are subclasses of `CircuitBreakerClient`
- Remove `CircuitBreaker.onFailure(Exception cause)` so that `CircuitBreaker` does not decide a failure, `CircuitBreakerStrategy` does the job
- Remove `CircuitBreakerConfig.exceptionFilter()` as well

Result:
- Robust circuit breaker system


core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Returns the {@link CircuitBreakerStrategy} that determines a {@link Response} as successful
+     * when its {@link HttpStatus} is not {@link HttpStatusClass#SERVER_ERROR} and there was no
+     * {@link Exception} raised.
+     */
+    static CircuitBreakerStrategy<HttpResponse> onServerErrorStatus() {
+        return onStatus(
+                (status, thrown) -> thrown == null &&
+                                    (status == null || status.codeClass() != HttpStatusClass.SERVER_ERROR));
```
hyangtack: I'm not sure but it seems that `status != null && status.codeClass() != HttpStatusClass.SERVER_ERROR` more makes sense. WDYT?

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClient.java
```diff
@@ -118,27 +71,33 @@ public O execute(ClientRequestContext ctx, I req) throws Exception {
...
+
+    /**
+     * Invoked when the {@link CircuitBreaker} is in closed state.
+     */
+    protected abstract O doExecute(ClientRequestContext ctx, I req, CircuitBreaker circuitBreaker)
+            throws Exception;
+
+    /**
+     * Reports a success or a failure to the specified {@link CircuitBreaker} according to the completed value
+     * of the specified {@code future}. If the completed value is {@code null}, this doesn't do anything.
```
hyangtack: Just curious. Is it any special reason of adding `null` as a return value? Because it seems that there were only success and failure before.

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClient.java
```diff
@@ -118,27 +71,33 @@ public O execute(ClientRequestContext ctx, I req) throws Exception {
...
+     */
+    protected static void reportSuccessOrFailure(CircuitBreaker circuitBreaker,
+                                                 CompletableFuture<Boolean> future) {
+        future.handle(voidFunction((success, unused) -> {
+            if (success != null) {
```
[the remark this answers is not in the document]
minwoox: Then I have to add another `else if`.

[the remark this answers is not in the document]
minwoox: Then I have to add another `else if`.
The intention was like this:
```
if (success == null) {
    // The user does not want to count the response as a success nor a failure, so ignore it.
} else {
    if (success) {
        circuitBreaker.onSuccess();
    } else {
        circuitBreaker.onFailure();
    }
}
```
Let me add a comment to clarify this. :-)

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+        return onStatus(
+                (status, thrown) -> status != null && status.codeClass() != HttpStatusClass.SERVER_ERROR);
+    }
+
+    /**
+     * Returns the {@link CircuitBreakerStrategy} that determines a {@link Response} as successful
+     * using the specified {@link BiFunction}.
+     *
+     * @param function the {@link BiFunction} that returns {@code true}, {@code false} or
+     *                 {@code null} according to the {@link HttpStatus} and {@link Throwable}
```
trustin: Please add the explanation for each return value.

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    static CircuitBreakerStrategy<HttpResponse> onStatus(
+            BiFunction<HttpStatus, Throwable, Boolean> function) {
+        return new HttpStatusBasedCircuitBreakerStrategy(function);
+    }
+
+    /**
+     * Returns a {@link CompletableFuture} that contains {@code true}, {@code false} or
+     * {@code null} which will be used to report the specified {@link Response} as a success, a failure or
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/NonBlockingCircuitBreaker.java
```diff
@@ -186,12 +174,12 @@ private void notifyStateChanged(CircuitState circuitState) {
             try {
                 listener.onStateChanged(this, circuitState);
             } catch (Throwable t) {
-                logger.warn("An error occured when notifying a state changed event", t);
+                logger.warn("An error occurred when notifying a state changed event", t);
```
trustin: How about `StateChange event`?

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/NonBlockingCircuitBreaker.java
```diff
@@ -201,7 +189,7 @@ private void notifyCountUpdated(EventCount count) {
             try {
                 listener.onEventCountUpdated(this, count);
             } catch (Throwable t) {
-                logger.warn("An error occured when notifying an EventCount updated event", t);
+                logger.warn("An error occurred when notifying an EventCount updated event", t);
```
trustin: How about `EventCountUpdated event`?

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/NonBlockingCircuitBreaker.java
```diff
@@ -211,7 +199,7 @@ private void notifyRequestRejected() {
             try {
                 listener.onRequestRejected(this);
             } catch (Throwable t) {
-                logger.warn("An error occured when notifying a request rejected event", t);
+                logger.warn("An error occurred when notifying a request rejected event", t);
```
trustin: How about `RequestRejected event`?

trustin:
Thanks, @minwoox and reviewers :-)

PR #1195, opened 2018-05-14
Add access log format tokens for logging response header and RequestLog

### LINKED ISSUE #1193: Add format tokens corresponding to the values in RequestLog
It would be better to have format tokens for the values, such as:
- total duration
- request start/end time, duration
- response start/end time, duration
- request length

hyangtack:
Motivation:
The default access log format is not enough for logging a service request because it does not provide a value regarding `time` and `duration`.

Modifications:
- Add `%L` format token for logging RequestLog so that a user can specify `time` and `duration`-related values, such as `requestStartTimeMillis`, `requestDurationMillis`, `totalDurationMillis`, etc.
- Add `%o` format token for the response HTTP headers.

Result:
- Closes #1193


hyangtack:
@line/cdtf Ready for review. :-)

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -93,16 +99,101 @@ public Object getMessage(RequestLog log) {
...
+        }
+
+        static DateTimeFormatter findFormatter(@Nullable String variable) {
+            if (variable == null) {
+                return defaultDateTimeFormatter;
+            }
+            // Find pre-defined formatter first.
+            switch (variable) {
+                case "BASIC_ISO_DATE":
+                    return DateTimeFormatter.BASIC_ISO_DATE;
```
trustin: Perhaps we could pre-populate `Map<String, DateTimeFormatter>` by looking for the `public static final DateTimeFormatter` fields in `DataTimeFormatter`?

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogType.java
```diff
@@ -125,7 +144,7 @@ public boolean isConditionAvailable() {
         return isConditionAvailable;
     }
 
-    public boolean isVariableRequired() {
-        return isVariableRequired;
+    public VariableRequirement variableRequirement() {
```
trustin: `VariableRequirement` looks like a package-local class. Should this method be also package-local?

core/src/test/java/com/linecorp/armeria/server/logging/AccessLogFormatsTest.java
```diff
@@ -255,6 +276,114 @@ public void requestLogAvailabilityException() {
...
+        assertThat(AccessLogger.format(format, log)).isEqualTo(
+                String.join(" ",
+                            String.valueOf(requestStartTime.toEpochMilli()),
+                            String.valueOf(requestStartTime.plus(duration).toEpochMilli()),
+                            String.valueOf(duration.toMillis())));
+
+        format = AccessLogFormats.parseCustom("\"%{requestCause}L\"");
+        assertThat(AccessLogger.format(format, log)).isEqualTo("\"-\"");
+        when(log.requestCause()).thenReturn(new IllegalArgumentException("detail_message"));
+        assertThat(AccessLogger.format(format, log)).isEqualTo("\"detail_message\"");
```
trustin: Shouldn't we also include the exception type? e.g. `IllegalArgumentException: detail_message`

hyangtack: Good point, it would be better!

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogType.java
hyangtack: Got it.

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
hyangtack: Got it!

hyangtack:
@AmosDoan 
In this PR, `%t` token is changed to have a date/time pattern as its variable. Is it enough for your case? Would you let me know what you think?

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -138,47 +147,43 @@ static DateTimeFormatter findFormatter(@Nullable String variable) {
...
-                default:
-                    try {
-                        return DateTimeFormatter.ofPattern(variable, Locale.ENGLISH);
-                    } catch (Throwable cause) {
-                        throw new IllegalArgumentException("unexpected date/time format variable: " +
-                                                           variable, cause);
-                    }
+
+            if (builtinFormatters == null) {
+                initializePredefinedFormatters();
```
trustin: Perhaps we don't really need to initialize this lazily?

core/src/test/java/com/linecorp/armeria/server/logging/AccessLogFormatsTest.java
```diff
@@ -255,6 +276,115 @@ public void requestLogAvailabilityException() {
...
+    }
+
+    @Test
+    public void timestamp() {
+        final DummyRequestContext ctx = new DummyRequestContext();
+        final DefaultRequestLog log = spy(new DefaultRequestLog(ctx));
+
+        log.startRequest(channel, SessionProtocol.H2, "www.example.com");
+        // To generate the same datetime string always.
+        final long requestStartTimeMillis = 0L;
```
minwoox: How about changing to some value which is not `0`? something like `19830725`?

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -281,10 +364,101 @@ Object getMessage0(RequestLog log) {
...
+            if (cause == null) {
+                return null;
+            }
+            cause = Exceptions.peel(cause);
+            final String message = cause.getMessage();
+            return message != null ? cause.getClass().getSimpleName() + ": " + message
+                                   : cause.getClass().getSimpleName();
+        }
+
+        private static Function<RequestLog, Object> findResolver(String variable) {
```
minwoox: I don't know if it's a good idea or not, but how about using a `Map<String, Function<RequestLog, Object>>` here? 
If you think it takes too much memory and don't worth, just ignore this comment. :-)

site/src/sphinx/server-access-log.rst
```diff
@@ -163,6 +174,119 @@ the condition.
 |                                                   | nor ``304``.                                   |
 +---------------------------------------------------+------------------------------------------------+
 
+.. _request-log:
+
+Retrieving values from RequestLog
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:api:`RequestLog` holds information about the request, so a user may want to write these values on his or
```
minwoox: these values to?

site/src/sphinx/server-access-log.rst
```diff
@@ -163,6 +174,119 @@ the condition.
...
+|                               | completely, in milliseconds                                        |
++-------------------------------+--------------------------------------------------------------------+
+| ``requestDurationNanos``      | the duration that was taken to consume or produce the request      |
+|                               | completely, in nanoseconds                                         |
++-------------------------------+--------------------------------------------------------------------+
+| ``requestLength``             | the length of the request content                                  |
++-------------------------------+--------------------------------------------------------------------+
+| ``requestCause``              | the cause of request processing failure. The detail message of a   |
+|                               | ``Throwable`` or the class name of the cause if no message is      |
+|                               | specified.                                                         |
```
minwoox: How about `The class name of the cause and the detail message of it will be contained if exists.`?

site/src/sphinx/server-access-log.rst
```diff
@@ -163,6 +174,119 @@ the condition.
...
+| ``statusCode``                | the status code of the response. e.g. ``200``                      |
++-------------------------------+--------------------------------------------------------------------+
+
+.. _timestamp-format:
+
+Customizing timestamp format
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can specify a new date/time format for the ``%t`` token with DateTimeFormatter_. You can use one of the
+following formatter which is provided by JDK as a variable of the ``%t`` token, e.g. ``%{BASIC_ISO_DATE}t``.
```
minwoox: formatters?

site/src/sphinx/server-access-log.rst
```diff
@@ -163,6 +174,119 @@ the condition.
...
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_LOCAL_TIME``      | Time without offset              | ``10:15:30``                               |
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_OFFSET_TIME``     | Time with offset                 | ``10:15:30+01:00``                         |
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_TIME``            | Time with or without offset      | ``10:15:30+01:00``; ``10:15:30``           |
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_LOCAL_DATE_TIME`` | ISO Local Date and Time          | ``2011-12-03T10:15:30``                    |
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_OFFSET_DATE_TIME``| Date Time with Offset            | ``2011-12-03T10:15:30+01:00``              |
```
minwoox: `Offset` -> `offset`?

site/src/sphinx/server-access-log.rst
```diff
@@ -163,6 +174,119 @@ the condition.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can specify a new date/time format for the ``%t`` token with DateTimeFormatter_. You can use one of the
+following formatter which is provided by JDK as a variable of the ``%t`` token, e.g. ``%{BASIC_ISO_DATE}t``.
+If you want to use your own pattern, you can specify it as the variable, e.g. ``%{yyyy MM dd}t``.
+
++-------------------------+----------------------------------+--------------------------------------------+
+| Formatter               | Description                      | Example                                    |
++=========================+==================================+============================================+
+| ``BASIC_ISO_DATE``      | Basic ISO date                   | ``20111203``                               |
```
minwoox: Date?

site/src/sphinx/server-access-log.rst
```diff
@@ -163,6 +174,119 @@ the condition.
...
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_LOCAL_DATE_TIME`` | ISO Local Date and Time          | ``2011-12-03T10:15:30``                    |
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_OFFSET_DATE_TIME``| Date Time with Offset            | ``2011-12-03T10:15:30+01:00``              |
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_ZONED_DATE_TIME`` | Zoned Date Time                  | ``2011-12-03T10:15:30+01:00[Europe/Paris]``|
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_DATE_TIME``       | Date and time with ZoneId        | ``2011-12-03T10:15:30+01:00[Europe/Paris]``|
++-------------------------+----------------------------------+--------------------------------------------+
+| ``ISO_ORDINAL_DATE``    | Year and day of year             | ``2012-337``                               |
```
minwoox: day of year ranging between 1 and 366?

site/src/sphinx/server-access-log.rst
hyangtack: Actually I copied this table from javadoc: https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html
I don't know why they capitalize some words.. 

Does it better to change the description? WDYT?

site/src/sphinx/server-access-log.rst
hyangtack: Oh, I missed fixing this description! Thanks!

core/src/test/java/com/linecorp/armeria/server/logging/AccessLogFormatsTest.java
hyangtack: Ok. But I don't like `19830725`. :stuck_out_tongue_winking_eye:

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
hyangtack: Thanks! This method is called by `ServerBuilder` which means that it's only at start-up. So I think that it's better to let it as it is. :-)

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
hyangtack: Ok, got it!

site/src/sphinx/server-access-log.rst
minwoox: Then please forget my comments :)

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -93,16 +105,90 @@ public Object getMessage(RequestLog log) {
...
+        static final ZoneId defaultZoneId = ZoneId.systemDefault();
+
+        static boolean isSupported(AccessLogType type) {
+            return type == AccessLogType.REQUEST_TIMESTAMP;
+        }
+
+        private final boolean addQuote;
+        private final DateTimeFormatter formatter;
+
+        @Nullable
```
trustin: `predefinedFormatters` seems to be non-null.

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -93,16 +105,90 @@ public Object getMessage(RequestLog log) {
...
+        private static final Map<String, DateTimeFormatter> predefinedFormatters =
```
trustin: - `predefinedFormatters` seems to be non-null.
- Could be placed before the member fields?

hyangtack: Ooops!! Sorry, that's my fault. Thank you!

PR #1196, opened 2018-05-14
Skip pending request counting at health check service

imasahiro:
Motivations:
- Armeria server counts the number of pending requests including healthcheck requests.
- If a server has many healthcheck clients and these healthcheck clients send a request
  frequently, it makes difficult for a server to shutdown on time because graceful
  shutdown depends on the number of pending requests.

Modifications:
- Adds methods to track start/finish of processing requests to Server
- Stop to count a pending healthcheck request

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Service.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Service.java
@@ -140,19 +140,4 @@ public interface Service<I extends Request, O extends Response> {
     default boolean shouldCachePath(String path, @Nullable String query, PathMapping pathMapping) {
         return pathMapping.exactPath().isPresent() && query == null;
     }
-
-    /**
-     * Invoked when this {@link Service} starts to process a {@link Request}.
-     */
-    default void startToProcessRequest(HttpRequest req,
-                                       GracefulShutdownSupport gracefulShutdownSupport) {
-        gracefulShutdownSupport.inc();
-    }
-
-    /**
-     * Invoked when this {@link Service} finishes to process a {@link Request}.
-     */
-    default void finishToProcessRequest(GracefulShutdownSupport gracefulShutdownSupport) {
-        gracefulShutdownSupport.dec();
-    }
 }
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/server/Service.java
```diff
@@ -140,4 +140,19 @@ R decorate(Function<T, R> decorator) {
     default boolean shouldCachePath(String path, @Nullable String query, PathMapping pathMapping) {
         return pathMapping.exactPath().isPresent() && query == null;
     }
+
+    /**
+     * Invoked when this {@link Service} starts to process a {@link Request}.
+     */
+    default void startToProcessRequest(HttpRequest req,
```
imasahiro: I feel, these method names and javadocs are not saying correct behavior but I could not think of better names. Could you give me suggestions?

kojilin:
I prefer `boolean canXxx()` rather than make concrete `Service` knows ShutdownSupport , e.g. `canBeIgnoredWhenShutdown` . But still doesn't have good naming idea.

core/src/main/java/com/linecorp/armeria/server/Service.java
anuraaga: Agree that providing these as lifecycle methods seems unnecessary, we can have a method to provide a characteristic of the service

Do you think these requests should also be left out of e.g., server "total active requests"? If so then something like `boolean transientRequests()` maybe? The requests are there but transient and not included as server requests, similar to how transient fields are not included in serialization. Otherwise, more explicit would be `requestsNotActive()` or similar.

trustin:
How about just this:

```java
public interface TransientService<I extends Request, O extends Response> extends Service<I, O> {}

// in HttpServerHandler.java:
final boolean transient = service.as(TransientService.class).isPresent());
if (transient) { ...inc(); }
...
if (transient) { ...dec(); }
```

core/src/main/java/com/linecorp/armeria/server/TransientService.java
```diff
@@ -0,0 +1,26 @@
...
+ * under the License.
+ */
+
+package com.linecorp.armeria.server;
+
+import com.linecorp.armeria.common.Request;
+import com.linecorp.armeria.common.Response;
+
+/**
+ * A {@link Service} that indicates to handles transient requests, for example, health check request.
```
trustin: - indicates to handles -> handles
- health check request -> health check requests

core/src/main/java/com/linecorp/armeria/server/Service.java
imasahiro: Thanks for suggestions. Let me use a word, `transient` with `TransientService` idea from trustin.

trustin:

PR #1197, opened 2018-05-15
Update IntelliJ IDEA code style and inspection profile

trustin:
- Using IntelliJ IDEA 2018.1.2
- Code style
  - Made 'public static' methods come before non-static fields
- Inspection profile
  - Changed the level of 'Interface may be annotated @FunctionInterface'
    from WARN to WEAK WARN
  - Added 'Use of obsolete date-time API'
  - Added 'Local variable or parameter can be final'
    - Tweaked to limit the warning to local variables only
    - NOTE: You may see a lot of warnings if you did not declare your local
      variables as final. However, I believe it's a very good habit to do so.
  - Added 'Test-only class or method call in production code'
  - Added 'assertEquals() may be assertSame()'
  - Added 'JUnit 4 test can be JUnit 5'
  - Added 'Obsolete assertions in JUnit 5 tests'
  - Added 'Log condition does not match logging call'
  - Added 'Number of placeholders does not match number of arguments in
    logging call'
  - Added 'Suspicious underscore in number literal'
  - Added 'equals() or hashCode() called on java.net.URL object'
  - Added 'Map or Set may contain java.net.URL objects'
  - Added 'Constant conditions & exceptions' and 'Return of null' with tweaks:
    - NOTE: You are going to have to annotate your methods, fields and
      parameters with `@Nullable` annotation to make this function properly.
      You may find this cumbersome, but I think it's very good because:
      - Much easier to catch silly NPEs or to find dead code
      - Better interop with Kotlin if annotated correctly
  - Added 'Subtraction in compareTo()'
  - Added 'Convert TestNG Javadoc to 1.5 annotations'
  - Tweaked test class naming convention a little bit


trustin:
Let me merge this once CI passes since this has been approved internally.

PR #1202, opened 2018-05-16
Fix IntelliJ IDEA inspection warnings

trustin:
Fixed various inspection warnings which showed up since we updated our
inspection profile.

Notable changes:

- Fixed mismatching number of arguments for some log messages
- Added missing `@Nullable` annotations


trustin:
Given this is of mostly mechanical changes and thus will be very tedious to review. I'll merge this once CI passes.

core/src/test/java/com/linecorp/armeria/client/Http2ClientSettingsTest.java
```diff
@@ -149,7 +149,7 @@ public void maxFrameSize() throws Exception {
                     .build();
 
             final HttpClient client = HttpClient.of(clientFactory, "http://127.0.0.1:" + port);
-            final CompletableFuture<AggregatedHttpMessage> future = client.get("/").aggregate();
+            client.get("/").aggregate();
```
imasahiro: Hm.. no need to join this future?

trustin: No, because otherwise it will wait until the response is sent. We send the response below using a Socket, so join() will fail with timeout.

PR #1209, opened 2018-05-18
Remove CircuitBreakerBasedThrottlingStrategy completely

minwoox:
Motivation:
If a user want to use `CircuitBreakerBasedThrottlingStrategy`, he or she has to use it
in the way described in
https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/throttling/CircuitBreakerBasedThrottlingStrategy.java#L35
The `CircuitBreaker` instance used in `CircuitBreakerBasedThrottlingStrategy` and behind the `ThrottlingService`
should be same if the request goes to the same remote service.
But it is hard to retrieve same instance under different `Context` and `Request` values so that
it might not work as what a user expects.
Besides, `CircuitBreaker` and `ThrottlingSerivce` have same methods which are `onSuccess()` and `onFailure()`
and their behavior is totally different.
Therefore, I think we better not have `CircuitBreakerBasedThrottlingStrategy`.

Modification:
- Remove `CircuitBreakerBasedThrottlingStrategy`

Result:
- Removed `CircuitBreakerBasedThrottlingStrategy`


minwoox:
Please let me know if there's anyone using this class and how to use it.

trustin:
Thanks, @minwoox.

PR #1212, opened 2018-05-18
Publish Tomcat8.5 project

imasahiro:
Motivations:
- Because some Tomcat API were removed from Tomcat 9, ABI compatibility is broken up (for example, [`Context#addServletMapping(String pattern, String name)`](https://tomcat.apache.org/tomcat-8.5-doc/api/org/apache/catalina/Context.html#addServletMapping(java.lang.String,%20java.lang.String)) is used in TomcatEmbeddedServletContainerFactory(spring-boot-1.5.x) was removed from Tomcat 9). For this, if Tomcat 9 dependency comes first, fails to startup spring-boot application with NoSuchMethodException.

Modifications:
- Publish armeria-tomcat8.5


trustin:
Thanks, @imasahiro !

PR #1217, opened 2018-05-22
Removes old dependencies via Brave 5

adriancole:
Brave v5 is the same as v4 except no old deps. Notably, there is no
current constant file, so that's why literals are used in this change.

See https://github.com/openzipkin/brave/issues/699

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -158,12 +158,9 @@ io.reactivex.rxjava2:
 
 io.zipkin.brave:
   brave:
-    version: &BRAVE_VERSION '4.19.2'
+    version: &BRAVE_VERSION '5.0.0'
     javadocs:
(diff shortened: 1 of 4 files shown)


trustin:
ICLA confirmed. Thanks, @adriancole !

adriancole:
happy to help :)

trustin:
Good luck for today's meetup, and I hope you enjoyed your stay in Japan. :heart:

adriancole:
Thanks! PS I made an armeria example and gave you admin (feel free to add
others). It will be a nice way to collaborate and is exactly the same
example app we use for other tools like finagle and pyramid.

adriancole:
https://github.com/openzipkin-contrib/zipkin-armeria-example

PR #1220, opened 2018-05-23
Have DynamicEndpointGroup ignore updates that do not change the list …

### LINKED ISSUE #1216: DynamicEndpointGroup should only notify when endpoints changed
Currently, `DynamicEndpointGroup` notifies listeners anytime when `setEndpoints` is called. This happens even if the endpoints didn't change, which doesn't seem very useful. I think a couple options would be

1) (preferred) `DynamicEndpointGroup` doesn't call `notifyListeners` when endpoints didn't change (by using `Object.equals` to compare the old and new ones)
2) `DynamicEndpointGroup` notifies listeners of both new and old endpoints so they can make the decision of whether to execute business logic even if nothing changed.

What do you think?

anuraaga:
…of endpoints and order them for more stable comparison.

Currently, `DynamicEndpointGroup` listeners are notified anytime an endpoint refresh happens, even if the endpoints didn't change. This doesn't seem useful so now `DynamicEndpointGroup` compares the old and new endpoints and only notifies when they change.

I guess insertion sort can be used for `addEndpoint` to avoid a copy but the complexity didn't seem worth it for a method that shouldn't be called with that much concurrency, but let me know if you'd prefer it.

Fixes #1216

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
@@ -28,6 +28,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 
 import com.linecorp.armeria.client.Endpoint;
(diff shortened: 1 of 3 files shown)


trustin:
Thanks, @anuraaga and reviewers!

PR #1222, opened 2018-05-23
Inject `Cookie` header automatically

### LINKED ISSUE #1026: Support Cookie header injection by annotated HTTP service
We may automatically decode `Cookie` header as a Java object like `io.netty.handler.codec.http.cookie.Cookie`.

hyangtack:
Motivation:
It would be better to inject `Cookie` header value into a parameter automatically when a user uses an annotated HTTP service.

Modifications:
- Add `Cookies` class to let Armeria know where `Cookie` header is injected into.

Result:
- Closes #1026

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/Cookies.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/Cookies.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.annotation;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.Set;
+
+import com.google.common.collect.ForwardingSet;
+
+import io.netty.handler.codec.http.cookie.Cookie;
+
+/**
+ * A class which holds a set of {@link Cookie}s.
+ */
+public final class Cookies extends ForwardingSet<Cookie> {
+
+    private final Set<Cookie> delegate;
+
+    public Cookies(Set<Cookie> delegate) {
+        this.delegate = requireNonNull(delegate, "delegate");
+    }
+
+    @Override
+    protected Set<Cookie> delegate() {
+        return delegate;
+    }
+}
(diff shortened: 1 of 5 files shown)


hyangtack:
Please let me know whether this fix is reasonable or not. :-) Will add a document about this feature after that.

core/src/main/java/com/linecorp/armeria/server/annotation/Cookies.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Set;
+
+import com.google.common.collect.ForwardingSet;
+
+import io.netty.handler.codec.http.cookie.Cookie;
+
+/**
+ * A class which holds a set of {@link Cookie}s.
+ */
+public final class Cookies extends ForwardingSet<Cookie> {
```
trustin: - How about splitting this class into an interface and the default implementation?
- How about adding some convenience methods? Perhaps in a follow-up PR?

hyangtack: > How about adding some convenience methods? Perhaps in a follow-up PR?

I can't find out yet. I will add the methods later. :-)

core/src/main/java/com/linecorp/armeria/server/annotation/Cookies.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.netty.handler.codec.http.cookie.Cookie;
+
+/**
+ * An interface which holds decoded {@link Cookie} instances for an HTTP request.
+ */
+public interface Cookies extends Set<Cookie> {
+
+    /**
+     * Creates an instance with the specified set of {@link Cookie}s.
```
trustin: Could you add:

    Note that this factory method does not make a copy of the specified set. Any changes made in
    the set later will be visible to {@link Cookies}.

core/src/main/java/com/linecorp/armeria/server/annotation/Cookies.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    static Cookies from(Set<Cookie> cookies) {
```
trustin: How about:

- Adding:
  - `static Cookies of(Cookie... cookies)`
  - `static Cookies copyOf(Iterable<? extends Cookie> cookies)`
  - .. which make a copy of the given collection
- and removing `from()` and calling `new DefaultCookies()` directly internally, so we do not expose a potentially unsafe operation to our users?

hyangtack: > and removing from() and calling new DefaultCookies() directly internally, so we do not expose a potentially unsafe operation to our users?

Replaced `Cookies.from(...)` with `Cookies.copyOf(...)`. Because `DefaultCookies` is package-private, so it should be at the same package with `Cookies` interface. Because `ImmutableSet` is passed into `Cookies.copyOf(...)`, there might not perform actual copy operation.

core/src/main/java/com/linecorp/armeria/server/annotation/Cookies.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * An interface which holds decoded {@link Cookie} instances for an HTTP request.
+ */
+public interface Cookies extends Set<Cookie> {
+
+    /**
+     * Creates an instance with a copy of the specified set of {@link Cookie}s.
+     */
+    static Cookies of(Cookie... cookies) {
+        return new DefaultCookies(ImmutableSet.copyOf(cookies));
```
trustin: `ImmutableSet.copyOf(requireNonNull(cookies, "cookies"))`

because a user can still pass `null`: `Cookies.of((Cookie[]) null)`

trustin:
Thanks, @hyangtack and @minwoox

PR #1226, opened 2018-05-25
More sensible default DistributionStatisticConfig for Timer and Distr…

trustin:
…ibutionSummary

Related: https://github.com/micrometer-metrics/micrometer/commit/ef2dad5bd5542c6f88c02f11c6df93ebabe02e7c#diff-a428c30af987802231f5617fea57ea4dR38

Motivation:

Micrometer folks changed the default DistributionStatisticConfig in
Micrometer 1.0.3. Our users observed unexpectedly reduced accuracy in
their percentile values.

Modification:

- Define all properties related with percentile histograms so that our
  users are not surprised.

Result:

- Less surprises

Note: The values in this PR is not final. I'd like to listen to your preferences and choose the most sensible one. What `percentilePrecision`, `bufferLength` and `expiry` would you expect as a sensible default? /cc @imasahiro @taicki @huydx @anuraaga @jwills


huydx:
nits: How about added your comment at the Why? section to the code itself? @trustin

PR #1227, opened 2018-05-28
Add an example using CorsService in Grpc documentation

minwoox:


site/src/sphinx/server-grpc.rst
```diff
@@ -110,7 +116,33 @@ to just pass ``GrpcSerializationFormats.values()``.
 
 The server will support both native gRPC_ and gRPC-Web_ from the same endpoint. Use the unofficial
 gRPC-Web-Client_ to access the service from a browser. gRPC-Web_ does not support RPC methods with streaming
-requests.
+requests. gRPC-Web-Client_ might first send ``preflighted`` requests by the HTTP ``OPTIONS`` method, in order
```
anuraaga: I think this only happens when the request is actually cross-origin (grpc-web-client is just using browser APIs which deal with cors transparently I guess)

hyangtack: How about using `preflight`? I think it's more generally used.

minwoox:
>  I guess it's not gRPC-specific

I'm totally agree with you on that. That's why I'm not explaining about CORS here, and use the word `might`.

>what do you think about creating a separate section about CORS and link to it from here?

I was going to do in that way. But I think it takes time to make a document about CORS, so I did it in the reverse. The code that I inserted here is I think, the least amount of code when I make a document about CORS and link it here.
But if you feel this is unnecessary, I will remove this. :)

anuraaga:
I think it's fine to add it here first and move later as long as we're on the same page there :)

site/src/sphinx/server-grpc.rst
```diff
@@ -110,7 +116,33 @@ to just pass ``GrpcSerializationFormats.values()``.
 
 The server will support both native gRPC_ and gRPC-Web_ from the same endpoint. Use the unofficial
 gRPC-Web-Client_ to access the service from a browser. gRPC-Web_ does not support RPC methods with streaming
-requests.
+requests. gRPC-Web-Client_ might first send ``preflight`` requests by the HTTP ``OPTIONS`` method, in order
```
anuraaga: Can we add `if the origin of the Javascript and API server are different`? I don't want it to seem like `CorsService` should always be applied.

minwoox: Added! Thanks for the good point that I missed :)

trustin:
Thanks, @minwoox !

PR #1230, opened 2018-05-29
Fix bugs where a resolved Endpoint is not handled correctly

trustin:
Motivation:

Endpoint can have an IP address with its host name, which is called
'resolved Endpoint'. This is useful when you signify a specific host out
of many that have the same host name but different IP addresses, which
is often found when using a DnsEndpointGroup.

However, there are a few places in Armeria that the resolved Endpoints
with the same host name and different IP addresses are not handled
properly:

- KeyedCircuitBreakerMapping.KeySelector.HOST does not include an IP
  address in the return value. This can make a CircuitBreaker enter the
  OPEN state for all endpoints in a DnsEndpointGroup even if only one of
  them is faulty.
- HttpHealthCheckedEndpointGroup does not use an IP address when sending
  a health check request. This can make the health checker send a health
  check request to a wrong host.

Modifications:

- Fix KeyedCircuitBreakerMapping.KeySelector.HOST so that it includes an
  IP address in the return value
- Fix HttpHealthCheckedEndpointGroup so that it uses an IP address when
  sending a health check request
- Modify HttpClientDelegate to extract the remote host name from the
  `:authority` header so that SNI works as expected.
  - Add the test case to HttpClientSniTest
- Miscellaneous:
  - Allow specifying any integer value as a parameter of Endpoint.port(int)
    so that a user can use it to check if port was specified or not:

        if (endpoint.port(-1) == -1) { /* port not specified */ }

Result:

- No unexpected misbehavior of CircuitBreakers and HttpHealthCheckedEndpointGroup
- Client-side SNI works when a user specified a custom `:authority` header.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
@@ -22,6 +22,7 @@ import java.util.concurrent.CompletableFuture;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 
 import com.linecorp.armeria.client.pool.KeyedChannelPool;
@@ -57,7 +58,7 @@ final class HttpClientDelegate implements Client<HttpRequest, HttpResponse> {
             return HttpResponse.ofFailure(new IllegalArgumentException("invalid path: " + req.path()));
         }
 
-        final String host = extractHost(ctx, endpoint, req);
+        final String host = extractHost(ctx, req, endpoint);
         final EventLoop eventLoop = ctx.eventLoop();
         final PoolKey poolKey = new PoolKey(host, endpoint.ipAddr(),
                                             endpoint.port(), ctx.sessionProtocol());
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -80,6 +86,43 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
         return res;
     }
 
+    @VisibleForTesting
+    static String extractHost(ClientRequestContext ctx, HttpRequest req, Endpoint endpoint) {
+        final HttpHeaders additionalHeaders = ctx.additionalRequestHeaders();
+        String authority = additionalHeaders.get(HttpHeaderNames.AUTHORITY);
```
minwoox: `additionalHeader.authority()`?

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -80,6 +86,43 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
...
+            // Strip the port number.
+            return authority.substring(0, colonPos);
+        }
+        if (colonPos < 0) {
+            // authority does not have a port number.
+            return authority;
+        }
+
+        // Invalid authority - ':' is the first character.
+        return endpoint.host();
```
minwoox: Should check `authority = req.authority();` before?

PR #1232, opened 2018-05-30
Add more methods related with accessing IP addresses to Endpoint

trustin:
Motivation:

Endpoint internally knows:

- whether it is hostname-only endpoint or IP-address-only endpoint
- whether its IP address is IPv4 or IPv6

However, we do not provide any getters for such information.

Modifications:

- Add Endpoint.hasIpAddr()
- Add Endpoint.isIpAddrOnly()
- Add Endpoint.ipFamily()
- .. and use the new methods where applicable

Result:

- Efficiency


trustin:
Thanks for reviewing.

PR #1235, opened 2018-05-31
Fill all necessary pseudo headers in RequestLog headers / Deprecate RequestLog.host() in favor of authority()

trustin:
Motivation:

It is currently possible that `RequestLog.requestHeaders()` and
`responseHeaders()` return an empty `HttpHeaders`, which can trigger
various issues because we often assume they have at least essential
HTTP headers such as `:scheme`, `:authority`, `:method`, `:path` and
`:status`.

Modifications:

- Send a `400 Bad Request` response if a request does not have the
  `:method` header.
- Do not fail with an exception if a client sent a weird method name
  which is not recognized by `HttpMethod.valueOf()`.
  - Add `HttpMethod.UNKNOWN` to represent an 'unknown' method
  - Add `HttpStatus.UNKNOWN` to represent an 'unknown' status
- Auto-fill `:scheme` header if missing.
- Return a dummy `HttpHeaders` with the essential headers are filled
  instead of returning an empty `HttpHeaders`.
- Deprecate `RequestLog.host()` in favor of `RequestLog.authority()`

Result:

- Robustness

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpMethod.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpMethod.java
@@ -40,7 +40,7 @@ public enum HttpMethod {
     // Forked from Netty at 93b144b7b4872ea623a378c75b92d33bea28ab51
 
     /**
-     * The OPTIONS method represents a request for information about the communication options
+     * The OPTIONS method which represents a request for information about the communication options
      * available on the request/response chain identified by the Request-URI. This method allows
      * the client to determine the options and/or requirements associated with a resource, or the
      * capabilities of a server, without implying a resource action or initiating a resource
@@ -49,7 +49,7 @@ public enum HttpMethod {
     OPTIONS,
 
     /**
-     * The GET method means retrieve whatever information (in the form of an entity) is identified
+     * The GET method which means retrieve whatever information (in the form of an entity) is identified
      * by the Request-URI.  If the Request-URI refers to a data-producing process, it is the
      * produced data which shall be returned as the entity in the response and not the source text
      * of the process, unless that text happens to be the output of the process.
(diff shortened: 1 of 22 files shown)


core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -325,7 +331,12 @@ private void startRequest0(long requestStartTimeNanos, long requestStartTimeMill
         this.requestStartTimeMillis = requestStartTimeMillis;
         this.channel = channel;
         this.sessionProtocol = sessionProtocol;
-        this.host = host;
+        if (sessionProtocol.isTls()) {
+            // Switch to the dummy headers with ':scheme=https' if the connection is TLS.
+            if (requestHeaders == DUMMY_REQUEST_HEADERS_HTTP) {
+                requestHeaders = DUMMY_REQUEST_HEADERS_HTTPS;
```
huydx: nits: wonder what's better, fix the header by us or return to client that they're just sending malformed header?

huydx:
still LGTM

core/src/main/java/com/linecorp/armeria/common/HttpMethod.java
```diff
@@ -49,53 +49,59 @@
...
     /**
-     * The DELETE method requests that the origin server delete the resource identified by the
+     * The DELETE method which requests that the origin server delete the resource identified by the
      * Request-URI.
      */
     DELETE,
 
     /**
-     * The TRACE method is used to invoke a remote, application-layer loop- back of the request
+     * The TRACE method which is used to invoke a remote, application-layer loop- back of the request
```
hyangtack: `loop- back` -> `loop-back`?

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -439,7 +438,16 @@ private static String handleThrowable(@Nullable Throwable cause) {
                 case "scheme":
                     return RequestLog::scheme;
                 case "host":
-                    return RequestLog::host;
+                    return log -> {
+                        final String authority = log.responseHeaders().authority();
+                        if (!"?".equals(authority)) {
```
hyangtack: nit: how about flipping the condition?
```
if ("?".equals(authority)) {
  ...
} else {
  ...
}
```

core/src/main/java/com/linecorp/armeria/common/HttpMethod.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -292,6 +294,7 @@ public void setMaxRequestLength(long maxRequestLength) {
 
     @Override
     public HttpHeaders additionalResponseHeaders() {
+        final HttpHeaders additionalResponseHeaders = this.additionalResponseHeaders;
```
minwoox: Oops Thank you!

core/src/test/java/com/linecorp/armeria/common/logging/DefaultRequestLogTest.java
```diff
@@ -97,14 +119,13 @@ public void rpcFailure_endResponseDifferentCause() {
     }
 
     @Test
-    public void addChild() {
+    public void addChild() throws Exception {
```
minwoox: nit: can remove throws?

trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
trustin: :blush:

trustin:
Let me merge once CI passes.

trustin:
Thanks for reviewing!

PR #1237, opened 2018-06-01
Mitigate to return CompletionStage instead of CompletableFuture where…

minwoox:
… applicable

Motivation:
Consider following example:
```
new CircuitBreakerStrategy<RpcResponse>() {
    @Override
    public CompletableFuture<Boolean> shouldReportAsSuccess(RpcResponse response) {
        return response.thenApply(...)
                       .exceptionally(...)
                       .toCompletableFuture();
    }
};
```
The user who implements his or her own strategy has to put `toCompletableFuture` at the end. But
we use the methods of `CompletionStage` in returned codes, so the user doe not have to convert it.
So we can save one more method call.

Modification:
- Mitigate to return `CompletionStage` where applicable mostly in stategy classes

Result:
- Save the world by one more step

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClient.java
@@ -19,7 +19,7 @@ package com.linecorp.armeria.client.circuitbreaker;
 import static com.linecorp.armeria.common.util.Functions.voidFunction;
 import static java.util.Objects.requireNonNull;
 
-import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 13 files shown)


minwoox:
Thanks for reviewing!!

imasahiro:
@minwoox Sorry for late review. Realized during upgrading our services that, with this change, yeah, a user who write a `CircuitBreakerStrategy` don't need to call `toCompletableFuture` anymore. But this introduces another `toCompletableFuture` to a user who uses `CircuitBreaker` as a library. It looks like a open question but do you think moving to `CircuitBreaker` user side is more effective?

minwoox:
@imasahiro That was my worry as well when I made this PR :) because this makes users to call `toCompletableFuture()` if they need it when they don't use our `CircuitBreakClient`.
But I thought that it would be better to return `CompletionStage`.

If we return `CompletableFuture`, we have to call `toCompletableFuture()` in every situation.
If we return `CompletionStage`, `toCompletableFuture()` is only called when it's really needed (in user-side).
Besides, it encourages people to use non-blocking mechanism because they have to call `toCompletableFuture()` if they want to use `get()` or `join()`.

Issue #1185, opened 2018-05-08
Armeria performance, slower than OKHTTP?


rogerbosman:
Hi,

I have run tests to verify armeria's performance.
For this, I have two scenario's written in RxKotlin. One uses Square's "regular" retrofit, and one uses Armeria's. Both use the RxJavaCallAdapterFactory. The only change for the scenario's is what Retrofit instance is loaded. If Retrofit's performance is 100%, I am measuring 110% for Armeria. Is this right? Are we sacrificing response time for total capacity?

Here are my two Retrofit configs:

`  Retrofit.Builder().baseUrl(baseUrl).addCallAdapterFactory(RxJavaCallAdapterFactory.create()).addConverterFactory(GsonConverterFactory.create(gson)).build()
`

`ArmeriaRetrofitBuilder().baseUrl(baseUrl).addCallAdapterFactory(RxJavaCallAdapterFactory.create()).addConverterFactory(GsonConverterFactory.create(gson)).build()`

trustin:
There's some theoretical limit on what we can achieve with Retrofit integration because Armeria has to convert OkHttp classes into Armeria classes and vice versa, whereas the builtin implementation doesn't have to perform any conversion.

However, we are very interested in minimizing the conversion overhead and in achieving higher throughput by scaling better.

trustin:
Let me leave this issue open until we have a better picture about the performance of our Retrofit integration.

PR #1238, opened 2018-06-02
Add a simple benchmark to check retrofit performance.

anuraaga:
For #1185 - it's good to have some benchmarks to establish a baseline. This first one doesn't intend to be representative of all use cases, but has minimal business logic to try to stress the integration itself as much as possible. Indeed seeing about a 10% improvement in performance using upstream okhttp, similar to the report.

Edit: Was stupidly creating new client every benchmark iteration, ignore emails.

```
Benchmark                                            Mode  Cnt     Score     Error  Units
c.l.a.r.downstream.DownstreamSimpleBenchmark.empty  thrpt   20  3406.076 ▒ 105.094  ops/s
c.l.a.r.upstream.UpstreamSimpleBenchmark.empty      thrpt   20  3785.867 ▒ 130.965  ops/s
```

By the way, to not feel too down, here are the results with 20 JMH threads - as expected our throughput is way better.

```
Benchmark                                            Mode  Cnt      Score     Error  Units
c.l.a.r.downstream.DownstreamSimpleBenchmark.empty  thrpt   20  12852.210 ▒ 261.126  ops/s
c.l.a.r.upstream.UpstreamSimpleBenchmark.empty      thrpt   20   7064.704 ▒ 112.538  ops/s
```

### DIFF
--- a/benchmarks/build.gradle
+++ b/benchmarks/build.gradle
@@ -24,7 +24,11 @@ dependencies {
 
 dependencies {
     compile project(':grpc')
+    compile project(':retrofit2')
     compile project(':thrift')
+
+    compile 'com.squareup.retrofit2:adapter-java8'
+    compile 'com.squareup.retrofit2:converter-jackson'
 
     if (grpcUpstream) {
         // Since project(':core') is automatically added in the top-level build, it's difficult to exclude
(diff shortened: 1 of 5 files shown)


anuraaga:
JFYI trying on Windows Subsystem for Linux, had the same behavior of connections never being reused. Will be able to check real Linux on Monday

trustin:
Oops. Could you fix the conflicts, @anuraaga?

anuraaga:
Yup, merged

trustin:
Thank you!

PR #1239, opened 2018-06-02
Use gRPC shaded in benchmarks to avoid needing classpath workarounds …

anuraaga:
…in build script. For #799 

Guava version conflict also seems to be fixed. Being able to run both benchmarks at the same time is very convenient.

We're still slow, hmm :/ 

```
Benchmark                                           (clientType)   Mode  Cnt     Score     Error  Units
c.l.a.g.downstream.DownstreamSimpleBenchmark.empty        OKHTTP  thrpt   20  3859.725 ± 182.180  ops/s
c.l.a.g.upstream.UpstreamSimpleBenchmark.empty            OKHTTP  thrpt   20  5577.127 ± 147.109  ops/s
```

### DIFF
--- a/benchmarks/build.gradle
+++ b/benchmarks/build.gradle
@@ -11,26 +11,14 @@ apply plugin: 'me.champeau.gradle.jmh'
 
 def jmhInclude = rootProject.findProperty('jmh.include') ?: ''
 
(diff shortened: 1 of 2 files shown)


trustin:
Thank you, @anuraaga!

PR #1240, opened 2018-06-02
Only enable verbose exceptions for tests.

anuraaga:
Currently, verbose exceptions are enabled for all Java executions. This even applies to manual execution with the Gradle runner in IntelliJ, and when doing profiling, it causes the profiles to have extra noise. Now only test have verbose exceptions enabled, which is probably where they are useful.

Also fix `leak` property should be read from `rootProject` as with all gradle properties AFAIU. I recall us not getting leak traces at some point maybe related to this.


trustin:
Thanks, @anuraaga. :-)

PR #1242, opened 2018-06-03
Fix a bug where RetryingClient does not stop when StreamMessage is ab…

minwoox:
…orted

Motivation:
If the `Client` aborts the `Request` or `Response`, `RetryingClient` has to
stop retrying.

Modifications:
- Track the `Request` from the `Client` and `Respone` which handed over to the `Client` to check if it's aborted

Result:
- Fixed another bug

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
@@ -42,6 +42,7 @@ import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpRequestDuplicator;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpResponseDuplicator;
+import com.linecorp.armeria.common.stream.AbortedStreamException;
 import com.linecorp.armeria.internal.HttpHeaderSubscriber;
 
 /**
@@ -112,16 +113,21 @@ public final class RetryingHttpClient extends RetryingClient<HttpRequest, HttpRe
         final CompletableFuture<HttpResponse> responseFuture = new CompletableFuture<>();
         final HttpResponse res = HttpResponse.from(responseFuture);
         final HttpRequestDuplicator reqDuplicator = new HttpRequestDuplicator(req, 0, ctx.eventLoop());
-        doExecute0(ctx, reqDuplicator, responseFuture);
+        doExecute0(ctx, reqDuplicator, req, res, responseFuture);
         return res;
     }
 
     private void doExecute0(ClientRequestContext ctx, HttpRequestDuplicator rootReqDuplicator,
-                            CompletableFuture<HttpResponse> res) {
+                            HttpRequest originalReq, HttpResponse returnedRes,
+                            CompletableFuture<HttpResponse> future) {
+        if (originalReq.completionFuture().isCompletedExceptionally() || returnedRes.isComplete()) {
+            // The request or response is aborted by the client before it receives a response, so stop retrying.
+            actionOnException(ctx, rootReqDuplicator, future).accept(AbortedStreamException.get());
+            return;
+        }
+
         if (!setResponseTimeout(ctx)) {
-            onRetryingComplete(ctx);
-            res.completeExceptionally(ResponseTimeoutException.get());
-            rootReqDuplicator.close();
+            actionOnException(ctx, rootReqDuplicator, future).accept(ResponseTimeoutException.get());
             return;
         }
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -146,24 +152,32 @@ private void doExecute0(ClientRequestContext ctx, HttpRequestDuplicator rootReqD
...
+                                                                  returnedRes, future),
+                                                 nextDelay);
                            } else {
-                               finishRetryWithCurrentResponse(ctx, rootReqDuplicator, res, resDuplicator);
+                               finishRetryWithCurrentResponse(ctx, rootReqDuplicator, future, resDuplicator);
                            }
                        });
     }
 
+    private static Consumer<? super Throwable> actionOnException(ClientRequestContext ctx,
```
trustin: Can just return `Consumer<Throwable>` in this case.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -146,24 +152,32 @@ private void doExecute0(ClientRequestContext ctx, HttpRequestDuplicator rootReqD
...
+                                                                 HttpRequestDuplicator rootReqDuplicator,
+                                                                 CompletableFuture<HttpResponse> future) {
+        return cause -> {
+            onRetryingComplete(ctx);
+            future.completeExceptionally(cause);
+            rootReqDuplicator.close();
```
trustin: How about extracting this into a method like `handleException()` so that we don't need to do `actionOnException().accept()`?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
```diff
@@ -118,6 +120,14 @@ private void doExecute0(ClientRequestContext ctx, RpcRequest req, CompletableFut
         });
     }
 
+    private static Consumer<? super Throwable> actionOnException(ClientRequestContext ctx,
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
```diff
@@ -118,6 +120,14 @@ private void doExecute0(ClientRequestContext ctx, RpcRequest req, CompletableFut
...
+                                                                 CompletableFuture<RpcResponse> future) {
+        return cause -> {
+            onRetryingComplete(ctx);
+            future.completeExceptionally(cause);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
```diff
@@ -60,9 +67,20 @@
 public class RetryingHttpClientTest {
 
     // use different eventLoop from server's so that clients don't hang when the eventLoop in server hangs
-    private static ClientFactory clientFactory = new ClientFactoryBuilder()
+    private static final ClientFactory clientFactory = new ClientFactoryBuilder()
             .workerGroup(EventLoopGroups.newEventLoopGroup(2), true).build();
 
+    private static final RetryStrategy<HttpRequest, HttpResponse> retryAlways =
```
trustin: ~alwaysRetry?~

core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
```diff
@@ -357,6 +407,59 @@ public void shouldGetExceptionWhenFactoryIsClosed() {
...
+                    public void onNext(HttpObject httpObject) {
+                    }
+
+                    @Override
+                    public void onError(Throwable t) {
+                    }
+
+                    @Override
+                    public void onComplete() {
+                    }
```
trustin: Could these be `{}` for fewer lines?

core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
```diff
@@ -357,6 +407,59 @@ public void shouldGetExceptionWhenFactoryIsClosed() {
...
+
+        await().untilAsserted(() -> assertThat(subscriberCancelServiceCallCounter.get()).isEqualTo(3));
+    }
+
+    @Test
+    public void doNotRetryWhenRequestIsAborted() throws Exception {
+        final HttpClient client = client(retryAlways);
+
+        final HttpRequestWriter req = HttpRequest.streaming(
+                HttpHeaders.of(HttpMethod.GET, "/request-abort"));
```
trustin: Could be just: `HttpRequest.streaming(HttpMEthod.GET, "/request-abort")`

core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
```diff
@@ -357,6 +407,59 @@ public void shouldGetExceptionWhenFactoryIsClosed() {
...
+        req.write(HttpData.ofUtf8("I'm going to abort this request"));
+        req.abort();
+        client.execute(req);
+
+        TimeUnit.SECONDS.sleep(1);
+        // No request is made.
+        assertThat(responseAbortServiceCallCounter.get()).isZero();
```
trustin: You don't need to call `get()` for assertion. assertj knows how to deal with `Atomic*`

minwoox: All fixed! Thank you!

minwoox:
Thanks for reviewing!

PR #1246, opened 2018-06-05
Only compute advertised encodings header once since it doesn't change.

anuraaga:
The advertised encodings of a server or client are constant throughout multiple calls, so we don't need to do slightly expensive computation of the header string every time.

Just a micro-optimization

After (only 1 fork isn't enough for a reliable measurement)
```
Benchmark                                           (clientType)   Mode  Cnt      Score     Error  Units
c.l.a.g.downstream.DownstreamSimpleBenchmark.empty        OKHTTP  thrpt   20  17927.598 ± 670.447  ops/s
c.l.a.g.upstream.UpstreamSimpleBenchmark.empty            OKHTTP  thrpt   20  22614.173 ± 705.504  ops/s
```

Before
```
Benchmark                                           (clientType)   Mode  Cnt      Score      Error  Units
c.l.a.g.downstream.DownstreamSimpleBenchmark.empty        OKHTTP  thrpt   20  17344.866 ±  568.339  ops/s
c.l.a.g.upstream.UpstreamSimpleBenchmark.empty            OKHTTP  thrpt   20  21301.553 ± 1204.546  ops/s
```

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaChannel.java
+++ b/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaChannel.java
@@ -72,6 +72,7 @@ class ArmeriaChannel extends Channel implements ClientBuilderParams {
     private final SerializationFormat serializationFormat;
     @Nullable
     private final MessageMarshaller jsonMarshaller;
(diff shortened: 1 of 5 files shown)


trustin:
Thanks, @anuraaga ! :+1:

PR #1247, opened 2018-06-07
Reject most duplicate path mappings

### LINKED ISSUE #529: Show/Throw Error if register Service to the same path
When a user register multiple service to the same path through `ServerBuilder#serviceAt`, armeria creates a `Server` with services but PathMappings only map a request to a service that is registered first.
```java
    @ClassRule
    public static final ServerRule server = new ServerRule() {
        @Override
            sb.serviceAt("/api", new EchoService("1"))
              .serviceAt("/api", new EchoService("2"))
              .serviceAt("/api", new EchoService("3"));
        }
    };
    @Test
    public void test1() throws Exception {
        try (CloseableHttpClient hc = HttpClients.createMinimal()) {
            final HttpGet req = new HttpGet(server.uri("/api"));
            try (CloseableHttpResponse res = hc.execute(req)) {
                assertThat(EntityUtils.toString(res.getEntity()), is("1"));
            }
        }
    }
    private static class EchoService extends AbstractHttpService {
        private final String s;

        public EchoService(String s) {
            this.s = s;
        }

        @Override
        protected final void doGet(ServiceRequestContext ctx, HttpRequest req, HttpResponseWriter res) {
            res.write(HttpHeaders.of(HttpStatus.OK));
            res.write(HttpData.of(s.getBytes()));
            res.close();
        }
    }
```

It would be better to show message (or throw Exception) when ServerBuilder detects same path.

trustin:
Motivation:

When a user binds more than one service at the same path mapping,
`ServerBuilder` should reject such a configuration.

Modifications:

- Add duplicate mapping detection to `Routers.routers()`
- Miscellaneous:
  - Weaken `PathMapping` types in `AnnotatedHttpServiceFactory`

Result:

- Almost fixes #529
- Note that this changeset does not detect all possible cases. For
  example, it does not consider the relationship between media types.
  It will detect most silly mistakes though, which should be enough.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Routers.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Routers.java
@@ -26,7 +26,10 @@ import java.io.PrintWriter;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 
 import javax.annotation.Nullable;
@@ -89,6 +92,8 @@ public final class Routers {
      */
     @VisibleForTesting
     static <V> List<Router<V>> routers(Iterable<V> values, Function<V, PathMapping> pathMappingResolver) {
+        rejectDuplicateMapping(values, pathMappingResolver);
+
         final ImmutableList.Builder<Router<V>> builder = ImmutableList.builder();
         final List<V> group = new ArrayList<>();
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/linecorp/armeria/server/Routers.java
```diff
@@ -116,6 +121,75 @@
...
+                if (!headerMapping.consumeTypes().isEmpty() &&
+                    headerMapping.consumeTypes().stream().noneMatch(
+                            mediaType -> existingHeaderMapping.consumeTypes().contains(mediaType))) {
+                    // No overlap in consume types.
+                    continue;
+                }
+                if (!headerMapping.produceTypes().isEmpty() &&
+                    headerMapping.produceTypes().stream().noneMatch(
+                            mediaType -> existingHeaderMapping.produceTypes().contains(mediaType))) {
+                    // No overlap in consume types.
```
hyangtack: consume -> produce

trustin:
Thank you, reviewers!

PR #1248, opened 2018-06-08
Optimize complexity of client endpoint WeightedRoundRobinStrategy

linxGnu:
Motivation:

Client using EndpointGroup with current WeightedRoundRobinStrategy is slow. 
* Current algorithm complexity for selectEndpoint operation is O(N * MaxWeight) where N is number of endpoints. Arbitrary weight value (>1000 for example) cause it slow. 

Modifications:
- Sort endpoints by weight, then by host and port.
- Find the correct endpoint faster by binary search on accumulated sum of weight.
- The complexity is O(log(N)) and not depend on weight value.

Algorithm details:
- In comment code block

Result:
* Faster selectEndpoint operation.

Benchmarks:

Before
```
Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.randomMainly1Max30":
  392327.218 ±(99.9%) 9226.013 ops/s [Average]

Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.randomMax10":
  289799.381 ±(99.9%) 3761.818 ops/s [Average]

Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.randomMax100":
  75949.628 ±(99.9%) 1970.120 ops/s [Average]

Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.sameWeight":
  54412957.573 ±(99.9%) 1253690.454 ops/s [Average]

Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.unique":
  21152.272 ±(99.9%) 11486.626 ops/s [Average]
```

After
```
Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.randomMainly1Max30":
  28586820.346 ±(99.9%) 145391.823 ops/s [Average]

Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.randomMax10":
  25691640.444 ±(99.9%) 574921.960 ops/s [Average]

Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.randomMax100":
  24246013.498 ±(99.9%) 994566.750 ops/s [Average]

Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.sameWeight":
  54149030.441 ±(99.9%) 1580181.334 ops/s [Average]

Result "com.linecorp.armeria.core.client.endpoint.WeightedRoundRobinStrategyBenchmark.unique":
  21813513.256 ±(99.9%) 1958592.548 ops/s [Average]
```

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/WeightedRoundRobinStrategyBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/WeightedRoundRobinStrategyBenchmark.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 LINE Corporation
+ * Copyright 2018 LINE Corporation
  *
  * LINE Corporation licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/WeightedRoundRobinStrategyBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/WeightedRoundRobinStrategyBenchmark.java
@@ -36,20 +36,20 @@ public class WeightedRoundRobinStrategyBenchmark {
 
     final int numEndpoints = 500;
 
-    // Case 1: normal round robin, all weight: 3
-    EndpointSelector selector1;
+    // normal round robin, all weight: 3
+    EndpointSelector selectorSameWeight;
 
-    // Case 3: mainly weight: 1, max weight: 30
-    EndpointSelector selector2;
+    // mainly weight: 1, max weight: 30
+    EndpointSelector selectorRandomMainly1Max30;
 
-    // Case 4: randomly, max weight: 10
-    EndpointSelector selector4;
+    // randomly, max weight: 10
+    EndpointSelector selectorRandomMax10;
 
-    // Case 5: randomly, max weight: 100
-    EndpointSelector selector5;
+    // randomly, max weight: 100
+    EndpointSelector selectorRandomMax100;
 
-    // Case 6: all weights are unique
-    EndpointSelector selector6;
+    // all weights are unique
+    EndpointSelector selectorUnique;
 
     interface EndpointGenerator {
         Endpoint generate(int id);
@@ -74,49 +74,55 @@ public class WeightedRoundRobinStrategyBenchmark {
     public void setupCases() {
         Random rand = new Random();
 
-        selector1 = getEndpointSelector(generateEndpoints(id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
-                3
-        )), "group1");
+        selectorSameWeight = getEndpointSelector(generateEndpoints(
+                id -> Endpoint.of("127.0.0.1", id + 1)
+                        .withWeight(
+                                30
+                        )), "same-weight");
 
-        selector2 = getEndpointSelector(generateEndpoints(id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
-                1 + (id % 50 == 0 ? 29 : 0)
-        )), "group3");
+        selectorRandomMainly1Max30 = getEndpointSelector(generateEndpoints(
+                id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
+                        1 + (id % 50 == 0 ? 29 : 0)
+                )), "main-1-max-30");
 
-        selector4 = getEndpointSelector(generateEndpoints(id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
-                1 + rand.nextInt(10)
-        )), "group4");
+        selectorRandomMax10 = getEndpointSelector(generateEndpoints(
+                id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
+                        1 + rand.nextInt(10)
+                )), "random-max-10");
 
-        selector5 = getEndpointSelector(generateEndpoints(id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
-                1 + rand.nextInt(100)
-        )), "group5");
+        selectorRandomMax100 = getEndpointSelector(generateEndpoints(
+                id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
+                        1 + rand.nextInt(100)
+                )), "random-max-100");
 
-        selector6 = getEndpointSelector(generateEndpoints(id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
-                id + 1
-        )), "group6");
+        selectorUnique = getEndpointSelector(generateEndpoints(
+                id -> Endpoint.of("127.0.0.1", id + 1).withWeight(
+                        id + 1
+                )), "unique");
     }
 
     @Benchmark
     public Endpoint sameWeight() throws Exception {
-        return selector1.select(null);
+        return selectorSameWeight.select(null);
     }
 
     @Benchmark
     public Endpoint randomMainly1Max30() throws Exception {
-        return selector2.select(null);
+        return selectorRandomMainly1Max30.select(null);
     }
 
     @Benchmark
     public Endpoint randomMax10() throws Exception {
-        return selector4.select(null);
+        return selectorRandomMax10.select(null);
     }
 
     @Benchmark
     public Endpoint randomMax100() throws Exception {
-        return selector5.select(null);
+        return selectorRandomMax100.select(null);
     }
 
     @Benchmark
     public Endpoint unique() throws Exception {
-        return selector6.select(null);
+        return selectorUnique.select(null);
     }
 }
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/WeightedRoundRobinStrategyBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/WeightedRoundRobinStrategyBenchmark.java
@@ -36,7 +36,7 @@ public class WeightedRoundRobinStrategyBenchmark {
 
     final int numEndpoints = 500;
 
-    // normal round robin, all weight: 3
+    // normal round robin, all weight: 300
     EndpointSelector selectorSameWeight;
 
     // mainly weight: 1, max weight: 30
@@ -77,7 +77,7 @@ public class WeightedRoundRobinStrategyBenchmark {
         selectorSameWeight = getEndpointSelector(generateEndpoints(
                 id -> Endpoint.of("127.0.0.1", id + 1)
                         .withWeight(
-                                30
+                                300
                         )), "same-weight");
 
         selectorRandomMainly1Max30 = getEndpointSelector(generateEndpoints(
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -71,49 +74,109 @@ public Endpoint select(ClientRequestContext ctx) {
...
+                        .filter(endpoint -> endpoint.weight() > 0) // only process endpoint with weight > 0
+                        .sorted(Comparator
+                                .comparing(Endpoint::weight)
+                                .thenComparing(Endpoint::host)
+                                .thenComparingInt(Endpoint::port))
+                        .collect(Collectors.toList());
+                int numEndpoints = endps.size();
+
+                // accumulation
+                LinkedList<EndpointsGroupByWeight> accumulatedGroups = new LinkedList<>();
```
imasahiro: Can we replace first `LinkedList` to `List`? And can you use `new ArrayList` instead?

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -71,49 +74,109 @@ public Endpoint select(ClientRequestContext ctx) {
...
+                        currentGroup = new EndpointsGroupByWeight(numEndpoints - rest,
+                                endpoint.weight(), totalWeight);
+                        accumulatedGroups.addLast(currentGroup);
+                    }
+
+                    rest--;
+                }
+
+                this.endpoints = endps;
+                this.endpointsGroupByWeight = accumulatedGroups.toArray(
```
imasahiro: You don't need to convert to raw array if you use `ArrayList`, I think.

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -71,49 +74,109 @@ public Endpoint select(ClientRequestContext ctx) {
         private static final class EndpointsAndWeights {
             private final List<Endpoint> endpoints;
             private final boolean weighted;
-            private final int maxWeight;
-            private final int totalWeight;
+            private final long totalWeight; // prevent overflow by using long
+
+            private static final class EndpointsGroupByWeight {
```
imasahiro: Can you move this class out of `EndpointsAndWeights` for less indenting?

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -71,49 +74,109 @@ public Endpoint select(ClientRequestContext ctx) {
...
+                int startIndex;
```
imasahiro: Think you can add `final` to fields.

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
linxGnu: Thank you very much. I would change to use List interface.

However I think I would keep using LinkedList to prevent grow() (the nature of ArrayList does grow operation if reach capacity, which take Array.copy() operation)

trustin: Perhaps you could specify a better initial capacity, so it doesn't expand too often (or even never)?

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
linxGnu: I think ArrayList is much more better in term of abstraction and grow. 

However in this PR, I just want to make this micro optimization as good as possible, not calling additional function to archive element.

I think I would count number of distinct weight and do preallocation.

anuraaga: I think using an `ArrayList` will generally be faster than traversing a `LinkedList` that happens here. It seems reasonable to just allocate `new ArrayList<(endps.size())` and call `ImmutableList.copyOf` here.

Also, since this should only be called when endpoints change, which isn't often, I don't think we need to look at overoptimizations in this constructor.

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -71,49 +74,109 @@ public Endpoint select(ClientRequestContext ctx) {
...
+                                .comparing(Endpoint::weight)
+                                .thenComparing(Endpoint::host)
+                                .thenComparingInt(Endpoint::port))
+                        .collect(Collectors.toList());
+                int numEndpoints = endps.size();
+
+                // accumulation
+                LinkedList<EndpointsGroupByWeight> accumulatedGroups = new LinkedList<>();
+                EndpointsGroupByWeight currentGroup = null;
+                int rest = numEndpoints;
```
anuraaga: `long rest`?

linxGnu: Number of Endpoints is int (in signature of size() of List too). So I think int is ok.

anuraaga: Yeah but you only use it by casting to `long` - if you use `long` here, I think you can remove all the casts.

linxGnu: Yeah, and I need to change type of `startIndex` to `long` too to remove other casts.

linxGnu:
Thank you very much @imasahiro @anuraaga @trustin 
I have updated source code and added JMH as your suggestion. Benchmark is also included in comment.
Please take a review.

benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/WeightedRoundRobinStrategyBenchmark.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright 2016 LINE Corporation
```
huydx: 2018

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -68,26 +71,112 @@ public Endpoint select(ClientRequestContext ctx) {
...
             EndpointsAndWeights(Iterable<Endpoint> endpoints) {
+                long totalWeight = 0;
                 int minWeight = Integer.MAX_VALUE;
                 int maxWeight = Integer.MIN_VALUE;
-                int totalWeight = 0;
-                for (Endpoint endpoint : endpoints) {
+                int numberDistinctWeight = 0;
+
+                // prepare endpoints by copying
+                this.endpoints = ImmutableList.copyOf(endpoints)
```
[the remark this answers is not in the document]
linxGnu: I am sorry, I didn't notice it :(

And do you mean (I couldn't find Streams.of from Guava and JDK)
```
this.endpoints = Streams
                        .stream(endpoints)
                        .filter(e -> e.weight() > 0)
                        .sorted(Comparator
                                .comparing(Endpoint::weight)
                                .thenComparing(Endpoint::host)
                                .thenComparingInt(Endpoint::port))
                        .collect(ImmutableList.toImmutableList());
```

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -68,26 +71,112 @@ public Endpoint select(ClientRequestContext ctx) {
...
+                // accumulation
+                EndpointsGroupByWeight[] accumulatedGroups = new EndpointsGroupByWeight[numberDistinctWeight];
+                EndpointsGroupByWeight currentGroup = null;
+                int index = -1;
+
+                long rest = numEndpoints;
+                for (Endpoint endpoint : this.endpoints) {
+                    if (currentGroup == null || currentGroup.weight != endpoint.weight()) {
+                        totalWeight += currentGroup == null ?
+                                (long) endpoint.weight() * rest
```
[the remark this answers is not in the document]
linxGnu: Thank you, I would remove it.

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -68,26 +71,112 @@ public Endpoint select(ClientRequestContext ctx) {
...
+                                : (long) (endpoint.weight() - currentGroup.weight) * rest;
+                        currentGroup = new EndpointsGroupByWeight(
+                                numEndpoints - rest, endpoint.weight(), totalWeight
+                        );
+                        accumulatedGroups[++index] = currentGroup;
```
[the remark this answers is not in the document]
linxGnu: In this case, array give better performance. However, to fully guarantee immutability, I would change it to `ImmutableList` using `ImmutableList.Builder` as your suggestion.

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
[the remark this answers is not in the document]
trustin: Looks good to me. nits:

- No line break between `Streams` and `.stream(...)`
- No line break between `Comparator` and `.comparing()`
. Static-import `ImmutableList.toImmutableList`.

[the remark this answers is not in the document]
anuraaga: I did indeed mean `Streams.stream`, sorry for the confusion!

core/src/main/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategy.java
```diff
@@ -68,26 +72,112 @@ public Endpoint select(ClientRequestContext ctx) {
...
             EndpointsAndWeights(Iterable<Endpoint> endpoints) {
+
+                // prepare immutable endpoints
+                this.endpoints = Streams.stream(endpoints)
+                        .filter(e -> e.weight() > 0) // only process endpoint with weight > 0
+                        .sorted(Comparator.comparing(Endpoint::weight)
+                                .thenComparing(Endpoint::host)
+                                .thenComparingInt(Endpoint::port))
+                        .collect(toImmutableList());
+                long numEndpoints = this.endpoints.size();
```
hyangtack: nit: would you add `final` keyword to the variable which holds the value not to be changed like here?

linxGnu: I would change and commit soon. Thank you very much.

trustin:
Thanks you, @linxGnu and reviewers. :smile:

PR #1249, opened 2018-06-10
Refactor TomcatService into ManagedTomcatService & UnmanagedTomcatSer…

jmostella:
…vice, for compatibility with SpringBoot.

Issues resolved:

1. 2 Tomcat servers starting. (Spring & Tomcat9)
2. NPE due to Spring removing connector from tomcat instance. (Spring & Tomcat8.5)
3. Avoid unnecessary management of Tomcat server which is already being managed. (SpringBoot)

* ArmeriaAutoConfiguration added error logging for server startup failure.
* ArmeriaSettings fixed documented yaml.
* it/spring/boot-tomcat/ SpringBoot app which reproduces the issue for Tomcat 9.0.
* it/spring/boot-tomcat8.5/ SpringBoot app which reproduces the issue for Tomcat 8.5.
* it/server/ moved from it/
* ManagedTomcatService added.
* UnmanagedTomcatService added.
* TomcatService functionality moved to subclasses.
* TomcatVersion made public and VisibleForTesting added. Moved to com.linecorp.armeria.internal.tomcat
* TomcatServiceTest renamed to ManagedTomcatServiceTest


jmostella:
## Summary
Spring creates the `Tomcat` starter and uses it to start the tomcat server. Afterward, removes the connectors are removed from the service during startup.

After Spring is fully initialized it then re-associates the connectors, to the service.

### Tomcat9
A new connector is created during the call to `Tomcat.getConnector()`.
Resulting in  2 tomcat connectors. If both are using the default 8080 port, then a port collision occurs.

_refer_ [Tomcat9](https://github.com/apache/tomcat/blob/trunk/java/org/apache/catalina/startup/Tomcat.java#L443)

### Tomcat 8.5
A `null` is return from `Tomcat.getConnector()`. This will throw an error on startup, in the current code.

_refer_ [Tomcat85](https://github.com/apache/tomcat85/blob/trunk/java/org/apache/catalina/startup/Tomcat.java#L438)

As a follow up PR I will add the `ArmeriaServerConfigurator` to `boot-autoconfigure`.
Also I will fix async (Servlet 3.0) requests.

spring/boot-tomcat-sample/src/main/resources/config/application.yml
```diff
@@ -0,0 +1,6 @@
+armeria:
+  ports:
+    - port: 8081
+      protocol: HTTP
+server:
+  port: 8080
```
imasahiro: Please add missing newline

settings.gradle
```diff
@@ -30,6 +30,7 @@ includeWithFlags ':zookeeper',                  'java', 'publish', 'relocate'
 includeWithFlags ':benchmarks',       'java'
 includeWithFlags ':it',               'java', 'relocate'
 includeWithFlags ':testing-internal', 'java', 'relocate'
+includeWithFlags ':spring:boot-tomcat-sample', 'java', 'relocate'
```
trustin: Could align the flags of other unpublished Java projects?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
```diff
@@ -237,7 +238,12 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req)
 
         final Server s = server.build();
 
-        s.start().join();
+        s.start().handle((result, throwable) -> {
+            if (!Objects.isNull(throwable)) {
+                logger.error("Armeria server startup error: ", throwable);
```
trustin: `"Armeria server failed to start:", throwable` ?

build.gradle
```diff
@@ -5,6 +5,7 @@ buildscript {
     dependencies {
         classpath 'com.google.gradle:osdetector-gradle-plugin:1.6.0'
         classpath 'io.spring.gradle:dependency-management-plugin:1.0.5.RELEASE'
+        classpath 'org.springframework.boot:spring-boot-gradle-plugin:2.0.2.RELEASE'
```
trustin: How about moving this into `spring/boot-tomcat-sample/build.gradle`?

spring/boot-tomcat-sample/src/main/java/com/linecorp/armeria/spring/tomcat/demo/SpringTomcatApplication.java
```diff
@@ -0,0 +1,63 @@
...
+                                                          TomcatService.forTomcat(tomcat));
+        }
+        return serverBuilder -> { };
+    }
+
+    /**
+     * Main method.
+     * @param args program args.
+     */
+    @SuppressWarnings("checkstyle:uncommentedmain")
```
trustin: Could we fix this violation rather than suppressing it?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -175,43 +162,39 @@ public static TomcatService forFileSystem(Path docBase) {
      * Creates a new {@link TomcatService} from an existing {@link Tomcat} instance.
      * If the specified {@link Tomcat} instance is not configured properly, the returned {@link TomcatService}
      * may respond with '503 Service Not Available' error.
+     *
+     * @return tomcat service, will not manage the tomcat.
      */
     public static TomcatService forTomcat(Tomcat tomcat) {
```
trustin: `a new {@link TomcatService}, which will not manage the specified {@link Tomcat}`

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -254,125 +237,21 @@ static String toString(org.apache.catalina.Server server) {
...
-    @Nullable
-    private String hostname;
-    @Nullable
-    private Connector connector;
-    @Nullable
-    private String engineName;
-    private boolean started;
-
-    private TomcatService(@Nullable String hostname, Function<String, Connector> connectorFactory) {
-        this(hostname, connectorFactory, unused -> { /* unused */ });
```
trustin: Could you add a package-local default constructor, so that the constructor will not be shown in our Javadoc?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -254,125 +237,21 @@ static String toString(org.apache.catalina.Server server) {
...
-        }
-
-        try {
-            logger.info("Stopping an embedded Tomcat: {}", toString(server));
-            server.stop();
-        } catch (Exception e) {
-            logger.warn("Failed to stop an embedded Tomcat: {}", toString(server), e);
-        }
+    @Nullable
+    protected abstract Connector connector();
```
trustin: Could be a `public` method for backward compatibility?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -254,125 +237,21 @@ static String toString(org.apache.catalina.Server server) {
...
 
-        postStopTask.accept(connector);
-    }
+    @Nullable
+    protected abstract String hostName();
```
trustin: Could be package local to hide it from Javadoc?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatVersion.java
```diff
@@ -23,7 +23,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-final class TomcatVersion {
+import com.google.common.annotations.VisibleForTesting;
+
+@VisibleForTesting
+public final class TomcatVersion {
```
trustin: Could you move this class to `com.linecorp.armeria.internal.tomcat` so it's not visible from Javadoc?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/UnmanagedTomcatService.java
```diff
@@ -0,0 +1,60 @@
...
+package com.linecorp.armeria.server.tomcat;
+
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+import org.apache.catalina.connector.Connector;
+import org.apache.catalina.startup.Tomcat;
+
+public final class UnmanagedTomcatService extends TomcatService {
```
trustin: Could be package-local

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/ManagedTomcatService.java
```diff
@@ -0,0 +1,161 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Sets;
+
+import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.ServerListener;
+import com.linecorp.armeria.server.ServerListenerAdapter;
+import com.linecorp.armeria.server.ServiceConfig;
+
+public final class ManagedTomcatService extends TomcatService {
```
trustin: Could be package-local.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
```diff
@@ -237,7 +238,12 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req)
 
         final Server s = server.build();
 
-        s.start().join();
+        s.start().handle((result, throwable) -> {
```
anuraaga: We commonly use `t` for the exception in a future handler.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
```diff
@@ -237,7 +238,12 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req)
...
+            if (!Objects.isNull(throwable)) {
```
anuraaga: `t != null`

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
```diff
@@ -237,7 +238,12 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req)
...
+                logger.error("Armeria server startup error: ", throwable);
+            }
+            return result;
```
anuraaga: Doesn't this cause the code to continue, ignoring the error? How about instead of using `logger`, you just do something like `throw new IllegalStateException("Armeria server failed to start", t)` so the logging is added while still causing the exception to fail initialization?

spring/boot-tomcat-sample/src/main/java/com/linecorp/armeria/spring/tomcat/demo/SpringTomcatApplication.java
anuraaga: Also wonder for this type of reason and just avoiding building the example every time we build armeria, should we create a separate `armeria-examples` repo?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/UnmanagedTomcatService.java
```diff
@@ -0,0 +1,60 @@
...
+    @Nullable
+    private Tomcat tomcat;
```
anuraaga: Looks like these can be `final`

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/UnmanagedTomcatService.java
```diff
@@ -0,0 +1,60 @@
...
+    @Nullable
+    private String hostName;
+    @Nullable
+    private Connector connector;
+
+    UnmanagedTomcatService(Tomcat tomcat) {
+        this.tomcat = tomcat;
```
anuraaga: I think this can be `this(tomcat.getEngine().getDefaultHost(), tomcat.getConnector())` and the other methods are just getters.

spring/boot-tomcat-sample/src/main/java/com/linecorp/armeria/spring/tomcat/demo/SpringTomcatApplication.java
jmostella: Yes the main method is needed to launch spring boot. Do you mean adding to `checkstyle-suppressions.xml`

I think a separate repo is a good idea, except that since an example project like this is meant to reproduce issues, if Armeria breaks an example, then the Armeria build should fail also IMO.

anuraaga: Isn't an example to just show how to use the code? If we have integration test cases we could add to the `it` module

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
trustin: Thought we use `cause` more commonly? `t` is fine though.

spring/boot-tomcat-sample/src/main/java/com/linecorp/armeria/spring/tomcat/demo/SpringTomcatApplication.java
jmostella: I was thinking the example could function as both. But I also see your point.

since `it` is a single project, I think we need to refactoring `it` to have several sub-projects

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/UnmanagedTomcatService.java
jmostella: The reason as mention as mentioned in the **Summary**, is because during Spring startup, it removes the connectors from the `Tomcat`, and after startup ends re-associates them.

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/UnmanagedTomcatService.java
jmostella: see following comment

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatVersion.java
jmostella: `com.linecorp.armeria.tomcat.internal`?

trustin: Our convention uses `com.linecorp.armeria.internal.*`. See `armeria-thrift` for an example.

it/spring/boot-tomcat8.5/src/test/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+armeria-tomcat.version.major=8
+armeria-tomcat.version.minor=5
```
trustin: Missing LF at the end

settings.gradle
```diff
@@ -27,9 +27,12 @@ includeWithFlags ':zipkin',                     'java', 'publish', 'relocate'
 includeWithFlags ':zookeeper',                  'java', 'publish', 'relocate'
 
 // Unpublished Java projects
-includeWithFlags ':benchmarks',       'java'
-includeWithFlags ':it',               'java', 'relocate'
-includeWithFlags ':testing-internal', 'java', 'relocate'
+includeWithFlags ':benchmarks',                 'java'
+includeWithFlags ':it:server',                  'java', 'relocate'
+includeWithFlags ':it:spring:boot-tomcat',      'java', 'relocate'
+includeWithFlags ':it:spring:boot-tomcat8.5',   'java', 'relocate'
```
trustin: Thanks a lot for reorganization!

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -254,125 +240,21 @@ static String toString(org.apache.catalina.Server server) {
...
-        }
-
-        try {
-            logger.info("Stopping an embedded Tomcat: {}", toString(server));
-            server.stop();
-        } catch (Exception e) {
-            logger.warn("Failed to stop an embedded Tomcat: {}", toString(server), e);
-        }
+    @Nullable
+    public abstract Connector connector();
```
trustin: The previous behavior of this method was to throw `IllegalStateException` if `connector` is `null`. Would it be possible to keep the old behavior and remove `@Nullable`?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/UnmanagedTomcatService.java
```diff
@@ -0,0 +1,63 @@
...
+    @Nullable
+    private final Connector connector;
+
+    UnmanagedTomcatService(Tomcat tomcat) {
+        this.hostName = null;
+        this.connector = null;
+        this.tomcat = tomcat;
+    }
+
+    UnmanagedTomcatService(String hostName, Connector connector) {
```
trustin: `hostName` could be `@Nullable`?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
jmostella: My impression from https://github.com/line/armeria/pull/1249/files#diff-d20f3c660ecfc12f2baf4a22607a08a2L374
was that we wanted to detect this situation and `throw HttpStatusException.of(HttpStatus.SERVICE_UNAVAILABLE)`?

jmostella: https://github.com/line/armeria/pull/1249/files#diff-d20f3c660ecfc12f2baf4a22607a08a2R259

settings.gradle
jmostella: no problem!

it/spring/boot-tomcat/src/main/java/com/linecorp/armeria/spring/tomcat/demo/SpringTomcatApplication.java
```diff
@@ -0,0 +1,63 @@
...
+                                                          TomcatService.forTomcat(tomcat));
+        }
+        return serverBuilder -> { };
+    }
+
+    /**
+     * Main method.
+     * @param args program args.
+     */
+    @SuppressWarnings("checkstyle:uncommentedmain")
```
jmostella: Will update to suppress everything under `it`

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
trustin: I see. How about then returning an `Optional<Connector>`? It's a breaking change anyway.

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/UnmanagedTomcatService.java
```diff
@@ -0,0 +1,66 @@
...
+    UnmanagedTomcatService(@Nullable String hostName, Connector connector) {
+        this.hostName = hostName;
+        this.connector = connector;
+        this.tomcat = null;
+    }
+
+    @Override
+    public Optional<Connector> connector() {
+        if (connector != null) {
+            return Optional.of(connector);
```
anuraaga: How about storing the `Optional`s as fields instead of Nullable since we need to create `Optional` either way (I guess there's no conceivable usage where we won't create `Optional`)

jmostella: Thats a good idea

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/UnmanagedTomcatService.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public String hostName() {
+        if (hostName != null) {
+            return hostName;
+        }
+
+        // If connector w/o hostName
+        return tomcat.map(Tomcat::getEngine)
```
anuraaga: While I'm not sure what the escape analysis will do about overcreation of `Optional`, I'd usually recommend only using one call to `map` for these sort of cases since the readability is basically the same (or better?) `t -> t.getEngine().getDefaultHost()`

trustin: +1. What do you think, @jmostella ?

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatService.java
```diff
@@ -254,125 +240,19 @@ static String toString(org.apache.catalina.Server server) {
...
-    }
+    @Nullable
+    abstract String hostName();
 
     @Override
     public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
-        final Adapter coyoteAdapter = connector().getProtocolHandler().getAdapter();
-        if (coyoteAdapter == null) {
+        final Optional<Adapter> coyoteAdapter = connector().map(Connector::getProtocolHandler)
+                                                           .map(ProtocolHandler::getAdapter);
```
jmostella: Also need to condense this one to `connector().map(c -> c.getProtocolHandler().getAdapter())`

trustin:
Thanks, @jmostella !

PR #1250, opened 2018-06-11
Don't allocate / copy when constructing retrofit url with no query an…

anuraaga:
…d don't copy headers into a map to iterate them.

Very tiny but obvious microoptimizations.

After
```
Benchmark                         Mode  Cnt      Score     Error  Units
DownstreamSimpleBenchmark.empty  thrpt   20  11639.101 ± 389.186  ops/s
```

Before
```
Benchmark                         Mode  Cnt      Score     Error  Units
DownstreamSimpleBenchmark.empty  thrpt   20  11426.577 ± 329.674  ops/
```

### DIFF
--- a/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactory.java
+++ b/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactory.java
@@ -120,11 +120,12 @@ final class ArmeriaCallFactory implements Factory {
             final HttpUrl httpUrl = request.url();
             final URI uri = httpUrl.uri();
             final HttpClient httpClient = callFactory.getHttpClient(uri.getAuthority(), uri.getScheme());
-            final StringBuilder uriBuilder = new StringBuilder(httpUrl.encodedPath());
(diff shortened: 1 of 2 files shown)


retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactory.java
```diff
@@ -151,8 +152,11 @@ private static HttpResponse doCall(ArmeriaCallFactory callFactory, Request reque
                 default:
                     throw new IllegalArgumentException("Invalid HTTP method:" + request.method());
             }
-            request.headers().toMultimap().forEach(
-                    (key, values) -> headers.add(HttpHeaderNames.of(key), values));
+            final int numHeaders = request.headers().size();
```
trustin: `request.headers()` could be extracted into a local variable?

anuraaga: Fixeed

trustin:
Thanks, @anuraaga !

PR #1251, opened 2018-06-12
Fix a bug where HttpClient connects to a wrong host when custom authority is used

trustin:
Motivation:

When a user specifies both:

- an Endpoint without an IP address and
- a custom authority

the expected behavior is to use the custom authority as a host name (so
that it sends the correct server name on a TLS connection) and to
connect to the IP address of the host specified by the Endpoint.

However, currently, HttpClientDelegate always uses the IP address of the
host name in the custom authority. As a result, HttpClient ends up with
connecting to a wrong host.

Modifications:

- Made sure that HttpClientDelegate resolves the IP address of the
  Endpoint.
  - PoolKey.ipAddr is now non-null.
  - HttpSessionChannelFactory now assumes PoolKey always has an IP address.
- Added test cases
- Miscellaneous
  - Removed small code duplication in HttpClientDelegate

Result:

- HttpClient does not connect to a wrong host anymore.


trustin:
Thank you, reviewers!

Issue #1254, opened 2018-06-13
Armeria server does not handle the `Expect: 100-continue` header correctly.


trustin:
`Http1ObjectDecoder` and `Http2ObjectDecoder` does not send `100 Continue` informational header when a client sends a request with the `Expect: 100-continue` header, which is obviously a violation of the HTTP specification, unless we want to reject the request.

Reported by @jwills

trustin:
One problem with implementing this at the decoder level is that a `Service` will not be given a way to reject the request before a client sends the body because the decoder will always just respond with `100 Continue`. This is not ideal for a service with large request payloads such as file upload service.

trustin:
Alternatively, we may be able to allow a `Service` to choose to handle an `Expect` header by handling the `Expect` header in `HttpServerHandler`:

- `HttpServerHandler` queries a service is it desires to handle the `Expect` header by itself.
- If so, pass the request as it is and do not perform any additional processing.
- If not, remove the `Expect` header from the request headers so that the `Service` does not see the `Expect` header and send `100 Continue` response automatically.

One tiny remaining problem is handling an HTTP/1.0 request. According to the RFC, we have to ignore the `Expect` header if the client is an HTTP/1.0 client. However, `HttpServerHandler` does not have such information; it just knows whether it's HTTP/1.x or HTTP/2. Therefore, we may pass the exact HTTP version to `HttpServerHandler`.

trustin:
How do we let a user tell us his or her `Service` will handle an `Expect` header by itself? I think we can introduce a tag interface like we did `TransientService`, such as, `ExpectAwareHttpService`? Naming suggestion appreciated.

trustin:
After thinking a little bit more, I realize it's harder than initially thought. For example, let's say a user decorates his or her `Service` with various decorators which may want to reject the request, such as `HttpAuthService`. In this setup, some decorators may need to check the `Expect` header but others may not want to do so.

trustin:
Again, this is also harder than expected, because the `Service` implementation would become way more complicated due to different way of building a response, e.g.

```java
if (expectHeaderExists) {
    HttpResponseWriter res = HttpResponse.streaming();
    res.write(HttpHeaders.of(100));
    // ... handle the request and write to 'res' ...
    return res;
} else {
    // ... handle the request ...
    HttpResponse res = HttpResponse.of(200, ...);
    return res;
}
```

jwills:
Is this something I can simply handle myself in my own server impl?

[...]

PR #1255, opened 2018-06-14
Send '100 Continue' response on 'Expect: 100-continue'

trustin:
Motivation:

As explained in #1254, Armeria server currently does not handle the
requests with 'Expect: 100-continue' header. Although it's not ideal, we
can mitigate the problem by always sending '100 Continue' response.

Modifications:

- Modified `Http1RequestDecoder` and `Http2RequestDecoder` to always
  send '100 Continue' response if 'Expect: 100-continue' header is
  present.
- Modified `Http1RequestDecoder` to use `Http1ObjectEncoder` instead of
  writing Netty HTTP objects directly, so that `Http1ObjectEncoder`
  makes sure the response traffic ordering is not messed up.

Result:

- The clients that send their requests with 'Expect: 100-continue' are
  handled correctly.
- Fixes partially: #1254


core/src/main/java/com/linecorp/armeria/server/Http1RequestDecoder.java
```diff
@@ -223,29 +230,42 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
-            final FullHttpResponse res = new DefaultFullHttpResponse(
-                    HttpVersion.HTTP_1_1, status, content);
-
-            final HttpHeaders headers = res.headers();
-            headers.set(HttpHeaderNames.CONNECTION, HttpHeaderValues.CLOSE);
-            headers.set(HttpHeaderNames.CONTENT_TYPE, MediaType.PLAIN_TEXT_UTF_8);
-            headers.setInt(HttpHeaderNames.CONTENT_LENGTH, content.readableBytes());
-
-            future = ctx.writeAndFlush(res);
+        // '100-continue' is the only allowed expection.
```
hyangtack: expectation?

trustin: Fixed :sweat_smile:

PR #1257, opened 2018-06-14
Allow more memory to tests, except streaming tests

trustin:
Motivation:

We currently allow only 128 MiB of heap for tests, which is because we
do not want the streaming tests in `HttpServerTest` take too long
streaming `2 * maxMemory` bytes.

We don't really need to restrict the heap size for other tests. Small
heap size increases the test time for some complex tests that demands
more memory.

Modifications:

- Extract the streaming tests in `HttpServerTest` into
  `HttpServerStreamingTest`
- Exclude `HttpServerStreamingTest` from the `test` task
- Add `testStreaming` task which runs `HttpServerStreamingTest` only
- Allow only 64 MiB to `testStreaming` while allowing 512 MiB to other
  tests

Result:

- Streaming tests end sooner.
- Reduced memory pressure while running non-streaming tests

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/server/HttpServerStreamingTest.java
+++ b/core/src/test/java/com/linecorp/armeria/server/HttpServerStreamingTest.java
@@ -1,0 +1,384 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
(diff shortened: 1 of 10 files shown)


core/src/test/java/com/linecorp/armeria/server/HttpServerStreamingTest.java
```diff
@@ -0,0 +1,385 @@
+/*
+ * Copyright 2016 LINE Corporation
```
hyangtack: 2018...?

trustin: Fixed

PR #1259, opened 2018-06-14
Reproduce Tomcat 9.0 Issue w/o predefined ports.

jmostella:
* application.yml armeria.ports.port=0.
* SpringTomcatApplication cleanup to use @Autowired, for consistency.
* SpringTomcatApplicationItTest added test that validated that only one connector is present.
  Tomcat starts on random port.

### DIFF
--- a/it/spring/boot-tomcat/src/main/resources/config/application.yml
+++ b/it/spring/boot-tomcat/src/main/resources/config/application.yml
@@ -1,6 +1,4 @@
 armeria:
   ports:
-    - port: 8081
+    - port: 0
       protocol: HTTP
-server:
-  port: 8080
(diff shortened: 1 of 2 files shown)


it/spring/boot-tomcat/src/main/java/com/linecorp/armeria/spring/tomcat/demo/SpringTomcatApplication.java
```diff
@@ -32,8 +31,7 @@
 
 @SpringBootApplication
 public class SpringTomcatApplication {
-
-    @Inject
+    @Autowired
```
minwoox: Is there any reason that you changed to this spring specific annotation instead of the annotation in JSR-330?

trustin: +1 for using `@Inject` everywhere.

jmostella: For consistency. If `@Inject` is preferred, then I can change them all to `@Inject`.

trustin: Yes, please.

trustin:
Thanks, @jmostella and reviewers!

PR #1262, opened 2018-06-17
Make zipkin's current context can be nested

kojilin:
Apply this change will fix TraceContext leak in thread problem.
There is a case that we keep armeria's request context aware CompletableFuture
for fetching data, and let multiple request register callback on that future.
So there is a problem that future uses first request's eventloop and RequestContext
to call waiting handlers, but each handler may wrap with it's RequestContext, so
there is nested RequestContext on same thread(onEnter->onEnter->onExit->onExit).

Another solution is application developer should always jump back to current request's
context aware eventloop. e.g.

```

cachedFuture.acceptAsync((r, e)->{
  ...
}, RequestContext.current.contextAwareEventloop);

```

I'm not sure if armeria side need to take care of this or not...

### DIFF
--- a/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanInScopeWrapper.java
+++ b/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanInScopeWrapper.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.internal.tracing;
+
+import javax.annotation.Nullable;
+
+import brave.Tracer.SpanInScope;
+
+/**
+ * SpanInScope Wrapper for keeping previous span scope.
+ */
+public final class SpanInScopeWrapper implements AutoCloseable {
+
+    private final SpanInScope spanInScope;
+
+    @Nullable
+    private final SpanInScopeWrapper previous;
+
+    public SpanInScopeWrapper(SpanInScope current, @Nullable SpanInScopeWrapper previous) {
+        this.spanInScope = current;
+        this.previous = previous;
+    }
+
+    @Nullable
+    public SpanInScopeWrapper getPrevious() {
+        return previous;
+    }
+
+    @Override
+    public void close() {
+        spanInScope.close();
+    }
+}
--- a/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanInScopeWrapper.java
+++ b/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanInScopeWrapper.java
@@ -21,7 +21,7 @@ import javax.annotation.Nullable;
 import brave.Tracer.SpanInScope;
 
 /**
- * SpanInScope Wrapper for keeping previous span scope.
+ * {@link SpanInScope} wrapper that keeps the previous {@link SpanInScope}.
  */
 public final class SpanInScopeWrapper implements AutoCloseable {
 
@@ -36,7 +36,7 @@ public final class SpanInScopeWrapper implements AutoCloseable {
     }
 
     @Nullable
-    public SpanInScopeWrapper getPrevious() {
+    public SpanInScopeWrapper previous() {
         return previous;
     }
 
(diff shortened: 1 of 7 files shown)


zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanInScopeWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+    @Nullable
+    private final SpanInScopeWrapper previous;
+
+    public SpanInScopeWrapper(SpanInScope current, @Nullable SpanInScopeWrapper previous) {
+        this.spanInScope = current;
+        this.previous = previous;
+    }
+
+    @Nullable
+    public SpanInScopeWrapper getPrevious() {
```
trustin: nit: `previous()`?

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanInScopeWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package com.linecorp.armeria.internal.tracing;
+
+import javax.annotation.Nullable;
+
+import brave.Tracer.SpanInScope;
+
+/**
+ * SpanInScope Wrapper for keeping previous span scope.
```
trustin: `{@link SpanInScope} wrapper that keeps the previous {@link SpanInScope}.`

anuraaga:
Can you add test cases with the offending behavior? It'd also help document the affected case.

anuraaga:
I'm also sort of wondering what it means in terms of tracing to restore a parent context after a child context.

Assuming `server_span`, `client_span_1`, and `client_span_2`, I think we end up with something like

```
<------------------server_span----------------------->
        <--------client_span_1------------------>
                             <--server_span?----->
                               <--client_span_2-->
```

`client_span_2`'s parent will be `server_span`, but I think it's supposed to be `client_span_1`. Maybe this is correct, but conceptually I'm having trouble grokking it - what do you think about this? Personally I _always_ run my callbacks on the request thread explicitly because I'd expect tracing to break in any other threading model.

/cc @adriancole if you have any thoughts.

kojilin:
Add test. So if we don't keep previous scope and restore,  then there is a scope can't be close.

Not sure the question, but I don't think there is a situation that client scope has another client scope ?

zipkin/src/test/java/com/linecorp/armeria/it/tracing/HttpTracingNestedContextIntegrationTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    private final CountDownLatch waitCreateCache = new CountDownLatch(1);
+
+    @Rule
+    public final ServerRule server = new ServerRule() {
+        @Override
+        protected void configure(ServerBuilder sb) throws Exception {
+
+            final CountDownLatch countDownLatch = new CountDownLatch(2);
+            final AtomicReference<CompletableFuture<HttpStatus>> cache = new AtomicReference<>();
```
kojilin: We use caffeine, so this is similar situation.

zipkin/src/test/java/com/linecorp/armeria/it/tracing/HttpTracingNestedContextIntegrationTest.java
```diff
@@ -0,0 +1,184 @@
...
+                }
+            }));
+
+            sb.service("/read-cache", decorate("service/read-cache", new AbstractHttpService() {
+                @Override
+                protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req) throws Exception {
+                    try {
+                        final RequestContext requestContext = RequestContext.current();
+                        return HttpResponse.from(
+                                cache.get().thenApply(status -> {
```
kojilin: Or the application dev must always know using `thenApplyAsync` with current evetloop.

zipkin/src/test/java/com/linecorp/armeria/it/tracing/HttpTracingNestedContextIntegrationTest.java
```diff
@@ -0,0 +1,184 @@
...
+                                    try (SafeCloseable ignored = RequestContext.push(requestContext)) {
```
anuraaga: I see, the example seems to have two unrelated server context's on the same request context stack. Is this allowed? We have features like `RequestContext.onChild`, which I don't think support this sort of behavior, pushing a context here is like saying we're creating a new child context, so the unrelated "parent"'s `onChildCallbacks` would run too from what I can tell.

kojilin: oops, right. Didn't notice about `#onChild`.  
Also I should use `propagateContextIfNotPresent` in here too? https://github.com/line/armeria/blob/master/rxjava/src/main/java/com/linecorp/armeria/common/rxjava/RequestContextCompletable.java#L38

anuraaga: Yeah going with propagate if not present seems safer for that sort of callback, so while new contexts don't get set up properly, at least old contexts won't be affected poorly which I guess is usually better. Maybe some sort of logging could be added for such a case.

FWIW, the reason I added this comment a long time ago is because in my experience it's extremely difficult if not impossible to get "context" to work in asynchronous code in Java (which can only try mapping them to `ThreadLocal`) without jumping threads. Perhaps we could call this out in the documentation better - wonder what other's thoughts are

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/RequestContext.java#L271

kojilin: Thanks, it's so easy to forget using Async callback so thats why we want to use this kind of hook for chaining sync/async operations.

I will close this PR and try to make RxJava hook strictly.

PR #1266, opened 2018-06-18
Make RequestContext's RxJava auto inject hook strictly

kojilin:
Currently RequestContext RxJava hook pushes context to stack without
checking current context. Because armeria doesn't predict the case that
unrelated RequestContexts push on the same thread nestedly, it may cause
some unexpected problems.

e.g.
* Current zipkin related module doesn't expect nested context, if there
are two different contexts called sequentially, the outer span's scope
won't be closed.

* Application may utilize `RequestContext#onChild` to propagate
attribute from parent to child, so the 2nd context will see unrelated
request's context.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/RequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/common/RequestContext.java
@@ -224,7 +224,6 @@ public interface RequestContext extends AttributeMap {
      * Pushes the specified context to the thread-local stack. To pop the context from the stack, call
      * {@link SafeCloseable#close()}, which can be done using a {@code try-with-resources} block.
      *
-     * @see #push(boolean)
      * @deprecated Use {@link #push(boolean)}.
      */
     @Deprecated


core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -276,6 +276,25 @@ default EventLoop contextAwareEventLoop() {
         return new RequestContextAwareEventLoop(this, eventLoop());
     }
 
+    /**
+     * Pushes this context to the thread-local stack if there is no current context. If there is, this method
+     * will throw IllegalStateException. To pop the context from the stack, call {@link SafeCloseable#close()},
+     * which can be done using a {@code try-with-resources} block.
+     */
+    default SafeCloseable propagateContextIfNotPresent() {
```
trustin: - How about `pushIfAbsent()`?
- I'm also curious if we should add `RequestContext.push()` that deprecates `RequestContext.push(RequestContext)`. What do you think? /cc @anuraaga

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -276,6 +276,25 @@ default EventLoop contextAwareEventLoop() {
...
+        return mapCurrent(currentContext -> {
+            if (currentContext != this) {
+                throw new IllegalStateException(
+                        "Trying to call object made with makeContextAware or object on executor made with " +
+                        "makeContextAware with context " + this +
+                        ", but context is currently set to " + currentContext + ". This means the " +
+                        "callback was passed from one invocation to another which is not allowed. Make " +
+                        "sure you are not saving callbacks into shared state.");
```
trustin: I guess this exception message needs to change?

kojilin:
fix test, and I think we don't expect push has null context as parameter right?

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -276,6 +209,114 @@ default EventLoop contextAwareEventLoop() {
         return new RequestContextAwareEventLoop(this, eventLoop());
     }
 
+    /**
+     * Pushes the specified context to the thread-local stack. To pop the context from the stack, call
+     * {@link SafeCloseable#close()}, which can be done using a {@code try-with-resources} block.
+     *
+     * @see #push()
+     * @deprecated Use {@link #push()} instead.
```
trustin: - Please remove 'instead' for consistency. (Originally suggested by @anuraaga)
- Probably OK to remove the `@see` tag because it's mentioned by `@deprecated`?

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -276,6 +209,114 @@ default EventLoop contextAwareEventLoop() {
...
+    static SafeCloseable push(RequestContext ctx) {
+        return ctx.push(true);
+    }
+
+    /**
+     * Pushes the specified context to the thread-local stack. To pop the context from the stack, call
+     * {@link SafeCloseable#close()}, which can be done using a {@code try-with-resources} block.
+     *
+     * @see #push(boolean)
+     * @deprecated Use {@link #push(boolean)} instead.
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -276,6 +209,114 @@ default EventLoop contextAwareEventLoop() {
...
+            if (oldCtx != null) {
+                return () -> RequestContextThreadLocal.set(oldCtx);
+            } else {
+                return RequestContextThreadLocal::remove;
+            }
+        }
+    }
+
+    /**
+     * Pushes this context to the thread-local stack if there is no current context. If there is, this method
```
trustin: If there is, -> If there is and it is not same with this context (i.e. not reentrance),

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -276,6 +209,114 @@ default EventLoop contextAwareEventLoop() {
...
+     * will throw IllegalStateException. To pop the context from the stack, call {@link SafeCloseable#close()},
```
trustin: IllegalStateException -> `an {@link IllegalStateException}`

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -276,6 +209,114 @@ default EventLoop contextAwareEventLoop() {
...
+     * which can be done using a {@code try-with-resources} block.
+     */
+    default SafeCloseable pushIfAbsent() {
+        return mapCurrent(currentContext -> {
```
trustin: Now that this method is moved from the subclass, we don't need to use `mapCurrent()` anymore but use `RequestContextThreadLocal.get()` directly, which is supposed to be easier to read:

```java
final RequestContext oldCtx = RequestContextThreadLocal.getAndSet(this);
if (oldCtx != null && oldCtx != this) {
    throw new IllegalStateException(...);
}
return push();
```

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -276,6 +209,113 @@ default EventLoop contextAwareEventLoop() {
...
+    @Deprecated
+    static SafeCloseable push(RequestContext ctx) {
+        return ctx.push(true);
+    }
+
+    /**
+     * Pushes the specified context to the thread-local stack. To pop the context from the stack, call
+     * {@link SafeCloseable#close()}, which can be done using a {@code try-with-resources} block.
+     *
+     * @see #push(boolean)
```
trustin: This could be removed.

PR #1268, opened 2018-06-19
Do not confuse an IDE by having a source directory with two output directories

trustin:
Motivation:

IntelliJ IDEA generates an error event like the following when a source
directory has more than one output directory:

    java.lang.Throwable: Module 'boot-tomcat_main' and module 'boot-tomcat8.5_main' have the same content root: file:///home/trustin/Workspaces/armeria/it/spring/boot-tomcat/src/main/java

Modifications:

- Do not use the `sourceSets` directive for a derived project.
- Miscellaneous
  - Reorder the directives in `:it:spring:boot-tomcat8.5` so that build
    plugin part comes first
  - Remove an unused dependency

Result:

- IDE devs are happy.

### DIFF
--- a/it/spring/boot-tomcat8.5/build.gradle
+++ b/it/spring/boot-tomcat8.5/build.gradle
@@ -1,3 +1,15 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'org.springframework.boot:spring-boot-gradle-plugin:2.0.2.RELEASE'
+    }
+}
+
+apply plugin: 'java'
+apply plugin: 'org.springframework.boot'
+
 dependencyManagement {
     // Override Tomcat versions.
     dependencies {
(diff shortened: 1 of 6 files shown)


jmostella:
LGTM

jmostella:
@trustin could you add the version of IntelliJ IDEA that this applies to?

it/spring/boot-tomcat8.5/build.gradle
```diff
@@ -17,21 +6,43 @@ buildscript {
...
     compile('org.springframework.boot:spring-boot-starter-web') {
         exclude group: 'org.springframework.boot', module: 'spring-boot-starter-tomcat'
     }
     testCompile('org.springframework.boot:spring-boot-starter-test')
 }
+
+// Use the sources from ':it:spring:boot-tomcat'.
+// NB: We should not add these in the 'sourceSets' directive because that will make the directories mentioned
+//     below are added to two projects and having a source directory with two output directories will confuse
+//     the IDE.
```
jmostella: `IntelliJ IDEA version x.`

trustin: I think it was around IntelliJ IDEA 12, but I forgot. Let me update anyway.

trustin: Actually, it's not 12 but some other version before they used year.

trustin: Well, I'm not really sure. Let me just mention without version numbers.

trustin: Fixed

trustin:
Thanks for reviewing :slightly_smiling_face:

PR #1270, opened 2018-06-27
Use the reference to the dedicated page about service discovery in th…

trustin:
…e site front page

.. because it covers both DNS and ZooKeeper


trustin:
Will merge once build passes since it's trivial.

PR #1272, opened 2018-06-27
Use getServiceDescriptor to get stub methods instead of removed API. …

anuraaga:
…getServiceDescriptor has existed since 1.0.0.

I must have missed this method when first implementing since it's at the end of the file...


trustin:
Thanks a lot for a quick fix, @anuraaga !

PR #1273, opened 2018-06-27
Remove Javassist from dependencies

trustin:
.. because Netty does not depend on it, even optionally, since Feb 22 2017


trustin:
I added 'defect' label because Javassist 3.23.0 had a compatibility issue with Java 8 and a user who imports our BOM and uses Javassist will be affected, although there won't be many.

PR #1276, opened 2018-07-03
Make Armeria Retrofit subscriber not buffer response

kojilin:
Current implementation will buffer all data and call callback when complete.
So it's not suitable for downloading large files.

This PR will make subscriber requests data one by one from response.　And
because `SourceBuffer#read` and `Subscriber#onNext` are using same thread, we need to
using CommonPools to run `SourceBuffer#read` for avoiding deadlock.


trustin:
@kojilin Hmm? What's going on?

kojilin:
Too many edge case, let me add more test and complete the original code

Issue #1275, opened 2018-07-03
AbstractMethodError in latest tcnative


anuraaga:
I tried upgrading armeria, which gets the latest netty and tcnative, but I get the below exception on all client requests. JDK10 issue? Fixing tcnative to 2.0.10 works fine.

Java: JDK10
armeria: 0.67.1
netty: 4.1.25
tcnative: 2.0.11

```
java.lang.AbstractMethodError: null
	at io.netty.internal.tcnative.SSL.doHandshake(Native Method) ~[netty-tcnative-boringssl-static-2.0.11.Final.jar:2.0.11.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.handshake(ReferenceCountedOpenSslEngine.java:1563) ~[netty-handler-4.1.25.Final.jar:4.1.25.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.wrap(ReferenceCountedOpenSslEngine.java:652) ~[netty-handler-4.1.25.Final.jar:4.1.25.Final]
	at javax.net.ssl.SSLEngine.wrap(SSLEngine.java:511) ~[?:?]
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:997) ~[netty-handler-4.1.25.Final.jar:4.1.25.Final]
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:803) ~[netty-handler-4.1.25.Final.jar:4.1.25.Final]
	at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:771) ~[netty-handler-4.1.25.Final.jar:4.1.25.Final]
	at io.netty.handler.ssl.SslHandler.handleUnwrapThrowable(SslHandler.java:1207) ~[netty-handler-4.1.25.Final.jar:4.1.25.Final]
	at io.netty.handler.ssl.SslHandler.decodeNonJdkCompatible(SslHandler.java:1191) ~[netty-handler-4.1.25.Final.jar:4.1.25.Final]
	at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1223) ~[netty-handler-4.1.25.Final.jar:4.1.25.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:489) ~[netty-codec-4.1.25.Final.jar:4.1.25.Final]
[...]

PR #1277, opened 2018-07-05
Fix build errors on JDK 10

### LINKED ISSUE #1267: Compile failure at spring-boot-autoconfigure with JDK10
I didn't have time to dig much but I failed to build armeria with JDK10. Do you have any idea to fix this error?

```sh
$ git log --oneline | head -1
201af602 Remove @VisibleForTesting from TomcatService (#1261)
```

```sh
$ ./gradlew --version
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (file:/Users/JP20695/.gradle/wrapper/dists/gradle-4.7-all/4cret0dgl5o3b21weaoncl7ys/gradle-4.7/lib/groovy-all-2.4.12.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

------------------------------------------------------------
Gradle 4.7
------------------------------------------------------------

Build time:   2018-04-18 09:09:12 UTC
Revision:     b9a962bf70638332300e7f810689cb2febbd4a6c

Groovy:       2.4.12
Ant:          Apache Ant(TM) version 1.9.9 compiled on February 2 2017
JVM:          10.0.1 ("Oracle Corporation" 10.0.1+10)
OS:           Mac OS X 10.13.4 x86_64
```

```
$ ./gradlew build
> Task :spring:boot-autoconfigure:compileJava FAILED
/Users/JP20695/src/armeria/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ThriftServiceUtils.java:91: error: incompatible types: invalid method reference
                      .map(Class::getEnclosingClass)
                           ^
    method getEnclosingClass in class Class<T> cannot be applied to given types
      required: no arguments
      found: Object
      reason: actual and formal argument lists differ in length
  where T is a type-variable:
    T extends Object declared in class Class
Note: /Users/JP20695/src/armeria/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ThriftServiceUtils.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':spring:boot-autoconfigure:compileJava'.
> Compilation failed; see the compiler error output for details.

* Try:

* Get more help at https://help.gradle.org

BUILD FAILED in 12s
133 actionable tasks: 2 executed, 131 up-to-date
```

For some reason, entry was recognized as `Object`, not `Class` at https://github.com/line/armeria/blob/armeria-0.66.0/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ThriftServiceUtils.java#L91

If you rewrite code as follow, it can pass compilation.
```diff
-                      .map(Class.class::cast)
-                      .map(Class::getEnclosingClass)
-                      .map(Class::getName)
+                      .map(e -> ((Class<?>) e).getEnclosingClass().getName())
```

trustin:
- Fixed a Javadoc error about a `<table>` tag
- Fixed a compilation error in `ThriftServiceUtils`
- Changed the CI services to use JDK 10
- Update netty-tcnative-boringssl-static to 2.0.12, which may solve #1275
- Relaxed `ArmeriaMeterBindersConfigurationTest.testDefaultMetrics()`
- Add `--release 8` option to `javac` command.
- Require JDK 10 to release.

Result:

- Fixes #1267


trustin:
@anuraaga I have no idea why, but I do not get the `AbstractMethodError` with this PR. Could you paste your `java -version` output? Mine is:

```
java version "10.0.1" 2018-04-17
Java(TM) SE Runtime Environment 18.3 (build 10.0.1+10)
Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.1+10, mixed mode)
```

anuraaga:
It looks the same as mine (OS is Linux)

```
java version "10.0.1" 2018-04-17
Java(TM) SE Runtime Environment 18.3 (build 10.0.1+10)
Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.1+10, mixed mode)
```
Any pointers on what else I can look at? Too bad the exception is thrown in a native method :(

anuraaga:
Also one special factor is I am using client SSL auth - I'm not sure if we have any tests of that in the codebase yet

anuraaga:
By the way, we are using EC256 certs for both server and client-side authentication. I tried setting up a test of client auth using `SelfSignedCertificate`, which uses RSA and couldn't repro the error, but will see if I can get closer to my environment.

trustin:
@anuraaga That's interesting. I just ran `./gradlew check` since it should include all SSL tests, but I guess it doesn't cover enough. Are you interested in writing a test case or a simple reproducer?

anuraaga:
Thanks for the pointer - will also try it out!

trustin:
Ready for reviews.

build.gradle
```diff
@@ -94,6 +96,17 @@ configure(projectsWithFlags('java')) {
         testCompile 'org.assertj:assertj-core'
         testCompile 'org.mockito:mockito-core'
         testCompile 'org.apache.httpcomponents:httpclient'
+
+        // Fail the build if Java 9+ APIs are used.
+        signature 'org.codehaus.mojo.signature:java18:1.0@signature'
+    }
+
+    animalsniffer {
```
anuraaga: If I'm not mistaken, JDK9+ added the `--release` flag which prevents compiling against newer APIs.

http://openjdk.java.net/jeps/247

trustin: I'm curious if we can still use the JDK9+ APIs selectively using the `--release` option, i.e. use faster implementation in JDK 9/10 if possible, or fall back to less efficient one.

anuraaga: Yeah it does allow multi release jars too. Though it might just increase jar size without that much benefit, e.g., eventually they'll optimize string concatenation in newer Java, buy haven't heard of it happening yet

trustin: Updated to use `--release` option. We now reject the build if the current JDK is older than 9, because JDK 8 javac does not recognize the `--release` option.

trustin: Realized the we can just remove the `--release` option for JDK 8. Updated again.

trustin:
I also add a release-time constraint that JDK 10 must be used when release is performed. Please review again.

trustin:
Thanks, reviewers!

PR #1280, opened 2018-07-11
Update dependencies

trustin:
- Netty 4.1.25 -> 4.1.26
  - Netty TCNative BoringSSL 2.0.10 -> 2.0.12
- Bouncy Castle 1.59 -> 1.60
- Tomcat 8.0.52 -> 8.0.53
- Build-time only dependencies
  - Apache HttpComponents 4.5.5 -> 4.5.6
  - Awaitility 3.1.0 -> 3.1.1
  - Checker framework 2.5.2 -> 2.5.3
  - Checkstyle 8.10.1 -> 8.11

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -63,7 +63,7 @@ com.google.protobuf:
   protobuf-gradle-plugin: { version: '0.8.6' }
 
 com.puppycrawl.tools:
-  checkstyle: { version: '8.10.1' }
+  checkstyle: { version: '8.11' }
 
 com.spotify:
   completable-futures:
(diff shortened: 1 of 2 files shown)


dependencies.yml
```diff
@@ -132,7 +132,7 @@ io.micrometer:
     - org.springframework:spring-webmvc
 
 io.netty:
-  netty-codec-http2: { version: &NETTY_VERSION '4.1.25.Final' }
+  netty-codec-http2: { version: &NETTY_VERSION '4.1.26.Final' }
```
anuraaga: Ah was the problem just incompatibility with Netty itself?

trustin: Yeah, it seems on. Like using wrong version of protoc with gRPC :-)

PR #1281, opened 2018-07-11
Fix a bug where a RST frame for a non-existent stream can be sent

trustin:
Motivation:

When a client-side subscriber of a response stream closes the response
stream even before any HEADERS frame is sent to the stream, Armeria
tries to send a RST frame to that stream, which is a violation of
HTTP/2.

Modifications:

- Do not send a RST frame if the stream does not exist yet.
- Do not fail a client-side HTTP/2 connection even if
  Http2ResponseDecoder failed to write a received HttpObject to a
  response stream.
- Remove the test skip code in GrpcClientTest

Result:

- GrpcClientTest is not flaky anymore.
- HTTP/2 connections are not closed unexpectedly.


trustin:
Will run the build several more times so that we are sure this fixed the flakiness.

trustin:
Had no failures for 5 consecutive builds for both Travis CI and AppVeyor. Let me merge :-)

PR #1284, opened 2018-07-12
Do not fail but warn when Optional and Default are used redundantly

trustin:
Motivation:

Armeria currently raises an exception when a user creates an annotated
service with redundant use of Optional and Default:

    // No need to use Optional because 'value' is always present.
    @Get("/foo/:value")
    public String foo(@Param Optional<String> value) {}

    // No need to use @Default because 'value' is always present.
    @Get("/bar/:value")
    public String bar(@Param @Default("defVal") String value) {}

    // Optional and @Default were used together. 'value' will always be present.
    @Get("/baz")
    public String baz(@Param @Default("defVal") Optional<String> value) {}

We could relax a little bit and just warn so that the server starts up
anyway because it wouldn't harm and it would give a user less surprise.

Modifications:

- Log a warning message instead of raising an exception when a redundant
  usage of `@Default` and `Optional` was detected.
- Clean up the warning message a little bit so it is easier to
  understand.

Result:

- Better UX

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
@@ -571,12 +571,12 @@ final class AnnotatedValueResolver {
     }
 
     @VisibleForTesting
-    public boolean shouldExist() {
+    boolean shouldExist() {
         return shouldExist;
     }
 
     @VisibleForTesting
-    public boolean shouldWrapValueAsOptional() {
+    boolean shouldWrapValueAsOptional() {
         return shouldWrapValueAsOptional;
     }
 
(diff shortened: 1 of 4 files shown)


trustin:
Example warning messages:

```
10:43:38.030 [main] WARN  c.l.a.server.AnnotatedValueResolver - Optional is redundant for path variable 'var1' because the value is always present.
10:43:38.031 [main] WARN  c.l.a.server.AnnotatedValueResolver - @Default is redundant for path variable 'var1' because the value is always present.
10:43:38.032 [main] WARN  c.l.a.server.AnnotatedValueResolver - @Default was used with 'java.util.Optional<java.lang.String> value'. Optional is redundant because the value is always present.
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -574,6 +575,11 @@ public boolean shouldExist() {
         return shouldExist;
     }
 
+    @VisibleForTesting
+    public boolean shouldWrapValueAsOptional() {
```
minwoox: Can we make this as package-local?

trustin:
Thanks for reviewing!

PR #1285, opened 2018-07-12
Do not add HTTP port if TLS was enabled and no port was specified

### LINKED ISSUE #1279: ServerBuilder.tls should cause server to only start with HTTPS?
I was confused recently when I used `ServerBuilder.tls(ctx)` and it started with both HTTP and HTTPS. I was expecting behavior similar to when calling `ServerBuilder.https(0)` in that when TLS is requested, it is the only one enabled. Does anyone else feel the same confusion?

trustin:
Motivation:

When a user sets up his or her server like the following:

    new ServerBuilder()
            .tls(...)
            .service(...)
            .build();

The server will listen at two ephemeral ports, one for HTTP and the
other for HTTPS.

This is a surprising behavior - a user would expect the server to listen
at an HTTPS port only.

Modifications:

- Do not add HTTP port but only HTTPS port if TLS was enabled and no
  port was specified.

Result:

- Fixes #1279

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
@@ -1037,7 +1037,7 @@ public final class ServerBuilder {
             if (!this.ports.isEmpty()) {
                 ports = ImmutableList.copyOf(this.ports);
             } else {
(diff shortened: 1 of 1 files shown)


trustin:
Thanks for reviewing!

Issue #654, opened 2017-06-30
Migrate DocService to more scalable frontend framework


trustin:
.. so we can add more features and organize them easily. Vue.js or Angular?

trustin:
http://vuematerial.io/ and https://material.angular.io/

anuraaga:
Or React? :) www.material-ui.com/

Just to add the option, I guess there's never a right or wrong answer on which framework to pick (which makes it hard to pick...).

trustin:
Haha, yeah. I have some experience with Angular.js, so I was rather inclined to Angular than to React, but you know, Vue.js is also hot these days. :-D I think it's just a matter of who's gonna do the job and he or she prefers which.

PR #1288, opened 2018-07-12
Migrate docs client webapp to React / Typescript / Material UI

anuraaga:
This large PR migrates the docs client from a Javascript/jQuery/Bootstrap-based architecture to Typescript/React/MaterialUI. 

It adds a separate `docs-client` project which is a standard NodeJS / webpack based build, run using `gradle-node-plugin`. Currently it's a separate JAR, but if it makes more sense to add it into the `core` JAR pretty sure we can do that too.

I changed the URLs a little, `/method` -> `/methods`, `/namedType` -> `/structs` + `/enums`. Since struct and enum pages are quite different, it made sense to separate the URLs out. The old URls are still supported and redirect to the new ones.

The TypeScript configuration is as strict as possible, including whitespace / semantic linting based on AirBnb style + Prettier. It's a pretty common configuration nowadays.

The `README` covers development concepts, let me know if there's anything else to add. 

To try out the client, I'd recommend pulling the branch and just stuffing a `Thread.sleep(Long.MAX_VALUE)` in one of `ThriftDocServiceTest` or `GrpcDocServiceTest`.

For #654


docs-client/3rd-party-licenses.txt
```diff
@@ -0,0 +1,1618 @@
+THE FOLLOWING SETS FORTH ATTRIBUTION NOTICES FOR THIRD PARTY SOFTWARE THAT MAY BE CONTAINED IN PORTIONS OF THE @ARMERIA/DOCS CLIENT PRODUCT.
```
trustin: Question: Is this auto-generated? Would it be possible to have this generated into `/licenses`?

docs-client/src/components/VariableList/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+import Table from '@material-ui/core/Table';
+import TableBody from '@material-ui/core/TableBody';
+import TableCell from '@material-ui/core/TableCell';
+import TableHead from '@material-ui/core/TableHead';
+import TableRow from '@material-ui/core/TableRow';
+import Typography from '@material-ui/core/Typography';
+import React from 'react';
+
+import { Specification } from '../../lib/specification';
```
trustin: Global comment: The curly braces are redundant?

docs-client/src/index.css
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+pre, code, kbd, var, samp, tt, .code {
+  font-family: 'Roboto Mono', 'Consolas', 'Menlo', 'Monaco', 'Lucida Console', 'Liberation Mono',
+    'DejaVu Sans Mono', monospace !important;
```
trustin: Could you avoid using `!important`?

docs-client/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "allowSyntheticDefaultImports": true,
+    "declaration": false,
+    "esModuleInterop": true,
+    "experimentalDecorators": true,
+    "forceConsistentCasingInFileNames": true,
+    "jsx": "preserve",
+    "lib": ["esnext", "es2017", "es2016", "es2015", "dom"],
+    "module": "esnext",
+    "moduleResolution": "node",
+    "newLine": "LF",
```
trustin: Question: Does this determine the line separators of compilation output or the `.tsx` files?

settings.gradle
```diff
@@ -7,6 +7,7 @@ includeWithFlags ':bom',                       'bom'
 
 // Published Java projects
 includeWithFlags ':core',                       'java', 'publish', 'shade', 'trim'
+includeWithFlags ':docs-client',                'publish'
```
trustin: Could we avoid publishing this but just use the build output into the resource set of `:core`?

gradle/scripts/lib/java-shade.gradle
```diff
@@ -257,7 +257,7 @@ private Configuration createShadedTestRuntimeConfiguration(
 
     if (recursedProject.tasks.findByName('trimShadedJar')) {
         project.dependencies.add(shadedTestRuntime.name, files(recursedProject.tasks.trimShadedJar.outJarFiles))
-    } else {
+    } else if (recursedProject.tasks.findByName('shadedJar')) {
         project.dependencies.add(shadedTestRuntime.name, files(recursedProject.tasks.shadedJar.archivePath))
```
trustin: Thanks!

docs-client/3rd-party-licenses.txt
anuraaga: The file is automatically generated by the command in the README. Do you mean move this file into `armeria/licenses/web-licenses.txt`? Can definitely do that.

docs-client/src/components/VariableList/index.tsx
anuraaga: I didn't add a "default export" to specification since it exports multiple things, and I've found in that case exporting a default can result in ugly looking imports like `import Specification, { Method, Service } from 'specification'`. I could add a default export though what do you think?

docs-client/src/index.css
anuraaga: Since we're using material-ui components for the majority of our styling, and it defines it's own styles, it can be tricky to override fonts - the styles for each component are local to the component, so generally you would define stylized versions of each component you want to override. I thought it's probably overkill for this one single easy-to-grok CSS rule but can play with it some more if it's still a concern.

docs-client/tsconfig.json
anuraaga: The compilation output, but since we only pipe though webpack anyways this doesn't matter so removed.

settings.gradle
anuraaga: Yep both would work, though it would require a bit more configuration. I thought this might be a good idea to separate though since it's then very easy to exclude the jar to exclude the webapp if you don't care about it. What do you think?

gradle/scripts/lib/java-shade.gradle
anuraaga: I actually wasn't sure if I should edit this - should I edit the other gradle scripts repo instead?

docs-client/3rd-party-licenses.txt
trustin: Yes, please. Could you also update `NOTICE.txt` to point at the `web-licenses.txt` there?

docs-client/src/components/VariableList/index.tsx
trustin: I actually never knew `{ Something }` and `Something` are different. If they have different meaning, please leave them as they are.

docs-client/src/index.css
trustin: Does this solution work? https://stackoverflow.com/a/45403932

docs-client/tsconfig.json
trustin: Thanks!

settings.gradle
trustin: I'd still prefer making this part of the core JAR unless it adds too much. ;-)

gradle/scripts/lib/java-shade.gradle
trustin: Let me send a PR for this to upstream.

trustin: Fix has been pushed to `master`. Please rebase.

docs-client/src/index.css
anuraaga: It applies the font to all typography, but this one is only for code snippets. By the way, the default font already uses `Roboto` per material-ui convention, so I didn't change the default typography.

trustin:
@anuraaga Please rebase your PR so that Travis CI builds do not fail anymore due to JDK download failure.

docs-client/src/index.css
trustin: Ah I see. Let's leave this as it is for now.

core/build.gradle
```diff
@@ -1,4 +1,6 @@
 dependencies {
+    compile project(':docs-client')
```
trustin: This has to be removed because otherwise the pom will refer to an unpublished artifact.

anuraaga: Completely forgot to remove this! It's not needed since `builtBy` takes care of the task dependency.

trustin:
Miscellaneous issues found:
- There's no message when there are no parameters/exceptions/members, e.g. 'This method does not have parameters.'
- 'Use these HTTP headers for all functions.' checkbox does not work.

trustin:
From aesthetic PoV, I think it looks worse. :sweat_smile: But I guess we can improve on that later.

trustin:
screenshot_2018-07-19 armeria documentation service

anuraaga:
Thanks - I found in the old code the only one that shows "There are no..." is the method parameters, the rest are just hidden so tried to reproduce that.

Also, sticky headers should be working if you submit the form. In the old code, it coincidentally sort of worked even if you didn't submit because state wasn't reinitialized, but if you went to a page with debug form to one without and back, they'd be gone. Is it better if the sticky kicks in before hitting submit button? Having three sources of the information (URL, JS state, specification) isn't terrible but a bit complicated.

And yeah happy to tweak aesthetic in this or followup PR (maybe a material design designer sits next to you? :D). In particular, I'm thinking of adding `Paper` in certain places, I think the grey background makes it look weird in many places.

trustin:
> Thanks - I found in the old code the only one that shows "There are no..." is the method parameters, the rest are just hidden so tried to reproduce that.

I see. It would be still nicer if 'there are no...' is shown for exceptions and fields as well.

> Also, sticky headers should be working if you submit the form. In the old code, it coincidentally sort of worked even if you didn't submit because state wasn't reinitialized, but if you went to a page with debug form to one without and back, they'd be gone. Is it better if the sticky kicks in before hitting submit button?

I think so. A user may have filled all the headers and then realized that he or she was on a wrong service method. :-)

> And yeah happy to tweak aesthetic in this or followup PR.

No worries. I'm not very sure what changes would make it look better yet anyway. Let me play with it a little bit once merged. :bowing_man:

anuraaga:
Thanks - think I addressed the comments. Also added some `Paper` so it's less grey and I'm sure we'll come up with more aesthetic tweaks in the future :)

trustin:
Thanks! It looks much better. :-)

Some more bugs:

- If you choose Enums -> `ConsistencyLevel` and scroll down the right pane, the left pane looks truncated.
- The left pane does not disappear into [hamburger](https://en.wikipedia.org/wiki/Hamburger_button) even if you shrink the width of the window.
- The HTTP headers text area should be visible if it's not empty. For example, if you tick the 'Use these HTTP headers for all functions' checkbox in `HelloService.hello()` and switch to `SleepService.sleep()`, the HTTP headers text area should be visible.

trustin:
As you see below, the left pane's height is not synchronized with the right pane's height.
image

anuraaga:
Thanks for the thorough check, had forgotten about the navigation. Now it's responsive, and also fixed the headers textbox so it shows when there are headers.

And after lots of playing, figured out the CSS magic to make the heights work I think!

trustin:
Great job! :-)

But, in mobile layout, after I click the hamburger, I find no way to hide the method list. I also see excessive top margin. Perhaps appbar should be there so a user can click the hamburger again to hide the method list?

anuraaga:
Ok render the mobile drawer at 80% width so clicking off of it can close it, this seems to be what e.g., Google Maps does. I also had the top-space just copied from the `material-ui.com` example, not sure why they had it seems fine without it :) Not just trying to figure out why their example renders sticky subheaders correctly but not mine, if I can't figure it out I'll just disable it.

anuraaga:
Got the method names to not overlap - think it's ready for another look

trustin:
More minor issues. :pray: 

Could we reduce the left margin of the hamburger button so it is aligned with the page content?

image

Could we reduce the left margin of the left pane so that method names are placed better? Also, when a cursor hovers at a method name, the gray selection does not cover the row completely, leaving white background on the left of the method name.

Could we reduce the font size of the items in the left pane? It looks too big comparing to the right pane content. On the other hand, could the service names in the left pane be bigger?

I see a performance issue when pressing the hamburger button. It takes about a second since I clicked the button for the left pane to show up. Could we make it appear sooner?

anuraaga:
Fixed the method margin, it was introduced by extra `ul` tags that should have been `div`.

Tweaked the fonts so service name is larger and method names are smaller. Typography is a complex problem, and hoping to use the defaults as much as possible for now and customize it better in the future - WDYT?

And I'm not sure if I'm seeing the same performance issue, it feels like around 2-300ms for me, but I tried adding a prop that might make it faster for slower devices.

docs-client/src/containers/App/index.tsx
```diff
@@ -102,26 +107,31 @@ function AppDrawer({ classes, navigateTo, specification }: AppDrawerProps) {
             </ListItemText>
           </ListItem>
           {specification.getServices().map((service) => (
-            <li key={service.name}>
-              <ul>
-                <ListSubheader className={classes.methodHeader}>
+            <div key={service.name}>
+              <ListSubheader className={classes.methodHeader}>
+                <Typography variant="title">
```
trustin: `subheading` looks better to me. What do you think?

anuraaga: You're right, switched to it and also added `<code>`

anuraaga:
Thanks a lot for the detailed review!

anuraaga:
As a preview, looking forward to sending https://github.com/anuraaga/armeria/pull/5 with pluggable auth support after this is in :)

trustin:
@minwoox @hyangtack Gentle ping :stuck_out_tongue_winking_eye:

docs-client/src/containers/App/index.tsx
```diff
@@ -0,0 +1,352 @@
...
+  createStyles,
+  Theme,
+  withStyles,
+  WithStyles,
+} from '@material-ui/core/styles';
+import Toolbar from '@material-ui/core/Toolbar';
+import Typography from '@material-ui/core/Typography';
+import MenuIcon from '@material-ui/icons/Menu';
+import React from 'react';
+import { hot } from 'react-hot-loader';
```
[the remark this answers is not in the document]
anuraaga: Yeah - the module name is on the right side. It is almost always all lowercase, but if it only exports a single class, sometimes it is uppercase (the case for the material ui exports). There is some history behind material ui exporting a class per module because webpack used to fail at tree shaking them when importing from the top-level. I feel this is probably not the case anymore, but they still document their imports in this form. In general it is considered bad practice for a library to not export everything from the top level though.

I stick to all lowercase all the time for modules for simplicity.

The left side is what is imported. A function is all lowercase while a class is camelcase.

Hope that answers the question :)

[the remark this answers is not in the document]
minwoox: Ah! Didn't know that. Thank you for letting me know. :)

trustin:
Finally-! Thank you so much for your patience and contribution, @anuraaga !

PR #1290, opened 2018-07-13
Allow to register duplicated endpoints under HealthCheckedEndpointGroup

imasahiro:
Motivations:
- A user sometimes want to register duplicated entries to EndpointGroup
  to implement `virtual node` to balance requests. But if duplicated
  entries are added to an `EndpointGroups`, it stops to update EndpointGroup
  refresh.

Modifications:
- Apply merging function not to throw exception when duplicated entries
  are found in endpoint group list.


trustin:
Thanks, @imasahiro !

PR #1291, opened 2018-07-15
Add the git checkout action to post-release-msg

minwoox:
So that a user(me) never forget to checkout gh-pages before publishes the site.

### DIFF
--- a/.post-release-msg
+++ b/.post-release-msg
@@ -18,6 +18,7 @@
 5. Copy the web site generated to the 'gh-pages' branch, e.g.
 
    cd ../site-armeria
+   git checkout gh-pages
    rm -fr .buildinfo .doctrees .gradle *
    rsync -aiP --exclude '.*' ../upstream-armeria/site/build/site/ .
    git add -A .


anuraaga:
Just wondering, is it a good time to implement a `publishSite` Gradle task that executes those commands?

minwoox:
I am not sure. There are some stuffs that we have to do manually such as visiting the sonatype site, clicking close, release buttons, etc.
And we usually clone the repository again and remove some files using `rm -fr .buildinfo .doctrees .gradle *` to publish site. I haven't thought about it doing those stuff with a gradle task :) any idea?

anuraaga:
Ah I was mainly looking at step 5, which is many commands that could be run by Gradle instead (only git would be an exec command, and they could all be combined into one task).

Though it'd be possible to script everything too, pretty easily if we decided to use bintray as an intermediary - it can sync to maven central through it's API / Gradle plugin. Not quite low hanging fruit though :) https://github.com/bintray/gradle-bintray-plugin

minwoox:
I never know there's such thing :) Let's have a talk when trustin got back. Meanwhile this is [how we release](https://github.com/minwoox/misc/blob/master/armeria_release.md#uploading-to-the-staging-repository) FYI.

trustin:
Thanks, @minwoox !

Yeah, I think we can automate most of this process. Anybody interested? :wink:

PR #1292, opened 2018-07-16
Add SAML configurator in order to easily enable SAML authentication o…

hyangtack:
…n Armeria services

Motivation:
Nowdays a service doesn't need to provide a login service by itself because there are single sign-on specifications such as OAuth and SAML already.
So it would be better to support those specifications by Armeria so that a user can enable SSO on his or her services easily.

Modifications:
- Add saml module based on opensaml 3.x
- Add SamlServiceProvider class which helps a user enable SAML authentication for his or her services

Result:
Armeria provides SAML authentication for service providers.

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -8,6 +8,12 @@ ch.qos.logback:
     javadocs:
     - https://logback.qos.ch/apidocs/
 
+com.auth0:
+  java-jwt:
+    version: '3.4.0'
+    javadocs:
+    - https://www.javadoc.io/doc/com.auth0/java-jwt/3.4.0/
+
 com.fasterxml.jackson.core:
   jackson-annotations:
     version: &JACKSON_VERSION '2.9.6'
@@ -203,6 +209,9 @@ net.javacrumbs.json-unit:
 net.sf.proguard:
   proguard-gradle: { version: '6.0.3' }
 
+net.shibboleth.utilities:
+  java-support: { version: '7.3.0' }
+
 org.apache.curator:
   curator-recipes:
     version: '4.0.1'
@@ -309,6 +318,9 @@ org.jctools:
     - from: org.jctools
       to: com.linecorp.armeria.internal.shaded.jctools
 
+org.jsoup:
+  jsoup: { version: '1.11.3' }
+
 org.mockito:
   mockito-core: { version: '2.21.0' }
 
@@ -318,6 +330,15 @@ org.mortbay.jetty.alpn:
 org.openjdk.jmh:
   jmh-core: { version: &JMH_VERSION '1.21' }
   jmh-generator-annprocess: { version: *JMH_VERSION }
+
+org.opensaml:
+  opensaml-core: { version: &OPENSAML_VERSION '3.3.0' }
+  opensaml-saml-api: { version: *OPENSAML_VERSION }
+  opensaml-saml-impl: { version: *OPENSAML_VERSION }
+  opensaml-messaging-api: { version: *OPENSAML_VERSION }
+  opensaml-messaging-impl: { version: *OPENSAML_VERSION }
+  opensaml-soap-api: { version: *OPENSAML_VERSION }
+  opensaml-soap-impl: { version: *OPENSAML_VERSION }
 
 org.reactivestreams:
   reactive-streams: { version: &REACTIVE_STREAMS_VERSION '1.0.2' }
--- a/saml/src/main/java/com/linecorp/armeria/server/saml/HttpPostBindingUtil.java
+++ b/saml/src/main/java/com/linecorp/armeria/server/saml/HttpPostBindingUtil.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.saml;
+
+import static com.linecorp.armeria.server.saml.SamlHttpParameterNames.RELAY_STATE;
+import static com.linecorp.armeria.server.saml.SamlMessageUtil.deserialize;
+import static com.linecorp.armeria.server.saml.SamlMessageUtil.serialize;
+import static com.linecorp.armeria.server.saml.SamlMessageUtil.sign;
+import static java.util.Objects.requireNonNull;
+import static net.shibboleth.utilities.java.support.xml.SerializeSupport.nodeToString;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
+import javax.annotation.Nullable;
+
+import org.opensaml.messaging.context.MessageContext;
+import org.opensaml.saml.common.SAMLObject;
+import org.opensaml.saml.common.SignableSAMLObject;
+import org.opensaml.saml.common.messaging.context.SAMLBindingContext;
+import org.opensaml.security.credential.Credential;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.escape.Escaper;
+import com.google.common.html.HtmlEscapers;
+
+import com.linecorp.armeria.common.AggregatedHttpMessage;
+import com.linecorp.armeria.common.HttpData;
+import com.linecorp.armeria.server.saml.SamlService.SamlParameters;
+
+/**
+ * A utility class which supports HTTP POST binding protocol.
+ */
+final class HttpPostBindingUtil {
+
+    private static final ImmutableList<String> XHTML = ImmutableList.of(
+            // 0
+            "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
+            "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\" \"DTD/xhtml1-strict.dtd\">" +
+            "<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\">" +
+            "<head><meta http-equiv=\"content-type\" content=\"text/html;charset=utf-8\" /></head>" +
+            "<body onload=\"document.forms[0].submit()\">",
+            // 1, 2
+            "<form method=\"post\" action=\"", "\">",
+            // 3, 4, 5
+            "<input type=\"hidden\" name=\"", "\" value=\"", "\" />",
+            // 6
+            "</form></body></html>");
+
+    private static final Escaper HTML_ESCAPER = HtmlEscapers.htmlEscaper();
+
+    /**
+     * Returns an {@link HttpData} which holds a SSO form.
+     */
+    static HttpData getSsoForm(String remoteEndpointUrl,
+                               String paramName, String paramValue,
+                               @Nullable String relayState) {
+        requireNonNull(remoteEndpointUrl, "remoteEndpointUrl");
+        requireNonNull(paramName, "paramName");
+        requireNonNull(paramValue, "paramValue");
+
+        final StringBuilder html = new StringBuilder();
+        html.append(XHTML.get(0))
+            .append(XHTML.get(1)).append(HTML_ESCAPER.escape(remoteEndpointUrl)).append(XHTML.get(2))
+            .append(XHTML.get(3)).append(HTML_ESCAPER.escape(paramName))
+            .append(XHTML.get(4)).append(HTML_ESCAPER.escape(paramValue)).append(XHTML.get(5));
+
+        if (relayState != null) {
+            html.append(XHTML.get(3)).append(RELAY_STATE)
+                .append(XHTML.get(4)).append(HTML_ESCAPER.escape(relayState)).append(XHTML.get(5));
+        }
+        html.append(XHTML.get(6));
+
+        return HttpData.ofUtf8(html.toString());
+    }
+
+    /**
+     * Signs the specified {@link SignableSAMLObject} with the specified {@link Credential} and
+     * {@code signatureAlgorithm}, and then encodes the object into a base64 string.
+     */
+    static String toSignedBase64(SignableSAMLObject signableObj,
+                                 Credential signingCredential,
+                                 String signatureAlgorithm) throws SamlException {
+        sign(signableObj, signingCredential, signatureAlgorithm);
+        final String messageStr = nodeToString(serialize(signableObj));
+        return Base64.getEncoder().encodeToString(messageStr.getBytes(StandardCharsets.UTF_8));
+    }
+
+    /**
+     * Converts an {@link AggregatedHttpMessage} which is received from the remote entity to
+     * a {@link SAMLObject}.
+     */
+    static <T extends SAMLObject> MessageContext<T> toSamlObject(AggregatedHttpMessage msg,
+                                                                 String name) throws SamlException {
+        final SamlParameters parameters = new SamlParameters(msg);
+        final byte[] decoded;
+        try {
+            decoded = Base64.getDecoder().decode(parameters.getFirstValue(name));
+        } catch (IllegalArgumentException e) {
+            throw new SamlException("failed to decode a base64 string of the parameter: " + name, e);
+        }
+
+        @SuppressWarnings("unchecked")
+        final T message = (T) deserialize(decoded);
+
+        final MessageContext<T> messageContext = new MessageContext<>();
+        messageContext.setMessage(message);
+
+        final String relayState = parameters.getFirstValueOrNull(RELAY_STATE);
+        if (relayState != null) {
+            final SAMLBindingContext context = messageContext.getSubcontext(SAMLBindingContext.class, true);
+            assert context != null;
+            context.setRelayState(relayState);
+        }
+
+        return messageContext;
+    }
+
+    private HttpPostBindingUtil() {}
+}
--- a/saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java
+++ b/saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java
@@ -1,0 +1,290 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.saml;
+
+import static com.linecorp.armeria.server.saml.SamlHttpParameterNames.RELAY_STATE;
+import static com.linecorp.armeria.server.saml.SamlHttpParameterNames.SIGNATURE;
+import static com.linecorp.armeria.server.saml.SamlHttpParameterNames.SIGNATURE_ALGORITHM;
+import static java.util.Objects.requireNonNull;
+import static net.shibboleth.utilities.java.support.xml.SerializeSupport.nodeToString;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Map;
+import java.util.zip.Deflater;
+import java.util.zip.DeflaterOutputStream;
+import java.util.zip.Inflater;
+import java.util.zip.InflaterOutputStream;
+
+import javax.annotation.Nullable;
+
+import org.opensaml.core.xml.XMLObject;
+import org.opensaml.core.xml.io.MarshallingException;
+import org.opensaml.core.xml.util.XMLObjectSupport;
+import org.opensaml.messaging.context.MessageContext;
+import org.opensaml.saml.common.SAMLObject;
+import org.opensaml.saml.common.messaging.context.SAMLBindingContext;
+import org.opensaml.saml.saml2.core.Issuer;
+import org.opensaml.saml.saml2.core.RequestAbstractType;
+import org.opensaml.saml.saml2.core.StatusResponseType;
+import org.opensaml.security.SecurityException;
+import org.opensaml.security.credential.Credential;
+import org.opensaml.xmlsec.crypto.XMLSigningUtil;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import com.linecorp.armeria.common.AggregatedHttpMessage;
+import com.linecorp.armeria.common.HttpHeaderNames;
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.server.saml.SamlService.SamlParameters;
+
+import io.netty.handler.codec.http.HttpHeaderValues;
+import io.netty.handler.codec.http.QueryStringEncoder;
+
+/**
+ * A utility class which supports HTTP-Redirect binding protocol.
+ */
+final class HttpRedirectBindingUtil {
+
+    private static final String DEFAULT_CACHE_CONTROL =
+            String.join(",", HttpHeaderValues.NO_CACHE, HttpHeaderValues.NO_STORE);
+    private static final String DEFAULT_PRAGMA = HttpHeaderValues.NO_CACHE.toString();
+
+    /**
+     * Returns an {@link HttpHeaders} with the specified {@code location}, the default {@code cache-control}
+     * and the default {@code pragma} headers.
+     */
+    static HttpHeaders headersWithLocation(String location) {
+        return HttpHeaders.of(HttpStatus.FOUND)
+                          .add(HttpHeaderNames.LOCATION, location)
+                          .add(HttpHeaderNames.CACHE_CONTROL, DEFAULT_CACHE_CONTROL)
+                          .add(HttpHeaderNames.PRAGMA, DEFAULT_PRAGMA);
+    }
+
+    /**
+     * Returns an {@link HttpResponse} with the specified {@code location}, the default {@code cache-control}
+     * and the default {@code pragma} headers.
+     */
+    static HttpResponse responseWithLocation(String location) {
+        return HttpResponse.of(headersWithLocation(location));
+    }
+
+    /**
+     * Returns a redirected URL which includes a deflated base64 string that is converted from the specified
+     * {@link SAMLObject}. The URL must contain a signature of the generated query string.
+     */
+    static String toRedirectionUrl(SAMLObject msg,
+                                   String endpointUrl,
+                                   String messageParamName,
+                                   Credential signingCredential,
+                                   String signatureAlgorithm,
+                                   @Nullable String relayState) throws SamlException {
+        requireNonNull(msg, "msg");
+        requireNonNull(endpointUrl, "endpointUrl");
+        requireNonNull(messageParamName, "messageParamName");
+        requireNonNull(signingCredential, "signingCredential");
+        requireNonNull(signatureAlgorithm, "signatureAlgorithm");
+
+        final QueryStringEncoder encoder = new QueryStringEncoder("");
+        encoder.addParam(messageParamName, toDeflatedBase64(msg));
+
+        if (relayState != null) {
+            // RelayState data MAY be included with a SAML protocol message transmitted with this binding.
+            // The value MUST NOT exceed 80 bytes in length and SHOULD be integrity protected by the entity
+            // creating the message independent of any other protections that may or may not exist
+            // during message transmission.
+            if (relayState.length() > 80) {
+                throw new IllegalArgumentException("too long relayState string: " + relayState.length());
+            }
+            encoder.addParam(RELAY_STATE, relayState);
+        }
+
+        encoder.addParam(SIGNATURE_ALGORITHM, signatureAlgorithm);
+
+        // Use URL-encoded query string as input.
+        final String input = encoder.toString().substring(1);
+        final String output = generateSignature(signingCredential, signatureAlgorithm, input);
+        encoder.addParam(SIGNATURE, output);
+
+        return endpointUrl + encoder;
+    }
+
+    /**
+     * Validates a signature in the specified {@link AggregatedHttpMessage}.
+     */
+    private static void validateSignature(Credential validationCredential,
+                                          SamlParameters parameters,
+                                          String messageParamName) throws SamlException {
+        requireNonNull(validationCredential, "validationCredential");
+        requireNonNull(parameters, "parameters");
+        requireNonNull(messageParamName, "messageParamName");
+
+        final String signature = parameters.getFirstValue(SIGNATURE);
+        final String sigAlg = parameters.getFirstValue(SIGNATURE_ALGORITHM);
+
+        // The order is one of the followings:
+        // - SAMLRequest={value}&RelayState={value}=SigAlg={value}
+        // - SAMLResponse={value}&RelayState={value}=SigAlg={value}
+        final QueryStringEncoder encoder = new QueryStringEncoder("");
+        encoder.addParam(messageParamName, parameters.getFirstValue(messageParamName));
+
+        final String relayState = parameters.getFirstValueOrNull(RELAY_STATE);
+        if (relayState != null) {
+            encoder.addParam(RELAY_STATE, relayState);
+        }
+        encoder.addParam(SIGNATURE_ALGORITHM, sigAlg);
+
+        final byte[] input = encoder.toString().substring(1).getBytes(StandardCharsets.UTF_8);
+
+        try {
+            final byte[] decodedSignature = Base64.getDecoder().decode(signature);
+            if (!XMLSigningUtil.verifyWithURI(validationCredential, sigAlg, decodedSignature, input)) {
+                throw new SamlException("failed to validate a signature");
+            }
+        } catch (IllegalArgumentException e) {
+            throw new SamlException("failed to decode a base64 signature string", e);
+        } catch (SecurityException e) {
+            throw new SamlException("failed to validate a signature", e);
+        }
+    }
+
+    /**
+     * Generates a signature of the specified {@code input}.
+     */
+    @VisibleForTesting
+    static String generateSignature(Credential signingCredential,
+                                    String algorithmURI, String input) throws SamlException {
+        try {
+            final byte[] signature =
+                    XMLSigningUtil.signWithURI(signingCredential, algorithmURI,
+                                               input.getBytes(StandardCharsets.UTF_8));
+            return Base64.getEncoder().encodeToString(signature);
+        } catch (SecurityException e) {
+            throw new SamlException("failed to generate a signature", e);
+        }
+    }
+
+    /**
+     * Encodes the specified {@code message} into a deflated base64 string.
+     */
+    static String toDeflatedBase64(SAMLObject message) throws SamlException {
+        requireNonNull(message, "message");
+
+        final String messageStr;
+        try {
+            messageStr = nodeToString(XMLObjectSupport.marshall(message));
+        } catch (MarshallingException e) {
+            throw new SamlException("failed to serialize a SAML message", e);
+        }
+
+        final ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
+        try (DeflaterOutputStream deflaterStream =
+                     new DeflaterOutputStream(Base64.getEncoder().wrap(bytesOut),
+                                              new Deflater(Deflater.DEFLATED, true))) {
+            deflaterStream.write(messageStr.getBytes(StandardCharsets.UTF_8));
+        } catch (IOException e) {
+            throw new SamlException("failed to deflate a SAML message", e);
+        }
+        return bytesOut.toString();
+    }
+
+    /**
+     * Decodes, inflates and deserializes the specified base64-encoded message to an {@link XMLObject}.
+     */
+    static XMLObject fromDeflatedBase64(String base64Encoded) throws SamlException {
+        requireNonNull(base64Encoded, "base64Encoded");
+
+        final byte[] base64decoded;
+        try {
+            base64decoded = Base64.getDecoder().decode(base64Encoded);
+        } catch (IllegalArgumentException e) {
+            throw new SamlException("failed to decode a deflated base64 string", e);
+        }
+
+        final ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
+        try (InflaterOutputStream inflaterOutputStream =
+                     new InflaterOutputStream(bytesOut, new Inflater(true))) {
+            inflaterOutputStream.write(base64decoded);
+        } catch (IOException e) {
+            throw new SamlException("failed to inflate a SAML message", e);
+        }
+
+        return SamlMessageUtil.deserialize(bytesOut.toByteArray());
+    }
+
+    /**
+     * Converts an {@link AggregatedHttpMessage} which is received from the remote entity to
+     * a {@link SAMLObject}.
+     */
+    @SuppressWarnings("unchecked")
+    static <T extends SAMLObject> MessageContext<T> toSamlObject(
+            AggregatedHttpMessage msg, String name,
+            Map<String, SamlIdentityProviderConfig> idpConfigs,
+            @Nullable SamlIdentityProviderConfig defaultIdpConfig) throws SamlException {
+        requireNonNull(msg, "msg");
+        requireNonNull(name, "name");
+        requireNonNull(idpConfigs, "idpConfigs");
+
+        final SamlParameters parameters = new SamlParameters(msg);
+        final T message = (T) fromDeflatedBase64(parameters.getFirstValue(name));
+
+        final MessageContext<T> messageContext = new MessageContext<>();
+        messageContext.setMessage(message);
+
+        final Issuer issuer;
+        if (message instanceof RequestAbstractType) {
+            issuer = ((RequestAbstractType) message).getIssuer();
+        } else if (message instanceof StatusResponseType) {
+            issuer = ((StatusResponseType) message).getIssuer();
+        } else {
+            throw new SamlException("invalid message type: " + message.getClass().getSimpleName());
+        }
+
+        // Use the default identity provider config if there's no issuer.
+        final SamlIdentityProviderConfig config;
+        if (issuer != null) {
+            final String idpEntityId = issuer.getValue();
+            config = idpConfigs.get(idpEntityId);
+            if (config == null) {
+                throw new SamlException("a message from unknown identity provider: " + idpEntityId);
+            }
+        } else {
+            if (defaultIdpConfig == null) {
+                throw new SamlException("failed to get an Issuer element");
+            }
+            config = defaultIdpConfig;
+        }
+
+        // If this message is sent via HTTP-redirect binding protocol, its signature parameter should
+        // be validated.
+        validateSignature(config.signingCredential(), parameters, name);
+
+        final String relayState = parameters.getFirstValueOrNull(RELAY_STATE);
+        if (relayState != null) {
+            final SAMLBindingContext context = messageContext.getSubcontext(SAMLBindingContext.class, true);
+            assert context != null;
+            context.setRelayState(relayState);
+        }
+
+        return messageContext;
+    }
+
+    private HttpRedirectBindingUtil() {}
+}
--- a/saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
+++ b/saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.saml;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static io.netty.util.internal.MacAddressUtil.defaultMachineId;
+import static java.util.Objects.requireNonNull;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Base64;
+import java.util.Base64.Encoder;
+import java.util.Date;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.auth0.jwt.JWT;
+import com.auth0.jwt.JWTVerifier;
+import com.auth0.jwt.algorithms.Algorithm;
+import com.google.common.base.MoreObjects;
+
+import io.netty.util.internal.ThreadLocalRandom;
+
+/**
+ * A {@link SamlRequestIdManager} implementation based on JSON Web Tokens specification.
+ * See <a href="https://jwt.io/">JSON Web Tokens</a> for more information.
+ */
+final class JwtBasedSamlRequestIdManager implements SamlRequestIdManager {
+    private static final Logger logger = LoggerFactory.getLogger(JwtBasedSamlRequestIdManager.class);
+
+    private static final String CLAIM_NAME_UNIQUIFIER1 = "un1";
+    private static final String CLAIM_NAME_UNIQUIFIER2 = "un2";
+
+    private final String issuer;
+    private final Algorithm algorithm;
+    private final int validSeconds;
+    private final int leewaySeconds;
+
+    private final String un1;
+    private final JWTVerifier verifier;
+
+    JwtBasedSamlRequestIdManager(String issuer, Algorithm algorithm,
+                                 int validSeconds, int leewaySeconds) {
+        this.issuer = requireNonNull(issuer, "issuer");
+        this.algorithm = requireNonNull(algorithm, "algorithm");
+        this.validSeconds = validSeconds;
+        this.leewaySeconds = leewaySeconds;
+
+        checkArgument(validSeconds > 0,
+                      "invalid valid duration: " + validSeconds + " (expected: > 0)");
+        checkArgument(leewaySeconds >= 0,
+                      "invalid leeway duration:" + leewaySeconds + " (expected: >= 0)");
+
+        un1 = getUniquifierPrefix();
+        verifier = JWT.require(algorithm)
+                      .withIssuer(issuer)
+                      .acceptLeeway(leewaySeconds)
+                      .build();
+    }
+
+    @Override
+    public String newId() {
+        final Instant now = Instant.now();
+        final int un2 = ThreadLocalRandom.current().nextInt(Integer.MAX_VALUE) & 0x7fffffff;
+        return JWT.create()
+                  .withIssuer(issuer)
+                  .withIssuedAt(Date.from(now))
+                  .withExpiresAt(Date.from(now.plus(validSeconds, ChronoUnit.SECONDS)))
+                  // To make multiple tokens issued in the same second unique, we add uniquifiers.
+                  .withClaim(CLAIM_NAME_UNIQUIFIER1, un1)
+                  .withClaim(CLAIM_NAME_UNIQUIFIER2, un2)
+                  .sign(algorithm);
+    }
+
+    @Override
+    public boolean validateId(String id) {
+        requireNonNull(id, "id");
+        try {
+            // Verifier will check whether its issuer is me and also check whether it has been expired or not.
+            verifier.verify(id);
+            return true;
+        } catch (Throwable cause) {
+            logger.trace("JWT token validation failed", cause);
+            return false;
+        }
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                          .add("issuer", issuer)
+                          .add("algorithm", algorithm)
+                          .add("validSeconds", validSeconds)
+                          .add("leewaySeconds", leewaySeconds)
+                          .toString();
+    }
+
+    private static String getUniquifierPrefix() {
+        // To make a request ID globally unique, we will add MAC-based machine ID and a random number.
+        // The random number tries to make this instance unique in the same machine and process.
+        final byte[] r = new byte[6];
+        ThreadLocalRandom.current().nextBytes(r);
+        final Encoder encoder = Base64.getEncoder();
+        return new StringBuilder().append(encoder.encodeToString(defaultMachineId()))
+                                  .append(encoder.encodeToString(r))
+                                  .toString();
+    }
+}
--- a/saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
+++ b/saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.saml;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.opensaml.security.credential.CredentialResolver;
+import org.opensaml.security.credential.impl.KeyStoreCredentialResolver;
+
+/**
+ * A builder class which creates a new {@link KeyStoreCredentialResolver} instance.
+ */
+public final class KeyStoreCredentialResolverBuilder {
+
+    private final String path;
+    private final String keystorePassword;
+    private final Map<String, String> keyPasswords = new HashMap<>();
+
+    /**
+     * Creates a builder instance for {@link KeyStoreCredentialResolver}.
+     */
+    public KeyStoreCredentialResolverBuilder(String path, String keystorePassword) {
+        this.path = requireNonNull(path, "path");
+        this.keystorePassword = requireNonNull(keystorePassword, "keystorePassword");
+    }
+
+    /**
+     * Adds a key name and its password to the {@link KeyStoreCredentialResolverBuilder}.
+     */
+    public KeyStoreCredentialResolverBuilder addKeyPassword(String name, String password) {
+        requireNonNull(name, "name");
+        requireNonNull(password, "password");
+        checkArgument(!keyPasswords.containsKey(name), "key already exists: %s", name);
+        keyPasswords.put(name, password);
+        return this;
+    }
+
+    /**
+     * Creates a new {@link KeyStoreCredentialResolver}.
+     */
+    public CredentialResolver build() throws IOException, GeneralSecurityException {
+        final KeyStore ks = KeyStore.getInstance(KeyStore.getDefaultType());
+        try (InputStream is = KeyStoreCredentialResolverBuilder.class.getClassLoader()
+                                                                     .getResourceAsStream(path)) {
+            ks.load(is, keystorePassword.toCharArray());
+        }
+        return new KeyStoreCredentialResolver(ks, keyPasswords);
+    }
+}
--- a/saml/src/main/java/com/linecorp/armeria/server/saml/SamlAssertionConsumerConfig.java
+++ b/saml/src/main/java/com/linecorp/armeria/server/saml/SamlAssertionConsumerConfig.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.saml;
+
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.base.MoreObjects;
+
+/**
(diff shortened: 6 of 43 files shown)


saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+    public SamlServiceProviderConfigurator build() {
+        checkArgument(!authorizers.isEmpty(), "at least one authorizer should be specified");
+
+        if (credentialResolver == null) {
+            throw new IllegalArgumentException(CredentialResolver.class.getSimpleName() + " is not specifed");
+        }
+
+        // If there is only one ACS, it will be automatically a default ACS.
+        if (acsBuilders.size() == 1) {
+            Iterables.getFirst(acsBuilders, null).asDefault();
```
trustin: `acsBuilders.get(0).asDefault()`?

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+                                                          .collect(Collectors.toSet());
+        checkArgument(acsEndpoints.size() == acsBuilders.size(),
+                      "duplicated access consumer services");
+
+        final ImmutableMap.Builder<String, IdentityProviderConfig> profiles = ImmutableMap.builder();
+        IdentityProviderConfig defaultIdentityProviderConfig = null;
+        for (final IdentityProviderConfigBuilder builder : profileBuilders) {
+            if (builder.acsEndpoint != null) {
+                checkArgument(acsEndpoints.contains(builder.acsEndpoint),
+                              "unspecified access consumer service at " + builder.acsEndpoint);
```
trustin: `... at %s", builder.acsEndpoint`

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+
+            validateSignatureAlgorithm(signatureAlgorithm, config.signingCredential());
+            validateSignatureAlgorithm(signatureAlgorithm, config.encryptionCredential());
+
+            profiles.put(config.entityId(), config);
+
+            // If there is only one idp, it will be automatically a default idp.
+            if (builder.isDefault || profileBuilders.size() == 1) {
+                checkArgument(defaultIdentityProviderConfig == null,
+                              "too many default on identity provider configuration");
```
trustin: How about: `There has to be only one default identity provider.`

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+                defaultIdentityProviderConfig = config;
+            }
+        }
+
+        if (idpConfigSelector == null) {
+            checkArgument(defaultIdentityProviderConfig != null,
+                          "default identity provider configuration should exist");
```
trustin: How about: `Default identity provider does not exist.`

saml/src/main/java/com/linecorp/armeria/saml/SamlSingleLogoutFunction.java
```diff
@@ -0,0 +1,166 @@
...
+                } else {
+                    final String value = signAndBase64Encode(logoutResponse, signingCredential,
+                                                             signatureAlgorithm);
+                    final HttpData body = getHtmlForm(sloResEndpoint.urlAsString(),
+                                                      PARAM_SAML_RESPONSE, value, null);
+                    return HttpResponse.of(HttpStatus.OK, MediaType.HTML_UTF_8, body);
+                }
+            }));
+        } catch (SAMLException e) {
+            return HttpResponse.of(HttpStatus.BAD_REQUEST);
```
trustin: logging?

saml/src/main/java/com/linecorp/armeria/saml/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * A package which is used to enable SAML authentication on a server instance.
```
trustin: How about just `SAML-based single sign-on service.`?

saml/src/main/java/com/linecorp/armeria/saml/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+@NonNullByDefault
+package com.linecorp.armeria.saml;
```
trustin: `com.linecorp.armeria.server.saml`?

saml/src/test/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorTest.java
```diff
@@ -0,0 +1,546 @@
...
+                final int port = ss.getLocalPort();
+                sb.port(port, SessionProtocol.HTTP);
+
+                final SamlServiceProviderConfigurator authConfigurator = SamlServiceProviderConfigurator
+                        .builder()
+                        // A request will be authenticated if it contains 'test=test' cookie in Cookie header.
+                        .addAuthorizer(new CookieBasedAuthorizer("test", "test"))
+                        .authManager(new CookieBasedAuthManager("test", "test"))
+                        // My entity ID
+                        .entityId("http://127.0.0.1:" + port)
```
trustin: All these URLs could be auto-filled:

- scheme could be detected automatically between `http` and `https` from the active `ServerPort`s.
- hostname could be detected automatically from the `VirtualHost` where the SAML service/decorator is added.
- port could be detected automatically from the active `ServerPort`s.
- All path could have their own default values so that a user doesn't have to specify them.
- Both scheme and hostname could be overridden if a user does not like the auto-detected values.

saml/src/test/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorTest.java
```diff
@@ -0,0 +1,546 @@
...
+
+        @Override
+        public HttpResponse onFailure(ServiceRequestContext ctx, Throwable cause) {
+            // Handle as an error so that a test client can detect the failure.
+            return HttpResponse.of(HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    static class RandomRequestIdManager implements SamlRequestIdManager {
+        private final RandomIdentifierGenerationStrategy generator = new RandomIdentifierGenerationStrategy();
```
trustin: This uses `SecureRandom`, which may hang.

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfigurator.java
```diff
@@ -0,0 +1,316 @@
...
+import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.Service;
+import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.ServiceWithPathMappings;
+
+/**
+ * A simple service provider configurator. It helps a user configure SAML authentication for his or her
+ * services easily.
+ */
+public final class SamlServiceProviderConfigurator {
```
trustin: How about just `SamlServiceProvider` and `SamlServiceProviderBuilder`?

```java
SamlServiceProvider sp = ...;
sp.newDecorator();
sp.newService();
```

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfigurator.java
```diff
@@ -0,0 +1,316 @@
...
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.Service;
+import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.ServiceWithPathMappings;
+
+/**
+ * A simple service provider configurator. It helps a user configure SAML authentication for his or her
+ * services easily.
```
trustin: How about:

```java
/**
 * A SAML service provider implementation. See <a href="http://line.github.io/armeria/...">...</a> for more information.
```

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+        }
+
+        final Credential signingCredential = resolveCredential(signingKey);
+        final Credential encryptionCredential = resolveCredential(encryptionKey);
+        validateSignatureAlgorithm(signatureAlgorithm, signingCredential);
+        validateSignatureAlgorithm(signatureAlgorithm, encryptionCredential);
+
+        // Collect assertion consumer service endpoints for checking duplication and existence.
+        final Set<SamlEndpoint> acsEndpoints = acsBuilders.stream()
+                                                          .map(b -> b.endpoint)
```
trustin: `SamlEndpoint::endpoint`?

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+                                                          .collect(Collectors.toSet());
```
trustin: `toImmutableSet()`?

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+        try {
+            final Signature signature = Signature.getInstance(jcaAlgorithmID);
+            final PrivateKey key = credential.getPrivateKey();
+            if (key != null) {
+                signature.initSign(key);
+            } else {
+                signature.initVerify(credential.getPublicKey());
+            }
+        } catch (Exception e) {
+            throw new IllegalArgumentException("unsupported signature algorithm", e);
```
trustin: Could we make the exception message more detailed? e.g.

- `unsupported signature algorithm: <signatureAlgorithm>` if `Signature.getInstance()` fails.
- `failed to initialize a signature` if `initSign()` or `initVerify()` fails.

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+        }
+    }
+
+    public class IdentityProviderConfigBuilder {
```
trustin: - Make the default constructor package-local so that a user can't call it directly.
- How about making top-level and prepending `Saml`?

saml/src/main/java/com/linecorp/armeria/saml/SamlServiceProviderConfiguratorBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+
+        /**
+         * Returns a {@link SamlServiceProviderConfigurator} which is the parent of this builder.
+         */
+        public SamlServiceProviderConfiguratorBuilder and() {
+            return SamlServiceProviderConfiguratorBuilder.this;
+        }
+    }
+
+    public class AssertionConsumerConfigBuilder {
```
trustin: - The default constructor has to be package-local.
- Top-level and `Saml` prefix?

saml/src/main/java/com/linecorp/armeria/saml/HttpPostBindingUtil.java
```diff
@@ -0,0 +1,142 @@
...
+            return Exceptions.throwUnsafely(
+                    new MessageEncodingException("Unable to Base64 encode SAML message", e));
+        }
+    }
+
+    /**
+     * Converts an {@link AggregatedHttpMessage} which is received from the remote entity to
+     * a {@link SAMLObject}.
+     */
+    @SuppressWarnings("unchecked")
```
anuraaga: I think this may only apply to one line so maybe better to put it there

saml/src/main/java/com/linecorp/armeria/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,292 @@
...
+        requireNonNull(signingCredential, "signingCredential");
+        requireNonNull(signatureAlgorithm, "signatureAlgorithm");
+
+        try {
+            final QueryStringEncoder encoder = new QueryStringEncoder(endpointUrl);
+
+            final String samlRequest = deflateAndBase64Encode(msg);
+            encoder.addParam(messageParamName, samlRequest);
+
+            if (relayState != null) {
```
anuraaga: How about `checkArgument(relayState == null || relayState.length() <= 80)`. Do you want to ignore this too long case?

saml/src/main/java/com/linecorp/armeria/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,292 @@
...
+                if (relayState.length() > 80) {
+                    logger.warn("too long relayState string: " + relayState.length());
+                }
+                encoder.addParam(PARAM_RELAY_STATE, relayState);
+            }
+
+            encoder.addParam(PARAM_SIGNATURE_ALGORITHM, signatureAlgorithm);
+
+            // Use URL-encoded query string as input.
+            final String input = encoder.toUri().getRawQuery();
```
anuraaga: I think you can use `encoder.toString()` (maybe need substring if there's a `?` but still much more efficient than going through the URI parser)

saml/src/main/java/com/linecorp/armeria/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,292 @@
...
+        encoder.addParam(messageParamName, parameters.get(messageParamName).get(0));
+
+        final List<String> relayState = parameters.get(PARAM_RELAY_STATE);
+        if (relayState != null && !relayState.isEmpty()) {
+            encoder.addParam(PARAM_RELAY_STATE, relayState.get(0));
+        }
+        encoder.addParam(PARAM_SIGNATURE_ALGORITHM, sigAlg.get(0));
+
+        try {
+            final byte[] input = encoder.toUri().getRawQuery().getBytes(StandardCharsets.UTF_8);
```
anuraaga: Ditto

saml/src/main/java/com/linecorp/armeria/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * Marshals, deflates and encodes the specified {@code message} to base64-encoded form.
+     */
+    static String deflateAndBase64Encode(SAMLObject message) {
+        try {
+            final String messageStr = nodeToString(XMLObjectSupport.marshall(message));
+            final ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
+            final Deflater deflater = new Deflater(Deflater.DEFLATED, true);
+            final DeflaterOutputStream deflaterStream = new DeflaterOutputStream(bytesOut, deflater);
```
anuraaga: I think instead of manually finishing, you can do

```java
try (DeflaterOutputStream deflaterStream = ...) {
  stream.write();
}
```

saml/src/main/java/com/linecorp/armeria/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,292 @@
...
+    /**
+     * Decodes, inflates and unmarshals the specified base64-encoded message to a {@link XMLObject}.
+     */
+    static XMLObject inflateAndBase64Decode(String base64Encoded) {
+        try {
+            final byte[] base64decoded = Base64Support.decode(base64Encoded);
+
+            final ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
+            final Inflater inflater = new Inflater(true);
+            final InflaterOutputStream inflaterOutputStream = new InflaterOutputStream(bytesOut, inflater);
```
anuraaga: Ditto

saml/src/main/java/com/linecorp/armeria/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,292 @@
...
+    static String deflateAndBase64Encode(SAMLObject message) {
+        try {
+            final String messageStr = nodeToString(XMLObjectSupport.marshall(message));
+            final ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
+            final Deflater deflater = new Deflater(Deflater.DEFLATED, true);
+            final DeflaterOutputStream deflaterStream = new DeflaterOutputStream(bytesOut, deflater);
+            deflaterStream.write(messageStr.getBytes(StandardCharsets.UTF_8));
+            deflaterStream.finish();
+
+            return Base64Support.encode(bytesOut.toByteArray(), Base64Support.UNCHUNKED);
```
anuraaga: As I asked before, I'm not too sure if this `Base64Support` is special, but if you can use JDK, you can use `Base64.Encoder/Decoder.wrap`.

saml/src/main/java/com/linecorp/armeria/saml/JwtBasedSamlRequestIdManager.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * A {@link SamlRequestIdManager} implementation based on JSON Web Tokens specification.
+ * See <a href="https://jwt.io/">JSON Web Tokens</a> for more information.
+ */
+final class JwtBasedSamlRequestIdManager implements SamlRequestIdManager {
+
+    private static final String CLAIM_NAME_SALT = "salt";
+
+    private final String issuer;
+    private final int validSeconds;
```
anuraaga: Since we don't need to worry about allocation overhead, consider using `Duration`

saml/src/main/java/com/linecorp/armeria/saml/JwtBasedSamlRequestIdManager.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public String generateNewId() {
+        final Instant now = Instant.now();
+        return JWT.create()
+                  .withIssuer(issuer)
+                  .withIssuedAt(Date.from(now))
+                  .withExpiresAt(Date.from(now.plus(validSeconds, ChronoUnit.SECONDS)))
+                  .withClaim(CLAIM_NAME_SALT, ThreadLocalRandom.current().nextInt(Integer.MAX_VALUE))
```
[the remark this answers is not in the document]
anuraaga: Is it ok not to use `SecureRandom`?

saml/src/main/java/com/linecorp/armeria/saml/JwtBasedSamlRequestIdManager.java
```diff
@@ -0,0 +1,101 @@
...
+                  .sign(algorithm);
+    }
+
+    @Override
+    public boolean validateId(String id) {
+        requireNonNull(id, "id");
+        try {
+            // Verifier will check whether its issuer is me and also check whether it has been expired or not.
+            final DecodedJWT jwt = verifier.verify(id);
+            return !jwt.getClaim(CLAIM_NAME_SALT).isNull();
```
anuraaga: What does this salt do? If we don't actually compare it to anything, it doesn't seem to add any checking.

saml/src/main/java/com/linecorp/armeria/saml/HttpPostBindingUtil.java
```diff
@@ -0,0 +1,142 @@
...
+     * {@code signatureAlgorithm}, and then encode the object as base64 format.
+     */
+    static String signAndBase64Encode(SignableSAMLObject signableObj,
+                                      Credential signingCredential,
+                                      String signatureAlgorithm) {
+        sign(signableObj, signingCredential, signatureAlgorithm);
+        try {
+            final String messageStr = nodeToString(XMLObjectSupport.marshall(signableObj));
+            return Base64Support.encode(messageStr.getBytes(), Base64Support.UNCHUNKED);
+        } catch (Exception e) {
```
anuraaga: Global comment, I think the code would be easier to reason about if we use multicatch instead of `Exception/Throwable` throughout.

saml/src/main/java/com/linecorp/armeria/saml/SamlAssertionConsumerFunction.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        // The <Response> element MUST conform to the following:
+        //
+        // - The <Issuer> element MAY be omitted, but if present it MUST contain the unique identifier
+        //   of the issuing identity provider; the Format attribute MUST be omitted or have a value of
+        //   urn:oasis:names:tc:SAML:2.0:nameid-format:entity.
+        final Issuer issuer = response.getIssuer();
+        if (issuer == null || issuer.getValue() == null) {
+            throw new SAMLException("no issuer");
```
anuraaga: I may not be understanding the comment above, but is it ok to throw an exception even though `<Issuer> element MAY be omitted`?

saml/src/main/java/com/linecorp/armeria/saml/SamlAssertionConsumerFunction.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        validateSignature(idp.signingCredential(), response);
+
+        final DateTime now = new DateTime();
+        final DateTime issueInstant = response.getIssueInstant();
+        if (issueInstant == null) {
+            throw new SAMLException("no issue instant");
+        }
+        if (Math.abs(now.getMillis() - issueInstant.getMillis()) > 60_000) {
```
anuraaga: Would extract `MILLIS_IN_MINUTE = TimeUnit.MINUTE.toMillis(1)`

saml/src/main/java/com/linecorp/armeria/saml/SamlDecorator.java
```diff
@@ -0,0 +1,155 @@
...
+                                    signingCredential, configurator.signatureAlgorithm(), relayState));
+                        } else {
+                            final String value = signAndBase64Encode(request, signingCredential,
+                                                                     configurator.signatureAlgorithm());
+                            final HttpData body = getHtmlForm(endpoint.urlAsString(),
+                                                              PARAM_SAML_REQUEST, value, relayState);
+                            return HttpResponse.of(HttpStatus.OK, MediaType.HTML_UTF_8, body);
+                        }
+                    }));
+        }, ctx.blockingTaskExecutor()));
```
anuraaga: I don't think we're supposed to force `delegate().serve` onto the blocking executor.

saml/src/main/java/com/linecorp/armeria/saml/SamlSessionManager.java
```diff
@@ -0,0 +1,50 @@
...
+import org.opensaml.saml.saml2.core.LogoutRequest;
+import org.opensaml.saml.saml2.core.Response;
+
+/**
+ * An interface which is called back when a user is logged in or out.
+ */
+public interface SamlSessionManager {
+
+    /**
+     * A {@link SamlSessionManager} which stores a session into a memory.
```
anuraaga: Perhaps we could add something on when this would be used? I guess it's only for testing?

hyangtack:
@trustin @anuraaga Thanks to review. I'm in the middle of refactoring based on your comments. I'll let you know when this PR is ready for review again. :-)

saml/src/main/java/com/linecorp/armeria/saml/HttpRedirectBindingUtil.java
hyangtack: From SAML specification, it must not exceed 80 bytes in length, but I found that OpenSAML library allows a string which exceeds 80 bytes in length. So I just added a log message here. 

On second thought, it would be better to follow the specification, so let me fix the condition as what you told.

Thanks!

saml/src/main/java/com/linecorp/armeria/saml/JwtBasedSamlRequestIdManager.java
hyangtack: I wanted to use `Duration` but JWT library is based on `seconds`. So `Duration` may be confused to our user. It may look like that we can support `milliseconds` for expiration.

saml/src/main/java/com/linecorp/armeria/saml/JwtBasedSamlRequestIdManager.java
hyangtack: As I said [here](https://github.com/line/armeria/pull/1292#discussion_r203948109), JWT library is based on `seconds`. It means that it generates same IDs in a seconds because we have the same `issuer` and time values in a second. So the `salt` makes the ID different for every time. 

We can do that as simply adding a sequentially increasing number as a claim. Is it better? Let me know what you think. :-)

saml/src/main/java/com/linecorp/armeria/saml/SamlAssertionConsumerFunction.java
hyangtack: It was my fault. Thank you!

saml/src/main/java/com/linecorp/armeria/saml/SamlSessionManager.java
hyangtack: Sorry for not enough explanation. I'm trying to figure out how to integrate this SAML module with Central Dogma. So I think I can find how can we use the session manager while the integration is going on. For now, let me move the default session manager to the test code.

saml/src/main/java/com/linecorp/armeria/saml/SamlDecorator.java
hyangtack: Good point. Will not use the blocking executor here. Thanks!

saml/src/main/java/com/linecorp/armeria/server/saml/SamlRequestIdManager.java
```diff
@@ -0,0 +1,69 @@
...
+     * Returns a {@link SamlRequestIdManager} implementation based on JSON Web Tokens specification with
+     * the {@link Algorithm} instance using {@code HmacSHA384}.
+     *
+     * @param issuer the ID who issues a token
+     * @param secret the secret which is used to generate a signature
+     * @param validSeconds the valid period of a token in seconds
+     * @param leewaySeconds the leeway when there is a clock skew times between the signer and the verifier,
+     *                      in seconds.
+     */
+    static SamlRequestIdManager ofJwtBased(
```
trustin: How about just `ofJwt`?

saml/src/main/java/com/linecorp/armeria/server/saml/SamlServiceProviderBuilder.java
```diff
@@ -0,0 +1,479 @@
...
+        }
+
+        final Credential signingCredential = credentialResolver.apply(signingKey);
+        if (signingCredential == null) {
+            throw new IllegalArgumentException("cannot resolve a " + Credential.class.getSimpleName() +
+                                               " for signing: " + signingKey);
+        }
+        final Credential encryptionCredential = credentialResolver.apply(encryptionKey);
+        if (encryptionCredential == null) {
+            throw new IllegalArgumentException("cannot resolve a " + Credential.class.getSimpleName() +
```
trustin: Use `IllegalStateException` instead of `IllegalArgumentException` in `build()`.

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+    public CredentialResolver build() {
+        try {
+            final KeyStore ks = KeyStore.getInstance(KeyStore.getDefaultType());
+            try (InputStream is =
+                         KeyStoreCredentialResolverBuilder.class.getClassLoader().getResourceAsStream(path)) {
+                ks.load(is, password.toCharArray());
+            }
+            return new KeyStoreCredentialResolver(ks, keyPasswords);
+        } catch (Throwable cause) {
+            return Exceptions.throwUnsafely(cause);
```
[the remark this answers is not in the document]
hyangtack: Will make `build()` throws `GeneralSecurityException` and `IOException`. 
`KeyStoreException`, `CertificateException`, `NoSuchAlgorithmException` and `IOException` are thrown in the `build()` method, and the first three exceptions are subclasses of `GeneralSecurityException`.

saml/src/main/java/com/linecorp/armeria/server/saml/SamlAssertionConsumerConfig.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public boolean isDefault() {
+        return isDefault;
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                          .add("endpoint", endpoint)
+                          .add("isDefault", isDefault)
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlAssertionConsumerFunction.java
```diff
@@ -0,0 +1,278 @@
...
+                                              Credential decryptionCredential) throws SamlException {
+        final StaticKeyInfoCredentialResolver keyInfoCredentialResolver =
+                new StaticKeyInfoCredentialResolver(decryptionCredential);
+        final Decrypter decrypter =
+                new Decrypter(null, keyInfoCredentialResolver, new InlineEncryptedKeyResolver());
+        decrypter.setRootInNewDocument(true);
+        try {
+            return decrypter.decrypt(encryptedAssertion);
+        } catch (DecryptionException e) {
+            throw new SamlException("failed to decrypt assertion", e);
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlBindingProtocol.java
```diff
@@ -0,0 +1,45 @@
...
+    HTTP_POST("urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST");
+
+    private final String id;
+
+    SamlBindingProtocol(String id) {
+        this.id = id;
+    }
+
+    /**
+     * Returns the identification of this binding protocol.
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlBindingProtocol.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public String id() {
+        return id;
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlDecorator.java
```diff
@@ -0,0 +1,207 @@
...
+                                signingCredential, sp.signatureAlgorithm(), relayState));
+                    } else {
+                        final String value = toSignedBase64(request, signingCredential,
+                                                            sp.signatureAlgorithm());
+                        final HttpData body = getHtmlForm(endpoint.toUriString(),
+                                                          SAML_REQUEST, value, relayState);
+                        return HttpResponse.of(HttpStatus.OK, MediaType.HTML_UTF_8, body);
+                    }
+                } catch (SamlException e) {
+                    return HttpResponse.of(HttpStatus.BAD_REQUEST);
```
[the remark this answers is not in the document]
hyangtack: Added as TRACE level.

saml/src/main/java/com/linecorp/armeria/server/saml/SamlEndpoint.java
```diff
@@ -0,0 +1,146 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof SamlEndpoint)) {
+            return false;
+        }
+        final SamlEndpoint that = (SamlEndpoint) o;
+        return uri().equals(that.uri) && bindingProtocol() == that.bindingProtocol;
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlException.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private static final long serialVersionUID = -6694912876733624005L;
+
+    public SamlException(String message) {
+        super(message);
+    }
+
+    public SamlException(String message, Throwable cause) {
+        super(message, cause);
+    }
```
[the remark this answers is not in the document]
hyangtack: Added

saml/src/main/java/com/linecorp/armeria/server/saml/SamlIdentityProviderConfigBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+     */
+    public SamlServiceProviderBuilder and() {
+        return parent;
+    }
+
+    /**
+     * Builds a {@link SamlIdentityProviderConfig}.
+     */
+    SamlIdentityProviderConfig build(CredentialResolverAdapter credentialResolver) {
+        checkArgument(entityId != null, "IdP entity ID is missing");
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlInitializer.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Returns {@code true} if and only if the OpenSAML library is available.
+     */
+    static boolean isAvailable() {
+        return UNAVAILABILITY_CAUSE == null;
+    }
+
+    /**
+     * Ensure that the OpenSAML library is available.
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlInitializer.java
```diff
@@ -0,0 +1,80 @@
...
+    static void ensureAvailability() {
+        if (UNAVAILABILITY_CAUSE != null) {
+            throw new Error("failed to initialize OpenSAML library", UNAVAILABILITY_CAUSE);
+        }
+    }
+
+    /**
+     * Returns the cause of unavailability of the OpenSAML library.
+     */
+    static Throwable unavailabilityCause() {
```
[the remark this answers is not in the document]
hyangtack: Added

saml/src/main/java/com/linecorp/armeria/server/saml/SamlNameIdFormat.java
```diff
@@ -0,0 +1,79 @@
...
+
+    SamlNameIdFormat(String id) {
+        this.id = id;
+    }
+
+    /**
+     * Returns the identification of this name ID format.
+     */
+    public String id() {
+        return id;
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlRequestIdManager.java
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlServiceProviderBuilder.java
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlService.java
```diff
@@ -0,0 +1,171 @@
...
+            final SamlPortConfig portConfig = portConfigHolder.config().get();
+            if (portConfig.scheme().isTls() != ctx.sessionProtocol().isTls()) {
+                return HttpResponse.of(HttpStatus.BAD_REQUEST);
+            }
+
+            // Use user-specified hostname if it exists.
+            // If there's no hostname set by a user, the default virtual hostname will be used.
+            final String defaultHostname = firstNonNull(sp.hostname(), ctx.virtualHost().defaultHostname());
+            return func.serve(ctx, msg, defaultHostname, portConfig);
+        }, ctx.blockingTaskExecutor()));
```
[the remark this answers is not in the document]
hyangtack: Sorry for missing comments. I also wanted to prevent event loops from a user's blocking calls in the user handlers, such as `SamlSingleSignOnHandler` and `SamlSingleLogoutHandler`. 

I also talked with @trustin about this, then we make every callback methods return asynchronous like `CompletionStage<Void>` instead of `void`, in order to lower the possibility of blocking event loop.

Thanks.

saml/src/main/java/com/linecorp/armeria/server/saml/SamlSingleLogoutHandler.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Invoked when the single logout request is succeeded. It can do the local logout using session indexes
+     * containing in the {@link LogoutRequest}.
+     *
+     * @param ctx the {@link ServiceRequestContext} of {@code req}
+     * @param req the {@link AggregatedHttpMessage} being handled
+     * @param message the {@link MessageContext} of the {@link LogoutRequest} received from the identity
+     *                provider.
+     */
+    void sloSucceeded(ServiceRequestContext ctx, AggregatedHttpMessage req,
```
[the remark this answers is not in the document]
hyangtack: Ok, will change. :-)

saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
```diff
@@ -0,0 +1,99 @@
...
+                  .sign(algorithm);
+    }
+
+    @Override
+    public boolean validateId(String id) {
+        requireNonNull(id, "id");
+        try {
+            // Verifier will check whether its issuer is me and also check whether it has been expired or not.
+            final DecodedJWT jwt = verifier.verify(id);
+            return !jwt.getClaim(CLAIM_NAME_SALT).isNull();
```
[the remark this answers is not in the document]
hyangtack: Let's assume that there's no `salt` claim. In this case, we have 3 inputs as follows:
- issuer (string)
- issuedAt (datetime, in seconds)
- expiredAt (datetime, in seconds)

If the 3 inputs are not changed, JWT library produces the same output every time. But, as you see, the 2 time values are in **seconds**, not a **us** or **ms**. So it's possible to produce the same JWT library tokens in a second.

To make JWT library generates different tokens every time, I added the `salt` claim. And I used to random numbers to guess a little harder, but actually it doesn't have no more meaning than making different tokens. So, I just checked only whether the `salt` claim exists here since the token has been validated by `verifier`.

Let me know if you have a concern. :-)

saml/src/main/java/com/linecorp/armeria/server/saml/HttpPostBindingUtil.java
```diff
@@ -0,0 +1,127 @@
...
+        return Base64.getEncoder().encodeToString(messageStr.getBytes(StandardCharsets.UTF_8));
+    }
+
+    /**
+     * Converts an {@link AggregatedHttpMessage} which is received from the remote entity to
+     * a {@link SAMLObject}.
+     */
+    static <T extends SAMLObject> MessageContext<T> toSamlObject(AggregatedHttpMessage msg,
+                                                                 String name) throws SamlException {
+        final Map<String, List<String>> parameters = SamlService.getParameters(msg);
```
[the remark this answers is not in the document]
anuraaga: I'm not sure but otherwise we can define a static utility to replace parameters.get that returns the nullable value instead of list, to avoid the same null or empty check over and over.

Actually I've written a similar function in several codebases I've added armeria to, having it in core might be nice :)

saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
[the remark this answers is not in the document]
hyangtack: > is this a convention in SSO?

There's no convention about the ID in SSO. We need to **generate a new ID** when sending a new authentication request to an IdP(identity provider), and also we need to **validate the ID** when receiving a response from the IdP.

As you might know, SAML request/response is different from the general client-server request/response. They are independent HTTP request(GET or POST). So, in general, we can put the generated ID into the global session storage then validate the received ID by comparing to the generated one. But we want to validate the ID regardless of the global session storage, so I used JWT as an ID of a SAML message.

> Another approach would be to add a claim of id and fill it with UUID and use it rather than the whole token

Yes, it's possible. But I think that it doesn't have an advantage for achieving our goal I explained the above.

> IIUC since the signature is verified, actually there is no extra information from checking the claim is not null.

Yes, you're right. I just added the code for double checking. Let me remove it.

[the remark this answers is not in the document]
anuraaga: Cool makes sense. I like avoiding security checks that don't provide extra information for easier reasoning (I remember one code I've seen that encrypted, hashed, then encrypted again which made it harder to realize that none of the steps used a salt...).

saml/src/main/java/com/linecorp/armeria/server/saml/HttpPostBindingUtil.java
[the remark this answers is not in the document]
hyangtack: I added a simple wrapper class which can resolve what you pointed out. :-)

https://github.com/line/armeria/pull/1292/files#diff-6cb4c52b1669f6f34a67773b653f3585R160

We may have a better utility class in `core` in the future. At that time, we can replace the current implementation with new one. :-)

saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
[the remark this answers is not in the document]
hyangtack: To make clear, I renamed the claim from `salt` to `un1`/`un2` which means `uniquifier`. (@trustin 's suggestion)

`un1` is pre-computed when the class is instantiated, using a MAC address and random bytes. It helps to make our `request ID`(a.k.a JWT token) unique in a single machine.

`un2` is the same as the previous `salt` which is generated randomly for every request ID in order to make the ID unique in a second.

hyangtack:
Seems that I addressed all comments from @trustin and @anuraaga. Would you review again when you have some time? :-)

saml/src/main/java/com/linecorp/armeria/server/saml/SamlService.java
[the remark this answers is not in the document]
hyangtack: Oops! I missed fixing here. 😱 Anyway, the reason why I used `blockingTaskExecutor()` here was due to worry about slow XML handling and user's blocking calls in user handlers. Will replace `handleAsync` with `handle`. Thanks!

saml/src/main/java/com/linecorp/armeria/server/saml/SamlServiceProvider.java
```diff
@@ -0,0 +1,256 @@
...
+                        SamlSingleLogoutHandler sloHandler) {
+        this.authorizer = requireNonNull(authorizer, "authorizer");
+        this.entityId = requireNonNull(entityId, "entityId");
+        this.hostname = hostname;
+        this.signingCredential = requireNonNull(signingCredential, "signingCredential");
+        this.encryptionCredential = requireNonNull(encryptionCredential, "encryptionCredential");
+        this.signatureAlgorithm = requireNonNull(signatureAlgorithm, "signatureAlgorithm");
+        this.portConfigAutoFiller = requireNonNull(portConfigAutoFiller, "portConfigAutoFiller");
+        this.metadataPath = PathMapping.ofExact(requireNonNull(metadataPath, "metadataPath"));
+        this.idpConfigs = requireNonNull(idpConfigs, "idpConfigs");
```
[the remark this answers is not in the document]
hyangtack: It did in `SamlServiceProviderBuilder#build` so I didn't copy any collections here. Will check again if there's missing one.

saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private static String getUniquifierPrefix() {
+        // The following table is used just once when instantiating this object, so make it local variable.
+        final char[] digits =
+                { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f' };
+        // To make a request ID globally unique, we will add MAC-based machine ID and a random number.
+        // The random number tries to make this instance unique in the same machine and process.
+        final byte[] r = new byte[8];
+        ThreadLocalRandom.current().nextBytes(r);
```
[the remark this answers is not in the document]
trustin: Probably not worth it? Its sole purpose is uniquification and validation relies on encryption performed by JWT.

saml/src/main/java/com/linecorp/armeria/server/saml/HttpPostBindingUtil.java
```diff
@@ -0,0 +1,133 @@
...
+     * a {@link SAMLObject}.
+     */
+    static <T extends SAMLObject> MessageContext<T> toSamlObject(AggregatedHttpMessage msg,
+                                                                 String name) throws SamlException {
+        final SamlParameters parameters = new SamlParameters(msg);
+        final byte[] decoded;
+        try {
+            decoded = Base64.getDecoder().decode(parameters.getFirstValue(name));
+        } catch (IllegalArgumentException e) {
+            throw new SamlException("failed to decode a base64 string of the parameter " + name, e);
```
trustin: nit: How about wrapping `name` with single quotes, or appending `:` after `parameter`?

saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,341 @@
...
+     * Encodes the specified {@code message} into a deflated base64 string.
+     */
+    static String toDeflatedBase64(SAMLObject message) throws SamlException {
+        requireNonNull(message, "message");
+
+        final String messageStr;
+        try {
+            messageStr = nodeToString(XMLObjectSupport.marshall(message));
+        } catch (MarshallingException e) {
+            throw new SamlException("failed to marshal a SAML message", e);
```
trustin: I'm curious which is correct, marshal or marshall? Alternatively, we could just use 'serialize' and 'deserialize'. :-D

saml/src/main/java/com/linecorp/armeria/server/saml/SamlMessageUtil.java
```diff
@@ -0,0 +1,183 @@
...
+    static XMLObject unmarshalMessage(byte[] bytes) throws SamlException {
+        requireNonNull(bytes, "bytes");
+        final ParserPool parserPool = XMLObjectProviderRegistrySupport.getParserPool();
+        assert parserPool != null;
+
+        final InputStream is = new ByteArrayInputStream(bytes);
+        try {
+            return XMLObjectSupport.unmarshallFromInputStream(parserPool, is);
+        } catch (XMLParserException | UnmarshallingException e) {
+            throw new SamlException("failed to unmarshal a XML document bytes into a SAML object", e);
```
trustin: an XML

saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        return messageContext;
+    }
+
+    /**
+     * Build a query string which does not start with '?'.
+     */
+    static final class QueryStringBuilder {
+        private static final String UTF_8 = StandardCharsets.UTF_8.name();
```
trustin: Perhaps we could use Netty's `QueryStringEncoder` and append a substring? e.g.

```java
QueryStringEncoder qse = new QueryStringEncoder("");
qse.addParam();
qse.toString().substring(1);
...

```

saml/src/main/java/com/linecorp/armeria/server/saml/SamlMessageUtil.java
```diff
@@ -0,0 +1,183 @@
...
+        if (message.getDOM() != null) {
+            // Return cached DOM if it exists.
+            return message.getDOM();
+        }
+
+        final Marshaller marshaller =
+                XMLObjectProviderRegistrySupport.getMarshallerFactory().getMarshaller(message);
+        if (marshaller == null) {
+            throw new SamlException(
+                    "Unable to marshal message, no marshaller registered for message object: " +
```
trustin: U -> u, for consistency?

saml/src/main/java/com/linecorp/armeria/server/saml/SamlMessageUtil.java
```diff
@@ -0,0 +1,183 @@
...
+                    message.getElementQName());
+        }
+
+        try {
+            return marshaller.marshall(message);
+        } catch (MarshallingException e) {
+            throw new SamlException(
+                    "failed to marshal a SAML object into XML document", e);
```
trustin: XML -> an XML

saml/src/main/java/com/linecorp/armeria/server/saml/SamlMetadataServiceFunction.java
```diff
@@ -0,0 +1,208 @@
...
+
+    @Override
+    public HttpResponse serve(ServiceRequestContext ctx, AggregatedHttpMessage msg,
+                              String defaultHostname, SamlPortConfig portConfig) {
+        final HttpData metadata = metadataMap.computeIfAbsent(defaultHostname, h -> {
+            try {
+                final Element element =
+                        SamlMessageUtil.marshalMessage(buildMetadataEntityDescriptorElement(h, portConfig));
+                final HttpData newMetadata = HttpData.ofUtf8(nodeToString(element));
+                logger.debug("SAML service provider metadata has been prepared for virtual host {}.", h);
```
trustin: How about: `for: {}`, because VirtualHost.toString() starts with `VirtualHost(`

saml/src/main/java/com/linecorp/armeria/server/saml/SamlServiceProvider.java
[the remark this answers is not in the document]
trustin: `ImmutableList.copyOf()` does not perform any copy operation if the given list is already immutable. So, it's perhaps OK to do `copyOf()` here for safety?

saml/src/main/java/com/linecorp/armeria/server/saml/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * SAML-based single sign-on service.
```
trustin: How about: `<a href="https://en.wikipedia.org/wiki/Security_Assertion_Markup_Language">SAML</a>-based`?

saml/src/main/java/com/linecorp/armeria/server/saml/SamlServiceProvider.java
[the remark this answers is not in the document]
hyangtack: @trustin Ok, will add it!

saml/src/main/java/com/linecorp/armeria/server/saml/HttpPostBindingUtil.java
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java
hyangtack: Fixed using 'serialize' and 'deserialize'. Thanks!

saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java
hyangtack: Fixed. Thanks!

saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public String newId() {
+        final Instant now = Instant.now();
+        final int un2 = ThreadLocalRandom.current().nextInt(Integer.MAX_VALUE) & 0x7fffffff;
+        return JWT.create()
+                  .withIssuer(issuer)
+                  .withIssuedAt(Date.from(now))
+                  .withExpiresAt(Date.from(now.plus(validSeconds, ChronoUnit.SECONDS)))
+                  // To make this token unique in a second, uniquifiers will be added.
```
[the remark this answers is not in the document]
hyangtack: Fixed. Thanks!

saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
```diff
@@ -0,0 +1,132 @@
...
+        // To make a request ID globally unique, we will add MAC-based machine ID and a random number.
+        // The random number tries to make this instance unique in the same machine and process.
+        final byte[] r = new byte[8];
+        ThreadLocalRandom.current().nextBytes(r);
+        return new StringBuilder().append(encodeHex(defaultMachineId(), digits))
+                                  .append(encodeHex(r, digits))
+                                  .toString();
+    }
+
+    private static char[] encodeHex(byte[] data, char[] digits) {
```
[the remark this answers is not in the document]
hyangtack: Fixed. Thank you for letting me know that!

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+public final class KeyStoreCredentialResolverBuilder {
+
+    private final String path;
+    private final String password;
+    private final Map<String, String> keyPasswords = new HashMap<>();
+
+    /**
+     * Creates a builder instance for {@link KeyStoreCredentialResolver}.
+     */
+    public KeyStoreCredentialResolverBuilder(String path, String password) {
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlMessageUtil.java
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlMessageUtil.java
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlMetadataServiceFunction.java
hyangtack: Good point! Fixed.

saml/src/main/java/com/linecorp/armeria/server/saml/package-info.java
hyangtack: Fixed. Thanks!

saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
```diff
@@ -0,0 +1,121 @@
...
+                          .add("leewaySeconds", leewaySeconds)
+                          .toString();
+    }
+
+    private static String getUniquifierPrefix() {
+        // To make a request ID globally unique, we will add MAC-based machine ID and a random number.
+        // The random number tries to make this instance unique in the same machine and process.
+        final byte[] r = new byte[8];
+        ThreadLocalRandom.current().nextBytes(r);
+        final BaseEncoding encoder = BaseEncoding.base16();
```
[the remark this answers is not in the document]
trustin: If it can be any string, we could just use base64, or even just a raw `byte[]` if that's allowed?

dependencies.yml
```diff
@@ -8,6 +8,12 @@ ch.qos.logback:
     javadocs:
     - https://logback.qos.ch/apidocs/
 
+com.auth0:
+  java-jwt:
+    version: '3.4.0'
+    javadocs:
+    - https://www.javadoc.io/doc/com.auth0/java-jwt/3.3.0/
```
trustin: https://www.javadoc.io/doc/com.auth0/java-jwt/3.4.0/

saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,293 @@
...
+import com.linecorp.armeria.server.saml.SamlService.SamlParameters;
+
+import io.netty.handler.codec.http.HttpHeaderValues;
+import io.netty.handler.codec.http.QueryStringEncoder;
+
+/**
+ * A utility class which supports HTTP-Redirect binding protocol.
+ */
+final class HttpRedirectBindingUtil {
+    private static final Logger logger = LoggerFactory.getLogger(HttpRedirectBindingUtil.class);
```
minwoox: nit: this is not used

saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java
```diff
@@ -0,0 +1,293 @@
...
+                                              new Deflater(Deflater.DEFLATED, true))) {
+            deflaterStream.write(messageStr.getBytes(StandardCharsets.UTF_8));
+        } catch (IOException e) {
+            throw new SamlException("failed to deflate a SAML message", e);
+        }
+        return bytesOut.toString();
+    }
+
+    /**
+     * Decodes, inflates and deserialize the specified base64-encoded message to a {@link XMLObject}.
```
minwoox: nit: deserializesto an {@link SMLObject}

saml/src/main/java/com/linecorp/armeria/server/saml/SamlService.java
```diff
@@ -0,0 +1,222 @@
...
+                throw new SamlException("failed to get the value of a parameter: " + name);
+            }
+            return value;
+        }
+
+        /**
+         * Returns the first value of the parameter with the specified {@code name}. If it does not exist,
+         * {@code null} is returned.
+         *
+         * @throws NullPointerException if the specified {@code name} is {@code null}
```
minwoox: Perhaps, we can omit this statement like we did in other methods in which use `requireNonNull`.

saml/src/main/java/com/linecorp/armeria/server/saml/SamlService.java
```diff
@@ -0,0 +1,222 @@
...
+                decoder = new QueryStringDecoder(path, true);
+            }
+
+            parameters = decoder.parameters();
+        }
+
+        /**
+         * Returns the first value of the parameter with the specified {@code name}.
+         *
+         * @throws NullPointerException if the specified {@code name} is {@code null}
```
minwoox: ditto

saml/src/main/java/com/linecorp/armeria/server/saml/SamlServiceProvider.java
```diff
@@ -0,0 +1,259 @@
...
+import com.linecorp.armeria.server.ServiceWithPathMappings;
+import com.linecorp.armeria.server.auth.Authorizer;
+
+/**
+ * A SAML service provider implementation.
+ *
+ * @see <a href="https://line.github.io/armeria/advanced-saml.html">SAML Single Sign-On</a>
+ */
+public final class SamlServiceProvider {
+    private static final Logger logger = LoggerFactory.getLogger(SamlServiceProvider.class);
```
minwoox: not used

saml/src/main/java/com/linecorp/armeria/server/saml/SamlIdentityProviderConfig.java
```diff
@@ -0,0 +1,137 @@
...
+import org.opensaml.security.credential.Credential;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * A configuration for an identity provider.
+ */
+public final class SamlIdentityProviderConfig {
+
+    // IdP(Identity Prodiver) related:
```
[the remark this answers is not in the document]
hyangtack: Yeah, I want to be! 😆

saml/src/main/java/com/linecorp/armeria/server/saml/JwtBasedSamlRequestIdManager.java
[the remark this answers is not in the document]
hyangtack: Base64 sounds good to me. Will change to it and use 6 random bytes in order not to produce unnecessary paddings.

saml/src/main/java/com/linecorp/armeria/server/saml/SamlAssertionConsumerFunction.java
```diff
@@ -0,0 +1,278 @@
...
+
+import com.linecorp.armeria.common.AggregatedHttpMessage;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+/**
+ * A service which receives an assertion from the remote identity provider.
+ */
+final class SamlAssertionConsumerFunction implements SamlServiceFunction {
+    private static final Logger logger = LoggerFactory.getLogger(SamlAssertionConsumerFunction.class);
```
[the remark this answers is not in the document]
hyangtack: Removed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlAssertionConsumerFunction.java
```diff
@@ -0,0 +1,278 @@
...
+            }
+        }
+        throw new SamlException("failed to find identity provider from configuration " +
+                                issuer.getValue());
+    }
+
+    private Assertion getValidatedAssertion(Response response, String endpointUri) throws SamlException {
+        final Status status = response.getStatus();
+        final String statusCode = status.getStatusCode().getValue();
+        if (!StatusCode.SUCCESS.equals(status.getStatusCode().getValue())) {
```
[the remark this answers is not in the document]
hyangtack: Oops, thanks!

saml/src/main/java/com/linecorp/armeria/server/saml/SamlMessageUtil.java
```diff
@@ -0,0 +1,182 @@
...
+import org.w3c.dom.Element;
+
+import net.shibboleth.utilities.java.support.xml.ParserPool;
+import net.shibboleth.utilities.java.support.xml.XMLParserException;
+
+/**
+ * A utility class for SAML messages.
+ */
+final class SamlMessageUtil {
+    private static final Logger logger = LoggerFactory.getLogger(SamlMessageUtil.class);
```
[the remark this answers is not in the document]
hyangtack: Removed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlNameIdFormat.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    EMAIL("urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"),
+
+    /**
+     * X509 subject name format.
+     */
+    X509_SUBJECT("urn:oasis:names:tc:SAML:1.1:nameid-format:X509SubjectName"),
+
+    /**
+     * Windows domain qualified name name format.
```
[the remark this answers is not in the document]
hyangtack: Fixed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlRequestIdManager.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * An interface which generates and validates a request ID when transferring a SAML message between
+ * a service provider and an identity provider.
+ */
+public interface SamlRequestIdManager {
+    /**
+     * Returns a {@link SamlRequestIdManager} implementation based on JSON Web Tokens specification.
+     *
+     * @param issuer the ID who issues a token
```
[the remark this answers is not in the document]
hyangtack: Changed to `the ID of the entity who issues a token`

saml/src/main/java/com/linecorp/armeria/server/saml/SamlService.java
hyangtack: Ok, removed

saml/src/main/java/com/linecorp/armeria/server/saml/SamlServiceProvider.java
hyangtack: Removed

trustin:
Thumbs up!

PR #1296, opened 2018-07-18
Make Armeria Retrofit subscriber not buffer response

kojilin:
Current implementation will buffer all data and call callback when complete.
So it's not suitable for downloading large files.

This PR will make subscriber requests data one by one from response.　And
because `SourceBuffer#read` and `Subscriber#onNext` are using same thread, we need to
using CommonPools to run `SourceBuffer#read` to avoiding deadlock.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/CommonPools.java
+++ b/core/src/main/java/com/linecorp/armeria/common/CommonPools.java
@@ -51,8 +51,8 @@ public final class CommonPools {
     }
 
     /**
-     * Returns the common blocking task {@link Executor} which is used when
-     * {@link ServerBuilder#blockingTaskExecutor(Executor)} is not specified.
+     * Returns the default common blocking task {@link Executor} which is used for
+     * running a potentially long-running tasks which may block I/O threads.
      */
     public static Executor blockingTaskExecutor() {
         return BLOCKING_TASK_EXECUTOR;
--- a/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
+++ b/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
@@ -30,6 +30,7 @@ import com.linecorp.armeria.client.ClientFactory;
 import com.linecorp.armeria.client.ClientOptionsBuilder;
 import com.linecorp.armeria.client.HttpClient;
 import com.linecorp.armeria.client.endpoint.EndpointGroup;
+import com.linecorp.armeria.common.CommonPools;
 import com.linecorp.armeria.common.Scheme;
 import com.linecorp.armeria.common.SessionProtocol;
 
@@ -40,6 +41,7 @@ import retrofit2.Callback;
 import retrofit2.Converter;
 import retrofit2.Retrofit;
 import retrofit2.Retrofit.Builder;
+import retrofit2.http.Streaming;
 
 /**
  * A helper class for creating a new {@link Retrofit} instance with {@link ArmeriaCallFactory}.
@@ -77,6 +79,8 @@ public final class ArmeriaRetrofitBuilder {
     private final ClientFactory clientFactory;
     @Nullable
     private String baseUrl;
+    private boolean streaming;
+    private Executor callbackExecutor = CommonPools.blockingTaskExecutor();
     private BiFunction<String, ? super ClientOptionsBuilder, ClientOptionsBuilder> configurator =
             DEFAULT_CONFIGURATOR;
 
@@ -154,7 +158,26 @@ public final class ArmeriaRetrofitBuilder {
      * @see Retrofit.Builder#callbackExecutor(Executor)
      */
     public ArmeriaRetrofitBuilder callbackExecutor(Executor executor) {
-        retrofitBuilder.callbackExecutor(requireNonNull(executor, "executor"));
+        callbackExecutor = requireNonNull(executor, "executor");
+        retrofitBuilder.callbackExecutor(callbackExecutor);
+        return this;
+    }
+
+    /**
+     * Sets the streaming flag to make Armeria client fully support {@link Streaming}.
+     * If this flag is {@code false}, Armeria client will buffer all data and call a callback after receiving
+     * the data completely, even if a service method was annotated with {@link Streaming}.
+     * By enabling this flag, Armeria client will use the {@link Executor} specified with
+     * {@link ArmeriaRetrofitBuilder#callbackExecutor(Executor)} to read the data in a blocking way.
+     *
+     * <p>Note: It is not recommended to have the service methods both with and without the
+     * {@link Streaming} annotation in the same interface. Consider separating them into different
+     * interfaces and using different builder to build the service.</p>
+     *
+     * @see Streaming
+     */
+    public ArmeriaRetrofitBuilder streaming(boolean streaming) {
+        this.streaming = streaming;
         return this;
     }
 
@@ -179,7 +202,10 @@ public final class ArmeriaRetrofitBuilder {
         final HttpClient baseHttpClient = HttpClient.of(
                 clientFactory, fullUri, configurator.apply(fullUri, new ClientOptionsBuilder()).build());
         return retrofitBuilder.baseUrl(convertToOkHttpUrl(baseHttpClient, uri.getPath(), GROUP_PREFIX))
-                              .callFactory(new ArmeriaCallFactory(baseHttpClient, clientFactory, configurator))
+                              .callFactory(new ArmeriaCallFactory(
+                                      baseHttpClient, clientFactory, configurator,
+                                      streaming ? SubscriberFactory.streaming(callbackExecutor)
+                                                : SubscriberFactory.blocking()))
                               .build();
     }
 
--- a/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/StreamingCallSubscriber.java
+++ b/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/StreamingCallSubscriber.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.client.retrofit2;
+
+import java.io.IOException;
+import java.util.concurrent.Executor;
+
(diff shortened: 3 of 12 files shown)


retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallSubscriber.java
```diff
@@ -104,12 +109,25 @@ public void onNext(HttpObject httpObject) {
             if (contentLength == NO_CONTENT_LENGTH) {
                 contentLength = headers.getLong(HttpHeaderNames.CONTENT_LENGTH, NO_CONTENT_LENGTH);
             }
-
             return;
         }
+        if (!callbackCalled) {
+            responseBuilder.body(ResponseBody.create(
+                    Strings.isNullOrEmpty(contentType) ? null : MediaType.parse(contentType),
+                    contentLength, Okio.buffer(new ForwardingSource(pipeBuffer.source()) {
```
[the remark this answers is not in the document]
kojilin: I didn't find any clue and https://github.com/square/okhttp/wiki/Recipes#asynchronous-get looks like there is blocking at read.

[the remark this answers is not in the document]
trustin: That's a sad story. :sob: No worries, then.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallSubscriber.java
```diff
@@ -119,40 +137,44 @@ public void onError(Throwable throwable) {
...
+        pipeBuffer.close(canceledException);
     }
 
     private void safeOnFailure(IOException e) {
         if (callbackCalled) {
             return;
         }
         callbackCalled = true;
-        callback.onFailure(armeriaCall, e);
+        CommonPools.blockingTaskExecutor().execute(() -> callback.onFailure(armeriaCall, e));
```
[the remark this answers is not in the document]
kojilin: I will add at blockingTaskExecutor,
What I thought is retrofit already provides `callbackExecutor` setting, and armeria retrofit also expose it already, so if they use that executor, it becomes only callback at commons pools and blocking at user defined executor. or we should also use same executor here?

https://github.com/line/armeria/blob/master/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java#L156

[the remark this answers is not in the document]
trustin: Ah, using the `callbackExecutor` sounds better to me. I'm curious what the default would be if we do not set it at all. If Retrofit has a sensible default, we won't need to update `CommonPools` at all and leave `blockingTaskExecutor` on the server side.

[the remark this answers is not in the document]
kojilin: if we don't set, retrofit using it's net thread worker directly(from http://square.github.io/retrofit/)
> On Android, callbacks will be executed on the main thread. On the JVM, callbacks will happen on the same thread that executed the HTTP request.
So we will have deadlock when callback -> start reading from buffer(waiting onNext, but thread halt)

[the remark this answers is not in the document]
trustin: I guess in that case we could default to `CommonPools.blockingTaskExecutor()` as discussed initially. That way, we wouldn't need to add a new setter method to `ArmeriaRetrofitBuilder` but just reuse `ArmeriaRetrofitBuilder.callbackExecutor()`. Sounds good?

core/src/main/java/com/linecorp/armeria/common/CommonPools.java
```diff
@@ -52,7 +52,10 @@
 
     /**
      * Returns the common blocking task {@link Executor} which is used when
-     * {@link ServerBuilder#blockingTaskExecutor(Executor)} is not specified.
+     * {@link ServerBuilder#blockingTaskExecutor(Executor)} is not specified, or
+     * {@link com.linecorp.armeria.client.retrofit2.ArmeriaRetrofitBuilder#streaming(boolean)} is true and
+     * {@link com.linecorp.armeria.client.retrofit2.ArmeriaRetrofitBuilder#callbackExecutor(Executor)}
```
trustin: Doesn't this make `./gradlew :core:javadoc` fail, because `ArmeriaRetrofit` is not available? How about simplifying this way:

```
Returns the default common blocking task {@link Executor} which is used for
running a potentially long-running tasks which may block I/O threads.
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
```diff
@@ -154,7 +161,24 @@ public ArmeriaRetrofitBuilder addCallAdapterFactory(CallAdapter.Factory factory)
...
      */
     public ArmeriaRetrofitBuilder callbackExecutor(Executor executor) {
-        retrofitBuilder.callbackExecutor(requireNonNull(executor, "executor"));
+        callbackExecutor = requireNonNull(executor, "executor");
+        return this;
+    }
+
+    /**
+     * Sets the streaming flag to make Armeria client fully support {@link Streaming}.
+     * If this flag is false, even service method has {@link Streaming} annotation, armeria client will buffer
```
trustin:
```
If this flag is {@code false}, Armeria client will buffer all data and call a callback after receiving the data completely, even if a service method was annotated with {@link Streaming}.
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
```diff
@@ -154,7 +161,24 @@ public ArmeriaRetrofitBuilder addCallAdapterFactory(CallAdapter.Factory factory)
...
+     * all data and call callback after receiving all data.
+     * Enable this flag will make Armeria client use {@link ArmeriaRetrofitBuilder#callbackExecutor} or
+     * {@link CommonPools#blockingTaskExecutor()} to read the data in blocking way.
```
trustin:
```
By enabling this flag, Armeria client will use the {@link Executor} specified with
{@link ArmeriaRetrofitBuilder#callbackExecutor(Executor)} to read the data in a blocking way.
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
```diff
@@ -154,7 +161,24 @@ public ArmeriaRetrofitBuilder addCallAdapterFactory(CallAdapter.Factory factory)
...
+     *
+     * <p>Note: Using this flag when service method has {@link Streaming} only, separate service methods have
+     * {@link Streaming} or not and using different {@link ArmeriaRetrofitBuilder} may be better.</p>
```
trustin:
```
<p>Note: It is not recommended to have the service methods both with and without the
{@link Streaming} annotation in the same interface. Consider separating them into different
interfaces.</p>
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
```diff
@@ -75,6 +79,9 @@
 
     private final Retrofit.Builder retrofitBuilder;
     private final ClientFactory clientFactory;
+    private boolean streaming;
+    @Nullable
+    private Executor callbackExecutor;
```
trustin: Shouldn't this default to `CommonPools.blockingTaskExecutor()`?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/StreamingCallSubscriber.java
```diff
@@ -0,0 +1,186 @@
...
+            try {
+                callback.onResponse(armeriaCall, response);
+            } catch (IOException e) {
+                callback.onFailure(armeriaCall, e);
+            }
+        });
+    }
+
+    private static IOException newCanceledException() {
+        return new IOException("Canceled");
```
trustin: - `cancelled`? (To be consistent with `CompletableFuture.isCancelled()`)
- `newCancelledException`

retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactoryTest.java
```diff
@@ -284,15 +292,28 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req) throws
...
+
     private Service service;
 
+    public ArmeriaCallFactoryTest(boolean streaming) {
+        this.streaming = streaming;
+    }
+
+    @Parameters(name = "{index}: streaming={0}")
+    public static Collection<Boolean> parameters() {
+        return ImmutableList.of(true, false);
```
trustin: This method could be placed before non-static members.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/StreamingCallSubscriber.java
```diff
@@ -0,0 +1,186 @@
...
+            onCancel(subscription);
+            return;
+        }
+        if (httpObject instanceof HttpHeaders) {
+            subscription.request(1);
+            final HttpHeaders headers = (HttpHeaders) httpObject;
+            final HttpStatus status = headers.status();
+            if (!nonInformationalHeadersStarted) { // If not received a non-informational header yet
+                // Ignore informational headers or the headers without :status.
+                if (status == null || status.codeClass() == HttpStatusClass.INFORMATIONAL) {
```
trustin: I see some code duplication between this class and `BlockingCallSubscriber`. Would it be possible to extract the common logic somewhere, like introducing a parent abstract class or a utility class for common logic?

retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactoryLargeStreamTest.java
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright 2016 LINE Corporation
```
hyangtack: 2018?

core/src/main/java/com/linecorp/armeria/common/CommonPools.java
```diff
@@ -51,8 +51,8 @@
     }
 
     /**
-     * Returns the common blocking task {@link Executor} which is used when
-     * {@link ServerBuilder#blockingTaskExecutor(Executor)} is not specified.
+     * Returns the default common blocking task {@link Executor} which is used for
+     * running a potentially long-running tasks which may block I/O threads.
```
minwoox: nit: used for a potentially long-running? (remove the duplicate running)

trustin:
Thanks, @kojilin !

core/src/main/java/com/linecorp/armeria/common/CommonPools.java
trustin: Fixed in 9576e472ad1ec07311acb045fc2fe29c0a1521e6 (Typo in the commit message though :sob:)

PR #1297, opened 2018-07-19
Clean up logging related with TLS handshake failures

trustin:
Motivation:

On TLS handshake failure, we see at least two log messages - one from
the handshake failure event and another from `exceptionCaught()`.
There's no point of logging about TLS-related exceptions after TLS
handshake failure, since the connection is irrecoverable and will be
closed automatically by Netty.

Modifications:

- Do not log TLS-related exceptions after TLS handshake failure, while
  logging non-TLS exceptions properly.

Result:

Less noisy log


trustin:
Thanks for reviewing.

PR #1298, opened 2018-07-19
Add RejectedPathMappingHandler

### LINKED ISSUE #1283: Add an option to disable duplicate path mapping detection
As proposed by @anuraaga at https://github.com/line/armeria/issues/529#issuecomment-399030430

trustin:
Motivation:

When Armeria is used as a part of another framework/library, the
framework/library may want to have control over how duplicate path
mappings are handled, rather than just throwing an exception.

Modifications:

- Add `RejectedPathMappingHandler` which provides a callback method for
  rejected duplicate path mappings
- Change the default behavior from 'throw an exception' to 'log a warning'
- Add `ServerBuilder.rejectedPathMappingHandler()`

Result:

- Fixes #1283 
- More control over duplicate path mappings

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/RejectedPathMappingHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/server/RejectedPathMappingHandler.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/linecorp/armeria/server/RejectedPathMappingHandler.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * A {@link RejectedPathMappingHandler} that logs a warning message for a problematic {@link PathMapping}.
+     */
+    RejectedPathMappingHandler WARN = (virtualHost, mapping, existingMapping) -> {
+        final Logger logger = LoggerFactory.getLogger(RejectedPathMappingHandler.class);
+        final String a = mapping.toString();
+        final String b = existingMapping.toString();
+        final String hostnamePattern = virtualHost.hostnamePattern();
+        if (a.equals(b)) {
+            logger.warn("Virtual host '{}' has a duplicate path mapping: {}",
```
anuraaga: Since this only really happens once, maybe we don't need to worry about the efficiency and could just share the same code that uses concatenation for constructing the error message.

trustin: We are still in Java 8 and we can't have a private method in an interface, so let me leave this as it is for now rather than introducing an extra utility class. Will add a todo. :-)

trustin:
Thanks for reviewing, folks. :smile:

PR #1299, opened 2018-07-19
Fix a bug where Endpoint.toString() does not include IP address

trustin:
Motivation:

Endpoint.toString() should include the IP address of an endpoint if
necessary. It currently does not include it at all.

Modifications:

- Fix incorrect `if` statement.
- Update the test case.

Result:

- The string representation of an `Endpoint` now includes its IP
  address.


trustin:
Thanks for reviewing :-)

PR #1300, opened 2018-07-20
Fix flaky test where retry is made before cancelled in RetryingRpcCli…

minwoox:
…entTest

Motivation:
In the test which is  `doNotRetryWhenResponseIsCancelled()`, retry shouldn't be made
because `RpcResponse` is cancelled. However the threads who are calling cancel and sending
the requests are different, the retry can happen.

Modification:
- Increase the delay for the next retry

Result:
- Fixed flaky test


trustin:
Thanks, @minwoox!

PR #1303, opened 2018-07-24
Add Authorizer.orElse() / Use composition for HttpAuthService

### LINKED ISSUE #506: Allow a user to specify alternate success/failure logic when building HttpAuthService with HttpAuthServiceBuilder
`HttpAuthService` has two protected methods that allows a user to override the default behavior of when authentication succeeds or fails: `HttpAuthService.onSuccess()` and `HttpAuthService.onFailure()`.

However, it is impossible for a user to override the default behavior when building it using `HttpAuthServiceBuilder`. A user has to extend it which is less convenient.

We could define a callback interface like the following so that a user can specify it when building with `HttpAuthServiceBuilder`:

```java
public interface HttpAuthHandler { // better name?
    static HttpAuthHandler DEFAULT = new HttpAuthHandler() {
        @Override
        public HttpResponse onSuccess(ServiceRequestContext ctx, HttpRequest req) throws Exception {
        }

        @Override
        public HttpResponse onFailure(ServiceRequestContext ctx, HttpRequest req, @Nullable Throwable cause)
            throws Exception {
            // Log the cause if necessary.
            final DefaultHttpResponse res = new DefaultHttpResponse();
            res.respond(HttpStatus.UNAUTHORIZED);
            return res;
        }
    };

    HttpResponse onSuccess(ServiceRequestContext ctx, HttpRequest req) throws Exception;
    HttpResponse onFailure(ServiceRequestContext ctx, HttpRequest req, @Nullable Throwable cause)
            throws Exception;
}
```

trustin:
Motivation:

- The composition of `Authorizer`s are currently done by
  `HttpAuthService` but they could be better moved to `Authorizer` just
  like `Function`s are composed with `andThen()` or `compose()`.
- Type hierarchy of `HttpAuthService` is weird and it sometimes
  discourages the use of `HttpAuthServiceBuilder`, which is an API
  design mistake.

Modifications:

- Add `Authorizer.orElse()` for functional composition of multiple
  `Authorizer`s
- Make `HttpAuthService` final
  - Add `AuthSuccessHandler` and `AuthFailureHandler` which can be
    configured via `HttpAuthServiceBuilder.onSuccess()` and `onFailure()`

Result:

- Functional composition of `Authorizer`s
- Better API design
- Breaking change: `HttpAuthService` is not extensible anymore and must
  be built with `HttpAuthServiceBuilder`.
- Fixes #506


trustin:
Based on @anuraaga's comment about `CompositeAuthorizer` in #1292

anuraaga:
Also finally fixes #506 :D

core/src/main/java/com/linecorp/armeria/server/auth/Authorizer.java
```diff
@@ -37,4 +40,55 @@
...
+    default Authorizer<T> orElse(Authorizer<T> nextAuthorizer) {
+        final Authorizer<T> self = this;
+        return new Authorizer<T>() {
+            @Nullable
+            private String strVal;
+
+            @Override
+            public CompletionStage<Boolean> authorize(ServiceRequestContext ctx, T data) {
+                return self.authorize(ctx, data).thenComposeAsync(result -> {
+                    if (result != null && result) {
```
anuraaga: Our javadoc for `Authorizer` only mentions `true` or `false`, and I think an `Authorizer` that returns `null` should be considered invalid. Do we need this `null` check?

core/src/main/java/com/linecorp/armeria/server/auth/HttpAuthService.java
```diff
@@ -66,52 +66,44 @@
...
+        this.successHandler = successHandler;
+        this.failureHandler = failureHandler;
     }
 
     @Override
     public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
-        return HttpResponse.from(authorize(req, ctx).handleAsync((result, t) -> {
+        return HttpResponse.from(authorize(ctx, req).handleAsync((result, t) -> {
             try {
                 if (t != null || !result) {
```
anuraaga: Here we're not checking `result != null`

core/src/main/java/com/linecorp/armeria/server/auth/HttpAuthServiceBuilder.java
```diff
@@ -116,22 +132,52 @@ public HttpAuthServiceBuilder addOAuth2(Authorizer<? super OAuth2Token> authoriz
...
+     * By default, the request will be delegated to the next {@link Service}.
+     */
+    public HttpAuthServiceBuilder onSuccess(AuthSuccessHandler<HttpRequest, HttpResponse> successHandler) {
+        this.successHandler = requireNonNull(successHandler, "successHandler");
+        return this;
+    }
+
+    /**
+     * Sets the {@link AuthFailureHandler} which handles the requests with failed authorization.
+     * By default, the cause of authorization failure is logged at WARN level (if any) and a
```
anuraaga: `an exception thrown during authorization is logged at...`

I guess `result == false` is also an authorization failure but we don't log anything.

core/src/test/java/com/linecorp/armeria/server/auth/AuthorizerTest.java
```diff
@@ -0,0 +1,144 @@
...
+public class AuthorizerTest {
+
+    @Nullable
+    private static EventLoop eventLoop;
+    @Nullable
+    private static ServiceRequestContext serviceCtx;
+
+    @BeforeClass
+    public static void setEventLoop() {
+        eventLoop = new DefaultEventLoop();
```
anuraaga: Looking forward to a new `@Rule` for this :) (actually maybe not since I use junit5 now...)

core/src/test/java/com/linecorp/armeria/server/auth/AuthorizerTest.java
```diff
@@ -0,0 +1,144 @@
...
+        serviceCtx = mock(ServiceRequestContext.class);
+        when(serviceCtx.contextAwareEventLoop()).thenReturn(eventLoop);
```
anuraaga: Though I've always been wondering whether it was really a good idea for `ServiceRequestContext` to return `EventLoop`. Mocking in unit tests would be way easier if it returned a more generic thing like `Executor` or `ScheduledExecutorService`.

core/src/test/java/com/linecorp/armeria/server/auth/AuthorizerTest.java
```diff
@@ -0,0 +1,144 @@
...
+        if (eventLoop != null) {
+            eventLoop.shutdownGracefully();
+            eventLoop = null;
+        }
+    }
+
+    @Test
+    public void orElseFirst() {
+        final AtomicBoolean executedA = new AtomicBoolean();
+        final Authorizer<Object> a = (ctx, data) -> {
```
anuraaga: Using mocks for `Authorizer` might be more readable (can use `when` and `verify` as usual without for example this `executedA` counting guard.

core/src/main/java/com/linecorp/armeria/server/auth/Authorizer.java
trustin: Added some null checks and helpful exception messages.

core/src/main/java/com/linecorp/armeria/server/auth/HttpAuthService.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/auth/HttpAuthServiceBuilder.java
trustin: Thanks for suggestion. Fixed.

core/src/main/java/com/linecorp/armeria/server/auth/AuthorizerUtil.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2017 LINE Corporation
```
minwoox: 2018

core/src/test/java/com/linecorp/armeria/server/auth/AuthorizerTest.java
trustin: That's a good question. It is internally used often for using the desired event loop when acquiring a connection from the pool. Perhaps we can downcast there and expose `ScheduledExecutorService` there? One downside of this change would be that `ScheduledExecutorService` returns `Future` rather than `CompletableFuture`. Netty's `EventLoop` returns Netty's own `Future` so it fits better for asynchronous programming.

core/src/test/java/com/linecorp/armeria/server/auth/AuthorizerTest.java
```diff
@@ -0,0 +1,168 @@
...
+    @Test
+    public void orElseFirst() {
+        final Authorizer<Object> a = newMock();
+        when(a.authorize(any(), any())).thenReturn(completedFuture(true));
+        final Authorizer<Object> b = newMock();
+
+        final Boolean result = a.orElse(b).authorize(serviceCtx, new Object())
+                                .toCompletableFuture().join();
+        assertThat(result).isTrue();
+        verify(a, times(1)).authorize(any(), any());
```
anuraaga: We should go ahead and verify the arguments too (might want to use a string instead of object to make it easier).

core/src/test/java/com/linecorp/armeria/server/auth/AuthorizerTest.java
anuraaga: Yeah I guess replacing the method is too much but adding new methods with the downcasts could make it easier to write tests. Especially, like here `Executor` is used a ton and is usually fine being `MoreExecutors.directExecutor`.

core/src/test/java/com/linecorp/armeria/server/auth/AuthorizerTest.java
```diff
@@ -0,0 +1,168 @@
...
+        final Exception expected = new Exception();
+        final Authorizer<String> a = newMock();
+        when(a.authorize(any(), any())).thenReturn(exceptionallyCompletedFuture(expected));
+        final Authorizer<String> b = newMock();
+
+        assertThatThrownBy(() -> a.orElse(b).authorize(serviceCtx, "data").toCompletableFuture().join())
+                .isInstanceOf(CompletionException.class)
+                .hasCause(expected);
+        verify(a, times(1)).authorize(serviceCtx, "data");
+        verify(b, never()).authorize(any(), any());
```
anuraaga: You can use `verifyZeroInteractions` too if you want (maybe it's semantically simpler to say "Authorizer was never used" instead of "Authorizer's authorize method was never called). But either way's fine.

trustin: +1. Fixed :-)

trustin:
Thanks, reviewers.

PR #1307, opened 2018-07-25
Expose context-aware Executor for easier tests

anuraaga:
It is extremely common to use the `RequestContext`-aware eventloop as just an `Executor` for callbacks (e.g., `thenApplyAsync`). Unfortunately, when writing unit tests for such code, it can be cumbersome to mock `RequestContext` because there is no direct executing event loop (an event loop with an immediate executor usually deadlocks since callbacks get blocked by the previously running item in the loop). It slows down tests to create threads and is cumbersome to shut them down.

This exposes the exact same event loop as an `Executor`. It is purely to allow easier writing of tests where it them becomes possible to just do `when(ctx.contextAwareExecutor()).thenReturn(MoreExecutors.directExecutor())` for many cases where direct execution doesn't deadlock, allowing for faster, simpler tests.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/RequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/common/RequestContext.java
@@ -199,11 +199,22 @@ public interface RequestContext extends AttributeMap {
     }
 
     /**
-     * Returns an {@link EventLoop} that will make sure this {@link RequestContext} is set as the current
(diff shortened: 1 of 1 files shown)


trustin:
Perhaps we should add `RequestContext.executor()` as well? LGTM besides that.

anuraaga:
Added `executor()` as well

trustin:
Thanks, @anuraaga and reviewers.

PR #1308, opened 2018-07-26
Log the first gRPC request and response object

trustin:
Motivation:

Instead of logging no parameters and return values, we can keep the
first request and response object in a gRPC call, so that a user gets at
least some information about the call.

Modifications:

- Defer both request and response content logging for gRPC
- Set the first request/responseContent()
- Add LoggingService to GrpcServiceServerTest so that a dev can check if
  request/responseContent is recorded correctly

Result:

- More information in gRPC RequestLog

### DIFF
--- a/grpc/src/test/java/com/linecorp/armeria/client/grpc/GrpcClientTest.java
+++ b/grpc/src/test/java/com/linecorp/armeria/client/grpc/GrpcClientTest.java
@@ -817,7 +817,8 @@ public class GrpcClientTest {
         assertThat(((StatusRuntimeException) t).getStatus().getCode())
                 .isEqualTo(Status.DEADLINE_EXCEEDED.getCode());
 
-        checkRequestLogError((rpcReq, headers, cause) -> {
+        checkRequestLogError((headers, rpcReq, cause) -> {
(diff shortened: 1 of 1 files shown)


grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
```diff
@@ -530,7 +530,7 @@ private void reportCount() {
 
         private void verifySize() {
             if (count > maxMessageSize) {
-                throw Status.INTERNAL.withDescription(String.format(
+                throw Status.RESOURCE_EXHAUSTED.withDescription(String.format(
```
anuraaga: Thanks!

trustin:
Adding more test cases..

trustin:
I have relaxed some test cases because a gRPC `RequestLog` may not have `RpcRequest` or `RpcResponse` depending on when an error occurs, and it's not always easy to control the timing precisely.

trustin:
Let me re-run the builds a few more times just in case I broke anything.

PR #1309, opened 2018-07-27
add numMaxConnections Flag

demongaorui:
Inspried by https://line-armeria.slack.com/archives/C1NGPBUH2/p1527204753000180 and @chokoswitch 's idea in https://github.com/curioswitch/curiostack/issues/144#issuecomment-392628889 , try to add numMaxConnections Flag. 

And set the default value as 10000 according to https://stackoverflow.com/a/25765451 
> The defaults are maxConnections=10,000 and maxThreads=200.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -66,6 +66,10 @@ public final class Flags {
     private static final boolean USE_OPENSSL = getBoolean("useOpenSsl", OpenSsl.isAvailable(),
                                                           value -> OpenSsl.isAvailable() || !value);
 
+    private static final int DEFAULT_NUM_MAX_CONNECTIONS = 10000; // from Tomcat default maxConnections
+    private static final int NUM_MAX_CONNECTIONS =
+            getInt("numMaxConnections", DEFAULT_NUM_MAX_CONNECTIONS, value -> value > 0);
+
     private static final int DEFAULT_NUM_COMMON_WORKERS = NUM_CPU_CORES * 2;
     private static final int NUM_COMMON_WORKERS =
             getInt("numCommonWorkers", DEFAULT_NUM_COMMON_WORKERS, value -> value > 0);
@@ -236,6 +240,17 @@ public final class Flags {
      */
     public static boolean useOpenSsl() {
         return USE_OPENSSL;
+    }
+
+    /**
+     * Returns the default server-side maximum number of connections.
+     *
+     * <p>The default value of this flag is {@value #DEFAULT_NUM_MAX_CONNECTIONS}. Specify the
+     * {@code -Dcom.linecorp.armeria.numMaxConnections=<integer>} JVM option to override
+     * the default value.
+     */
+    public static int numMaxConnections() {
+        return NUM_MAX_CONNECTIONS;
     }
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -66,9 +66,9 @@ public final class Flags {
     private static final boolean USE_OPENSSL = getBoolean("useOpenSsl", OpenSsl.isAvailable(),
                                                           value -> OpenSsl.isAvailable() || !value);
 
-    private static final int DEFAULT_NUM_MAX_CONNECTIONS = 10000; // from Tomcat default maxConnections
-    private static final int NUM_MAX_CONNECTIONS =
-            getInt("numMaxConnections", DEFAULT_NUM_MAX_CONNECTIONS, value -> value > 0);
+    private static final int DEFAULT_MAX_NUM_CONNECTIONS = Integer.MAX_VALUE;
+    private static final int MAX_NUM_CONNECTIONS =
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -238,6 +242,17 @@ public static boolean useOpenSsl() {
...
     }
 
+    /**
+     * Returns the default server-side maximum number of connections.
+     *
+     * <p>The default value of this flag is {@value #DEFAULT_NUM_MAX_CONNECTIONS}. Specify the
+     * {@code -Dcom.linecorp.armeria.numMaxConnections=<integer>} JVM option to override
+     * the default value.
+     */
+    public static int numMaxConnections() {
```
trustin: `maxNumConnections`

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -66,6 +66,10 @@
     private static final boolean USE_OPENSSL = getBoolean("useOpenSsl", OpenSsl.isAvailable(),
                                                           value -> OpenSsl.isAvailable() || !value);
 
+    private static final int DEFAULT_NUM_MAX_CONNECTIONS = 10000; // from Tomcat default maxConnections
+    private static final int NUM_MAX_CONNECTIONS =
+            getInt("numMaxConnections", DEFAULT_NUM_MAX_CONNECTIONS, value -> value > 0);
```
trustin: `MAX_NUM_CONNECTIONS` and `maxNumConnections`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -924,6 +921,10 @@ public ServerBuilder serverListener(ServerListener serverListener) {
         return this;
     }
 
+    public int getMaxNumConnections() {
+        return maxNumConnections;
```
trustin: We don't usually provide a getter in a builder. If you are exposing it for testing, please make it package-private and add `@VisibleForTesting`. Also, please rename the method to `maxNumConnections()` because we do not prepend `get` for the getter methods mostly.

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -66,6 +66,10 @@
     private static final boolean USE_OPENSSL = getBoolean("useOpenSsl", OpenSsl.isAvailable(),
                                                           value -> OpenSsl.isAvailable() || !value);
 
+    private static final int DEFAULT_NUM_MAX_CONNECTIONS = 10000; // from Tomcat default maxConnections
```
trustin: I'm not sure about this default value, because we often get more connections than this. Could we keep using `Integer.MAX_VALUE`?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -238,6 +242,17 @@ public static boolean useOpenSsl() {
         return USE_OPENSSL;
     }
 
+    /**
+     * Returns the default server-side maximum number of connections.
+     *
+     * <p>The default value of this flag is {@value #DEFAULT_NUM_MAX_CONNECTIONS}. Specify the
+     * {@code -Dcom.linecorp.armeria.numMaxConnections=<integer>} JVM option to override
```
trustin: `maxNumConnections`

demongaorui:
@trustin   Thanks for your comments, updated.

core/src/main/java/com/linecorp/armeria/common/Flags.java
demongaorui: BTW, what hardware baseline you used to match these default flag values?  Like, how many vCores with how much memory?

trustin: Some of them are from Netty defaults. Some others are quite arbitrary although we tried to make it as sensible as possible. The connection timeout of 3.2 seconds is special though, which is related with the default value of some Linux socket parameter. Please feel free to let us know if you find any of them doesn't make sense.

trustin:
Thanks, @demongaorui !

core/src/main/java/com/linecorp/armeria/common/Flags.java
demongaorui: I'm thinking that maybe we could do some tests to make a baseline and recommendations tunning parameter values of Armeria. It will be very helpful for users of Armeria I think. 

Given hardware(or VM) env, with certain Armeria flag values configuration, provide some kind of performance benchmark report.

For example, for a GKE pod(4 GCP Core, 16GB Memory) with Armeria flag default values, and general web service load pattern(peek RPS, concurrency, cup load, etc. Maybe could use tools like `Load impact`), provide the performance test report to users.  

This approach will help us to decide default values for the flags based on solid evn and test feedbacks, also the performance test report could give the users a general clue of the server capacity. 

This is only an initial idea. Please feel free to share your ideas 😄

trustin: I agree in general. Would be nice if we can handle this as a part of https://github.com/line/armeria/issues/799

PR #1312, opened 2018-07-30
Allow DocService to inject user-specified scripts into the client and… 

anuraaga:
… provide an extension point to provide custom HTTP headers.

Many services require authentication using short-lived access tokens. It is not practical to set these using the http header field in the debug form. This change allows users to register custom JavaScript to run and allows them to register a customizer of HTTP headers. For example, a firebase auth user would just add this injected script to have debug requests automatically authenticated with a valid access token

```js
armeria.registerHeaderProvider(function() {
  return firebase.auth().currentUser.getIdToken().then(token => { authorization: 'Bearer ' + token });
});
```


core/src/main/java/com/linecorp/armeria/server/docs/DocServiceBuilder.java
```diff
@@ -241,6 +244,46 @@ public DocServiceBuilder exampleRequest(Iterable<?> exampleRequests) {
...
+     *   });
+     * }</pre>
+     */
+    public DocServiceBuilder injectedScript(String... scripts) {
+        requireNonNull(scripts, "scripts");
+        return injectedScript(ImmutableList.copyOf(scripts));
+    }
+
+    /**
+     * Adds a Javascript script to inject into the end of the debug page HTML. This can be used to customize
```
trustin: Aren't the scripts added to the `<head/>` section because you used `react-helmet`?

anuraaga: Oops fixed

trustin:
Thanks, @anuraaga !

Issue #1313, opened 2018-07-31
The new DocService can't send `{}` as body


rfkm:
I tried the new DocService and found an issue where it doesn't send a request whose body is `{}`.
So, I can't test 0-arity RPC methods.

anuraaga:
Thanks for reporting!

I found that there is a display issue when sending `{}` to a method that has a sample request with parameters and sent a change. But it doesn't sound like the issue you're seeing.

I tried sending an empty body and it sent the request as expected. Do you have any extra information to try to debug?

http://localhost:3000/docs/#/methods/com.linecorp.armeria.service.test.thrift.main.HelloService/hello?args=%7B%7D&http_headers=%7B%22c%22%3A%22d%22%7D

![image](https://user-images.githubusercontent.com/198344/43451150-3fed2526-94ef-11e8-8477-a62496878c0a.png)

![image](https://user-images.githubusercontent.com/198344/43451173-4b22bce4-94ef-11e8-8135-f49f24f3a427.png)

rfkm:
No tested yet, but #1314 may resolve the issue.
In my environment, I observed the following behavior:

Case A:
1. Put `{}` to the form and then click `SUBMIT`
2. Moved to `/#/.../...?args=&http_headers=...` (Note that `args` is empty) and nothing happens

Case B:
1. Put `{}\n` to the form and then click `SUBMIT`
2. Moved to `/#/.../...?args={}&http_headers=...` (Note that `args` is `{}`) and a request was successfully sent

anuraaga:
Thanks for the repro, added a fix for it :)

rfkm:
Great, thank you @anuraaga !

trustin:
Released! http://repo.maven.apache.org/maven2/com/linecorp/armeria/armeria/0.68.1/

Thanks @rfkm and @anuraaga

PR #1314, opened 2018-07-31
Don't return empty string when prettifying empty object.

anuraaga:
While migrating `prettify` to typescript, I may have changed something so it prettifies `{}` as ``, though I can't see what might cause it. This tweaks it so it'll ignore `{}` which is pretty enough as is.

Found while looking at #1313

### DIFF
--- a/docs-client/src/lib/json-prettify.ts
+++ b/docs-client/src/lib/json-prettify.ts
@@ -43,6 +43,10 @@ function doPrettify(ch: string, indentation: number): [string, number] {
 
 // A modified version of JSON.minify() by Kyle Simpson that prettifies a JSON string without fully parsing it.
 export default function prettify(json: string) {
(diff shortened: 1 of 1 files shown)


docs-client/src/lib/json-prettify.ts
```diff
@@ -43,6 +43,10 @@ function doPrettify(ch: string, indentation: number): [string, number] {
 
 // A modified version of JSON.minify() by Kyle Simpson that prettifies a JSON string without fully parsing it.
 export default function prettify(json: string) {
+  if (json === '{}') {
```
trustin: Do we need this? I thought this function works without this shortcut?

trustin: Never mind. Your PR description explains it all. But how about adding why we needed this as an inline comment?

PR #1322, opened 2018-08-06
Reworks tracing to use RequestContext natively

adriancole:
Before, we used lifecycle hooks to ensure trace contexts didn't leak.
Now, we have Brave use `RequestContext` directly. Yay.


adriancole:
this is the first output of hacktime in Fukuoka with @kojilin and @anuraaga!

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,81 @@
...
+import com.linecorp.armeria.common.RequestContext;
+import com.linecorp.armeria.internal.tracing.LogRequestContextWarningOnce;
+import com.linecorp.armeria.internal.tracing.PingPongExtra;
+
+import brave.propagation.CurrentTraceContext;
+import brave.propagation.TraceContext;
+import io.netty.util.AttributeKey;
+
+/**
+ * {@link brave.Tracing.Builder#currentTraceContext(brave.propagation.CurrentTraceContext) Tracing
```
trustin: `linkplain`?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,81 @@
...
+ * context} implemented with {@link com.linecorp.armeria.common.RequestContext}.
+ *
+ * <p>This is an alternative to synchronizing brave's try/finally style with request scope lifecycle hooks such
+ * as {@link RequestContext#onEnter(Consumer)} and {@link RequestContext#onExit(Consumer)}. Instead, this defers
+ * control to Armeria, which manages the attribute {@link #TRACE_CONTEXT_KEY}.
+ */
+public final class RequestContextCurrentTraceContext extends CurrentTraceContext {
+
+    public static final AttributeKey<TraceContext> TRACE_CONTEXT_KEY =
```
trustin: Probably OK to make private?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public CurrentTraceContext.Scope newScope(TraceContext currentSpan) {
+        // Handle inspection added to ensure we can fail-fast if this isn't installed.
+        if (PingPongExtra.maybeSetPong(currentSpan)) { return Scope.NOOP; }
+
+        if (hasCurrentRequestContextOrWarnOnce()) {
+            return Scope.NOOP;
+        }
+
+        final TraceContext previous = RequestContext.current().attr(TRACE_CONTEXT_KEY).get();
```
trustin: Could you extract `RequestContext.current().attr(TRACE_CONTEXT_KEY)` into a local variable?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,81 @@
...
+            @Override
+            public void close() {
+                RequestContext.current().attr(TRACE_CONTEXT_KEY).set(previous);
+            }
+        }
+        return new RequestContextTraceContextScope();
+    }
+
+    /** Armeria code should always have a request context available, and this won't work without it. */
+    static boolean hasCurrentRequestContextOrWarnOnce() {
```
trustin: Could be private?

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/LogRequestContextWarningOnce.java
```diff
@@ -0,0 +1,36 @@
...
+     * This won't be referenced until {@link #get()} is called. If there's only one classloader, the initializer
+     * will only be called once.
+     */
+    static class ClassLoaderHack {
+        static void loadMe() {
+        }
+
+        static {
+            logger.warn("Attempted to propagate trace context, but no request context available. " +
+                        "Did you remember to use RequestContext.contextAwareExecutor() or RequestContext.makeContextAware()");
```
trustin: - Missing `?`
- remember to use -> forget to use?

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/LogRequestContextWarningOnce.java
```diff
@@ -0,0 +1,36 @@
...
+    public RequestContext get() {
+        ClassLoaderHack.loadMe();
+        return null;
+    }
+
+    /**
+     * This won't be referenced until {@link #get()} is called. If there's only one classloader, the initializer
+     * will only be called once.
+     */
+    static class ClassLoaderHack {
```
trustin: OK to be private?

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/LogRequestContextWarningOnce.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.linecorp.armeria.common.RequestContext;
+import com.linecorp.armeria.common.tracing.RequestContextCurrentTraceContext;
+
+public enum LogRequestContextWarningOnce implements Supplier<RequestContext> {
```
trustin: - Could be package-local
- Given this is only used by `RequestContextCurrentTraceContext`, how about moving this into it?

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/PingPongExtra.java
```diff
@@ -0,0 +1,24 @@
+package com.linecorp.armeria.internal.tracing;
+
+import brave.propagation.TraceContext;
+
+/** Hack to allow us to peek inside a current trace context implementation */
+public final class PingPongExtra {
+    private boolean pong;
```
trustin: Please move a member field after static members.

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
```diff
@@ -16,32 +16,32 @@
...
-                spanInScope.close();
-                threadLocalSpan.remove();
-            }
-        });
+    public static void ensureScopeUsesRequestContext(Tracing tracing) {
+        PingPongExtra extra = new PingPongExtra();
+        // trace contexts are not recorded until Tracer.toSpan, so this won't end up as junk data
+        TraceContext dummyContext = TraceContext.newBuilder().traceId(1).spanId(1)
+                                                .extra(Collections.singletonList(extra)).build();
+        boolean scopeUsesRequestContext;
```
trustin: The local variables in this method could be all final?

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
```diff
@@ -16,32 +16,32 @@
...
+        try (CurrentTraceContext.Scope scope = tracing.currentTraceContext().newScope(dummyContext)) {
+            scopeUsesRequestContext = extra.isPong();
+        }
+        if (!scopeUsesRequestContext) {
+            throw new IllegalStateException("Please initialize Tracing.Builder.currentTraceContext("
+                                            + "new " + RequestContextCurrentTraceContext.class.getSimpleName()
+                                            + "()");
```
trustin: How about:

    "Tracing.currentTraceContext is not a " + RequestContextCurrentTraceContext.class.getSimpleName() + ". " +
    "Please call Tracing.Builder.currentTraceContext(new " + .... + ")."

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * {@link brave.Tracing.Builder#currentTraceContext(brave.propagation.CurrentTraceContext) Tracing
+ * context} implemented with {@link com.linecorp.armeria.common.RequestContext}.
+ *
+ * <p>This is an alternative to synchronizing brave's try/finally style with request scope lifecycle hooks such
+ * as {@link RequestContext#onEnter(Consumer)} and {@link RequestContext#onExit(Consumer)}. Instead, this defers
+ * control to Armeria, which manages the attribute {@link #TRACE_CONTEXT_KEY}.
+ */
+public final class RequestContextCurrentTraceContext extends CurrentTraceContext {
```
trustin: How about making this a singleton? i.e. `RequestContextCurrentTraceContext.INSTANCE`

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,81 @@
...
+import brave.propagation.TraceContext;
+import io.netty.util.AttributeKey;
+
+/**
+ * {@link brave.Tracing.Builder#currentTraceContext(brave.propagation.CurrentTraceContext) Tracing
+ * context} implemented with {@link com.linecorp.armeria.common.RequestContext}.
+ *
+ * <p>This is an alternative to synchronizing brave's try/finally style with request scope lifecycle hooks such
+ * as {@link RequestContext#onEnter(Consumer)} and {@link RequestContext#onExit(Consumer)}. Instead, this defers
+ * control to Armeria, which manages the attribute {@link #TRACE_CONTEXT_KEY}.
```
trustin: I feel this tells too much implementation-level detail. How about replacing this paragraph with:

    <p>This {@link CurrentTraceContext} stores/loads the trace context into/from a
    {@link RequestContext}'s attribute so that there's no need for thread local variables
    which can lead to unpredictable behavior in asynchronous programming.

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/PingPongExtra.java
```diff
@@ -0,0 +1,24 @@
...
+
+    /** If the input includes only this extra, set {@link #isPong() pong = true} */
+    public static boolean maybeSetPong(TraceContext context) {
+        if (context.extra().size() != 1) {
```
kojilin: `== 1` ?

adriancole: hehe yes.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
adriancole: this is shared by the client and server tracers as they have to set the attribute to be inherited.. @anuraaga mentioned it is better to do once there vs do it every time something is scoped. If there is another approach, then we can handle the concern internally here and mark it private

zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -100,7 +105,8 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
         final String method = ctx.method().name();
         span.kind(Kind.CLIENT).name(method).start();
 
-        SpanContextUtil.setupContext(SPAN_IN_THREAD, ctx, span, tracer);
+        // Ensure the trace context propagates to children
+        ctx.onChild((oldCtx, newCtx) -> newCtx.attr(TRACE_CONTEXT_KEY).set(oldCtx.attr(TRACE_CONTEXT_KEY).get()));
```
adriancole: @trustin this is the thing that sets up propagation.. done here and also the server

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
trustin: How about adding a public static method like this, then:

```java
public static void copy(RequestContext src, RequestContext dst) {
    dst.attr(TRACE_CONTEXT_KEY).set(src.attr(TRACE_CONTEXT_KEY).get());
}
```

adriancole: SGTM!

adriancole: done in latest commit. when you say go, if this is in the general shape seems mergeable, I'll backfill unit tests.

trustin: Cool. Please go ahead :-)

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,127 @@
...
+            @Override
+            public void close() {
+                contextAttribute.set(previous);
+            }
+        }
+        return new RequestContextTraceContextScope();
+    }
+
+    /** Armeria code should always have a request context available, and this won't work without it. */
+    private static boolean hasCurrentRequestContextOrWarnOnce() {
```
anuraaga: Also use `ensure`? `hasCurrentRequestContext` returns `false` when there's a request context, which doesn't seem right.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        final Attribute<TraceContext> contextAttribute = RequestContext.current().attr(TRACE_CONTEXT_KEY);
+        final TraceContext previous = contextAttribute.get();
+        contextAttribute.set(currentSpan);
+
+        // Don't remove the outer-most context (client or server request)
+        if (previous == null) { return Scope.NOOP; }
+
+        // Allow sub-spans (aka local spans) to close when Brave's scope does.
```
anuraaga: `// Removes sub-spans (i.e. local spans) from the current context when Brave's scope does. If an asynchronous sub-span, it may still complete later.`

adriancole:
fyi a comprehensive test on the trace context will need to wait for the next brave release as it had a bug

Here's what it will look like, and this passes with https://github.com/openzipkin/brave/pull/749
```java
public class RequestContextCurrentTraceContextTest extends CurrentTraceContextTest {
    DefaultAttributeMap map = new DefaultAttributeMap();
    @Rule
    public MockitoRule mockitoRule = MockitoJUnit.rule();

    @Mock(answer = Answers.CALLS_REAL_METHODS)
    RequestContext mockRequestContext;

    @Before
    public void setup() {
        when(mockRequestContext.attr(isA(AttributeKey.class))).thenAnswer(
                (Answer<Attribute>) invocation -> map.attr(invocation.getArgument(0)));
        mockRequestContext.push();
    }

    @Override
    protected CurrentTraceContext newCurrentTraceContext() {
        return RequestContextCurrentTraceContext.INSTANCE;
    }
}
```

adriancole:
ok I think we are good to go!

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
```diff
@@ -16,32 +16,38 @@
...
+                                                      .extra(Collections.singletonList(extra)).build();
+        final boolean scopeUsesRequestContext;
+        try (Scope scope = tracing.currentTraceContext().newScope(dummyContext)) {
+            scopeUsesRequestContext = extra.isPong();
+        }
+        if (!scopeUsesRequestContext) {
+            throw new IllegalStateException(
+                    "Tracing.currentTraceContext is not a " + RequestContextCurrentTraceContext.class
+                            .getSimpleName() + " scope. " +
+                    "Please call Tracing.Builder.currentTraceContext" + RequestContextCurrentTraceContext.class
```
kojilin: `currentTraceContext(`

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,160 @@
...
+        // Handle inspection added to ensure we can fail-fast if this isn't installed.
+        if (currentSpan != null && PingPongExtra.maybeSetPong(currentSpan)) {
+            return Scope.NOOP;
+        }
+
+        if (!ensureCurrentRequestContextOrWarnOnce()) {
+            return INCOMPLETE_CONFIGURATION_SCOPE;
+        }
+
+        final Attribute<TraceContext> contextAttribute = RequestContext.current().attr(TRACE_CONTEXT_KEY);
```
anuraaga: It might be slightly better to use the result of `mapCurrent` instead of reading from the thread-local twice.

zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
```diff
@@ -67,6 +68,17 @@ public void tearDown() {
         Tracing.current().close();
     }
 
+    @Test(expected = IllegalStateException.class)
+    public void newDecorator_shouldFailFastWhenRequestContextCurrentTraceContextNotConfigured() {
+        HttpTracingClient.newDecorator(Tracing.newBuilder().build());
```
anuraaga: Use `assertThatThrownBy`

zipkin/src/test/java/com/linecorp/armeria/server/tracing/HttpTracingServiceTest.java
```diff
@@ -57,6 +59,22 @@
 
     private static final String TEST_METHOD = "hello";
 
+    @After
+    public void tearDown() {
+        Tracing.current().close();
```
anuraaga: What does this do?

adriancole: clears the state caused when we initialize tracers.. this was in the HttpTracingClient file.. I copied it. It is slightly less crap than holding a ref to each tracing instance just to close them.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
adriancole: very nice

adriancole:
> What about avoiding nested ifs with early termination?
I started to do this, but elicited a "yik" comment locally :P It is a
matter of style I personally prefer, but can understand if others
don't like

zipkin/src/test/java/com/linecorp/armeria/internal/tracing/PingPongExtraTest.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2016 LINE Corporation
```
hyangtack: 2018

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,157 @@
...
+     * }</pre>
+     */
+    public static void copy(RequestContext src, RequestContext dst) {
+        dst.attr(TRACE_CONTEXT_KEY).set(src.attr(TRACE_CONTEXT_KEY).get());
+    }
+
+    @Override
+    @Nullable
+    public TraceContext get() {
+        Attribute<TraceContext> traceContextAttribute = getTraceContextAttributeOrWarnOnce();
```
hyangtack: nit: Add `final`?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Override
+    public Scope newScope(@Nullable TraceContext currentSpan) {
+        // Handle inspection added to ensure we can fail-fast if this isn't installed.
+        if (currentSpan != null && PingPongExtra.maybeSetPong(currentSpan)) {
+            return Scope.NOOP;
+        }
+
+        Attribute<TraceContext> traceContextAttribute = getTraceContextAttributeOrWarnOnce();
```
hyangtack: nit: Add `final`?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,157 @@
...
+        // If an asynchronous sub-span, it may still complete later.
+        class RequestContextTraceContextScope implements Scope {
+            @Override
+            public void close() {
+                // re-lookup the attribute to avoid holding a reference to the request if this scope is leaked
+                getTraceContextAttributeOrWarnOnce().set(previous);
+            }
+        }
+
+        return new RequestContextTraceContextScope();
```
hyangtack: I think this can be simply replaced with a lambda `return () -> getTraceContextAttributeOrWarnOnce().set(previous);`. Do you have any reason for using local inner class?

adriancole:
the reason for inner class is toString that's all. (also brave is by
default java 6, so sometimes I forget) :P

>
your call.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,157 @@
...
+         * initializer will only be called once.
+         */
+        static class ClassLoaderHack {
+            static void loadMe() {
+            }
+
+            static {
+                logger.warn("Attempted to propagate trace context, but no request context available. " +
+                            "Did you remember to use RequestContext.contextAwareExecutor() or " +
+                            "RequestContext.makeContextAware()");
```
trustin: - Missing trailing question mark? i.e. `RequestContext.makeContextAware()?`
- 'Did you forget to...' sounds better to me. What do you think?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Distributed tracing clients based on <a href="https://github.com/openzipkin/brave">Brave</a>,
+ * a Java tracing library compatible with <a href="http://zipkin.io/">Zipkin</a>.
```
trustin: How about just:

    Common classes for distributed tracing support based on <a href="...">Brave</a>.

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
```diff
@@ -16,32 +16,38 @@
...
-                                    Tracer tracer) {
-        ctx.onEnter(unused -> threadLocalSpan.set(tracer.withSpanInScope(span)));
-        ctx.onExit(unused -> {
-            final SpanInScope spanInScope = threadLocalSpan.get();
-            if (spanInScope != null) {
-                spanInScope.close();
-                threadLocalSpan.remove();
-            }
-        });
+    public static void ensureScopeUsesRequestContext(Tracing tracing) {
```
trustin: Perhaps we could move this method into `RequestContextCurrentTraceContext`? It seems more relevant to `RequestContextCurrentTraceContext`. Then we could also move `PingPongExtra` and its test class to `common.tracing` and make `PingPongExtra` package-local.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
adriancole: sgtm

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
adriancole: very nice

adriancole:
@trustin I think I got all your notes. very good ones!

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -41,10 +44,16 @@
  */
 public final class RequestContextCurrentTraceContext extends CurrentTraceContext {
 
+    /**
+     * Use this singleton when building a {@linkplain brave.Tracing} instance for use with
+     * {@linkplain HttpTracingService} or {@linkplain HttpTracingClient}.
```
trustin: Uh, these are type names, so we should use `@link` rather than `@linkplain` here.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -69,6 +78,28 @@ public String toString() {
         }
     };
 
+    /**
+     * Trace context scopes are often wrapped, for example to decorate MDC. This creates a dummy scope to
+     * ensure {@link RequestContextCurrentTraceContext#INSTANCE} is configured, even if it is wrapped.
```
trustin: Now that this method is part of public API, it needs to be more user-centric. How about:

```
Ensures the specified {@link Tracing} uses a {@link RequestContextCurrentTraceContext}.

@throws IllegalStateException if {@code tracing} does not use {@link RequestContextCurrentTraceContext}
```

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -149,9 +180,32 @@ static void loadMe() {
...
-                            "Did you remember to use RequestContext.contextAwareExecutor() or " +
-                            "RequestContext.makeContextAware()");
+                            "Did you forget to use RequestContext.contextAwareExecutor() or " +
+                            "RequestContext.makeContextAware()?");
             }
         }
     }
+
+    /** Hack to allow us to peek inside a current trace context implementation. */
+    static final class PingPongExtra {
```
trustin: `@VisibleForTesting`, because it would have been `private` without the test cases.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -149,9 +180,32 @@ static void loadMe() {
...
+        /**
+         * If the input includes only this extra, set {@link #isPong() pong = true}.
+         */
+        public static boolean maybeSetPong(TraceContext context) {
```
trustin: Could be package-local?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -149,9 +180,32 @@ static void loadMe() {
...
+                    ((PingPongExtra) extra).pong = true;
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        private boolean pong;
+
+        public boolean isPong() {
```
trustin: Could be package-local?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
adriancole: hah we were just discussing this :P

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
adriancole: ack

adriancole:
okie dokie.. pushed based on last notes. we are off to lunch and will handle anything else when back!

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,216 @@
...
+                                                      .extra(Collections.singletonList(extra)).build();
+        final boolean scopeUsesRequestContext;
+        try (Scope scope = tracing.currentTraceContext().newScope(dummyContext)) {
+            scopeUsesRequestContext = extra.isPong();
+        }
+        if (!scopeUsesRequestContext) {
+            throw new IllegalStateException(
+                    "Tracing.currentTraceContext is not a " + RequestContextCurrentTraceContext.class
+                            .getSimpleName() + " scope. " +
+                    "Please call Tracing.Builder.currentTraceContext(" + RequestContextCurrentTraceContext.class
```
minwoox: Please call Tracing.newBuilder()?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,216 @@
...
+                            .getSimpleName() + ".INSTANCE).");
```
minwoox: .INSTANCE).build() ?

zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
```diff
@@ -67,6 +69,21 @@ public void tearDown() {
         Tracing.current().close();
     }
 
+    @Test
+    public void newDecorator_shouldFailFastWhenRequestContextCurrentTraceContextNotConfigured() {
+        assertThatThrownBy(() -> HttpTracingClient.newDecorator(Tracing.newBuilder().build()))
+                .isInstanceOf(IllegalStateException.class).hasMessage(
+                "Tracing.currentTraceContext is not a RequestContextCurrentTraceContext scope. " +
+                "Please call Tracing.Builder.currentTraceContext(RequestContextCurrentTraceContext.INSTANCE)."
```
minwoox: ditto

adriancole:
I didn't want to mention newBuilder.build because this will create an
incomplete instance of tracing. more intended to say how to patch where one
is already creating their builder.

ex there is no valid reason to call
tracing.newBuilder().currentTraceContext(something).build() as it is
missing a couple things most notably where to send the data (ex
builder.spanReporter(foo))

make sense?

minwoox:
> make sense?

Of course. Thanks for the explanation!

trustin:
Thanks a lot, @adriancole and reviewers!

adriancole:
heh should note @anuraag paired on this and was controlling my hands
also @kojilin helped review locally too thanks everyone!

adriancole:
https://github.com/openzipkin/brave/releases/tag/5.2.0 talks about this. feel free to reuse the content in any way you like

PR #1326, opened 2018-08-08
Add ArmeriaWebServer which can be a reactive web server for Spring Boot application

hyangtack:
Motivation:
Spring Boot provides a way to add a custom web server, so it would be helpful to let our Spring user have Armeria web server using `armeria-spring-boot-starter`.

Modifications:
- Add `ArmeriaWebServer` and its factory to `boot-webflux-autoconfigure`.
  - Common classes are copied from `boot-autoconfigure` to `boot-webflux-autoconfigure/gen-src` before building the module.
- Add `boot-webflux-starter`.
- Add ArmeriaClientHttpConnector which is used when creating WebClient.
- Add a document how to configure Armeria as a web server of a reactive Spring application.

Result:
- A user can start a reactive Armeria server with spring properties.


hyangtack:
This is WIP. I'm not familiar with Spring, so I want to get comments from Spring specialists about this feature. Is this way okay to go? Feel free to leave any comments about this PR! 😃 

/cc @imasahiro @taicki @anuraaga

spring/boot-autoconfigure/build.gradle
```diff
@@ -7,10 +20,34 @@ dependencies {
...
+    include '**/ArmeriaServerConfigurator.java'
+    include '**/ArmeriaSettings.java'
+    include '**/MeterIdPrefixFunctionFactory.java'
+    include '**/ThriftServiceUtils.java'
+}
+
+tasks.compileJava.dependsOn(copyFiles)
+
+project.sourceSets.all {
+    sourceSet -> sourceSet.java.srcDir file("${project.ext.genSrcDir}/${sourceSet.name}/java")
```
trustin: No need for this once you rebase. :wink:

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,281 @@
...
+        this.beanFactory = requireNonNull(beanFactory, "beanFactory");
+    }
+
+    @Override
+    public WebServer getWebServer(HttpHandler httpHandler) {
+        final ServerBuilder sb = new ServerBuilder();
+
+        // TODO(hyangtack) Change factory and DataBuffer to Armeria-specific ones later if necessary.
+        final ArmeriaHttpHandlerAdapter handler =
+                new ArmeriaHttpHandlerAdapter(httpHandler, new DefaultDataBufferFactory());
```
trustin: Curious if `DefaultDataBufferFactory` could be instantiated only once and reused.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,281 @@
...
+            configureTls(sb, ssl, getSslStoreProvider());
+            protocol = SessionProtocol.HTTPS;
+        } else {
+            protocol = SessionProtocol.HTTP;
+        }
+
+        final Http2 http2 = getHttp2();
+        if (http2 != null && !http2.isEnabled()) {
+            logger.warn(
+                    "Cannot disable HTTP/2 protocol from Armeria server. It will be enabled automatically.");
```
trustin: from -> for

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,281 @@
...
+    private <T> Optional<T> findBean(Class<T> clazz) {
+        final String[] names = beanFactory.getBeanNamesForType(clazz);
+        if (names.length == 0) {
+            return Optional.empty();
+        }
+        if (names.length == 1) {
+            return Optional.of(beanFactory.getBean(names[0], clazz));
+        }
+        throw new IllegalStateException("Too many " + clazz.getSimpleName() + " beans: " +
+                                        String.join(",", names) + " (expected: 1)");
```
trustin: `","` -> `", "` for prettier output

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,281 @@
...
+        return port;
+    }
+
+    private static void configureTls(ServerBuilder sb,
+                                     Ssl ssl, @Nullable SslStoreProvider sslStoreProvider) {
+        try {
+            if (sslStoreProvider == null &&
+                ssl.getKeyStore() == null && ssl.getTrustStore() == null) {
+                logger.warn("Configure TLS with a self-signed certificates " +
+                            "because there are no key/trust store specified.");
```
trustin: - Configure -> Configuring
- a self-signed certificate (singular form)
- because no key or trust store was specified

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaServerHttpRequest.java
```diff
@@ -0,0 +1,110 @@
...
+        httpHeaders.forEach(entry -> newHttpHeaders.add(entry.getKey().toString(), entry.getValue()));
+        return newHttpHeaders;
+    }
+
+    @Override
+    protected MultiValueMap<String, HttpCookie> initCookies() {
+        final MultiValueMap<String, HttpCookie> cookies = new LinkedMultiValueMap<>();
+        final List<String> values = req.headers().getAll(HttpHeaderNames.COOKIE);
+        values.stream()
+              .map(ServerCookieDecoder.STRICT::decode)
```
trustin: Probably better using LAX?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaServerHttpRequest.java
```diff
@@ -0,0 +1,110 @@
...
+    public String getMethodValue() {
+        return req.method().name();
+    }
+
+    @Override
+    public Flux<DataBuffer> getBody() {
+        return Flux.from(req).map(httpObject -> {
+            // Expect HttpData only because this method is to get the body of the request.
+            assert httpObject instanceof HttpData;
+            return factory.wrap(((HttpData) httpObject).array());
```
trustin: You need to take the `arrayOffset()` and `length()` into account.

```java
final HttpData data = (HttpData) httpObject;
final byte[] array = data.array();
return factory.wrap(array, data.arrayOffset(), data.length());
```

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/DummyAddressResolverGroup.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+
+import io.netty.resolver.AddressResolver;
+import io.netty.resolver.AddressResolverGroup;
+import io.netty.resolver.InetNameResolver;
+import io.netty.resolver.InetSocketAddressResolver;
+import io.netty.util.concurrent.EventExecutor;
+import io.netty.util.concurrent.Promise;
+
+// TODO(hyangtack) Copied from HttpClientSniTest. Move this to testing module?
```
trustin: Yes, please go ahead.

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ReactiveWebServerAutoConfigurationTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @Rule
+    public TestRule globalTimeout = new DisableOnDebug(new Timeout(10, TimeUnit.SECONDS));
+
+    @LocalServerPort
+    int port;
+
+    @Test
+    public void shouldGetHelloViaH1() throws Exception {
+        final HttpClient client =
+                clientFactory.newClient("none+h1://example.com:" + port, HttpClient.class);
```
trustin: Could use `HttpClient.newClient()`

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ReactiveWebServerAutoConfigurationTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+        final AggregatedHttpMessage response = client.get("/hello").aggregate().join();
+        assertThat(response.content().toStringUtf8()).isEqualTo("hello");
+    }
+
+    @Test
+    public void shouldGetHelloViaH2() throws Exception {
+        final HttpClient client =
+                clientFactory.newClient("none+h2://example.com:" + port, HttpClient.class);
```
trustin: Could use `HttpClient.newClient()`

spring/boot-autoconfigure/src/test/resources/config/application-test_reactive.yml
```diff
@@ -0,0 +1,17 @@
+server:
+  address: 127.0.0.1
+  ssl:
+    # Armeria will use a self-signed certificate when there's no keystore configuration.
+    enabled: true
+
+
+armeria:
+  ports:
```
trustin: Question: Now we do not even need to specify any ports because `server.address` will be picked up, right?

hyangtack: Yes, right. We don't need to specify `armeria.*` properties. :-)

hyangtack:
I think that it's not easy to fully implement servlet Armeria web server for now. To provide this new starter in phases, it'd be better to separate the reactive Armeria web server from the existing modules. So, let me move the changes in this PR to `boot-webflux-autoconfigure` (and `boot-starter-webflux-armeria`).

hyangtack:
I realized that we need to make `WebClient` based on Armeria because reactor-netty is excluded from dependencies. Let me do more work on that and will let you know if the PR is ready for review.

spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/SseIntegrationTests.java
```diff
@@ -0,0 +1,282 @@
...
+                        assertEquals("bar 1", event.comment());
+                        assertNull(event.event());
+                        assertNull(event.retry());
+                    })
+                    .thenCancel()
+                    .verify(Duration.ofSeconds(5L));
+    }
+
+    @Test // SPR-16494
+    @Ignore
```
hyangtack: Will be removed once https://github.com/line/armeria/pull/1350 is merged.

spring/boot-webflux-autoconfigure/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,2 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
+com.linecorp.armeria.spring.web.reactive.ArmeriaReactiveWebServerFactoryAutoConfiguration
```
trustin: Indentation for readability?

spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/SseIntegrationTests.java
```diff
@@ -0,0 +1,280 @@
...
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.MonoProcessor;
+import reactor.test.StepVerifier;
+
+@SuppressWarnings("all")
+/**
+ * @author Sebastien Deleuze
+ */
+public class SseIntegrationTests extends AbstractHttpHandlerIntegrationTests {
```
trustin: I had no idea we supported ServerSentEvents :laughing: /cc @tokuhirom

testing-internal/src/main/java/com/linecorp/armeria/testing/internal/MockAddressResolverGroup.java
```diff
@@ -0,0 +1,87 @@
...
+
+import io.netty.resolver.AddressResolver;
+import io.netty.resolver.AddressResolverGroup;
+import io.netty.resolver.InetNameResolver;
+import io.netty.resolver.InetSocketAddressResolver;
+import io.netty.util.concurrent.EventExecutor;
+import io.netty.util.concurrent.Promise;
+
+/**
+ * An {@link AddressResolverGroup} which returns an IP address resolved by the resolver function.
```
trustin: the resolver function -> a given resolver function?

testing-internal/src/main/java/com/linecorp/armeria/testing/internal/MockAddressResolverGroup.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+public final class MockAddressResolverGroup extends AddressResolverGroup<InetSocketAddress> {
+
+    private final Function<String, InetAddress> resolver;
+
+    /**
+     * Creates an {@link AddressResolverGroup} which always returns {@code 127.0.0.1} for any hostname.
+     */
+    public MockAddressResolverGroup() {
```
trustin: How about making the constructors private and adding static factory methods so that a user knows what it resolves to? e.g. `MockAddressResolverGroup.localhost()`, `always(String ip)` and `of(Function)`

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientHttpConnector.java
```diff
@@ -0,0 +1,123 @@
...
+        final String path = uri.getRawPath();
+        final String query = uri.getRawQuery();
+
+        checkArgument(!Strings.isNullOrEmpty(authority), "URI is not absolute: " + uri);
+        checkArgument(!Strings.isNullOrEmpty(path), "path is undefined: " + uri);
+
+        final HttpRequestWriter writer =
+                HttpRequest.streaming(com.linecorp.armeria.common.HttpMethod.valueOf(method.name()),
+                                      Strings.isNullOrEmpty(query) ? path : path + '?' + query);
+        final URI baseUri = new URI(Strings.isNullOrEmpty(scheme) ? authority : scheme + "://" + authority);
```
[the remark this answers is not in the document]
hyangtack: Fixed. Thanks!

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientHttpResponse.java
```diff
@@ -0,0 +1,98 @@
...
+    public Flux<DataBuffer> getBody() {
+        return body;
+    }
+
+    @Override
+    public HttpHeaders getHeaders() {
+        // Lazy copy the headers.
+        final HttpHeaders httpHeaders = new HttpHeaders();
+        headers.forEach(entry -> httpHeaders.add(entry.getKey().toString(), entry.getValue()));
+        return httpHeaders;
```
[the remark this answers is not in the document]
hyangtack: I referred to [ReactorClientHttpResponse.java](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpResponse.java), so I thought that it's called once. But it'd be more defensive to cache the result here. :-) Thanks!

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpHandlerAdapter.java
```diff
@@ -0,0 +1,71 @@
...
+        this.httpHandler = requireNonNull(httpHandler, "httpHandler");
+        this.factory = requireNonNull(factory, "factory");
+    }
+
+    Mono<Void> handle(ServiceRequestContext ctx, HttpRequest req, HttpResponseWriter res) {
+        final ArmeriaServerHttpRequest convertedRequest;
+        try {
+            convertedRequest = new ArmeriaServerHttpRequest(ctx, req, factory);
+        } catch (URISyntaxException e) {
+            logger.warn("{} Invalid URL", ctx, e);
```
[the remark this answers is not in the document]
hyangtack: Fixed

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,334 @@
...
+        }
+
+        final Http2 http2 = getHttp2();
+        if (http2 != null && !http2.isEnabled()) {
+            logger.warn(
+                    "Cannot disable HTTP/2 protocol for Armeria server. It will be enabled automatically.");
+        }
+
+        final InetAddress address = getAddress();
+        final int port = ensureValidPort(getPort());
```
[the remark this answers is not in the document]
hyangtack: It's set from [ReactiveWebServerFactoryCustomizer.java#L51](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveWebServerFactoryCustomizer.java#L51). So I'm not sure whether it's possible. Let me take a look.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,334 @@
...
+                                       meterIdPrefixFunctionFactory);
+        configureServerWithArmeriaSettings(sb, settings,
+                                           findBean(MeterRegistry.class).orElse(Metrics.globalRegistry),
+                                           findBeans(HealthChecker.class));
+
+        findBeans(ArmeriaServerConfigurator.class).forEach(configurator -> configurator.configure(sb));
+    }
+
+    private <T> Optional<T> findBean(Class<T> clazz) {
+        final String[] names = beanFactory.getBeanNamesForType(clazz);
```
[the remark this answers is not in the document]
hyangtack: Fixed

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpRequest.java
```diff
@@ -0,0 +1,107 @@
...
+public class ArmeriaServerHttpRequest extends AbstractServerHttpRequest {
+
+    private final ServiceRequestContext ctx;
+    private final HttpRequest req;
+    private final Flux<DataBuffer> body;
+
+    ArmeriaServerHttpRequest(ServiceRequestContext ctx,
+                             HttpRequest req,
+                             DataBufferFactory factory) throws URISyntaxException {
+        super(new URI(requireNonNull(req, "req").path()),
```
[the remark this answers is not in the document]
hyangtack: Fixed

spring/boot-webflux-autoconfigure/src/main/resources/META-INF/spring.factories
hyangtack: Yes, I referred to [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/resources/META-INF/spring.factories).

testing-internal/src/main/java/com/linecorp/armeria/testing/internal/MockAddressResolverGroup.java
hyangtack: Fixed

hyangtack:
Will address remaining comments tomorrow. :-)

spring/boot-webflux-autoconfigure/src/main/resources/META-INF/spring.factories
hyangtack: Fixed

testing-internal/src/main/java/com/linecorp/armeria/testing/internal/MockAddressResolverGroup.java
hyangtack: Good suggestion. :-)

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientHttpConnector.java
```diff
@@ -0,0 +1,119 @@
...
+        this(configurator, DataBufferFactories.DEFAULT);
+    }
+
+    /**
+     * Creates an {@link ArmeriaClientHttpConnector}.
+     *
+     * @param configurator the configurator to be used to build an {@link HttpClient}
+     * @param bufferFactory the factory to be used to create a {@link DataBuffer}
+     */
+    public ArmeriaClientHttpConnector(Consumer<HttpClientBuilder> configurator,
```
trustin: Which would be better in this case, `Function` or `Consumer`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpHandlerAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+        this.httpHandler = requireNonNull(httpHandler, "httpHandler");
+        this.factory = requireNonNull(factory, "factory");
+    }
+
+    Mono<Void> handle(ServiceRequestContext ctx, HttpRequest req, HttpResponseWriter res) {
+        final ArmeriaServerHttpRequest convertedRequest;
+        try {
+            convertedRequest = new ArmeriaServerHttpRequest(ctx, req, factory);
+        } catch (Exception e) {
+            logger.warn("{} Invalid URL: {}", ctx, req.path(), e);
```
trustin: How about `Invalid request path`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,335 @@
...
+        this.beanFactory = requireNonNull(beanFactory, "beanFactory");
+    }
+
+    @Override
+    public WebServer getWebServer(HttpHandler httpHandler) {
+        final ServerBuilder sb = new ServerBuilder();
+
+        final SessionProtocol protocol;
+        final Ssl ssl = getSsl();
+        if (ssl != null && ssl.isEnabled()) {
```
trustin: Should we warn if `ssl == null && ssl.isEnabled()`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,335 @@
...
+                        break;
+                    case WANT:
+                        sslBuilder.clientAuth(io.netty.handler.ssl.ClientAuth.OPTIONAL);
+                        break;
+                }
+            }
+
+            sb.tls(sslBuilder.build());
+        } catch (Exception e) {
+            throw new IllegalStateException(e);
```
trustin: How about adding a message like `"Failed to configure TLS: " + e`

spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/ReactiveWebServerAutoConfigurationTest.java
```diff
@@ -0,0 +1,178 @@
...
+import reactor.core.publisher.Mono;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest(
+        webEnvironment = WebEnvironment.RANDOM_PORT,
+        // https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-testing.html
+        // #boot-features-testing-spring-boot-applications-detecting-web-app-type
+        properties = "spring.main.web-application-type=reactive"
+)
+@ActiveProfiles({ "test_reactive" })
```
trustin: Perhaps we can omit the curly braces if there's only one element?

testing-internal/src/main/java/com/linecorp/armeria/testing/internal/MockAddressResolverGroup.java
```diff
@@ -0,0 +1,101 @@
...
+    public static MockAddressResolverGroup localhost() {
+        return always("127.0.0.1");
+    }
+
+    /**
+     * Creates an {@link AddressResolverGroup} which always returns the specified {@code ipAddress}
+     * for any hostname.
+     */
+    public static MockAddressResolverGroup always(String ipAddress) {
+        requireNonNull(ipAddress, "ipAddress");
```
trustin: How about adding an additional check using Netty's `NetUtil`?

```java
checkArgument(NetUtil.isValidIpV4Address(ip) || NetUtil.isValidIpV6Address(ip),
              "not an IP address: %s", ip);
```

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientHttpConnectorCustomizer.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * A customizer to customize an {@link HttpClientBuilder} for an {@link ArmeriaClientHttpConnector}.
+ */
+@FunctionalInterface
+public interface ArmeriaClientHttpConnectorCustomizer {
+    /**
+     * A default customizer which does nothing.
+     */
+    ArmeriaClientHttpConnectorCustomizer IDENTITY = builder -> builder;
```
trustin: Could we hide this from the public API? I guess a user will not find a user for this.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientHttpConnector.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    private ArmeriaClientHttpRequest createRequest(HttpMethod method, URI uri) {
+        final String scheme = uri.getScheme();
+        final String authority = uri.getRawAuthority();
+        final String path = uri.getRawPath();
+        final String query = uri.getRawQuery();
+
+        checkArgument(!Strings.isNullOrEmpty(authority), "URI is not absolute: " + uri);
```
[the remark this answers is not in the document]
hyangtack: Do you mean a request path like `http:///api/v4/groups/1`? If so, it will also be failed with the same error message. Actually I wanted to use a message `authority is undefined`, but [this test](https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientIntegrationTests.java#L734) checks the message, so I followed it.

core/src/main/java/com/linecorp/armeria/server/encoding/HttpEncodingService.java
```diff
@@ -32,8 +33,13 @@
 
 /**
  * Decorates a {@link Service} to apply HTTP encoding (e.g., gzip) to an {@link HttpService}.
- * HTTP encoding will be applied if the client specifies it, the response content type is a reasonable
- * type to encode, and the response either has no fixed content length or the length is larger than 1KB.
+ * HTTP encoding will be applied if:
+ * <ul>
+ *     <li>the client specifies it</li>
+ *     <li>the response content type is a reasonable type to encode</li>
+ *     <li>the request headers are also acceptable</li>
```
minwoox: Could remove also?

core/src/main/java/com/linecorp/armeria/server/encoding/HttpEncodingService.java
```diff
@@ -32,8 +33,13 @@
 
 /**
  * Decorates a {@link Service} to apply HTTP encoding (e.g., gzip) to an {@link HttpService}.
- * HTTP encoding will be applied if the client specifies it, the response content type is a reasonable
- * type to encode, and the response either has no fixed content length or the length is larger than 1KB.
+ * HTTP encoding will be applied if:
+ * <ul>
+ *     <li>the client specifies it</li>
+ *     <li>the response content type is a reasonable type to encode</li>
```
minwoox: I feel a little vague when I read the `reasonable` here, so how about
the response content type is one of the encodable types?

examples/spring-boot-webflux-starter/src/main/java/example/springframework/boot/armeria/HelloConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+@Configuration
+public class HelloConfiguration {
+
+    /**
+     * A user can configure a {@link Server} by providing an {@link ArmeriaServerConfigurator} bean.
+     */
+    @Bean
+    public ArmeriaServerConfigurator armeriaServerConfigurator() {
+        return builder -> { /* noop */ };
```
minwoox: How about `Customize the server using the given ServerBuilder.` like you do in the documentation?

examples/spring-boot-webflux-starter/src/test/java/example/springframework/boot/armeria/WebTestClientTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(
+        webEnvironment = WebEnvironment.RANDOM_PORT,
+        properties = "spring.main.web-application-type=reactive"
+)
+public class WebTestClientTest {
+
+    @RestController
+    static class HelloController {
```
trustin: Shouldn't we test `example.springframework.boot.armeria.HelloController` rather than this one?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/ArmeriaWebServer.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param server the Armeria server
+     * @param protocol the session protocol which is used for the primary port
+     * @param address the primary local address that the server will be bound to
+     * @param port the primary local port that the server will be bound to
+     */
+    public ArmeriaWebServer(Server server, SessionProtocol protocol, @Nullable InetAddress address, int port) {
+        this.server = requireNonNull(server, "server");
+        this.protocol = requireNonNull(protocol, "protocol");
+        this.address = address;
+        this.port = port;
```
minwoox: Just wondering if we need to validate the port here?

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,80 @@
...
+`spring-boot-webflux-starter` to the dependencies.
+
+For Maven:
+
+.. code-block:: xml
+
+    <dependency>
+        <groupId>com.linecorp.armeria</groupId>
+        <artifactId>spring-boot-webflux-starter</artifactId>
+    </dependency>
```
trustin: version?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/ArmeriaWebServer.java
```diff
@@ -0,0 +1,108 @@
...
+                    // Replace the specified port number with the primary port number.
+                    // Server#activePort doesn't return the first added port, so we need to find that.
+                    final Optional<ServerPort> port =
+                            server.activePorts().values().stream()
+                                  .filter(p -> p.protocols().contains(protocol))
+                                  .filter(p -> address == null ||
+                                               Arrays.equals(address.getAddress(),
+                                                             p.localAddress().getAddress().getAddress()))
+                                  .findFirst();
+                    assert port.isPresent() : "the primary port doest not exist";
```
minwoox: please add `.` because it's a sentence? :)

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,80 @@
...
+
+For Gradle:
+
+.. code-block:: groovy
+
+    dependencies {
+        compile 'com.linecorp.armeria:spring-boot-webflux-starter'
```
trustin: version?

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,80 @@
...
+
+    @Configuration
+    public class ArmeriaConfiguration {
+        /**
+         * A user can configure the server by providing an ArmeriaServerConfigurator bean.
+         */
+        @Bean
+        public ArmeriaServerConfigurator armeriaServerConfigurator() {
+            return builder -> {
+                // Customize the server using the given ServerBuilder.
```
trustin: How about adding some example builder calls?

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,80 @@
...
+
+    @Configuration
+    public class ArmeriaConfiguration {
+        /**
+         * A user can configure the client by providing an ArmeriaClientConfigurator bean.
+         */
+        @Bean
+        public ArmeriaClientConfigurator armeriaClientConfigurator() {
+            return builder -> {
+                // Customize the client using the given HttpClientBuilder.
```
trustin: How about adding some example builder calls?

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,80 @@
+.. _`Spring Boot Reference Guide`: https://docs.spring.io/spring-boot/docs/current/reference/html/
+.. _`armeria-examples`: https://github.com/line/armeria-examples
+
+.. _advanced-spring-webflux-integration:
+
+Using Armeria as a web server of a reactive Spring application
+==============================================================
+
+A user can use Armeria as a web server of his or her reactive Spring application by adding
+`spring-boot-webflux-starter` to the dependencies.
```
trustin: How about adding why this is useful to users? e.g. HTTP/2, leveraging all the nice bells and whistles in Armeria, etc.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ByteBufferBasedArmeriaBufferFactory.java
```diff
@@ -0,0 +1,72 @@
...
+public class ByteBufferBasedArmeriaBufferFactory
+        extends AbstractArmeriaBufferFactory<DefaultDataBufferFactory> {
+
+    ByteBufferBasedArmeriaBufferFactory(DefaultDataBufferFactory delegate) {
+        super(delegate);
+    }
+
+    @Override
+    protected DataBuffer wrap0(ByteBufHttpData httpData) {
+        return delegate().wrap(ByteBufUtil.getBytes(httpData.content()));
```
trustin: We could check if `httpData.content().hasArray()` is true and avoid making a copy if so.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ByteBufferBasedArmeriaBufferFactory.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public HttpData unwrap(DataBuffer dataBuffer) {
+        requireNonNull(dataBuffer, "dataBuffer");
+        final ByteBuffer src =
+                dataBuffer instanceof DefaultDataBuffer ? ((DefaultDataBuffer) dataBuffer).getNativeBuffer()
+                                                        : dataBuffer.asByteBuffer();
+        if (src.hasArray()) {
+            return HttpData.of(src.array());
```
trustin: `HttpData.of(src.array(), src.arrayOffset() + src.position(), src.remaining())` ?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ByteBufferBasedArmeriaBufferFactory.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        final byte[] dst = new byte[src.remaining()];
```
trustin: We can take advantage of `Unsafe` here:

```java
final byte[] dst = PlatformDependent.allocateUninitializedArray(src.remaining());
```

or we could avoid memory copy completely by doing this:

```java
new ByteBufHttpData(Unpooled.wrappedBuffer(src), false);
```

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ByteBufferBasedArmeriaBufferFactory.java
```diff
@@ -0,0 +1,72 @@
...
+        src.get(dst);
```
trustin: If we retrieved `src` from `DefaultDataBuffer.getNativeBuffer()`, wouldn't this affect the behavior of `dataBuffer` in an unexpected way? I think it's fine if Spring does not touch `dataBuffer` at all after unwrapping though.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/DefaultSslInfo.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        return sb.toString();
+    }
+
+    @Nullable
+    private static X509Certificate[] initCertificates(SSLSession session) {
+        Certificate[] certificates;
+        try {
+            certificates = session.getPeerCertificates();
+        } catch (Throwable ex) {
```
trustin: `Exception`, because we do not want to swallow a critical exceptions like OOME here.

spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientHttpRequestTest.java
```diff
@@ -0,0 +1,175 @@
...
+                    .expectNext("d").thenRequest(1)
+                    .expectNext("e").thenRequest(1)
+                    .expectComplete()
+                    .verify();
+
+        await().until(() -> httpRequest.completionFuture().isDone());
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
```
trustin: Could we move this annotation inside so we suppress only where necessary?

spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactoryTest.java
```diff
@@ -0,0 +1,236 @@
...
+    public void shouldReturnBadRequestDueToException() {
+        final ArmeriaReactiveWebServerFactory factory = factory();
+        runServer(factory, AlwaysFailureHandler.INSTANCE, server -> {
+            final HttpClient client = httpClient(server);
+
+            final AggregatedHttpMessage res1 = client.post("/hello", "hello").aggregate().join();
+            assertThat(res1.status()).isEqualTo(com.linecorp.armeria.common.HttpStatus.BAD_REQUEST);
+
+            final AggregatedHttpMessage res = client.get("/hello").aggregate().join();
+            assertThat(res.status()).isEqualTo(com.linecorp.armeria.common.HttpStatus.BAD_REQUEST);
```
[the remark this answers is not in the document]
trustin: or `res2`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/AbstractArmeriaBufferFactory.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public final DataBuffer wrap(HttpData httpData) {
+        requireNonNull(httpData, "httpData");
+        if (httpData instanceof ByteBufHttpData) {
+            return wrap0((ByteBufHttpData) httpData);
+        }
+        if (httpData instanceof DefaultHttpData) {
+            return wrap0((DefaultHttpData) httpData);
```
trustin: How about changing:

- `wrap0(ByteBufHttpData)` to `wrapByteBufHttpData(ByteBufHttpData)`
- `wrap0(DefaultHttpData)` to `wrapHttpData(Httpdata)`

and handle any `HttpData` implementations that are not `ByteBufHttpData` in `wrapHttpData()`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/NettyBasedArmeriaBufferFactory.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+class NettyBasedArmeriaBufferFactory extends AbstractArmeriaBufferFactory<NettyDataBufferFactory> {
+
+    NettyBasedArmeriaBufferFactory(NettyDataBufferFactory delegate) {
+        super(delegate);
+    }
+
+    @Override
+    protected DataBuffer wrap0(ByteBufHttpData httpData) {
+        return delegate().wrap(httpData.content());
```
trustin: How and when the `ByteBuf` returned by `httpData.content()` is released?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/NettyBasedArmeriaBufferFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public HttpData unwrap(DataBuffer dataBuffer) {
+        requireNonNull(dataBuffer, "dataBuffer");
+        if (dataBuffer instanceof NettyDataBuffer) {
+            return new ByteBufHttpData(((NettyDataBuffer) dataBuffer).getNativeBuffer(), false);
+        }
+
+        final ByteBuf byteBuf = delegate().getByteBufAllocator().buffer(dataBuffer.readableByteCount());
+        byteBuf.writeBytes(dataBuffer.asByteBuffer());
```
trustin: Can we wrap without making a copy?

```java
Unpooled.wrappedBuffer(dataBuffer.asByteBuffer());
```

examples/spring-boot-webflux-starter/src/main/resources/config/application.yml
```diff
@@ -0,0 +1,11 @@
+spring.profiles.active: local
```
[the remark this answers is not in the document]
hyangtack: When I added the property to [here](https://github.com/line/armeria/pull/1326/files#diff-9f44dfc8409d0632e6ce30ab22e1729cR74), the dependencies for `reactive` and `servlet` types existed together in our module at that time. 

But now it can be removed because Spring resolves the type automatically when `spring-webflux` module is only used.

So, we also do not have to specify `spring.main.web-application-type: reactive` here. I will also remove the property in the test code. Thanks! :-)

examples/spring-boot-webflux-starter/src/test/java/example/springframework/boot/armeria/WebTestClientTest.java
hyangtack: Oops, my mistake. I copied the test code without thinking. Thanks!

trustin:
@rstoyanchev Could you sign the ICLA now that your documentation update is a part of this pull request?

site/src/sphinx/advanced-spring-webflux-integration.rst
hyangtack: I added an example builder calls in the `example`: ...g-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java

Is it okay to copy that to here? What do you think?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,352 @@
...
+
+        final Predicate<HttpHeaders> encodableRequestHeadersPredicate;
+        final String[] excludedUserAgents = compression.getExcludedUserAgents();
+        if (excludedUserAgents == null || excludedUserAgents.length == 0) {
+            encodableRequestHeadersPredicate = headers -> true;
+        } else {
+            encodableRequestHeadersPredicate = headers -> {
+                final String userAgent = headers.get(HttpHeaderNames.USER_AGENT);
+                return !Strings.isNullOrEmpty(userAgent) &&
+                       Arrays.stream(excludedUserAgents).noneMatch(userAgent::equalsIgnoreCase);
```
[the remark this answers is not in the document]
hyangtack: Agreed. I referred to https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/netty/CompressionCustomizer.java#L82. But the other implementation seems like using pattern. Will fix soon!

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+@Configuration
+public class HelloConfiguration {
+
+    /**
+     * A user can configure a {@link Server} by providing an {@link ArmeriaServerConfigurator} bean.
+     */
+    @Bean
+    public ArmeriaServerConfigurator armeriaServerConfigurator() {
+        // An example which adds a LoggingService decorator to the ServerBuilder in order to log every request.
+        return builder -> builder.decorator(LoggingService.newDecorator())
```
trustin: How about adding an additional example that shows:

- a user can bind a gRPC or Thrift service as a comment
- HAPROXY support
- DocService

..?

e.g.

```
builder -> {
    builder.serviceUnder("/docs", new DocService())
           .decorator(...)
           .accessLogWriter(...);

    // You can also bind an asynchronous RPC services such as gRPC and Thrift:
    // builder.service(new GrpcServiceBuilder() .... build());
    // builder.service(THttpService.of(...));

    return builder;
}
```

examples/spring-boot-webflux/src/main/resources/config/application.yml
```diff
@@ -0,0 +1,11 @@
+spring.profiles.active: local
+server:
+  port: 8443
+  ssl:
+    enabled: true # Use self-signed certificate because no key-store configuration is specified.
```
trustin: Use -> Uses

examples/spring-boot-webflux/src/main/resources/config/application.yml
```diff
@@ -0,0 +1,11 @@
...
+
+# Additional server ports specified with Armeria configuration.
+armeria:
+  ports:
+  - port: 8443    # Port 8443 will serve both HTTPS and HTTP.
```
trustin: This is interesting. Is this an expected behavior to serve both HTTP and HTTPS on 8443 in this case? I guess a user may expect HTTPS only?

examples/spring-boot-webflux/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,11 @@
...
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <root level="debug">
+    <appender-ref ref="STDOUT" />
+  </root>
+</configuration>
```
trustin: no eol at the end

site/src/sphinx/advanced-spring-webflux-integration.rst
trustin: Yes, let's make a copy here.. Can't think of a better way yet.

site/src/sphinx/advanced-spring-webflux-integration.rst
trustin: You can insert the real version number instead of `x.y.z`. See https://github.com/line/armeria/blob/master/site/src/sphinx/setup-maven.rst

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. note::
+
+    Visit `armeria-examples <https://github.com/line/armeria-examples>`_ to find the fully working example code.
+
+Spring framework provides powerful features which are necessary for building a web application, such as
+dependency injection, data binding, AOP, transaction, etc. In addition, if your Spring application integrates
+with Armeria, you can leverage the following:
+
+- rich support for Apache `Thrift <https://thrift.apache.org/>`_ and `gRPC <https://grpc.io/>`_,
```
trustin: - We could capitalize all first words in the items.

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,95 @@
...
+  including the documentation service
+- a way to run HTTP REST service and RPC service in the same port
```
trustin: A way -> Ability

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,95 @@
...
+.. note::
+
+    Visit `armeria-examples <https://github.com/line/armeria-examples>`_ to find the fully working example code.
+
+Spring framework provides powerful features which are necessary for building a web application, such as
+dependency injection, data binding, AOP, transaction, etc. In addition, if your Spring application integrates
+with Armeria, you can leverage the following:
+
+- rich support for Apache `Thrift <https://thrift.apache.org/>`_ and `gRPC <https://grpc.io/>`_,
+  including the documentation service
```
trustin: How about:

- including [the fancy web console](https://github.com/line/armeria/blob/master/site/src/sphinx/server-docservice.rst) that enables you to send Thrift and gRPC requests from web browser.

site/src/sphinx/advanced-spring-webflux-integration.rst
trustin: Ditto - let's make a copy.

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,95 @@
...
+                // Customize the client using the given HttpClientBuilder.
+            }
+        }
+    }
+
+.. note::
+
+    If you are not familiar with Spring Boot and Spring WebFlux, please refer to
+    `Spring Boot Reference Guide <https://docs.spring.io/spring-boot/docs/current/reference/html/>`_ and
+    `Spring Framework Documentation <https://docs.spring.io/spring/docs/current/spring-framework-reference/>`_.
```
trustin: Shouldn't we link here - https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html ? (because we said 'Spring Boot and Spring WebFlux')

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AbstractServiceRegistrationBean.java
```diff
@@ -30,10 +30,10 @@
 import com.linecorp.armeria.server.Service;
 
 /**
- * An abstract bean with information for registering an annotated service object. It enables micrometer
- * monitoring of the service automatically.
+ * An abstract bean with information for registering a service object of type {@code T} using a builder
+ * of type {@code U}. It enables micrometer monitoring of the service automatically.
```
trustin: You can use `@param <T>` instead and keep the first sentence short.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/NettyBasedArmeriaBufferFactory.java
hyangtack: Spring framework will consume it if a user wants to read HTTP body, then release the `DataBuffer` which is returned here using https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java#L406.

According to the code, the `DataBuffer` instance, which holds a `ByteBuf`, should be a subclass of `PooledDataBuffer`. 

So, I left comment on https://github.com/line/armeria/pull/1326/files#diff-4938ea1a82e4c011c7439b3a255b8a16R89 and made the return type `PooledDataBuffer` explicitly.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -0,0 +1,352 @@
...
+
+    @Nullable
+    private static KeyStore loadKeyStore(
+            @Nullable String type,
+            @Nullable String resource,
+            @Nullable String password) throws IOException, GeneralSecurityException {
+        if (resource == null) {
+            return null;
+        }
+        final KeyStore store = KeyStore.getInstance(firstNonNull(type, "JKS"));
```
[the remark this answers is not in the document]
hyangtack: It seems that making Spring user confused, because they are familiar with the current way which is based on keystore. So it would be better to improve this later, by extending `ArmeriaSettings`.

core/src/test/java/com/linecorp/armeria/server/ServerBuilderTest.java
```diff
@@ -16,29 +16,54 @@
...
+        final Server server = new ServerBuilder()
+                .http(8080)
+                .https(8080)
+                .tlsSelfSigned()
+                .service("/", (ctx, req) -> HttpResponse.of(HttpStatus.OK))
+                .build();
+
+        final List<ServerPort> ports = server.config().ports();
+        assertThat(ports.size()).isOne(); // merged
+        assertThat(Iterables.getFirst(ports, null).protocols())
```
minwoox: `ports.get(0)`?

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -24,21 +33,42 @@
...
     public ArmeriaServerConfigurator armeriaServerConfigurator() {
-        // An example which adds a LoggingService decorator to the ServerBuilder in order to log every request.
-        return builder -> builder.decorator(LoggingService.newDecorator())
-                                 .accessLogWriter(AccessLogWriter.combined(), false);
+        // Customize the server using the given ServerBuilder. For example:
+        return builder -> {
+            // Add DocService that enables you to send Thrift and gRPC requests from web browser.
+            builder.serviceUnder("/docs", new DocService());
+
+            // Logging every request and response which is received by the server.
```
minwoox: `Logging` -> `Log` because we used `add` and `write`?

`which are received`? or `which the server receives.`?

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -24,21 +33,42 @@
...
-        return builder -> builder.factory(
-                new ClientFactoryBuilder()
-                        .sslContextCustomizer(b -> b.trustManager(InsecureTrustManagerFactory.INSTANCE))
-                        .build());
+        // Customize the client using the given HttpClientBuilder. For example:
+        return builder -> {
+            // Use circuit breaker for every endpoint.
+            final Function<String, CircuitBreaker> factory = key -> CircuitBreaker.of("my-cb-" + key);
+            final CircuitBreakerStrategy strategy = CircuitBreakerStrategy.onServerErrorStatus();
+            builder.decorator(CircuitBreakerHttpClient.newPerHostDecorator(factory, strategy));
```
minwoox: We can just do `builder.decorator(new CircuitBreakerHttpClientBuilder(strategy).build())` and remove the `factory` part up there. :)

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -24,21 +33,42 @@
...
+
+            // Automatically retry a request when the server returns a 5xx response.
+            builder.decorator(RetryingHttpClient.newDecorator(RetryStrategy.onServerErrorStatus()));
+
+            // Use a custom client factory in order not to validate a certificate received from the server.
```
minwoox: in order to disable the certificate validation?

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -11,10 +11,11 @@ Spring framework provides powerful features which are necessary for building a w
...
 with Armeria, you can leverage the following:
 
-- rich support for Apache `Thrift <https://thrift.apache.org/>`_ and `gRPC <https://grpc.io/>`_,
+- Rich support for Apache `Thrift <https://thrift.apache.org/>`_ and `gRPC <https://grpc.io/>`_,
   including the documentation service
-- a way to run HTTP REST service and RPC service in the same port
-- full HTTP/2 support for both server-side and client-side, including HTTP/2 on cleartext connections
+- Ability to run HTTP REST service and RPC service in the same port
+- Full HTTP/2 support for both server-side and client-side, including :ref:`the fancy web console <server-docservice>`
+  that enables you to send Thrift and gRPC requests from web browser.
```
minwoox: a web browser?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1171,6 +1169,38 @@ public Server build() {
         return server;
     }
 
+    /**
+     * Returns a list of {@link ServerPort} which consists of distinct port numbers except for the port
```
minwoox: `{@link ServerPort}s`?

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
hyangtack: Thank you! I didn't notice the default `CircuitBreakerMapping`. 😆

trustin:
Thank you for signing ICLA, @rstoyanchev.

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Spring framework provides powerful features which are necessary for building a web application, such as
+dependency injection, data binding, AOP, transaction, etc. In addition, if your Spring application integrates
+with Armeria, you can leverage the following:
+
+- Rich support for Apache `Thrift <https://thrift.apache.org/>`_ and `gRPC <https://grpc.io/>`_,
+  including the documentation service
+- Ability to run HTTP REST service and RPC service in the same port
+- Full HTTP/2 support for both server-side and client-side, including :ref:`the fancy web console <server-docservice>`
+  that enables you to send Thrift and gRPC requests from a web browser.
```
trustin: - The current 'including ...' clause should be in the first bullet about Thrift and gRPC support.
- 'including `h2c` (plaintext HTTP/2)' ?

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,122 @@
...
+- `PROXY protocol <https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt>`_ support which is
+  interoperable with load balancers such as `HAProxy <https://www.haproxy.org/>`_ and
```
trustin: whish provides interoperability with?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
```diff
@@ -66,67 +68,76 @@
...
+         * Register a list of {@link SessionProtocol}s that the {@link Server} uses.
+         */
+        public Port setProtocols(List<SessionProtocol> protocols) {
+            this.protocols = protocols;
+            return this;
         }
 
         /**
-         * Register a {@link SessionProtocol} that {@link Server} use.
+         * Register a {@link SessionProtocol} that the {@link Server} uses.
```
trustin: Registers

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
```diff
@@ -66,67 +68,76 @@
...
          */
-        public SessionProtocol getProtocol() {
-            return protocol;
+        @Nullable
+        public List<SessionProtocol> getProtocols() {
+            return protocols;
+        }
+
+        /**
+         * Register a list of {@link SessionProtocol}s that the {@link Server} uses.
```
trustin: Registers

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/DataBufferFactoryWrapper.java
```diff
@@ -0,0 +1,105 @@
...
+    DataBufferFactoryWrapper(T delegate) {
+        this.delegate = requireNonNull(delegate, "delegate");
+        converter = delegate instanceof NettyDataBufferFactory ? this::withNettyDataBufferFactory
+                                                               : this::withDataBufferFactory;
+    }
+
+    /**
+     * Returns the underlying {@link DataBufferFactory}.
+     */
+    public T dataBufferFactory() {
```
trustin: `delegate()`?

site/src/sphinx/advanced-spring-webflux-integration.rst
hyangtack: Oops, my mistake. Thanks!

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+    @Bean
+    public ArmeriaClientConfigurator armeriaClientConfigurator() {
+        // Customize the client using the given HttpClientBuilder. For example:
+        return builder -> {
+            // Use a circuit breaker for each remote host.
+            final CircuitBreakerStrategy strategy = CircuitBreakerStrategy.onServerErrorStatus();
+            builder.decorator(new CircuitBreakerHttpClientBuilder(strategy).newDecorator());
+
+            // Automatically retry a request when the server returns a 5xx response.
+            builder.decorator(RetryingHttpClient.newDecorator(RetryStrategy.onServerErrorStatus()));
```
trustin: I guess RetryingHttpClient should decorate before CircuitBreakerHttpClient decorates, or only one of them should be used?

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+
+            // Use a custom client factory in order to disable the certificate validation,
+            // which means any certificate received from the server will be accepted.
+            final ClientFactory clientFactory = new ClientFactoryBuilder().sslContextCustomizer(
+                    b -> b.trustManager(InsecureTrustManagerFactory.INSTANCE)).build();
```
trustin: This client factory will not be cleaned up when application context is terminated. How about providing it as a bean so that Spring calls `ClientFactory.close()` for us when the application terminates?

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,122 @@
...
+        @Bean
+        public ArmeriaClientConfigurator armeriaClientConfigurator() {
+            // Customize the client using the given HttpClientBuilder. For example:
+            return builder -> {
+                // Use a circuit breaker for each remote host.
+                final CircuitBreakerStrategy strategy = CircuitBreakerStrategy.onServerErrorStatus();
+                builder.decorator(new CircuitBreakerHttpClientBuilder(strategy).newDecorator());
+
+                // Automatically retry a request when the server returns a 5xx response.
+                builder.decorator(RetryingHttpClient.newDecorator(RetryStrategy.onServerErrorStatus()));
```
trustin: Ditto

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -0,0 +1,122 @@
...
+
+                // Use a custom client factory in order to disable the certificate validation,
+                // which means any certificate received from the server will be accepted.
+                final ClientFactory clientFactory = new ClientFactoryBuilder().sslContextCustomizer(
+                        b -> b.trustManager(InsecureTrustManagerFactory.INSTANCE)).build();
```
trustin: and ditto.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientHttpResponse.java
```diff
@@ -0,0 +1,133 @@
...
+        return this.httpHeaders;
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this).omitNullValues()
+                          .add("status", status)
+                          .add("headers", headers)
+                          .add("cookies", cookies)
+                          .add("httpHeaders", httpHeaders)
```
trustin: Is this for a debugging purpose? I guess a user will see both `headers` and `httpHeaders` which may be confusing.

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
hyangtack: As discussed in LINE chat room, let me remove retrying client. :-)

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -58,14 +56,18 @@ public ArmeriaClientConfigurator armeriaClientConfigurator() {
...
-            // Automatically retry a request when the server returns a 5xx response.
-            builder.decorator(RetryingHttpClient.newDecorator(RetryStrategy.onServerErrorStatus()));
-
-            // Use a custom client factory in order to disable the certificate validation,
-            // which means any certificate received from the server will be accepted.
-            final ClientFactory clientFactory = new ClientFactoryBuilder().sslContextCustomizer(
-                    b -> b.trustManager(InsecureTrustManagerFactory.INSTANCE)).build();
-            builder.factory(clientFactory);
+            // Set a custom client factory.
+            builder.factory(clientFactory());
```
trustin: Probably better letting Spring inject the `ClientFactory` as a parameter of `armeriaClientConfigurator()` method?

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -103,16 +103,20 @@ in your configuration as follows:
...
-                // Automatically retry a request when the server returns a 5xx response.
-                builder.decorator(RetryingHttpClient.newDecorator(RetryStrategy.onServerErrorStatus()));
-
-                // Use a custom client factory in order to disable the certificate validation,
-                // which means any certificate received from the server will be accepted.
-                final ClientFactory clientFactory = new ClientFactoryBuilder().sslContextCustomizer(
-                        b -> b.trustManager(InsecureTrustManagerFactory.INSTANCE)).build();
-                builder.factory(clientFactory);
+                // Set a custom client factory.
+                builder.factory(clientFactory());
```
trustin: Ditto

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -58,14 +56,18 @@ public ArmeriaClientConfigurator armeriaClientConfigurator() {
...
         };
     }
+
+    /**
+     * Returns a custom client factory which is configured as disabling the certificate validation,
+     * which means any certificate received from the server will be accepted.
+     */
+    @Bean
+    public ClientFactory clientFactory() {
```
trustin: .. and for readability, we could move this method before `armeriaClientConfigurator()`.

site/src/sphinx/advanced-spring-webflux-integration.rst
```diff
@@ -103,16 +103,20 @@ in your configuration as follows:
...
             };
         }
+
+        /**
+         * Returns a custom client factory which is configured as disabling the certificate validation,
+         * which means any certificate received from the server will be accepted.
+         */
+        @Bean
+        public ClientFactory clientFactory() {
```
trustin: Ditto

trustin:
@imasahiro Care to review again?

PR #1327, opened 2018-08-09
Add a new flag: 'com.linecorp.armeria.verboseResponses'

trustin:
Motivation:

Exposing the server-side stack trace in a response can be a security
concern depending on server configuration.

Modifications:

- Add a new system property called `com.linecorp.armeria.verboseResponses`
  so that a user can go back to the previous behavior.
- Do not include stack traces in `THttpService` responses by default.
- Include stack traces in `GrpcService` responses if the flag is set.
- Miscellaneous:
  - Un-deprecate `Exceptions.traceText()` and make it always use `\n` as a line delimiter
  - Replace `Throwables.getStackTraceAsString()` and `System.lineSeparator()` with
    `Exceptions.traceText()` and `\n`
Result:

- A little bit more out-of-the-box security
- gRPC response stack traces are sent if enabled.
- Platform independent stack trace responses

@anuraaga Anything to change on gRPC side? I can make a fix here.

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
+++ b/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
@@ -38,8 +38,10 @@ import org.slf4j.LoggerFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
+import com.google.common.base.Throwables;
 
 import com.linecorp.armeria.common.DefaultHttpHeaders;
+import com.linecorp.armeria.common.Flags;
 import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpHeaders;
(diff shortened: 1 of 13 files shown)


thrift/src/main/java/com/linecorp/armeria/server/thrift/THttpService.java
```diff
@@ -420,10 +421,15 @@ protected HttpResponse doPost(ServiceRequestContext ctx, HttpRequest req) {
         ctx.logBuilder().deferRequestContent();
         req.aggregateWithPooledObjects(ctx.eventLoop(), ctx.alloc()).handle(voidFunction((aReq, cause) -> {
             if (cause != null) {
-                responseFuture.complete(
-                        HttpResponse.of(
-                                HttpStatus.INTERNAL_SERVER_ERROR,
-                                MediaType.PLAIN_TEXT_UTF_8, Throwables.getStackTraceAsString(cause)));
+                final HttpResponse errorRes;
+                if (!Flags.verboseResponses()) {
```
hyangtack: nit: How about flipping the condition? (`if (Flags.verboseResponses())`)

grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
```diff
@@ -455,14 +457,32 @@ static HttpHeaders statusToTrailers(Status status, boolean headersSent) {
...
+    private static String statusToDescription(Status status) {
+        final String description = status.getDescription();
+        if (Flags.verboseResponses()) {
+            final Throwable cause = status.getCause();
+            if (cause != null) {
+                final String trace = Throwables.getStackTraceAsString(cause);
+                if (Strings.isNullOrEmpty(description)) {
+                    return "Caused by: " + trace;
+                } else {
+                    return description + System.lineSeparator() + "Caused by: " + trace;
```
anuraaga: Oops one point - I'd go with `\n` instead of `System`

trustin: I thought about this and realized `Throwables.getStackTraceAsString()` uses platform dependent line ending, so just used `System.lineSeparator()` for consistency. I think I can add something similar to `com.linecorp.armeria.common.util.Exceptions` to produce a platform-independent output. Sounds good?

trustin: - Un-deprecated `Exceptions.traceText()` and made it always use `\n` as a line delimiter.
- Replaced all `Throwables.getStackTraceAsString()` and most `System.lineSeparator()` calls

trustin:
Thanks for reviewing :-)

PR #1329, opened 2018-08-09
More strict path validation

### LINKED ISSUE #1323: Provides more strict path pattern validation to PathAndQuery
If a user send a HTTP request to `\n` contained uri with `HttpClient`, `HttpClient` requests the uri without validation or url-encoding and it breaks HTTP protocol like 
```
GET /foo\n
bar HTTP/1.1\r\n
```

Currently, we have few path validation to uri path part but it woud be required to provide more strict parsing/validating uri path in for example, `PathAndQuery`.

```java
    @Test
    public void parse_newline() {
        assertThat(PathAndQuery.parse("/a\nb")).isNull();
        // should be null but currently, return { path: "/a\nb", query: null }
    }
```
cc: @mauhiz

trustin:
Motivation:

- A user can send a request to an invalid URL using HttpClient:

  ```java
  HttpClient.newClient(...).get("/a\nb");
  ```

- A server accepts a path that contains control characters.
- A server does not filter out the paths with double dots when they are
  encoded, e.g. `/%2e%2e/%2e%2e/secret.txt`

Modifications:

- Reject control characters
- Decode percent-encoded characters before complex validation
- Encode a raw Unicode string to UTF-8
- Various optimizations
  - Slash deduplication while decoding
  - Percent encoding validation while decoding
- Add test cases

Result:

- Fixes #1323
- A user cannot send an invalid request with Armeria.
- Armeria filters out more potentially malicious requests.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
@@ -35,7 +35,8 @@ import com.linecorp.armeria.common.metric.MeterIdPrefix;
 import com.linecorp.armeria.internal.metric.CaffeineMetricSupport;
 
 import io.micrometer.core.instrument.MeterRegistry;
-import it.unimi.dsi.fastutil.bytes.ByteArrayList;
+import it.unimi.dsi.fastutil.Arrays;
+import it.unimi.dsi.fastutil.bytes.ByteArrays;
 
 /**
  * A parser of the raw path and query components of an HTTP path. Performs validation and allows caching of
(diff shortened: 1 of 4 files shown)


trustin:
This needs more work. Will ping once it's ready.

trustin:
Now ready for a review.

anuraaga:
Would you be able to add to or just run https://github.com/line/armeria/blob/master/benchmarks/src/jmh/java/com/linecorp/armeria/internal/PathParsingBenchmark.java#L37 to see what happens? Curious how the performance changes.

trustin:
@anuraaga Sure. Will get back with the result soon.

trustin:
Let me try to optimize a little bit more taking advantage of this chance.

trustin:
Thanks, reviewers. :-)

PR #1330, opened 2018-08-10
Add examples of annotated HTTP service

hyangtack:
Motivation:
We provide official document for annotated HTTP service, but it is not ready-to-use on IDE. So, it would be useful to beginners if we provide working examples.

Modifications:
- Add `example.armeria.server` package

### DIFF
--- a/examples/annotated-http-service/src/main/java/example/armeria/server/annotated/ExceptionHandlerService.java
+++ b/examples/annotated-http-service/src/main/java/example/armeria/server/annotated/ExceptionHandlerService.java
@@ -16,6 +16,9 @@ import example.armeria.server.annotated.ExceptionHandlerService.GlobalExceptionH
 
 /**
  * Examples how to use {@link ExceptionHandler}.
+ *
+ * @see <a href="https://line.github.io/armeria/server-annotated-service.html#handling-exceptions">
+ *      Handling exceptions</a>
  */
 @LoggingDecorator(
         requestLogLevel = LogLevel.INFO,            // Log every request sent to this service at INFO level.
@@ -88,6 +91,7 @@ public class ExceptionHandlerService {
             if (cause instanceof GloballyGeneralException) {
                 return HttpResponse.of(HttpStatus.FORBIDDEN);
             }
+            // To the next exception handler.
             return ExceptionHandlerFunction.fallthrough();
         }
     }
@@ -101,6 +105,7 @@ public class ExceptionHandlerService {
             if (cause instanceof LocallyGeneralException) {
                 return HttpResponse.of(HttpStatus.BAD_REQUEST);
             }
+            // To the next exception handler.
             return ExceptionHandlerFunction.fallthrough();
         }
     }
--- a/examples/annotated-http-service/src/main/java/example/armeria/server/annotated/ExceptionHandlerService.java
+++ b/examples/annotated-http-service/src/main/java/example/armeria/server/annotated/ExceptionHandlerService.java
@@ -1,0 +1,107 @@
+package example.armeria.server.annotated;
+
+import com.linecorp.armeria.common.HttpRequest;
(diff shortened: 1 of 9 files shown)


hyangtack:
This is WIP for getting comments from others. 😃 Will ping if this is ready to review.

examples/spring-boot-tomcat/src/main/java/example/armeria/server/MessageConverterService.java
```diff
@@ -0,0 +1,94 @@
...
+import com.linecorp.armeria.server.annotation.RequestObject;
+import com.linecorp.armeria.server.annotation.ResponseConverter;
+import com.linecorp.armeria.server.annotation.decorator.LoggingDecorator;
+
+/**
+ * Examples how to use {@link RequestConverter} and {@link ResponseConverter}.
+ */
+@LoggingDecorator(
+        requestLogLevel = LogLevel.INFO,            // Log every request sent to this service as INFO level.
+        successfulResponseLogLevel = LogLevel.INFO  // Log every response sent from this service as INFO level.
```
trustin: as -> at

hyangtack:
> Could we create a separate example project for this, which does not involve Spring?

Of course, I misunderstood the structure of the example project. :-)

hyangtack:
@minwoox Ready for review. :-)

examples/annotated-http-service/src/main/java/example/armeria/server/annotated/ExceptionHandlerService.java
```diff
@@ -0,0 +1,107 @@
...
+     * Throws a {@link GloballyGeneralException} which can be handled by the {@link GlobalExceptionHandler}.
+     */
+    @Get("/globallyGeneral")
+    @ExceptionHandler(LocalExceptionHandler.class)
+    public String exception3() {
+        throw new GloballyGeneralException();
+    }
+
+    /**
+     * Throws an {@link IllegalArgumentException} which can be handled by the default exception handler.
```
minwoox: I'm wondering whether we need to tell the reader more about default exception handlers such as what kinds of exceptions they handles, and what happens if there's a specific exception that default exception handlers cannot handle.
Probably overkill?

examples/annotated-http-service/src/main/java/example/armeria/server/annotated/ExceptionHandlerService.java
```diff
@@ -0,0 +1,107 @@
...
+        private static final long serialVersionUID = 5879797158322612975L;
+    }
+
+    static final class GlobalExceptionHandler implements ExceptionHandlerFunction {
+        @Override
+        public HttpResponse handleException(RequestContext ctx, HttpRequest req, Throwable cause) {
+            if (cause instanceof GloballyGeneralException) {
+                return HttpResponse.of(HttpStatus.FORBIDDEN);
+            }
+            return ExceptionHandlerFunction.fallthrough();
```
minwoox: I'm wondering if we need to add a comment to explain what this means.

examples/annotated-http-service/src/main/java/example/armeria/server/annotated/ExceptionHandlerService.java
hyangtack: They are explained in our official web site(https://line.github.io/armeria/server-annotated-service.html). So I think it would be better to add links to the documents. Thank you for the suggestion!

examples/annotated-http-service/src/main/java/example/armeria/server/annotated/ExceptionHandlerService.java
hyangtack: Ok, will add it!

minwoox:
Thanks @hyangtack!

PR #1334, opened 2018-08-13
Update post release msg

minwoox:
Motivation:
Because we have armeria-examples in a different repository, we need to include the instruction which copies the examples.


hyangtack:
Thanks, @minwoox !

PR #1338, opened 2018-08-16
Push ClientRequestContext when initiating gRPC HTTP request and calli…

anuraaga:
…ng onReady.

`listener.onReady()` is for completeness since all callbacks should have context, though it's a rarely used one.

Pushing when calling `httpClient.execute` is more important - currently, decorators that expect the context to be filled will have unexpected behavior since without calling `.push`, `onChild` callbacks aren't run. I'm not sure if this is a good approach though - we pass `ctx` into `execute` so it shouldn't have to be pushed, yet I think having had `onChild` called is important.

Should `onChild` be called when a `DefaultClientRequestContext` is constructed, not when it's pushed? Also, perhaps we should have a first-class API for getting attributes recursively following parents instead of users having to copy attributes in `onChild` which can be repetitive.

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
+++ b/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
@@ -164,7 +164,7 @@ class ArmeriaClientCall<I, O> extends ClientCall<I, O>
         prepareHeaders(req.headers(), compressor);
         listener = responseListener;
         final HttpResponse res;
-        try {
+        try (SafeCloseable ignored = ctx.push()) {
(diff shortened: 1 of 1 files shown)


trustin:
@anuraaga 

> Should `onChild` be called when a `DefaultClientRequestContext` is constructed, not when it's pushed?

I'm not sure if it's a good idea to do that. We may create a `DefaultClientRequestContext` and push it some time later.

> Also, perhaps we should have a first-class API for getting attributes recursively following parents instead of users having to copy attributes in `onChild` which can be repetitive.

Sounds like a good idea. Interested in sending a pull request? :wink:

trustin:
Thanks, @anuraaga !

PR #1340, opened 2018-08-19
Generate config metadata for Spring

matsumana:
If metadata is generated, we can use code completion and code jump feature in IntelliJ.
I think that is helpful.

---

# As-Is

# To-Be

Unfortunately, the `ports` property defined with List doesn't seem to be complemented.

---

Official documentation of Spring Boot:
- [Spring 1.5.x](https://docs.spring.io/spring-boot/docs/1.5.15.RELEASE/reference/html/configuration-metadata.html#configuration-metadata-annotation-processor)
- [Spring 2.0.x](https://docs.spring.io/spring-boot/docs/2.0.4.RELEASE/reference/html/configuration-metadata.html#configuration-metadata-annotation-processor)


adriancole:
hehe I did not know this.. LG!

hyangtack:
Thanks, @matsumana !

PR #1344, opened 2018-08-23
Add StartStopSupport

trustin:
Motivation:

It would be useful to provide a general construct that helps a user
implement asynchronous startup/shutdown life cycle, just like we do for
`Server.start()` and `stop()`.

Modifications:

- Add `StartStopSupport`
- Use `StartStopSupport` in `Server`
- Miscellaneous:
  - Do not run streaming tests in `:code:shadedTest`
  - Fix a bug where `DocStringExtractor` uses a wrong `ClassLoader`

Result:

- Reusability
- Slightly reduced build time
- Fixes build failures due to the test failures in `GrpcDocStringExtractorTest`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Server.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Server.java
@@ -16,7 +16,6 @@
 
 package com.linecorp.armeria.server;
 
-import static com.linecorp.armeria.common.util.Functions.voidFunction;
 import static java.util.Objects.requireNonNull;
 
 import java.net.InetSocketAddress;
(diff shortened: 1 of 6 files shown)


trustin:
Getting test failures in `com.linecorp.armeria.server.grpc.GrpcDocStringExtractorTest` which seem completely unrelated with this PR. No idea why. Can't reproduce locally.

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -486,144 +263,203 @@ public String toString() {
...
-            return "(" + type + ", " + future + ')';
-        }
-    }
-
-    private final class StateManager {
+        protected CompletionStage<Void> doStart() throws Exception {
+            final CompletableFuture<Void> future = new CompletableFuture<>();
+            // Initialize the server sockets asynchronously.
+            final List<ServerPort> ports = config().ports();
+            final AtomicInteger remainingPorts = new AtomicInteger(ports.size());
```
minwoox: nit: we can move these two lines below.

trustin: Fixed

trustin:
Thanks, reviewers!

PR #1345, opened 2018-08-27
Apply collapse to side menu

namidairo777:
This PR is to add the collapse feature to side menu which will be useful for developers when having a large list of `services` `enums` `structs` `exceptions`.
There are mainly three changes:
- Collapse feature for `services` `enums` `structs` `exceptions` sidemenu.
- Add `dense` to each `ListItem` which changed `padding-top` and `padding-down` from `12px` to `8px`

### Before (Apply `dense`)
sidemenu-dense
### After (Apply `dense`)
sidemenu-dense-new

### Collapse Feature
- All ListItems expand
sidemenu-new
- Click first Item
sidemenu-new1

### DIFF
--- a/docs-client/src/containers/App/index.tsx
+++ b/docs-client/src/containers/App/index.tsx
@@ -91,10 +91,10 @@ const styles = (theme: Theme) =>
 interface State {
   mobileDrawerOpen: boolean;
   specification?: Specification;
-  servicesState: boolean;
-  enumsState: boolean;
-  structsState: boolean;
-  exceptionsState: boolean;
+  servicesOpen: boolean;
+  enumsOpen: boolean;
+  structsOpen: boolean;
+  exceptionsOpen: boolean;
 }
 
 type Props = WithStyles<typeof styles> & RouteComponentProps<{}>;
@@ -121,9 +121,9 @@ function AppDrawer({
             <ListItemText disableTypography>
               <Typography variant="headline">Services</Typography>
             </ListItemText>
-            {state.servicesState ? <ExpandLess /> : <ExpandMore />}
+            {state.servicesOpen ? <ExpandLess /> : <ExpandMore />}
           </ListItem>
-          <Collapse in={state.servicesState} timeout="auto" unmountOnExit>
+          <Collapse in={state.servicesOpen} timeout="auto" unmountOnExit>
             {specification.getServices().map((service) => (
               <div key={service.name}>
                 <ListSubheader className={classes.methodHeader}>
@@ -161,9 +161,9 @@ function AppDrawer({
             <ListItemText disableTypography>
               <Typography variant="headline">Enums</Typography>
             </ListItemText>
-            {state.enumsState ? <ExpandLess /> : <ExpandMore />}
+            {state.enumsOpen ? <ExpandLess /> : <ExpandMore />}
           </ListItem>
-          <Collapse in={state.enumsState} timeout="auto" unmountOnExit>
+          <Collapse in={state.enumsOpen} timeout="auto" unmountOnExit>
             {specification.getEnums().map((enm) => (
               <ListItem
                 dense
@@ -190,9 +190,9 @@ function AppDrawer({
             <ListItemText disableTypography>
               <Typography variant="headline">Structs</Typography>
             </ListItemText>
-            {state.structsState ? <ExpandLess /> : <ExpandMore />}
+            {state.structsOpen ? <ExpandLess /> : <ExpandMore />}
           </ListItem>
-          <Collapse in={state.structsState} timeout="auto" unmountOnExit>
+          <Collapse in={state.structsOpen} timeout="auto" unmountOnExit>
             {specification.getStructs().map((struct) => (
               <ListItem
                 dense
@@ -219,9 +219,9 @@ function AppDrawer({
             <ListItemText disableTypography>
               <Typography variant="headline">Exceptions</Typography>
             </ListItemText>
-            {state.exceptionsState ? <ExpandLess /> : <ExpandMore />}
+            {state.exceptionsOpen ? <ExpandLess /> : <ExpandMore />}
           </ListItem>
-          <Collapse in={state.exceptionsState} timeout="auto" unmountOnExit>
+          <Collapse in={state.exceptionsOpen} timeout="auto" unmountOnExit>
             {specification.getExceptions().map((struct) => (
               <ListItem
                 dense
@@ -250,10 +250,10 @@ class App extends React.PureComponent<Props, State> {
   public state: State = {
     mobileDrawerOpen: false,
     specification: undefined,
-    servicesState: true,
-    enumsState: true,
-    structsState: true,
-    exceptionsState: true,
+    servicesOpen: true,
+    enumsOpen: true,
+    structsOpen: true,
+    exceptionsOpen: true,
   };
 
   public componentWillMount() {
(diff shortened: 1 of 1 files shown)


docs-client/src/containers/App/index.tsx
```diff
@@ -88,25 +91,35 @@ const styles = (theme: Theme) =>
 interface State {
   mobileDrawerOpen: boolean;
   specification?: Specification;
+  servicesState: boolean;
```
anuraaga: servicesOpen, enumsOpen, etc

docs-client/src/containers/App/index.tsx
```diff
@@ -88,25 +91,35 @@ const styles = (theme: Theme) =>
...
+  structsState: boolean;
+  exceptionsState: boolean;
 }
 
 type Props = WithStyles<typeof styles> & RouteComponentProps<{}>;
 
 interface AppDrawerProps extends WithStyles<typeof styles> {
   specification: Specification;
   navigateTo: (url: string) => void;
+  state: State;
```
anuraaga: Pass in the four props individually to prevent unrelated state changes from affecting this component

docs-client/src/containers/App/index.tsx
```diff
@@ -280,6 +317,8 @@ class App extends React.PureComponent<Props, State> {
               classes={classes}
               specification={specification}
               navigateTo={(url: string) => this.navigateTo(url)}
+              state={this.state}
+              handleCollapse={(itemName: string) => this.handleCollapse(itemName)}
```
anuraaga: Just this.handleCollapse

docs-client/src/containers/App/index.tsx
```diff
@@ -351,6 +392,31 @@ class App extends React.PureComponent<Props, State> {
       mobileDrawerOpen: !this.state.mobileDrawerOpen,
     });
   };
+
+  private handleCollapse(itemName: string) {
```
anuraaga: `handleCollapse = (itemName) => {}` so it's bound to this

docs-client/src/containers/App/index.tsx
```diff
@@ -351,6 +392,31 @@ class App extends React.PureComponent<Props, State> {
...
+    switch(itemName) {
+      case "services":
```
anuraaga: Let's make the type a union of these values instead of `string`

docs-client/src/containers/MethodPage/index.tsx
```diff
@@ -203,6 +203,9 @@ export default class MethodPage extends React.PureComponent<Props, State> {
                   onChange={this.onDebugFormChange}
                   inputProps={{
                     className: 'code',
+                    style: {
+                      fontSize: '14px',
```
anuraaga: Does it work to set this in the stylesheet instead of using inline styles?

trustin:
@namidairo777 Hello! Thanks for a nice pull request. Would you mind signing the CLA first?

docs-client/src/containers/App/index.tsx
namidairo777: I tried to pass four props individually but turned out that the DOM will not refresh by value change (E.g. `servivesOpen: true` to `servivesOpen: false`), then click some redirection menu (one item in `Services`), the menu will be collapsed. 
But by passing them inside the `state`, this works.

docs-client/src/containers/App/index.tsx
trustin: I see `navigateTo(to: string)` does not use a fat arrow while some other methods use fat arrows. Would you mind educating me when a fat arrow should be used? I thought it's OK not to use a fat arrow for prototype methods.

anuraaga: Looking again, I think `navigateTo` should use an arrow too and then we don't need an arrow in the above places where we reference it, oops. 

Unfortunately, prototype methods when used as callbacks are not bound to `this` and `this` can't be used in them, while arrow functions are. I guess we may want to follow a practice that all methods on a component class should always be arrow functions to avoid worrying about it.

trustin: I'm curious why callback methods are not bound to `this`. Can't just the party that invokes the callback do `component.callbackMethod();` then `this` could be `component`?

anuraaga: If `component` was passed around, then it works fine. But most JS APIs pass function references around, so for example you end up doing `onClick={component.callbackMethod}` (passing reference to the function, not component). As soon as you do this, for a normal prototype method it becomes unbound. One of the most frustrating aspects of Javascript (and Typescript doesn't fix it to preserve compatibility)

trustin: I see. Thanks for explanation, @anuraaga!

docs-client/src/containers/MethodPage/index.tsx
namidairo777: This works in my local env. 
I removed it eventually, because everyone has a different taste in design. 
Maybe only me who thinks the `font-size` of `debug` and `response` may look too big.

namidairo777:
@trustin @anuraaga 
I found a warning when I open browser's developer tool:
```
Warning: Each child in an array or iterator should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
    in React.Fragment
```

docs-client/src/containers/App/index.tsx
anuraaga: I checked out the branch and am trying to reproduce this, but can't find any problem. Can you describe in more detail with 1), 2), 3) etc what steps cause the four boolean approach to not work? Code-wise, I don't see any reason it wouldn't.

This is the code I'm trying and not finding problem with
https://github.com/anuraaga/armeria/blob/namidairo777-Apply_Collapse_to_side_menu/docs-client/src/containers/App/index.tsx#L116

docs-client/src/containers/App/index.tsx
```diff
@@ -297,6 +334,10 @@ class App extends React.PureComponent<Props, State> {
               classes={classes}
               specification={specification}
               navigateTo={(url: string) => this.navigateTo(url)}
+              state={this.state}
+              handleCollapse={(itemName: string) =>
```
anuraaga: You can replace this with `{this.handleCollapse}` too

docs-client/src/containers/App/index.tsx
```diff
@@ -88,127 +91,155 @@ const styles = (theme: Theme) =>
...
-                  onClick={() =>
-                    navigateTo(`/methods/${service.name}/${method.name}`)
-                  }
-                >
-                  <ListItemText
-                    inset
-                    primaryTypographyProps={{
-                      variant: 'body1',
-                    }}
+          <Collapse in={state.servicesOpen} timeout="auto" unmountOnExit>
```
anuraaga: Let's remove `unmountOnExit` throughout. I guess while it is pretty complicated DOM, it's not unreasonable so the optimization of unmounting may not be worth the fact that the sidebar width changes when collapsing if we do unmount.

namidairo777: You are right! It's better to keep the contents in DOM but just collapse.
Using `unmountOnExit `  may cause the sidebar width change which looks a little bit wired

namidairo777:
@anuraaga Thanks for your detail explanation. I misunderstood before. 
I thought you gave the advice to define the `xxxOpen` values separately from `State` and pass them to the function

namidairo777:
I will fix this later (https://github.com/line/armeria/pull/1345#discussion_r212851824) tomorrow 😄

docs-client/src/containers/App/index.tsx
anuraaga: Happened to find this detailed guide too

https://github.com/Microsoft/TypeScript/wiki/'this'-in-TypeScript

trustin: Thanks for sharing, @anuraaga. Very helpful.

trustin:
Thanks, @namidairo777 and reviewers!

PR #1349, opened 2018-08-28
Completes stopFuture immediately if Server does not start server serverChannels yet

imasahiro:
Motivations:
- When a server starts multiple times, it fails to start later one but
  start/stopFuture for later one will be stuck since there is no chance to
  complete the future.

Modifications:
- Completes a stopFuture immediately if Server does not start server serverChannels yet
- Shutdown server correctly even if fails to startup

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/server/ServerTest.java
+++ b/core/src/test/java/com/linecorp/armeria/server/ServerTest.java
@@ -23,7 +23,6 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.PrintWriter;
-import java.net.BindException;
 import java.net.Socket;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
@@ -317,8 +316,7 @@ public class ServerTest {
                 .service("/", (ctx, res) -> HttpResponse.of(""))
                 .build();
         assertThatThrownBy(() -> duplicatedPortServer.start().join())
-                .hasCauseInstanceOf(BindException.class)
-                .hasMessageContaining("Address already in use");
+                .hasCauseInstanceOf(IOException.class);
     }
 
     private static void testSimple(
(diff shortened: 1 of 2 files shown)


imasahiro:
cc: @fukuang

core/src/test/java/com/linecorp/armeria/server/ServerTest.java
```diff
@@ -308,6 +310,17 @@ public void testUnsupportedMethod() throws Exception {
...
 
+    @Test
+    public void duplicatedPort() {
+        Server duplicatedPortServer = new ServerBuilder()
+                .http(server.httpPort())
+                .service("/", (ctx, res) -> HttpResponse.of(""))
+                .build();
+        assertThatThrownBy(() -> duplicatedPortServer.start().join())
+                .hasCauseInstanceOf(BindException.class)
+                .hasMessageContaining("Address already in use");
```
trustin: I guess this may fail because I/O exception messages are often platform dependent.

trustin:
Please fix the test failure:

```
com.linecorp.armeria.server.ServerTest > duplicatedPort FAILED
    java.lang.AssertionError: 
    Expecting a throwable with cause being an instance of:
     <java.net.BindException>
    but was an instance of:
     <io.netty.channel.unix.Errors$NativeIoException: bind(..) failed: Address already in use>
        at com.linecorp.armeria.server.ServerTest.duplicatedPort(ServerTest.java:320)
```

Probably using `IOException.class` is enough?

imasahiro:
Updated. 
> Do you want us to release 0.70.1?

Maybe we can wait next round. How do you think? @fukuang

fukuang:
> Do you want us to release 0.70.1?

@trustin @imasahiro 

I appreciate if you can release 0.70.1 if it's not too troublesome.

Our service is about to be released. It would be nice if we can include this fix :)

trustin:
Let me merge this and release now. :wink:

trustin:
@fukuang @imasahiro Just released 0.70.1. Enjoy :-)

imasahiro:
Thanks @trustin !!

PR #1354, opened 2018-09-04
Handle `%26` and `%3D` in a query string correctly

trustin:
Motivation:

The following query string:

    a=1%26a=2

should be interpreted as a single query entry whose key is `a` and value
is `1&a=2`. However, `PathAndQuery` currently handles it as two query
entries: `a=1` and `a=2`.

It is because `PathAndQuery` decodes `%26` into `&` internally, making `%26`
indistinguishable from `&`.

Also, `%3D` in a query string should be similarly due to the case where
a query string key contains `=`: `a%3Db=1`

Modifications:

- Decode `%26` and `%3D` into a special marker character so that it is encoded
  back into `%26` and `%3D` in `PathAndQuery`
- Update test cases
  - ArmeriaCallFactory has been updated by @kojilin.

Result:

- A query parameter that contains an ampersand or an equal sign is interpreted correctly.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/internal/PathAndQueryTest.java
+++ b/core/src/test/java/com/linecorp/armeria/internal/PathAndQueryTest.java
@@ -189,14 +189,42 @@ public class PathAndQueryTest {
 
     @Test
     public void plus() {
-        final PathAndQuery res = PathAndQuery.parse("/+?+");
+        final PathAndQuery res = PathAndQuery.parse("/+?a+b=c+d");
         assertThat(res).isNotNull();
         assertThat(res.path()).isEqualTo("/+");
-        assertThat(res.query()).isEqualTo("+");
+        assertThat(res.query()).isEqualTo("a+b=c+d");
 
-        final PathAndQuery res2 = PathAndQuery.parse("/%2b?%2b");
+        final PathAndQuery res2 = PathAndQuery.parse("/%2b?a%2bb=c%2bd");
         assertThat(res2).isNotNull();
         assertThat(res2.path()).isEqualTo("/+");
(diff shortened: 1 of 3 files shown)


retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactoryTest.java
```diff
@@ -361,6 +361,18 @@ public void queryStringEncoded() throws Exception {
 
         response = service.queryStringEncoded("Foo+Bar", 33).get();
         assertThat(response).isEqualTo(new Pojo("Foo Bar", 33));
+
+        response = service.queryString("Foo%26name%3DBar", 33).get();
```
kojilin: move this to `#queryString_withSpecialCharacter ` ?

core/src/test/java/com/linecorp/armeria/internal/PathAndQueryTest.java
```diff
@@ -189,14 +189,56 @@ public void space() {
...
+    }
+
+    @Test
+    public void semicolon() {
+        final PathAndQuery res = PathAndQuery.parse("/;?a=b;c=d");
+        assertThat(res).isNotNull();
+        assertThat(res.path()).isEqualTo("/;");
+        assertThat(res.query()).isEqualTo("a=b;c=d");
+
+        // '%26' in a query string should never be decoded into '&'.
```
kojilin: nit comment is `%3B`

trustin: Oops :-)

trustin:
Will merge once CI passes. @kojilin, thanks again for reporting this issue.

PR #1355, opened 2018-09-05
Use thread local when creating scopes off of the request context thread.

anuraaga:
While the `RequestContext` can be treated as the only thread-local for the request thread, this isn't the case for other threads, i.e., when making callbacks run in a threadpool context-aware. For such callbacks, we need to use a separate thread-local to prevent the multiple threads from clashing the context.

### DIFF
--- a/zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
+++ b/zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
@@ -159,14 +159,14 @@ public final class RequestContextCurrentTraceContext extends CurrentTraceContext
     @Override
     @Nullable
     public TraceContext get() {
-        RequestContext ctx = getRequestContextOrWarnOnce();
+        final RequestContext ctx = getRequestContextOrWarnOnce();
         if (ctx == null) {
             return null;
         }
         if (ctx.eventLoop().inEventLoop()) {
             return ctx.attr(TRACE_CONTEXT_KEY).get();
         } else {
-            TraceContext threadLocalContext = THREAD_LOCAL_CONTEXT.get();
+            final TraceContext threadLocalContext = THREAD_LOCAL_CONTEXT.get();
             if (threadLocalContext != null) {
                 return threadLocalContext;
             }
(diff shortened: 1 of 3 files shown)


zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -148,11 +152,27 @@ public static void copy(RequestContext src, RequestContext dst) {
...
+    private RequestContextCurrentTraceContext(Builder builder) {
+        super(builder);
+    }
+
     @Override
     @Nullable
     public TraceContext get() {
-        final Attribute<TraceContext> traceContextAttribute = getTraceContextAttributeOrWarnOnce();
-        return traceContextAttribute != null ? traceContextAttribute.get() : null;
+        RequestContext ctx = getRequestContextOrWarnOnce();
```
hyangtack: nit: please add missing `final`.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -162,11 +182,24 @@ public Scope newScope(@Nullable TraceContext currentSpan) {
             return Scope.NOOP;
         }
 
-        final Attribute<TraceContext> traceContextAttribute = getTraceContextAttributeOrWarnOnce();
-        if (traceContextAttribute == null) {
+        RequestContext ctx = getRequestContextOrWarnOnce();
```
hyangtack: nit: ditto

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -182,27 +215,57 @@ public void close() {
...
+    @Nullable
+    private static RequestContext getRequestContextOrWarnOnce() {
+        return RequestContext.mapCurrent(Function.identity(), LogRequestContextWarningOnce.INSTANCE);
     }
 
-    private RequestContextCurrentTraceContext(Builder builder) {
-        super(builder);
+    /** Armeria code should always have a request context available, and this won't work without it. */
+    @Nullable private static Attribute<TraceContext> getTraceContextAttributeOrWarnOnce() {
+        RequestContext ctx = RequestContext.mapCurrent(
```
hyangtack: nit: ditto

anuraaga:
Fixed

trustin:
Thanks, @anuraaga and @kojilin !

PR #1356, opened 2018-09-05
Add decodedPath() and decodedMappedPath() / Decode path params

### LINKED ISSUE #756: URL encoded path mapping
As described in [RFC3986](https://tools.ietf.org/html/rfc3986), a path component might be percent-encoded. If a regex path mapping has any reserved characters like `*` in its pattern and a request path is percent-encoded, the request might not be mapped to the path mapping.

trustin:
Motivation:

- It is fairly often for a user to decode the request path.
  `RequestContext.path()`, `mappedPath()` and other path getters in
  Armeria always return a raw path, which may require an extra work on
  user-side.
- Path parameters are currently provided in an encoded form. In this
  case, a user would almost always expect a decoded form.

Modifications:

- Add `RequestContext.decodedPath()`, `ServiceRequestContext.decodedMappedPath()`,
  `RequestLog.decodedPath()` and `PathMappingResult.decodedPath()` which
  return the decoded path string.
- Update all `PathMapping` implementations so that path parameters are
  always decoded.
- Add `PathMappingResultBuilder` to build a non-trivial mapping result
  more conveniently.
- Add `PathMappingResult.of(String path)` for convenience.
- Add `ArmeriaHttpUtil.decodePath()`
- Fix a bug where `HttpFileService` does not handle UTF8-encoded paths
  correctly.
- Miscellaneous:
  - Fix missing `@Nonnull` annotations for some `RequestContext`s.
  - Remove `PathAndQuery.isHexadecimal()` which was unnecessary.
  - Fix inefficient string conversion in `StringResponseConverterFunction`

Result:

- Fixes #756
- More convenient to retrieve a decoded request path
- `HttpFileService` handles CJK paths correctly.
- Breaking change: path parameters are not encoded anymore.


core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
```diff
@@ -169,6 +169,16 @@ default String path() {
         return context().path();
     }
 
+    /**
+     * Returns the absolute path part of the current {@link Request} URI, excluding the query part,
+     * decoded in UTF-8.
+     * This method is a shortcut to {@code context().path()}.
```
minwoox: path() -> decodedPath() ?

core/src/main/java/com/linecorp/armeria/server/PathMappingResult.java
```diff
@@ -44,6 +46,13 @@ public static PathMappingResult empty() {
         return EMPTY;
     }
 
+    /**
+     * Creates a new instance with the specified {@code path}, without a query any path parameters.
```
minwoox: any -> and ?

trustin: It was actually 'or any'. Thanks :-)

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
trustin: Fixed. Thanks!

trustin:
Thanks for reviewing :-)

PR #1357, opened 2018-09-07
Add AsyncMethodCallbacks which bridges the gap between CompletionStage and AsyncMethodCallback

### LINKED ISSUE #508: Provide a utility that bridges the gap between CompletionStage and AsyncMethodCallback
So that a user can delegate the result of `CompletionStage` to `AsyncMethodCallback`. e.g.

```java

public void hello(String name, AsyncMethodCallback<String> resultHandler) throws TException {
    CompletableFuture<String> future = asyncHello(name);
    BiFunction<String, Throwable, Void> converted = convert(resultHandler); // Name?
    future.handle(converted);
}
```

trustin:
Motivation:

The following is a very common idiom when implementing an asynchronous
Thrift service:

    public void myThriftServiceMethod(AsyncMethodCallback<String> cb) {
        CompletableFuture<String> f = someAsyncOp();
        f.whenComplete((res, cause) -> {
            if (cause != null) {
                if (cause instanceof Exception) {
                    cb.onError(cause);
                } else {
                    cb.onError(new CompletionException(cause));
                }
            } else {
                cb.onComplete(res);
            }
        });
    }

We could provide some utility for this:

    public void myThriftServiceMethod(AsyncMethodCallback<String> cb) {
        AsyncMethodCallbacks.transfer(someAsyncOp(), cb);
    }

Modifications:

- Add `AsyncMethodCallbacks.transfer()` and `transferAsync()`
- Add `AsyncMethodCallbacks.invokeOnError()` for deduplication with
  `THttpClientInvocationHandler`.

Result:

- Fixes #508
- Slightly better Thrift experience

### DIFF
--- a/thrift/src/test/java/com/linecorp/armeria/common/thrift/AsyncMethodCallbacksTest.java
+++ b/thrift/src/test/java/com/linecorp/armeria/common/thrift/AsyncMethodCallbacksTest.java
@@ -17,7 +17,6 @@ package com.linecorp.armeria.common.thrift;
 
 import static com.spotify.futures.CompletableFutures.exceptionallyCompletedFuture;
 import static java.util.concurrent.CompletableFuture.completedFuture;
-import static org.awaitility.Awaitility.await;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.ArgumentMatchers.isA;
@@ -27,8 +26,6 @@ import static org.mockito.Mockito.only;
 import static org.mockito.Mockito.verify;
 
 import java.util.concurrent.CompletionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 
 import org.apache.thrift.async.AsyncMethodCallback;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


thrift/src/test/java/com/linecorp/armeria/common/thrift/AsyncMethodCallbacksTest.java
```diff
@@ -0,0 +1,110 @@
...
+                   "foo".equals(argument.getCause().getMessage());
+        }));
+    }
+
+    @Test
+    public void transferCallbackError() {
+        @SuppressWarnings("unchecked")
+        final AsyncMethodCallback<String> callback = mock(AsyncMethodCallback.class);
+        doThrow(new AnticipatedException()).when(callback).onComplete(any());
+        AsyncMethodCallbacks.transfer(completedFuture("foo"), callback);
```
imasahiro: Is it better to add assertions here?
`verify(callback, only()).onComplete("foo");`

trustin: Sounds good. Added a verification as suggested.

kojilin:
LGTM, just curious whats the purpose of `transferAsync` ? I think it usually safe to call in any thread?

trustin:
> LGTM, just curious whats the purpose of `transferAsync` ? I think it usually safe to call in any thread?

That's a good question. I added them just in case an implementor of `AsyncMethodCallback` want to make sure the callbacks are always invoked from the same thread. I'm not sure it will be useful or not.

I guess it's safe to remove them given that `AsyncMethodCallback` is supposed to be thread-safe and thus there's no point of exposing them to users.

trustin:
Removed `transferAsync()` as suggested. Let me know if there's any concern about the removal.

trustin:
@imasahiro Any other comments?

trustin:
Thanks for reviewing :-)

imasahiro:
LGTM~

PR #1358, opened 2018-09-07
Fix an assertion error caused by simultaneous `StartStopSupport.start()` or `stop()`

### LINKED ISSUE #1351: Assertion error in StartStopSupport
```
java.lang.AssertionError: null
	at com.linecorp.armeria.common.util.StartStopSupport.lambda$stop$10(StartStopSupport.java:196)
	at io.netty.util.concurrent.GlobalEventExecutor$TaskRunner.run(GlobalEventExecutor.java:248)
	at java.base/java.lang.Thread.run(Thread.java:844)
```

trustin:
Motivation:

An `AssertionError` can be triggered when a user calls `StartStopSupport.start()`
or `stop()` because we do not switch the state into `STARTING` or `STOPPING`
in a single transaction.

Modifications:

- Enter `STARTING` or `STOPPING` before submitting a start/stop task.
- Roll back to `STOPPED` or `STARTED` if the task submission failed.

Result:

- Fixes #1351

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/StartStopSupport.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/StartStopSupport.java
@@ -118,13 +118,16 @@ public abstract class StartStopSupport<V, L> implements AutoCloseable {
                              .thenComposeAsync(unused -> start(failIfStarted), executor);
         }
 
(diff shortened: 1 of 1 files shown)


trustin:
Thanks!

PR #1362, opened 2018-09-10
Allow TAB, CR and LF in a query string

trustin:
Motivation:

The validation logic in `PathAndQuery` is still too restrictive. It
should allow `%09`, `%0A` and `%0D` (TAB, LF and CR) which can be
included in a query string part, e.g. a user can enter a multi-line
text in a form.

Modifications:

- Allow TAB, CR and LF in a query string.
  - They are still unallowed in a path part.

Result:

- Innocent web form submissions are not rejected.


trustin:
Thanks for swift reviews. Will cut a release tomorrow morning. :sob:

anuraaga:
Just wondering, are we reinventing a really complicated wheel with this method?

trustin:
Maybe.. I hope this is the last hole to close and it leaves less concern of path and query sanitization to users.

PR #1367, opened 2018-09-14
Fix wrong assert condition in AremriaCallFactoryTest

minwoox:
The test should check if the content type is `null`, but it checks query.


trustin:
Thanks, @minwoox !

PR #1369, opened 2018-09-19
Fix NPE in HttpRequestSubscriber.onSubscribe()

trustin:
Motivation:

`HttpRequestSubscriber` retrieves the last `ChannelHandlerContext` in a
`ChannelPipeline` using `ChannelPipeline.lastContext()`. However,
`lastContext()` may return `null` when called on a closed `Channel`.

Modifications:

- Do not use `ChannelPipeline.lastContext()`.
- Make `HttpObjectEncoder` implementations keep the non-null reference
  to `Channel` or `ChannelHandlerContext` for 1) no change of NPE and 2)
  simpler method signatures.

Result:

- NPE is gone.
- Slightly simpler encoder code.


trustin:
Thank you, reviewers!

Issue #1368, opened 2018-09-15
Thrift oneway sync call need to wait server handle complete???


yanfeipeng:
```java
    private static void invokeSynchronously(
            ServiceRequestContext ctx, Object impl,
            ThriftFunction func, TBase<?, ?> args, DefaultRpcResponse reply) {

        final ProcessFunction<Object, TBase<?, ?>> f = func.syncFunc();
        ctx.blockingTaskExecutor().execute(() -> {
            if (reply.isDone()) {
                // Closed already most likely due to timeout.
                return;
            }

            try {
                final TBase<?, ?> result = f.getResult(impl, args);
                if (func.isOneWay()) {
                    reply.complete(null);
                } else {
                    reply.complete(func.getResult(result));
                }
            } catch (Throwable t) {
                reply.completeExceptionally(t);
            }
        });
    }
````

something wrong with 

```java
                final TBase<?, ?> result = f.getResult(impl, args);
                if (func.isOneWay()) {
                    reply.complete(null);
                } else {
                    reply.complete(func.getResult(result));
                }
```

trustin:
@yanfeipeng What would be the expected behavior then? Sending an HTTP response immediately after the service started to handle the request?

yanfeipeng:
> @yanfeipeng What would be the expected behavior then? Sending an HTTP response immediately after the service started to handle the request?

Nearly.
1.Sometimes I donot care about the respone, just sending the request successfully is OK.
2.oneway call is diff from primitive Thrift, make coder confuse

PR #1372, opened 2018-09-20
fix: fix thrift oneway sync call need to way service handle complete …

yanfeipeng:
…#1368

### DIFF
--- a/thrift/src/main/java/com/linecorp/armeria/server/thrift/ThriftCallService.java
+++ b/thrift/src/main/java/com/linecorp/armeria/server/thrift/ThriftCallService.java
@@ -166,13 +166,17 @@ public final class ThriftCallService implements Service<RpcRequest, RpcResponse>
             }
 
             try {
-                final TBase<?, ?> result = f.getResult(impl, args);
                 if (func.isOneWay()) {
                     reply.complete(null);
+                    f.getResult(impl, args);
                 } else {
+                    final TBase<?, ?> result = f.getResult(impl, args);
                     reply.complete(func.getResult(result));
                 }
             } catch (Throwable t) {
+                if (func.isOneWay()) {
+                    throw new RuntimeException(t);
+                }
                 reply.completeExceptionally(t);
             }
         });


thrift/src/main/java/com/linecorp/armeria/server/thrift/ThriftCallService.java
```diff
@@ -166,13 +166,17 @@ private static void invokeSynchronously(
...
                 if (func.isOneWay()) {
                     reply.complete(null);
+                    f.getResult(impl, args);
                 } else {
+                    final TBase<?, ?> result = f.getResult(impl, args);
                     reply.complete(func.getResult(result));
                 }
             } catch (Throwable t) {
+                if (func.isOneWay()) {
+                    throw new RuntimeException(t);
```
trustin: The exception will go out of our hand if we rethrow here. How about logging instead?

```java
if (func.isOneWay()) {
    logger.warn("{} Unexpected exception from a one-way function:", ctx, t);
} else {
    reply.completeExceptionally(t);
}
```

yanfeipeng: logging is more perfect. My first fix just want to log, But need to import logger, so, give up

trustin: Thanks!

I believe `invokeAsynchronously()` should be changed like this:

```java
    private static final AsyncMethodCallback<Object> NOOP_CALLBACK = new AsyncMethodCallback<Object> {
        @Override
        public void onComplete(Object response) {}

        @Override
        public void onError(Exception e) {}
    };

    private static void invokeAsynchronously(
            Object impl, ThriftFunction func, TBase<?, ?> args, DefaultRpcResponse reply) throws TException {

        final AsyncProcessFunction<Object, TBase<?, ?>, Object> f = func.asyncFunc();
        if (func.isOneWay()) {
            f.start(impl, args, NOOP_CALLBACK);
            reply.complete(null);
        } else {
            f.start(impl, args, new AsyncMethodCallback<Object>() {
                @Override
                public void onComplete(Object response) {
                    reply.complete(response);
                }

                @Override
                public void onError(Exception e) {
                    reply.completeExceptionally(e);
                }
            });
        }
    }
```

trustin:
Could you also sign the CLA?

thrift/src/main/java/com/linecorp/armeria/server/thrift/ThriftCallService.java
yanfeipeng: My opinion:
It is a bad way to use oneway call in async. Because async call donot wait the invoke result.Something like sync oneway call not interested with the result.And in Thrift, oneway Method Must be Void, the return has no scense.
But how to measure the oneway call complete?Just send out or up to the server side?
The sync oneway call change is the 'up to server side' way, for Unifying Concept， The invokeAsynchronously() change is the good way!!!
How to choose a more reasonable?@trustin

trustin: Because it's a server-side change, the behavior has to be identical for sync and async calls, i.e. my proposed change in `invokeAsynchronously()` is required. Otherwise, a synchronous client that calls an asynchronous one-way function will block until the server finishes handling the request.

yanfeipeng: yeah I see,  Agree! I update my fix with that 'proposed change' code!!!

yanfeipeng:
> Could you also sign the CLA?

how to ?

trustin:
@yanfeipeng It's here: https://cla-assistant.io/line/armeria?pullRequest=1372

trustin:
Gentle ping

trustin:
@yanfeipeng Ping?

trustin:
Final call. Will close if no response within a week.

trustin:
Superceded by https://github.com/line/armeria/pull/1455

PR #1375, opened 2018-10-05
Report health for all endpoints, not just ones with unique authority.

anuraaga:
For an endpoint group like a DNS one, multiple endpoints have the same authority but different IP address. The current logic will only report health for one endpoint in such a case. Really, we want to make sure we report health for all endpoints, regardless of content, so this changes the logic to key on `Endpoint` instead of authority.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
@@ -17,6 +17,7 @@ package com.linecorp.armeria.client.endpoint.healthcheck;
 
 import static java.util.Objects.requireNonNull;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -28,6 +29,7 @@ import com.linecorp.armeria.client.Endpoint;
 import com.linecorp.armeria.common.metric.MeterIdPrefix;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.MeterBinder;
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
```diff
@@ -69,23 +71,31 @@ public void bindTo(MeterRegistry registry) {
...
-                registry.gauge(idPrefix.name(), idPrefix.tags("authority", authority),
-                               this, unused -> healthMap.get(authority) ? 1 : 0);
+            endpointsToUpdate.forEach((endpoint, healthy) -> {
+                healthMap.put(endpoint, healthy);
+                List<Tag> tags = new ArrayList<>(2);
+                tags.add(Tag.of("authority", endpoint.authority()));
+                if (endpoint.hasIpAddr()) {
+                    String address = endpoint.ipAddr();
+                    assert address != null;
+                    tags.add(Tag.of("address", address));
```
trustin: I wonder if `ip` or `ipAddr` is better. Thoughts?

anuraaga: `ip` seems good - I avoided `ip_address` since it felt verbose.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
```diff
@@ -69,23 +71,31 @@ public void bindTo(MeterRegistry registry) {
...
             });
 
             // Process the newly appeared endpoints.
-            endpointsToUpdate.forEach((authority, healthy) -> {
-                healthMap.put(authority, healthy);
-                registry.gauge(idPrefix.name(), idPrefix.tags("authority", authority),
-                               this, unused -> healthMap.get(authority) ? 1 : 0);
+            endpointsToUpdate.forEach((endpoint, healthy) -> {
+                healthMap.put(endpoint, healthy);
+                List<Tag> tags = new ArrayList<>(2);
```
hyangtack: nit: `final`?

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
```diff
@@ -69,23 +71,31 @@ public void bindTo(MeterRegistry registry) {
...
+                tags.add(Tag.of("authority", endpoint.authority()));
+                if (endpoint.hasIpAddr()) {
+                    String address = endpoint.ipAddr();
```
hyangtack: nit: `final`?

PR #1377, opened 2018-10-05
Use MIME type base64 decoder for SAML messages

hyangtack:
Motivation:
The `Basic` type base64 decoder rejects data that contains characters outside the base64 alphabet.
However, some identity provider produces base64 with line feed character, so we need to use
`MIME` type base64 decoder instead of the `Basic` type.

Modifications:
- Get base64 decoder by calling Base64.getMimeDecoder() instread of Base64.getDecoder().
- Improve KeyStoreCredentialResolverBuilder:
  - Able to read a keystore file from file system.
  - Able to specify keystore type.

### DIFF
--- a/examples/annotated-http-service/src/main/java/example/armeria/server/annotated/MessageConverterService.java
+++ b/examples/annotated-http-service/src/main/java/example/armeria/server/annotated/MessageConverterService.java
@@ -112,13 +112,13 @@ public class MessageConverterService {
      *
      * <p>If you want to specify converters for every methods in a service class, you can specify them
      * above the class definition as follows:
-     * <pre><code>
-     * {@literal @}RequestConverter(CustomRequestConverter.class)
-     * {@literal @}ResponseConverter(CustomResponseConverter.class)
-     * public class MyAnnotatedService {
-     *     ...
-     * }
-     * </code></pre>
+     * <pre>{@code
+     * > @RequestConverter(CustomRequestConverter.class)
+     * > @ResponseConverter(CustomResponseConverter.class)
+     * > public class MyAnnotatedService {
+     * >     ...
+     * > }
+     * }</pre>
      */
     @Post("/custom")
     @RequestConverter(CustomRequestConverter.class)
--- a/examples/annotated-http-service/src/main/java/example/armeria/server/annotated/MessageConverterService.java
+++ b/examples/annotated-http-service/src/main/java/example/armeria/server/annotated/MessageConverterService.java
@@ -112,13 +112,13 @@ public class MessageConverterService {
      *
      * <p>If you want to specify converters for every methods in a service class, you can specify them
      * above the class definition as follows:
-     * <pre>{@code
-     * @RequestConverter(CustomRequestConverter.class)
-     * @ResponseConverter(CustomResponseConverter.class)
+     * <pre><code>
+     * {@literal @}RequestConverter(CustomRequestConverter.class)
+     * {@literal @}ResponseConverter(CustomResponseConverter.class)
      * public class MyAnnotatedService {
      *     ...
      * }
-     * }</pre>
+     * </code></pre>
      */
     @Post("/custom")
     @RequestConverter(CustomRequestConverter.class)
--- a/saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
+++ b/saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
@@ -44,7 +44,8 @@ public final class KeyStoreCredentialResolverBuilder {
     private final ClassLoader classLoader;
 
     private String type = KeyStore.getDefaultType();
-    private String password = "";
+    @Nullable
+    private String password;
     private final Map<String, String> keyPasswords = new HashMap<>();
 
     /**
@@ -76,8 +77,8 @@ public final class KeyStoreCredentialResolverBuilder {
     /**
      * Sets a password of the {@link KeyStore}.
      */
-    public KeyStoreCredentialResolverBuilder password(String password) {
-        this.password = requireNonNull(password, "password");
+    public KeyStoreCredentialResolverBuilder password(@Nullable String password) {
+        this.password = password;
         return this;
     }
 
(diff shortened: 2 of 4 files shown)


examples/annotated-http-service/src/main/java/example/armeria/server/annotated/MessageConverterService.java
```diff
@@ -112,13 +112,13 @@ public Response json3(@RequestObject Request request) {
      *
      * <p>If you want to specify converters for every methods in a service class, you can specify them
      * above the class definition as follows:
-     * <pre>{@code
-     * @RequestConverter(CustomRequestConverter.class)
-     * @ResponseConverter(CustomResponseConverter.class)
+     * <pre><code>
+     * {@literal @}RequestConverter(CustomRequestConverter.class)
+     * {@literal @}ResponseConverter(CustomResponseConverter.class)
```
trustin: You can use `>` trick. :-)

```
<pre>{@code
> @RequestConverter(...)
...
}</pre>
```

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
```diff
@@ -56,15 +67,32 @@ public KeyStoreCredentialResolverBuilder addKeyPassword(String name, String pass
...
+        final KeyStore ks = KeyStore.getInstance(type);
+        try (InputStream is = open(path)) {
             ks.load(is, keystorePassword.toCharArray());
         }
         return new KeyStoreCredentialResolver(ks, keyPasswords);
     }
+
+    private static InputStream open(String path) throws IOException {
+        final File file = new File(path);
+        if (file.isFile() && file.exists()) {
```
trustin: Instead of auto-detection, how about providing two constructors? One that accepts `File` or `Path` and the other that accepts `String path` and `ClassLoader`

examples/annotated-http-service/src/main/java/example/armeria/server/annotated/MessageConverterService.java
minwoox: please use `>`

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
```diff
@@ -18,31 +18,68 @@
...
+     * Creates a builder with the specified {@code resourcePath} and {@link ClassLoader}.
+     */
+    public KeyStoreCredentialResolverBuilder(String resourcePath, ClassLoader classLoader) {
+        this.resourcePath = requireNonNull(resourcePath, "path");
+        this.classLoader = requireNonNull(classLoader, "classLoader");
+        file = null;
+    }
+
+    /**
+     * Sets a type of the {@link KeyStore}.
```
trustin: How about adding `If not set, the default value retrieved from {@link KeyStore#getDefaultType()} will be used.`

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
```diff
@@ -56,15 +93,33 @@ public KeyStoreCredentialResolverBuilder addKeyPassword(String name, String pass
...
         }
         return new KeyStoreCredentialResolver(ks, keyPasswords);
     }
+
+    private InputStream open() throws IOException {
+        if (file != null) {
+            return new FileInputStream(file);
+        }
+
+        assert classLoader != null : "classLoader";
```
anuraaga: Looking at these extra null checks, I don't know if it's a better pattern but came up with this. Any thoughts?

```java

private final Supplier<InputStream> keystoreStream;

public KeyStoreCredentialResolverBuilder(File file) {
  requireNonNull(file, "file);
  keystoreStream = () -> new FileInputStream(file);
}

public KeyStoreCredentialResolverBuilder(String, ClassLoader) {
  keystoreStream = () -> classLoader.getResourceAsStream(resourcePath);
}
```

Actually, after writing it though, I guess the annoying `IOException` makes it non-trivial 😿

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
```diff
@@ -18,31 +18,68 @@
...
+        resourcePath = null;
+        classLoader = null;
+    }
+
+    /**
+     * Creates a builder with the specified {@code resourcePath} and {@link ClassLoader}.
+     */
+    public KeyStoreCredentialResolverBuilder(String resourcePath, ClassLoader classLoader) {
+        this.resourcePath = requireNonNull(resourcePath, "path");
+        this.classLoader = requireNonNull(classLoader, "classLoader");
```
trustin: `classLoader` could be nullable? If `null`, could try the context class loader:

```java
ClassLoader classLoader = this.classLoader;
if (classLoader == null) {
    classLoader = Thread.current().getContextClassLoader();
}
if (ClassLoader == null) {
    classLoader = getClass().getClassLoader();
}
```

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
trustin: That's a good question. It may be useful to provide an abstraction for this:

```java
public class FileOrResource { ... }

public KeyStoreCredentialResolverBuilder(File file) {
  requireNonNull(file, "file);
  keystoreResource = new FileOrResource(file);
}

public KeyStoreCredentialResolverBuilder(String, ClassLoader) {
  keystoreResource = new FileOrResource(resourcePath, classLoader);
}
...

keysotreResource.newInputStream();
```

hyangtack: @anuraaga @trustin Both sound good to me. But it seems that it's time to release. So, let me make a fix for the next version. Or, @anuraaga , you can take it if you are interested. :-)

trustin:
Thanks, @hyangtack !

PR #1382, opened 2018-10-10
Add a flag `com.linecorp.armeria.annotatedServiceExceptionVerbosity`

hyangtack:
Motivation:
A user may want to know why his or her annotated service returns 400 or 500 status code from the log files.

Modifications:
- Add a flag `com.linecorp.armeria.annotatedServiceExceptionVerbosity`
- Add `ExceptionVerbosity` enum to control logging

Result:
A user can get every exception when adding `com.linecorp.armeria.annotatedServiceExceptionVerbosity=all` to the system properties.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -43,6 +43,7 @@ import com.linecorp.armeria.client.retry.RetryingRpcClient;
 import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.server.PathMappingContext;
 import com.linecorp.armeria.server.ServiceConfig;
+import com.linecorp.armeria.server.annotation.ExceptionHandler;
 import com.linecorp.armeria.server.annotation.ExceptionLoggingMode;
 
 import io.netty.channel.epoll.Epoll;
@@ -531,13 +532,20 @@ public final class Flags {
     }
 
     /**
-     * Returns the value of the {@code annotatedServiceExceptionLoggingMode} parameter which equals to
-     * one of the {@code all}, {@code unhandled} or {@code none}.
+     * Returns the value of the {@code annotatedServiceExceptionLoggingMode} parameter. It would be used to
+     * log an exception which is raised from annotated HTTP services. If it is set to
+     * {@link ExceptionLoggingMode#ALL}, all exceptions raised from annotated HTTP services are logged as
+     * {@code warn} level. If it is set to {@link ExceptionLoggingMode#UNHANDLED}, exceptions, which are
+     * not handled by {@link ExceptionHandler}s provided by a user and not well-known exceptions,
+     * would be logged as {@code warn} level. If it is set to {@link ExceptionLoggingMode#NONE},
+     * no log would be written.
      *
      * <p>The default value of this flag is {@value DEFAULT_ANNOTATED_SERVICE_EXCEPTION_LOGGING_MODE}.
      * Specify the
-     * {@code -Dcom.linecorp.armeria.annotatedServiceExceptionLoggingMode=<spec>} JVM option to override
+     * {@code -Dcom.linecorp.armeria.annotatedServiceExceptionLoggingMode=<mode>} JVM option to override
      * the default value.
+     *
+     * @see ExceptionLoggingMode
      */
     public static ExceptionLoggingMode annotatedServiceExceptionLoggingMode() {
         return ANNOTATED_SERVICE_EXCEPTION_LOGGING_MODE;
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -532,17 +532,20 @@ public final class Flags {
     }
 
     /**
-     * Returns the verbosity of exceptions logged by annotated HTTP services. If it is set to
-     * {@link ExceptionVerbosity#ALL}, all exceptions raised from annotated HTTP services are logged as
-     * {@code warn} level. If it is set to {@link ExceptionVerbosity#UNHANDLED}, exceptions, which are
-     * not handled by {@link ExceptionHandler}s provided by a user and not well-known exceptions,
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -523,6 +530,19 @@ public static int defaultMaxTotalAttempts() {
         return COMPOSITE_SERVICE_CACHE_SPEC;
     }
 
+    /**
+     * Returns the value of the {@code annotatedServiceExceptionLoggingMode} parameter which equals to
+     * one of the {@code all}, {@code unhandled} or {@code none}.
+     *
+     * <p>The default value of this flag is {@value DEFAULT_ANNOTATED_SERVICE_EXCEPTION_LOGGING_MODE}.
+     * Specify the
+     * {@code -Dcom.linecorp.armeria.annotatedServiceExceptionLoggingMode=<spec>} JVM option to override
```
trustin: `<spec>` -> `<mode>`?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -523,6 +530,19 @@ public static int defaultMaxTotalAttempts() {
         return COMPOSITE_SERVICE_CACHE_SPEC;
     }
 
+    /**
+     * Returns the value of the {@code annotatedServiceExceptionLoggingMode} parameter which equals to
```
trustin: Could you rephrase this to better explain what this flag does and why it is necessary?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpService.java
```diff
@@ -76,18 +77,21 @@
     private final ResponseConverterFunction providedResponseConverter;
 
     private final ResponseType responseType;
+    private final ExceptionLoggingMode exceptionLoggingMode;
```
trustin: Perhaps no need to keep it as a field because it will never change?

core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpService.java
```diff
@@ -250,6 +254,12 @@ private HttpResponse convertResponse(ServiceRequestContext ctx, @Nullable Object
      */
     private HttpResponse convertException(RequestContext ctx, HttpRequest req, Throwable cause) {
         final Throwable peeledCause = Exceptions.peel(cause);
+
+        if (exceptionLoggingMode == ExceptionLoggingMode.ALL &&
+            logger.isWarnEnabled()) {
+            logger.warn("An exception from {}:", method.getName(), peeledCause);
```
trustin: How about including the class name as well as the method name?

core/src/main/java/com/linecorp/armeria/server/annotation/DefaultExceptionHandler.java
```diff
@@ -19,23 +19,51 @@
...
+        }
+
+        if (cause instanceof HttpResponseException) {
+            return ((HttpResponseException) cause).httpResponse();
+        }
+
+        if (exceptionLoggingMode == ExceptionLoggingMode.UNHANDLED &&
+            logger.isWarnEnabled()) {
+            logger.warn("No exception handler exists for the cause. " +
+                        DefaultExceptionHandler.class.getName() + " is handling it.",
```
trustin: Could use `{}` instead of concatenation?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -523,6 +531,25 @@ public static int defaultMaxTotalAttempts() {
         return COMPOSITE_SERVICE_CACHE_SPEC;
     }
 
+    /**
+     * Returns the verbosity of exceptions logged by annotated HTTP services. If it is set to
+     * {@link ExceptionVerbosity#ALL}, all exceptions raised from annotated HTTP services are logged as
+     * {@code warn} level. If it is set to {@link ExceptionVerbosity#UNHANDLED}, exceptions, which are
```
trustin: `at {@code WARN} level`

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -523,6 +531,25 @@ public static int defaultMaxTotalAttempts() {
         return COMPOSITE_SERVICE_CACHE_SPEC;
     }
 
+    /**
+     * Returns the verbosity of exceptions logged by annotated HTTP services. If it is set to
+     * {@link ExceptionVerbosity#ALL}, all exceptions raised from annotated HTTP services are logged as
```
trustin: Could be simplified using `<ul><li>`:

```
The value of this property is one of the following:
<ul>
  <li>{@link ExceptionVerbosity#ALL} - ...</li>
</ul>
```

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -523,6 +531,25 @@ public static int defaultMaxTotalAttempts() {
...
+     * {@code warn} level. If it is set to {@link ExceptionVerbosity#UNHANDLED}, exceptions, which are
+     * not handled by {@link ExceptionHandler}s provided by a user and not well-known exceptions,
+     * would be logged as {@code warn} level. If it is set to {@link ExceptionVerbosity#NONE},
```
trustin: `at {@code WARN} level`

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -190,6 +193,11 @@
             CSV_SPLITTER.splitToList(getNormalized(
                     "cachedHeaders", DEFAULT_CACHED_HEADERS, CharMatcher.ascii()::matchesAllOf));
 
+    private static final String DEFAULT_ANNOTATED_SERVICE_EXCEPTION_VERBOSITY = "unhandled";
```
trustin: `ExceptionVerbosity.UNHADLED.name()`?

hyangtack: It makes `{@value ...}` not work. So I intentionally used the string here. :-)

PR #1387, opened 2018-10-15
`EmptyFixedStreamMessage.completionFuture()` should be succeeded afte…

hyangtack:
…r sending an HttpRequest

Motivation:
When an HttpClient sends an HttpRequest without its body, the completion future of the request cannot be succeeded because the HttpRequestSubscriber cancels the subscription after sending HTTP headers. In order to fix that, check whether any of `onComplete` or `onError` has been invoked when the request has been sent. If it has not been invoked, invoke `subscription.request(1)` in order to let the publisher have a chance to produce the last call.

Modifications:
- Invoke `subscription.request(1)` one more time after the request has been sent.
- Correct misbehavior of the `write timeout` timer.
- Add HttpRequestSubscriberTest.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -62,20 +62,20 @@ public interface ClientRequestContext extends RequestContext {
     String fragment();
 
     /**
-     * Returns the amount of time allowed until starting to send out the current {@link Request}.
-     * This value is initially set from {@link ClientOption#DEFAULT_WRITE_TIMEOUT_MILLIS}.
+     * Returns the amount of time allowed until the initial write attempt of the current {@link Request}
+     * succeeds. This value is initially set from {@link ClientOption#DEFAULT_WRITE_TIMEOUT_MILLIS}.
      */
     long writeTimeoutMillis();
 
     /**
-     * Sets the amount of time allowed until starting to send out the current {@link Request}.
-     * This value is initially set from {@link ClientOption#DEFAULT_WRITE_TIMEOUT_MILLIS}.
+     * Returns the amount of time allowed until the initial write attempt of the current {@link Request}
+     * succeeds. This value is initially set from {@link ClientOption#DEFAULT_WRITE_TIMEOUT_MILLIS}.
      */
     void setWriteTimeoutMillis(long writeTimeoutMillis);
 
     /**
-     * Sets the amount of time allowed until starting to send out the current {@link Request}.
-     * This value is initially set from {@link ClientOption#DEFAULT_WRITE_TIMEOUT_MILLIS}.
+     * Returns the amount of time allowed until the initial write attempt of the current {@link Request}
+     * succeeds. This value is initially set from {@link ClientOption#DEFAULT_WRITE_TIMEOUT_MILLIS}.
      */
     void setWriteTimeout(Duration writeTimeout);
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -57,7 +56,8 @@
     enum State {
         NEEDS_TO_WRITE_FIRST_HEADER,
         NEEDS_DATA_OR_TRAILING_HEADERS,
-        DONE
+        DONE_1,
+        DONE_2
```
minwoox: I think we can remove `DONE_2`. This is needed to prevent calling `subscription.cancel()` twice, but according to the [Reactive Streams JVM spec 3.5](https://github.com/reactive-streams/reactive-streams-jvm#3.5), `subscription.cancel()` must be idempotent, so it doesn't matter.

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -73,6 +73,7 @@
     @Nullable
     private ScheduledFuture<?> timeoutFuture;
     private State state = State.NEEDS_TO_WRITE_FIRST_HEADER;
+    private boolean isCompleted;
```
minwoox: How about renaming something like `isSubscriptionCompeted`?
I think this makes me confused as it's hard to figure out the difference with `isDone()`.

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
hyangtack: On second thought after reading your comment, it would be better to remove `DONE_2` because the situation I wanted to prevent is rare. Thanks!

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
hyangtack: It makes sense. It can be confused because it's too general. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -62,19 +62,19 @@
     String fragment();
 
     /**
-     * Returns the amount of time allowed until sending out the current {@link Request} completely.
+     * Returns the amount of time allowed until starting to send out the current {@link Request}.
```
trustin: How about `the initial write attempt of the current {@link Request} succeeds`?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -62,19 +62,19 @@
...
     /**
-     * Sets the amount of time allowed until sending out the current {@link Request} completely.
+     * Sets the amount of time allowed until starting to send out the current {@link Request}.
      * This value is initially set from {@link ClientOption#DEFAULT_WRITE_TIMEOUT_MILLIS}.
      */
     void setWriteTimeoutMillis(long writeTimeoutMillis);
 
     /**
-     * Sets the amount of time allowed until sending out the current {@link Request} completely.
+     * Sets the amount of time allowed until starting to send out the current {@link Request}.
```
trustin: Ditto

hyangtack:
Thanks, reviewers!

PR #1389, opened 2018-10-16
Allow configuring HTTP2 SETTINGS frame.

### LINKED ISSUE #1241: Default flow control window in Http2Settings?
I was looking through upstream gRPC code and noticed they always set a default initial flow control window size in their `Http2Settings`

https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyServerBuilder.java#L58

Do you think armeria should set a default too? From what I can tell, the default is 64K, which seems pretty small and might affect connection warmup performance.

https://http2.github.io/http2-spec/#SETTINGS

```
SETTINGS_INITIAL_WINDOW_SIZE (0x4):
Indicates the sender's initial window size (in octets) for stream-level flow control. The initial value is 216-1 (65,535) octets.
```

It might be good to also allow configuring max streams. I wonder if max frame size should also be configurable, 16K seems a little small

```
ETTINGS_MAX_FRAME_SIZE (0x5):
Indicates the size of the largest frame payload that the sender is willing to receive, in octets.

The initial value is 214 (16,384) octets. 
```

For reference, gRPC's settings initialization.

https://github.com/grpc/grpc-java/blob/60a0b0c471d720b0546ee3a5b4fa4283635dfbcf/netty/src/main/java/io/grpc/netty/NettyServerHandler.java#L216

anuraaga:
I didn't find any real difference in my usual benchmark, but doesn't hurt to have knobs. The defaults change though

- Initial window size is 1MB instead of HTTP/2 default of 64K
- Max header size is 8K, just like our default HTTP/1 max

I just followed gRPC upstream but it'd be fine to change these defaults too.

Fixes #1241

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -154,6 +154,25 @@ public final class Flags {
     private static final boolean DEFAULT_USE_HTTP2_PREFACE = getBoolean("defaultUseHttp2Preface", true);
     private static final boolean DEFAULT_USE_HTTP1_PIPELINING = getBoolean("defaultUseHttp1Pipelining", false);
 
+    private static final int DEFAULT_DEFAULT_HTTP2_INITIAL_WINDOW_SIZE = 1048576; // 1MiB
+    private static final int DEFAULT_HTTP2_INITIAL_WINDOW_SIZE =
+            getInt("defaultHttp2InitialWindowSize",
+                   DEFAULT_DEFAULT_HTTP2_INITIAL_WINDOW_SIZE,
+                   value -> value > 0);
+
+    private static final int DEFAULT_DEFAULT_HTTP2_MAX_STREAMS_PER_CONNECTION = Integer.MAX_VALUE;
+    private static final int DEFAULT_HTTP2_MAX_STREAMS_PER_CONNECTION =
+            getInt("defaultHttp2MaxStreamsPerConnection",
+                   DEFAULT_DEFAULT_HTTP2_MAX_STREAMS_PER_CONNECTION,
+                   value -> value > 0);
+
+    // from Netty default maxHeaderSize
+    private static final int DEFAULT_DEFAULT_HTTP2_MAX_HEADER_LIST_SIZE = 8192;
+    private static final int DEFAULT_HTTP2_MAX_HEADER_LIST_SIZE =
+            getInt("defaultHttp2MaxHeaderListSize",
+                   DEFAULT_DEFAULT_HTTP2_MAX_HEADER_LIST_SIZE,
+                   value -> value > 0);
+
     private static final String DEFAULT_DEFAULT_BACKOFF_SPEC =
             "exponential=200:10000,jitter=0.2";
     private static final String DEFAULT_BACKOFF_SPEC =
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -430,6 +437,27 @@ public int defaultMaxHttp1ChunkSize() {
...
 
+    /**
+     * Returns the initial HTTP/2 flow control window size.
+     */
+    public int http2InitialWindowSize() {
+        return http2InitialWindowSize;
+    }
+
+    /**
+     * Returns the max concurrent streams per HTTP/2 connection.
```
hyangtack: `max`->`maximum number of`?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -495,6 +498,36 @@ public ServerBuilder maxHttp1ChunkSize(int maxHttp1ChunkSize) {
...
+     * the expense of being easier to overload the server. Defaults to
+     * {@link Flags#defaultHttp2InitialWindowSize()}.
+     */
+    public ServerBuilder http2InitialWindowSize(int windowSize) {
+        this.http2InitialWindowSize = validateNonNegative(windowSize, "windowSize");
+        return this;
+    }
+
+    /**
+     * Sets the max concurrent streams per HTTP/2 connection. Unset means there is no limit on
```
hyangtack: `max`->`maximum number of`?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -458,6 +477,18 @@ public static boolean defaultUseHttp1Pipelining() {
         return DEFAULT_USE_HTTP1_PIPELINING;
     }
 
+    public static int defaultHttp2InitialWindowSize() {
```
hyangtack: How about adding javadoc for the newly added methods?

anuraaga: Oops forgot to come back after defining the linked methods, added.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -495,6 +498,36 @@ public ServerBuilder maxHttp1ChunkSize(int maxHttp1ChunkSize) {
...
+     * multiplexed over each.
+     */
+    public ServerBuilder http2MaxStreamsPerConnection(int maxStreams) {
+        this.http2MaxStreamsPerConnection = validateNonNegative(maxStreams, "maxStreams");
+        return this;
+    }
+
+    /**
+     * Sets the maximum size of headers that can be received. Defaults to
+     * {@link Flags#defaultHttp2MaxHeaderListSize()} ()}.
```
minwoox: Could you remove `()}`?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -495,6 +498,36 @@ public ServerBuilder maxHttp1ChunkSize(int maxHttp1ChunkSize) {
         return this;
     }
 
+    /**
+     * Sets the initial HTTP/2 flow control window size. Larger values can stream warmup time at
```
minwoox: can lower?

trustin:
Thanks, @anuraaga !

PR #1390, opened 2018-10-17
Special-case marshalling of empty protos.

anuraaga:
I noticed gRPC upstream has this optimization so figured we should too. Unfortunately, it didn't change the result of the `empty` benchmark but probably still worth doing.

https://github.com/grpc/grpc-java/blob/master/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java#L200

After
```
Benchmark                                           (clientType)   Mode  Cnt      Score      Error  Units
c.l.a.g.downstream.DownstreamSimpleBenchmark.empty        OKHTTP  thrpt    5  29995.439 ±  263.738  ops/s
c.l.a.g.upstream.UpstreamSimpleBenchmark.empty            OKHTTP  thrpt    5  74463.286 ± 4337.546  ops/s
```

Before
```
Benchmark                                           (clientType)   Mode  Cnt      Score       Error  Units
c.l.a.g.downstream.DownstreamSimpleBenchmark.empty        OKHTTP  thrpt    5  29153.925 ±   542.103  ops/s
```


trustin:
Thanks!

PR #1391, opened 2018-10-17
Remove unnecessary firstFrame optimization.

anuraaga:
Actually, `firstFrame` was never set because I forgot to remove line 197 >< I guess the optimization previously in #1167 was just from the removal of `nextFrame`.

I tried finishing the `firstFrame` optimization but there was no difference so just getting rid of it.


trustin:
Thanks, @anuraaga

PR #1392, opened 2018-10-17
Handle GOAWAY frames correctly

trustin:
Motivation:

Armeria always disconnects the connection immediately when it receives a
GOAWAY frame, which is a violation of HTTP/2 specification.

See https://httpwg.org/specs/rfc7540.html#GOAWAY

Modifications:

- `AbstractHttp2ConnectionHandler` does not close the connections
  immediately on a GOAWAY frame anymore.
  - It still closes the connections immediately during
    `ClientFactory.close()` and `Server.stop()`.
    - Added `HttpClientFactory.isClosing()`
    - Added `GracefulShutdownSupport.isShuttingDown()`
  - It still closes the connections immediately if the error code of the
    GOAWAY frame is not NO_ERROR.
    - Updated `Http2(Request|Response)Decoder` to store the error code
      of the received GOAWAY frame.
- `Http2ResponseDecoder` does not send a redundant RST_STREAM frame for
  the requests with higher stream ID than `lastStreamId`.
- `Http2ResponseDecoder` closes `HttpResponse` with a new exception
  called `UnprocessedRequestException` to indicate that the request can
  be retried safely, because the request had a higher stream ID than
  `lastStreamId`.
  - Updated `HttpStatusBasedCircuitBreakerStrategy` to ignore an
    `UnprocessedRequestException`.
  - Added `RetryStrategy.onUnprocessed()`
- Miscellaneous:
  - Use `@linkplain` instead of `@link` where applicable.
  - Instantiate and configure `Http2(Request|Response)Decoder` inside
    `Http2(Server|Client)ConnectionHandler`.
  - Use `Http2Settings.defaultSettings()` instead of `new
    Http2Settings()` if possible.
  - Fixed an unexpected `BlockingOperationException` during
    `DefaultKeyedChannelPool.close()`.
  - Do not include stack trace information in the `GOAWAY` frames sent
    by Armeria, respecting `Flags.verboseResponses()`.
  - Made `Http2ObjectEncoder` easier to understand.
  - Made `GracefulShutdownSupport` package-local because it is not meant
    to be part of public API.
  - Replaced `HttpServerHandler.unfinishedRequests` and `unfinishedResponses`
    with a single `IdentityHashMap`.
  - Cleaned up `Http2ClientSettingsTest`
  - Skip some tests on WSL (Windows Subsystem for Linux)
  - gRPC services yield `CANCELLED` status rather than `UNKNOWN` when a
    client sends too large message now.

Result:

- GOAWAY frames are handled correctly.
- A user can choose to retry the request with higher stream ID than the
  `lastStreamId`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Http2ResponseDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Http2ResponseDecoder.java
@@ -53,6 +53,7 @@ final class Http2ResponseDecoder extends HttpResponseDecoder implements Http2Con
 
     private final Http2Connection conn;
     private final Http2ConnectionEncoder encoder;
+    private long goAwayErrorCode = -1; // -1 if not received GOAWAY yet
 
     Http2ResponseDecoder(Http2Connection conn, Channel channel, Http2ConnectionEncoder encoder) {
         super(channel);
@@ -79,7 +80,9 @@ final class Http2ResponseDecoder extends HttpResponseDecoder implements Http2Con
 
                 // Reset the stream.
                 final int streamId = idToStreamId(id);
-                if (conn.streamMayHaveExisted(streamId)) {
+                final int lastStreamId = conn.local().lastStreamKnownByPeer();
+                if (lastStreamId < 0 || // Did not receive a GOAWAY yet or
+                    streamId <= lastStreamId) { // received a GOAWAY and the request's streamId <= lastStreamId
                     final ChannelHandlerContext ctx = channel().pipeline().lastContext();
                     if (ctx != null) {
                         encoder.writeRstStream(ctx, streamId, Http2Error.CANCEL.code(), ctx.newPromise());
@@ -91,6 +94,10 @@ final class Http2ResponseDecoder extends HttpResponseDecoder implements Http2Con
             }
         }, channel().eventLoop());
         return resWrapper;
+    }
+
+    boolean receivedErrorGoAway() {
+        return goAwayErrorCode > Http2Error.NO_ERROR.code();
     }
 
     @Override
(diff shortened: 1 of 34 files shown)


core/src/main/java/com/linecorp/armeria/client/Http2ResponseDecoder.java
```diff
@@ -53,6 +53,7 @@
 
     private final Http2Connection conn;
     private final Http2ConnectionEncoder encoder;
+    private volatile long goAwayErrorCode = -1; // -1 if not received GOAWAY yet
```
anuraaga: Since this is a single connection, won't this always be read / write from the same thread?

core/src/main/java/com/linecorp/armeria/server/Http2RequestDecoder.java
```diff
@@ -58,13 +58,18 @@
     private final InboundTrafficController inboundTrafficController;
     private final IntObjectMap<DecodedHttpRequest> requests = new IntObjectHashMap<>();
     private int nextId;
+    private volatile boolean receivedErrorGoAway;
```
anuraaga: Ditto

core/src/main/java/com/linecorp/armeria/client/Http2ResponseDecoder.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/Http2RequestDecoder.java
trustin: Fixed. Thanks!

trustin:
Also added `RetryStrategy.onUnprocessed()`. Please let me know if you have a better name for it.

trustin:
Thanks for reviewing :-)

PR #1393, opened 2018-10-18
Make more HTTP/2 settings configurable and make them consistent

trustin:
- Made sure both client and server have the following HTTP/2 settings
  - Connection-level initial window size
  - Stream-level initial window size
  - Max frame size
  - Max header list size
- Reordered the fields and methods related with HTTP settings so that
  HTTP/2 settings come right before HTTP/1 settings.
- (breaking) Made the protocol-related property names consistent so that
  they all start with protocol name first, e.g.
  - `maxHttp1ChunkSize` -> `http1MaxChunkSize`
  - `initialHttp2ConnectionWindowSize` -> `http2InitialConnectionWindowSize`
- Renamed `ServerBuilder.http2InitialWindowSize` to `http2InitialStreamWindowSize`
- Added `ServerBuilder.http2InitialConnectionWindowSize` and implemented
  connection-level initial window size update in `HttpServerHandler`.
- Improve/fix validation of HTTP/2-related properties
- Change the type of `http2MaxStreamsPerConnection` and `http2MaxHeaderListSize`
  to `long` because its maximum value is `0xFFFFFFFFL`
- (breaking) Removed `default` prefix for HTTP/1-related properties because
  they should've not been named so from the beginning.
- Miscellaneous:
  - Fixed a race between `Http2RequestDecoder` and `HttpServerHandler`
    trying to write an error response, causing HTTP/2 violation.
    - Reproduced by `HttpServerTest.testTooLargeContentToNonExistentService()`
      which makes `Http2RequestDecoder` and `HttpServerHandler` send
      `413 Entity Too Large` and `404 Not Found`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
@@ -17,7 +17,6 @@
 package com.linecorp.armeria.client;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_MAX_FRAME_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_WINDOW_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.MAX_FRAME_SIZE_LOWER_BOUND;
 import static io.netty.handler.codec.http2.Http2CodecUtil.MAX_FRAME_SIZE_UPPER_BOUND;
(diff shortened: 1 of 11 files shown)


trustin:
@anuraaga :rofl: No worries! It was interesting to know sending 0xFFFFFFFF MAX_CONCURRENT_STREAM breaks OKHTTP by the way.

core/src/main/java/com/linecorp/armeria/server/HttpServerPipelineConfigurator.java
```diff
@@ -190,14 +190,19 @@ private Http2ConnectionHandler newHttp2ConnectionHandler(ChannelPipeline pipelin
...
-            settings.maxConcurrentStreams(maxStreamsPerConnection);
+
+        final int maxFrameSize = config.http2MaxFrameSize();
+        if (maxFrameSize != Http2CodecUtil.DEFAULT_MAX_FRAME_SIZE) {
+            settings.maxFrameSize(maxFrameSize);
         }
+
+        // Not using the value greater than 2^31-1 because some HTTP/2 client implementations use a signed
+        // 32-bit integer to represent an HTTP/2 SETTINGS parameter value.
+        settings.maxConcurrentStreams(Math.min(config.http2MaxStreamsPerConnection(), Integer.MAX_VALUE));
```
anuraaga: I thought I left a comment here but can't find it anymore - sorry if I double-comment.

If we clamp to int anyways shouldn't we leave the type as int instead of long?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -97,12 +96,13 @@
     private Function<? super EventLoopGroup,
             ? extends AddressResolverGroup<? extends InetSocketAddress>> addressResolverGroupFactory =
             DEFAULT_ADDRESS_RESOLVER_GROUP_FACTORY;
-    private int initialHttp2ConnectionWindowSize = DEFAULT_WINDOW_SIZE;
-    private int initialHttp2StreamWindowSize = DEFAULT_WINDOW_SIZE;
-    private int http2MaxFrameSize = DEFAULT_MAX_FRAME_SIZE;
-    private int maxHttp1InitialLineLength = Flags.defaultMaxHttp1InitialLineLength();
-    private int maxHttp1HeaderSize = Flags.defaultMaxHttp1HeaderSize();
-    private int maxHttp1ChunkSize = Flags.defaultMaxHttp1ChunkSize();
+    private int initialHttp2ConnectionWindowSize = Flags.defaultHttp2InitialConnectionWindowSize();
```
hyangtack: nit: `initialHttp2ConnectionWindowSize` -> `http2InitialConnectionWindowSize`, 
`initialHttp2StreamWindowSize` -> `http2InitialStreamWindowSize`?

trustin:
Thanks for reviewing.

PR #1400, opened 2018-10-21
Allow specifying injected docs scripts per request.

anuraaga:
It can be useful to be able to inject a per-request script into the docs page (e.g., a randomly generated token for bypassing auth checks).


trustin:
Thanks, @anuraaga! Let me merge this after releasing 0.74.1 today which includes some hot fix for us.

PR #1402, opened 2018-10-23
Do not treat a class as automatically-convertible bean if it has a me…

hyangtack:
…mber which is not injectable

Motivation:
The server cannot be started if there is a parameter when there is a parameter unable to be injected on a constructor or method of a bean specified as `@RequestObject`. This behavior makes a user confused when he or she adds a member variable of one of the automatically-injectable types, such as `HttpRequest` and `RequestContext`, to his or her class which is not an automatically-convertiable class. We need to skip checking the class for this case because the class may be converted by one of the user-defined request converters.

Modifications:
- Throw `NoAnnotatedParameterException` instead of `IllegalArgumentException` when there is a parameter unable to be injected on a constructor or method of a bean specified as `@RequestObject`.


trustin:
Thanks, @hyangtack !

PR #1403, opened 2018-10-23
Overall cleanup of HTTP/2 handlers / Do not send GOAWAY to a closed connection

trustin:
Motivation:

- We currently add 2 `Http2Connection.Listener`s, `Http2GoAwayListener`
  and `Http2(Request|Response)Decoder`. We could add only a single
  listener so that Netty does not waste time to notify the events which
  is uninteresting to `Http2GoAwayListener`.
- We handle GOAWAY frames in more than one place, which is not great for
  readability.
- Netty invokes `onGoAwaySent()` callback even if the connection has
  been closed already and thus a GOAWAY frame is never sent. This makes
  `Http2GoAwayListener` log a false-positive message at WARN level.

Modifications

- Renamed `Http2GoAwayListener` to `Http2GoAwayHandler`.
- `Http2GoAwayHandler` is not an `Http2ConnectionAdapter` anymore.
- Moved all GOAWAY-related code to `Http2GoAwayHandler`.
- Clean-up
  - Reordered some constructor parameters so that the parameters of superclass
    appears first.
  - Removed `Http2Connection` from parameter list if there is
    `Http2ConnectionEncoder` in the parameter list, because
    `Http2Connection` can be retrieved via `Http2ConnectionEncoder.connection()`.
  - Cleaned up the code that injects verbose debug data into
    `Http2Exception`.
- Miscellaneous:
  - Fixed a bug where HTTP/2 graceful shutdown timeout it set to 0 when
    Armeria idle timeout is disabled.

Result:

- Fewer number of `Http2Connection.Listener`s
- Less scattered code
- Less noisy log messages

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/AbstractHttp2ConnectionHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/AbstractHttp2ConnectionHandler.java
@@ -20,20 +20,19 @@ import static io.netty.handler.codec.http2.Http2Error.INTERNAL_ERROR;
 
 import javax.annotation.Nullable;
 
-import com.google.common.base.MoreObjects;
-
 import com.linecorp.armeria.client.ClientFactory;
 import com.linecorp.armeria.common.Flags;
 import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.server.Server;
 
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
 import io.netty.handler.codec.http2.Http2ConnectionDecoder;
 import io.netty.handler.codec.http2.Http2ConnectionEncoder;
 import io.netty.handler.codec.http2.Http2ConnectionHandler;
 import io.netty.handler.codec.http2.Http2Exception;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/linecorp/armeria/internal/AbstractHttp2ConnectionHandler.java
```diff
@@ -80,22 +79,21 @@ protected void onConnectionError(ChannelHandlerContext ctx, boolean outbound,
...
-            } else if (http2Ex instanceof ClosedStreamCreationException) {
-                final ClosedStreamCreationException e = (ClosedStreamCreationException) http2Ex;
-                http2Ex = new ClosedStreamCreationException(e.error(), goAwayDebugData(e, cause), cause);
-            } else {
-                http2Ex = new Http2Exception(
-                        http2Ex.error(), goAwayDebugData(http2Ex, cause), cause, http2Ex.shutdownHint());
-            }
+    private static Http2Exception filterHttp2Exception(Throwable cause, @Nullable Http2Exception http2Ex) {
+        if (http2Ex != null) {
+            return http2Ex;
```
trustin: `Http2Exception` and its subtypes already have enough debug data, so I skipped filtering.

trustin:
Thanks for reviewing. Let me release 0.74.1 soon.

PR #1404, opened 2018-10-23
Invalidate the cached if there is new file in HttpVfs

linna-li:
Hi!
I want add the customized expire time in HttpFileServiceBuilder because want to use as:
because we found the 404 not found page was cached.
I am sorry i have not added unit test because i am not sure Could i do in this way or maybe there is better idea?

Any comment is appreciated!!

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
@@ -250,13 +250,13 @@ public final class HttpFileService extends AbstractHttpService {
         }
 
         final PathAndEncoding pathAndEncoding = new PathAndEncoding(path, contentEncoding);
-        CachedEntry entry = cache.getIfPresent(pathAndEncoding);
+        final CachedEntry entry = cache.getIfPresent(pathAndEncoding);
         if (entry == null) {
-            entry = cache.get(pathAndEncoding);
+            return cache.get(pathAndEncoding);
         } else {
-            if (config.vfs().get(path, contentEncoding).lastModifiedMillis() > entry.lastModifiedMillis()) {
+            if (config.vfs().get(path, contentEncoding).lastModifiedMillis() != entry.lastModifiedMillis()) {
                 cache.invalidate(pathAndEncoding);
-                entry = cache.get(pathAndEncoding);
+                return cache.get(pathAndEncoding);
             }
         }
         assert entry != null; // Non-existent entry will have lastModifiedMillis of 0.
--- a/core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
@@ -253,13 +253,13 @@ public final class HttpFileService extends AbstractHttpService {
         final CachedEntry entry = cache.getIfPresent(pathAndEncoding);
         if (entry == null) {
             return cache.get(pathAndEncoding);
-        } else {
-            if (config.vfs().get(path, contentEncoding).lastModifiedMillis() != entry.lastModifiedMillis()) {
-                cache.invalidate(pathAndEncoding);
-                return cache.get(pathAndEncoding);
-            }
         }
-        assert entry != null; // Non-existent entry will have lastModifiedMillis of 0.
+
+        if (config.vfs().get(path, contentEncoding).lastModifiedMillis() != entry.lastModifiedMillis()) {
+            cache.invalidate(pathAndEncoding);
+            return cache.get(pathAndEncoding);
+        }
+
         return entry;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
@@ -24,6 +24,7 @@ import java.nio.file.Path;
 import java.text.ParseException;
 import java.util.EnumSet;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
 
@@ -111,6 +112,7 @@ public final class HttpFileService extends AbstractHttpService {
         if (config.maxCacheEntries() != 0) {
             cache = Caffeine.newBuilder()
                             .maximumSize(config.maxCacheEntries())
+                            .expireAfterWrite(config.cacheExpireTime(), TimeUnit.MINUTES)
                             .recordStats()
                             .build(this::getEntryWithoutCache);
         } else {
--- a/core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
@@ -24,7 +24,6 @@ import java.nio.file.Path;
 import java.text.ParseException;
 import java.util.EnumSet;
 import java.util.Objects;
-import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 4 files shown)


linna-li:
@trustin , thanks for your suggestion very much! and i think we can use your idea! 
just one question, 
I am not very familiar with `HttpVfs`, is there always save the newest file?  so the `config.vfs().get()` can get the good html but not 404, am i understand right?

trustin:
@linna-li wrote:

> I am not very familiar with `HttpVfs`, is there always save the newest file?  so the `config.vfs().get()` can get the good html but not 404, am i understand right?

That's correct. `HttpVfs` does not cache anything unless implemented so. We have two `HttpVfs` implementations and `FileSystemHttpVfs` does not cache.

linna-li:
@trustin Thank you, i have revert the old change but check whether there is new file from user.

core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
```diff
@@ -249,7 +249,16 @@ private Entry getEntry(String path, @Nullable String contentEncoding) {
             return config.vfs().get(path, contentEncoding);
         }
 
-        final CachedEntry entry = cache.get(new PathAndEncoding(path, contentEncoding));
+        final PathAndEncoding pathAndEncoding = new PathAndEncoding(path, contentEncoding);
+        CachedEntry entry = cache.getIfPresent(pathAndEncoding);
+        if (entry == null) {
+            entry = cache.get(pathAndEncoding);
```
trustin: Could just return here and then the `else` block can be unwrapped.

core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
```diff
@@ -249,7 +249,16 @@ private Entry getEntry(String path, @Nullable String contentEncoding) {
...
+        } else {
+            if (config.vfs().get(path, contentEncoding).lastModifiedMillis() > entry.lastModifiedMillis()) {
```
trustin: Not sure which is better but perhaps `!=` is OK here?

core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
```diff
@@ -249,7 +249,16 @@ private Entry getEntry(String path, @Nullable String contentEncoding) {
...
+                cache.invalidate(pathAndEncoding);
+                entry = cache.get(pathAndEncoding);
```
trustin: Could just return without assignment.

core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
```diff
@@ -249,7 +249,16 @@ private Entry getEntry(String path, @Nullable String contentEncoding) {
             return config.vfs().get(path, contentEncoding);
         }
 
-        final CachedEntry entry = cache.get(new PathAndEncoding(path, contentEncoding));
+        final PathAndEncoding pathAndEncoding = new PathAndEncoding(path, contentEncoding);
+        CachedEntry entry = cache.getIfPresent(pathAndEncoding);
```
trustin: Because `entry` is used only once, we could inline this.

core/src/main/java/com/linecorp/armeria/server/file/HttpFileService.java
linna-li: yes, think all of the `cached file in entry` should from` config.vfs().get()` so there should be no possible that `config.vfs().get(path, contentEncoding).lastModifiedMillis() < entry.lastModifiedMillis()` right?

linna-li:
Sorry, another question: how is the `astModifiedMillis`property of an existing file can be updated?

trustin:
@linna-li wrote:

> Sorry, another question: how is the `lastModifiedMillis`property of an existing file can be updated?

If you modified the file, your O/S will do that for you.

If you just want to invalidate regardless if the file was modified, you can use `touch` command on your file or something equivalent.

If you created a new file, the `lastModifiedMillis` of a non-existing file is currently `0`, so it will become non-zero, invalidating the cache.

trustin:
So, if you want to do it programmatically in your test case, you can just modify the file by updating its content, like opening it with `FileOutputStream` and writing something into it. If it does not work, you can also use `File.setLastModified()` or `Files.setLastModifiedTime(Path, FileTime)` in JDK.

core/src/test/java/com/linecorp/armeria/server/file/HttpFileServiceTest.java
```diff
@@ -324,6 +325,26 @@ public void testFileSystemGet() throws Exception {
         }
     }
 
+    @Test
+    public void testFileSystemGet_newFile() throws Exception {
```
trustin: It does not test the case of adding a new file but modifying an existing file. Could you:

1. Rename this method to `testFileSystemGet_modifiedFile` and
2. Add another test method `testFileSystemGet_newFile`, which makes sure the first request returns 404 and the second request returns 200

..?

linna-li:
Updated, I added another unit test and firstly it passed but finally there is the following info from IntelliJ
I am not sure maybe some setting is wrong in my IntelliJ?
```
> Task :core:testStreaming FAILED
> Task :core:testNg FAILED
FAILURE: Build completed with 2 failures.
1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':core:testStreaming'.
> No tests found for given includes: [**/HttpServerStreamingTest**](include rules) [com.linecorp.armeria.server.file.HttpFileServiceTest.testFileSystemGet_newFile](filter.includeTestsMatching)
* Try:
==============================================================================
2: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':core:testNg'.
> No tests found for given includes: [com.linecorp.armeria.server.file.HttpFileServiceTest.testFileSystemGet_newFile](filter.includeTestsMatching)
* Try:
==============================================================================
```

core/src/test/java/com/linecorp/armeria/server/file/HttpFileServiceTest.java
linna-li: Thank you! I will add another test for newFile case,
I would like to confirm:
From the unit test, my understand is `final File barFile = new File(tmpDir, "bar.html");` this is like the place that saving cached file, so if we return new html, the content in this path will be refreshed, am I right?

trustin:
> Test failure

I think it's because you used the same file path (`bar.html`) for more than one test method. You need to use different file name for each test method so they do not clash.

linna-li:
I updated the path, but the exception is still here.
```
FAILURE: Build completed with 2 failures.
1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':core:testStreaming'.
> No tests found for given includes: [**/HttpServerStreamingTest**](include rules) [com.linecorp.armeria.server.file.HttpFileServiceTest](filter.includeTestsMatching)
* Try:
==============================================================================
2: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':core:testNg'.
> No tests found for given includes: [com.linecorp.armeria.server.file.HttpFileServiceTest](filter.includeTestsMatching)
* Try:
==============================================================================
```

core/src/test/java/com/linecorp/armeria/server/file/HttpFileServiceTest.java
trustin: Yes. To be precise, the cached content will be refreshed.

trustin:
I guess the failure is not relevant to this change but your build environment (IDE setup?).

trustin:
Thanks, @linna-li ! Looking forward to seeing your next pull request. ;-)

PR #1405, opened 2018-10-24
Deprecate @RequestObject in favor of @RequestBean

hyangtack:
Motivation:
There is no reason to put `@RequestObject` for a parameter which needs to be converted into an object with a request converter, because all parameters in the annotated service method should be injectable and the others can be identified by annotations or their types. So, any parameter which does not have an annotation can be tried to be converted with a request converter. 
However, in this case, we have an ambiguity which parameter can be converted into an object with a request converter or can be converted into a bean. So we need to introduce a new annotation `@RequestBean` which indicates a parameter needs to be converted into a bean, which is used relatively less than using a request converter.

Modifications:
- Deprecate `@RequestObject`.
- Add `@RequestBean` which is used for converting a request into a bean. 
- `@RequestConverter` can be put on a parameter of an annotated service method.
- Remove `@RequestObject` from tests, except for minimal test cases for testing backward compatibility.

Result:
A user does not need to put `@RequestObject` annotation on a parameter which needs to be converted into an object with a request converter any more.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
@@ -125,8 +125,8 @@ final class AnnotatedValueResolver {
 
     /**
      * Returns a list of {@link AnnotatedValueResolver} which is constructed with the specified
-     * {@link Executable}, {@code pathParams} and {@code objectResolvers}. The {@link Executable} can be
-     * one of {@link Constructor} or {@link Method}.
+     * {@link Executable}, {@code pathParams}, {@code converters} and {@code applyConverterForNoAnnotation}.
+     * The {@link Executable} can be one of {@link Constructor} or {@link Method}.
      */
     static List<AnnotatedValueResolver> of(Executable constructorOrMethod, Set<String> pathParams,
                                            List<RequestConverterFunction> converters,
@@ -215,7 +215,7 @@ final class AnnotatedValueResolver {
 
     /**
      * Returns a list of {@link AnnotatedValueResolver} which is constructed with the specified
-     * {@link Parameter}, {@code pathParams} and {@code converters}.
+     * {@link Parameter}, {@code pathParams}, {@code converters} and {@code applyConverterForNoAnnotation}.
      */
     static Optional<AnnotatedValueResolver> of(Parameter parameter, Set<String> pathParams,
                                                List<RequestConverterFunction> converters,
@@ -226,7 +226,7 @@ final class AnnotatedValueResolver {
 
     /**
      * Returns a list of {@link AnnotatedValueResolver} which is constructed with the specified
-     * {@link Field}, {@code pathParams} and {@code objectResolvers}.
+     * {@link Field}, {@code pathParams}.
      */
     static Optional<AnnotatedValueResolver> of(Field field, Set<String> pathParams) {
         // 'Field' is only used for converting a bean. So we need to pass 'applyConverterForNoAnnotation'
@@ -260,7 +260,7 @@ final class AnnotatedValueResolver {
         requireNonNull(typeElement, "typeElement");
         requireNonNull(type, "type");
         requireNonNull(pathParams, "pathParams");
-        requireNonNull(converters, "objectResolvers");
+        requireNonNull(converters, "converters");
 
         final Param param = annotatedElement.getAnnotation(Param.class);
         if (param != null) {
@@ -288,7 +288,8 @@ final class AnnotatedValueResolver {
             assert parameterizedType instanceof ParameterizedType : String.valueOf(parameterizedType);
             final Type actual = ((ParameterizedType) parameterizedType).getActualTypeArguments()[0];
             if (autoInjectableTypes.containsKey(actual)) {
-                return Optional.empty();
+                throw new IllegalArgumentException("Optional is not allowed for: " +
+                                                   actual.getClass().getSimpleName());
             }
         } else {
             // There should be no '@Default' annotation on 'annotatedElement' if 'annotatedElement' is
@@ -296,7 +297,7 @@ final class AnnotatedValueResolver {
             // So, 'typeElement' should be used when finding an injectable type because we need to check
             // syntactic errors like below:
             //
-            // void method1(@Default("a") ServiceRequestContext) { ... }
+            // void method1(@Default("a") ServiceRequestContext ctx) { ... }
             //
             final BiFunction<AnnotatedElement, Class<?>,
                     AnnotatedValueResolver> factory = autoInjectableTypes.get(type);
@@ -327,7 +328,7 @@ final class AnnotatedValueResolver {
             final BeanFactoryId beanFactoryId = AnnotatedBeanFactory.register(type, pathParams);
             AnnotatedBeanFactory.find(beanFactoryId)
                                 .orElseThrow(() -> new IllegalArgumentException(
-                                        "Cannot be registered to " +
+                                        "Failed to register to " +
                                         AnnotatedBeanFactory.class.getSimpleName() + ": " + annotatedElement));
             return Optional.of(ofRequestBean(annotatedElement, type, pathParams, beanFactoryId));
         }
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
@@ -278,7 +278,7 @@ final class AnnotatedValueResolver {
         }
 
         if (type == Optional.class) {
-            // Do not allow Optional<?> for the following auto-injectable types:
+            // Warn if Optional<?> is used for the following auto-injectable types:
             //  - RequestContext and ServiceRequestContext
             //  - Request and HttpRequest
             //  - AggregatedHttpMessage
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -239,77 +240,152 @@
...
+            //  - RequestContext and ServiceRequestContext
+            //  - Request and HttpRequest
+            //  - AggregatedHttpMessage
+            //  - HttpParameters
+            //  - Cookies
+            final Type parameterizedType = parameterizedTypeOf(annotatedElement);
+            assert parameterizedType instanceof ParameterizedType : String.valueOf(parameterizedType);
+            final Type actual = ((ParameterizedType) parameterizedType).getActualTypeArguments()[0];
+            if (autoInjectableTypes.containsKey(actual)) {
+                return Optional.empty();
```
minwoox: Don't we have to throw an exception because we do not allow Optional<?> for auto-injectable types?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -239,77 +240,152 @@
...
+            }
+        } else {
+            // There should be no '@Default' annotation on 'annotatedElement' if 'annotatedElement' is
+            // different from 'typeElement', because it was checked before calling this method.
+            // So, 'typeElement' should be used when finding an injectable type because we need to check
+            // syntactic errors like below:
+            //
+            // void method1(@Default("a") ServiceRequestContext) { ... }
```
minwoox: Do we need to put variable name here? `ServiceRequestContext ctx`

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -239,77 +240,152 @@
...
+            }
+        }
+
+        final RequestBean requestBean = annotatedElement.getAnnotation(RequestBean.class);
+        if (requestBean != null) {
+            // In this case, this element must be registered to AnnotatedBeanFactory.
+            final BeanFactoryId beanFactoryId = AnnotatedBeanFactory.register(type, pathParams);
+            AnnotatedBeanFactory.find(beanFactoryId)
+                                .orElseThrow(() -> new IllegalArgumentException(
+                                        "Cannot be registered to " +
```
minwoox: fail to register.... for consistency?

core/src/main/java/com/linecorp/armeria/server/annotation/RequestObject.java
```diff
@@ -26,7 +26,11 @@
  *
  * @see RequestConverterFunction
  * @see RequestConverter
+ * @deprecated No more put {@link RequestObject} on the parameter which needs to be converted by one
```
minwoox: No more putting or Do not put?

core/src/main/java/com/linecorp/armeria/server/annotation/RequestObject.java
```diff
@@ -26,7 +26,11 @@
...
+ *             of the {@link RequestConverterFunction}s. However, use {@link RequestBean}
```
minwoox: `Instead, use`? or just `Use`?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -439,39 +439,38 @@ converter is not able to convert the request.
     }
 
 Then, you can write your service method as follows. Note that a request converter will work on the parameters
-which are annotated with :api:`@RequestObject`.
+which are not annotated and are not automatically injectable types.
```
minwoox: which are either not annotated or not ...?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -239,77 +240,152 @@
...
                                                        Set<String> pathParams,
-                                                       List<RequestObjectResolver> objectResolvers) {
+                                                       List<RequestConverterFunction> converters,
+                                                       boolean applyConverterForNoAnnotation) {
         requireNonNull(annotatedElement, "annotatedElement");
         requireNonNull(typeElement, "typeElement");
         requireNonNull(type, "type");
         requireNonNull(pathParams, "pathParams");
-        requireNonNull(objectResolvers, "objectResolvers");
+        requireNonNull(converters, "objectResolvers");
```
minwoox: "converters"?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
hyangtack: Changed to throw IAE. Thanks!

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -239,77 +240,153 @@
...
         final Header header = annotatedElement.getAnnotation(Header.class);
         if (header != null) {
-            return Optional.of(ofHeader(header, annotatedElement, typeElement, type));
+            return Optional.of(ofHeader(findName(header, typeElement), annotatedElement, typeElement, type));
         }
 
+        if (type == Optional.class) {
+            // Do not allow Optional<?> for the following auto-injectable types:
+            //  - RequestContext and ServiceRequestContext
+            //  - Request and HttpRequest
```
trustin: Perhaps we could just inject an `Optional` that's always present, although it's not the most efficient setup? (We could warn about it though.)

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -239,77 +240,153 @@
...
+            AnnotatedBeanFactory.find(beanFactoryId)
+                                .orElseThrow(() -> new IllegalArgumentException(
+                                        "Failed to register to " +
+                                        AnnotatedBeanFactory.class.getSimpleName() + ": " + annotatedElement));
+            return Optional.of(ofRequestBean(annotatedElement, type, pathParams, beanFactoryId));
+        }
+
+        final RequestConverter requestConverter = annotatedElement.getAnnotation(RequestConverter.class);
+        if (requestConverter != null) {
+            // If we are analyzing a service method, we need to deal an element, which does not have any
```
trustin: deal -> deal with

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -460,36 +557,27 @@ private static AnnotatedValueResolver ofInjectableTypes(AnnotatedElement annotat
...
-            if (value != null) {
-                return value;
-            }
-            throw new IllegalArgumentException("No suitable request converter found for a @" +
-                                               RequestObject.class.getSimpleName() + " '" +
-                                               resolver.elementType().getSimpleName() + '\'');
-        };
+    static boolean isRequestConverterType(@Nullable Class<? extends Annotation> type) {
+        if (type == null) {
+            return false;
```
trustin: OK to remove this `if` block.

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -683,10 +766,16 @@ private Builder(AnnotatedElement annotatedElement, Type type) {
...
+         *     <li>{@link Param}</li>
+         *     <li>{@link Header}</li>
+         *     <li>{@link RequestBean}</li>
+         *     <li>{@link RequestConverter}</li>
+         * </ul>
          */
-        private Builder annotation(Annotation annotation) {
-            this.annotation = annotation;
+        private Builder annotationType(Class<? extends Annotation> annotationType) {
+            this.annotationType = annotationType;
```
trustin: I'm curious if it's a good idea to add an assertion here to make sure `annotationType` is one of the four types.

core/src/main/java/com/linecorp/armeria/server/annotation/RequestObject.java
```diff
@@ -26,7 +26,11 @@
  *
  * @see RequestConverterFunction
  * @see RequestConverter
+ * @deprecated No more putting {@link RequestObject} on the parameter which needs to be converted by one
```
trustin: No more putting .. -> `There's no need to put {@link RequestObject} annotation anymore on a parameter converted by a {@link RequestConverterFunction}.`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -439,39 +439,38 @@ converter is not able to convert the request.
     }
 
 Then, you can write your service method as follows. Note that a request converter will work on the parameters
-which are annotated with :api:`@RequestObject`.
+which are neither annotated nor automatically injectable types.
```
trustin: injectable -> injected?

hyangtack:
@trustin and I talked about this issue at office. And finally we realized that it's not necessary to deprecate `@RequestObject` by introducing new annotations like `@RequestBean`. Instead, it's enough to make `@RequestObject` annotation optional in the parameter list of a service method. 

So let me close this PR now, and then make a new PR for making `@RequestObject` annotation optional in the parameter list of a service method. 

Thanks, reviewers. :-)

PR #1406, opened 2018-10-25
Fail with UnprocessedRequestException when writing first request head…

trustin:
…er to an inactive session

Motivation:

`HttpRequestSubscriber` currently fails the request with
`ClosedSessionException` when it's writing the first request headers to
an inactive session. However, the request has to fail with
`UnprocessedRequestException` in this case because it is very sure that
the request was not sent at all.

Modifications:

- Fail with `UnprocessedRequestException` instead of `ClosedSessionException`

Result:

- More safely retriable requests


trustin:
Thanks!

PR #1412, opened 2018-10-25
`@RequestObject` can be omitted in the parameter list of an annotated…

hyangtack:
… method

Motivation:
We tried to deprecate `@RequestObject` in #1405 at first, but it was not useful different from our intention.
It would be more useful make `@RequestObject` able to be omitted in the parameter list of an annotated method.

Modifications:
- A parameter which is not annotated will be handled the same as when a `@RequestObject` is specified, only for a service method.
- Revert some tests which was fixed in #1402 in order to raise an exception when we can guess a user's intention.

Result:
A user can omit `@RequestObject` in the parameter list of an annotated method.


core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -201,27 +203,51 @@
...
+                //     ...
+                //     BeanA(@Param("a") int a, int b) { ... }
+                // }
+                //
+                // But for the automatically injected types such as RequestContext and HttpRequest, etc.
+                // it is not easy to understand what a user intends for. So we ignore that.
+                //
+                // class BeanB {
+                //     ...
+                //     BeanB(ServiceRequestContext ctx, int b) { ... }
```
minwoox: Could you add a comment that this bean will not be automatically injected to clarify?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -129,11 +129,12 @@
...
      * Returns a list of {@link AnnotatedValueResolver} which is constructed with the specified
-     * {@link Executable}, {@code pathParams} and {@code objectResolvers}. The {@link Executable} can be
-     * one of {@link Constructor} or {@link Method}.
+     * {@link Executable}, {@code pathParams}, {@code objectResolvers} and {@code noAnnotationAsRequestObject}.
+     * The {@link Executable} can be one of {@link Constructor} or {@link Method}.
      */
     static List<AnnotatedValueResolver> of(Executable constructorOrMethod, Set<String> pathParams,
-                                           List<RequestObjectResolver> objectResolvers) {
+                                           List<RequestObjectResolver> objectResolvers,
+                                           boolean noAnnotationAsRequestObject) {
```
trustin: `implicitRequestObjectAnnotation`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -201,27 +203,51 @@
...
         if (list.isEmpty()) {
             throw new NoAnnotatedParameterException(constructorOrMethod.toGenericString());
         }
         if (list.size() != parameters.length) {
-            throw new NoAnnotatedParameterException("Unsupported parameter exists: " +
-                                                    constructorOrMethod.toGenericString());
+            if (list.stream().anyMatch(r -> r.annotationType() != null)) {
+                //
+                // If a user specify one of @Param, @Header or @RequestObject on the parameter list,
+                // it clearly means that the user wanted the class to be converted into a bean. e.g.
```
trustin: wants to convert the parameter into a bean.

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -245,11 +273,16 @@
      * @param pathParams       a set of path variables.
      * @param objectResolvers  a list of {@link RequestObjectResolver} to be evaluated for the objects which
      *                         are annotated with {@link RequestObject} annotation.
+     * @param noAnnotationAsRequestObject a flag which indicates how to handle an element which is not
+     *                                    annotated. If it is {@code true}, it would be handled the same as
+     *                                    when a {@link RequestObject} is specified. Otherwise, the element
+     *                                    would be ignored.
```
trustin: Not sure this is the best description, but:

```
@param implicitRequestObjectAnnotation {@code true} if an element is always treated like it is annotated
                                       with {@link RequestObject} so that conversion is performed.
                                       {@code false} if an element has to be annotated with
                                       {@link RequestObject} explicitly to get converted.
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -439,7 +439,8 @@ converter is not able to convert the request.
     }
 
 Then, you can write your service method as follows. Note that a request converter will work on the parameters
-which are annotated with :api:`@RequestObject`.
+which are annotated with :api:`@RequestObject`. You can omit the annotation in the parameter list
+of your service method.
```
trustin: How about omitting the 'Note that...' sentence completely so our users avoid using `@RequestObject` where unnecessary?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -489,12 +518,19 @@ Injecting value of parameters and HTTP headers into a Java object
 
 Armeria provides a generic built-in request converter for Java objects which is activated by
```
trustin: Ditto. How about avoid mentioning `@RequestObject` completely here?

```
Armeria provides a generic built-in request converter that converts a request into a Java bean.
Just define a plain old Java class and specify it as a parameter of your service method.
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -489,12 +518,19 @@ Injecting value of parameters and HTTP headers into a Java object
...
 
     public class MyAnnotatedService {
-        @Post("/hello")
-        public HttpResponse hello(@RequestObject MyRequestObject myRequestObject) { ... }
+        @Post("/hello1")
+        public HttpResponse hello1(@RequestObject MyRequestObject myRequestObject) { ... }
+
+        @Post("/hello2")
+        public HttpResponse hello2(MyRequestObject myRequestObject) {
+            // Will work the same as 'hello1'.
```
trustin: We could revert this part.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -489,12 +518,19 @@ Injecting value of parameters and HTTP headers into a Java object
...
+        }
     }
 
 Besides the annotated service class, you also need to create ``MyRequestObject`` and put :api:`@Param` or
```

```
We also need to define a class called ``MyRequestObject`` which was referred in our service method
``hello()``. To tell Armeria which constructor parameters, setter methods or fields has to be 
injected with what value, we should put :api:`@Param`, :api:`@Header`, :api:`@RequestObject`
annotations on any of the following elements:
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -517,7 +553,7 @@ or HTTP headers:
         @Header("age") // This field will be injected by the value of HTTP header "age".
         private int age;
 
-        @RequestObject // This field will be injected by another request converter.
+        @RequestObject // This field will be injected by another request converter. You cannot omit this annotation.
```
trustin: Probably OK to omit the added sentence :-)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -464,8 +465,36 @@ which are annotated with :api:`@RequestObject`.
 Armeria also provides built-in request converters such as, a request converter for a Java Bean,
 :api:`JacksonRequestConverterFunction` for a JSON document, :api:`StringRequestConverterFunction`
 for a string and :api:`ByteArrayRequestConverterFunction` for binary data. They will be applied
-after your request converters by default, so you can use these built-in converters by just putting
-:api:`@RequestObject` annotation on the parameters which you want to convert.
+after your request converters by default, so you can use these built-in converters by just specifying
```
trustin: Perhaps OK to omit 'by default'?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -464,8 +465,36 @@ which are annotated with :api:`@RequestObject`.
...
+a parameter with desired type to be converted in the parameter list of your service method.
```
trustin: 'by just specifying ..' sounds too complicated. How about:

```
so you don't have to specify any `@RequestConverter` annotations:
```

site/src/sphinx/server-annotated-service.rst
```diff
@@ -464,8 +465,36 @@ which are annotated with :api:`@RequestObject`.
...
+        // 'application/binary' or none.
+        @Post("/hello5")
+        public HttpResponse hello5(byte[] body) { ... }
+
+        @Post("/hello6")
+        public HttpResponse hello6(HttpData body) { ... }
+    }
 
 In some cases, :api:`@RequestObject` annotation may have a request converter as its value.
 Assume that you have a Java class named ``MyRequest`` that it is usually able to be converted by
```
trustin: This section could be rewritten or removed entirely.

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
hyangtack: Added [here](https://github.com/line/armeria/pull/1412/files#diff-9df84bbcae0217f0942af4e4d6a3f8c0R215). :-)

site/src/sphinx/server-annotated-service.rst
hyangtack: Will fix soon!

hyangtack: Fixed

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -174,7 +177,27 @@
             }
 
             resolver = of(constructorOrMethod,
-                          parameters[0], parameters[0].getType(), pathParams, objectResolvers);
+                          parameters[0], parameters[0].getType(), pathParams, objectResolvers,
+                          implicitRequestObjectAnnotation);
+        } else if (constructorOrMethod.getAnnotationsByType(RequestConverter.class).length > 0 &&
+                   parameters.length == 1) {
+            //
+            // Filter out like the following case:
```
trustin: Filter out the cases like the following:

core/src/main/java/com/linecorp/armeria/server/annotation/RequestConverter.java
```diff
@@ -28,12 +28,38 @@
  * Specifies a {@link RequestConverterFunction} class which converts an {@link AggregatedHttpMessage} to
  * an object.
  *
+ * <p>It can be specified on a class, a method and a parameter in an annotated service.
+ * Depending on where it is specified, its scope can be determined. e.g.
```
trustin: `Its scope is determined by where it is specified, e.g.` ?

core/src/main/java/com/linecorp/armeria/server/annotation/RequestObject.java
```diff
@@ -22,18 +22,51 @@
...
+ * <p>A {@link RequestObject} can be implicitly applied when:
+ * <ul>
+ *     <li>A parameter is neither annotated nor automatically injected type in an annotated method.</li>
+ *     <li>A {@link RequestConverter} annotation is specified on a parameter in an annotated method
+ *     or a setter of a bean.</li>
+ *     <li>A {@link RequestConverter} annotation is specified on a field of a bean.</li>
+ *     <li>A {@link RequestConverter} annotation is specified on a method or constructor which has
+ *     only one parameter.</li>
+ * </ul>
+ * The following example shows when to apply a {@link RequestObject} implicitly.
```
trustin: explicitly?

core/src/main/java/com/linecorp/armeria/server/annotation/RequestObject.java
```diff
@@ -22,18 +22,51 @@
...
+ * >
+ * >     @RequestConverter(BobConverter.class)
+ * >     private Bob b;
+ * >
+ * >     private Charlie c;
+ * >     private David d;
+ * >     private Erin e;
+ * >
+ * >     @RequestObject
+ * >     private Frank f;
```
trustin: How about adding some comment here why `@RequestObject` has to be specified explicitly?

core/src/main/java/com/linecorp/armeria/server/AnnotatedValueResolver.java
```diff
@@ -201,27 +224,52 @@
...
                                     .map(Optional::get)
                                     .collect(Collectors.toList()));
         if (list.isEmpty()) {
             throw new NoAnnotatedParameterException(constructorOrMethod.toGenericString());
         }
         if (list.size() != parameters.length) {
-            throw new NoAnnotatedParameterException("Unsupported parameter exists: " +
-                                                    constructorOrMethod.toGenericString());
+            // There are parameters which cannot be resolved, so we cannot accept this constructor or method
+            // as an annotated bean or method. But we can handle this kind of situation as two ways as follows.
```
minwoox: How about `... or method. We handle this case in two ways as follows.` ?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -424,82 +424,109 @@ converter is not able to convert the request.
...
 
         @Post("/hola")
-        @RequestConverter(MySpanishRequestConverter.class)
-        public HttpResponse hola(@RequestObject MySpanishObject myObject) {
-            // MySpanishRequestConverter will be tried to convert a request first.
-            // MyRequestConverter will be used if MySpanishRequestConverter fell through.
+        @RequestConverter(ToSpanishConverter.class)
+        public HttpResponse hola(Greeting greeting) {
+            // ToSpanishConverter will be tried to convert a request first.
+            // ToEnglishConverter will be used if ToSpanishConverter fell through.
```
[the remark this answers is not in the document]
trustin: I guess 500 will be returned if a converter throws an exception rather than falling through? 'fell through' is correct in my opinion.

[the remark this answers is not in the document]
minwoox: Ah, I thought it was more clear using the word `fail`, but now I realized that it can confuse others, so fell through is correct. :)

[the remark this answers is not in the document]
hyangtack: Yes. 500 or other status code will be returned, depending on what exception is raised and how exception handlers handle it.

core/src/main/java/com/linecorp/armeria/server/annotation/RequestObject.java
hyangtack: It's hard to explain why, because it must be specified by default. So, I added comments on the element that `@RequestObject` is implicitly applied. What do you think?

core/src/main/java/com/linecorp/armeria/server/annotation/RequestObject.java
hyangtack: Please refer to https://github.com/line/armeria/pull/1412#discussion_r230257679

site/src/sphinx/server-annotated-service.rst
```diff
@@ -523,10 +523,11 @@ Just define a plain old Java class and specify it as a parameter of your service
...
     }
 
-We also need to define a class called ``MyRequestObject`` which was referred in our service method
-``hello()``. To tell Armeria which constructor parameters, setter methods or fields has to be
-injected with what value, we should put :api:`@Param`, :api:`@Header`, :api:`@RequestObject`
-annotations on any of the following elements:
+
+
+We also need to define the ``MyRequestObject`` class which was used in the method ``hello()`` above.
+To tell Armeria which constructor parameters, setter methods or fields has to be injected with what value,
```
trustin: parameter, setter method or field has to be injected

PR #1414, opened 2018-10-27
Fix failures in HttpFileServiceTest

trustin:
Motivation:

f5251f1b83adf4fa9229927d8ac3134a13ed231f introduced the failure of
`HttpFileServiceTest.testFileSystemGet_modifiedFile()` on Windows.

Modifications:

- Update the last write time of the file explicitly so that there is no
  chance of having the same last write time.

Result:

- Build stability


trustin:
Thanks!

PR #1415, opened 2018-10-27
Fix a bug where RequestLog is not complete under certain circumstances

minwoox:
Motivation:
The `RequestLog` should be complete because the active request count of an `EventLoop`
is decreased when the `ReqeustLog` is complete.
However, if an exception is raised even before sending an `HttpRequest` by specifying
invalid URI, failing to resolve URI, etc, the `RequestLog` is not complete.

Modifications:
- Fix to complete `RequestLog`
- Fix to abort the `HttpRuest`

Result:

- Less bugs

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
@@ -30,10 +30,10 @@ import com.google.common.base.Strings;
 
 import com.linecorp.armeria.client.pool.KeyedChannelPool;
 import com.linecorp.armeria.client.pool.PoolKey;
-import com.linecorp.armeria.common.ClosedSessionException;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.SessionProtocol;
+import com.linecorp.armeria.common.logging.RequestLogBuilder;
 import com.linecorp.armeria.internal.PathAndQuery;
 
 import io.netty.channel.Channel;
@@ -58,8 +58,9 @@ final class HttpClientDelegate implements Client<HttpRequest, HttpResponse> {
     @Override
     public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Exception {
         if (!sanitizePath(req)) {
-            req.abort();
-            return HttpResponse.ofFailure(new IllegalArgumentException("invalid path: " + req.path()));
+            final IllegalArgumentException cause = new IllegalArgumentException("invalid path: " + req.path());
+            handleEarlyRequestException(ctx, req, cause);
+            return HttpResponse.ofFailure(cause);
         }
 
         final Endpoint endpoint = ctx.endpoint().resolve(ctx)
@@ -94,7 +95,9 @@ final class HttpClientDelegate implements Client<HttpRequest, HttpResponse> {
         if (resolveFuture.isSuccess()) {
             executeWithIpAddr(ctx, endpoint, resolveFuture.getNow().getAddress().getHostAddress(), req, res);
         } else {
-            res.close(resolveFuture.cause());
+            final Throwable cause = resolveFuture.cause();
+            handleEarlyRequestException(ctx, req, cause);
+            res.close(cause);
         }
     }
 
@@ -118,7 +121,9 @@ final class HttpClientDelegate implements Client<HttpRequest, HttpResponse> {
             final Channel ch = channelFuture.getNow();
             invoke0(ch, ctx, req, res, poolKey);
         } else {
-            res.close(channelFuture.cause());
+            final Throwable cause = channelFuture.cause();
+            handleEarlyRequestException(ctx, req, cause);
+            res.close(cause);
         }
     }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/client/UserClient.java
```diff
@@ -158,7 +160,15 @@ protected final O execute(@Nullable EventLoop eventLoop,
         try (SafeCloseable ignored = ctx.push()) {
             return delegate().execute(ctx, req);
         } catch (Throwable cause) {
-            ctx.logBuilder().endResponse(cause);
+            final RequestLogBuilder logBuilder = ctx.logBuilder();
+            if (!ctx.log().isAvailable(RequestLogAvailability.REQUEST_START)) {
+                // An exception is raised even before sending a request, so end the request with the exception.
+                if (req instanceof HttpRequest) {
```
hyangtack: nit: I'm not sure what's the best. How about moving these code to fallback of `DefaultHttpClient` and `DefaultTHttpClient`? Because I'm not sure that it's good choice that `HttpRequest` is used here.

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -188,6 +193,14 @@ private static boolean sanitizePath(HttpRequest req) {
         return true;
     }
 
+    private static void handleEarlyException(ClientRequestContext ctx,
+                                             HttpRequest req, Throwable cause) {
+        req.abort();
+        final RequestLogBuilder logBuilder = ctx.logBuilder();
+        logBuilder.endRequest(cause);
```
trustin: I'm curious if `req.abort()` will trigger `logBuilder.endRequest()` anyway. Could you check?

core/src/main/java/com/linecorp/armeria/client/UserClient.java
trustin: Agreed.

core/src/test/java/com/linecorp/armeria/client/HttpClientWithRequestLogTest.java
```diff
@@ -0,0 +1,136 @@
...
+        assertThatThrownBy(() -> client.execute(req).aggregate().get()).isInstanceOf(Exception.class);
+
+        await().untilAsserted(() -> assertThat(requestCauseHolder.get()).isNotNull());
+        await().untilAsserted(() -> assertThat(responseCauseHolder.get()).isNotNull());
+        await().untilAsserted(() -> assertThat(req.isComplete()).isTrue());
+    }
+
+    @Test
+    public void connectionError() {
+        // According to rfc7805, TCP port number 1 is not used so a connection error always happens.
```
trustin: TIL :smile:

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -199,7 +212,9 @@ void invoke0(Channel channel, ClientRequestContext ctx,
             if (sessionProtocol == null) {
                 needsRelease = false;
                 try {
-                    res.close(ClosedSessionException.get());
+                    final UnprocessedRequestException cause = UnprocessedRequestException.get();
+                    handleEarlyException(ctx, req, cause);
```
trustin: Seems like this line is not covered according to codecov.io report. Could you try to add a test case that covers this?

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -188,6 +193,14 @@ private static boolean sanitizePath(HttpRequest req) {
         return true;
     }
 
+    private static void handleEarlyException(ClientRequestContext ctx,
```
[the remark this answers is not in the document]
minwoox: This is called when an exception is raised even before sending the `HttpRequest`. I think there should be a word implying that. At first, I was trying to name it as something like `handleEaylyRequestException` but it was too long so I shortened it. What do you think?

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
minwoox: Unfortunately, `req.abort()` does not trigger it unless it is subscribed by the `HttpRequestSubscriber`. So this is needed it, or I can change the `ReqeustLogBuilder.endResponse(cause)` to end request when the `Request` has not started. What do you think?

core/src/main/java/com/linecorp/armeria/client/UserClient.java
minwoox: Got it~!

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
trustin: Thanks for explanation. Let's leave it as it is.

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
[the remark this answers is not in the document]
hyangtack: I see. Thank you for explanation. :-)

trustin:
Thanks, @minwoox!

PR #1417, opened 2018-10-29
Don't rewrite the user's path if it is valid since it may be escaped.

### LINKED ISSUE #1416: Cannot issue HTTP request with escaped slashes in path
I think it happened somewhere around #1356 but whereas previously `HttpClientDelegate` sent the user's request path correctly, now it decodes the user's path, which is not correct. It means currently it's not possible to issue a request with, for example, escaped path parameters as they get decoded and sent as a raw path instead.

While refactoring `PathAndQuery` I guess we forgot that `HttpClientDelegate` also uses it for sanitizing.

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java#L172

anuraaga:
Users may have a specific reason to use special escaping in a client request (it may be the contract of the server such as [here](https://cloud.google.com/storage/docs/json_api/#encoding)). While the client should make sure request paths are valid, it should not overwrite whatever escaping the user may have used.

A side effect is double slashes are sent as is instead of rewriting them to single slashes. It's conceivable that a server requires double slashes for whatever reason and the client should support it (grudgingly).

Fixes #1416

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
@@ -68,6 +68,12 @@ public final class PathAndQuery {
 
     /**
      * A special byte which tells {@link #encodeToPercents(Bytes, boolean)} to translate it to
+     * {@code "%2F"}.
+     */
+    private static final int ENCODED_SLASH = 0xFC;
+
+    /**
+     * A special byte which tells {@link #encodeToPercents(Bytes, boolean)} to translate it to
      * {@code "%26"}.
      */
     private static final int ENCODED_AMPERSAND = 0xFD;
--- a/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
@@ -68,12 +68,6 @@ public final class PathAndQuery {
 
     /**
      * A special byte which tells {@link #encodeToPercents(Bytes, boolean)} to translate it to
-     * {@code "%2F"}.
-     */
-    private static final int ENCODED_SLASH = 0xFC;
-
-    /**
-     * A special byte which tells {@link #encodeToPercents(Bytes, boolean)} to translate it to
      * {@code "%26"}.
      */
     private static final int ENCODED_AMPERSAND = 0xFD;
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -300,6 +300,15 @@ public void onComplete() {
                 }, ctx.eventLoop());
                 return res;
             });
+
+            sb.service("glob:/oneparam/**", ((ctx, req) -> {
+                // The client was able to send a request with an escaped path param. Armeria servers always
+                // decode the path so ctx.path == '/oneparam/foo/bar' here.
+                if (req.headers().path().equals("/oneparam/foo%2Fbar")) {
```
trustin: How about checking `decodedPath()` as well?

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -66,6 +66,12 @@
     private static final Bytes EMPTY_QUERY = new Bytes(0);
     private static final Bytes ROOT_PATH = new Bytes(new byte[] { '/' });
 
+    /**
+     * A special byte which tells {@link #encodeToPercents(Bytes, boolean)} to translate it to
+     * {@code "%2F"}.
+     */
+    private static final int ENCODED_SLASH = 0xFC;
```
trustin: Seems unused? :-)

anuraaga: Oops, you found my experimentation ;)

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
anuraaga: Added the check for `ctx.path()` - it does make me wonder, though, why is there a `decodedPath()` if `path()` is already decoded? :thinking:

anuraaga:
Oops - should have run the whole test too. The change prevents double slash suppression - is it the job of a client library to do such editing of a URL? Guess browsers don't - https://stackoverflow.com/questions/20523318/is-a-url-with-in-the-path-section-valid

trustin:
> Oops - should have run the whole test too. The change prevents double slash suppression - is it the job of a client library to do such editing of a URL? Guess browsers don't - stackoverflow.com/questions/20523318/is-a-url-with-in-the-path-section-valid

I think we don't need double slash suppression on the client side, although it doesn't look like a good idea to rely on that especially on the server side. So yeah, it looks good as it is.

trustin:
I resolved the conflict with #1415. Could you confirm it just in case I made any mistakes, @anuraaga ?

trustin:
Thanks, @anuraaga !

PR #1418, opened 2018-10-30
Split RetryStrategy into RetryStrategy and RetryStrategyWithContent

### LINKED ISSUE #1408: Use StreamMessageDuplicator only where it's needed
We use `StreamMessageDuplicator` in `RetryingHttpClient` and `CircuitBreakerHttpClient`. The `StreamMessageDuplicator` is needed to inspect the content of `HttpRequest` and `HttpResponse` because the nature of the stream messages which is they can be subscribed only once, so we use the `StreamMessageDuplicator` to subscribe the content multiple times.
However, in most cases, the `RetryingHttpClient` and `CircuitBreakerHttpClient` do not need to look at the content, but just use the headers to decide what to do.
If we can remove `StreamMessageDuplicator` and use it only where it's needed, the performance of Armeria Client will go up.

So how do we do that?
We can add a method to `*Strategy` to decide whether the strategy uses the duplicator or not.
Also we can change the signature of the `shouldRertry` from
```
public interface RetryStrategy<I extends Request, O extends Response> {
    CompletionStage<Backoff> shouldRetry(I request, O response);
}
```
to
```
public interface RetryStrategy<T extends Response> {
    CompletionStage<Backoff> shouldRetry(RequestContext ctx, @Nullable T response);
}
```
so that the strategy can use the log information from the `ctx` to make a decision.
Therefore, new `RetryStrategy` will be:
```
public interface RetryStrategy<I extends Request, O extends Response> {

    CompletionStage<Backoff> shouldRetry(I request, O response) {..}

    boolean needsResponseStreaming(RequestContext ctx);
}
```

minwoox:
Motivation:
When using `RetryingHttpClient`, we use `HttpResponseDuplicator` to duplicate the `HttpResponse` in order to determine to retry or not.
However, in most cases, we just need the response headers to decide and don't need the content.
If we see the content only when we really have to, we can minimize the usage of duplicator which reduces the performance. This approach works for `CircuitBreakerHttpClient` as well.

Modifications:
- Change the method of `RetryStrategy` to `shouldRetry(RequestContext ctx, Throwable cause)`
- Add `RetryStrategyWithContent`
- Add `CircuitBreakerClientBuiler`
- Change the method of `CircuitBreakerStrategy` to `shouldReportAsSuccess(RequestContext ctx, Throwable cause)`
- Add `CircuitBreakerStrategyWithContent`
- Add a benchmark to see the difference between using `Duplicator` and without it

Result:
- Better performance
```
Benchmark                 Mode  Cnt      Score     Error  Units
WithDuplicator.empty     thrpt  100  87399.115 ± 101.651  ops/s
WithoutDuplicator.empty  thrpt  100  91787.170 ±  73.129  ops/s
```
- Close #1408

Todo:
- Change `FixedStreamMessage` duplicable
- Don't let `RetryingHttpClient` use `HttpRequestDuplicator` when the content-length is zero

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/retry/RetryingHttpClientBase.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/retry/RetryingHttpClientBase.java
@@ -14,7 +14,7 @@
  * under the License.
  */
 
-package com.linecorp.armeria.retrofit2.shared;
+package com.linecorp.armeria.core.client.retry;
 
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.Scope;
@@ -22,23 +22,23 @@ import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.TearDown;
 
+import com.linecorp.armeria.client.HttpClient;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.server.Server;
 import com.linecorp.armeria.server.ServerBuilder;
 import com.linecorp.armeria.server.ServerPort;
 
 @State(Scope.Benchmark)
-public abstract class SimpleBenchmarkBase {
+public abstract class RetryingHttpClientBase {
 
     private Server server;
-    private SimpleBenchmarkClient client;
+    private HttpClient client;
 
     @Setup
-    public void start() throws Exception {
+    public void start() {
         server = new ServerBuilder()
-                .https(0)
+                .http(0)
                 .service("/empty", (ctx, req) -> HttpResponse.of("\"\""))
-                .tlsSelfSigned()
                 .build();
         server.start().join();
         client = newClient();
@@ -49,17 +49,17 @@ public abstract class SimpleBenchmarkBase {
         server.stop().join();
     }
 
-    protected abstract SimpleBenchmarkClient newClient() throws Exception;
+    protected abstract HttpClient newClient();
 
     protected String baseUrl() {
         final ServerPort httpPort = server.activePorts().values().stream()
-                                          .filter(ServerPort::hasHttps).findAny()
+                                          .filter(ServerPort::hasHttp).findAny()
                                           .get();
-        return "https://127.0.0.1:" + httpPort.localAddress().getPort();
+        return "h2c://127.0.0.1:" + httpPort.localAddress().getPort();
     }
 
     @Benchmark
     public void empty() {
-        client.empty().join();
+        client.get("/empty").aggregate().join();
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClientBuilder.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.client.circuitbreaker;
+
+import static com.google.common.base.Preconditions.checkState;
+import static java.util.Objects.requireNonNull;
+
+import java.util.function.Function;
+
+import javax.annotation.Nullable;
+
+import com.google.common.base.MoreObjects;
+
+import com.linecorp.armeria.client.Client;
+import com.linecorp.armeria.common.Request;
+import com.linecorp.armeria.common.Response;
+
+/**
+ * Builds a new {@link CircuitBreakerClient} or its decorator function.
+ *
+ * @param <T> the type of {@link CircuitBreakerClientBuilder}
+ * @param <U> the type of the {@link Client} that this builder builds or decorates
+ * @param <I> the type of outgoing {@link Request} of the {@link Client}
+ * @param <O> the type of incoming {@link Response} of the {@link Client}
+ */
+public abstract class CircuitBreakerClientBuilder<
+        T extends CircuitBreakerClientBuilder<T, U, I, O>, U extends CircuitBreakerClient<I, O>,
+        I extends Request, O extends Response> {
+
+    @Nullable
+    private final CircuitBreakerStrategy strategy;
+
+    @Nullable
+    private final CircuitBreakerStrategyWithContent<O> strategyWithContent;
+
+    private CircuitBreakerMapping mapping = CircuitBreakerMapping.ofDefault();
+
+    /**
+     * Creates a new builder with the specified {@link CircuitBreakerStrategy}.
+     */
+    CircuitBreakerClientBuilder(CircuitBreakerStrategy strategy) {
+        this(requireNonNull(strategy, "strategy"), null);
+    }
+
+    /**
+     * Creates a new builder with the specified {@link CircuitBreakerStrategyWithContent}.
+     */
+    CircuitBreakerClientBuilder(CircuitBreakerStrategyWithContent<O> strategyWithContent) {
+        this(null, requireNonNull(strategyWithContent, "strategyWithContent"));
+    }
+
+    private CircuitBreakerClientBuilder(@Nullable CircuitBreakerStrategy strategy,
+                                        @Nullable CircuitBreakerStrategyWithContent<O> strategyWithContent) {
+        this.strategy = strategy;
+        this.strategyWithContent = strategyWithContent;
+    }
+
+    @SuppressWarnings("unchecked")
+    final T self() {
+        return (T) this;
+    }
+
+    CircuitBreakerStrategy strategy() {
+        checkState(strategy != null, "strategy is not set.");
+        return strategy;
+    }
+
+    CircuitBreakerStrategyWithContent<O> strategyWithContent() {
+        checkState(strategyWithContent != null, "strategyWithContent is not set.");
+        return strategyWithContent;
+    }
+
+    /**
+     * Sets the {@link CircuitBreakerMapping}. If unspecified, {@link CircuitBreakerMapping#ofDefault()}
+     * will be used.
+     *
+     * @return {@link T} to support method chaining.
+     */
+    public T circuitBreakerMapping(CircuitBreakerMapping mapping) {
+        this.mapping = requireNonNull(mapping, "mapping");
+        return self();
+    }
+
+    CircuitBreakerMapping circuitBreakerMapping() {
+        return mapping;
+    }
+
+    /**
+     * Returns a newly-created {@link CircuitBreakerClient} based on the properties of this builder.
+     */
+    abstract U build(Client<I, O> delegate);
+
+    /**
+     * Returns a newly-created decorator that decorates a {@link Client} with a new {@link CircuitBreakerClient}
+     * based on the properties of this builder.
+     */
+    abstract Function<Client<I, O>, U> newDecorator();
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this).omitNullValues()
+                          .add("strategy", strategy)
+                          .add("strategyWithContent", strategyWithContent)
+                          .add("mapping", mapping)
+                          .toString();
+    }
+}
--- a/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerMapping.java
+++ b/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerMapping.java
@@ -16,7 +16,12 @@
 
 package com.linecorp.armeria.client.circuitbreaker;
 
+import static java.util.Objects.requireNonNull;
+
+import java.util.function.Function;
+
 import com.linecorp.armeria.client.ClientRequestContext;
+import com.linecorp.armeria.client.circuitbreaker.KeyedCircuitBreakerMapping.KeySelector;
 import com.linecorp.armeria.common.Request;
 
 /**
@@ -24,6 +29,43 @@ import com.linecorp.armeria.common.Request;
  */
 @FunctionalInterface
 public interface CircuitBreakerMapping {
+
+    /**
+     * Returns the default {@link CircuitBreakerMapping}.
+     */
+    static CircuitBreakerMapping ofDefault() {
+        return KeyedCircuitBreakerMapping.defaultMapping;
+    }
+
+    /**
+     * Creates a new {@link CircuitBreakerMapping} which maps {@link CircuitBreaker}s with method name.
+     *
+     * @param factory A function that takes a method name and creates a new {@link CircuitBreaker}
+     */
+    static CircuitBreakerMapping perMethod(Function<String, CircuitBreaker> factory) {
+        return new KeyedCircuitBreakerMapping<>(KeySelector.METHOD, requireNonNull(factory, "factory"));
+    }
+
+    /**
+     * Creates a new {@link CircuitBreakerMapping} which maps {@link CircuitBreaker}s with the remote host name.
+     *
+     * @param factory A function that takes a host name and creates a new {@link CircuitBreaker}
+     */
+    static CircuitBreakerMapping perHost(Function<String, CircuitBreaker> factory) {
+        return new KeyedCircuitBreakerMapping<>(KeySelector.HOST, requireNonNull(factory, "factory"));
+    }
+
+    /**
+     * Creates a new {@link CircuitBreakerMapping} which maps {@link CircuitBreaker}s with the remote host and
+     * method name.
+     *
+     * @param factory A function that takes the remote host and method name and
+     *                creates a new {@link CircuitBreaker}
+     */
+    static CircuitBreakerMapping perHostAndMethod(Function<String, CircuitBreaker> factory) {
+        return new KeyedCircuitBreakerMapping<>(KeySelector.HOST_AND_METHOD,
+                                                requireNonNull(factory, "factory"));
+    }
 
     /**
      * Returns the {@link CircuitBreaker} mapped to the given parameters.
--- a/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
+++ b/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
@@ -19,26 +19,28 @@ package com.linecorp.armeria.client.circuitbreaker;
 import java.util.concurrent.CompletionStage;
 import java.util.function.BiFunction;
 
-import com.linecorp.armeria.common.HttpResponse;
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.client.ClientRequestContext;
+import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.HttpStatusClass;
 import com.linecorp.armeria.common.Response;
 
 /**
  * Determines whether a {@link Response} should be reported as a success or a failure to a
- * {@link CircuitBreaker}.
- *
- * @param <T> the response type
+ * {@link CircuitBreaker}. If you need to determine whether the request was successful by looking into the
+ * {@link Response} content, use {@link CircuitBreakerStrategyWithContent}.
  */
 @FunctionalInterface
-public interface CircuitBreakerStrategy<T extends Response> {
+public interface CircuitBreakerStrategy {
 
     /**
      * Returns the {@link CircuitBreakerStrategy} that determines a {@link Response} as successful
      * when its {@link HttpStatus} is not {@link HttpStatusClass#SERVER_ERROR} and there was no
      * {@link Exception} raised.
      */
-    static CircuitBreakerStrategy<HttpResponse> onServerErrorStatus() {
+    static CircuitBreakerStrategy onServerErrorStatus() {
         return onStatus(
                 (status, thrown) -> status != null && status.codeClass() != HttpStatusClass.SERVER_ERROR);
     }
@@ -50,21 +52,39 @@ public interface CircuitBreakerStrategy<T extends Response> {
      * @param function the {@link BiFunction} that returns {@code true}, {@code false} or
      *                 {@code null} according to the {@link HttpStatus} and {@link Throwable}. If {@code true}
      *                 is returned, {@link CircuitBreaker#onSuccess()} is called so that the
-     *                 {@link CircuitBreaker} increases its success count and use it to make a decision to
-     *                 close or open the switch. If {@code false} is returned, it works the other way around.
+     *                 {@link CircuitBreaker} increases its success count and uses it to make a decision to
+     *                 close or open the circuit. If {@code false} is returned, it works the other way around.
      *                 If {@code null} is returned, the {@link CircuitBreaker} ignores it.
      */
-    static CircuitBreakerStrategy<HttpResponse> onStatus(
-            BiFunction<HttpStatus, Throwable, Boolean> function) {
+    static CircuitBreakerStrategy onStatus(BiFunction<HttpStatus, Throwable, Boolean> function) {
         return new HttpStatusBasedCircuitBreakerStrategy(function);
     }
 
     /**
      * Returns a {@link CompletionStage} that contains {@code true}, {@code false} or
-     * {@code null} according to the specified {@link Response}. If {@code true} is returned,
+     * {@code null} which indicates a {@link Response} is successful or not. If {@code true} is returned,
      * {@link CircuitBreaker#onSuccess()} is called so that the {@link CircuitBreaker} increases its success
-     * count and use it to make a decision to close or open the switch. If {@code false} is returned, it works
+     * count and uses it to make a decision to close or open the circuit. If {@code false} is returned, it works
      * the other way around. If {@code null} is returned, the {@link CircuitBreaker} ignores it.
+     * To retrieve the response {@link HttpHeaders}, you can use the specified {@link ClientRequestContext}:
+     *
+     * <pre>{@code
+     * CompletionStage<Backoff> shouldReportAsSuccess(ClientRequestContext ctx, @Nullable Throwable cause) {
+     *     if (cause != null) {
+     *         return CompletableFuture.completedFuture(false);
+     *     }
+     *
+     *     HttpHeaders responseHeaders = ctx.log().responseHeaders();
+     *     if (responseHeaders.status().codeClass() == HttpStatusClass.SERVER_ERROR) {
+     *         return CompletableFuture.completedFuture(false);
+     *     }
+     *     ...
+     * }
+     * }</pre>
+     *
+     * @param ctx the {@link ClientRequestContext} of this request
+     * @param cause the {@link Throwable} which is raised while sending a request. {@code null} if there's no
+     *              exception.
      */
-    CompletionStage<Boolean> shouldReportAsSuccess(T response);
+    CompletionStage<Boolean> shouldReportAsSuccess(ClientRequestContext ctx, @Nullable Throwable cause);
 }
--- a/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategyWithContent.java
+++ b/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategyWithContent.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.client.circuitbreaker;
+
+import java.util.concurrent.CompletionStage;
+
+import com.linecorp.armeria.client.ClientRequestContext;
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.Response;
+
+/**
+ * Determines whether a {@link Response} should be reported as a success or a failure to a
+ * {@link CircuitBreaker} using the content of a {@link Response}. If you just need the {@link HttpHeaders}
+ * to make a decision, use {@link CircuitBreakerStrategy} for efficiency.
+ *
+ * @param <T> the response type
+ */
+@FunctionalInterface
+public interface CircuitBreakerStrategyWithContent<T extends Response> {
+
+    /**
+     * Returns a {@link CompletionStage} that contains {@code true}, {@code false} or
+     * {@code null} according to the specified {@link Response}.
+     * If {@code true} is returned, {@link CircuitBreaker#onSuccess()} is called so that the
+     * {@link CircuitBreaker} increases its success count and uses it to make a decision
+     * to close or open the circuit. If {@code false} is returned, it works the other way around.
+     * If {@code null} is returned, the {@link CircuitBreaker} ignores it.
+     *
+     * @param ctx the {@link ClientRequestContext} of this request
+     * @param response the {@link Response} from the server
+     */
+    CompletionStage<Boolean> shouldReportAsSuccess(ClientRequestContext ctx, T response);
+}
--- a/core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java
@@ -16,6 +16,7 @@
 package com.linecorp.armeria.client.retry;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.linecorp.armeria.client.retry.DefaultBackoffHolder.defaultBackoff;
 import static com.linecorp.armeria.client.retry.FixedBackoff.NO_DELAY;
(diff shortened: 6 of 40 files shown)


benchmarks/src/jmh/java/com/linecorp/armeria/core/client/retry/RetryingHttpClientBase.java
```diff
@@ -0,0 +1,65 @@
...
+        server.start().join();
+        client = client();
+    }
+
+    @TearDown
+    public void stop() {
+        server.stop().join();
+    }
+
+    protected abstract HttpClient client();
```
trustin: `newClient()`?

benchmarks/src/jmh/java/com/linecorp/armeria/core/client/retry/RetryingHttpClientBase.java
```diff
@@ -0,0 +1,65 @@
...
+
+    protected String baseUrl() {
+        final ServerPort httpPort = server.activePorts().values().stream()
+                                          .filter(ServerPort::hasHttp).findAny()
+                                          .get();
+        return "http://localhost:" + httpPort.localAddress().getPort();
```
trustin: `h2c://127.0.0.1:`? (I'd prefer using an IP address instead of `localhost` just in case `localhost` is configured poorly in `/etc/hosts`.)

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -65,6 +70,27 @@
      * {@link CircuitBreaker#onSuccess()} is called so that the {@link CircuitBreaker} increases its success
      * count and use it to make a decision to close or open the switch. If {@code false} is returned, it works
      * the other way around. If {@code null} is returned, the {@link CircuitBreaker} ignores it.
+     *
+     * @param ctx the {@link ClientRequestContext} of this request
+     * @param response the {@link Response} from the server.
+     *                 If {{@link #needsResponseStreaming(ClientRequestContext)}} returns {@code false} and
+     *                 the type of {@link Response} is {@link HttpResponse}, this will be {@code null}.
+     *                 This will be the actual {@link Response} otherwise.
+     * @param cause the {@link Throwable} which is raised while sending a request
```
trustin: Could you mention when `cause` can be `null` although it's easily guessed.

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -44,13 +46,18 @@
...
-        final HttpHeaderSubscriber subscriber = new HttpHeaderSubscriber(future);
-        response.completionFuture().whenComplete(subscriber);
-        response.subscribe(subscriber);
-
-        return future.handle((headers, thrown) -> backoffFunction
-                .apply(headers != null ? headers.status() : null, thrown));
+    public CompletionStage<Backoff> shouldRetry(ClientRequestContext ctx, @Nullable HttpResponse response,
+                                                @Nullable Throwable cause) {
+        if (cause instanceof UnprocessedRequestException) {
+            // We can just retry immediately when the UNprocessedRequestException is raised.
```
trustin: `UN` -> `Un`

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
```diff
@@ -44,13 +46,18 @@
...
+            return CompletableFuture.completedFuture(Backoff.withoutDelay());
```
trustin: There's a chance of thrashing a server in this case. A server may be in graceful-shutdown phase, rejecting any incoming requests with GOAWAY, which will make the server suffer from a high spike of retries. How about consulting `backoffFunction`?

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -113,10 +114,31 @@
      * If the condition does not match, this will return {@code null} to stop retry attempt.
      * Note that {@link ResponseTimeoutException} is not retriable for the whole retry, but each attempt.
      *
+     * @param ctx the {@link ClientRequestContext} of this request
+     * @param response the {@link Response} from the server.
+     *                 If {{@link #needsResponseStreaming(ClientRequestContext)}} returns {@code false} and
+     *                 the type of {@link Response} is {@link HttpResponse}, this will be {@code null}.
```
trustin: How about:

```
{@code null} if {@link #needsResponseStreaming(...)} ....
The actual {@link Response} otherwise.
```

Please also note the double curly braces in `{{@link #needsResponseStreaming...}}`

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -65,6 +70,27 @@
      * {@link CircuitBreaker#onSuccess()} is called so that the {@link CircuitBreaker} increases its success
      * count and use it to make a decision to close or open the switch. If {@code false} is returned, it works
      * the other way around. If {@code null} is returned, the {@link CircuitBreaker} ignores it.
+     *
+     * @param ctx the {@link ClientRequestContext} of this request
+     * @param response the {@link Response} from the server.
+     *                 If {{@link #needsResponseStreaming(ClientRequestContext)}} returns {@code false} and
+     *                 the type of {@link Response} is {@link HttpResponse}, this will be {@code null}.
+     *                 This will be the actual {@link Response} otherwise.
```
trustin: How about:

```
{@code null} if {@link #needsResponseStreaming(...)} ....
The actual {@link Response} otherwise.
```

Please also note the double curly braces in `{{@link #needsResponseStreaming...}}`

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -113,10 +114,31 @@
...
      *
      * @see <a href="https://line.github.io/armeria/advanced-retry.html#per-attempt-timeout">Per-attempt
      *      timeout</a>
      */
-    CompletionStage<Backoff> shouldRetry(I request, O response);
+    CompletionStage<Backoff> shouldRetry(ClientRequestContext ctx, @Nullable T response,
+                                         @Nullable Throwable cause);
+
+    /**
+     * Returns {@code true} if you need the content in an {@link HttpResponse} in order to use in
```
trustin: content in -> content of (or just `{@link HttpResponse}` content`?)

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -113,10 +114,31 @@
...
+     * {@link #shouldRetry(ClientRequestContext, Response, Throwable)}. That will invoke
+     * {@link #shouldRetry(ClientRequestContext, Response, Throwable)} with the actual {@link Response}
+     * so that you can use it. If the {@link Response} is the instance of an {@link RpcResponse},
```
trustin: use it -> subscribe to it and retrieve its content?

anuraaga:
Happy to see performance PRs landing :D

site/src/sphinx/client-circuit-breaker.rst
```diff
@@ -111,40 +111,81 @@ is raised or the status is ``5xx``, succeeds when the status is ``2xx`` and igno
...
-        public CompletionStage<Boolean> shouldReportAsSuccess(HttpResponse response) {
-            return response.aggregate().handle((res, cause) -> {
-                if (cause != null) { // A failure if an Exception is raised.
-                    return false;
+        public CompletionStage<Boolean> shouldReportAsSuccess(ClientRequestContext ctx,
+                                                              @Nullable HttpResponse response,
+                                                              @Nullable Throwable cause) {
+            if (cause != null && !(cause instanceof UnprocessedRequestException)) {
+                // A failure if an Exception is raised.
+                return CompletableFuture.completedFuture(false);
```
trustin: How about adding `return CompletableFuture.completedFuture(null)` for `UnprocessedRequestException` around here? We could also add some comments about why `null` has to be returned on `UnprocessedRequestException`.

benchmarks/src/jmh/java/com/linecorp/armeria/core/client/retry/RetryingHttpClientBase.java
minwoox: Just copied the codes without much thinking. Thank you~!

core/src/main/java/com/linecorp/armeria/client/retry/HttpStatusBasedRetryStrategy.java
minwoox: Oops my big mistake. Fixed. Thank you~!

site/src/sphinx/client-circuit-breaker.rst
```diff
@@ -111,40 +111,81 @@ is raised or the status is ``5xx``, succeeds when the status is ``2xx`` and igno
...
+            // Neither a success nor a failure. Do not take this response into account.
+            return CompletableFuture.completedFuture(null);
         }
     };
 
 If you want to treat a :api:`Response` as a success, return ``true``. Return ``false`` to treat as a failure.
 Note that :api:`CircuitBreakerStrategy` can return ``null`` as well. It won't be counted as a success nor
 a failure.
 
+You can use the actual ``HttpResponse`` to make a decision overriding
```
trustin: How about:

If you need to determine if the request was successful by looking into the response content, you need to override ...

site/src/sphinx/client-circuit-breaker.rst
```diff
@@ -111,40 +111,81 @@ is raised or the status is ``5xx``, succeeds when the status is ``2xx`` and igno
...
+        }
+
+        @Override
+        public CompletionStage<Boolean> shouldReportAsSuccess(ClientRequestContext ctx,
+                                                              @Nullable HttpResponse response,
+                                                              @Nullable Throwable cause) {
+            return response.aggregate().handle((res, thrown) -> {
+                if (thrown != null && !(thrown instanceof UnprocessedRequestException)) {
+                    // A failure if an Exception is raised.
+                    return false;
```
trustin: Ditto - return null explicitly on `UnprocessedRequestException` and explain why briefly.

site/src/sphinx/client-retry.rst
```diff
@@ -98,30 +104,63 @@ This will retry when the response's status is ``409 Conflict`` or :api:`Response
...
+                return CompletableFuture.completedFuture(backoffOnServerErrorOrTimeout);
+            } else if (status == HttpStatus.CONFLICT) {
+                return CompletableFuture.completedFuture(backoffOnConflict);
+            }
+
+            return CompletableFuture.completedFuture(null); // Return null to stop retrying.
+        }
+    };
+
+If you want to use the actual ``HttpResponse`` to make a decision, you need to override
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -65,6 +70,45 @@
      * {@link CircuitBreaker#onSuccess()} is called so that the {@link CircuitBreaker} increases its success
      * count and use it to make a decision to close or open the switch. If {@code false} is returned, it works
      * the other way around. If {@code null} is returned, the {@link CircuitBreaker} ignores it.
+     * To retrieve the response {@link HttpHeaders} without {@link Response}, you can use the specified
+     * {@link ClientRequestContext}:
+     *
+     * <pre>{@code
+     * > CompletionStage<Backoff> shouldReportAsSuccess(ClientRequestContext ctx, @Nullable T response,
+     * >                                                 @Nullable Throwable cause) {
```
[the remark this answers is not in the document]
minwoox: Fixed :)

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -112,11 +114,49 @@
...
      *      timeout</a>
      */
-    CompletionStage<Backoff> shouldRetry(I request, O response);
+    CompletionStage<Backoff> shouldRetry(ClientRequestContext ctx, @Nullable T response,
+                                         @Nullable Throwable cause);
+
+    /**
+     * Returns {@code true} if you need to determine whether you need to retry by looking into the
+     * {@link HttpResponse} content in {@link #shouldRetry(ClientRequestContext, Response, Throwable)}.
+     * That will invoke {@link #shouldRetry(ClientRequestContext, Response, Throwable)} with the actual
```
trustin: `If this method returns {@code true}, {@link #shouldRetry...} will be invoked with ...`

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -112,11 +114,49 @@
...
+     * {@link Response} so that you can subscribe to it and retrieve its content. If the {@link Response} is
+     * the instance of an {@link RpcResponse}, {@link #shouldRetry(ClientRequestContext, Response, Throwable)}
```
trustin: the instance of an -> an

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -112,11 +114,49 @@
...
+     * is called with the actual {@link Response} regardless of the return value.
```
trustin: the return value -> what this method returns

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -112,11 +114,49 @@
...
+     * Please note that if you just need the {@link HttpHeaders} to make a decision, return {@code false},
+     * because it will increase the performance.
```
trustin: it will increase the performance -> content streaming reduces performance ?

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -65,6 +70,45 @@
...
+     *              exception.
      */
-    CompletionStage<Boolean> shouldReportAsSuccess(T response);
+    CompletionStage<Boolean> shouldReportAsSuccess(ClientRequestContext ctx, @Nullable T response,
+                                                   @Nullable Throwable cause);
+
+    /**
+     * Returns {@code true} if you need to determine whether the request was successful by looking into the
+     * {@link HttpResponse} content in
+     * {@link #shouldReportAsSuccess(ClientRequestContext, Response, Throwable)}. That will invoke
```
trustin: `If this method returns {@code true}, {@link #shouldReportAsSuccess...} will be invoked with ...`

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -65,6 +70,45 @@
...
+     * {@link #shouldReportAsSuccess(ClientRequestContext, Response, Throwable)} with the actual
+     * {@link Response} so that you can subscribe to it and retrieve its content. If the {@link Response} is
+     * the instance of an {@link RpcResponse},
```
trustin: the instance of an -> an

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -65,6 +70,45 @@
...
+     * {@link #shouldReportAsSuccess(ClientRequestContext, Response, Throwable)} is called
+     * with the actual {@link Response} regardless of the return value. Please note that if you just need the
+     * {@link HttpHeaders} to make a decision, return {@code false}, because it will increase the performance.
```
trustin: it will increase the performance -> content streaming reduces performance ?

minwoox:
Will update the documentation soon.

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClientBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Creates a new builder with the specified {@link CircuitBreakerStrategyWithContent}.
+     */
+    CircuitBreakerClientBuilder(CircuitBreakerStrategyWithContent<O> strategyWithContent) {
+        this(null, requireNonNull(strategyWithContent, "strategyWithContent"));
+    }
+
+    private CircuitBreakerClientBuilder(@Nullable CircuitBreakerStrategy strategy,
```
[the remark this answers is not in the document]
minwoox: I just wanted to make a user specify either `RetryStrategy` or `RetryStrategyWithContent` when they create a builder so that only one of those strategies can be set.
I can do the same thing using setters and check the state inside the setters but I didn't want it because I thought it can confuse the user.
If you mind this constructor, I just can remove this private constructor and assign the variables in other constructors. What do you think?

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerMapping.java
```diff
@@ -25,6 +30,40 @@
 @FunctionalInterface
 public interface CircuitBreakerMapping {
 
+    /**
+     * The default {@link CircuitBreakerMapping} implementation.
+     */
+    CircuitBreakerMapping defaultMapping = CircuitBreakerMapping.perHost(CircuitBreaker::of);
```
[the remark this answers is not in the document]
trustin: How about providing as a method? `ofDefault()` like in `MeterIdPrefixFunction`. We are going to have to keep the default instance somewhere else though.

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -19,26 +19,28 @@
...
 import com.linecorp.armeria.common.HttpStatusClass;
 import com.linecorp.armeria.common.Response;
 
 /**
  * Determines whether a {@link Response} should be reported as a success or a failure to a
- * {@link CircuitBreaker}.
- *
- * @param <T> the response type
+ * {@link CircuitBreaker}. If you need to determine whether the request was successful by looking into the
+ * {@link Response}, use {@link CircuitBreakerStrategyWithContent}.
```
trustin: `{@link Response} content`?

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategy.java
```diff
@@ -50,21 +52,39 @@
...
         return new HttpStatusBasedCircuitBreakerStrategy(function);
     }
 
     /**
      * Returns a {@link CompletionStage} that contains {@code true}, {@code false} or
-     * {@code null} according to the specified {@link Response}. If {@code true} is returned,
+     * {@code null} which indicates a {@link Response} is successful or not. If {@code true} is returned,
      * {@link CircuitBreaker#onSuccess()} is called so that the {@link CircuitBreaker} increases its success
-     * count and use it to make a decision to close or open the switch. If {@code false} is returned, it works
+     * count and uses it to make a decision to close or open the switch. If {@code false} is returned, it works
```
trustin: switch -> circuit?

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategyWithContent.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.concurrent.CompletionStage;
+
+import com.linecorp.armeria.client.ClientRequestContext;
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.Response;
+
+/**
+ * Determines whether a {@link Response} should be reported as a success or a failure to a
+ * {@link CircuitBreaker} using the content of a {@link Response}. If you just need the {@link HttpHeaders}
+ * to make a decision, use {@link CircuitBreakerStrategy}.
```
trustin: .. for efficiency.

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerStrategyWithContent.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@FunctionalInterface
+public interface CircuitBreakerStrategyWithContent<T extends Response> {
+
+    /**
+     * Returns a {@link CompletionStage} that contains {@code true}, {@code false} or
+     * {@code null} according to the specified {@link Response}.
+     * If {@code true} is returned, {@link CircuitBreaker#onSuccess()} is called so that the
+     * {@link CircuitBreaker} increases its success count and uses it to make a decision
+     * to close or open the switch. If {@code false} is returned, it works the other way around.
```
trustin: switch -> circuit

core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java
```diff
@@ -24,11 +24,19 @@
...
 /**
  * Controls back off between attempts in a single retry operation.
  */
 @FunctionalInterface
 public interface Backoff {
+
+    /**
+     * The default {@link Backoff} implementation.
+     */
+    Backoff defaultBackoff = Backoff.of(Flags.defaultBackoffSpec());
```
trustin: Ditto - `ofDefault()`

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -21,48 +21,53 @@
...
+public interface RetryStrategy {
 
-    Backoff defaultBackoff = Backoff.of(Flags.defaultBackoffSpec());
+    /**
+     * The default {@link Backoff} implementation.
+     *
+     * @deprecated use {@link Backoff#defaultBackoff}
+     */
+    @Deprecated
+    Backoff defaultBackoff = Backoff.defaultBackoff;
```
trustin: :+1:

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategyWithContent.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param <T> the response type
+ */
+@FunctionalInterface
+public interface RetryStrategyWithContent<T extends Response> {
+
+    /**
+     * Returns a {@link CompletionStage} that contains {@link Backoff} which will be used for retry.
+     * If the condition does not match, this will return {@code null} to stop retry attempt.
+     * Note that {@link ResponseTimeoutException} is not retriable for the whole retry, but each attempt.
```
trustin: but -> but only for (for clarity?)

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -85,28 +111,33 @@ public O execute(ClientRequestContext ctx, I req) throws Exception {
...
         return retryStrategy;
     }
 
+    /**
+     * Returns the {@link RetryStrategyWithContent}.
+     *
+     * @throws IllegalStateException if the {@link RetryStrategyWithContent} is not set
+     */
+    protected RetryStrategyWithContent<O> retryStrategyWithContent() {
+        checkState(retryStrategyWithContent != null, "retryStrategyWithContent is not set");
```
trustin: Missing period, because it's a sentence.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -85,28 +111,33 @@ public O execute(ClientRequestContext ctx, I req) throws Exception {
...
     }
 
-    protected RetryStrategy<I, O> retryStrategy() {
+    /**
+     * Returns the {@link RetryStrategy}.
+     *
+     * @throws IllegalStateException if the {@link RetryStrategy} is not set
+     */
+    protected RetryStrategy retryStrategy() {
+        checkState(retryStrategy != null, "retryStrategy is not set");
```
trustin: Missing period, because it's a sentence.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -59,16 +73,21 @@ public RetryingHttpClientBuilder useRetryAfter(boolean useRetryAfter) {
...
+     *
      * @return {@link RetryingHttpClientBuilder} to support method chaining
+     *
+     * @throws IllegalStateException if this builder is created with a {@link RetryStrategy}
+     * @throws IllegalArgumentException if the specified {@code contentPreviewLength} is 0
      */
     public RetryingHttpClientBuilder contentPreviewLength(int contentPreviewLength) {
-        checkArgument(contentPreviewLength >= 0,
-                      "contentPreviewLength: %s (expected: >= 0)", contentPreviewLength);
+        checkState(needsContentInStrategy, "cannot set contentPreviewLength when RetryStrategy is used.");
```
trustin: How about: `.. is used; Use RetryStrategyWithContent to enable this feature.`

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -59,16 +73,21 @@ public RetryingHttpClientBuilder useRetryAfter(boolean useRetryAfter) {
...
      * this and there's no retry condition matched, it will hand over the stream to the client.
      * @param contentPreviewLength the content length to preview. Unlike other length related arguments,
-     *                             {@code 0} does not disable the length limit. It means that it will not
-     *                             look up the content
+     *                             {@code 0} does not disable the length limit.
+     *
      * @return {@link RetryingHttpClientBuilder} to support method chaining
+     *
+     * @throws IllegalStateException if this builder is created with a {@link RetryStrategy}
+     * @throws IllegalArgumentException if the specified {@code contentPreviewLength} is 0
```
trustin: - 0 is a literal and thus has to be `{@code 0}`.
- `is` -> `is equal to or less than`

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -59,16 +73,21 @@ public RetryingHttpClientBuilder useRetryAfter(boolean useRetryAfter) {
     }
 
     /**
-     * Sets the length of content to look up whether retry or not. If the total length of content exceeds
+     * Sets the length of content to look up whether retry or not when you create this builder with a
```
trustin: How about:

```
... content required to determine whether to retry or not. If the total length of ...
hand over the stream to the client. Note that this property is useful only if you specified
{@link RetryStrategyWithContent} when calling this builder's constructor. The default value
of this property is {@literal #...}.

@param ...
```

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -59,16 +73,21 @@ public RetryingHttpClientBuilder useRetryAfter(boolean useRetryAfter) {
...
+     * {@link RetryStrategyWithContent}. If the total length of content exceeds
      * this and there's no retry condition matched, it will hand over the stream to the client.
      * @param contentPreviewLength the content length to preview. Unlike other length related arguments,
-     *                             {@code 0} does not disable the length limit. It means that it will not
-     *                             look up the content
+     *                             {@code 0} does not disable the length limit.
+     *
      * @return {@link RetryingHttpClientBuilder} to support method chaining
+     *
+     * @throws IllegalStateException if this builder is created with a {@link RetryStrategy}
```
trustin: `... rather than {@link RetryStrategyWithContent}` ?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -59,16 +73,21 @@ public RetryingHttpClientBuilder useRetryAfter(boolean useRetryAfter) {
...
 
     /**
-     * Sets the length of content to look up whether retry or not. If the total length of content exceeds
+     * Sets the length of content to look up whether retry or not when you create this builder with a
+     * {@link RetryStrategyWithContent}. If the total length of content exceeds
      * this and there's no retry condition matched, it will hand over the stream to the client.
      * @param contentPreviewLength the content length to preview. Unlike other length related arguments,
-     *                             {@code 0} does not disable the length limit. It means that it will not
-     *                             look up the content
+     *                             {@code 0} does not disable the length limit.
```
trustin: `wnlike other ...` could be removed.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -36,11 +39,22 @@
 
     private int contentPreviewLength = DEFAULT_CONTENT_PREVIEW_LENGTH;
```
trustin: `DEFAULT_CONTENT_PREVIEW_LENGTH` is currently `0`, which is not the valid `contentPreviewLength`. Now that we have a separate strategy interface that has no content streaming, how about using `Integer.MAX_VALUE` as the default?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
```diff
@@ -73,19 +73,23 @@ public RetryingHttpClientBuilder useRetryAfter(boolean useRetryAfter) {
...
     /**
-     * Sets the length of content to look up whether retry or not when you create this builder with a
-     * {@link RetryStrategyWithContent}. If the total length of content exceeds
-     * this and there's no retry condition matched, it will hand over the stream to the client.
-     * @param contentPreviewLength the content length to preview. Unlike other length related arguments,
-     *                             {@code 0} does not disable the length limit.
+     * Sets the length of content required to determine whether to retry or not. If the total length of content
+     * exceeds this length and there's no retry condition matched, it will hand over the stream to the client.
+     * Note that this property is useful only if you specified {@link RetryStrategyWithContent} when calling
+     * this builder's constructor. The default value of this property is {@link Integer#MAX_VALUE}.
```
trustin: `{@literal #DEFAULT_CONTENT_PREVIEW_LENGTH}`?

site/src/sphinx/client-retry.rst
```diff
@@ -98,34 +104,72 @@ This will retry when the response's status is ``409 Conflict`` or :api:`Response
...
+                return null; // Return null to stop retrying.
             });
         }
     };
 
+    final HttpClient client = new HttpClientBuilder(...)
+            .decorator(new RetryingHttpClientBuilder(strategy).newDecorator()) // Specify the strategy.
+            .build();
+
+    client.execute(...).aggregate().join();
```
trustin: `final AggregatedHttpMessage res = client.execute(...)...;`

site/src/sphinx/client-circuit-breaker.rst
```diff
@@ -111,40 +110,91 @@ is raised or the status is ``5xx``, succeeds when the status is ``2xx`` and igno
...
+                        return null;
+                    });
+                }
+            };
+
+    final HttpClient client = new HttpClientBuilder(...)
+            .decorator(new CircuitBreakerHttpClientBuilder(myStrategy).newDecorator()) // Specify the strategy
+            .build();
+
+    client.execute(...).aggregate().join();
```
trustin: `final AggregatedHttpResponse res = client.execute(...)...;`

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClientBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+        T extends CircuitBreakerClientBuilder<T, U, I, O>, U extends CircuitBreakerClient<I, O>,
+        I extends Request, O extends Response> {
+
+    @Nullable
+    final CircuitBreakerStrategy strategy;
+
+    @Nullable
+    final CircuitBreakerStrategyWithContent<O> strategyWithContent;
+
+    CircuitBreakerMapping mapping = CircuitBreakerMapping.ofDefault();
```
trustin: Could you:

- make these 3 fields private and
- add proper package-local getters?

IIRC our inspection profile should warn about this. Let me know it did not warn you about this, so I double-check the inspection profile.

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -21,48 +21,53 @@
...
  */
 @FunctionalInterface
-public interface RetryStrategy<I extends Request, O extends Response> {
+public interface RetryStrategy {
 
-    Backoff defaultBackoff = Backoff.of(Flags.defaultBackoffSpec());
+    /**
+     * The default {@link Backoff} implementation.
+     *
+     * @deprecated use {@link Backoff#ofDefault()}
```
trustin: - use -> Use
- missing period

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClientBuilder.java
minwoox: Change to `{@value ...}` Thank you~!

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClientBuilder.java
minwoox: Fixed Thank you~!

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
minwoox: Fixed Thank you~!

site/src/sphinx/client-circuit-breaker.rst
minwoox: Fixed Thank you~!

core/src/main/java/com/linecorp/armeria/common/HttpHeaderNames.java
```diff
@@ -254,7 +254,7 @@
     /**
      * {@code "keep-alive"}.
      *
-     * @deprecated Use {@link #CONNECTION} instead.
+     * @deprecated Use {@link #CONNECTION}.
```
trustin: :+1:

minwoox:
```
Benchmark                 Mode  Cnt      Score     Error  Units
WithDuplicator.empty     thrpt  100  87399.115 ± 101.651  ops/s
WithoutDuplicator.empty  thrpt  100  91787.170 ±  73.129  ops/s
```

trustin:
Thanks, @minwoox!

PR #1423, opened 2018-11-02
Add logging of when the first bytes are written / read from a transpo…

### LINKED ISSUE #1421: Add log availability for writing first byte and reading first byte from channel
I am trying to debug a lot of untraced time in spans (e.g., time between client send <-> server receive) and it would be very useful to have wire send (ws) and wire receive (wr) annotations which I believe are semi-standard annotations added by other zipkin instrumentation. For armeria, I think it would involve adding FIRST_BYTE_SENT and FIRST_BYTE_RECEIVED request log availability, propagate `RequestContext` into the `Http*Encoder` classes and call the log methods just after writing to the channel, or receiving from it. If there is anywhere more appropriate than the encoders (ideally it should be done in the very last actual netty channel handler but not sure if we can propagate request context that far) let me know.

With such callbacks, it'd be simple to also update the zipkin library to add annotations during those events. If the general idea seems fine, would be glad to implement.

/cc @adriancole

anuraaga:
…rt channel.

When doing detailed latency debugging, for example using traces, it can be useful to record the precise moment bytes were transfered on a channel. This helps differentiate time spent in the framework and time spent in network infrastructure.

@trustin Can you check the general approach? If the approach looks ok will add tests and zipkin integration.

Fixes #1421


core/src/main/java/com/linecorp/armeria/client/HttpClientFirstTransferLogger.java
```diff
@@ -0,0 +1,51 @@
...
+            reqCtx.logBuilder().responseHeadersFirstBytesTransferred();
+            loggedResponseHeadersFirstBytesTransferred = true;
+        }
+        super.read(ctx);
+    }
+
+    @Override
+    public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+        if (!loggedRequestHeadersFirstBytesTransferred) {
+            reqCtx.logBuilder().requestHeadersFirstBytesTransferred();
```
trustin: At this point no data was sent to the wire. It is merely added to the outbound queue of Netty. Although not perfect, it'd better adding a `ChannelFutureListener` to the write future of the first response headers.

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -207,6 +207,8 @@ void invoke0(Channel channel, ClientRequestContext ctx,
                 return;
             }
 
+            channel.pipeline().addFirst(new HttpClientFirstTransferLogger(ctx));
```
trustin: I think this does not work as intended because:

- HTTP/2 is multiplex protocol. There's no way to tell the byte you just received is yours.
- We will have as many `*FirstTransferLogger`s as the number of pending requests, i.e. very deep call stack and poor performance.

core/src/main/java/com/linecorp/armeria/client/HttpClientFirstTransferLogger.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private boolean loggedResponseHeadersFirstBytesTransferred;
+    private boolean loggedRequestHeadersFirstBytesTransferred;
+
+    HttpClientFirstTransferLogger(ClientRequestContext reqCtx) {
+        this.reqCtx = reqCtx;
+    }
+
+    @Override
+    public void read(ChannelHandlerContext ctx) throws Exception {
```
trustin: I guess you meant to override `channelRead()`?

anuraaga:
@trustin Thanks for the tips! I indeed confused myself a lot when trying to get as precise as possible, I guess logging after a full frame is sort of inevitable since we at least need the stream ID. Can you take another skim through and if it looks ok will add more?

anuraaga:
Thanks for the initial look! Sorry for not getting back to this PR yet but will get to it soon

anuraaga:
Thanks - this should be ready for review.

/cc @adriancole

anuraaga:
image

adriancole:
this is interesting in so far as the annotations are server-side and
enclosed by the server events. Where I've seen this in the past is
client-side. No problem, just mentioning this.

zipkin/src/main/java/com/linecorp/armeria/server/tracing/HttpTracingService.java
```diff
@@ -84,10 +85,16 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
...
+                log -> SpanTags.logWireReceive(span, log.requestFirstBytesTransferredTimeNanos(), log),
+                RequestLogAvailability.REQUEST_FIRST_BYTES_TRANSFERRED);
 
         // Ensure the trace context propagates to children
         ctx.onChild(RequestContextCurrentTraceContext::copy);
 
+        ctx.log().addListener(
+                log -> SpanTags.logWireSend(span, log.responseFirstBytesTransferredTimeNanos(), log),
+                RequestLogAvailability.RESPONSE_FIRST_BYTES_TRANSFERRED);
         ctx.log().addListener(log -> SpanContextUtil.closeSpan(span, log), RequestLogAvailability.COMPLETE);
```
trustin: How about logging wireReceive and wireSend when closing a span to reduce the number of log listeners? Is it important to add wire annotations on time?

zipkin/src/test/java/com/linecorp/armeria/it/tracing/HttpTracingIntegrationTest.java
```diff
@@ -275,6 +275,37 @@ public void testClientInitiatedTrace() throws Exception {
...
+        long clientWireReceiveTime = clientFooSpan.annotations().stream().filter(a -> a.value().equals("wr"))
+                                               .findFirst().get().timestamp();
+        long clientEndTime = clientStartTime + clientFooSpan.durationAsLong();
+
+        long serverStartTime = serviceFooSpan.timestampAsLong();
+        long serverWireSendTime = serviceFooSpan.annotations().stream().filter(a -> a.value().equals("ws"))
+                                                .findFirst().get().timestamp();
+        long serverWireReceiveTime = serviceFooSpan.annotations().stream().filter(a -> a.value().equals("wr"))
+                                                   .findFirst().get().timestamp();
+        long serverEndTime = serverStartTime + serviceFooSpan.durationAsLong();
```
trustin: Could these variables final?

anuraaga: Ah didn't know we did that in tests too.

zipkin/src/main/java/com/linecorp/armeria/server/tracing/HttpTracingService.java
anuraaga: Good point, moved to completion listener.

trustin:
@anuraaga Compilation errors? :-)

anuraaga:
Urgh, new IntelliJ didn't finish indexing in time for me to check before pushing ;)

trustin:
Thanks a lot, @anuraaga !

PR #1427, opened 2018-11-03
Set RequestLog wall time in micros and use RequestLog timestamps for s…

anuraaga:
…pan timing.

JDK9+ supports returning wall time with microseconds precision. We may as well use it in logging instead of millis so users can take advantage of extra precision.

I also went ahead and updated the zipkin tracing to use the new values for spans. However I'm not exactly sure if this is good or bad - I think it means time spent in decorators can't be traced as the `RequestLog` is tied to the lifecycle of `StreamMessage`, not `Service`. On the flip side, maybe that is more natural behavior and it does make adding annotations like in #1423 much simpler. Appreciate thoughts on it.

/cc @adriancole

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
@@ -29,10 +29,13 @@ import static com.linecorp.armeria.common.logging.RequestLogAvailability.RESPONS
 import static com.linecorp.armeria.common.logging.RequestLogAvailability.SCHEME;
 import static java.util.Objects.requireNonNull;
 
+import java.time.Clock;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.Function;
 
@@ -88,14 +91,14 @@ public class DefaultRequestLog implements RequestLog, RequestLogBuilder {
     private volatile boolean requestContentDeferred;
     private volatile boolean responseContentDeferred;
 
-    private long requestStartTimeMillis;
+    private long requestStartTimeMicros;
     private long requestStartTimeNanos;
     private long requestEndTimeNanos;
     private long requestLength;
     @Nullable
     private Throwable requestCause;
 
-    private long responseStartTimeMillis;
+    private long responseStartTimeMicros;
     private long responseStartTimeNanos;
     private long responseEndTimeNanos;
     private long responseLength;
@@ -149,7 +152,7 @@ public class DefaultRequestLog implements RequestLog, RequestLogBuilder {
 
     private void propagateRequestSideLog(RequestLog child) {
         child.addListener(log -> {
-            startRequest0(log.requestStartTimeNanos(), log.requestStartTimeMillis(), log.channel(),
+            startRequest0(log.requestStartTimeNanos(), log.requestStartTimeMicros(), log.channel(),
                           log.sessionProtocol(), true);
         }, REQUEST_START);
         child.addListener(log -> serializationFormat(log.serializationFormat()), SCHEME);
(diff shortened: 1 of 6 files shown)


adriancole:
I might still use brave's clock because it is guaranteed to not go backwards in the same trace. We freeze it and use ticks from there for example.

anuraaga:
Yeah I think `RequestLog` timings are similar, we get wall time and nano time at the start and later measurements are nano time.

trustin:
How about undeprecating millis getters given most timing APIs in Java uses millis in many places?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -910,4 +922,10 @@ public String toStringResponseOnly(Function<HttpHeaders, HttpHeaders> headersSan
             this.interestedFlags = interestedFlags;
         }
     }
+
+    // This is millisecond precision on Java 8, but it's the best we can do.
+    private static long currentTimeMicros() {
+        Instant now = Clock.systemUTC().instant();
+        return TimeUnit.SECONDS.toMicros(now.getEpochSecond()) + TimeUnit.NANOSECONDS.toMicros(now.getNano());
```
trustin: How about using `PlatformDependent.javaVersion()` and just use `System.currentTimeMillis()` up to Java 8?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -538,16 +545,21 @@ private void startResponse0(long responseStartTimeNanos, long responseStartTimeM
         }
 
         this.responseStartTimeNanos = responseStartTimeNanos;
-        this.responseStartTimeMillis = responseStartTimeMillis;
+        this.responseStartTimeMicros = responseStartTimeMillis;
```
minwoox: Could you change the name of the parameter(`responseStartTimeMillis`) as well?

hyangtack: `this.responseStartTimeMicros = responseStartTimeMicros;`?

anuraaga: Oops thanks

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanContextUtil.java
```diff
@@ -16,19 +16,33 @@
...
      * The span cannot be used further after this method has been called.
      */
     public static void closeSpan(Span span, RequestLog log) {
         SpanTags.addTags(span, log);
-        span.finish();
+        span.finish(wallTimeMicros(log, log.responseEndTimeNanos()));
+    }
+
+    private static long wallTimeMicros(RequestLog log, long timeNanos) {
+        long relativeTimeNanos = timeNanos - log.requestStartTimeNanos();
```
minwoox: nit: is this `log.totalDurationNanos()`?

anuraaga: Currently yes since I only pass `responseEndTimeNanos()` into this function. Later I want to also pass `requestFirstBytesTransferredNanos()` and `responseFirstBytesTransferredNanos()`. Is it clearer if I use that now and replace it with a generic approach later?

trustin: Oh I see. Now I get it. Fine as it is.

minwoox: So am I :)

anuraaga:
Hmm - appveyor is having trouble with reusing the nodejs cache it seems even when retrying. I run gradle + node on Windows all the time without problem, but maybe something with how they package it up...

trustin:
Let me zap the cache tomorrow. Please ignore meanwhile..

adriancole:
@anuraaga great progress! so will there be a separate change for the "ws" "wr" annotation part on client side?

PR #1429, opened 2018-11-04
Reduce memory copies in `Http2ResponseDecoder` and gRPC

trustin:
Motivation:

We perform unnecessary memory copies while handling a gRPC request.

Modifications:

- Use `ByteBufHttpData` instead of `HttpData.of()` in
  `Http2ResponseDecoder`
- Use a queue of `ByteBuf`s instead of `CompositeByteBuf` for deframing
  in `ArmeriaMessageDeframer`, because the buffers we get from
  `CompositeByteBuf` is often copied or wrapped with indirection.
- Use `CompositeByteBuf` to prepend a frame header if the frame is
  larger than 128 bytes

Result:

Higher performance.

Before:

    Benchmark                     (wrapBuffer)   Mode  Cnt   Score   Error  Units
    LargePayloadBenchmark.normal         false  thrpt  100  35.266 ± 0.191  ops/s
    LargePayloadBenchmark.normal          true  thrpt  100  40.839 ± 0.228  ops/s

After:

    Benchmark                     (wrapBuffer)   Mode  Cnt    Score   Error  Units
    LargePayloadBenchmark.normal         false  thrpt  100   95.581 ± 0.700  ops/s
    LargePayloadBenchmark.normal          true  thrpt  100  117.076 ± 0.806  ops/s

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
+++ b/grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
@@ -53,7 +53,9 @@ import static java.util.Objects.requireNonNull;
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayDeque;
 import java.util.Objects;
+import java.util.Queue;
 
 import javax.annotation.Nullable;
 
@@ -69,7 +71,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufHolder;
 import io.netty.buffer.ByteBufInputStream;
-import io.netty.buffer.CompositeByteBuf;
+import io.netty.buffer.Unpooled;
 
 /**
  * A deframer of messages transported in the gRPC wire format. See
@@ -177,7 +179,8 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
     private boolean endOfStream;
 
     @Nullable
-    private CompositeByteBuf unprocessed;
+    private Queue<ByteBuf> unprocessed;
+    private int unprocessedBytes;
 
     private long pendingDeliveries;
     private boolean deliveryStalled = true;
@@ -191,7 +194,7 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
         this.maxMessageSizeBytes = maxMessageSizeBytes;
         this.alloc = requireNonNull(alloc, "alloc");
 
-        unprocessed = alloc.compositeBuffer();
+        unprocessed = new ArrayDeque<>();
     }
 
     /**
@@ -236,16 +239,17 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
 
         startedDeframing = true;
 
-        if (!data.isEmpty()) {
+        final int dataLength = data.length();
+        if (dataLength != 0) {
             final ByteBuf buf;
             if (data instanceof ByteBufHolder) {
                 buf = ((ByteBufHolder) data).content();
             } else {
-                buf = alloc.buffer(data.length());
-                buf.writeBytes(data.array(), data.offset(), data.length());
+                buf = Unpooled.wrappedBuffer(data.array(), data.offset(), dataLength);
             }
             assert unprocessed != null;
-            unprocessed.addComponent(true, buf);
+            unprocessed.add(buf);
+            unprocessedBytes += dataLength;
         }
 
         // Indicate that all of the data for this stream has been received.
@@ -261,7 +265,7 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
     public void close() {
         try {
             if (unprocessed != null) {
-                unprocessed.release();
+                unprocessed.forEach(ByteBuf::release);
             }
         } finally {
             unprocessed = null;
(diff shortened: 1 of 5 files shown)


grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
```diff
@@ -177,7 +179,8 @@ public int hashCode() {
     private boolean endOfStream;
 
     @Nullable
-    private CompositeByteBuf unprocessed;
+    private Queue<ByteBuf> unprocessed;
```
anuraaga: This is really surprising - I feel like you've reimplemented `CompositeByteBuf`. Is this not a use case for that class?

Also, can you compare against the original code with a Netty snapshot? By total coincidence looks like there have been major optimizations lately

https://github.com/netty/netty/commits/4.1/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
```diff
@@ -379,6 +385,42 @@ private void readHeader() {
...
+    private long readHeaderAsLong() {
+        unprocessedBytes -= HEADER_LENGTH;
+        assert unprocessed != null;
+        final ByteBuf firstBuf = unprocessed.peek();
+        assert firstBuf != null;
+        final int firstBufLen = firstBuf.readableBytes();
+
+        if (firstBufLen >= HEADER_LENGTH) {
+            final long header = (firstBuf.readUnsignedInt() << 8) | firstBuf.readUnsignedByte();
+            if (firstBufLen == HEADER_LENGTH) {
```
anuraaga: Probably a bit clearer to use `firstBuf.isReadable` like below

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageFramer.java
```diff
@@ -147,23 +147,37 @@ private ByteBuf writeUncompressed(ByteBuf message) {
...
+                buf.writeInt(messageLength);
+                buf.writeBytes(message);
+                return buf;
+            } finally {
+                message.release();
+            }
+        }
+
+        // Frame is fairly large that composition might reduce memory footprint.
+        return new CompositeByteBuf(alloc, true, 2,
```
anuraaga: I figured a composite buf would always be copied once into a continuous buffer before writing to the socket. Do you think that doesn't happen with large buffers?

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
trustin: `CompositeByteBuf` is a general purpose buffer implementation that has to maintain a lot more information than the implementation in this PR. It's much simpler because we always read from the beginning and we do not have to worry about managing the component list after `readerIndex` advances beyond a component.

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
trustin: Fixed

trustin:
```
Benchmark                     (wrapBuffer)   Mode  Cnt    Score   Error  Units
LargePayloadBenchmark.normal         false  thrpt  100   95.581 ± 0.700  ops/s
LargePayloadBenchmark.normal          true  thrpt  100  117.076 ± 0.806  ops/s
```

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
```diff
@@ -394,10 +436,51 @@ private void readBody() {
...
+        final ByteBuf merged = alloc.buffer(length);
+        for (;;) {
+            final ByteBuf buf = unprocessed.peek();
+            assert buf != null;
+
+            final int bufLen = buf.readableBytes();
+            final int remaining = merged.writableBytes();
+
+            if (bufLen <= remaining) {
+                merged.writeBytes(buf, bufLen);
```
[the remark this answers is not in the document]
trustin: Indeed. Removed the second parameter. I thought `writeBytes` will try to read as much as `merged.writableBytes()` which can be greater than `buf.readableBytes()`.

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
```diff
@@ -379,6 +385,42 @@ private void readHeader() {
...
+    private long readHeaderAsLongSlowPath() {
+        assert unprocessed != null;
+
+        long header = 0;
+        int readBytes = 0;
+        do {
+            final ByteBuf buf = unprocessed.peek();
+            assert buf != null;
+            header <<= 8;
+            header |=  buf.readUnsignedByte();
```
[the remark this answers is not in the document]
trustin: Fixed

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
```diff
@@ -379,6 +385,42 @@ private void readHeader() {
...
+            if (!firstBuf.isReadable()) {
+                unprocessed.remove().release();
+            }
+            return header;
+        }
+
+        return readHeaderAsLongSlowPath();
+    }
+
+    private long readHeaderAsLongSlowPath() {
```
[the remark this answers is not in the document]
trustin: Did a quick perf check and there seems no change.

PR #1431, opened 2018-11-06
Make HttpHealthcheckEndpointGroup port customizable

huydx:
As we discussed internally @trustin. 
The use case is when we wrap HttpHealthCheckEndpoint by DnsEndpoint and use different port by DnsEndpoint (for example,  UDP).

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
@@ -26,6 +26,7 @@ import java.util.function.Function;
 import com.linecorp.armeria.client.Client;
 import com.linecorp.armeria.client.ClientFactory;
 import com.linecorp.armeria.client.ClientOptionsBuilder;
+import com.linecorp.armeria.client.Endpoint;
 import com.linecorp.armeria.client.HttpClient;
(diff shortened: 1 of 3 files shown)


huydx:
@trustin Endpoint mapping is a good generic idea, but it's quite confusing from user view, I would like an obvious way to custom Http check target, and IMHO port and path seems quite obvious and enough.

trustin:
OK. Then let's keep it as it is. Will continue reviewing.

huydx:
all comments resolved

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
```diff
@@ -52,6 +54,17 @@ public HttpHealthCheckedEndpointGroupBuilder(EndpointGroup delegate, String heal
         this.healthCheckPath = requireNonNull(healthCheckPath, "healthCheckPath");
     }
 
+    /**
+     * If used, a health check request will be sent to the specified port instead of the port of the
```
trustin: Sorry but how about rephrasing a little bit:

```
Sets the port where a health check request will be sent instead of the original port number
specified by {@link EndpointGroup}'s {@link Endpoint}s. This property is useful when your
server listens to health check requests on a different port.
```

huydx:
@trustin PTAL~

trustin:
Thanks, @huydx.

PR #1433, opened 2018-11-09
Fix a bug where authority header is not set properly when RetryingClient is used

minwoox:
Motivation
When an `HttpClient` sends an `HttpRequest`, the authority header is set automatically in `HttpRequestSubscriber`.
If we retry to the same host every time, we do not have to reset the authority header, but the reality is not.
So we should remove the authority header when it wasn't set by a user so that the request is sent with proper authority header.

Modifications:
- Remove the authority header in `RetryingHttpClient` when the authority header is not set by a user
- Fix flaky test

Result:
- less bugs
- less flaky

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
@@ -18,6 +18,7 @@ package com.linecorp.armeria.client.retry;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.linecorp.armeria.common.HttpHeaders.EMPTY_HEADERS;
 import static com.linecorp.armeria.internal.ClientUtil.executeWithFallback;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -131,6 +137,9 @@
 
     @Override
     protected HttpResponse doExecute(ClientRequestContext ctx, HttpRequest req) throws Exception {
+        final String authority = req.headers().authority();
+        ctx.attr(AUTHORITY_EXISTS).set(!isNullOrEmpty(authority));
```
trustin: How about setting only when the authority header exists and checking with `hasAttr()`?

minwoox: Fixed Thank you~!

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -131,6 +137,10 @@
 
     @Override
     protected HttpResponse doExecute(ClientRequestContext ctx, HttpRequest req) throws Exception {
+        if (!isNullOrEmpty(req.headers().authority())) {
+            ctx.attr(AUTHORITY_EXISTS).set(true);
```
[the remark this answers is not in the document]
minwoox: Oops you're right. We don't need it.

trustin:
Thanks, @minwoox !

PR #1434, opened 2018-11-09
Check for empty remote service name which isn't allowed.

anuraaga:
Empty service name will cause an exception to be thrown by `Span.remoteServiceName`


trustin:
Thanks, @anuraaga !

PR #1436, opened 2018-11-13
Do not use UnpooledByteBufAllocator

trustin:
Motivation:

`DefaultClientRequestContext.alloc()` returns
`UnpooledByteBufAllocator.DEFAULT` when the current `Channel` is not
determined yet. This can lead to poor performance when a `Client` tries
to get the allocator early.

Modifications:

- Return `PooledByteBufAllocator.DEFAULT` when `Channel` is not
  determined yet.

Result:

- Better client-side gRPC performance (apx. 5% improvement)

master:

    DownstreamSimpleBenchmark.empty        NORMAL  thrpt  100  75881.642 ± 178.229  ops/s

This PR:

    DownstreamSimpleBenchmark.empty        NORMAL  thrpt  100  78863.742 ± 179.395  ops/s


minwoox:
Will investigate the test failure.

trustin:
Thank you for reviewing

PR #1439, opened 2018-11-14
Throw FileNotFoundException when KeyStoreCredentialResolverBuilder ca…

hyangtack:
…nnot open a resource

Motivation:
`ClassLoader#getResourceAsStream` returns `null` if the given resource could not be found.
But we have not checked the return value. It may make a user hard to find what is the problem when a resource does not exist.

Modifications:
- Throw FileNotFoundException when KeyStoreCredentialResolverBuilder cannot open a resource
- Add tests


trustin:
Thanks, @hyangtack !

PR #1440, opened 2018-11-15
Avoid using CompletableFuture.whenCompleteAsync()

### LINKED ISSUE #1426: CompletableFuture composition is expensive due to CompletionException.fillInStackTrace()
https://gist.github.com/trustin/52c5014a04331cf4c6a452e2cd6880b3 shows that `CompletionException.fillInStackTrace()` is taking 4-5% of samples.

`CompletionException` is instantiated when your callback is added to a child `CompletableFuture` of another future. See https://stackoverflow.com/questions/49230980/does-completionstage-always-wrap-exceptions-in-completionexception for more information.

Unfortunately, it is very hard to fix this problem because of the two reasons:

First, `CompletableFuture.encodeThrowable()`, which wraps an exception with a new `CompletionException` is a static package-local method that's not overridable.

```java
// From OpenJDK 10
static AltResult encodeThrowable(Throwable x) {
    return new AltResult((x instanceof CompletionException) ? x :
                         new CompletionException(x));
}
```

Second, even if we extend `CompletableFuture`, we have no control over what `CompletableFuture` instance is returned when future is chained. This is possible since Java 9, but our minimum requirement is Java 8. And the first problem is still not solved.

```java
OurCustomCompletableFuture f = ...;
// We can't make this return OurCustomCompletableFuture. It's always CompletableFuture.
CompletableFuture child = f.whenComplete(...); 
```

There are a few options for fixing this.

1. Implement our own `CompletionStage` implementation similar to `CompletableFuture`.
2. Consider adopting Reactor Core again. Replacing `StreamMessage` with `Flux` would
   - offload us from stream-related optimization to some extent
   - enable easier integration with other projects that rely on Reactor Core, reducing the cost of building adapters, etc.
   - provide rich stream manipulation capability of Reactor Core to our users
   - .. at the cost of huge refactoring or at the risk of getting worse performance.
   - Previous discussion: https://github.com/line/armeria/issues/970

trustin:
Related: #1426

Motivation:

`CompletableFuture` always creates a new `CompletionException` after it
calls the callback registered with `whenCompleteAsync()`, even if there
are no dependants and the `CompletionException` will never be used.

See https://hg.openjdk.java.net/jdk/jdk11/file/1ddf9a99e4ad/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java#l870
for more information.

Modifications:

- Use `CompletableFuture.handleAsync()` instead of `whenCompleteAsync()`.
- Miscellaneous:
  - Avoid using `voidFunction()` which increases the level of indirection

Result:

- Fixes #1426
- Better performance

master:

    DownstreamSimpleBenchmark.empty        NORMAL  thrpt  100  52167.567 ± 274.358  ops/s

This pull request:

    DownstreamSimpleBenchmark.empty        NORMAL  thrpt  100  58171.594 ± 365.014  ops/s

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Http2ResponseDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Http2ResponseDecoder.java
@@ -71,7 +71,7 @@ final class Http2ResponseDecoder extends HttpResponseDecoder implements Http2Con
         final HttpResponseWrapper resWrapper =
                 super.addResponse(id, req, res, logBuilder, responseTimeoutMillis, maxContentLength);
 
-        resWrapper.completionFuture().whenCompleteAsync((unused, cause) -> {
+        resWrapper.completionFuture().handleAsync((unused, cause) -> {
             // Cancel timeout future and abort the request if it exists.
             resWrapper.onSubscriptionCancelled();
 
@@ -94,6 +94,8 @@ final class Http2ResponseDecoder extends HttpResponseDecoder implements Http2Con
                     }
                 }
             }
+
+            return null;
         }, channel().eventLoop());
         return resWrapper;
     }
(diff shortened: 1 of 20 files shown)


trustin:
Ran the benchmark again today for #1444 and the throughput of this pull request is now around 54,000-56,000. Maybe not as dramatic as initially observed, although it's still quite an improvement.

PR #1441, opened 2018-11-15
Revamp connection pool

### LINKED ISSUE #818: HTTP/2 connection pool can create extra connections unnecessarily

trustin:
Related: #818

Motivation:

- Unlike HTTP/1 connections, HTTP/2 connections do not need to be
  removed from the connection pool. We has slight loss of efficiency
  because we currently remove an HTTP/2 connection from the pool and
  add it back.
- Current connection pool implementation has unnecessarily deep type
  hierarchy and high level of indirection which weren't very useful.
- There's circular package dependency between `armeria.client` and
  `armeria.client.pool` which is undesirable.

Modifications:

- Merged `HttpSessionChannelFactory`, `KeyedChannelPool` and
  `DefaultKeyedChannelPool` into `HttpChannelPool`.
  - Tuned the internal data structure of `DefaultKeyedChannelPool`.
    - `PoolKey` does not include `SessionProtocol` anymore. We keep
      separate `Map` for each `SessionProtocol`.
  - Improved the connection pool implementation so that the connections
    with the same session protocol are grouped together even if they
    were created from different URI schemes. For example, both `h2c://`
    and `http://` can create `h2c` connections, but their connections
    were treated differently, but not anymore.
  - Removed unnecessary synchronization since all code in
    `HttpChannelPool` should be run in an event loop.
- Replaced `KeyedChannelPoolHandler` and its related classes with
  `ConnectionPoolListener` et al.
  - We do not notify the listeners about acquisition and release because
    it has no meaning for HTTP/2 and it is more meaningful to retrieve
    such information from `RequestLog`. Let me know if you have a
    suggestions for more listener methods.

Result:

- Fixes #818
- Connection pool listener API has a breaking change.


core/src/main/java/com/linecorp/armeria/client/SafeConnectionPoolListener.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.linecorp.armeria.common.SessionProtocol;
+
+import io.netty.util.AttributeMap;
+
+/**
+ * Helper Handler delegate event to decorated  {@link ConnectionPoolListener}.
```
hyangtack: Need to revise this Javadoc?

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -103,27 +97,37 @@ private void finishResolve(ClientRequestContext ctx, Endpoint endpoint,
...
+
+        final PoolKey key = new PoolKey(host, ipAddr, port);
+        final PooledChannel pooledChannel = pool.acquireNow(protocol, key);
+        if (pooledChannel != null) {
+            invoke0(pooledChannel,ctx, req, res);
         } else {
-            final Throwable cause = channelFuture.cause();
-            handleEarlyRequestException(ctx, req, cause);
-            res.close(cause);
+            pool.acquire(protocol, key).whenComplete((newPooledChannel, cause) -> {
```
hyangtack: Just curious. Don't we need to use `handle` instead of `whenComplete` in order to avoid creating `CompletionException` internally?

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -103,27 +97,37 @@ private void finishResolve(ClientRequestContext ctx, Endpoint endpoint,
...
+
+        final String host = extractHost(ctx, req, endpoint);
+        final int port = endpoint.port();
+        final SessionProtocol protocol = ctx.sessionProtocol();
+        final HttpChannelPool pool = factory.pool(ctx.eventLoop());
+
+        final PoolKey key = new PoolKey(host, ipAddr, port);
+        final PooledChannel pooledChannel = pool.acquireNow(protocol, key);
+        if (pooledChannel != null) {
+            invoke0(pooledChannel,ctx, req, res);
```
minwoox: nit: space?

core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
```diff
@@ -92,10 +92,10 @@
      */
     private boolean needsRetryWithH1C;
 
-    HttpSessionHandler(HttpSessionChannelFactory channelFactory, Channel channel,
+    HttpSessionHandler(HttpChannelPool channelPool, Channel channel,
                        Promise<Channel> sessionPromise, ScheduledFuture<?> sessionTimeoutFuture) {
 
-        this.channelFactory = requireNonNull(channelFactory, "channelFactory");
+        this.channelPool = requireNonNull(channelPool, "channelFactory");
```
minwoox: "channelPool"?

core/src/main/java/com/linecorp/armeria/client/SafeConnectionPoolListener.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright 2015 LINE Corporation
```
minwoox: year?

core/src/main/java/com/linecorp/armeria/client/logging/ConnectionPoolLoggingListener.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2015 LINE Corporation
```
minwoox: one more?

core/src/main/java/com/linecorp/armeria/client/logging/ConnectionPoolLoggingListener.java
```diff
@@ -0,0 +1,89 @@
...
+        this.ticker = requireNonNull(ticker, "ticker");
+    }
+
+    @Override
+    public void connectionOpen(SessionProtocol protocol,
+                               InetSocketAddress remoteAddr,
+                               InetSocketAddress localAddr,
+                               AttributeMap attrs) throws Exception {
+        final int activeChannels = this.activeChannels.incrementAndGet();
+        attrs.attr(OPEN_NANOS).set(ticker.read());
```
minwoox: This line can be in the `if` statement as well?

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -0,0 +1,533 @@
...
+        final String host;
+        final String ipAddr;
+        final int port;
+        final int hashCode;
+
+        PoolKey(String host, String ipAddr, int port) {
+            this.host = host;
+            this.ipAddr = ipAddr;
+            this.port = port;
+            hashCode = ipAddr.hashCode() * 31 + port;
```
anuraaga: Is it ok to use `host` in `equals` but not in `hashCode`? If it is let's add a comment since it's not conventional.

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -0,0 +1,533 @@
...
+    private void acquire(SessionProtocol desiredProtocol, PoolKey key,
+                         CompletableFuture<PooledChannel> promise) {
+        final PooledChannel ch = acquireNow(desiredProtocol, key);
+        if (ch != null) {
+            promise.complete(ch);
+            return;
+        }
+
+        // Try to use the pending connection to avoid creating an extra connection.
+        final Future<Channel> pendingChannel = getPendingConnection(desiredProtocol, key);
```
anuraaga: Should we check `desiredProtocol.isMultiplex()`?

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
trustin: It's OK because we did not chain after calling `whenComplete`, but let me switch to `handle` anyway just in case we add more in the future.

core/src/main/java/com/linecorp/armeria/client/ConnectionPoolListenerAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+
+import io.netty.util.AttributeMap;
+
+/**
+ * A skeletal {@link ConnectionPoolListener} implementation to minimize the effort to implement this interface.
+ * Extend this class to implement only a few of the provided handler methods.
+ */
+public class ConnectionPoolListenerAdapter implements ConnectionPoolListener {
+
+    static final ConnectionPoolListenerAdapter NOOP = new ConnectionPoolListenerAdapter();
```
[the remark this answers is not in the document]
trustin: Only if we were allowed to have `private static` members in an interface...

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    private static boolean isHealthy(PooledChannel pooledChannel) {
+        final Channel ch = pooledChannel.get();
+        return ch.isActive() && HttpSession.get(ch).isActive();
+    }
+
+    @Nullable
+    private static SessionProtocol getProtocolIfHealthy(Channel ch) {
+        if (!ch.isActive()) {
```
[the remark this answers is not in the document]
trustin: It's not necessary because `HttpSession.protocol()` returns `null` is session is inactive.

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -0,0 +1,533 @@
...
+        assert future.isDone();
+        removePendingConnection(desiredProtocol, key);
+
+        try {
+            if (future.isSuccess()) {
+                final Channel channel = future.getNow();
+                final SessionProtocol protocol = getProtocolIfHealthy(channel);
+                if (closed || protocol == null) {
+                    channel.close();
+                    promise.completeExceptionally(ClosedSessionException.get());
```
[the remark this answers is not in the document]
trustin: Sounds like a good idea.

core/src/main/java/com/linecorp/armeria/client/logging/ConnectionPoolLoggingListener.java
trustin: Fixed. Also added some defensive code in `connectionClosed` just in case log level changes at runtime.

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
trustin: We do not know the protocol will be multiplex or not at this point if `desiredProtocol` is `HTTP` or `HTTPS`, but yeah we can do something after the pending connection is ready. Updated the logic and extracted it into `usePendingAcquisition()`. Please review.

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
trustin: It's OK, but I guess it doesn't do harm by adding `host.hashCode()`, so I added it. :-)

trustin: That is, without `host.hashCode()` it had higher chance of collision for the servers that utilize virtual hosts.

core/src/main/java/com/linecorp/armeria/client/SafeConnectionPoolListener.java
trustin: Removed `SafeConnectionPoolListener` since just wrapping the notification code with `try-catch` seemed fine.

core/src/main/java/com/linecorp/armeria/client/ConnectionPoolListenerAdapter.java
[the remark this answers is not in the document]
anuraaga: Sorry didn't notice this is a class, not interface!

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -165,24 +170,62 @@ private void acquire(SessionProtocol desiredProtocol, PoolKey key,
...
+                } else {
+                    // Try to acquire again because the connection was not HTTP/2.
+                    // We use the exact protocol (H1 or H1C) instead of 'desiredProtocol' so that
+                    // we do not waste our time looking for pending acquisitions for the host
+                    // that does not support HTTP/2.
+                    acquire(pch.protocol(), key, promise);
+                }
+            } else {
+                // Try to acquire again because the pending connection attempt has failed.
+                acquire(desiredProtocol, key, promise);
```
minwoox: Might be silly question, but just wondering if there's any chance that an event loop tries this `acquire` forever. :-)

trustin: I guess it is possible. Pushed a fix.

trustin:
@anuraaga You may want to check this commit as well which got into after your approval https://github.com/line/armeria/pull/1441/commits/96fa6c336bd3da487e80a8a53ca9088b205c6702

minwoox:
LGTM again~!

anuraaga:
LGTM

PR #1444, opened 2018-11-16
Even less usage of CompletableFuture.whenComplete()

trustin:
See #1440


trustin:
Getting slightly higher number:

master:

```
DownstreamSimpleBenchmark.empty        NORMAL  thrpt  100  54468.648 ± 338.609  ops/s
```

this pull request:

```
DownstreamSimpleBenchmark.empty        NORMAL  thrpt  100  55929.070 ± 342.300  ops/s
```

PR #1445, opened 2018-11-16
Add public methods for users to determine the channel type of an even…

### LINKED ISSUE #1438: Expose TransportType to users?
I needed to write a TCP client in an armeria server and ended up using the `TransportType` internal class to also take advantage of `CommonPools`. Since we expose `CommonPools`, does it make sense to expose `TransportType` for this sort of use case? Maybe it's overkill, just want to get it out there.

```java
    var workerGroup = CommonPools.workerGroup();
    var bootstrap = new Bootstrap();
    bootstrap.group(workerGroup);
    // NOTE(rag): This is an internal API, but should be stable. Will see about exposing it to
    // users.
    bootstrap.channel(TransportType.socketChannelType(workerGroup));
    bootstrap.option(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT);
    bootstrap.option(ChannelOption.SO_KEEPALIVE, true);
    bootstrap.handler(
        new ChannelInitializer<SocketChannel>() {
          @Override
          protected void initChannel(SocketChannel ch) {
            ch.pipeline().addLast(new Handler());
          }
        });

    channel = bootstrap.connect(target.getHost(), target.getPort()).syncUninterruptibly().channel();
```

anuraaga:
…t loop group.

This can be useful for users that want to use armeria's `CommonPools` with a custom netty pipeline, such as a raw TCP client.

Fixes #1438

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/EventLoopGroups.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/EventLoopGroups.java
@@ -23,7 +23,11 @@ import java.util.concurrent.ThreadFactory;
 
 import com.linecorp.armeria.internal.TransportType;
 
+import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.ServerChannel;
+import io.netty.channel.socket.DatagramChannel;
+import io.netty.channel.socket.SocketChannel;
 import io.netty.util.concurrent.DefaultThreadFactory;
 
 /**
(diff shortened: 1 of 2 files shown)


trustin:
Thanks, @anuraaga!

anuraaga:
Oops - fixed, though I wonder if the new order is actually an improvement

trustin:
Thanks, @anuraaga :-)

core/src/main/java/com/linecorp/armeria/common/util/EventLoopGroups.java
```diff
@@ -93,5 +97,29 @@ public static EventLoopGroup newEventLoopGroup(int numThreads, ThreadFactory thr
         return type.newEventLoopGroup(numThreads, unused -> threadFactory);
     }
 
+    /**
+     * Returns the {@link ServerChannel} class that is available for this {@code eventLoopGroup}, for use in
+     * configuring a custom {@link Bootstrap}.
+     */
+    public static Class<? extends ServerChannel> serverChannelClass(EventLoopGroup eventLoopGroup) {
```
trustin: Just realized that this has to be `serverChannelType`. Would you mind sending a follow-up pull request, @anuraaga ?

Issue #1446, opened 2018-11-19
StreamMessageDuplicator feeds null to its Subscriber


trustin:
```
java.lang.NullPointerException: null 
	at com.linecorp.armeria.common.HttpMessageAggregator.onData(HttpMessageAggregator.java:91) 
	at com.linecorp.armeria.common.HttpResponseAggregator.onData(HttpResponseAggregator.java:96) 
	at com.linecorp.armeria.common.HttpMessageAggregator.onNext(HttpMessageAggregator.java:78) 
	at com.linecorp.armeria.common.HttpMessageAggregator.onNext(HttpMessageAggregator.java:36) 
	at com.linecorp.armeria.common.stream.DeferredStreamMessage$ForwardingSubscriber.onNext(DeferredStreamMessage.java:302) 
	at com.linecorp.armeria.common.stream.AbstractStreamMessageDuplicator$DownstreamSubscription.doSignalSingle(AbstractStreamMessageDuplicator.java:747) 
	at com.linecorp.armeria.common.stream.AbstractStreamMessageDuplicator$DownstreamSubscription.doSignal(AbstractStreamMessageDuplicator.java:657) 
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163) 
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:404) 	at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:326) 
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:897) 
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) 
	at java.lang.Thread.run(Thread.java:745)
```

trustin:
Fixed via #1449

PR #1449, opened 2018-11-20
Fix NPE in AbstractStreamMessageDuplicator

minwoox:
Fix NPE in AbstractStreamMessageDuplicator
Motivation:
`StreamMessageDuplicator` feed null to its `Subscriber`. (related with #1446)

Modifications:
- Assign the `elements` to a local variable in the method `get(index)` in `SignalQueue` so that the reference is not changed during the get operation
- Add some `assert` statements which produce useful information for debugging purpose

Result:
- Fewer bugs

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.common.stream;
 
+import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -902,10 +920,15 @@ private void doubleCapacity() {
...
+        /**
+         * Invoked by the executor in {@link DownstreamSubscription}.
+         */
         Object get(int offset) {
             final int head = this.head;
             final int tail = this.tail;
+            final Object[] elements = this.elements;
+            assert elements != null : "elements is null. SignalQueue: " + this;
+
             final int length = elements.length;
```
minwoox: This was the problem. The reference to this `elements` can be changed before the current thread reaches line 943.

trustin:
Thanks, @minwoox !

PR #1458, opened 2018-11-25
Make armeria support retrofit 2.5

kojilin:
* Timeout.deadlineNanoTime is similar to ClientRequestContext.responseTimeoutMillis, but client-side timeout is not runtime-adjustable in Armeria, unlike Timeout API does.
* There's nothing in Armeria that matches the contract of Timeout.timeoutNanos.

Until we can support them, perhaps we could use Timeout.NONE

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -92,7 +92,7 @@ com.spotify:
 
 com.squareup.retrofit2:
   retrofit:
-    version: &RETROFIT2_VERSION '2.4.0'
+    version: &RETROFIT2_VERSION '2.5.0'
     javadocs:
     - https://square.github.io/retrofit/2.x/retrofit/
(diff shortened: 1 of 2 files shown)


trustin:
My findings so far:

- `Timeout.deadlineNanoTime` is similar to `ClientRequestContext.responseTimeoutMillis`, but client-side timeout is not runtime-adjustable in Armeria, unlike `Timeout` API does.
- There's nothing in Armeria that matches the contract of `Timeout.timeoutNanos`.

Until we can support them, perhaps we could use `Timeout.NONE`?

kojilin:
follow PR review comment. updated.

kojilin:
updated.

trustin:
Thank you, @kojilin !

PR #1459, opened 2018-11-26
Allow `reactor-netty` and `armeria-spring-boot-webflux-autoconfigure` to coexist

trustin:
Originally reported by @cj848
Motivation:

It is currently not possible to use `ArmeriaReactiveWebServerFactory`
without excluding `reactor-netty`.

Modifications:

- Do not import `ReactiveWebServerFactoryAutoConfiguration` when
  `ArmeriaReactiveWebServerFactoryAutoConfiguration` is loaded.
- Miscellaneous:
  - Remove trailing spaces from `README.md`
  - Remove Guava usage from examples
  - Remove unused `logback-test.xml`

Result:

- A user can use `reactor-netty` as a standalone library while using
  Armeria as a reactive `WebServer`.
- Less dependencies in examples


trustin:
Confirmed that `WebClient` is not affected by this issue. Hence no modification there.

PR #1461, opened 2018-11-27
Unescape grpc message description in client.

### LINKED ISSUE #1460: gRPC status description is not decoded.
When a user raises a status exception with non-ASCII description:

```java
observer.onError(Status.INTERNAL.withCause(e).withDescription("오류").asRuntimeException());
```

A user who uses Armeria client should get the status description `"오류"`, but gets its percent-encoded value.

Originally reported by @cj848

/cc @anuraaga

anuraaga:
Completely forgot to call the `unescape` method anywhere outside of unit tests.

Fixes #1460


trustin:
Thanks, @anuraaga !

PR #1466, opened 2018-11-29
Fix to add exampleHttpHeaders to the DocService debug page

minwoox:
Motivation:
At the debug page in the documentation service, only service level `exampleHttpHeaders` are shown.

Modification:
- Fix to add method level and entire level `exampleHttpHeaders` to the debug page

Result:
- Better debug page

### DIFF
--- a/docs-client/package.json
+++ b/docs-client/package.json
@@ -15,6 +15,7 @@
     "jsonminify": "^0.4.1",
     "react": "^16.6.0",
     "react-dom": "^16.6.0",
+    "react-dropdown": "^1.6.2",
     "react-helmet": "^5.2.0",
     "react-hot-loader": "^4.3.12",
     "react-router-dom": "^4.3.1",
--- a/docs-client/src/containers/MethodPage/index.tsx
+++ b/docs-client/src/containers/MethodPage/index.tsx
@@ -220,7 +220,7 @@ export default class MethodPage extends React.PureComponent<Props, State> {
                     <>
                       {this.state.exampleHeaders.length > 0 && (
                         <Dropdown
-                          placeholder="Select an example headers"
+                          placeholder="Select an example headers..."
                           options={this.state.exampleHeaders}
                           onChange={this.onSelectedHeadersChange}
                         />
--- a/docs-client/src/containers/MethodPage/index.tsx
+++ b/docs-client/src/containers/MethodPage/index.tsx
@@ -14,6 +14,8 @@
  * under the License.
  */
 
+import 'react-dropdown/style.css';
+
 import Button from '@material-ui/core/Button';
 import Checkbox from '@material-ui/core/Checkbox';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
@@ -34,6 +36,7 @@ import DeleteSweepIcon from '@material-ui/icons/DeleteSweep';
 import FileCopyIcon from '@material-ui/icons/FileCopy';
 import jsonMinify from 'jsonminify';
 import React, { ChangeEvent } from 'react';
+import Dropdown, { Option } from 'react-dropdown';
 import { RouteComponentProps } from 'react-router-dom';
 import SyntaxHighlighter from 'react-syntax-highlighter';
 import githubGist from 'react-syntax-highlighter/styles/hljs/github-gist';
@@ -50,6 +53,7 @@ interface State {
   debugResponse: string;
   additionalHeadersOpen: boolean;
   additionalHeaders: string;
+  exampleHeaders: Option[];
   stickyHeaders: boolean;
 }
 
@@ -66,6 +70,7 @@ export default class MethodPage extends React.PureComponent<Props, State> {
     debugResponse: '',
     additionalHeadersOpen: false,
     additionalHeaders: '',
+    exampleHeaders: [],
     stickyHeaders: false,
   };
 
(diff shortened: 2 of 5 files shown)


docs-client/src/containers/MethodPage/index.tsx
```diff
@@ -431,10 +431,20 @@ export default class MethodPage extends React.PureComponent<Props, State> {
...
 
-    const exampleHeaders =
-      service.exampleHttpHeaders.length > 0
-        ? JSON.stringify(service.exampleHttpHeaders[0], null, 2)
-        : undefined;
+    let exampleHeaders;
+    if (method.exampleHttpHeaders.length > 0) {
+      exampleHeaders = JSON.stringify(method.exampleHttpHeaders[0], null, 2);
+    } else if (service.exampleHttpHeaders.length > 0) {
+      exampleHeaders = JSON.stringify(service.exampleHttpHeaders[0], null, 2);
```
trustin: Perhaps we could concat method/server/specification-level examples?

minwoox:
I just added the `exampleHttpHeaders` in the order of method, service and entire level. But I was wondering if I have to append the `exampleHttpHeaders` in the order of entire, service and method while overriding same headers.

docs-client/src/containers/MethodPage/index.tsx
```diff
@@ -431,16 +456,45 @@ export default class MethodPage extends React.PureComponent<Props, State> {
...
       : undefined;
 
-    const exampleHeaders =
-      service.exampleHttpHeaders.length > 0
-        ? JSON.stringify(service.exampleHttpHeaders[0], null, 2)
-        : undefined;
+    const exampleHeaders: Option[] = [];
+    if (method.exampleHttpHeaders.length > 0) {
+      exampleHeaders.push({
+        value: JSON.stringify(method.exampleHttpHeaders[0], null, 2),
```
hyangtack: Just curious. Is there any reason why we use the first header only?

docs-client/src/containers/MethodPage/index.tsx
```diff
@@ -212,6 +217,15 @@ export default class MethodPage extends React.PureComponent<Props, State> {
                 <Typography variant="body1" paragraph />
                 {this.state.additionalHeadersOpen && (
                   <>
+                    <>
+                      {this.state.exampleHeaders.length > 1 && (
```
hyangtack: Just curious. If `exampleHeaders` didn't set to `additionalHeader` at https://github.com/line/armeria/pull/1466/files#diff-f34c1c4217b18c494f39ebed6d5bfab9R494, isn't it necessary to provide this dropdown menu in order that a user can choose an example header?

minwoox: If there's only one example headers, it is set to `EDIT HTTP HEADERS` automatically.
If there are more than one example headers, this dropdown shows up so that a user can choose.

docs-client/package.json
```diff
@@ -15,6 +15,7 @@
     "jsonminify": "^0.4.1",
     "react": "^16.6.0",
     "react-dom": "^16.6.0",
+    "react-dropdown": "^1.6.2",
```
trustin: I guess you can use `<Select/>` provided by Material UI, which is our current UI toolkit? https://material-ui.com/demos/selects/ (I wish to migrate to something else like Blueprint later though..)

docs-client/src/containers/MethodPage/index.tsx
trustin: I guess it may still be useful to show the drop-down control even if there is only one example. So that a user can reset the additional headers field with the example headers after editing the additional headers field.

docs-client/package.json
trustin: Had a chat with @minwoox offline and he told me that he chose `react-dropdown` intentionally because it was easier to work with than `Select`. Given that we will move away from Material UI, I think it's fine as it is.

docs-client/src/containers/MethodPage/index.tsx
```diff
@@ -491,10 +487,7 @@ export default class MethodPage extends React.PureComponent<Props, State> {
       exampleHeaders,
       debugRequest: urlDebugRequest || method.exampleRequests[0] || '',
       debugResponse: '',
-      additionalHeaders:
-        urlHeaders ||
-        stateHeaders ||
-        (exampleHeaders.length === 1 ? exampleHeaders[0].value || '' : ''),
```
hyangtack: I'm not sure but isn't this okay to let as it is?

minwoox: Reverted it. Thanks~!

docs-client/src/containers/MethodPage/index.tsx
```diff
@@ -212,6 +217,15 @@ export default class MethodPage extends React.PureComponent<Props, State> {
                 <Typography variant="body1" paragraph />
                 {this.state.additionalHeadersOpen && (
                   <>
+                    <>
+                      {this.state.exampleHeaders.length > 0 && (
+                        <Dropdown
+                          placeholder="Select an example headers"
```
trustin: `Select an example headers...` ? (... to indicate that there's something to select after clicking on it?)

trustin:
@anuraaga Interested in reviewing?

docs-client/src/containers/MethodPage/index.tsx
minwoox: Updated. Thanks~!

anuraaga:
Oh can you run the command in the readme to update licenses?

trustin:
Thanks, @minwoox!

Issue #1082, opened 2018-03-12
Default ClientFactory's shutdown hook conflicts Spring's shutdown hook


trustin:
A Spring user who uses `ClientFactory.DEFAULT` may observe the problem where the default `ClientFactory` is closed before his/her Spring application is fully destroyed.

We should figure out what the best way to solve this. One option would be to provide an option to disable the shutdown hook, but it doesn't look elegant to me.

rfkm:
Let me explain the trouble I faced before.
While my project uses Armeria with a custom factory, I accidentally had set up `HttpHealthCheckedEndpointGroup` without specifying our own factory. 
As a result, all requests failed even though the custom factory was still alive during the shutdown process, because the default factory was already closed and that marked all endpoints unhealthy.

It still looks a good idea to me to provide an option to disable the default shutdown hook, though it may not help users to avoid such a mistake. I guess many frameworks that register a shutdown hook have such an option.

trustin:
Thanks for adding background information about this issue, @rfkm. Very helpful.

trustin:
Fixed via #1472

PR #1472, opened 2018-12-03
Allow disabling default ClientFactory shutdown hook

trustin:
See: #1082
Motivation:

A Spring user who uses `ClientFactory.DEFAULT` may observe the problem
where the default `ClientFactory` is closed before his/her Spring
application is fully destroyed.

Modifications:

- Added `ClientFactory.disableShutdownHook()`

Result:

- A user have full control over when the default `ClientFactory` can be
  closed.


trustin:
Sorry that it took so long, @rfkm.

trustin:
Thanks for reviewing!

PR #1474, opened 2018-12-04
Fix endpoint group health check metrics register different tag problem.

kojilin:
Sometime we may have some endpoint groups using ip, some not. But
MeterRegistry doesn't allow us to create label with different tags. So
we need to set blank to ip tag if there is no ip addr.


trustin:
```
[ant:checkstyle] [ERROR] C:\projects\armeria\core\src\test\java\com\linecorp\armeria\client\endpoint\healthcheck\HttpHealthCheckedEndpointGroupTest.java:166:95: Empty statement. [EmptyStatement]
```

PR #1475, opened 2018-12-04
Refine document of RetryStrategy#shouldRetry

kojilin:
Not return null directly but null is contained in `CompletionStage`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
@@ -113,7 +113,7 @@ public interface RetryStrategy {
 
     /**
      * Returns a {@link CompletionStage} that contains {@link Backoff} which will be used for retry.
-     * If the condition does not match, this will return a {@link CompletionStage} that completed with
+     * If the condition does not match, this will return a {@link CompletionStage} completed with
      * {@code null} to stop retry attempt.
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -113,7 +113,8 @@ static RetryStrategy onStatus(
 
     /**
      * Returns a {@link CompletionStage} that contains {@link Backoff} which will be used for retry.
-     * If the condition does not match, this will return {@code null} to stop retry attempt.
+     * If the condition does not match, this will return a {@link CompletionStage} that contains {@code null}
```
trustin: `that contains null` -> `completed with {@code null}` ?

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -113,7 +113,8 @@ static RetryStrategy onStatus(
 
     /**
      * Returns a {@link CompletionStage} that contains {@link Backoff} which will be used for retry.
-     * If the condition does not match, this will return {@code null} to stop retry attempt.
+     * If the condition does not match, this will return a {@link CompletionStage} that completed with
```
trustin: that completed -> completed (short form of 'that is completed')

trustin: Thanks for fixing :-)

PR #1478, opened 2018-12-05
Exception handlers handle an exception raised from a decorator

### LINKED ISSUE #1477: Exception handlers should handle an exception raised by a decorator
In annotated HTTP service, a user may use a decorator and an exception handler for his or her service. In this case, it would be more useful to let the exception handler handle an exception raised by the decorator.

hyangtack:
Motivation:
In annotated HTTP service, a user may use both a decorator and an exception handler together for his or her service.
In this case, it would be more useful to let the exception handler handle an exception raised from the decorator.

Modifications:
- Add `ExceptionFilteredHttpResponseDecorator` to the first of a decorator chain when the chain is not empty.
  - It wraps an `HttpResponse` with `ExceptionFilteredHttpResponse` if the response is not an instance of `ExceptionFilteredHttpResponse`, in order to handle an exception raised from a decorator.
- Avoid converting `HttpResponseException` to a response, because `HttpResponseException` holds an `HttpResponse` inside.

Result:
- A user can handle exceptions in a generalized way even if any exception is raised from a decorator.
- Closes #1477

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpService.java
@@ -301,7 +301,7 @@ final class AnnotatedHttpService implements HttpService {
     }
 
     /**
-     * Intercepts a {@link HttpResponse} and wraps the response with an {@link ExceptionFilteredHttpResponse}
+     * Intercepts an {@link HttpResponse} and wraps the response with an {@link ExceptionFilteredHttpResponse}
      * if it is not an instance of {@link ExceptionFilteredHttpResponse}. This decorator will make an
      * {@link Exception} to be handled by {@link ExceptionHandlerFunction}s even if the exception is raised
      * from a decorator.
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/server/AnnotatedHttpService.java
```diff
@@ -290,6 +291,42 @@ private HttpResponse convertException(RequestContext ctx, HttpRequest req, Throw
...
+     * Returns a {@link Function} which produces a {@link Service} wrapped with an
+     * {@link ExceptionFilteredHttpResponseDecorator}.
+     */
+    Function<Service<HttpRequest, HttpResponse>,
+            ? extends Service<HttpRequest, HttpResponse>> exceptionHandlingDecorator() {
+        return ExceptionFilteredHttpResponseDecorator::new;
+    }
+
+    /**
+     * Intercepts a {@link HttpResponse} and wraps the response with an {@link ExceptionFilteredHttpResponse}
```
minwoox: an {@link HttpResponse} ?

trustin:
Thanks, @hyangtack and @minwoox

PR #1480, opened 2018-12-06
Support Spring validation on annotated HTTP services

### LINKED ISSUE #1473: support for spring  validation framework
is there a way to integrate armeria with spring  validation framework?

currently, when I use annotation @org.springframework.validation.annotation.Validated with @RequestObject on http post body, it it not working.

what is the correct way to validate http post data in armeria?

thanks in advance.

hyangtack:
Motivation:
If a user uses Spring and Armeria together, he or she may want to apply Java Bean Validation framework to Armeria annotated HTTP service.

Modifications:
- Use `org.reflections.ReflectionUtils` when getting methods or constructors from a class.
- Check all super classes when getting annotations from a class.
  - `ReflectionUtils.getAllAnnotations()` returns a `HashSet` of `Annotations` which does not preserve their specified order.
- Add `HelloAnnotatedService` and related classes to `examples/spring-boo-webflux` in order to show how to use an annotated HTTP service with Spring framework.

Result:
- Closes #1473

### DIFF
--- a/examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloAnnotatedService.java
+++ b/examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloAnnotatedService.java
@@ -9,6 +9,12 @@ import com.linecorp.armeria.server.annotation.ExceptionHandler;
 import com.linecorp.armeria.server.annotation.Get;
 import com.linecorp.armeria.server.annotation.Param;
 
+/**
+ * Note that this is not a Spring WebFlux-based component but an annotated HTTP service that leverages
+ * Armeria's built-in annotations.
+ *
+ * @see <a href="https://line.github.io/armeria/server-annotated-service.html">Annotated HTTP Service</a>
+ */
 @Component
 @Validated
 @ExceptionHandler(ValidationExceptionHandler.class)
@@ -21,6 +27,6 @@ public class HelloAnnotatedService {
     public String hello(
             @Size(min = 3, max = 10, message = "name should have between 3 and 10 characters")
             @Param String name) {
-        return String.format("Hello, %s! This is from Armeria annotated service!", name);
+        return String.format("Hello, %s! This message is from Armeria annotated service!", name);
     }
 }
--- a/examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloAnnotatedService.java
+++ b/examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloAnnotatedService.java
@@ -1,0 +1,26 @@
+package example.springframework.boot.webflux;
+
+import javax.validation.constraints.Size;
+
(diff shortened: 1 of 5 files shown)


examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloConfiguration.java
```diff
@@ -73,4 +75,16 @@ public ArmeriaClientConfigurator armeriaClientConfigurator(ClientFactory clientF
...
     }
+
+    /**
+     * Returns an {@link AnnotatedServiceRegistrationBean} which will be used to add an annotated HTTP
+     * service object to a {@link ServerBuilder}. If you want to only use Spring controllers, you do not
+     * need to create this bean.
+     */
+    @Bean
+    public AnnotatedServiceRegistrationBean annotatedServiceRegistrationBean(HelloAnnotatedService service) {
+        return new AnnotatedServiceRegistrationBean().setServiceName("hello")
```
trustin: How about just adding the annotated service in `armeriaServerConfigurator` programmatically? We may want to remove/deprecate `*RegistrationBean` in favor of annotation-based scanning, like we discussed before.

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/ValidationExceptionHandler.java
```diff
@@ -0,0 +1,94 @@
...
+ * A sample exception handler which handles a {@link ValidationException}.
+ */
+public class ValidationExceptionHandler implements ExceptionHandlerFunction {
+
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    @Override
+    public HttpResponse handleException(RequestContext ctx, HttpRequest req, Throwable cause) {
+        if (cause instanceof ValidationException) {
+            final HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
```
trustin: Perhaps `BAD_REQUEST`? We could use `INTERNAL_SERVER_ERROR` on `JsonProcessingException` though.

examples/spring-boot-webflux/src/main/java/example/springframework/boot/webflux/HelloAnnotatedService.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.validation.annotation.Validated;
+
+import com.linecorp.armeria.server.annotation.ExceptionHandler;
+import com.linecorp.armeria.server.annotation.Get;
+import com.linecorp.armeria.server.annotation.Param;
+
+@Component
+@Validated
+@ExceptionHandler(ValidationExceptionHandler.class)
+public class HelloAnnotatedService {
```
trustin: Could you add a short Javadoc which explains this is not a Webflux-based service but an annotated service that leverages Armeria's built-in annotations? We could also add a link to our official documentation about annotated services.

PR #1481, opened 2018-12-06
Respect MAX_CONCURRENT_STREAMS on the client side

### LINKED ISSUE #1206: ArmeriaRetrofitBuilder & HTTP2 - Max Number of Streams
Hi,

I'm using the ArmeriaRetrofitBuilder to wrap an HTTP2 REST API call asynchronous:

```
Retrofit retrofit = new ArmeriaRetrofitBuilder()
        .baseUrl(url)
        .addConverterFactory(JacksonConverterFactory.create(mapper))
        .addCallAdapterFactory(Java8CallAdapterFactory.create())
        .build();

SomeService service = retrofit.create(SomeService.class);
Call<String> sayHello = server.sayHello();
sayHello.enqueue(this);

```

But I am getting this error:

`io.netty.handler.codec.http2.Http2Exception$StreamException: Maximum active streams violated for this endpoint.`

Am I using something wrong?
If not, how do I set the max number of active streams?

Thanks

trustin:
Related: #1206
Motivation:

Armeria HTTP client does not respect `MAX_CONCURRENT_STREAMS` HTTP/2
setting. As a result, it is possible for a user to create more streams
than allowed accidentally, which results it protocol violation.

Modifications:

- Do not reuse an existing HTTP/2 connection but create a new one when
  the connection already has many concurrent streams.
- Clean up acquisition logic in `HttpChannelPool`.

Result:

- Fixes #1206
- No more protocol violation

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/client/HttpClientMaxConcurrentStreamTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/HttpClientMaxConcurrentStreamTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 LINE Corporation
+ * Copyright 2018 LINE Corporation
  *
  * LINE Corporation licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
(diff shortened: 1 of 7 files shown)


core/src/test/java/com/linecorp/armeria/client/HttpClientMaxConcurrentStreamTest.java
```diff
@@ -0,0 +1,169 @@
+/*
+ * Copyright 2015 LINE Corporation
```
hyangtack: nit: 2018 :-)

PR #1483, opened 2018-12-07
Remove the usage of Retrofit `Java8CallAdapterFactory`

trustin:
Motivation:

Since Retrofit 2.5, there is no need to use `Java8CallAdapterFactory`
because it's now built-in.

Modifications:

- Remove the references to `Java8CallAdapterFactory`.
- Remove `adapter-java8` from the dependencies.

Result:

- Less cruft


trustin:
Thanks for reviewing!

Issue #1471, opened 2018-12-03
Propagate gRPC stacktrace using DebugInfo proto


anuraaga:
There seems to be a semi-official protocol for propagating stacktraces through gRPC using DebugInfo

https://github.com/grpc/grpc-java/blob/master/examples/src/main/java/io/grpc/examples/errorhandling/DetailErrorSample.java#L78

We should add support to the server to populate a stacktrace into DebugInfo and the client to reconstruct a `Throwable` cause from it.

anuraaga:
I am having second thoughts on this - currently setting verbose responses will let normal logging, like `LoggingClient` but with this proto it requires special handling. The proto is only a list of strings so it's tricky or not possible to reconstruct a proper `Throwable` in the client - I would probably end up parsing it and stuffing it into `message` there anyways to make sure it gets logged.

Any thoughts on this?

trustin:
You mean putting the stack trace into both `grpc-message` and `DebugInfo`?

trustin:
Shouldn't a gRPC exception be recorded in `RequestLog.responseCause`?

anuraaga:
I read more at the example and found this snippet

https://github.com/grpc/grpc-java/blob/master/examples/src/main/java/io/grpc/examples/errorhandling/DetailErrorSample.java#L52

I think the recommendation is to use an application-specific proto, not necessarily `DebugInfo` itself. This makes sense to me since `DebugInfo` is basically just repeated string and doesn't seem to provide much value IMO.

I'm thinking of creating an armeria-specific proto with enough information to reconstruct `StackTraceElement` in the client and attaching a `Throwable` as the cause when decoding an error status. That way, logging should work fine since the stack trace will be logged as the cause of the `Throwable` by logging libraries, similar to how currently they log the stack trace as the message itself.

How does this sound?

trustin:
Sounds good to me.

PR #1488, opened 2018-12-09
Allow users to set additional trailers using ServiceRequestContext.

### LINKED ISSUE #1468: Add ServiceRequestContext.addAdditionalResponseTrailer
Currently we have `ServiceRequestContext.addAdditionalResponseHeader` to make it easier for RPC users to be able to add HTTP headers. It would also be helpful to have the trailer version of it, especially for gRPC where trailers are actively used.

anuraaga:
It can be useful for RPC implementations, especially gRPC, to be able to specify additional trailers using `ServiceRequestContext`. 

This also updates `additionalResponseHeaders` manipulation to be thread-safe, which I don't think it was before. Compared to the previous implementation, `removeAdditional` always forces creation of headers but I think this is fine given usual usage of the method and makes sure it's thread-safe vs other add operations.

Fixes #1468

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -173,16 +173,7 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                 }
 
                 final HttpHeaders additionalHeaders = reqCtx.additionalResponseHeaders();
-                if (!additionalHeaders.isEmpty()) {
-                    if (headers.isImmutable()) {
-                        // All headers are already validated.
-                        final HttpHeaders temp = headers;
-                        headers = new DefaultHttpHeaders(false, temp.size() + additionalHeaders.size());
-                        headers.set(temp);
-                        o = headers;
-                    }
-                    headers.setAllIfAbsent(additionalHeaders);
-                }
+                o = fillAdditionalHeaders(headers, additionalHeaders);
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -226,6 +219,14 @@ public void onNext(HttpObject o) {
                 return;
         }
 
+        if (endOfStream && !(o instanceof HttpHeaders)) {
+            final HttpHeaders additionalTrailers = reqCtx.additionalResponseTrailers();
+            if (!additionalTrailers.isEmpty()) {
```
trustin: Perhaps this part could be handled better in the `case NEEDS_DATA_OR_TRAILING_HEADERS:` part?

```java
case NEEDS_DATA_OR_TRAILING_HEADERS: {
    if (o instanceof HttpHeaders) {
        ...
    } else if (endOfStream) { // Last DATA frame
        final HttpHeaders additionalTrailers = reqCtx.additionalResponseTrailers();
        if (!additionalTrailers.isEmpty()) {
            write(o, false);
            o = additionalTrailers;
        }
    }
    break;
}
```

anuraaga: Ah yeah

anuraaga:
Yeah that'll be separate. Since gRPC module creates trailers, it can just fill them directly rather than using this mechanism for #1471

trustin:
Thanks, @anuraaga !

PR #1489, opened 2018-12-11
Respect HTTP/2 connection window size

trustin:
Motivation:

`InboundTrafficController` uses a hard-coded water mark value of 128KiB,
which is much smaller than our current initial connection window size
(1MiB).

Modifications:

- Use the water mark values that matches the initial HTTP/2 connection
  window size.
  - HTTP/1 connections' water mark values stay same.
- Disable Netty `Channel` write buffer water mark to reduce unnecessary
  emission of `channelWritabilityChanged` events.

Result:

- Less frequent network read suspension which was unnecessary.
- Less frequent `channelWritabilityChanged` events.


trustin:
Thanks for reviewing.

PR #1492, opened 2018-12-14
Support Tomcat 9.0.14

trustin:
- Updated Tomcat from 9.0.13 to 9.0.14
- Added `ConfigFileLoaderInitializer` which sets
  `ConfigFileLoader.source` property for Tomcat 9+
- Added the methods required to support Tomcat 9.0.14 to `Tomcat90ProtocolHandler`


trustin:
Ready for reviews.

trustin:
Thanks for reviewing.

PR #1467, opened 2018-11-29
Determine a client address from Forwarded, X-Forwarded-For headers an…

hyangtack:
…d PROXY protocol

Modifications:
- Add `InetAddress clientAddress()` method to `ServiceRequestContext` interface.
- Can specify HTTP header names which are used to determine a client address.
- Can specify a remote address filter that evaluates whether a remote endpoint is trusted. Only the requests from the trust endpoint will be used to determine the client address.
- Add `Predicate<InetAddress>` to filter out the `InetAddress` that a user is not interested in.
- Add `server/HttpHeaderUtil` to provide a method which determines a client address from a request.
- Add `[C:x.x.x.x]` to `DefaultServiceRequestContext#toString` if the `#clientAddress`  is different from `#remoteAddress`.
- Add more tokens for access log format:
  - `%a`: client IP address of a request
  - `%{c}a`: remote IP address of the channel
  - `%A`: local IP address of the channel

Result:
- Closes #1077

PR #1494, opened 2018-12-19
Add CIDR based Predicate<InetAddress>

### LINKED ISSUE #1490: Add CIDR based `Predicate<InetAddress>`
Once #1467 is merged, we could add a utility class that provides static factory methods that creates `Predicate<InetAddress>` from a CIDR notation, such as `192.168.0.0/24`. We could perhaps fork `org.jboss.netty.handler.ipfilter.CIDR`.

hyangtack:
Motivation:
Once #1467 is merged, we could add a utility class that provides static factory methods that creates `Predicate<InetAddress>` from a CIDR notation, such as `192.168.0.0/24`.

Modifications:
- Add `InetAddressPredicates` which provides static factory methods that return `Predicate<InetAddress>`.

Result:
- Closes #1490

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
+++ b/core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
@@ -102,7 +102,7 @@ final class Inet4AddressBlock implements Predicate<InetAddress> {
         // The first 10 bytes should be 0.
         for (int i = 0; i < 10; i++) {
             if (addr[i] != 0x00) {
-                throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context" +
+                throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context: " +
                                                    address.getHostAddress());
             }
         }
@@ -118,7 +118,7 @@ final class Inet4AddressBlock implements Predicate<InetAddress> {
             return new byte[] { addr[12], addr[13], addr[14], addr[15] };
         }
 
-        throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context" +
+        throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context: " +
                                            address.getHostAddress());
     }
 
--- a/core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
+++ b/core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.misc;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+
+import java.net.Inet4Address;
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.util.function.Predicate;
+
+import com.google.common.base.MoreObjects;
+
+final class Inet4AddressBlock implements Predicate<InetAddress> {
+
+    private final Inet4Address baseAddress;
+    private final int maskBits;
+    private final int lowerBound;
+    private final int upperBound;
+
+    Inet4AddressBlock(Inet4Address baseAddress, int maskBits) {
+        this.baseAddress = requireNonNull(baseAddress, "baseAddress");
+        checkArgument(maskBits >= 0 && maskBits <= 32,
+                      "maskBits: %s (expected: 0-32)", maskBits);
+        this.maskBits = maskBits;
+
+        // Calculate the lower and upper bounds of this address block.
+        // e.g. mask is 256(0x00000100) if maskBits is 24.
+        final int mask = 1 << 32 - maskBits;
+        // e.g. (-mask) is 0xFFFFFF00 if mask is 256.
+        lowerBound = ipv4AddressToInt(baseAddress.getAddress()) & -mask;
+        // e.g. (mask-1) is 0x000000FF if mask is 256.
+        upperBound = lowerBound + mask - 1;
+    }
+
+    @Override
+    public boolean test(InetAddress address) {
+        requireNonNull(address, "address");
+        if (maskBits == 0) {
+            return true;
+        }
+        final int addr = inetAddressToInt(address);
+        return addr >= lowerBound && addr <= upperBound;
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                          .add("baseAddress", baseAddress)
+                          .add("maskBits", maskBits)
+                          .add("lowerBound", "0x" + Integer.toHexString(lowerBound))
+                          .add("upperBound", "0x" + Integer.toHexString(upperBound))
+                          .toString();
+    }
+
+    /**
+     * Returns the integer representation of the specified {@link InetAddress}.
+     */
+    private static int inetAddressToInt(InetAddress inetAddress) {
+        final byte[] address;
+        if (inetAddress instanceof Inet6Address) {
+            address = ipv6ToIpv4Address((Inet6Address) inetAddress);
+        } else {
+            address = inetAddress.getAddress();
+        }
+        return ipv4AddressToInt(address);
+    }
+
+    /**
+     * Returns IPv4 byte representation of the specified {@link Inet6Address}.
+     *
+     * @throws IllegalArgumentException if the IPv6 cannot be mapped to IPv6
+     */
+    private static byte[] ipv6ToIpv4Address(Inet6Address address) {
+        final byte[] addr = address.getAddress();
+        for (int i = 0; i < 9; i++) {
+            if (addr[i] != 0) {
+                throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context");
+            }
+        }
+        if (addr[10] != 0 && addr[10] != (byte) 0xFF ||
+            addr[11] != 0 && addr[11] != (byte) 0xFF) {
+            throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context");
+        }
+        return new byte[] { addr[12], addr[13], addr[14], addr[15] };
+    }
+
+    /**
+     * Returns the integer representation of the specified {@code address} which is a byte array of
+     * an IPv4 address.
+     */
+    private static int ipv4AddressToInt(byte[] address) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/linecorp/armeria/server/misc/InetAddressPredicates.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /**
+     * Returns a {@link Predicate} which returns {@code true} if the given {@link InetAddress} is in the
+     * range of a <a href="https://tools.ietf.org/html/rfc4632">Classless Inter-domain Routing (CIDR)</a> block.
+     *
+     * @param baseAddress the base {@link InetAddress} of a CIDR notation
+     * @param subnetMask the subnet mask, e.g. {@code 255.255.255.0}
+     */
+    public static Predicate<InetAddress> ofCidr(InetAddress baseAddress, String subnetMask) {
```
hyangtack: I added this method because [jboss.netty.handler.ipfilter.CIDR](https://github.com/netty/netty/blob/3.10/src/main/java/org/jboss/netty/handler/ipfilter/CIDR.java#L60) provides it. But I'm curious that is it necessary for us?
@trustin @minwoox Please let me know what you think! :-)

core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
```diff
@@ -0,0 +1,114 @@
...
+        } else {
+            address = inetAddress.getAddress();
+        }
+        return ipv4AddressToInt(address);
+    }
+
+    /**
+     * Returns IPv4 byte representation of the specified {@link Inet6Address}.
+     *
+     * @throws IllegalArgumentException if the IPv6 cannot be mapped to IPv6
```
minwoox: to IPv4?

core/src/main/java/com/linecorp/armeria/server/misc/InetAddressPredicates.java
minwoox: I think it's good to have this one. There should be someone who prefers this representation to bits.

trustin: :+1:

core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Returns IPv4 byte representation of the specified {@link Inet6Address}.
+     *
+     * @throws IllegalArgumentException if the IPv6 cannot be mapped to IPv4
+     */
+    private static byte[] ipv6ToIpv4Address(Inet6Address address) {
+        final byte[] addr = address.getAddress();
+        for (int i = 0; i < 9; i++) {
+            if (addr[i] != 0) {
+                throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context");
```
trustin: This will `test()` raise an exception rather than returning `false`.

core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+        if (addr[10] != 0 && addr[10] != (byte) 0xFF ||
+            addr[11] != 0 && addr[11] != (byte) 0xFF) {
+            throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context");
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/server/misc/InetAddressPredicates.java
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Returns a {@link Predicate} which returns {@code true} if the given {@link InetAddress} equals to
+     * the specified {@code address}.
+     *
+     * @param address the expected {@link InetAddress}
+     */
+    public static Predicate<InetAddress> ofExact(InetAddress address) {
+        requireNonNull(address, "address");
+        return address::equals;
```
trustin: I guess we at least should perform some conversion between IPv4/IPv6 addresses? e.g.

- What if a user called this method with an IPv6 address which actually contains an IPv4 address?
- What if a user called this method with an IPv4 address and he or she called the returned predicate with an IPv6 address which actually contains an IPv4 address?

Perhaps you might want to reused some of the code I put in `DnsAddressEndpointGroup`: https://github.com/line/armeria/blob/b971e19d763cab0ace19a53cf0ea575576be58c4/core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java#L162-L195

core/src/main/java/com/linecorp/armeria/server/misc/InetAddressPredicates.java
```diff
@@ -0,0 +1,199 @@
...
+     * range of a <a href="https://tools.ietf.org/html/rfc4632">Classless Inter-domain Routing (CIDR)</a> block.
+     *
+     * @param baseAddress the base {@link InetAddress} of a CIDR notation
+     * @param subnetMask the subnet mask, e.g. {@code 255.255.255.0}
+     */
+    public static Predicate<InetAddress> ofCidr(InetAddress baseAddress, String subnetMask) {
+        requireNonNull(baseAddress, "baseAddress");
+        requireNonNull(subnetMask, "subnetMask");
+        checkArgument(validIpV4Address.matcher(subnetMask).matches(),
+                      "subnetMask: %s (expected: IP v4 address string)", subnetMask);
```
trustin: IP v4 address string -> an IPv4 address string

core/src/main/java/com/linecorp/armeria/server/misc/Inet6AddressBlock.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Returns the {@link BigInteger} representation of the specified {@link InetAddress}.
+     */
+    private static BigInteger ipv6AddressToBigInteger(InetAddress inetAddress) {
+        final byte[] ipv6;
+        if (inetAddress instanceof Inet4Address) {
+            final byte[] ipv4 = inetAddress.getAddress();
+            // Fill the lower 4 bytes with the IPv4 address.
+            ipv6 = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ipv4[0], ipv4[1], ipv4[2], ipv4[3] };
```
trustin: Need to handle loopback addresses as well.

core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+        return new byte[] { addr[12], addr[13], addr[14], addr[15] };
```
trustin: Need to handle loopback address as well. Otherwise it will be translated into `0.0.0.1` rather than `127.0.0.1`. See: https://github.com/line/armeria/blob/b971e19d763cab0ace19a53cf0ea575576be58c4/core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java#L162-L195

core/src/main/java/com/linecorp/armeria/server/misc/Inet6AddressBlock.java
```diff
@@ -73,8 +85,13 @@ private static BigInteger ipv6AddressToBigInteger(InetAddress inetAddress) {
         final byte[] ipv6;
         if (inetAddress instanceof Inet4Address) {
             final byte[] ipv4 = inetAddress.getAddress();
-            // Fill the lower 4 bytes with the IPv4 address.
-            ipv6 = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ipv4[0], ipv4[1], ipv4[2], ipv4[3] };
+            if (ipv4[0] == 0x7F && ipv4[1] == 0x00 && ipv4[2] == 0x00 && ipv4[3] == 0x01) {
```
trustin: `ipv4[3]` can be any number actually, e.g. `127.0.0.2` and `127.0.0.3` are also localhost and thus needs to be translated into `::1`.

core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
```diff
@@ -0,0 +1,137 @@
...
+     * Returns IPv4 byte representation of the specified {@link Inet6Address}.
+     *
+     * @throws IllegalArgumentException if the IPv6 cannot be mapped to IPv4
+     */
+    private static byte[] ipv6ToIpv4Address(Inet6Address address) {
+        final byte[] addr = address.getAddress();
+        // The first 10 bytes should be 0.
+        for (int i = 0; i < 10; i++) {
+            if (addr[i] != 0x00) {
+                throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context" +
```
minwoox: `context: "`? so that the message is read easily?

core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
```diff
@@ -0,0 +1,137 @@
...
+                return localhost;
+            }
+            return new byte[] { addr[12], addr[13], addr[14], addr[15] };
+        }
+
+        if (addr[10] == (byte) 0xFF && addr[11] == (byte) 0xFF) {
+            return new byte[] { addr[12], addr[13], addr[14], addr[15] };
+        }
+
+        throw new IllegalArgumentException("This IPv6 address cannot be used in IPv4 context" +
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/server/misc/Inet6AddressBlock.java
hyangtack: Thanks! You're right. :-)

From https://tools.ietf.org/html/rfc5735#section-3:
> 127.0.0.0/8 - This block is assigned for use as the Internet host
   loopback address.  A datagram sent by a higher-level protocol to an
   address anywhere within this block loops back inside the host.  This
   is ordinarily implemented using only 127.0.0.1/32 for loopback.  As
   described in [RFC1122], Section 3.2.1.3, addresses within the entire
   127.0.0.0/8 block do not legitimately appear on any network anywhere.

It seems to be better if we check the first byte only. Please let me know if you have any concern.

core/src/main/java/com/linecorp/armeria/server/misc/Inet4AddressBlock.java
hyangtack: I missed that. Thanks!

core/src/main/java/com/linecorp/armeria/server/misc/InetAddressPredicates.java
```diff
@@ -0,0 +1,216 @@
...
+ * {@link InetAddress}.
+ */
+public final class InetAddressPredicates {
+
+    /**
+     * A {@link Pattern} of a valid IP address.
+     */
+    private static final Pattern validIpV4Address =
+            Pattern.compile("^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}" +
+                            "(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$");
```
trustin: You can use `NetUtil.isValidIpV4Address()` in Netty.

hyangtack:
Thanks, reviewers!

PR #1499, opened 2018-12-21
Send 'content-length' header for an empty PUT/POST/PATCH request

### LINKED ISSUE #1476: Armeria removes ContentLength header when performing PUT requests without content
When performing a PUT http request without content and with Content-Length=0 with armeria, the Content-Length header gets removed [here](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/Http1ObjectEncoder.java#L247). This was added in #226 which speaks of GET requests, but this logic is being applied to all request methods, including PUT, POST, etc., which are supposed to have content.

trustin:
Motivation:

Although not a protocol violation, Armeria does not send a
`content-length` header when a client sends an empty PUT/POST/PATCH
request, which may confuse some server implementations or users.

Modifications:

- Make sure that an empty PUT/POST/PATCH request always has the
  `content-length: 0` header.

Result:

- Fixes #1476


sschepens:
Thanks!

PR #1500, opened 2018-12-21
Revert 'null' to 'Function.identity()' which was fixed by #1478

hyangtack:
Motivation:
It was unnecessary because we are comparing the original service instance with decorators-applied service instance when deciding to add an exception handling decorator.

Modifications:
- Revert `null` to `Function.identity()` in `AbstractVirtualHostBuilder`.
- Add a test case for `ArmeriaConfigurationUtil#configureAnnotatedHttpServices`. (related to #1497)

### DIFF
--- a/spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtilTest.java
+++ b/spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtilTest.java
@@ -32,12 +32,13 @@ import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.internal.annotation.AnnotatedHttpService;
-import com.linecorp.armeria.server.AbstractHttpService;
 import com.linecorp.armeria.server.Server;
 import com.linecorp.armeria.server.ServerBuilder;
 import com.linecorp.armeria.server.Service;
 import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.SimpleDecoratingService;
 import com.linecorp.armeria.server.annotation.Get;
+import com.linecorp.armeria.server.metric.MetricCollectingService;
 import com.linecorp.armeria.spring.AnnotatedServiceRegistrationBean;
 import com.linecorp.armeria.spring.MeterIdPrefixFunctionFactory;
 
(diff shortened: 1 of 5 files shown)


spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtilTest.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * A decorator function which is the same as {@link #identity()} but is not a final class.
+     */
+    static class DecoratorFunction
+            implements Function<Service<HttpRequest, HttpResponse>, Service<HttpRequest, HttpResponse>> {
+        @Override
+        public Service<HttpRequest, HttpResponse> apply(Service<HttpRequest, HttpResponse> delegate) {
+            return delegate;
```
trustin: Could you also add the test case where a decorator function actually returns a decorated instance?

hyangtack: Thanks! Added.

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtilTest.java
```diff
@@ -0,0 +1,128 @@
...
+                .setServiceName("test")
+                .setService(new SimpleService())
+                .setDecorators(decorator);
+
+        final ServerBuilder sb1 = new ServerBuilder();
+        configureAnnotatedHttpServices(sb1, ImmutableList.of(bean), MeterIdPrefixFunctionFactory.DEFAULT);
+        verify(decorator).apply(any());
+        assertThat(service(sb1)).isNotInstanceOf(AnnotatedHttpService.class)
+                                .satisfies(s -> assertThat(s.getClass().getName())
+                                        .endsWith("$ExceptionFilteredHttpResponseDecorator"));
```
trustin: Could we:

- extract these three lines into a method to remove the duplicate assertion in `makesSureDecoratedServiceIsAdded()`?
- add a brief comment why an `AnnotatedHttpService` is decorated with `ExceptionFilteredHttpResponseDecorator` to the extracted method?
- add an extra assertion that checks if `DecoratingFunction` actually decorated `AnnotatedHttpService` to the extracted method? (To do this, you will have to define a subtype of `SimpleDecoratingService` instead of using an anonymous class.)
  - You can use `Service.as(Class)` method to check if a decorator exists in a service:
    ```java
    Service s = ...;
    assertThat(s.as(SomeDecoratorService.class)).isPresent();
    ```

hyangtack: Thanks! I totally forgot the 3rd option! 😱

hyangtack:
Thanks, reviewers!

PR #1501, opened 2018-12-21
Fix to support wildcard type parameter in DocService

minwoox:
Motivation:
The return type of a annotated HTTP method could be `CompletableFuture<?>`.

Modifications:
- Support wildcard type parameter

Result:
- Less bugs


core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpDocServicePlugin.java
```diff
@@ -77,12 +81,16 @@
     private static final String QUERY_PARAM = "query";
     private static final String HEADER_PARAM = "header";
 
+    // The formats defined by OpenAPI Specification
+    // (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#dataTypes):
     @VisibleForTesting
     static final TypeSignature VOID = TypeSignature.ofBase("void");
     @VisibleForTesting
     static final TypeSignature BOOL = TypeSignature.ofBase("boolean");
+    // Not defined in the spec, but added to support char and byte types.
```
trustin: Isn't `char` a 16-bit unsigned integer, which is not 8-bit?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpDocServiceTest.java
```diff
@@ -117,35 +127,50 @@ private static void addFooMethodInfo(Map<Class<?>, Set<MethodInfo>> methodInfos)
...
                 .availableMimeTypes(MediaType.JSON_UTF_8).build();
         Stream.of(HttpMethod.values())
               .filter(httpMethod -> httpMethod != HttpMethod.CONNECT && httpMethod != HttpMethod.UNKNOWN)
               .forEach(httpMethod -> {
                   final MethodInfo methodInfo =
-                          new MethodInfo("fooAll", toTypeSignature(String.class), ImmutableList.of(),
-                                         ImmutableList.of(), ImmutableList.of(endpoint), httpMethod, null);
+                          new MethodInfo("allMethods", TypeSignature.ofContainer("CompletableFuture",
+                                                                                 TypeSignature
+                                                                                         .ofUnresolved("")),
```
trustin: Could break after `"allMethods"` to make more space for `TypeSignature.ofContainer()`.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpDocServicePlugin.java
minwoox: Oops~ Thanks~

PR #1506, opened 2018-12-28
Make services in DocService collapsible

minwoox:
Motivation:
When there are many services to show in `DocService`, it's not easy to see the
main content page because when the scroll goes down the main content does not follow.
So if we can collapse the services, it will me much easier to look at the main content.

Modifications:

- Make services collapsible
- Make the `AppDrawer` scrollable

Result:

- Better UI

### DIFF
--- a/docs-client/src/containers/App/index.tsx
+++ b/docs-client/src/containers/App/index.tsx
@@ -23,7 +23,6 @@ import IconButton from '@material-ui/core/IconButton';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
-import ListSubheader from '@material-ui/core/ListSubheader';
 import {
   createStyles,
   Theme,
@@ -69,6 +68,8 @@ const styles = (theme: Theme) =>
       [theme.breakpoints.down('sm')]: {
         width: '80%',
       },
+      height: '100vh',
+      overflowY: 'auto',
     },
     content: {
       backgroundColor: theme.palette.background.default,
@@ -125,6 +126,7 @@ interface State {
   mobileDrawerOpen: boolean;
   specification?: Specification;
   servicesOpen: boolean;
+  serviceOpens: { [key: string]: boolean };
   enumsOpen: boolean;
   structsOpen: boolean;
   exceptionsOpen: boolean;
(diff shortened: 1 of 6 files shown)


minwoox:
screencapture-localhost-3000-docs-2018-12-28-11_23_25

docs-client/src/containers/App/index.tsx
```diff
@@ -309,13 +322,20 @@ class App extends React.PureComponent<Props, State> {
...
+    serviceOpens: {},
     enumsOpen: true,
     structsOpen: true,
     exceptionsOpen: true,
   };
 
   public componentWillMount() {
-    this.fetchSpecification();
+    this.fetchSpecification().then((specification) => {
+      let serviceOpens = {};
```
anuraaga: Would it be simpler to do this inside `fetchSpecification` which is an async method?

minwoox: Yes, it would.
Let me move the codes inside `fetchSpecification`.
I just thought that I had to change the name `fetchSpecification` to something else because it does one more job inside the method and I couldn't figure out the better name. :)

docs-client/src/containers/App/index.tsx
```diff
@@ -477,12 +496,15 @@ class App extends React.PureComponent<Props, State> {
...
+  private initSpecification = async () => {
     const httpResponse = await fetch('specification.json');
-    const specification: SpecificationData = await httpResponse.json();
-    this.setState({
-      specification: new Specification(specification),
+    const specificationData: SpecificationData = await httpResponse.json();
+    const specification = new Specification(specificationData);
+    let openServices = {};
+    specification.getServices().forEach((service) => {
+      openServices = { ...openServices, ...{ [service.name]: true } };
```
anuraaga: I think this works

```ts
openServices = { ...openServices, [service.name]: true };
```

Or also can use reduce.

```ts
const openServices = specification.getServices().reduce((services, service) => (
  { ...services, [service.name]: true }), {});
```

minwoox: Didn't know that I can do it in that way. Fixed, thank you~!

trustin:
@minwoox Could you squash the commits and update the commit message?

minwoox:
Done! Thank you~!

minwoox:
Thanks for reviewing :)

PR #1508, opened 2018-12-28
Add ServerConfig.verboseResponses()

### LINKED ISSUE #1507: Add `ServerBuilder.verboseResponses(boolean)`
.. so that a user can override the value of `Flags.verboseResponses()` programmatically.

trustin:
Motivation:

`verboseResponse` flag could be controlled programmatically when
creating a `Server` because a user may not want to rely on a system
property.

Modifications:

- Added `verboseResponses` property to `ServerBuilder` and `ServerConfig`
- Removed the stack trace from the `debugData` field of a `GOAWAY` frame
  and logged the connection error exceptions explicitly instead for
  easier protocol violation diagnosis.
  - As a result, `GOAWAY` frame logging does not rely on `verboseResponses`
    anymore.

Result:

- Closes #1507


trustin:
Thanks for reviewing.

PR #1509, opened 2018-12-28
Provide a way to return a successful response with alternative status…

hyangtack:
… code

Motivation:
Currently there is no way to return a response with alternative status code except that a user creates an `HttpResponse` instance by himself or herself. In that case, however, he or she cannot use response converters which are very handy function. So it would be better to provide a way to use response converters with alternative status code.

Modifications:
- (Breaking) Change the method signature of `ResponseConverterFunction`.
- Add an `HttpResult` interface to specify a user-specific status code, headers and body.
- Add a `@StatusCode` annotation to provide a way to specify alternative status code when a user is not using `HttpResult`.
```
@Get("/hello")
@StatusCode(201)
public String hello() {
    return "Hello, world!";
}
```
- (Breaking) Change the behavior when `null` object is returned from an annotated HTTP service method.
  In that case, `204 No Content` has been returned before, but now it depends on the `@StatusCode` annotation and the return type of the method.
  - If `@StatusCode` annotation is specified, use the status code by default.
  - If `@StatusCode` annotation is not specified, the default status code is to be:
    - `204 No Content` only if the return type of the method is `void` or `Void`
    - `200 Ok` for the other return types
```
@Get("/hello1")
// Implicitly specifies '@StatusCode(204)' because the return type is 'void'.
public void hello1() {
	// Will send '204 No Content' response with empty body.
    return;
}

@Get("/hello2")
@StatusCode(200)
public void hello2() {
	// Will send '200 Ok' response with empty body.
    return;
}

@Get("/hello3")
// Implicitly specifies '@StatusCode(200)' because the return type is not a 'void' or 'Void'.
public Object hello3() {
	// Will send '200 Ok' response with empty body.
    return null;
}

@Get("/hello4")
@StatusCode(204)
public Object hello4() {
	// Will send '204 No Content' response with empty body.
    return null;
}

@Get("/hello5")
@ResponseConverter(NullToNoContentResponseConverterFunction.class)
public Object hello5() {
	// Will send '204 No Content' response with empty body.
    return null;
}
```


core/src/main/java/com/linecorp/armeria/server/annotation/HttpResult.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * An interface which helps a user specify an {@link HttpStatus} and {@link HttpHeaders} for a response
+ * produced by an annotated HTTP service method. The HTTP content can be specified as {@code body} as well,
+ * and it would be converted into {@link HttpData} instances by a {@link ResponseConverterFunction}.
+ *
+ * @param <T> the type of a body which is to be contained in an {@link HttpResponse}
+ */
+@FunctionalInterface
+public interface HttpResult<T> {
```
trustin: How about adding some static factory methods which are shortcuts to `HttpResultBuilder`? e.g.

```java
static HttpResult<T> of(HttpStatus status, T body) { ... }
// ... add more methods like we have in `HttpResponse` ...
```

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceUtil.java
```diff
@@ -0,0 +1,39 @@
...
+     * Converts the specified {@link Type} to a {@link Class} instance.
+     */
+    public static Class<?> typeToClass(Type type) {
+        if (type instanceof Class) {
+            return (Class<?>) type;
+        }
+        if (type instanceof ParameterizedType) {
+            return (Class<?>) ((ParameterizedType) type).getRawType();
+        }
+        return Void.class;
```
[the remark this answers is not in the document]
hyangtack: As we talked in LINE chat, we don't need to handle that kinds of types. And this method has been moved to rxjava module. 😄

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonResponseConverterFunction.java
```diff
@@ -67,13 +70,17 @@ public HttpResponse convertResponse(ServiceRequestContext ctx,
...
                 }
             }
         } else if (result instanceof JsonNode) {
             // No media type is specified, but the result is a JsonNode type.
-            return HttpResponse.of(HttpStatus.OK, MediaType.JSON_UTF_8, mapper.writeValueAsBytes(result));
+            final HttpHeaders responseHeaders =
+                    headers.isImmutable() ? HttpHeaders.copyOf(headers).contentType(MediaType.JSON_UTF_8)
+                                          : headers.contentType(MediaType.JSON_UTF_8);
+            final HttpData body = HttpData.of(mapper.writeValueAsBytes(result));
+            return HttpResponse.of(responseHeaders, body, trailingHeaders);
```
trustin: I see some inconsistency in variable/parameter naming - `body` vs. `content`.

core/src/main/java/com/linecorp/armeria/server/annotation/NullToNoContentResponseConverterFunction.java
```diff
@@ -0,0 +1,43 @@
...
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+/**
+ * A response converter implementation which creates an {@link HttpResponse} of {@link HttpStatus#NO_CONTENT}
+ * when the specified {@code result} is {@code null}.
+ */
+public class NullToNoContentResponseConverterFunction implements ResponseConverterFunction {
```
trustin: `final`?

core/src/main/java/com/linecorp/armeria/server/annotation/StatusCode.java
```diff
@@ -0,0 +1,34 @@
...
+package com.linecorp.armeria.server.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * An annotation which specifies a default HTTP status code of a response produced by an annotated HTTP service.
+ * If this annotation is missing, {@code @StatusCode(200)} would be applied by default.
```
trustin: Could we also mention that 204 is used when the return type is `void`?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
```diff
@@ -168,6 +181,111 @@ public HttpData httpData() {
...
+                @Get("/expect-ok")
+                public Object expectOk() {
+                    // Will send '200 Ok' because there is no @StatusCode annotation, that means
+                    // '200 Ok' will be used by default because the return type is not a 'void' or 'Void'.
+                    return null;
+                }
+
+                @Get("/expect-specified-no-content")
+                @StatusCode(204)
+                public HttpResult<Object> expectSpecifiedNoContent() {
```
trustin: I'm curious what would happen if a user defines methods like:

```java
public HttpResult<?> foo() { ... }
public <T> HttpResult<T> bar() { ... }
```

Will they work as expected?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -104,6 +104,23 @@ a path and use the HTTP method annotations without a path to map multiple HTTP m
         public HttpResponse hello() { ... }
     }
 
+Every service method assumes that they returns an HTTP response with ``200 Ok`` status. So If you want to
```
trustin: So If -> If

site/src/sphinx/server-annotated-service.rst
```diff
@@ -104,6 +104,23 @@ a path and use the HTTP method annotations without a path to map multiple HTTP m
...
+return another status for a specific method, you can use :api:`@StatusCode` annotation as follows.
```
trustin: return an alternative status code for a method,

site/src/sphinx/server-annotated-service.rst
```diff
@@ -104,6 +104,23 @@ a path and use the HTTP method annotations without a path to map multiple HTTP m
...
+
+.. code-block:: java
+
+    public class MyAnnotatedService {
+
+        @Get("/user/{name}")
+        public HttpResponse getUser(@Param("name") String name) { ... }
```
trustin: This method could be removed?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -762,6 +788,12 @@ more response types which can be used in the annotated service.
     being sent, exception handlers will handle it. If no message has been sent to the client yet,
     the exception handler can send an :api:`HttpResponse` instead.
 
+- :api:`HttpResult`
+
+  - It contains the :api:`HttpHeaders` and the object which can be converted into :api:`HttpData` by
```
trustin: converted into HTTP response body by

site/src/sphinx/server-annotated-service.rst
```diff
@@ -762,6 +788,12 @@ more response types which can be used in the annotated service.
...
+    response converters. A user can customize the HTTP status and headers including the trailing headers,
+    with this type.
```
trustin: How about adding a small example here?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
hyangtack: Added methods for those cases. Thanks! :-)

site/src/sphinx/server-annotated-service.rst
hyangtack: Added! Thanks!

core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayResponseConverterFunction.java
```diff
@@ -31,29 +31,37 @@
...
-    private static MediaType mediaType(@Nullable MediaType mediaType) {
-        if (mediaType == null) {
-            return MediaType.APPLICATION_BINARY;
+    private static HttpResponse response(HttpHeaders headers, byte[] body,
+                                         HttpHeaders trailingHeaders) {
+        final MediaType contentType = headers.contentType();
+        if (contentType == null) {
+            final HttpHeaders responseHeaders =
+                    headers.isImmutable() ? HttpHeaders.copyOf(headers)
+                                                       .contentType(MediaType.APPLICATION_BINARY)
```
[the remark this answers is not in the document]
hyangtack: Changed. Thanks!

core/src/main/java/com/linecorp/armeria/server/annotation/HttpResult.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * Creates a new {@link HttpResult} with the specified {@code content}, the {@link HttpStatus#OK} status
+     * and the specified trailing {@link HttpHeaders}.
+     *
+     * @param content the content of the response
+     * @param trailingHeaders the trailing HTTP headers
+     */
+    static <T> HttpResult<T> of(T content, HttpHeaders trailingHeaders) {
```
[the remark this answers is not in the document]
hyangtack: Removed.

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonResponseConverterFunction.java
hyangtack: Oh, I missed fixing here. Thanks!

site/src/sphinx/server-annotated-service.rst
hyangtack: Added more examples. Could you check them again? Then, please let me know if it   still looks unnecessary. :-)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -104,21 +104,27 @@ a path and use the HTTP method annotations without a path to map multiple HTTP m
         public HttpResponse hello() { ... }
     }
 
-Every service method assumes that they returns an HTTP response with ``200 Ok`` status. So If you want to
-return another status for a specific method, you can use :api:`@StatusCode` annotation as follows.
+Every service method assumes that they returns an HTTP response with ``200 Ok`` or ``204 No Content`` status
+according to its return type. If the return type is ``void`` or ``Void``, ``204 No Content`` would be applied.
+``200 Ok`` would be applied for the other types. If you want to return an alternative status code for a method,
```
trustin: nit: Ok -> OK

core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayResponseConverterFunction.java
```diff
@@ -31,29 +31,37 @@
...
-        if (mediaType == null) {
-            return MediaType.APPLICATION_BINARY;
+    private static HttpResponse response(HttpHeaders headers, byte[] content,
+                                         HttpHeaders trailingHeaders) {
+        final MediaType contentType = headers.contentType();
+        if (contentType == null) {
+            final HttpHeaders responseHeaders =
+                    headers.isImmutable() ? HttpHeaders.copyOf(headers)
+                                                       .contentType(MediaType.OCTET_STREAM)
+                                          : headers;
```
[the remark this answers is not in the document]
hyangtack: Oops! I missed that. Thanks! 👍

trustin:
Great job, @hyangtack :-)

PR #1511, opened 2018-12-31
Specialize media type matching for a single range

anuraaga:
While looking at code, I found this low hanging fruit. Matching media types is much simpler for a single one than multiple, so specializing the code for a single argument improves performance of matching. I found most callers to be matching against a single format (`GrpcService`, `THttpService`) so it's some free (micro-) performance.

After
```
Benchmark                             Mode  Cnt         Score          Error  Units
MediaTypesBenchmark.all              thrpt    5  10201031.774 ±   334154.181  ops/s
MediaTypesBenchmark.complexMatch     thrpt    5  28234112.782 ±  4971845.081  ops/s
MediaTypesBenchmark.complexNotMatch  thrpt    5  25309947.033 ±  2968303.292  ops/s
MediaTypesBenchmark.simpleMatch      thrpt    5  93640157.481 ± 12225237.586  ops/s
MediaTypesBenchmark.simpleNotMatch   thrpt    5  79549461.487 ± 12319334.619  ops/s
```

Before
```
Benchmark                             Mode  Cnt         Score         Error  Units
MediaTypesBenchmark.all              thrpt    5   7383325.445 ±  204680.181  ops/s
MediaTypesBenchmark.complexMatch     thrpt    5  23564626.484 ±  418400.507  ops/s
MediaTypesBenchmark.complexNotMatch  thrpt    5  21315281.225 ± 2725154.281  ops/s
MediaTypesBenchmark.simpleMatch      thrpt    5  46150165.428 ± 3222182.395  ops/s
MediaTypesBenchmark.simpleNotMatch   thrpt    5  54324601.873 ± 1090724.032  ops/s
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/MediaTypeSet.java
+++ b/core/src/main/java/com/linecorp/armeria/common/MediaTypeSet.java
@@ -21,7 +21,6 @@ import static java.util.Objects.requireNonNull;
 
 import java.util.AbstractSet;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -31,11 +30,12 @@ import java.util.Set;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
+import com.google.common.collect.Lists;
 
 /**
  * An immutable {@link Set} of {@link MediaType}s which provides useful methods for content negotiation.
  *
- * <p>This {@link Set} provides {@link #match(MediaType...)} and {@link #matchHeaders(CharSequence...)}
+ * <p>This {@link Set} provides {@link #match(MediaType)} and {@link #matchHeaders(CharSequence...)}
  * so that a user can find the preferred {@link MediaType} that matches the specified media ranges. For example:
  * <pre>{@code
  * MediaTypeSet set = new MediaTypeSet(MediaType.HTML_UTF_8, MediaType.PLAIN_TEXT_UTF_8);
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/common/MediaTypeSet.java
```diff
@@ -177,23 +197,23 @@ public int size() {
         int matchNumParams = Integer.MIN_VALUE;    // higher = better
         for (MediaType range : ranges) {
             requireNonNull(range, "ranges contains null.");
+            float qValue = range.qualityFactor(Float.NEGATIVE_INFINITY);
```
anuraaga: If none of the candidates match, these few lines are wasteful. If all of them do, though it's better to only run them once. I figured it's more likely to match than not match and rearranged this code a little. WDYT?

trustin: Sounds reasonable.

core/src/main/java/com/linecorp/armeria/common/MediaTypeSet.java
```diff
@@ -150,15 +150,35 @@ public int size() {
...
+            }
+        }
+        return Optional.empty();
+    }
+
     /**
      * Finds the {@link MediaType} in this {@link List} that matches one of the specified media ranges.
      *
      * @return the most preferred {@link MediaType} that matches one of the specified media ranges.
      *         {@link Optional#empty()} if there are no matches or {@code ranges} does not contain
```
minwoox: Don't we need to update this doc as well because there's no `ranges` anymore?

trustin: @anuraaga Could you update the Javadoc?

anuraaga:
Thanks - updated

PR #1513, opened 2019-01-03
Do not return a connection to the pool if it's scheduled for disconnection

trustin:
Motivation:

It is currently possible to return a connection to the pool even if it's
scheduled for disconnection, e.g. a connection which received a response
with a `connection: close` header or a `GOAWAY` frame.

As a result, it is possible to acquire a connection which will be closed
very soon, causing an unexpected request failure.

Modifications:

- Renamed `HttpSession.isActive()` to `canSendRequest()` to make its
  meaning clearer.
- `HttpSession.canSendRequest()` will return `false` if disconnection
  has been scheduled, e.g. received a `connection: close` header or
  sent/received a `GOAWAY` frame.
- Renamed `HttpResponseDecoder.needsToDisconnect()` to `needsToDisconnectNow()`
  for clarity.
- Added `HttpResponseDecoder.needsToDisconnectWhenFinished()`, which is
  used to tell if the connection has been scheduled for disconnection.
- Do not close the connection immediately when returning an unhealthy
  connection, because it may be handling unfinished requests.
  - The connection will be closed anyway eventually after all unfinished
    requests are handled.
- Added `Http1ConnectionCloseHeaderTest` to ensure that Armeria client
  respects the `connection: close` header.

Result:

- Less chance of getting a bad connection from the pool.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
@@ -591,10 +591,10 @@ final class HttpChannelPool implements AutoCloseable {
         @Override
         public void release() {
             if (isHealthy(this)) {
-                // Channel turns out to be healthy, offering and releasing it.
+                // Channel turns out to be healthy. Add it back to the pool.
                 addToPool(protocol(), key, this);
             } else {
-                // Channel not healthy, just releasing it.
+                // Channel not healthy. Do not add it back to the pool.
             }
         }
     }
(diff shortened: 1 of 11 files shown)


core/src/test/java/com/linecorp/armeria/client/Http1ConnectionCloseHeaderTest.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2018 LINE Corporation
```
minwoox: Happy new year~

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -598,10 +595,6 @@ public void release() {
                 addToPool(protocol(), key, this);
             } else {
                 // Channel not healthy, just releasing it.
```
minwoox: Could modify the comment as well because there's no releasing code in this `else` block?

trustin:
Thanks for reviewing.

PR #1516, opened 2019-01-04
Make HttpHeaderNames.of() accept CharSequence in lieu of String

trustin:
Motivation:

By allowing a user to specify `AsciiString` or other `CharSequence`
implementations, we can perform more HTTP header name normalization,
e.g.

    AsciiString contentType = AsciiString.of("Content-Type");
    assert HttpHeaderNames.CONTENT_TYPE != contentType;
    AsciiString normContentType = HttpHeaderNames.of(contentType);
    assert HttpHeaderNames.CONTENT_TYPE == normContentType;

Modifications:

- Change the parameter type of `HttpHeaderNames.of()` from `String` to
  `CharSequence` and add fast path for the case of `AsciiString`.
- Use `HttpHeaderNames.of()` instead of `AsciiString.of()` wherever
  possible.
- Use `AsciiString.cached()` instead of `AsciiString.of()` when the type
  of the string is actually a `String`.
- Make `ClientBuilder.add/setHttpHeader()` accept `CharSequence` as a
  header name.

Result:

- Efficiency

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/client/endpoint/StickyEndpointSelectionStrategyTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/endpoint/StickyEndpointSelectionStrategyTest.java
@@ -28,20 +28,21 @@ import org.junit.Test;
 
 import com.linecorp.armeria.client.ClientRequestContext;
 import com.linecorp.armeria.client.Endpoint;
+import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.HttpRequest;
-
-import io.netty.util.AsciiString;
 
 public class StickyEndpointSelectionStrategyTest {
 
     private static final String STICKY_HEADER_NAME = "USER_COOKIE";
 
-    final ToLongFunction<ClientRequestContext> hasher =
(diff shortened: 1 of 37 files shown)


trustin:
TODO: Add a test case.

hyangtack:
LGTM so far. Let me review again after test case is added. :-)

trustin:
Ready for a review.

core/src/test/java/com/linecorp/armeria/client/endpoint/StickyEndpointSelectionStrategyTest.java
```diff
@@ -28,20 +28,21 @@
...
 
     private static final String STICKY_HEADER_NAME = "USER_COOKIE";
 
-    final ToLongFunction<ClientRequestContext> hasher =
-            (ClientRequestContext ctx) -> ((HttpRequest) ctx.request()).headers()
-                                                                       .get(AsciiString.of(STICKY_HEADER_NAME))
-                                                                       .hashCode();
+    final ToLongFunction<ClientRequestContext> hasher = (ClientRequestContext ctx) -> {
+        return ((HttpRequest) ctx.request()).headers()
+                                     .get(HttpHeaderNames.of(STICKY_HEADER_NAME))
```
minwoox: nit: indentation?

trustin: Fixed :-)

trustin:
Thanks for reviewing.

PR #1519, opened 2019-01-08
Add an example which uses Armeria with the minimal Spring Boot depend…

hyangtack:
…encies

Modifications:
- Add `spring-boot-minimal` example.
- Move `example.springframework.boot.webflux.HelloAnnotatedService` to `spring-boot-minimal` example.


adriancole:
thank you.. much better than my unmerged cruft locally :)

PR #1525, opened 2019-01-11
Add option to run gRPC callbacks in blocking executor.

### LINKED ISSUE #1524: Why armeria run all grpc service logic directly in the event loop?
I have plan to convert all my grpc service to armeria-grpc but i have an issue, armeria execute service logic directly in event loop and i should handle blocking part manually, this take a lot of from me.
Now want to now what is your reason for using event loop for execute service logic?

Thanks and sorry for my bad english.

anuraaga:
Fixes #1524

### DIFF
--- a/site/src/sphinx/server-grpc.rst
+++ b/site/src/sphinx/server-grpc.rst
@@ -176,9 +176,11 @@ for JSON POST body.
 Blocking service implementation
 -------------------------------
 
-Unlike upstream gRPC-java, Armeria does not run service logic in a separate thread pool. If your service
-implementation requires blocking, either run the individual blocking logic in a thread pool, or just wrap the
-entire service implementation in ``RequestContext.current().blockingTaskExecutor().submit``
+Unlike upstream gRPC-java, Armeria does not run service logic in a separate thread pool by default. If your
+service implementation requires blocking, either run the individual blocking logic in a thread pool, wrap the
+entire service implementation in ``RequestContext.current().blockingTaskExecutor().submit``, or set
+``GrpcServiceBuilder.useBlockingExecutor()`` so the above happens automatically for all service methods
+and lifecycle callbacks.
 
 .. code-block:: java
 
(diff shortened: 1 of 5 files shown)


trustin:
How about updating `.rst` files as well?

anuraaga:
Yeah - updated

site/src/sphinx/server-grpc.rst
```diff
@@ -200,6 +202,29 @@ entire service implementation in ``RequestContext.current().blockingTaskExecutor
...
+            HelloReply reply = HelloReply.newBuilder()
+                                         .setMessage("Hello, " + req.getName() + '!')
+                                         .build();
+            responseObserver.onNext(reply);
+            responseObserver.onCompleted();
+        }
+
+        public static void configureService(GrpcServiceBuilder builder) {
+            // All service methods will be run within the blocking executor.
+            builder.useBlockingExecutor(true);
```
trustin: Would you mind updating this example like this:

```java
public class MyHelloService ... {
   ...
}

sb.service(new GrpcServiceBuilder().addService(...)
                                   .useBlockingExecutor(true)
                                   .build());
```

.. so that we do not give a user an impression that `configureService` will be automatically invoked?

anuraaga: Sure - wasn't sure if the doc snippets are supposed to compile by itself or not

trustin: It's OK even if it doesn't compile in my opinion. We could add a fully working example to `examples/` and add a link in `.rst`, though.

anuraaga:
Thanks for review :) Merged master just to fix the flaky test

trustin:
@anuraaga One last request: How about renaming `useBlockingExecutor` to `useBlockingTaskExecutor()` so that it is clear that it's about `ctx.blockingTaskExecutor()`?

trustin:
Thanks, @anuraaga !

Issue #1139, opened 2018-04-10
Allow specifying different CORS policies for different origins


trustin:
Currently, it is not possible to specify different CORS policies for different origins, due to the limitation in our API design.

Related: https://github.com/netty/netty/issues/7785 https://github.com/netty/netty/pull/7800

We could change our API, for example:

```java
CorsServiceBuilder.forOrigin("foo")...
                  .andForOrigin("bar")...
                  .newDecorator();
```

I believe `CorsConfig` will have breaking changes.

/cc @jprante

tmdgus0118:
I'm currently working on it.

PR #1526, opened 2019-01-11
Support different CORS policy for different origins.

tmdgus0118:
Issue #1139 -
Motivation
A user may want to use different CORS policy for different origins.

Modification
- Added `CorsPolicy`, `AbstractCorsPolicyBuilder`, `ChainedCorsPolicyBuilder`, `CorsPolicyBuilder` classes.
- (Breaking change) Moved most of fields and their getters of `CorsConfig` and `CorsServiceBuilder` except for `anyOriginSupported`, `shortCircuit` to `CorsPolicy` and `AbstractCorsPolicyBuilder` class
- Added `Set<CorsPolicy>` for `CorsConfig` field to support multiple CORS policies.
- Added `andForOrigins` method which returns `ChainedCorsPolicyBuilder` for supporting method chaining. Like the usage of `ServerBuilder` and `ChainedVirtualHostBuilder`, a user can get `CorsServiceBuilder` back via `ChainedCorsPolicyBuilder.and()`.

Usage:
- A user can set a new policy for different origins via `andForOrigins`, and he or she has to use `and()` to make a new decorator or build.

- There is no different if a user wants to use just one policy. 

- A user can also create a `CorsPolicy` via CorsPolicyBuilder and add it via `CorsServiceBuilder.addPolicy`

For example,

```java
	// Usage 1 - 1 policy
ServerBuilder sb = new ServerBuilder().service(“/cors”, new AbstractHttpService() {
	…
}.decorator(CorsServiceBuilder.forOrigin(“http://example.com”)
			…
			.newDecorator());

// Usage 2 - multiple policies

ServerBuilder sb = new ServerBuilder().service(“/cors”, new AbstractHttpService() {
	…
}.decorator(CorsServiceBuilder.forOrigin(“http://example.com”)
				…
			.andForOrigins(“http://example2.com”)
				…
			.andForOrigins(“http://example3.com”, “http://example4.com”)
				…
			.and()
			.newDecorator());
```

** I will send another pull request for a document about CORS Service which is not written yet.


core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,383 @@
...
+     * Determines if the policy allows a null origin.
+     */
+    boolean isNullOriginAllowed() {
+        return nullOriginAllowed;
+    }
+
+    /**
+     * Web browsers may set the 'Origin' request header to 'null' if a resource is loaded
+     * from the local file system. Calling this method will enable a successful CORS response
+     * with a {@code "null"} value for the the CORS response header 'Access-Control-Allow-Origin'.
```
hyangtack: nit: `"null"` -> `null`, `'Access-Control-Allow-Origin'` -> `{@code Access-Control-Allow-Origin}`?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,383 @@
...
+
+    /**
+     * Determines if the policy allows a null origin.
+     */
+    boolean isNullOriginAllowed() {
+        return nullOriginAllowed;
+    }
+
+    /**
+     * Web browsers may set the 'Origin' request header to 'null' if a resource is loaded
```
hyangtack: nit: `'Origin'` -> `{@code Origin}`, `'null'` -> `{@code null}`?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,383 @@
...
+     *
+     * <p>During a simple CORS request, only certain response headers are made available by the
+     * browser, for example using:
+     * <pre>{@code
+     * xhr.getResponseHeader("Content-Type");
+     * }</pre>
+     *
+     * <p>The headers that are available by default are:
+     * <ul>
+     * <li>Cache-Control</li>
```
hyangtack: nit: indentation?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,383 @@
...
+            if (methods[i] == null) {
+                throw new NullPointerException("methods[" + i + ']');
+            }
+        }
+        Collections.addAll(allowedRequestMethods, methods);
+        return self();
+    }
+
+    /**
+     * Specifies the if headers that should be returned in the CORS 'Access-Control-Allow-Headers'
```
hyangtack: nit: `the if headers` -> `the headers`?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,383 @@
...
+                throw new NullPointerException("headers[" + i + ']');
+            }
+        }
+
+        Arrays.stream(headers).map(HttpHeaderNames::of).forEach(allowedRequestHeaders::add);
+        return self();
+    }
+
+    /**
+     * Specifies the if headers that should be returned in the CORS 'Access-Control-Allow-Headers'
```
hyangtack: Ditto

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,383 @@
...
+     * }</pre>
+     * the server will receive the above header name in the 'Access-Control-Request-Headers' of the
+     * preflight request. The server will then decide if it allows this header to be sent for the
+     * real request (remember that a preflight is not the real request but a request asking the server
+     * if it allow a request).
+     *
+     * @param headers the headers to be added to the preflight 'Access-Control-Allow-Headers' response header.
+     * @return {@link B} to support method chaining.
+     */
+    public B allowRequestHeaders(final String... headers) {
```
hyangtack: nit: How about changing `String` to `CharSequence` and then removing `public B allowRequestHeaders(final AsciiString... headers)`? We can use cached `AsciiString` for both of `String` and `AsciiString`.

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
tmdgus0118: I'm not sure about that because it literally means a String value "null". I think {@code null} may be more confusing. I also think that 'Access-Control-Allow-Origin' may be better with quotes for disambiguation.

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
tmdgus0118: you're right. Thanks!

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,383 @@
...
+     */
+    public B disablePreflightResponseHeaders() {
+        preflightResponseHeadersDisabled = true;
+        return self();
+    }
+
+    /**
+     * Returns a newly-created {@link CorsPolicy} based on the properties of this builder.
+     */
+    public CorsPolicy build() {
```
trustin: This should not be visible when a user uses `ChainedCorsPolicyBuilder`. We could make it package-local and increase the visibility at `CorsPolicyBuilder`?

core/src/main/java/com/linecorp/armeria/server/cors/CorsConfig.java
```diff
@@ -205,111 +96,42 @@ public boolean isCredentialsAllowed() {
...
-    public long maxAge() {
+    public Set<CorsPolicy> policies() {
         ensureEnabled();
-        return maxAge;
+        return policies;
     }
 
     /**
-     * Returns a set of headers to be exposed to calling clients.
+     * Returns the policy for the specific {@code origin}.
```
trustin: specified?

core/src/main/java/com/linecorp/armeria/server/cors/CorsConfig.java
```diff
@@ -205,111 +96,42 @@ public boolean isCredentialsAllowed() {
...
-     *
-     * @throws IllegalStateException if CORS support is not enabled
-     */
-    public HttpHeaders preflightResponseHeaders() {
-        ensureEnabled();
-
-        if (preflightResponseHeaders.isEmpty()) {
-            return HttpHeaders.EMPTY_HEADERS;
+    @Nullable
+    CorsPolicy getPolicy(String origin) {
```
trustin: - `requireNonNull` on `origin`
- Should we make this a public method (or not)?

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,285 @@
...
+import com.linecorp.armeria.common.DefaultHttpHeaders;
+import com.linecorp.armeria.common.HttpHeaderNames;
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.server.cors.CorsConfig.ConstantValueSupplier;
+import com.linecorp.armeria.server.cors.CorsConfig.InstantValueSupplier;
+
+import io.netty.util.AsciiString;
+
+public final class CorsPolicy {
```
trustin: Missing Javadoc

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,285 @@
...
+            this.preflightResponseHeaders = ImmutableMap.of(
+                    HttpHeaderNames.DATE, InstantValueSupplier.INSTANCE,
+                    HttpHeaderNames.CONTENT_LENGTH, ConstantValueSupplier.ZERO);
+        } else {
+            this.preflightResponseHeaders = ImmutableMap.copyOf(preflightResponseHeaders);
+        }
+    }
+
+    /**
+     * Returns the allowed origin. This can either be a wildcard or an origin value.
```
trustin: Could you add:

  This method returns the first specified origin if this policy has more than one origin.

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,285 @@
...
+     * Returns a set of headers to be exposed to calling clients.
+     *
+     * <p>During a simple CORS request only certain response headers are made available by the
+     * browser, for example using:
+     * <pre>{@code
+     * xhr.getResponseHeader("Content-Type");
+     * }</pre>
+     * The headers that are available by default are:
+     * <ul>
+     * <li>Cache-Control</li>
```
trustin: indentation and `{@code ...}`

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,285 @@
...
+     */
+    public Set<AsciiString> exposedHeaders() {
+        return exposedHeaders;
+    }
+
+    /**
+     * Returns the allowed set of Request Methods. The Http methods that should be returned in the
+     * CORS 'Access-Control-Request-Method' response header.
+     *
+     * @return the {@link HttpMethod}s that represent the allowed Request Methods.
```
trustin: `Request Methods` -> request methods

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,285 @@
...
+        final HttpHeaders preflightHeaders = new DefaultHttpHeaders(false);
+        for (Entry<AsciiString, Supplier<?>> entry : preflightResponseHeaders.entrySet()) {
+            final Object value = getValue(entry.getValue());
+            if (value instanceof Iterable) {
+                preflightHeaders.addObject(entry.getKey(), (Iterable<?>) value);
+            } else {
+                preflightHeaders.addObject(entry.getKey(), value);
+            }
+        }
+        return preflightHeaders;
```
trustin: Could we build an immutable `HttpHeaders` and reuse it, instead of building it every time?

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicyBuilder.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.cors;
+
+public final class CorsPolicyBuilder extends AbstractCorsPolicyBuilder<CorsPolicyBuilder> {
```
trustin: Javadoc with a simple example.

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /**
+     * This is a non CORS specification feature which enables the setting of preflight
+     * response headers that might be required by intermediaries.
+     *
+     * @param headers the {@link HttpHeaders} to which the preflight headers should be added.
+     */
+    void setPreflightHeaders(final HttpHeaders headers) {
```
trustin: It seems to me that `CorsService` is a better place for all these setter methods.

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -331,10 +291,9 @@ public CorsServiceBuilder allowRequestHeaders(final AsciiString... headers) {
      *
      * @param name the name of the HTTP header.
      * @param values the values for the HTTP header.
-     * @return {@link CorsServiceBuilder} to support method chaining.
+     * @return {@link AbstractCorsPolicyBuilder} to support method chaining.
```
trustin: Revert?

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -374,12 +321,10 @@ public CorsServiceBuilder preflightResponseHeader(final AsciiString name, final
      * @param name the name of the HTTP header.
      * @param values the values for the HTTP header.
      * @param <T> the type of values that the Iterable contains.
-     * @return {@link CorsServiceBuilder} to support method chaining.
+     * @return {@link AbstractCorsPolicyBuilder} to support method chaining.
```
trustin: Revert?

core/src/main/java/com/linecorp/armeria/server/cors/ChainedCorsPolicyBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public CorsServiceBuilder and() {
+        return serviceBuilder;
+    }
+
+    /**
+     * Create a new instance of {@link ChainedCorsPolicyBuilder} added to the parent {@link CorsServiceBuilder}.
+     * @return the created instance.
+     */
+    public ChainedCorsPolicyBuilder andForOrigins(final String... origins) {
```
trustin: Could you also add `andForOrigin(String origin)` for consistency with `CorsService.forOrigin(String)`?

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -68,48 +62,49 @@ public static CorsServiceBuilder forOrigin(final String origin) {
...
-            }
-        }
         return new CorsServiceBuilder(origins);
     }
 
-    final Set<String> origins;
     final boolean anyOriginSupported;
-    boolean nullOriginAllowed;
-    boolean credentialsAllowed;
+    final ChainedCorsPolicyBuilder defaultPolicyBuilder;
```
trustin: I'm curious if this class can be simplified by extending `AbstractCorsServiceBuilder` instead of delegating all builder methods. Thoughts?

core/src/main/java/com/linecorp/armeria/server/cors/CorsService.java
```diff
@@ -108,6 +109,58 @@ protected HttpObject filter(HttpObject obj) {
...
+        requireNonNull(policy, "policy");
+        for (Map.Entry<AsciiString, String> entry : policy.preflightResponseHeaders()) {
+            headers.add(entry.getKey(), entry.getValue());
+        }
+    }
+
+    void setCorsAllowCredentials(final CorsPolicy policy, final HttpHeaders headers) {
+        requireNonNull(policy, "policy");
+        if (policy.isCredentialsAllowed() &&
+            !headers.get(HttpHeaderNames.ACCESS_CONTROL_ALLOW_ORIGIN).equals(CorsService.ANY_ORIGIN)) {
```
trustin: `headers.get(...)` could return `null`? How about flipping the equals caller and callee?

core/src/main/java/com/linecorp/armeria/server/cors/CorsService.java
```diff
@@ -108,6 +109,58 @@ protected HttpObject filter(HttpObject obj) {
...
+        }
+    }
+
+    void setCorsExposeHeaders(final CorsPolicy policy, final HttpHeaders headers) {
+        requireNonNull(policy, "policy");
+        if (policy.exposedHeaders().isEmpty()) {
+            return;
+        }
+
+        headers.set(HttpHeaderNames.ACCESS_CONTROL_EXPOSE_HEADERS, HEADER_JOINER.join(policy.exposedHeaders()));
```
trustin: How about keeping the precalculated value of the headers set by `setCors*()` in `CorsPolicy`? Then, it would make sense to move the setter methods to `CorsPolicy`.

core/src/main/java/com/linecorp/armeria/server/cors/CorsService.java
```diff
@@ -172,60 +215,45 @@ private static HttpResponse forbidden() {
...
-            return false;
+            return null;
         }
 
         final String origin = request.headers().get(HttpHeaderNames.ORIGIN);
         if (origin != null) {
-            if (NULL_ORIGIN.equals(origin) && config.isNullOriginAllowed()) {
+            final CorsPolicy policy = config.getPolicy(origin);
+            if (policy == null) {
+                logger.debug("There is no CORS policy configured for the request origin [{}]]", origin);
```
trustin: Could you also log the current `ServiceRequestContext`?  You'll have to add `ServiceRequestContext` as the first parameter of `setCorsOrigin()`.

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -225,51 +176,39 @@ public CorsServiceBuilder exposeHeaders(final String... headers) {
...
-            }
-        }
-        Collections.addAll(allowedRequestMethods, methods);
+        firstPolicyBuilder.allowRequestMethods(methods);
         return this;
     }
 
     /**
-     * Specifies the if headers that should be returned in the CORS 'Access-Control-Allow-Headers'
+     * Specifies the if the headers that should be returned in the CORS {@code 'Access-Control-Allow-Headers'}
```
trustin: the if the -> the

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -16,28 +16,19 @@
...
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.server.Service;
-import com.linecorp.armeria.server.cors.CorsConfig.ConstantValueSupplier;
-
-import io.netty.util.AsciiString;
 
 /**
  * Builds a new {@link CorsService} or its decorator function.
```
trustin: Could you add a simple example that configures multiple origins with different policies? (using chaining)

trustin:
@jprante You might be interested in this pull request, which is a follow-up to your `CorsService` work to Armeria.

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,320 @@
...
+     * have certain headers set. This enables such headers to be added.
+     *
+     * @param name the name of the HTTP header.
+     * @param values the values for the HTTP header.
+     * @param <T> the type of values that the Iterable contains.
+     * @return {@code this} to support method chaining.
+     */
+    public <T> B preflightResponseHeader(final CharSequence name, final Iterable<T> values) {
+        requireNonNull(name, "name");
+        requireNonNull(values, "values");
```
minwoox: I'm wondering if we have to check the `values` contains `null` as we do in the method above.

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,320 @@
...
+                          .add("allowedRequestMethods", allowedRequestMethods)
+                          .add("allowedRequestHeaders", allowedRequestHeaders)
+                          .add("preflightResponseHeaders", preflightResponseHeaders)
+                          .add("preflightResponseHeadersDisabled", preflightResponseHeadersDisabled).toString();
+    }
+
+    @Override
+    public String toString() {
+        return toString(this, origins, nullOriginAllowed, credentialsAllowed, maxAge,
+                                   exposedHeaders, allowedRequestMethods, allowedRequestHeaders,
```
minwoox: indentation?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
tmdgus0118: I think we may have to check them. Thanks. 
Added the logic checking if `values` contains `null` and making a new `ArrayList` so that the constructor of `CorsPolicy` doesn't have to check if it is an instance of `Iterable`.

tmdgus0118: I think we need to check them. Added it.
Thanks

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+     * <p>Some values must be dynamically created when the HTTP response is created, for
+     * example the 'Date' response header. This can be accomplished by using a {@link Supplier}
+     * which will have its 'call' method invoked when the HTTP response is created.
+     *
+     * @param name the name of the HTTP header.
+     * @param valueSupplier a {@link Supplier} which will be invoked at HTTP response creation.
+     * @param <T> the type of the value that the {@link Supplier} can return.
+     * @return {@code this} to support method chaining.
+     */
+    public <T> B preflightResponseHeader(CharSequence name, Supplier<T> valueSupplier) {
```
trustin: `<T>` -> `<?>` (no point of defining a type parameter in this case.)

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,300 @@
...
+    private final String joinedAllowedRequestHeaders;
+    private final String joinedAllowedRequestMethods;
+    private final HttpHeaders preflightResponseHeaders;
+
+    CorsPolicy(final Set<String> origins, boolean credentialsAllowed, long maxAge,
+               boolean nullOriginAllowed, Set<AsciiString> exposedHeaders,
+               Set<AsciiString> allowedRequestHeaders, Set<HttpMethod> allowedRequestMethods,
+               boolean preflightResponseHeadersDisabled,
+               Map<AsciiString, Supplier<?>> preflightResponseHeaders) {
+        this.origins = origins;
```
trustin: `ImmutableSet.copyOf(origins)`

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,300 @@
...
+        this.credentialsAllowed = credentialsAllowed;
+        this.maxAge = maxAge;
+        this.nullOriginAllowed = nullOriginAllowed;
+        this.exposedHeaders = ImmutableSet.copyOf(exposedHeaders);
+        this.allowedRequestMethods = EnumSet.copyOf(allowedRequestMethods);
```
trustin: How about changing the type of the parameter `allowedRequestMethods` to `EnumSet<HttpMethod>` and using `ImmutableSet.copyOf()` here, so that Guava internally uses `ImmutableEnumSet`?

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,300 @@
...
+        if (preflightResponseHeadersDisabled) {
+            preflightResponseHeadersMap = Collections.emptyMap();
+        } else if (preflightResponseHeaders.isEmpty()) {
+            preflightResponseHeadersMap = ImmutableMap.of(
+                    HttpHeaderNames.DATE, InstantValueSupplier.INSTANCE,
+                    HttpHeaderNames.CONTENT_LENGTH, ConstantValueSupplier.ZERO);
+        } else {
+            preflightResponseHeadersMap = preflightResponseHeaders;
+        }
+        final HttpHeaders preflightHeaders;
```
trustin: I realized preflight headers cannot be cached unless it's disabled. For example, `InstantValueSupplier` above will have to produce a different timestamp for each request. We could cache it only when:

- preflight headers are disabled or empty, or
- all suppliers are `ConstantValueSupplier`

(or we could just give up caching preflight headers?)

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,300 @@
...
+
+    /**
+     * Determines if the policy allows a null origin.
+     */
+    public boolean isNullOriginAllowed() {
+        return nullOriginAllowed;
+    }
+
+    /**
+     * Returns HTTP response headers that should be added to a CORS preflight response.
```
trustin: Returns HTTP response headers -> Generates immutable HTTP response headers

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,300 @@
...
+     *
+     * @return {@link HttpHeaders} the HTTP response headers to be added.
+     */
+    public HttpHeaders preflightResponseHeaders() {
```
trustin: How about renaming to `generatePreflightResponseHeaders()`?

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,300 @@
...
+        return preflightResponseHeaders;
+    }
+
+    /**
+     * This is a non CORS specification feature which enables the setting of preflight
+     * response headers that might be required by intermediaries.
+     *
+     * @param headers the {@link HttpHeaders} to which the preflight headers should be added.
+     */
+    void setPreflightHeaders(final HttpHeaders headers) {
```
trustin: Perhaps we can remove this method and let the caller do the following?

```java
headers.add(policy.generatePreflightResponseHeaders());
```

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -16,31 +16,39 @@
...
+ *                            .allowCredentials()
+ *                            .allowRequestMethods(HttpMethod.GET, HttpMethod.POST)
+ *                            .allowRequestHeaders("allow_request_header1", "allow_request_header2")
+ *                            .andForOrigins("http://example3.com")
+ *                            .allowCredentials()
+ *                            .allowRequestMethods(HttpMethod.GET)
+ *                            .and();
+ * new ServerBuilder().service("/cors", new AbstractHttpService() {
+ *     ...
+ * }).decorator(cb.newDecorator())
```
trustin: How about:

```java
sb.service("/cors", myService.decorate(
        CorsServiceBuilder.forOrigin(...)
                          ...
                          .and()
                          .newDecorator()));
```

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -398,10 +292,8 @@ public CorsServiceBuilder preflightResponseHeader(final AsciiString name, final
      * @param <T> the type of the value that the {@link Supplier} can return.
      * @return {@link CorsServiceBuilder} to support method chaining.
      */
-    public <T> CorsServiceBuilder preflightResponseHeader(AsciiString name, Supplier<T> valueSupplier) {
-        requireNonNull(name, "name");
-        requireNonNull(valueSupplier, "valueSupplier");
-        preflightResponseHeaders.put(name, valueSupplier);
+    public <T> CorsServiceBuilder preflightResponseHeader(CharSequence name, Supplier<T> valueSupplier) {
```
trustin: `<T>` -> `<?>`

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+    private final Map<AsciiString, Supplier<?>> preflightResponseHeaders = new HashMap<>();
+    private boolean preflightResponseHeadersDisabled;
+
+    AbstractCorsPolicyBuilder() {
+        origins = Collections.emptySet();
+    }
+
+    AbstractCorsPolicyBuilder(String... origins) {
+        requireNonNull(origins, "origins");
+        for (int i = 0; i < origins.length; i++) {
```
trustin: Should we also reject when `origins` is empty?

```java
checkArgument(origins.length > 0, "origins is empty.");
```

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
tmdgus0118: I think this method would be better remaining for consistency because all of the other fields (such as allow methods, allow headers etc) do have the setters for the headers. What do you think?

trustin: Sounds reasonable to me. It's fine to keep this setter. :-)

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,293 @@
...
+     * Returns HTTP response headers that should be added to a CORS preflight response.
+     *
+     * @return {@link HttpHeaders} the HTTP response headers to be added.
+     */
+    public HttpHeaders generatePreflightResponseHeaders() {
+        final HttpHeaders headers = new DefaultHttpHeaders();
+        preflightResponseHeaders.forEach((key,value) -> {
+            final Object val = getValue(value);
+            if (val instanceof Iterable) {
+                headers.addObject(key, (Iterable<?>)val);
```
trustin: Missing space before `val`

core/src/main/java/com/linecorp/armeria/server/cors/CorsService.java
```diff
@@ -172,60 +155,48 @@ private static HttpResponse forbidden() {
...
         }
 
         final String origin = request.headers().get(HttpHeaderNames.ORIGIN);
         if (origin != null) {
-            if (NULL_ORIGIN.equals(origin) && config.isNullOriginAllowed()) {
+            final CorsPolicy policy = config.getPolicy(origin);
+            if (policy == null) {
+                logger.debug(
+                        "There is no CORS policy configured for the request origin [{}] with the context [{}]",
+                        origin, ctx);
```
trustin: `ctx` needs to come first:

```java
"{} There is no ... the request origin '{}'.", ctx, origin
```

core/src/main/java/com/linecorp/armeria/server/cors/CorsConfig.java
```diff
@@ -205,111 +98,43 @@ public boolean isCredentialsAllowed() {
...
-
-    /**
-     * Returns the allowed set of Request Headers.
+     * Returns the policies.
      *
-     * <p>The header names returned from this method will be used to set the CORS
-     * 'Access-Control-Allow-Headers' response header.
-     *
-     * @throws IllegalStateException if CORS support is not enabled
+     * @throws IllegalStateException if CORS support is not enabled.
```
trustin: This is untrue, too.

trustin: Not actually. Never mind.

core/src/main/java/com/linecorp/armeria/server/cors/ChainedCorsPolicyBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Builds a new {@link CorsPolicy}.
+ *
+ * <p>This class can only be created through the {@link CorsServiceBuilder#andForOrigins(String...)} or
+ * {@link CorsServiceBuilder#andForOrigin(String)} method of the {@link CorsServiceBuilder}.
+ *
+ * <p>Call {@link #and()} method can also return to {@link CorsServiceBuilder}.
+ *
```
trustin: nit: A redundant empty line.

core/src/main/java/com/linecorp/armeria/server/cors/ChainedCorsPolicyBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Builds a new {@link CorsPolicy}.
+ *
+ * <p>This class can only be created through the {@link CorsServiceBuilder#andForOrigins(String...)} or
+ * {@link CorsServiceBuilder#andForOrigin(String)} method of the {@link CorsServiceBuilder}.
+ *
+ * <p>Call {@link #and()} method can also return to {@link CorsServiceBuilder}.
```
trustin: Calling and() method will return the control to CorsServiceBuilder.

core/src/main/java/com/linecorp/armeria/server/cors/ChainedCorsPolicyBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Returns the parent {@link CorsServiceBuilder}.
+     */
+    public CorsServiceBuilder and() {
+        return serviceBuilder;
+    }
+
+    /**
+     * Create a new instance of {@link ChainedCorsPolicyBuilder} added to the parent {@link CorsServiceBuilder}.
```
trustin: Creates

core/src/main/java/com/linecorp/armeria/server/cors/ChainedCorsPolicyBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return the created instance.
+     */
+    public ChainedCorsPolicyBuilder andForOrigins(final String... origins) {
+        return and().andForOrigins(origins);
+    }
+
+    /**
+     * Create a new instance of {@link ChainedCorsPolicyBuilder} added to the parent {@link CorsServiceBuilder}.
```
trustin: Creates

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,307 @@
...
+    CorsPolicy build() {
+        return new CorsPolicy(origins, credentialsAllowed, maxAge, nullOriginAllowed,
+                              exposedHeaders, allowedRequestHeaders, allowedRequestMethods,
+                              preflightResponseHeadersDisabled, preflightResponseHeaders);
+    }
+
+    @Override
+    public String toString() {
+        return CorsPolicy.toString(this, origins, nullOriginAllowed, credentialsAllowed, maxAge,
+                        exposedHeaders, allowedRequestMethods, allowedRequestHeaders,
```
hyangtack: nit: alignment :-)

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,307 @@
...
+    /**
+     * Sets the CORS {@code "Access-Control-Max-Age"} response header and enables the
+     * caching of the preflight response for the specified time. During this time no preflight
+     * request will be made.
+     *
+     * @param maxAge the maximum time, in seconds, that the preflight response may be cached.
+     * @return {@code this} to support method chaining.
+     */
+    public B maxAge(final long maxAge) {
+        checkState(maxAge <= 0, "maxAge: %d (expected: > 0)", maxAge);
```
hyangtack: `maxAge > 0`? Would you add a test case which sets `maxAge` as well?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,307 @@
...
+        }
+        this.origins = Arrays.stream(origins).map(Ascii::toLowerCase).collect(toImmutableSet());
+    }
+
+    @SuppressWarnings("unchecked")
+    B self() {
+        return (B) this;
+    }
+
+    public Set<String> origins() {
```
hyangtack: Is this method necessary? If so, please add Javadoc.

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,293 @@
...
+import com.linecorp.armeria.server.cors.CorsConfig.InstantValueSupplier;
+
+import io.netty.util.AsciiString;
+
+/**
+ * Contains information of the CORS policy with the specified origins.
+ */
+public final class CorsPolicy {
+
+    private static final Logger logger = LoggerFactory.getLogger(CorsPolicy.class);
```
hyangtack: Seems not used. Could we remove this?

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,293 @@
...
+        } else {
+            this.preflightResponseHeaders = ImmutableMap.copyOf(preflightResponseHeaders);
+        }
+    }
+
+    /**
+     * Returns the allowed origin. This can either be a wildcard or an origin value.
+     * This method returns the first specified origin if this policy has more than one origin.
+     *
+     * @return the value that will be used for the CORS response header {@code 'Access-Control-Allow-Origin'}
```
hyangtack: nit: `'` -> `"`

core/src/main/java/com/linecorp/armeria/server/cors/CorsPolicy.java
```diff
@@ -0,0 +1,293 @@
...
+     * {@code true} cookies will be added to CORS requests. Setting this value to true will set the
+     * CORS {@code "Access-Control-Allow-Credentials"} response header to {@code true}.
+     *
+     * <p>Please note that cookie support needs to be enabled on the client side as well.
+     * The client needs to opt-in to send cookies by calling:
+     * <pre>{@code
+     * xhr.withCredentials = true;
+     * }</pre>
+     *
+     * <p>The default value for 'withCredentials' is {@code false} in which case no cookies are sent.
```
hyangtack: nit: `{@code 'withCredentials'}` for consistency?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -0,0 +1,310 @@
...
+     * have certain headers set. This enables such headers to be added.
+     *
+     * @param name the name of the HTTP header.
+     * @param values the values for the HTTP header.
+     * @return {@code this} to support method chaining.
+     */
+    public B preflightResponseHeader(CharSequence name, Iterable<?> values) {
+        requireNonNull(name, "name");
+        requireNonNull(values, "values");
+        checkArgument(values.iterator().hasNext(), "values should not be empty.");
```
minwoox: `!Iterables.isEmpty(values)`?

trustin:
Great job, @tmdgus0118 :+1:

PR #1527, opened 2019-01-12
Don't synchronize CountingSampler.

anuraaga:
Currently `CountingSampler` synchronizes threads to manipulate an integer counter such that it doesn't go over `100`. An `AtomicInteger` and `mod` can do the same with much less overhead.

See https://github.com/openzipkin/brave/pull/857


trustin:
Thanks, @anuraaga !

PR #1529, opened 2019-01-14
Less verbose error logging / Fix incorrect behavior when 'verboseExceptions' is set.

### LINKED ISSUE #1515: Do not log 'Sent a GOAWAY frame: debugData="Error flushing"'
.. because it basically means the connection is already over or the connection has been reset.

Related: #1401

trustin:
Motivation:

1)

`Exceptions.isExpected(Throwable)` should always return `false` when
`verboseExceptions` flag is set. However, it currently always returns
`true` by coding mistake.

2)

I found a few more exception messages which do not really need to be
logged at high level:

    SSLException: SSLEngine closed already
    Sent a GOAWAY frame: lastStreamId=..., errorCode=INTERNAL_ERROR(2), debugData="Error flushing"

Both errors can happen when a socket connection has been closed by a
remote peer, which should be treated in a similar manner with
'connection reset by peer' errors.

Modifications:

- Update `Exceptions.isExpected()` to always return `false` when
  `verboseExceptions` flag is set.
- Update `Exceptions.isExpected()` to return `true` for
  `SSLException: SSLEngine closed already`.
- Update `Http2GoAwayHandler` to log at DEBUG level for
  `Sent a GOAWAY frame: ... debugData="Error flushing"`.

Result:

- Fixes #1515
- `com.linecorp.armeria.verboseExceptions` flag works as expected.
- Cleaner log file with less WARN-level log messages.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/common/util/ExceptionsTest.java
+++ b/core/src/test/java/com/linecorp/armeria/common/util/ExceptionsTest.java
@@ -18,11 +18,22 @@ package com.linecorp.armeria.common.util;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
+import java.nio.channels.ClosedChannelException;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 
+import javax.net.ssl.SSLException;
+
 import org.junit.Test;
+
+import com.linecorp.armeria.common.ClosedSessionException;
+import com.linecorp.armeria.common.Flags;
+
+import io.netty.channel.ChannelException;
+import io.netty.handler.codec.http2.Http2Error;
+import io.netty.handler.codec.http2.Http2Exception;
 
 public class ExceptionsTest {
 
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/linecorp/armeria/common/util/ExceptionsTest.java
```diff
@@ -66,4 +77,31 @@ public void testTraceText() {
         assertThat(trace).startsWith(Exception.class.getName() + ": foo\n" +
                                      "\tat " + ExceptionsTest.class.getName());
     }
+
+    @Test
+    public void testIsExpected() {
+        final boolean expected = !Flags.verboseExceptions();
```
minwoox: Just wondering if we can just make this as `true`.(not using `!Flags.verboseExceptions()`)

trustin: There was actually a bug in `Exceptions.isExpected()` where a wrong boolean value is returned when `Flags.verboseExceptions()` is `true`.

trustin:
Thanks for reviewing.

PR #1530, opened 2019-01-14
[WIP] Support streaming response by default response converters

hyangtack:
Motivation:
Currently the default response converters do not support streaming response even if the returned object is a `Publisher` or `Stream`. So a user who wants to send a streaming response should use `HttpResponse` directly or write his or her own response converter. It would be useful if the default response converters support streaming response.

Modifications:
- The default response converters support `Publisher` and `Stream` by themselves if a user specifies the content type of a response with `@Produces` annotation.
  - They consumed the collected objects published by a `Publisher` or `Stream` before.
- `JacksonResponseConverterFunction` supports JSON Text Sequences (`https://tools.ietf.org/html/rfc7464`). Added `MediaType.JSON_SEQ` constant and `@ProducesJsonSequences` annotation for that.
- Add `ObjectCollectingUtil` and `ResponseConversionUtil` to `c.l.a.internal.annotation` package for default response converters.
- Add `reactor-test` dependency to `core` for testing.

Result:
- Streaming response support


hyangtack:
@trustin @minwoox Please let me know what you think about:

1. using `reactor-test` in `core` module
1. how to handle `Mono` of reactor (Check [here](https://github.com/line/armeria/pull/1530/files#diff-22f9e9aa0df98fb32ccfc8dbf74701f4R45) to see the current implementation). Would it better to add `reactor` module like `rxjava`?

core/src/main/java/com/linecorp/armeria/internal/annotation/ObjectCollectingUtil.java
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * The {@link Class} instance of {@code reactor.core.publisher.Mono} of
+     * <a href="https://projectreactor.io/">Project Reactor</a>.
+     */
+    private static final Class<?> MONO_CLASS;
+
+    static {
+        Class<?> mono = null;
+        try {
+            mono = Class.forName("reactor.core.publisher.Mono");
```
trustin: Please specify the class loader, e.g. `ObjectCollectingUtil.class.getClassLoader()`.

core/src/main/java/com/linecorp/armeria/internal/annotation/ObjectCollectingUtil.java
```diff
@@ -0,0 +1,163 @@
...
+        CollectingSingleObjectSubscriber(CompletableFuture<Object> future) {
+            super(future);
+        }
+
+        @Override
+        public void onNext(T t) {
+            if (result == null) {
+                result = t;
+            } else {
+                onError(new IllegalStateException("Too many objects are published."));
```
trustin: How about: `"Only one element can be published: " + t`?

hyangtack:
@line/cdtf Removed WIP tag. Please review this PR. 

FYI, I have a plan to send follow-up PRs, such as:
- Adding a new converter for Server Sent Events
- Adding documentation about streaming response converters

core/src/main/java/com/linecorp/armeria/internal/annotation/ResponseConversionUtil.java
```diff
@@ -0,0 +1,275 @@
...
+        requireNonNull(contentConverter, "contentConverter");
+
+        return aggregateFrom(collectFrom(publisher),
+                             headers, trailingHeaders, contentConverter);
+    }
+
+    private static HttpResponseWriter aggregateFrom(CompletableFuture<?> future,
+                                                    HttpHeaders headers, HttpHeaders trailingHeaders,
+                                                    Function<Object, HttpData> contentConverter) {
+        final HttpResponseWriter writer = HttpResponse.streaming();
```
trustin: Curious if `HttpResponse.from()` is a better option here. By changing so, all `aggregateFrom()` methods could return `HttpResponse` as well.

core/src/main/java/com/linecorp/armeria/internal/annotation/ResponseConversionUtil.java
```diff
@@ -0,0 +1,275 @@
...
+                    subscription.cancel();
+                }
+                return null;
+            });
+            try {
+                writer.onDemand(new Runnable() {
+                    @Override
+                    public void run() {
+                        assert subscription != null;
+                        subscription.request(1);
```
trustin: Not sure which is better, but you can use `s` here and remove the assertion.

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
```diff
@@ -513,4 +552,43 @@ public void charset() {
...
+                        assertThat(o).isInstanceOf(HttpHeaders.class);
+                        final HttpHeaders headers = (HttpHeaders) o;
+                        assertThat(headers.status()).isEqualTo(HttpStatus.OK);
+                        assertThat(headers.contentType()).isEqualTo(MediaType.JSON_SEQ);
+                    })
+                    .assertNext(o -> ensureExpectedHttpData.accept(o, "foo"))
+                    .assertNext(o -> ensureExpectedHttpData.accept(o, "bar"))
+                    .assertNext(o -> ensureExpectedHttpData.accept(o, "baz"))
+                    .assertNext(o -> ensureExpectedHttpData.accept(o, "qux"))
+                    .expectNext(HttpData.EMPTY_DATA)
```
trustin: Is this expected? Perhaps we could avoid emitting this?

core/src/test/java/com/linecorp/armeria/server/annotation/StringResponseConverterFunctionTest.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final ResponseConverterFunction function = new StringResponseConverterFunction();
+    private static final ServiceRequestContext ctx = mock(ServiceRequestContext.class);
+
+    private static final HttpHeaders PLAIN_TEXT_HEADERS =
+            HttpHeaders.of(HttpStatus.OK).contentType(MediaType.PLAIN_TEXT_UTF_8);
+    private static final HttpHeaders DEFAULT_TRAILING_HEADERS = HttpHeaders.EMPTY_HEADERS;
+
+    @BeforeClass
+    public static void setup() {
+        when(ctx.blockingTaskExecutor()).thenReturn(MoreExecutors.newDirectExecutorService());
```
trustin: `MoreExecutors.directExecutor()` should be enough.

core/src/test/java/com/linecorp/armeria/server/annotation/StringResponseConverterFunctionTest.java
```diff
@@ -0,0 +1,98 @@
...
+                .isInstanceOf(FallthroughException.class);
+    }
+
+    @Test
+    public void charset() throws Exception {
+        final HttpHeaders headers = HttpHeaders.of(HttpStatus.OK)
+                                               .contentType(MediaType.parse("text/plain; charset=euc-kr"));
+        StepVerifier.create(function.convertResponse(ctx, headers, "한글", DEFAULT_TRAILING_HEADERS))
+                    .expectNext(headers)
+                    .expectNext(HttpData.of("한글".getBytes(Charset.forName("euc-kr"))))
```
trustin: `HttpData.of(Charset.forName("euc-kr"), "한글")` ?

core/src/main/java/com/linecorp/armeria/internal/annotation/ResponseConversionUtil.java
hyangtack: I wanted to decorate a response with `ExceptionFilteredHttpResponse` on [here](https://github.com/line/armeria/pull/1530/files#diff-767b436cc8c0427e4457c63752a88bd2R303) only if the returned type is instance of `HttpResponseWriter`.
Is it better to use `HttpResponse.from()`?

core/src/test/java/com/linecorp/armeria/server/annotation/StringResponseConverterFunctionTest.java
hyangtack: hmm.. `blockingTaskExecutor()` should return `ExecutorService`, not `Executor`.

leecosy:
😚

hyangtack <notifications@github.com> 于2019年1月18日周五 下午4:46写道：

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
hyangtack: Added a comment as follows:
```
.assertNext(o -> {
    // On the server side, HttpResponseSubscriber emits a DATA frame with end of stream
    // flag when the HttpResponseWriter is closed.
    final HttpData lastContent = (HttpData) o;
    assertThat(lastContent.isEmpty()).isTrue();
    assertThat(lastContent.isEndOfStream()).isTrue();
})
```

core/src/test/java/com/linecorp/armeria/server/annotation/StringResponseConverterFunctionTest.java
hyangtack: Got it!

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonResponseConverterFunction.java
```diff
@@ -66,23 +91,59 @@ public HttpResponse convertResponse(ServiceRequestContext ctx,
...
+            return HttpResponse.of(toMutableHeaders(headers).contentType(MediaType.JSON_UTF_8),
+                                   toJsonHttpData(result), trailingHeaders);
         }
 
         return ResponseConverterFunction.fallthrough();
     }
+
+    private HttpData toJsonHttpData(@Nullable Object value) {
+        try {
+            return HttpData.of(mapper.writeValueAsBytes(value));
```
minwoox: I'm wondering if it's ok to produce the bytes which contain "null" string when the value is `null`.

hyangtack: Yes, it's okay because `null` string is also a JSON value. :-)

PR #1236, opened 2018-06-01
Fix a bug where RetryingClient does not stop when ClientFactory is closed

minwoox:
Motivation:
`Client` has to stop when the `ClientFacotry` is closed.

Modifications:
- Store the `ScheduledFuture` and add a lister which stops retrying

Result:
- Fixed another bug

PR #1532, opened 2019-01-15
Deprecate `ClosedClientFactoryException` and use `IllegalStateException`

trustin:
Motivation:

Previously at #1236, we added `ClosedClientFactoryException` which is
raised when a request in progress by `RetryingClient` is cancelled due
to the termination of `ClientFactory`. I see two issues here:

- Defining a new exception type for such an exceptional cause has
  dubious value. It is hard to expect for a user to catch
  `ClosedClientFactoryException` and handle it.
- Due to inevitable race conditions, we are already raising
  `IllegalStateException` in some places, such as the connection pool.
  Indeed, `IllegalStateException` with a message is a better choice
  here.

Modifications:

- Deprecated `ClosedClientFactoryException` and used `IllegalStateException`.
- Made `ClosedClientFactoryException` extend `IllegalStateException`.
- Updates the test cases.

Result:

- Less flakiness
- Leaner API


trustin:
Thanks for reviews.

PR #1535, opened 2019-01-16
Make Armeria(Client|Server)Configurator extend Spring's Ordered inter…

### LINKED ISSUE #1533: Make Armeria*Configurator extend Spring's Ordered interface
Let's imagine that a user specifies `ArmeriaServerConfigurator` bean several times, which makes Armeria configure `ServerBuilder` with a list of `ArmeriaServerConfigurator`s. 

Most of cases there's nothing to do more, but if you need to specify applying order of the configurators you need to use Spring's `@Order` annotation. The default value of the annotation is `Integer.MAX_VALUE` which means the lowest precedence. So if you want to move a bean to the last of the list, you need to add `@Order` annotation to the every other bean. 

It's definitely inconvenient. So we need to change the default order value of `Armeria*Configurator` to `0` in order to make our user happy.

hyangtack:
…face

Motivation:
When specifying evaluation order of the configurators using Spring's `@Order` annotation, it's definitely inconvenient if you want to make a configurator to be called at the last. Because the default value of `@Order` annotation is `Integer.MAX_VALUE` which means the lowest precedence. So it would be more convenient if we change the default order of our configurators to `0` so that a user can easily make a configurator to be called at the last by specifying `@Order(1)` or `@Order(Ordered.LOWEST_PRECEDENCE)`.

Modifications:
- Make `Armeria(Client|Server)Configurator` extend `org.springframework.core.Ordered`.
  - Implement `getOrder()` as a default method which returns `0`.
- (Breaking change) Rename `ArmeriaClientConfigurator#customize` to `configure` for consistency.
- A user can specify multiple `ArmeriaClientConfigurator` beans, like `ArmeriaServerConfigurator`.

Result:
- Closes #1533

### DIFF
--- a/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientConfigurator.java
+++ b/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientConfigurator.java
@@ -15,15 +15,36 @@
  */
 package com.linecorp.armeria.spring.web.reactive;
 
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.AnnotationAwareOrderComparator;
+import org.springframework.core.annotation.Order;
+
 import com.linecorp.armeria.client.HttpClientBuilder;
 
 /**
- * A customizer to customize an {@link HttpClientBuilder} for an {@link ArmeriaClientHttpConnector}.
+ * A configurator to configure an {@link HttpClientBuilder} for an {@link ArmeriaClientHttpConnector}.
  */
 @FunctionalInterface
-public interface ArmeriaClientConfigurator {
+public interface ArmeriaClientConfigurator extends Ordered {
     /**
      * Configures the client using the specified {@link HttpClientBuilder}.
      */
-    void customize(HttpClientBuilder builder);
+    void configure(HttpClientBuilder builder);
+
+    /**
+     * Returns the evaluation order of this configurator. A user can specify the order with an {@link Order}
+     * annotation when defining a bean with a {@link Bean} annotation.
+     *
+     * <p>Note that the default value of the {@link Order} annotation is {@link Ordered#LOWEST_PRECEDENCE}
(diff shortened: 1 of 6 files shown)


spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientConfigurator.java
```diff
@@ -15,15 +15,36 @@
...
+ * A configurator to configure an {@link HttpClientBuilder} for an {@link ArmeriaClientHttpConnector}.
  */
 @FunctionalInterface
-public interface ArmeriaClientConfigurator {
+public interface ArmeriaClientConfigurator extends Ordered {
     /**
      * Configures the client using the specified {@link HttpClientBuilder}.
      */
-    void customize(HttpClientBuilder builder);
+    void configure(HttpClientBuilder builder);
```
trustin: How about adding a deprecated default method:

```java
default void customize(builder) {
   configure(builder);
}
```

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaClientHttpConnector.java
```diff
@@ -119,7 +106,7 @@ private ArmeriaClientHttpRequest createRequest(HttpMethod method, URI uri) {
 
         final URI baseUri = URI.create(Strings.isNullOrEmpty(scheme) ? authority : scheme + "://" + authority);
         final HttpClientBuilder builder = new HttpClientBuilder(baseUri);
-        customizer.customize(builder);
+        configurators.forEach(c -> c.configure(builder));
```
trustin: Should be `c -> c.customize(builder)`

hyangtack: As we talked on LINE chat, let me remove the deprecated method `customize` because providing a default method is useless. :-)

trustin:
Thanks, @hyangtack !

PR #1537, opened 2019-01-17
Fix a bug where CorsService does not work with annotated HTTP services

### LINKED ISSUE #1531: CorsService does not work with annotated HTTP services
For the following annotated service method, a preflight request cannot be handled even if the method is decorated with `CorsService`, because Armeria cannot find the service for a path with `OPTIONS` method.

```
@Get("/hello")
public String hello() { ... }
```

hyangtack:
Motivation:
A preflight request cannot be handled even if the method is decorated with `CorsService` when no `@Options` annotation is specified on the service method.

Modifications:
- Add `isPreflight()` method to `PathMappingContext` interface.
- The router skips checking whether the HTTP method of a request is supported or not, if the request is preflight.

Result:
- Closes #1531


hyangtack:
Let me fix tests after #1526 PR is merged.

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -80,11 +80,13 @@
     public PathMappingResult apply(PathMappingContext mappingCtx) {
         final PathMappingResult result = pathStringMapping.apply(mappingCtx);
         if (!result.isPresent()) {
+            if (mappingCtx.isPreflight() && !mappingCtx.delayedThrowable().isPresent()) {
+                mappingCtx.delayThrowable(HttpStatusException.of(HttpStatus.FORBIDDEN));
```
trustin: Could you explain why we do not return `PathMappingResult.empty()` to trigger 404 here?

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -80,11 +80,13 @@
...
+            }
             return PathMappingResult.empty();
         }
 
-        // We need to check the method at the last in order to return '405 Method Not Allowed'.
-        if (!supportedMethods.contains(mappingCtx.method())) {
+        if (!mappingCtx.isPreflight() && !supportedMethods.contains(mappingCtx.method())) {
```
trustin: How about updating the comment instead of just removing it?

core/src/main/java/com/linecorp/armeria/server/PathMappingContext.java
```diff
@@ -102,4 +104,11 @@ public String path() {
             }
         };
     }
+
+    /**
+     * Returns {@code true} if this context is for a preflight request.
+     *
+     * @see CorsService#isCorsPreflightRequest(HttpRequest)
+     */
+    boolean isPreflight();
```
trustin: How about `isCorsPreflight()`? (because this class is not CORS-specific)

core/src/main/java/com/linecorp/armeria/server/cors/CorsService.java
```diff
@@ -59,6 +59,19 @@
...
     private static final Joiner HEADER_JOINER = Joiner.on(DELIMITER);
 
+    /**
+     * Check for a CORS preflight request.
+     *
+     * @param request the HTTP request with CORS info
+     *
+     * @return {@code true} if HTTP request is a CORS preflight request
+     */
+    public static boolean isCorsPreflightRequest(final HttpRequest request) {
```
trustin: How about `isPreflightRequest()`? (because the class name already includes 'CORS')

core/src/main/java/com/linecorp/armeria/server/DefaultPathMappingContext.java
```diff
@@ -57,20 +57,23 @@
...
+    private final boolean isPreflight;
     private final List<Object> summary;
     @Nullable
     private Throwable delayedCause;
 
     DefaultPathMappingContext(VirtualHost virtualHost, String hostname,
                               HttpMethod method, String path, @Nullable String query,
-                              @Nullable MediaType consumeType, @Nullable List<MediaType> produceTypes) {
+                              @Nullable MediaType consumeType, @Nullable List<MediaType> produceTypes,
+                              boolean isPreflight) {
```
trustin: - How about `isCorsPreflight`?
- How about adding overloaded constructor where `isCorsPreflight` is always `false`?

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
hyangtack: Sorry for missing comment. 🙇 
I referred to https://fetch.spec.whatwg.org/#http-responses
```
In case a server does not wish to participate in the CORS protocol, its HTTP response to the CORS or CORS-preflight request must not include any of the above headers. The server is encouraged to use the 403 status in such HTTP responses.
```
So I thought `403 forbidden` seems better than `404 not found` for a preflight request.
Let me add a comment about this!

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
hyangtack: Got it. Will update it. :-)

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -288,11 +291,31 @@ private static AnnotatedHttpServiceElement create(String pathPrefix, Object obje
                     return HttpStatus.valueOf(statusCode);
                 });
 
+        // A CORS preflight request can be received because we handle it exceptionally. The following
```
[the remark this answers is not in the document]
hyangtack: Fixed

adriancole:
lovely!

PR #1538, opened 2019-01-17
Prevent redundant injection with the same parameter in AnnotatedHttpS…

minwoox:
…ervice

Motivation:
Currently, the annotation paramter such as `@Param` and `@Header`, whose `name` is same,
can be annotated in multiple places which are constructor, methods, and fields.
This is incorrect usage so we should warn it and make the injection happens only once.

Modifications:
- Warn if a parameter used more than once in a bean when building `AnnotatedHttpService`
- Make the parameter injected only once by the order of consturctor, method and field

Result:
- Better usage

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedBeanFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedBeanFactory.java
@@ -27,12 +27,13 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.AbstractMap.SimpleImmutableEntry;
-import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
 
@@ -43,6 +44,8 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMap.Builder;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.MapMaker;
 
@@ -89,6 +92,25 @@ final class AnnotatedBeanFactory {
                                                                     : Optional.empty();
     }
 
+    static TreeSet<AnnotatedValueResolver> uniqueResolverSet() {
+        return new TreeSet<>((o1, o2) -> {
+            final String o1Name = o1.httpElementName();
+            final String o2Name = o2.httpElementName();
+            if (o1Name != null && o2Name != null && o1Name.equals(o2Name) &&
+                o1.annotationType() == o2.annotationType()) {
+                return 0;
+            }
+            // We are not ordering, but just finding duplicate elements.
+            return -1;
+        });
+    }
+
+    static void warnUsedRedundantly(AnnotatedValueResolver resolver, String genericString) {
+        logger.warn("Found redundantly used parameter in {}." +
+                    " httpElementName: {}, annotation: {}", genericString,
+                    resolver.httpElementName(), resolver.annotationType().getSimpleName());
+    }
+
     @Nullable
     private static <T> Function<ResolverContext, T> createFactory(BeanFactoryId beanFactoryId,
                                                                   List<RequestObjectResolver> objectResolvers) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedBeanFactory.java
```diff
@@ -89,6 +92,25 @@ static synchronized BeanFactoryId register(Class<?> clazz, Set<String> pathParam
...
+                o1.annotationType() == o2.annotationType()) {
+                return 0;
+            }
+            // We are not ordering, but just finding duplicate elements.
+            return -1;
+        });
+    }
+
+    static void warnUsedRedundantly(AnnotatedValueResolver resolver, String genericString) {
+        logger.warn("Found redundantly used parameter in {}." +
```
trustin: Found a redundant use of annotation in ...?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedBeanFactory.java
```diff
@@ -89,6 +92,25 @@ static synchronized BeanFactoryId register(Class<?> clazz, Set<String> pathParam
...
+            if (o1Name != null && o2Name != null && o1Name.equals(o2Name) &&
+                o1.annotationType() == o2.annotationType()) {
+                return 0;
+            }
+            // We are not ordering, but just finding duplicate elements.
+            return -1;
+        });
+    }
+
+    static void warnUsedRedundantly(AnnotatedValueResolver resolver, String genericString) {
```
trustin: warnRedundantUse?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedBeanFactory.java
```diff
@@ -141,17 +166,17 @@ static synchronized BeanFactoryId register(Class<?> clazz, Set<String> pathParam
                         constructor.getValue(), resolverContext);
                 final T instance = constructor.getKey().newInstance(constructorArgs);
 
-                for (final Entry<Field, AnnotatedValueResolver> field : fields) {
-                    final Object fieldArg = field.getValue().resolve(resolverContext);
-                    field.getKey().set(instance, fieldArg);
-                }
-
-                for (final Entry<Method, List<AnnotatedValueResolver>> method : methods) {
+                for (final Entry<Method, List<AnnotatedValueResolver>> method : methods.entrySet()) {
```
trustin: Could you add some comment about which is injected first? e.g. constructor, method, field

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedValueResolver.java
```diff
@@ -410,6 +424,16 @@ private static boolean isAnnotationPresent(AnnotatedElement element) {
                element.isAnnotationPresent(RequestObject.class);
     }
 
+    private static void warnIfUsedRedundantly(Executable constructorOrMethod,
```
trustin: warnOnRedundantUse?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceRequestConverterTest.java
```diff
@@ -513,6 +521,40 @@ public void validate() {
         }
     }
 
+    static class RequestBean4 {
+        private int foo;
+
+        // This field is not set because the foo param is already used in the constructor.
+        @Param("foo")
+        private int foo1;
```
trustin: - Could you update `RequestBean2`?
- How about adding a test case that makes sure `@Header("foo")` and `@Param("foo")` do not cause a problem?

minwoox: Updated Thanks~!

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedBeanFactory.java
minwoox: I added comments before `findMethods` and `findFields` that the method will warn and not include if there're parameters used redundantly.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedBeanFactory.java
```diff
@@ -223,13 +244,49 @@ static synchronized BeanFactoryId register(Class<?> clazz, Set<String> pathParam
...
                 if (!resolvers.isEmpty()) {
-                    ret.add(new SimpleImmutableEntry<>(method, resolvers));
+                    boolean redundant = false;
+                    for (AnnotatedValueResolver resolver : resolvers) {
+                        if (!uniques.add(resolver)) {
+                            redundant = true;
+                            warnRedundantUse(resolver, method.toGenericString());
+                        }
+                    }
+                    if (redundant && resolvers.size() == 1) {
```
hyangtack: Could you explain why `resolvers.size() == 1` condition is required?

PR #1539, opened 2019-01-18
Prevent `initCause()` on singleton exceptions

trustin:
Motivations:

Currently it is allowed to call `initCause()` on singleton exceptions
such as `ResponseTimeoutException`, e.g.

    ResponseTimeoutException e = ResponseTimeoutException.get();
    e.initCause(new Exception());

.. which should be prohibited because otherwise all timeouts will have
an extra cause when `verboseExceptions` is off.

Modifications:

- Made sure all singletone exceptions have its cause set to `null`.
- Updated `TimeoutException` and its subtypes to make it clear that they
  never have stack trace.
- Removed the redundant instantation logic in the subtypes of
  `TimeoutException`.
- Added `@Nullable` annotations where applicable.

Result:

- Fixes a bug where a user can attach a cause to a singleton exception.
- Fixes nullability of exception constructor parameters.


kojilin:
I'm checking RxJava2.
https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/exceptions/CompositeException.java#L136-L138

Looks like it will ignore overwrite exception, so maybe ok..hm.

PR #1541, opened 2019-01-20
Make javax.validation compileOnly

anuraaga:
Currently, users either need to provide `hibernate-validator` or exclude `javax.validation` when using `armeria-spring` because spring boot checks for the presence of the latter on the classpath and then requires something like the former. I think most users will have the dependencies if they want validation automatically without armeria enforcing it as a required dependency. Of course, exclusion is always an option too so maybe it's not a big deal. Thoughts?

Came up in https://github.com/openzipkin/zipkin/pull/2348


trustin:
Realized we can't remove hibernate-validator because our example requires it.

trustin:
Thanks, @anuraaga !

PR #1542, opened 2019-01-21
Fix NPE in HttpChannelPool

minwoox:
Motivation:
NPE is raised in `HttpChannelPool`. This is because the `HttpChannelPool` is designed to be accessed by just one thread, but it wasn't.
When the client uses HTTP/1 and subscribes the `HttpResponse` with another thread, it returns a connection to the `HttpChannelPool` with the different thread.

Modification:
- Fix to return the connection to the pool with the same eventloop

Result:
- Fewer NPEs

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
@@ -320,7 +320,7 @@ public final class ClientFactoryBuilder {
 
     /**
      * Sets whether to use <a href="https://en.wikipedia.org/wiki/HTTP_pipelining">HTTP pipelining</a> for
-     * HTTP/1 connections. This does not affect HTTP/2 connections. This option is enabled by default.
+     * HTTP/1 connections. This does not affect HTTP/2 connections. This option is disabled by default.
      */
     public ClientFactoryBuilder useHttp1Pipelining(boolean useHttp1Pipelining) {
         this.useHttp1Pipelining = useHttp1Pipelining;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -215,10 +215,10 @@ private void doExecute(PooledChannel pooledChannel, ClientRequestContext ctx,
                     // If pipelining is disabled, return after the response is fully received.
                     final CompletableFuture<Void> completionFuture =
                             factory.useHttp1Pipelining() ? req.completionFuture() : res.completionFuture();
-                    completionFuture.handle((ret, cause) -> {
+                    completionFuture.handleAsync((ret, cause) -> {
                         pooledChannel.release();
                         return null;
-                    });
+                    }, ctx.eventLoop());
```
trustin: Could we use direct invocation when the current thread is an event loop instead of always handling it asynchronously? For example, we could keep using `handle()` here and update the `PooledChannel.release()` implementation.

```java
@Override
public void release() {
    if (!eventLoop.inEventLoop()) {
        eventLoop.execute(() -> release0());
    } else {
        release0();
    }
}
```

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -320,7 +320,7 @@ public ClientFactoryBuilder useHttp2Preface(boolean useHttp2Preface) {
 
     /**
      * Sets whether to use <a href="https://en.wikipedia.org/wiki/HTTP_pipelining">HTTP pipelining</a> for
-     * HTTP/1 connections. This does not affect HTTP/2 connections. This option is enabled by default.
+     * HTTP/1 connections. This does not affect HTTP/2 connections. This option is disabled by default.
```
trustin: Thanks!

PR #1543, opened 2019-01-21
Remove deprecated methods in CircuitBreakerListener

minwoox:
Motivation
I made the methods in `CircuitBreakerListener` deprecated and add methods whose first parameter
is the `circuitBreakerName` to prevent another breaking change.
But it was not a good idea because I had to add new methods in `CircuitBreakerListener` so I cannot
avoid making breaking change anyway.

Modification
- Remove the deprecated methods

Result
- Neat APIs


minwoox:
Thanks, reviewer!

PR #1553, opened 2019-01-24
Enable DNS query traces by default / Add ClientFactoryBuilder.domainN…

### LINKED ISSUE #1512: DNS error with certain DNS servers that don't handle EDNS
Only on Windows, I often am unable to resolve DNS with Netty

```
Caused by: java.net.UnknownHostException: failed to resolve 'www.googleapis.com' after 3 queries
        at io.netty.resolver.dns.DnsResolveContext.finishResolve(DnsResolveContext.java:845)
        ... 32 more
```

Despite `nslookup www.googleapis.com` for example working fine.

This happens both when invoking commands from the command line in an `msys2` terminal as well as in IntelliJ.

Don't have much else to go on, any recommendations on how I can get some more logs or something to help debug?

### LINKED ISSUE #1552: Log DNS resolver traces
Related: #1512

A user may find it useful to log DNS resolver traces at DEBUG level.

/cc @anuraaga @mauhiz

trustin:
…ameResolverCustomizer()

Motivation:

- A user often wants to know why a domain name resolution attempt has
  failed in detail.
- `ClientFactoryBuilder.addressResolverGroupFactory()` provides a way to
  customize the `DomainNameResolver`, but it often involves more
  boilerplate code than a user expects.

Modifications:

- Added `ClientFactoryBuilder.domainNameResolverCustomizer()` which
  enables simpler customization of `DomainNameResolver`.
  - Made sure it is mutually exclusive with `addressResolverGroupFactory()`.
- Added `DefaultAddressResolverGroupFactory` which enables DNS query
  tracing by default.

Result:

- A user can easily figure out how a domain name resolution attempt
  failed by configuring the logging framework. For example, a Logback
  user can add the following:

      <logger name="io.netty.resolver.dns.TraceDnsQueryLifeCycleObserverFactory" level="DEBUG" />

- A user can easily customize `DomainNameResolver`, such as specifying
  non-default set of DNS servers.
- Closes #1512
- Closes #1552

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
@@ -16,7 +16,9 @@
 
 package com.linecorp.armeria.client;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_WINDOW_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.MAX_FRAME_SIZE_LOWER_BOUND;
 import static io.netty.handler.codec.http2.Http2CodecUtil.MAX_FRAME_SIZE_UPPER_BOUND;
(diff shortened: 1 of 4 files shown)


trustin:
> Are traces printed also for successful attempts?

No. Only failed or redirected queries are logged. I believe the amount of log will be small anyway because of cache.

> What would a sample trace look like?

See here for the complete logging code: https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/TraceDnsQueryLifecycleObserver.java

It's not a sample, but I guess you get the picture. Please feel free to let me know if you need a real sample.

trustin:
@minwoox Gentle ping

Issue #1550, opened 2019-01-23
Provide a way to configure additional response header using annotation


tmdgus0118:
I think it would be useful if a user can configure HTTP Response header via annotation, e.g.

```java
        Server s = new ServerBuilder()
                .annotatedService("/example", new Object() {
                    
                    @Get("/get")
                    @AdditionalHeader(name = "header_name", value = "value")
                    public void get() {}
                })
```

trustin:
Sounds like a great idea. :+1:

PR #1555, opened 2019-01-25
Support @AdditionalHeader for default response header in annotatedService.

tmdgus0118:
Motivation:
 Issues #1550 
 - Provide a way to configure default response header via `@AdditionalHeader`

Modification:
- Replaced `defaultResponseStatus` to `AnnotatedHttpServiceFactory` to create default `HttpHeaders` and it passes the headers to the constructor of `AnnotatedHttpService`.
- Added test about it.
- Added `@AdditionalHeader` & `@AdditionalTrailer`

Result:
- Usability

TODO: Add docs


core/src/main/java/com/linecorp/armeria/server/annotation/AdditionalHeader.java
```diff
@@ -16,12 +16,17 @@
...
 import java.lang.annotation.Target;
 
 /**
  * Annotation for an additional HTTP header.
  */
-@Target(ElementType.ANNOTATION_TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(AdditionalHeaders.class)
+@Target({ElementType.TYPE, ElementType.METHOD, ElementType.ANNOTATION_TYPE})
 public @interface AdditionalHeader {
```
trustin: How about changing the type of `value()` `String[]` instead of `String` so that a user can specify a header with multiple values?

```java
@AdditionalHeader(name = "foo", value = "bar")
@AdditionalHeader(name = "foo", value = "baz") // Does not work...
// vs.
@AdditionalHeader(name = "foo", value = { "bar", "baz" })
```

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -290,6 +340,14 @@ private static AnnotatedHttpServiceElement create(String pathPrefix, Object obje
...
                     return HttpStatus.valueOf(statusCode);
                 });
+        final HttpHeaders defaultHeaders = HttpHeaders.of(defaultResponseStatus(defaultResponseStatus, method));
+
+        final Consumer<AdditionalHeader> consumer = header -> {
+            defaultHeaders.set(HttpHeaderNames.of(header.name()), header.value());
+        };
+
+        findRepeatableAnnotations(clazz, AdditionalHeader.class).forEach(consumer);
+        findRepeatableAnnotations(method, AdditionalHeader.class).forEach(consumer);
```
trustin: Could we log a warning message when there are `@AdditionalHeader` annotations with a duplicate header name at the same level? e.g.

```java
@AdditionalHeader(name = "a", value = "b")
@AdditionalHeader(name = "c", value = "d")
@AdditionalHeader(name = "c", value = "e") // WARN
public class Foo {
    @AdditionalHeader(name = "a", value = "f") // DO NOT WARN
    @AdditionalHeader(name = "g", value = "h")
    @AdditionalHeader(name = "g", value = "i") // WARN
    public void method() {}
}
```

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
```diff
@@ -252,6 +254,7 @@ public Object expectOk() {
                 }
 
                 @Get("/expect-custom-header")
+                @AdditionalHeader(name = "x-custom-annotated-header", value = "annotated-value")
                 @ProducesJson
                 public HttpResult<Map<String, String>> expectCustomHeader() {
                     return HttpResult.of(HttpHeaders.of(AsciiString.of("x-custom-header"), "value"),
```
trustin: Global comment: For header names, you should use `HttpHeaderNames.of(CharSequence)`.

tmdgus0118:
> How about also adding `@AdditionalTrailer` and `@AdditionalTrailers` which adds additional trailer headers by taking advantage of this pull request?

Sounds good to me !

core/src/test/java/com/linecorp/armeria/server/cors/HttpServerCorsTest.java
```diff
@@ -222,7 +223,8 @@ public void testCorsDecoratorAnnotation() {
                      response.headers().get(HttpHeaderNames.ACCESS_CONTROL_ALLOW_HEADERS));
         assertEquals("*", response.headers().get(HttpHeaderNames.ACCESS_CONTROL_ALLOW_ORIGIN));
         assertEquals("3600", response.headers().get(HttpHeaderNames.ACCESS_CONTROL_MAX_AGE));
-        assertEquals("Hello CORS", response.headers().get(HttpHeaderNames.of("x-preflight-cors")));
+        assertEquals("[Hello CORS, Hello CORS2]",
+                     response.headers().getAll(HttpHeaderNames.of("x-preflight-cors")).toString());
```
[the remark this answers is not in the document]
tmdgus0118: Thanks for the good tip!

tmdgus0118:
> Looks good all in all. Please address the comments. Are you going to work on its documentation in a separate pull request?

Thanks for the reviews,  I'm going to add to this PR :)

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -215,6 +223,86 @@
...
+        return Optional.of(repeatable.get().value());
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T extends Annotation> List<T> findRepeatableAnnotations(AnnotatedElement element,
+                                                                            Class<T> annotationType) {
+        requireNonNull(element, "element");
+        requireNonNull(annotationType, "annotationType");
+        final Class<? extends Annotation> repeatable = getRepeatable(annotationType).orElseThrow(
+                () -> new IllegalArgumentException("annotationType is not an repeatable annotation."));
```
minwoox: an -> a?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -215,6 +223,86 @@
...
+                            " It will be ignored.",
+                            header.name(), level);
+                return;
+            }
+            headers.set(HttpHeaderNames.of(header.name()), header.value());
+            addedHeaderSets.add(header.name());
+        });
+    }
+
+    private static void setAdditionalTrailer(HttpHeaders headers, AnnotatedElement element, String level) {
```
minwoox: Perhaps we can combine these two methods into one adding a type parameter which indicates whether it is `AdditionalHeader.class` or `AdditionalTrailer.class` so that we can reduce the duplicated code?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,54 @@ Then, you can annotate your service method with your annotation as follows.
         @MyProducibleType  // the same as @Produces("application/xml")
         public MyResponse hello(MyRequest myRequest) { ... }
     }
+
+
+Annotation-based HTTP Response header/trailer
```
minwoox: headers/trailers? (plural)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,54 @@ Then, you can annotate your service method with your annotation as follows.
...
+    @AdditionalHeader(name = "custom-header", value = "custom-value")
+    @AdditionalTrailer(name = "custom-trailer", value = "custom-value")
+    public class MyAnnotatedService {
+        @Get("/hello")
+        @AdditionalHeader(name = "custom-header-2", value = "custom-value")
+        @AdditionalTrailer(name = "custom-trailer-2", value = "custom-value")
+        public HttpResponse hello() { ... }
+    }
+
+The :api:`@AdditionalHeader` or :api:`@AdditionalTrailer` specified at the method level take precedence
```
minwoox: take's' precedence over?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,54 @@ Then, you can annotate your service method with your annotation as follows.
...
+what's specified at the class level if they have the same name, e.g.
```
minwoox: if it has?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
trustin: e.g.

```java
private static void setAdditionalHeader(Class<?> annotationType, HttpHeaders headers, AnnotatedElement element, String level) {
    ...
    findRepeatableAnnotations(element, annotationType).forEach(header -> {
    });
    ...
}
```

core/src/main/java/com/linecorp/armeria/server/annotation/ResponseConverterFunction.java
```diff
@@ -36,6 +36,19 @@
      * Returns {@link HttpResponse} instance corresponds to the given {@code result}.
      * Calls {@link ResponseConverterFunction#fallthrough()} or throws a {@link FallthroughException} if
      * this converter cannot convert the {@code result} to the {@link HttpResponse}.
+     *
+     * @param headers The HTTP headers of the {@link HttpResult} returned by the service method.
+     *                Even if the service method does not return any {@link HttpResult},
+     *                the headers will basically include {@code "status"} and {@code "content-type"}
+     *                HTTP Response Header.
+     *                Note that the additional headers will not be included in this headers,
+     *                call {@link ServiceRequestContext#additionalResponseHeaders()} to access them.
```
minwoox: I think we should explain the default status is `200` unless a user specifies it with `@StatusCode`.
I'm not sure that this is best describable, but how about this?

```
@param headers the basic HTTP headers that you might want to use to create the {@link HttpResponse} with.
     *                The status of headers is {@link HttpStatus#OK} by default, unless you specify it with
     *                {@link StatusCode} on the method. The headers also will include a {@link MediaType}, if
     *                {@link ServiceRequestContext#negotiatedResponseMediaType()} returns it.
     *                If the method that you decorate with this function returns an {@link HttpResult},
     *                this headers is the same header from {@link HttpResult#headers()}.
     *                The headers could be immutable, so please call {@link HttpHeaders#toMutable()} before you modify it.
     *                Please note that the additional headers which are set by {@link ServiceRequestContext#addAdditionalResponseHeader(AsciiString, String)}
     *                and {@link AdditionalHeaders} are not included in this header.
```

hyangtack: @minwoox 👍 
How about this?
- If the method that you decorate with this function returns -> If the annotated method returns

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -215,6 +223,83 @@
...
+        requireNonNull(element, "element");
+        requireNonNull(level, "level");
+
+        final Set<String> addedHeaderSets = new HashSet<>();
+        findRepeatableAnnotations(element, annotation).forEach(header -> {
+            final String name = getAnnotationField(header, "name", String.class);
+            final String[] value = getAnnotationField(header, "value", String[].class);
+
+            if (addedHeaderSets.contains(name)) {
+                logger.warn("The additional {} named '{}' is already added at the same {} level." +
```
trustin: - `is already added at the same {} level. It will be ignored.` -> `is set at the same {} level already; ignoring.`
- How about telling the class name (and method) which has the problem as well?  e.g. in the format of `<className>.<methodName>()`

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -290,6 +375,21 @@ private static AnnotatedHttpServiceElement create(String pathPrefix, Object obje
...
+
+        final HttpHeaders defaultTrailingHeaders = HttpHeaders.of();
+
+        setAdditionalHeader(defaultHeaders, clazz, "header", "class", AdditionalHeader.class);
+        setAdditionalHeader(defaultHeaders, method, "header", "method", AdditionalHeader.class);
+        setAdditionalHeader(defaultTrailingHeaders, clazz, "trailer", "class", AdditionalTrailer.class);
+        setAdditionalHeader(defaultTrailingHeaders, method, "trailer", "method", AdditionalTrailer.class);
+
+        if (ArmeriaHttpUtil.isContentAlwaysEmpty(defaultHeaders.status()) && !defaultTrailingHeaders
+                .isEmpty()) {
```
trustin: Could look prettier if line is broken right after `&&`.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -290,6 +375,21 @@ private static AnnotatedHttpServiceElement create(String pathPrefix, Object obje
...
+            logger.warn("The content for HTTP status ({}) should be always empty. " +
+                        "Trailing headers for the method '{}' might be ignored.",
```
trustin: - `A response with HTTP status code '{}' cannot have a content. Trailing headers defined at '{}' will be ignored.`
- How about also mentioning the class name? e.g. in the format of `<className>.<methodName>()`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,54 @@ Then, you can annotate your service method with your annotation as follows.
         @MyProducibleType  // the same as @Produces("application/xml")
         public MyResponse hello(MyRequest myRequest) { ... }
     }
+
+
+Annotation-based HTTP Response headers/trailers
```
trustin: How about: `Specifying additional response headers`? I think we don't need to mention 'annotation-based' because this page is all about annotation-based services.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,54 @@ Then, you can annotate your service method with your annotation as follows.
...
+-----------------------------------------------
+
+Armeria provides a way to configure additional headers/trailers via annotation,
+:api:`@AdditionalHeader` for HTTP headers and :api:`@AdditionalTrailer` for HTTP trailers.
+
+You can annotate your service method with the annotations as follows.
+
+.. code-block:: java
+
+    @AdditionalHeader(name = "custom-header", value = "custom-value")
```
trustin: How about adding imports for the two annotations?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,54 @@ Then, you can annotate your service method with your annotation as follows.
...
+    public class MyAnnotatedService {
+        @Get("/hello")
+        @AdditionalHeader(name = "custom-header", value = "custom-overwritten")
+        @AdditionalTrailer(name = "custom-trailer", value = "custom-overwritten")
+        public HttpResponse hello() { ... }
+    }
+
+In this case, the values of the HTTP header named ``custom-header`` and the HTTP trailer named ``custom-trailer`` will be ``custom-overwritten``, not ``custom-value``.
+
+Note that the trailers will not be injected with the following HTTP status code:
```
trustin: injected into the responses with the following HTTP status code, because they always have an empty content.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -252,7 +252,8 @@ private static HttpStatus defaultResponseStatus(Optional<HttpStatus> defaultResp
             return annotation.getClass().getMethod(fieldName).invoke(annotation);
         } catch (Exception ex) {
             throw new IllegalStateException(
-                    String.format("%s() method invocation throws an exception", fieldName), ex);
+                    String.format("'%s.%s()' method invocation throws an exception",
```
trustin: `method invocation throws an exception` -> `raised an exception.`

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -290,9 +292,9 @@ private static HttpStatus defaultResponseStatus(Optional<HttpStatus> defaultResp
             final String[] value = getAnnotationField(header, "value", String[].class);
 
             if (addedHeaderSets.contains(name)) {
-                logger.warn("The additional {} named '{}' is already added at the same {} level." +
-                            " It will be ignored.",
-                            clsAlias, name, level);
+                logger.warn("The additional {} named '{}' at '{}' is set at the same {} level;" +
```
trustin: `;` -> ` already;`

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1233,8 +1233,8 @@ Then, you can annotate your service method with your annotation as follows.
     }
 
 
-Annotation-based HTTP Response headers/trailers
------------------------------------------------
+Specifying additional HTTP Response headers/trailers
```
trustin: - Response -> response
- OK to omit 'HTTP'?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,58 @@ Then, you can annotate your service method with your annotation as follows.
...
+    @AdditionalHeader(name = "custom-header", value = "custom-value")
+    @AdditionalTrailer(name = "custom-trailer", value = "custom-value")
+    public class MyAnnotatedService {
+        @Get("/hello")
+        @AdditionalHeader(name = "custom-header-2", value = "custom-value")
+        @AdditionalTrailer(name = "custom-trailer-2", value = "custom-value")
+        public HttpResponse hello() { ... }
+    }
+
+The :api:`@AdditionalHeader` or :api:`@AdditionalTrailer` specified at the method level takes precedence
```
hyangtack: nit: takes precedence **over**?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -252,7 +252,7 @@ private static HttpStatus defaultResponseStatus(Optional<HttpStatus> defaultResp
             return annotation.getClass().getMethod(fieldName).invoke(annotation);
         } catch (Exception ex) {
             throw new IllegalStateException(
-                    String.format("'%s.%s()' method invocation throws an exception",
+                    String.format("'%s.%s()' method invocation raised an exception",
```
trustin: - 'invocation' could be omitted.
- missing period

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
```diff
@@ -196,21 +196,19 @@ static HttpHeaders copyOf(HttpHeaders headers) {
...
-     *     }
+     * headers.names().forEach(name -> {
+     *      if (!contains(name)) {
+     *          set(name, headers.getAll(name));
+     *      }
      * });
      * }</pre>
      */
-    default HttpHeaders setAllIfAbsent(Headers<? extends AsciiString, ? extends String, ?> headers) {
+    default HttpHeaders setAllIfAbsent(Headers<AsciiString, ? extends String, ?> headers) {
```
minwoox: `? extends String` -> `String`?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,58 @@ Then, you can annotate your service method with your annotation as follows.
...
+        @Get("/hello")
+        @AdditionalHeader(name = "custom-header", value = "custom-overwritten")
+        @AdditionalTrailer(name = "custom-trailer", value = "custom-overwritten")
+        public HttpResponse hello() { ... }
+    }
+
+In this case, the values of the HTTP header named ``custom-header`` and the HTTP trailer named ``custom-trailer`` will be ``custom-overwritten``, not ``custom-value``.
+
+Note that the trailers will not be injected into the responses with the following HTTP status code,
+because they always have an empty content.
```
minwoox: they always do not have the content.?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -215,6 +223,85 @@
...
+                                                                   String clsAlias,
+                                                                   String elementAlias,
+                                                                   String level,
+                                                                   Class<T> annotation) {
+        requireNonNull(headers, "headers");
+        requireNonNull(element, "element");
+        requireNonNull(level, "level");
+
+        final Set<String> addedHeaderSets = new HashSet<>();
+        findRepeatableAnnotations(element, annotation).forEach(header -> {
```
hyangtack: I'm not sure if it works. But would you try to apply `AnnotationUtil#findAll` for this?

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java#L125

core/src/main/java/com/linecorp/armeria/server/annotation/AdditionalHeader.java
```diff
@@ -16,12 +16,17 @@
...
+import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
  * Annotation for an additional HTTP header.
  */
-@Target(ElementType.ANNOTATION_TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(AdditionalHeaders.class)
+@Target({ElementType.TYPE, ElementType.METHOD, ElementType.ANNOTATION_TYPE})
```
hyangtack: nit: `ElementType.ANNOTATION_TYPE` is not necessary because `ElementType.TYPE` includes it.

core/src/main/java/com/linecorp/armeria/server/annotation/AdditionalTrailer.java
```diff
@@ -0,0 +1,41 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for an additional HTTP trailing header.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(AdditionalTrailers.class)
+@Target({ ElementType.TYPE, ElementType.METHOD, ElementType.ANNOTATION_TYPE})
```
hyangtack: Ditto

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1232,3 +1231,58 @@ Then, you can annotate your service method with your annotation as follows.
...
+        @Get("/hello")
+        @AdditionalHeader(name = "custom-header", value = "custom-overwritten")
+        @AdditionalTrailer(name = "custom-trailer", value = "custom-overwritten")
+        public HttpResponse hello() { ... }
+    }
+
+In this case, the values of the HTTP header named ``custom-header`` and the HTTP trailer named ``custom-trailer`` will be ``custom-overwritten``, not ``custom-value``.
+
+Note that the trailers will not be injected into the responses with the following HTTP status code,
+because they do always have an empty content.
```
trustin: OK to omit `do`?

core/src/main/java/com/linecorp/armeria/server/annotation/ResponseConverterFunction.java
```diff
@@ -36,6 +40,28 @@
      * Returns {@link HttpResponse} instance corresponds to the given {@code result}.
      * Calls {@link ResponseConverterFunction#fallthrough()} or throws a {@link FallthroughException} if
      * this converter cannot convert the {@code result} to the {@link HttpResponse}.
+     *
+     * @param headers The basic HTTP headers that you might want to use to create the {@link HttpResponse}.
```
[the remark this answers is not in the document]
trustin: OK to omit `basic`?

core/src/main/java/com/linecorp/armeria/server/annotation/ResponseConverterFunction.java
```diff
@@ -36,6 +40,28 @@
...
+     *                The headers also will include a {@link MediaType} if
+     *                {@link ServiceRequestContext#negotiatedResponseMediaType()} returns it.
+     *                If the method returns {@link HttpResult}, this headers is the same headers from
+     *                {@link HttpResult#headers()}
+     *                The headers could be immutable, so please call {@link HttpHeaders#toMutable()}.
+     *                Please note that the additional headers set by
+     *                {@link ServiceRequestContext#addAdditionalResponseHeader(AsciiString, String)}
+     *                and {@link AdditionalHeader} are not included in this headers.
+     * @param result The result of the service method.
+     * @param trailingHeaders The basic HTTP trailers that you might want to use to create the
```
trustin: OK to omit `basic`? (or `default`?)

trustin:
Thanks, @tmdgus0118 !

PR #1560, opened 2019-01-29
Support an alias for a bunch of annotations

### LINKED ISSUE #1224: Support an alias for a bunch of annotations
Example
Define an alias annotation like below:
```
@ProduceJson
@LoggingDecorator(...)
@interface CategoryX {}
```

Use the alias like below:
```
@CategoryX
@Get(...)
public void ... {}
```

hyangtack:
Motivation:
A user may want to define his or her annotations as a custom annotation in order to apply the custom annotation to his or her annotated HTTP services. e.g:
```
// Define a custom annotation:
@ProducesJson
@LoggingDecorator
@interface MyApiSpecification {}

// Apply it to the annotated HTTP service:
@Get("/api")
@MyApiSpecification
public Something getSomething() {}
```

Modifications:
- Add `AnnotationUtil` for finding a specific annotation or getting all annotations from an `AnnotatedElement`.
- Use `AnnotationUtil` instead of `ReflectionUtils.getAllAnnotations()`. The old one does not respect the order of the annotations because its return type is a `Set<Annotation>` (more specifically it's `HashSet`).

Result:
- Won't depend on the `ReflectionUtils`'s implementation when getting annotations.
- Closes #1224

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
@@ -77,11 +77,14 @@ final class AnnotationUtil {
                             Retention.class, Repeatable.class, Target.class);
 
     /**
-     * Returns an annotation of the {@code annotationType} or {@link Optional#empty()}, which is found
-     * from the specified {@code element}.
-     *
-     * <p>Note that this method will <em>not</em> find annotations from both the super classes of
-     * the {@code element} and the meta-annotations.
+     * Returns an annotation of the {@code annotationType} if it is found from one of the following:
+     * <ul>
+     *     <li>the specified {@code element}</li>
+     *     <li>the super classes of the specified {@code element} if the {@code element} is a class</li>
+     *     <li>the meta-annotations of the annotations specified on the {@code element}
+     *     or its super classes</li>
+     * </ul>
+     * Otherwise, {@link Optional#empty()} will be returned.
      *
      * @param element the {@link AnnotatedElement} to find annotations
      * @param annotationType the type of the annotation to find
@@ -108,11 +111,13 @@ final class AnnotationUtil {
     }
 
     /**
-     * Returns the annotations of the {@code annotationType}, which are found from the specified
-     * {@code element}.
-     *
-     * <p>Note that this method will find annotations from both the super classes of the {@code element}
-     * and the meta-annotations.
+     * Returns all annotations of the {@code annotationType} which are found from the following.
+     * <ul>
+     *     <li>the specified {@code element}</li>
+     *     <li>the super classes of the specified {@code element} if the {@code element} is a class</li>
+     *     <li>the meta-annotations of the annotations specified on the {@code element}
+     *     or its super classes</li>
+     * </ul>
      *
      * @param element the {@link AnnotatedElement} to find annotations
      * @param annotationType the type of the annotation to find
@@ -123,11 +128,13 @@ final class AnnotationUtil {
     }
 
     /**
-     * Returns the annotations of the {@code annotationType}, which are found from the specified
-     * {@code element}.
+     * Returns all annotations of the {@code annotationType} if which are found from the following.
+     * <ul>
+     *     <li>the specified {@code element}</li>
+     *     <li>the super classes of the specified {@code element} if the {@code element} is a class</li>
+     * </ul>
      *
-     * <p>Note that this method will find annotations from the super classes of the {@code element}
-     * but will <em>not</em> find annotations from the meta-annotations.
+     * <p>Note that this method will <em>not</em> find annotations from the meta-annotations.
      *
      * @param element the {@link AnnotatedElement} to find annotations
      * @param annotationType the type of the annotation to find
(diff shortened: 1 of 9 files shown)


dependencies.yml
```diff
@@ -2,6 +2,9 @@
 # NB: Update NOTICE.txt and add/remove LICENSE.*.txt when adding/removing a dependency.
 #
 
+cglib:
+  cglib: { version: '3.2.10' }
```
trustin: Perhaps we should update NOTICE.txt and add LICENSE.txt?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
```diff
@@ -0,0 +1,358 @@
...
+    /**
+     * Built-in meta annotations defined in {@code java.lang.annotation} package.
+     */
+    private static final Set<Class<? extends Annotation>> BUILTIN_META_ANNOTATIONS =
+            ImmutableSet.of(Documented.class, Inherited.class, Native.class,
+                            Retention.class, Repeatable.class, Target.class);
+
+    /**
+     * Returns an annotation of the {@code annotationType} or {@link Optional#empty()}, which is found
+     * from the specified {@code element}.
```
trustin: I find the differences between `find*()`, `find*Declared()`, `find*Inherited()` are not clear looking from the Javadoc. Could you update the method description so that the user of this class understand the differences better?

dependencies.yml
hyangtack: I missed them. Thanks!

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
```diff
@@ -0,0 +1,369 @@
...
+     * Returns an annotation of the {@code annotationType} if it is found from one of the following:
+     * <ul>
+     *     <li>the specified {@code element}</li>
+     *     <li>the super classes of the specified {@code element} if the {@code element} is a class</li>
+     *     <li>the meta-annotations of the annotations specified on the {@code element}
+     *     or its super classes</li>
+     * </ul>
+     * Otherwise, {@link Optional#empty()} will be returned.
+     *
+     * @param element the {@link AnnotatedElement} to find annotations
```
minwoox: to find the first annotation?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    /**
+     * Returns an annotation of the {@code annotationType} if it is found from the specified {@code element}.
+     * Otherwise, the {@link Optional#empty()} will be returned.
+     *
+     * <p>Note that this method will <em>not</em> find annotations from both the super classes of
+     * the {@code element} and the meta-annotations.
+     *
+     * @param element the {@link AnnotatedElement} to find annotations
```
minwoox: ditto?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
```diff
@@ -0,0 +1,369 @@
...
+     * @param annotationType the type of the annotation to find
+     */
+    static <T extends Annotation> List<T> findAll(AnnotatedElement element, Class<T> annotationType) {
+        return find(element, annotationType, EnumSet.of(FindOption.LOOKUP_SUPER_CLASSES,
+                                                        FindOption.LOOKUP_META_ANNOTATIONS));
+    }
+
+    /**
+     * Returns all annotations of the {@code annotationType} if which are found from the following.
```
minwoox: Could remove `if`?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
```diff
@@ -0,0 +1,369 @@
...
+         * given {@link AnnotatedElement} will be collected at last. This option will work only with
+         * the {@link #LOOKUP_SUPER_CLASSES}.
+         */
+        COLLECT_SUPER_CLASSES_FIRST
+    }
+
+    /**
+     * Built-in meta annotations defined in {@code java.lang.annotation} package.
+     */
+    private static final Set<Class<? extends Annotation>> BUILTIN_META_ANNOTATIONS =
```
minwoox: `BUILT_IN`? not sure if it's correct but we have `BuiltInProperties` class. >_<

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
hyangtack: Fixed. Thanks!

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
hyangtack: Yes. It would be better. Fixed! Thanks! :-)

PR #1566, opened 2019-01-30
Do not set Content-length header when trailing headers exists

### LINKED ISSUE #1554: HttpResponse's trailing headers are not sent.
Originally reported by @southernkasaist

```java
// Trailer is sent.
ctx.addAdditionalResponseTrailer(HttpHeaderNames.WWW_AUTHENTICATE,
                                 "Basic realm=\"Authentication required for this page\"");
return HttpResponse.of(HttpStatus.UNAUTHORIZED,
                       MediaType.HTML_UTF_8,
                       HttpData.ofUtf8("401 Forbidden"));

// Trailer is not set.
return HttpResponse.of(HttpStatus.UNAUTHORIZED,
                       MediaType.HTML_UTF_8,
                       HttpData.ofUtf8("401 Forbidden"),
                       HttpHeaders.of(HttpHeaderNames.WWW_AUTHENTICATE,
                                      "Basic realm=\"Authentication required for this page\""));
```

In the second case, after writing `HttpData.ofUtf8("401 Forbidden")`, connection was closed and the trailing headers of `WWW-Authenticate` was subscribed but not written to response.

minwoox:
Motivations:
Some of the client implementations such as `curl` ignores trailing headers if the `Content-length` header is present.
We should not set `Content-length` header when trailing headers exists so that those clients can receive the trailing headers.

Modification:
- Do not set `Content-length` header when trailing headers exists
- Make a backlist which must not be included in trailing headers
  - https://tools.ietf.org/html/rfc7230#section-4.1.2
- Allow `Trailer` headers in response headers which was not before
- Set `Content-length` to `0` when the `HttpResponse` contains only a status so that the response is not transfer encoding

Result:
- Closes #1554

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
@@ -17,7 +17,7 @@
 package com.linecorp.armeria.common;
 
 import static com.linecorp.armeria.internal.ArmeriaHttpUtil.isContentAlwaysEmpty;
-import static com.linecorp.armeria.internal.ArmeriaHttpUtil.isContentAlwaysEmptyWithValidation;
+import static com.linecorp.armeria.internal.ArmeriaHttpUtil.setOrRemoveContentLengthInResponseHeaders;
 import static java.util.Objects.requireNonNull;
 
 import java.nio.charset.StandardCharsets;
@@ -217,13 +217,6 @@ public interface HttpResponse extends Response, StreamMessage<HttpObject> {
         requireNonNull(content, "content");
 
         final HttpHeaders headers = HttpHeaders.of(status).contentType(mediaType);
-        if (trailingHeaders.isEmpty()) {
-            // Some of the client implementation such as "curl" ignores trailing headers if
-            // the Content-length header is present. So set Content-length header only if trailingHeaders
-            // is empty.
-            headers.setInt(HttpHeaderNames.CONTENT_LENGTH, content.length());
-        }
-
         return of(headers, content, trailingHeaders);
     }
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -255,20 +259,32 @@ static HttpResponse of(HttpHeaders headers, HttpData content, HttpHeaders traili
...
-        if (isContentAlwaysEmptyWithValidation(status, content, trailingHeaders)) {
+        if (isContentAlwaysEmptyWithValidation(status, content, trailingHeaders) ||
+            (content.isEmpty() && trailingHeaders.isEmpty())) {
             ReferenceCountUtil.safeRelease(content);
-            return new OneElementFixedHttpResponse(headers);
-        } else if (!content.isEmpty()) {
-            if (trailingHeaders.isEmpty()) {
-                return new TwoElementFixedHttpResponse(headers, content);
+
+            final HttpHeaders mutableHeaders = headers.toMutable();
```
hyangtack: nit: How about calling `toMutable()` inside the below `if` statement?

minwoox:
All fixed. Thank you!

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -665,6 +723,57 @@ public static void toNettyHttp1(
...
+                                                       HttpHeaders trailingHeaders) {
+        requireNonNull(headers, "headers");
+        requireNonNull(content, "content");
+        requireNonNull(trailingHeaders, "trailingHeaders");
+
+        final HttpStatus status = headers.status();
+        assert status != null;
+
+        final HttpHeaders mutable = headers.toMutable();
+        if (isContentAlwaysEmptyWithValidation(status, content, trailingHeaders)) {
```
minwoox: I had to put this first to validate the status. >_<

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -723,6 +723,54 @@ public static void toNettyHttp1(
         }
     }
 
+    /**
+     * Returns a {@link HttpHeaders} which {@link HttpHeaderNames#CONTENT_LENGTH} is added or removed
```
trustin: - a -> an
- which -> whose

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -583,12 +626,24 @@ private static void setHttp2Scheme(io.netty.handler.codec.http.HttpHeaders in, U
...
      */
-    public static Http2Headers toNettyHttp2(HttpHeaders in) {
+    public static Http2Headers toNettyHttp2(HttpHeaders in, boolean server) {
         final Http2Headers out = new DefaultHttp2Headers(false, in.size());
-        out.set(in);
-        out.remove(HttpHeaderNames.CONNECTION);
-        out.remove(HttpHeaderNames.TRANSFER_ENCODING);
-        out.remove(HttpHeaderNames.TRAILER);
+
+        // Trailering headers if it does not have :status.
```
hyangtack: nit: Trailering -> Trailing

trustin:
Thanks, @minwoox !

PR #1568, opened 2019-01-30
Remove spring-boot-starter-actuator dependency from spring-boot modules

imasahiro:
Motivations:
- Sometimes spring-boot-actuator pulls unrelated metrics to MeterRegistry. For example, if a user add `armeria-spring-boot-starter` dependency to an application, because it also pulls spring-boot-actuator dependency, [KafkaMetricsAutoConfiguration](https://docs.spring.io/spring-boot/docs/2.2.0.BUILD-SNAPSHOT/api/org/springframework/boot/actuate/autoconfigure/metrics/KafkaMetricsAutoConfiguration.html) is enabled if kafka client bean is existed. And it causes unintentional conflicts with user metrics configurations. 

Changes:
- Remove spring-boot-starter-actuator dependency from spring-boot modules

Result:
- Less surprise

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -390,6 +390,7 @@ org.springframework.boot:
     version: &SPRING_BOOT_VERSION '2.1.2.RELEASE'
     javadocs:
     - https://docs.spring.io/spring/docs/current/javadoc-api/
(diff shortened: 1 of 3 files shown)


hyangtack:
@imasahiro Sounds good to me. Would you check the build failure?

PR #1569, opened 2019-01-31
Improve AnnotationUtil

hyangtack:
Motivation:
Follow-up work after #1560.

Modifications:
- `AnnotationUtil` will lookup meta-annotations of a meta-annotation to support the following case:
```
@ProducesJson
@Produces("text/plain")
@interface MyPostServiceSpecifications {}
```
  - `@ProducesJson` will also be found with this PR, which wasn't found before.
- Add `ElementType.ANNOTATION_TYPE` to `Target` annotation for `@Order` and `@StatusCode` so that a user can use them as a meta-annotation for his or her custom annotation.


trustin:
Thanks, @hyangtack!

PR #1575, opened 2019-02-08
Tell why a request is bad for a 400 Bad Request response

trustin:
Motivation:

It is sometimes hard for a client to know why a request is bad.

Modifications:

- Added short error messages to 400 Bad Request responses
- Used `ProtocolViolationException` instead of `IllegalArgumentException` for bad requests.
- Removed unnecessary validation of HTTP method in `HttpServerHandler`.
- Cleaned up response generation in `HttpServerHandler`.

Result:

- User friendliness.


trustin:
Thanks for reviewing.

PR #1576, opened 2019-02-08
Add more shortcut methods for convenience

trustin:
Motivation:

Skimmed through our test cases and thought a few shortcut methods would
be useful to reduce some tedious repetition.

Modifications:

- Added `contentType()` to:
  - `HttpRequest`
  - `AggregatedHttpMessage`
- Added `contentUtf8()`, `contentAscii()` and `content(Charset)` to
  `AggregatedHttpMessage`.
- Accepted `CharSequence` as well as String in:
  - `HttpRequest.of()`
  - `HttpRespons.of()`
  - `AggregatedHttpMessage.of()`
- Replaced the old usage with the new concise usage.

Result:

Brevity.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
@@ -107,8 +107,24 @@ public interface HttpResponse extends Response, StreamMessage<HttpObject> {
      * @param mediaType the {@link MediaType} of the response content
      * @param content the content of the response
      */
+    static HttpResponse of(HttpStatus status, MediaType mediaType, CharSequence content) {
+        if (content instanceof String) {
+            return of(status, mediaType, (String) content);
+        }
+
+        return of(status, mediaType,
+                  HttpData.of(mediaType.charset().orElse(StandardCharsets.UTF_8), content));
+    }
+
+    /**
+     * Creates a new HTTP response of the specified {@link HttpStatus} and closes the stream.
+     *
+     * @param mediaType the {@link MediaType} of the response content
+     * @param content the content of the response
+     */
     static HttpResponse of(HttpStatus status, MediaType mediaType, String content) {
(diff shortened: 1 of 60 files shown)


core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -101,14 +101,30 @@ static HttpResponse of(HttpStatus status) {
...
+     *
+     * @param mediaType the {@link MediaType} of the response content
+     * @param content the content of the response
+     */
+    static HttpResponse of(HttpStatus status, MediaType mediaType, CharSequence content) {
+        if (content instanceof String) {
+            return of(status, mediaType, (String) content);
+        }
+
+        return of(status, mediaType,
```
minwoox: `requireNonNull(mediaType, "mediaType");`?

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -101,14 +101,30 @@ static HttpResponse of(HttpStatus status) {
...
     /**
      * Creates a new HTTP response of the specified {@link HttpStatus} and closes the stream.
      *
      * @param mediaType the {@link MediaType} of the response content
      * @param content the content of the response
      */
     static HttpResponse of(HttpStatus status, MediaType mediaType, String content) {
-        return of(status, mediaType, content.getBytes(mediaType.charset().orElse(StandardCharsets.UTF_8)));
+        return of(status, mediaType,
+                  HttpData.of(mediaType.charset().orElse(StandardCharsets.UTF_8), content));
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
trustin: Fixed

trustin:
Thanks, @minwoox !

Issue #1586, opened 2019-02-12
Response closed without grpc-status (Headers only)


AhmadVatani:
Hi,
For send error to client, I'm using below code:
`responseObserver.onError(Status.INVALID_ARGUMENT.withDescription("Error description").asRuntimeException())`

In grpc web I got that error with below statusMessage and status 2:
`Response closed without grpc-status (Headers only)`

What should i do for get pretty format of error in grpc web?

trustin:
That looks like an unexpected behavior to me. Do you have a simple code that reproduces the problem, because we didn't see such an error before. /cc @anuraaga

AhmadVatani:
We are using [grpc-web](https://github.com/improbable-eng/grpc-web) implementation for TypeScript in front-end.

Here is grpc calling in TypeScript:
`grpc.unary(ProfileSecureService.ReadProfile, {
        host: ... ,
        request: ... ,
        metadata: ...,
        onEnd: res => {
          console.log(res);
        }
      }); `

And When get exception, in console log:

![screen shot 2019-02-12 at 16 07 45](https://user-images.githubusercontent.com/13574768/52636017-aeb28d80-2ee0-11e9-9128-8ba70b79f2db.png)

anuraaga:
Thanks will take a look. Also for reference, what version of the typescript gRPC-Web are you using?

AhmadVatani:
@anuraaga 
[grpc-web version 0.7.0](https://github.com/improbable-eng/grpc-web/releases/tag/0.7.0)
And Armeria version 0.79.0

iEAmi:
@anuraaga 
Armeria send grpc status in HTTP trailers but some browsers not support HTTP/2.0 trailers. I think this make problem.
I don't know netty-grpc how send grpc status!!!!!

anuraaga:
@AhmadVatani I have added a test case for grpc-web with an error in #1588. Can you compare it's business logic to yours, specifically https://github.com/line/armeria/blob/master/grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java#L159 ?

The server seems to be returning a correct response so I'm not sure why you're seeing the strange behavior on the browser. If you don't see anything obvious, it would be really great if you could set up a small repro with a HTML/JS and Java that could show the error - it'll take me some time to try to get those components set up to try myself, and I worry I won't be able to reproduce the error ><.

[...]

PR #1588, opened 2019-02-13
Add test for grpc-web error case.

anuraaga:
This test case exercises grpc-web with a top-level error. I don't see anything wrong with the server behavior at least for this case.

For #1586


trustin:
Build failure unrelated with this PR. Merging.

PR #1589, opened 2019-02-13
Add metrics related to timeout

### LINKED ISSUE #1580: Add timeout counter in RequestMetricSupport
/cc @kshgizmo

minwoox:
Motivation:
It would be nice if we provide timeout metrics by default.

Modifications:
- Add `requestTimeout` metric in server-side
- Add `responseTimeout` and `writeTimeout` metrics in client-side
- Add `ServiceRequestContext.isRequestTimedOut()`
  - Deprecate `RequestContext.isTimedOut()`

Result:
- More metrics
- Close #1580

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/AbstractRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/common/AbstractRequestContext.java
@@ -27,6 +27,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 
 import com.linecorp.armeria.common.util.SafeCloseable;
+import com.linecorp.armeria.server.DefaultServiceRequestContext;
 
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.EventLoop;
@@ -151,8 +152,11 @@ public abstract class AbstractRequestContext implements RequestContext {
     /**
      * Marks this {@link RequestContext} as having been timed out. Any callbacks created with
      * {@code makeContextAware} that are run after this will be failed with {@link CancellationException}.
+     *
+     * @deprecated Use {@link DefaultServiceRequestContext#setRequestTimedOut()}.
      */
-    public void setTimedOut() {
+    @Deprecated
+    protected void setTimedOut() {
         timedOut = true;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/common/AbstractRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/common/AbstractRequestContext.java
@@ -153,10 +153,10 @@ public abstract class AbstractRequestContext implements RequestContext {
      * Marks this {@link RequestContext} as having been timed out. Any callbacks created with
      * {@code makeContextAware} that are run after this will be failed with {@link CancellationException}.
      *
-     * @deprecated Use {@link DefaultServiceRequestContext#setRequestTimedOut()}.
+     * @deprecated Use {@link DefaultServiceRequestContext#setTimedOut()}.
      */
     @Deprecated
-    protected void setTimedOut() {
+    public void setTimedOut() {
         timedOut = true;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
@@ -89,7 +89,7 @@ public final class RequestMetricSupport {
                                                                           DefaultServiceRequestMetrics::new);
             updateMetrics(log, metrics);
             if (log.responseCause() instanceof RequestTimeoutException) {
-                metrics.requestTimeout().increment();
+                metrics.requestTimeouts().increment();
             }
             return;
         }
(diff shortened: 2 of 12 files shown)


core/src/main/java/com/linecorp/armeria/common/AbstractRequestContext.java
```diff
@@ -151,8 +152,11 @@ public boolean isTimedOut() {
     /**
      * Marks this {@link RequestContext} as having been timed out. Any callbacks created with
      * {@code makeContextAware} that are run after this will be failed with {@link CancellationException}.
+     *
+     * @deprecated Use {@link DefaultServiceRequestContext#setRequestTimedOut()}.
      */
-    public void setTimedOut() {
+    @Deprecated
+    protected void setTimedOut() {
```
trustin: Should still be public because otherwise it will be a breaking change.

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -200,4 +236,43 @@ public Timer totalDuration() {
...
+    private static class DefaultClientRequestMetrics
+            extends AbstractRequestMetrics implements ClientRequestMetrics {
+
+        private final Counter writeTimeout;
+        private final Counter responseTimeout;
+
+        DefaultClientRequestMetrics(MeterRegistry parent, MeterIdPrefix idPrefix) {
+            super(parent, idPrefix);
+            writeTimeout = parent.counter(idPrefix.name("writeTimeout"), idPrefix.tags());
+            responseTimeout = parent.counter(idPrefix.name("responseTimeout"), idPrefix.tags());
```
trustin: - `writeTimeouts` and `responseTimeouts`
- or shoule we name them `timeouts` and add a tag? e.g. `timeouts{cause=WriteTimeoutException}` and `timeouts{cause=ResponseTimeoutException}`

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -200,4 +236,43 @@ public Timer totalDuration() {
...
+    }
+
+    private static class DefaultServiceRequestMetrics
+            extends AbstractRequestMetrics implements ServiceRequestMetrics {
+
+        private final Counter requestTimeout;
+
+        DefaultServiceRequestMetrics(MeterRegistry parent, MeterIdPrefix idPrefix) {
+            super(parent, idPrefix);
+            requestTimeout = parent.counter(idPrefix.name("requestTimeout"), idPrefix.tags());
```
trustin: - `requestTimeouts`
- Or should we name it `timeouts` and add a tag? e.g. `timeouts{cause=ResponseTimeoutException}`

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -89,6 +90,8 @@
     private Runnable requestTimeoutHandler;
     private long maxRequestLength;
 
+    private boolean timedOut;
```
trustin: I think we still need to keep this in the super class because the users of the deprecated method will not want to see a breaking change.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -405,6 +408,20 @@ public void setRequestTimeoutHandler(Runnable requestTimeoutHandler) {
...
+    @Override
+    public boolean isRequestTimedOut() {
+        return timedOut;
+    }
+
+    /**
+     * Marks this {@link ServiceRequestContext} as having been timed out. Any callbacks created with
+     * {@code makeContextAware} that are run after this will be failed with {@link CancellationException}.
+     */
+    public void setRequestTimedOut() {
```
trustin: You don't really need to introduce new methods in this case. If we just overrode `setTimedOut()` and `isTimedOut()` without `@Deprecated` tag, `ServiceRequestContext.isTimedOut()` and `setTimedOut()` will not emit a deprecation warning:

```java
interface RequestContext {
    /**
     * Old Javadoc stays here.
     * @deprecated Use {@link ServiceRequestContext#isRequestTimedOut()}.
     */
    @Deprecated
    boolean isTimedOut();
}

interface ServiceRequestContext extends RequestContext {
    /**
     * Put your new Javadoc here.
     */
    @Override
    boolean isTimedOut();
}

ServiceRequestContext ctx = ...;
boolean timedOut1 = ctx.isTimedOut(); // No warning
boolean timeoutOut2 = ((RequestContext) ctx).isTimedOut(); // Warning
```

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContext.java
minwoox: Got it. Reverted it. Thanks!

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
minwoox: Great idea to name it as `timeouts` and add a tag. Fixed, thank you!

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
minwoox: Removed it. Thanks!

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
minwoox: Changed it as you suggested. Thanks!

trustin:
Thanks, @minwoox !

PR #1574, opened 2019-02-07
Support content preview for request/response in RequestLog

tmdgus0118:
Motivation:
related to #730 
- Currently, there's no way to log content using supported LoggingDecorators unless you use `RequestLogBuilder.deferRequestContent()` or `RequestLogBuilder.deferResponseContent()` which may not be good for a user who does not want to log all of content body.

Modification:
- Added an `ContentPreviewer` interface.
- Added a package-private `BinaryContentPreviewer` class which implements `ContentPreviewer` for reproducing preview a user may want. A user can create this via `ContentPreviewer.ofBinary()`
- Added a package-private `StringContentPreviewer` class which implements `ContentPreviewer` for collecting string with the specified max length and the charset based on `content-type` header (or defaultCharset). A user can create this via `ContentPreviewer.ofText()`
- Added an `ContentPreviewerFactory` interface which creates a new previewer for the request/responses with specific content-type or on the condition specified by a user.
- Added a package-private `CompositeContentPreviewerFactory` class which wraps a list of `ContentPreviewerFactory`s
- Added a package-private `TextualContentPreviewerFactory` class which creates `StringContentPreviewer` when textual content is received.
- Added a package-private `MappedContentPreviewerFactory` class which creates a new previewer through supplier when the content-type header matches any `MediaType` key.

Usage:
By default, previewing is disabled.
A user can enable it when configuring `Server`, `VirtualHost`, or `Client`

```java
ServerBuilder sb = new ServerBuilder();
// Enables previewing the content with the maximum length of 100 for textual contents.
sb.contentPreview(100);

VirtualHostBuilder vhb = new VirtualHostBuilder();
// In this case, the property of virtual host takes precedence over that of server.
vhb.contentPreview(150);

HttpClientBuilder client = new HttpClientBuilder("http://example.com");
client.contentPreview(100);

// A user can use their customized previewer factory.
sb.contentPreviewerFactory((ctx, headers) -> {
    ...
    // the previewer which produces the preview through customized function
    // when the contents have been aggregated more than specific bytes or the stream has been ended.
    // In this case, the preview is produced when the contents have been aggregated more than 100 bytes
    // Or the stream has been ended.
    return ContentPreviewer.ofBinary(100, byteBuf -> {
        return ...
    });
});
```

two or more `ContentPreviewerFactory`s can be combined to 1 `ContentPreviewerFactory`.

```java

ContentPreviewerFactory factory1 = ...;
// if factory1 returns disabled, the previewer is created by factory2.
[...]

PR #1593, opened 2019-02-14
Add total-attempts header/Add totalAttempts in log when RetryingClien…

### LINKED ISSUE #1583: Make RetryingClient set a header that tells the number of attempts so far
.. so that a server can make use of that information to reject excessive retries, etc.

minwoox:
…t is used

Motivation:
If we set a header which indicates the number of attempts so far when using `RetryingClient`,
the server can use it to reject excessive retries, etc.
Also, the log and metrics are now insufficient to contain total attempts info, we should fix it.

Modifications:
- Add total-attempts header only when retry happens(from the second attempt)
- Log `totalAttempts` in response
- Add `actualRequests` counter to metric so that a user can calculate how many retries happened
- Fix a bug where `requestLength` and `responseLength` is not set properly

Result:
- Close #1583
- More information in log and metric

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
@@ -34,6 +34,7 @@ import com.linecorp.armeria.client.SimpleDecoratingClient;
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Response;
 
+import io.netty.util.AsciiString;
 import io.netty.util.AttributeKey;
 import io.netty.util.concurrent.ScheduledFuture;
 
@@ -47,6 +48,12 @@ public abstract class RetryingClient<I extends Request, O extends Response>
         extends SimpleDecoratingClient<I, O> {
 
     private static final Logger logger = LoggerFactory.getLogger(RetryingClient.class);
+
+    /**
+     * An Armeria specific header. This header is used to represent the total attempts.
+     * The server might use this value to reject excessive retries, etc.
+     */
+    public static final AsciiString TOTAL_ATTEMPTS = AsciiString.cached("total-attempts");
 
     private static final AttributeKey<State> STATE =
             AttributeKey.valueOf(RetryingClient.class, "STATE");
@@ -228,6 +235,13 @@ public abstract class RetryingClient<I extends Request, O extends Response>
         }
 
         return nextDelay;
+    }
+
+    /**
+     * Returns the total attempts number of the current request.
+     */
+    protected static int getTotalAttempts(ClientRequestContext ctx) {
+        return ctx.attr(STATE).get().totalAttemptNo;
     }
 
     private static class State {
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
@@ -31,6 +31,7 @@ import com.linecorp.armeria.client.Client;
 import com.linecorp.armeria.client.ClientFactory;
 import com.linecorp.armeria.client.ClientRequestContext;
 import com.linecorp.armeria.client.SimpleDecoratingClient;
+import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Response;
 
@@ -50,10 +51,10 @@ public abstract class RetryingClient<I extends Request, O extends Response>
     private static final Logger logger = LoggerFactory.getLogger(RetryingClient.class);
 
     /**
-     * An Armeria specific header. This header is used to represent the total attempts.
+     * The header which indicates the retry count of the {@link Request}.
      * The server might use this value to reject excessive retries, etc.
      */
-    public static final AsciiString TOTAL_ATTEMPTS = AsciiString.cached("total-attempts");
+    public static final AsciiString ARMERIA_RETRY_COUNT = HttpHeaderNames.of("armeria-retry-count");
 
     private static final AttributeKey<State> STATE =
             AttributeKey.valueOf(RetryingClient.class, "STATE");
@@ -89,8 +90,8 @@ public abstract class RetryingClient<I extends Request, O extends Response>
      * Creates a new instance that decorates the specified {@link Client}.
      */
     private RetryingClient(Client<I, O> delegate, @Nullable RetryStrategy retryStrategy,
-                             @Nullable RetryStrategyWithContent<O> retryStrategyWithContent,
-                             int maxTotalAttempts, long responseTimeoutMillisForEachAttempt) {
+                           @Nullable RetryStrategyWithContent<O> retryStrategyWithContent,
+                           int maxTotalAttempts, long responseTimeoutMillisForEachAttempt) {
         super(delegate);
         this.retryStrategy = retryStrategy;
         this.retryStrategyWithContent = retryStrategyWithContent;
@@ -238,10 +239,17 @@ public abstract class RetryingClient<I extends Request, O extends Response>
     }
 
     /**
-     * Returns the total attempts number of the current request.
+     * Returns the total number of attempts of the current request represented by the specified
+     * {@link ClientRequestContext}.
      */
     protected static int getTotalAttempts(ClientRequestContext ctx) {
-        return ctx.attr(STATE).get().totalAttemptNo;
+        final State state = ctx.attr(STATE).get();
+        if (state == null) {
+            throw new IllegalStateException("state is null. You should call " +
+                                            "RetryingClient.execute(ClientRequestContext ctx, I req) " +
+                                            "before this method is called.");
+        }
+        return state.totalAttemptNo;
     }
 
     private static class State {
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
@@ -51,7 +51,7 @@ public abstract class RetryingClient<I extends Request, O extends Response>
     private static final Logger logger = LoggerFactory.getLogger(RetryingClient.class);
 
     /**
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -48,6 +49,12 @@
 
     private static final Logger logger = LoggerFactory.getLogger(RetryingClient.class);
 
+    /**
+     * An Armeria specific header. This header is used to represent the total attempts.
+     * The server might use this value to reject excessive retries, etc.
+     */
+    public static final AsciiString TOTAL_ATTEMPTS = AsciiString.cached("total-attempts");
```
minwoox: Not sure this suits for this context. We should name it like `retry-count` to clearly indicate to the server?

trustin: `retry-count` sounds better to me. Should we use `armeria-retry-count` or just `retry-count`?

trustin: - `HttpHeaderNames.of(...)`
- First sentence is too vague. Could you reorganize the Javadoc?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -230,6 +237,13 @@ protected final long getNextDelay(ClientRequestContext ctx, Backoff backoff, lon
         return nextDelay;
     }
 
+    /**
+     * Returns the total attempts number of the current request.
```
trustin: `Returns the total number of attempts of the request represented by the specified {@link ClientRequestContext}.` ?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -230,6 +237,13 @@ protected final long getNextDelay(ClientRequestContext ctx, Backoff backoff, lon
...
+     */
+    protected static int getTotalAttempts(ClientRequestContext ctx) {
+        return ctx.attr(STATE).get().totalAttemptNo;
```
trustin: `get()` could return `null`?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -166,7 +166,10 @@ private void propagateRequestSideLog(RequestLog child) {
         child.addListener(log -> requestHeaders(log.requestHeaders()), REQUEST_HEADERS);
         child.addListener(log -> requestContent(log.requestContent(), log.rawRequestContent()),
                           REQUEST_CONTENT);
-        child.addListener(log -> endRequest0(log.requestCause(), log.requestEndTimeNanos()), REQUEST_END);
+        child.addListener(log -> {
+            increaseRequestLength(log.requestLength());
```
trustin: @tmdgus0118 You may want to rebase and update this part at #1574.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -197,6 +200,7 @@ private void propagateResponseSideLog(RequestLog lastChild) {
         }
 
         if (lastChild.isAvailable(RESPONSE_END)) {
+            increaseResponseLength(lastChild.responseLength());
```
trustin: @tmdgus0118 and here.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -207,8 +211,10 @@ private void propagateResponseSideLog(RequestLog lastChild) {
         lastChild.addListener(log -> responseHeaders(log.responseHeaders()), RESPONSE_HEADERS);
         lastChild.addListener(log -> responseContent(
                 log.responseContent(), log.rawResponseContent()), RESPONSE_CONTENT);
-        lastChild.addListener(log -> endResponse0(
-                log.responseCause(), log.responseEndTimeNanos()), RESPONSE_END);
+        lastChild.addListener(log -> {
+            increaseResponseLength(lastChild.responseLength());
```
trustin: @tmdgus0118 and here.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1004,6 +1010,12 @@ public String toStringResponseOnly(Function<HttpHeaders, HttpHeaders> headersSan
         }
         buf.append('}');
 
+        if (children != null && children.size() > 1) {
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -892,16 +898,16 @@ public String toString() {
            .append(res)
            .append('}');     // 1 char
         if (children != null && children.size() > 1) {
```
trustin: - You could define a variable upfront and reuse it everywhere you call `children.size()`:

  ```java
  final int numChildren = children != null ? children.size() : 0;
  if (numChildren != 0) {
      ...
      buf.append(numChildren);
      ...
      for (int i = 0; i < numChildren; i++) { ... }
  }
  ```

site/src/sphinx/client-retry.rst
```diff
@@ -331,12 +331,12 @@ do the reverse:
             .decorator(LoggingClient.newDecorator())
             .build();
 
-This will produce only single request and response log pair regardless how many attempts are made:
+This will produce only single request and response log pair and the total attempts count:
```
trustin: How about:

```
... and the total number of attempts only, regardless how many attempts are made.
```

..?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
minwoox: `armeria-retry-count` sounds better to me. let me use it with the value which is `totalAttempts - 1`. Thanks!

minwoox:
> Could you also update `client-retry.rst` to mention the HTTP header `RetryingClient` adds?

Yes, I will. I was actually going to add it once the header name is fixed. 😆

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
tmdgus0118: How about using `requestLength(log.requestLength())` instead for readability?

tmdgus0118: How about using `requestLength(log.requestLength())` instead for readability? Could it be different?

minwoox: No. Just forgot that we had it. Thanks :-)

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -50,10 +51,10 @@
     private static final Logger logger = LoggerFactory.getLogger(RetryingClient.class);
 
     /**
-     * An Armeria specific header. This header is used to represent the total attempts.
+     * The header which indicates the retry count of the {@link Request}.
```
trustin: the request -> a request?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -238,10 +239,17 @@ protected final long getNextDelay(ClientRequestContext ctx, Backoff backoff, lon
...
     /**
-     * Returns the total attempts number of the current request.
+     * Returns the total number of attempts of the current request represented by the specified
+     * {@link ClientRequestContext}.
      */
     protected static int getTotalAttempts(ClientRequestContext ctx) {
-        return ctx.attr(STATE).get().totalAttemptNo;
+        final State state = ctx.attr(STATE).get();
+        if (state == null) {
+            throw new IllegalStateException("state is null. You should call " +
```
trustin: Probably better returning 0?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
```diff
@@ -109,7 +109,7 @@ private void doExecute0(ClientRequestContext ctx, RpcRequest req,
 
         final int totalAttempts = getTotalAttempts(ctx);
         if (totalAttempts > 1) {
-            derivedCtx.addAdditionalRequestHeader(TOTAL_ATTEMPTS, Integer.toString(totalAttempts));
+            derivedCtx.addAdditionalRequestHeader(ARMERIA_RETRY_COUNT, Integer.toString(totalAttempts - 1));
```
trustin: Could use `setAdditionalRequestHeader()` instead of `addAdditionalRequestHeader()`?

site/src/sphinx/client-retry.rst
```diff
@@ -312,11 +312,17 @@ This will produce following logs when there are three attempts:
...
+    LoggingClient - Request: {startTime=..., ..., headers=[..., armeria-retry-count=1, ...]
     LoggingClient - Response: {startTime=..., length=..., duration=..., headers=[:status=500, ...]
-    LoggingClient - Request: {startTime=..., length=..., duration=..., scheme=..., host=..., headers=[...]
+    LoggingClient - Request: {startTime=..., ..., headers=[..., armeria-retry-count=2, ...]
     LoggingClient - Response: {startTime=..., length=..., duration=..., headers=[:status=200, ...]
 
+.. note::
+
+    Did you notice that the ``armeria-retry-count`` header is inserted from the second request?
+    :api:`RetryingClient` inserts the header to indicate the retry count of the request.
```
trustin: the header -> it ?

trustin:
One last comment - `actualRequests` meter should not be created when a user does not use `RetryingClient`. Perhaps we should register it lazily?

trustin:
Thanks, @minwoox !

PR #1594, opened 2019-02-14
Add `responseCauseSanitizer` to `LoggingDecoratorBuilder`

trustin:
Motivation:

A user wants to sanitize the stack trace of `RequestLog.responseCause()`
or avoid logging the stack trace completely.

Modifications:

- Added LoggingDecoratorBuilder.responseCauseSanitizer()` and applied
  the specified `Function` to the response cause.
- Widened the type parameters of sanitizer functions.
- Miscellaneous:
  - Fixed a bug where a response content sanitizer is applied on request
    content on a certain case.
  - Added `contentSanitizer()` and `headersSanitizer()` shortcut methods.
  - Javadoc clean-up

Result:

- A user can sanitize/drop the stack trace of `RequestLog.responseCause()`.
- Request content is always sanitized with an intended sanitizer.
- More shortcut methods in `LoggingClientBuilder` and `LoggingServiceBuilder`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/logging/LoggingClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/logging/LoggingClient.java
@@ -46,9 +46,6 @@ public final class LoggingClient<I extends Request, O extends Response> extends 
 
     private static final Logger logger = LoggerFactory.getLogger(LoggingClient.class);
 
-    private static final String REQUEST_FORMAT = "{} Request: {}";
-    private static final String RESPONSE_FORMAT = "{} Response: {}";
-
     /**
      * Returns a new {@link Client} decorator that logs {@link Request}s and {@link Response}s at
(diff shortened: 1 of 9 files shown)


trustin:
@anuraaga I found we do not log `RequestLog.requestCause()`, so I did not add `requestCauseSanitizer()`. Do you think we need to add `requestCauseSanitizer()` and start to log a request cause? I'm curious why we did not so far. Any clues?

anuraaga:
Hmm - I don't know if I've ever thought of the existence of request cause. Any request exception would probably be logged in the response itself so it seems likely to double-log if the request cause was logged.

trustin:
Yeah. Tha's probably why. Let me keep it this way then. Thanks for a reply.

trustin:
Thanks for reviewing :-)

PR #1600, opened 2019-02-17
Execute blocking gRPC callbacks in order

### LINKED ISSUE #1524: Why armeria run all grpc service logic directly in the event loop?

anuraaga:
Fixes #1524 hopefully doesn't break other things though ><

I was expecting this to just be running callbacks in order, but it turned out also needing to rework many aspects of deframing to allow the deframing and business logic to happen on separate threads. It seems like upstream also has added similar logic after our fork of the class, though the context seems to be moving deframing off the transport thread which confuses me since I would expect the ordering issue to happen when it is _on_ the transport thread. Oh well, didn't dig deeper on that.

https://github.com/grpc/grpc-java/pull/3145

Now, when the HTTP reader is finished, the business logic listener is not notified right away, but only if all messages have first been delivered. Previously, the ordering worked out that this was always the case but when message processing happens on a different thread, it isn't anymore.

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
+++ b/grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
@@ -177,13 +177,13 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
 
     private boolean compressedFlag;
     private boolean endOfStream;
+    private boolean closeWhenComplete;
 
     @Nullable
     private Queue<ByteBuf> unprocessed;
     private int unprocessedBytes;
 
     private long pendingDeliveries;
-    private boolean deliveryStalled = true;
     private boolean inDelivery;
     private boolean startedDeframing;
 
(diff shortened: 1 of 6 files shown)


grpc/src/main/java/com/linecorp/armeria/internal/grpc/ArmeriaMessageDeframer.java
```diff
@@ -257,18 +257,33 @@ public void deframe(HttpData data, boolean endOfStream) {
         deliver();
     }
 
+    /** Close when any messages currently queued have been requested and delivered. */
```
hyangtack: nit: Close -> Closes?

anuraaga: Thanks updated. I was a bit unsure since I just copied this line from upstream, but it makes sense for us to make tweaks since it's quite diverged at this point.

minwoox:
Thanks @anuraaga

PR #1607, opened 2019-02-19
Update server-access-log.rst

perlun:
The `ServerBuilder.accessLogWriter()` method now expects two parameters, the latter being `shutdownOnStop`. I assume `true` is a reasonable default for this, but feel free to change to `false` if you so prefer. The key element is to make the examples be correct.

Thanks for a nice project! :+1:

### DIFF
--- a/site/src/sphinx/server-access-log.rst
+++ b/site/src/sphinx/server-access-log.rst
@@ -97,9 +97,9 @@ using :api:`ServerBuilder`. You may use one of the pre-defined log formats.
 
     ServerBuilder sb = new ServerBuilder();
     // Use NCSA common log format.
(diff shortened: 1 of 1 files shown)


trustin:
`true` is a reasonable default in most cases. Thanks for your first pull request! Would love to pull in once you sign the CLA.

perlun:
> Would love to pull in once you sign the CLA.

There we go, CLA signed. :+1:

trustin:
Much appreciated. Thanks for your first contribution!

perlun:
> Much appreciated. Thanks for your first contribution!

Thanks, hopefully not the last! :smile: :+1:

PR #1610, opened 2019-02-20
Fix a bug where SSE is incorrectly encoded

hyangtack:
Motivation:
- When encoding SSE, there was a missing `end-of-line` character for each event.
- `ResponseConversionUtil#StreamingSubscriber` is possible to make lots of `request`s to its publisher.
- We didn't have a working example for Server-Sent Events.

Modifications:
- Add a missing `end-of-line` character when emitting an SSE.
- Register on-demand callback to `HttpResponseWriter` which requests a demand to its publisher only after `onNext` is called.
- Add a working example which shows how to send SSE with Armeria and how to receive the events with simple Javascript.

Result:
- Less bugs
- Performance
- Documentation


examples/annotated-http-service/build.gradle
```diff
@@ -3,7 +3,6 @@ dependencies {
     compile 'io.projectreactor:reactor-core'
```
minwoox: Could remove this as well?

site/src/sphinx/server-sse.rst
```diff
@@ -49,3 +53,37 @@ use the annotation.
     public Publisher<ServerSentEvent> sse() {
         return Flux.just(ServerSentEvent.ofData("foo"), ServerSentEvent.ofData("bar"));
     }
+
+Adjusting the request timeout
+-----------------------------
+
+The events may being sent for a longer time depending on the application. Even it can continue infinitely,
```
trustin: - The events -> An event stream
- being -> be
- a longer time -> a longer period than the configured timeout
- Even it -> It even

examples/annotated-http-service/build.gradle
```diff
@@ -3,7 +3,6 @@ dependencies {
...
     runtime 'org.slf4j:slf4j-simple'
```
minwoox: and this?

site/src/sphinx/server-sse.rst
```diff
@@ -49,3 +53,37 @@ use the annotation.
...
+for example streaming stock quotes. But the server allows receiving a request and sending the corresponding
+response completely only within the amount of time, which is configured as the request timeout property.
+Even if you do not configure the property, it is configured to ``10`` seconds by default.
```
trustin: 'But the server .... seconds by default.' could be simplified like this:

```
Such a long running stream may be terminated prematurely because Armeria has
the default request timeout of ``10`` seconds, i.e. your stream will be broken
after 10 seconds at most.
```

site/src/sphinx/server-sse.rst
```diff
@@ -49,3 +53,37 @@ use the annotation.
...
+So you must adjust the timeout if your event stream is possible to exceed it because the response will be
```
trustin: - So -> Therefore,
- if your event stream lasts longer than the configured timeout.
- because the response will be ... -> Could be removed, because we explained it above already?

site/src/sphinx/server-sse.rst
```diff
@@ -49,3 +53,37 @@ use the annotation.
...
+terminated if it lasts longer than the configured timeout. The following example shows how to adjust
+the timeout of a single request. As you might know, it is not only for `Server-Sent Events`_, so you can
+use this way for any request which you want to adjust its timeout.
```
trustin: - this way -> this method
- any requests which you want to adjust timeout for.

site/src/sphinx/server-sse.rst
```diff
@@ -49,3 +53,37 @@ use the annotation.
...
+
+    import java.time.Duration;
+    import com.linecorp.armeria.common.sse.ServerSentEvent;
+    import com.linecorp.armeria.server.Server;
+    import com.linecorp.armeria.server.ServerBuilder;
+    import com.linecorp.armeria.server.streaming.ServerSentEvents;
+    import reactor.core.publisher.Flux;
+
+    Server server = new ServerBuilder()
+            // This service infinitely sends numbers as the data of events for every second.
```
trustin: for every second -> every second

site/src/sphinx/server-sse.rst
```diff
@@ -49,3 +53,37 @@ use the annotation.
...
+            .service("/long-sse", (ctx, req) -> {
+                // Note that you MUST adjust the request timeout if you want to send events for a
+                // longer time than the configured request timeout. The timeout can be unset by
```
trustin: - longer time -> longer period
- unset -> disabled

examples/server-sent-events/src/main/java/example/armeria/server/sse/Main.java
```diff
@@ -0,0 +1,83 @@
...
+
+    static Server newServer(int httpPort, int httpsPort, Duration sendingInterval, long eventCount,
+                            Supplier<String> randomStringSupplier) throws Exception {
+        return new ServerBuilder()
+                .http(httpPort)
+                .https(httpsPort)
+                .tlsSelfSigned()
+                .service("/long", (ctx, req) -> {
+                    // Note that you MUST adjust the request timeout if you want to send events for a
+                    // longer time than the configured request timeout. The timeout can be unset by
```
minwoox: can be disabled?

site/src/sphinx/server-sse.rst
```diff
@@ -49,3 +53,37 @@ use the annotation.
...
+                // setting 0 like the below, but it is NOT RECOMMENDED in the real world application.
```
trustin: ... application, because it can leave a lot of unfinished requests.

hyangtack:
Thanks, @trustin and @minwoox. Resolved all comments. :-)

PR #1611, opened 2019-02-21
Add docs about a composite annotation

### LINKED ISSUE #1608: Write docs about using annotation alias
User guide of #1560

hyangtack:
Modifications:
- Add `Using a composite annotation` section to `server-annotated-service.rst`.

Result:
- Documentation
- Closes #1608

### DIFF
--- a/site/src/sphinx/server-annotated-service.rst
+++ b/site/src/sphinx/server-annotated-service.rst
@@ -1294,8 +1294,8 @@ because they always have an empty content.
 Using a composite annotation
 ----------------------------
 
-In some cases, you may want to make a composite annotation which includes several types of annotations.
-Let's assume that there is a service class like the below:
+To avoid specifying a common set of annotations repetitively, you may want to create a composite annotation
+which is annotated by other annotations. For example, let's assume that there is a service class like the below:
 
 .. code-block:: java
 
@@ -1316,9 +1316,8 @@ Let's assume that there is a service class like the below:
         public HttpResponse update() { ... }
     }
 
-In the example, you need to add the same 4 annotations to the two different methods. It is obviously
-uncomfortable and noisy. So Armeria provides a way to make a composite annotation. You can create your own
-composite annotation as follows.
+In the above example, you had to add the same 4 annotations to the two different methods. It is obviously
+too verbose and duplicate, so we could simplify them by creating a composite annotation like the following:
 
 .. code-block:: java
 
@@ -1329,8 +1328,9 @@ composite annotation as follows.
     @MyAuthenticationDecorator
     public @interface MyCreateOrUpdateApiSpec {}
 
-Now let's rewrite the service class with the composite annotation. It is definitely more clear than before.
-Moreover it makes you can manage your annotations in the better way.
+Now, let's rewrite the service class with the composite annotation. It is definitely less verbose than before.
+Moreover, you don't need to update both ``create()`` and ``update()`` but only ``MyCreateOrUpdateApiSpec``
+when you add more common annotations to them.
 
 .. code-block:: java
 
--- a/site/src/sphinx/server-annotated-service.rst
+++ b/site/src/sphinx/server-annotated-service.rst
@@ -1291,3 +1291,56 @@ because they always have an empty content.
 | 304          | Not modified   |
 +--------------+----------------+
 
+Using a composite annotation
+----------------------------
+
+In some cases, you may want to make a composite annotation which includes several types of annotations.
+Let's assume that there is a service class like the below:
+
+.. code-block:: java
+
+    public class MyAnnotatedService {
+
+        @Post("/create")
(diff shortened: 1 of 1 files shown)


hyangtack:
@line/cdtf I picked `Using a composite annotation` as the title of new section. But I'm not sure whether `a composite annotation` is suitable naming or not. Please let me know if you have any idea about the name. :-)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1291,3 +1291,56 @@ because they always have an empty content.
...
+
+        @Post("/update")
+        @ConsumesJson
+        @ProducesJson
+        @LoggingDecorator
+        @MyAuthenticationDecorator
+        public HttpResponse update() { ... }
+    }
+
+In the example, you need to add the same 4 annotations to the two different methods. It is obviously
```
trustin: - the above example
- you need to -> you had to

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1291,3 +1291,56 @@ because they always have an empty content.
...
+uncomfortable and noisy. So Armeria provides a way to make a composite annotation. You can create your own
```
trustin: too verbose and duplicate, so we could simplify them by creating a composite annotation like the following:

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1291,3 +1291,56 @@ because they always have an empty content.
 | 304          | Not modified   |
 +--------------+----------------+
 
+Using a composite annotation
+----------------------------
+
+In some cases, you may want to make a composite annotation which includes several types of annotations.
```
trustin: To avoid specifying a common set of annotations repetitively, you may want to create a composite annotation which is annotated by other annotations.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1291,3 +1291,56 @@ because they always have an empty content.
...
+.. code-block:: java
+
+    @Retention(RetentionPolicy.RUNTIME)
+    @ConsumesJson
+    @ProducesJson
+    @LoggingDecorator
+    @MyAuthenticationDecorator
+    public @interface MyCreateOrUpdateApiSpec {}
+
+Now let's rewrite the service class with the composite annotation. It is definitely more clear than before.
```
trustin: - Comma between Now and let's.
- more clear -> less verbose?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1291,3 +1291,56 @@ because they always have an empty content.
 | 304          | Not modified   |
 +--------------+----------------+
 
+Using a composite annotation
+----------------------------
+
+In some cases, you may want to make a composite annotation which includes several types of annotations.
+Let's assume that there is a service class like the below:
```
trustin: For example, let's assume ...

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1291,3 +1291,56 @@ because they always have an empty content.
...
+
+    @Retention(RetentionPolicy.RUNTIME)
+    @ConsumesJson
+    @ProducesJson
+    @LoggingDecorator
+    @MyAuthenticationDecorator
+    public @interface MyCreateOrUpdateApiSpec {}
+
+Now let's rewrite the service class with the composite annotation. It is definitely more clear than before.
+Moreover it makes you can manage your annotations in the better way.
```
trustin: Moreover, you don't need to update both `create()` and `update()` but only `MyCreateOrUpdateApiSpec` when you add more common annotations to them.

('in the better way' was too vague.)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -1291,3 +1291,56 @@ because they always have an empty content.
...
+    @Retention(RetentionPolicy.RUNTIME)
+    @ConsumesJson
+    @ProducesJson
+    @LoggingDecorator
+    @MyAuthenticationDecorator
+    public @interface MyCreateOrUpdateApiSpec {}
+
+Now, let's rewrite the service class with the composite annotation. It is definitely less verbose than before.
+Moreover, you don't need to update both ``create()`` and ``update()`` but only ``MyCreateOrUpdateApiSpec``
+when you add more common annotations to them.
```
minwoox: How about removing these two lines? I think it's a little verbose and does not contain useful information(people will know about this without this sentence.).
It's just my opinion and if you think it needs you can leave it. :)

hyangtack: Thanks to the comment! I agree with you to a certain degree. It seems a little verbose but I think it'd be better to leave it as it is because it can be used to emphasized the benefit when using a composite annotation. :-)

PR #1612, opened 2019-02-21
Less noisy TLS handshake error logging

trustin:
Motivation:

Server might leave a log message like the following when a client
disconnects prematurely during TLS handshake:

    TLS handshake failed:
    java.nio.channels.ClosedChannelException: null
        at io.netty.handler.ssl.SslHandler.channelInactive(...)

.. which pollutes the server log file.

Modifications:

- Guarded the TLS handshake failure logging code with
  `!Exceptions.isExpected()`.

Result:

- Less noise


trustin:
Thanks :smile:

Issue #723, opened 2017-08-14
io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED   


sunning9001:
env:  JDK 1.8   os:  windows    armeria  version:  0.52.0   data format: google protocal buffer

on the rpc client ,I  set idleTimeout =50 second , and  rpc  server   ,I set  idleTimeout =50 secord ,requestTimeout =50 secord .  but  10 second pass, the  **io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED**   occur.

I also know  that  only  the  default  IDLE_TIMEOUT_MILLIS is  ten second. 

why   the parameters  I had set  do not work ??

the rpc client code:

	    	ClientFactory factory=  new ClientFactoryBuilder().idleTimeout( Duration.ofSeconds(50)).build();
		
				ClientBuilder builder = new ClientBuilder("gproto+http://127.0.0.1:8080/")
						.decorator(HttpRequest.class, HttpResponse.class, LoggingClient.newDecorator()).factory(factory);
				
				HelloServiceFutureStub helloService = builder.build(HelloServiceFutureStub.class);																			// HelloServiceFutureStub.class
					
				HelloRequest request = HelloRequest.newBuilder().setName("Armerian World").build();
				ListenableFuture<HelloReply> reply = helloService.hello(request);
				reply.get();

the rpc server code:

		ServerBuilder sb = new ServerBuilder().idleTimeout( Duration.ofSeconds(50)).defaultRequestTimeout( Duration.ofSeconds(50));
		sb.port(8080, HTTP);
		sb.serviceUnder("/", new GrpcServiceBuilder().addService(new MyHelloService()).build());

		Server server = sb.build();
		server.start();

the  exception  code:
java.util.concurrent.ExecutionException: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:503)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:482)
	at GrpcClient.main(GrpcClient.java:40)
时间11
Caused by: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED
	at io.grpc.Status.asRuntimeException(Status.java:543)
	at io.grpc.stub.ClientCalls$UnaryStreamToFuture.onClose(ClientCalls.java:442)
[...]

Issue #1134, opened 2018-04-09
GrpcService with default serialization formats does not show a debug form in DocService


trustin:
Unlike `THttpService`, a user has to enable more serialization formats to make `DocService` show the debug form for a `GrpcService`. Perhaps we should:

- make it less surprising,
- update our documentation, or
- add some instructions to the DocService page without the debug form

What do you think, @anuraaga ?

anuraaga:
Yeah with some more documentation and with text in `DocService` or at least a `logger.info`, it would be much clearer. I think the default for it to be off is safer, at least for servers that might be exposed to external traffic since exposing unframed JSON can be a huge DDoS opportunity - it's the main reason I didn't default those on.

I have been thinking for some time, and a little related to this, perhaps we could have the concept of modes in armeria. Something along the lines of
- Internal vs external
- RPC vs REST
- Unary vs streaming
- ... etc (though not too many or it's just spaghetti soup)

which group together best-practice defaults / flags for each use case. In this case, I would default on in `Internal` and off in `external`. Defaults for things like the path parsing cache, timeouts, logging, and probably many things could also be tuned for these use cases.

PR #1613, opened 2019-02-21
[docs] Add info about client timeouts

perlun:
Slightly related to #723, this PR describes how the default client timeouts can be overridden.

(This wasn't obvious to me at first, so adding this in the hope that it helps others. Note that the `ClientBuilder` class can also be used to tweak this; it's not 100% clear to me when to use that one and when to go with the `Clients.newClient` method.)

### DIFF
--- a/site/src/sphinx/client-timeouts.rst
+++ b/site/src/sphinx/client-timeouts.rst
@@ -1,0 +1,47 @@
+.. _client-timeouts:
+
+Overriding client timeouts
+==========================
+Sometimes, the default timeouts used by the Armeria-generated client does not suit a particular scenario well. In these cases, you might want to override the timeout settings.
+
+Using ``ClientOptionsBuilder``
+------------------------------
+
+.. code-block:: java
+
+    import java.time.Duration;
+
+    import com.linecorp.armeria.common.util.SafeCloseable
+    import com.linecorp.armeria.client.ClientOptionsBuilder;
+    import com.linecorp.armeria.client.Clients;
+
+    // Defaults are defined in com.linecorp.armeria.common.Flags and
+    // com.linecorp.armeria.client.ClientOptions
+    int responseTimeout = 15;
+    int writeTimeout = 1;
+
+    HelloService.Iface client = Clients.newClient(
+            "tbinary+http://example.com/hello",
+            HelloService.Iface.class,
+            new ClientOptionsBuilder()
+                    .defaultResponseTimeout(Duration.ofSeconds(responseTimeout))
+                    .defaultWriteTimeout(Duration.ofSeconds(writeTimeout))
+                    .build()
+    );
+
+Using ``ClientBuilder``
+-----------------------
+
+.. code-block:: java
+
+    import java.time.Duration;
+
+    import com.linecorp.armeria.client.ClientBuilder;
+
+    int responseTimeout = 15;
+    int writeTimeout = 1;
+
+    HelloService.Iface client = new ClientBuilder("tbinary+http://example.com/hello")
+            .defaultResponseTimeout(Duration.ofSeconds(responseTimeout))
+            .defaultWriteTimeout(Duration.ofSeconds(writeTimeout))
+            .build(HelloService.Iface.class);
--- a/site/src/sphinx/client-custom-http-headers.rst
+++ b/site/src/sphinx/client-custom-http-headers.rst
@@ -2,7 +2,7 @@
 
 Sending custom HTTP headers
 ===========================
-When send an RPC request, it is sometimes required to send HTTP headers with it, such as authentication token.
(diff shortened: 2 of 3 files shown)


anuraaga:
Thanks, this looks very useful! Just wondering did you run into this with a streaming method?

Wondering if #1134 is related, I've always wished it took less knobs to accomplish common use cases.

site/src/sphinx/client-timeouts.rst
```diff
@@ -0,0 +1,26 @@
+.. _client-timeouts:
+
+Overriding client timeouts
+==========================
+Sometimes, the default timeouts used by the Armeria-generated client does not suit a particular scenario well. In these cases, you might want to override the timeout settings.
+
+Using ``ClientOptionsBuilder``
+-----------------------------------
```
trustin: Mismatching number of hyphens

site/src/sphinx/client-timeouts.rst
```diff
@@ -0,0 +1,26 @@
...
+
+    import com.linecorp.armeria.common.util.SafeCloseable
+    import com.linecorp.armeria.client.ClientOptionsBuilder;
+    import com.linecorp.armeria.client.Clients;
+
+    // Defaults are defined in com.linecorp.armeria.common.Flags and com.linecorp.armeria.client.ClientOptions
+    int responseTimeout = 15;
+    int writeTimeout = 1;
+
+    HelloService.Iface client = Clients.newClient("gproto+http://example.com/hello",
```
trustin: 1) `tbinary+http://...`? :-)

2) How about breaking the line after `(` so that the example does not overflow? You can preview your documentation by running `./gradlew site` at `site/build/site/index.html`.

3) You can also use `ClientBuilder` which may simplify the construction of this client:

```java
HelloService.Iface client = new ClientBuilder("tbinary+http://example.com/hello")
    .defaultResponseTimeout(Duration.ofSeconds(responseTimeout))
    .defaultWriteTimeout(Duration.ofSeconds(writeTimeout))
    .build(HelloService.Iface.class);
```

perlun: Thanks for the feedback! :+1: Going on vacation shortly (Winter holidays here in the Nordics coming up), will try to pick this up at the end of next week.

perlun: @trustin 

> 1. `tbinary+http://...`? :-)

What's the difference, Thrift vs gRPC? Anyway, I did the change (we are using gRPC ourselves).

> 2. How about breaking the line after `(` so that the example does not overflow? You can preview your documentation by running `./gradlew site` at `site/build/site/index.html`.

Good point, and big thanks for the preview suggestion! Made things a lot easier.

> 3. You can also use `ClientBuilder` which may simplify the construction of this client:

I added a `ClientBuilder` example now also, thanks for that. To be honest, I don't fully comprehend when `ClientBuilder` would be preferable to the `Clients.newClient` factory method. I guess they are pretty much interchangeable?

Anyway, changes done - will push in a second.

perlun:
@anuraaga 

> Thanks, this looks very useful! Just wondering did you run into this with a streaming method?

Nope, just regular gRPC calls. I ran into this when putting an IntelliJ breakpoint in the server-side code; it was annoying that the client would time out in this case, which is why I started digging into this in the first place.

trustin:
Thanks, @perlun !

site/src/sphinx/client-timeouts.rst
```diff
@@ -0,0 +1,47 @@
+.. _client-timeouts:
+
+Overriding client timeouts
+==========================
+Sometimes, the default timeouts used by the Armeria-generated client does not suit a particular scenario well. In these cases, you might want to override the timeout settings.
```
anuraaga: Should we also document the system properties that can control the defaults? Especially for the case where debugging in IntelliJ or something, using system properties in the debug configuration can be even simpler than changing the code.

anuraaga: For reference, the properties are documented around here

https://line.github.io/armeria/apidocs/com/linecorp/armeria/common/Flags.html#defaultRequestTimeoutMillis()

site/src/sphinx/client-timeouts.rst
trustin: > What's the difference, Thrift vs gRPC? Anyway, I did the change (we are using gRPC ourselves).

Yes. You used Thrift interface (`*.AsyncIface`), so I thought it is more realistic to use Thrift URI.

> I guess they are pretty much interchangeable?

Yes. I thought most users will prefer the version that used `ClientBuilder`.

site/src/sphinx/client-timeouts.rst
trustin: Let me send a follow-up pull request after merging this one.

trustin:
Thanks, @perlun !

PR #1615, opened 2019-02-22
Add CacheControl, CacheControlBuilder and their subtypes

trustin:
Motivation:

It will be useful if a user can generate `cache-control` header value
programatically.

Modifications:

- Added `CacheControl`, `ClientCacheControl` and `ServerCacheControl`.
- Added `CacheControlBuilder`, `ClientCacheControlBuilder` and
  `ServerCacheControlBuilder`.
- Added `cacheControl()` to `HttpFileServiceBuilder` and `AbstractHttpFileBuilder`.
- Made `HttpHeaders.setObject()` supports `CacheControl`.
- Miscellaneous:
  - Added a way to specify additional headers to `HttpFileServiceBuilder`.
  - Added `AbstractHttpFileBuilder.contentType()`.
  - `Content-Type` -> `content-type` for consistency

Result:

- It is more convenient to set `cache-control` or `content-type` headers.
- A user can now specify additional headers to the responses created by
  `HttpFileService`.
- `HttpVfs.get()` has a breaking change.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientCacheControl.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientCacheControl.java
@@ -19,9 +19,13 @@ import static com.linecorp.armeria.internal.ArmeriaHttpUtil.parseCacheControl;
 import static com.linecorp.armeria.internal.ArmeriaHttpUtil.parseCacheControlSeconds;
 import static java.util.Objects.requireNonNull;
 
+import java.util.Map;
+import java.util.function.BiConsumer;
+
 import javax.annotation.Nullable;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 
 import com.linecorp.armeria.common.CacheControl;
 
@@ -53,6 +57,48 @@ public final class ClientCacheControl extends CacheControl {
             .maxStaleSeconds(Integer.MAX_VALUE)
             .build();
 
+    private static final Map<String, BiConsumer<ClientCacheControlBuilder, String>> DIRECTIVES =
+            ImmutableMap.<String, BiConsumer<ClientCacheControlBuilder, String>>builder()
+                    .put("no-cache", (b, v) -> b.noCache())
+                    .put("no-store", (b, v) -> b.noStore())
+                    .put("no-transform", (b, v) -> b.noTransform())
+                    .put("max-age", (b, v) -> {
+                        final long maxAgeSeconds = parseCacheControlSeconds(v);
+                        if (maxAgeSeconds >= 0) {
+                            b.maxAgeSeconds(maxAgeSeconds);
+                        }
+                    })
+                    .put("only-if-cached", (b, v) -> b.onlyIfCached())
+                    .put("max-stale", (b, v) -> {
+                        if (v == null) {
+                            b.maxStale();
+                        } else {
+                            final long maxStaleSeconds = parseCacheControlSeconds(v);
+                            if (maxStaleSeconds >= 0) {
+                                b.maxStaleSeconds(maxStaleSeconds);
+                            }
+                        }
+                    })
+                    .put("min-fresh", (b, v) -> {
+                        final long minFreshSeconds = parseCacheControlSeconds(v);
+                        if (minFreshSeconds >= 0) {
+                            b.minFreshSeconds(minFreshSeconds);
+                        }
+                    })
+                    .put("stale-while-revalidate", (b, v) -> {
+                        final long staleWhileRevalidateSeconds = parseCacheControlSeconds(v);
+                        if (staleWhileRevalidateSeconds >= 0) {
+                            b.staleWhileRevalidateSeconds(staleWhileRevalidateSeconds);
+                        }
+                    })
+                    .put("stale-if-error", (b, v) -> {
+                        final long staleIfErrorSeconds = parseCacheControlSeconds(v);
+                        if (staleIfErrorSeconds >= 0) {
+                            b.staleIfErrorSeconds(staleIfErrorSeconds);
+                        }
+                    })
+                    .build();
+
     /**
      * Parses the specified {@code "cache-control"} header values into a {@link ClientCacheControl}.
      * Note that any unknown directives will be ignored.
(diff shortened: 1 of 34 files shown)


hyangtack:
> Made `HttpHeaders.setObject()` supports `CacheControl`.

I couldn't find this change. Would you let me know where it is if I missed it? :-)

core/src/main/java/com/linecorp/armeria/server/file/AbstractHttpFileBuilder.java
```diff
@@ -199,7 +204,46 @@ public final B setHeader(CharSequence name, Object value) {
...
+     * Sets the {@code "content-type"} header. This method is a shortcut of:
+     * <pre>{@code
+     * builder.autoDetectedContentType(false);
+     * builder.setHeader(HttpHeaderNames.CONTENT_TYPE, contentType);
+     * }</pre>
+     */
+    public final B contentType(MediaType contentType) {
+        requireNonNull(contentType, "contentType");
+        autoDetectedContentType(false);
+        return setHeader(HttpHeaderNames.CONTENT_TYPE, contentType);
```
hyangtack: Just curious. What if we use `getOrCreateHeaders().contentType(contentType)`?

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -130,18 +142,84 @@ public HttpFileServiceBuilder autoIndex(boolean autoIndex) {
...
+
+    /**
+     * Sets the {@code "cache-control"} header. This method is a shortcut of:
+     * <pre>{@code
+     * builder.setHeader(HttpHeaderNames.CACHE_CONTROL, cacheControl);
+     * }</pre>
+     */
+    public HttpFileServiceBuilder cacheControl(CacheControl cacheControl) {
+        requireNonNull(cacheControl, "cacheControl");
+        return setHeader(HttpHeaderNames.CACHE_CONTROL, cacheControl);
```
hyangtack: nit: How about adding `HttpHeaders.cacheContrl(cc)` as well? It would be useful because our user doesn't need to write the header name. What do you think?

trustin: Sure. Let me add one. :-)

core/src/main/java/com/linecorp/armeria/server/file/AbstractHttpFileBuilder.java
trustin: No problem. :-)

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
trustin: I changed my mind not to add it because:

- we cannot add a getter even if I added the parser because we cannot determine which parser should be used. `ClientCacheControl.parse()` vs `ServerCacheControl.parse()`.
- it feels weird to have only a setter while all other methods have setters.
- we will eventually make a large change on `HttpHeaders` for immutability, so perhaps it's not a good idea to add more setters.

hyangtack: Agreed, especially the last one. :-)

core/src/main/java/com/linecorp/armeria/client/ClientCacheControl.java
```diff
@@ -0,0 +1,268 @@
...
+     * Note that any unknown directives will be ignored.
+     *
+     * @return the {@link ClientCacheControl} decoded from the specified header values.
+     */
+    public static ClientCacheControl parse(Iterable<String> directives) {
+        requireNonNull(directives, "directives");
+        final ClientCacheControlBuilder builder = new ClientCacheControlBuilder();
+        for (String d : directives) {
+            parseCacheControl(d, (name, value) -> {
+                switch (name) {
```
hyangtack: Just curious. What if using `Map<String, BiConsumer<ClientCacheControlBuilder, String>>` here? Would it be better than `switch-case`? What do you think?

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -374,6 +367,118 @@ public static boolean isContentAlwaysEmptyWithValidation(
...
+            if (!name.isEmpty()) {
+                callback.accept(Ascii.toLowerCase(name), Strings.emptyToNull(value));
+            }
+        }
+    }
+
+    /**
+     * Converts the specified {@code "cache-control"} directive value into a long integer.
+     *
+     * @return the converted value or {@code -1} if {@code value} is {@code null} or not an integer.
```
hyangtack: nit: the converted value if {@code value} is {@code 0} or a positive integer. Otherwise, {@code -1} is returned.?

core/src/main/java/com/linecorp/armeria/client/ClientCacheControl.java
trustin: Sounds good to me.

core/src/main/java/com/linecorp/armeria/server/ServerCacheControlBuilder.java
```diff
@@ -0,0 +1,185 @@
...
+        return cachePrivate(true);
+    }
+
+    /**
+     * Enables or disables the {@code "private"} directive.
+     *
+     * @param cachePrivate {@code true} to enable or {@code false} to disable.
+     */
+    public ServerCacheControlBuilder cachePrivate(boolean cachePrivate) {
+        this.cachePrivate = cachePrivate;
```
minwoox: Just wondering if we need to check the exclusivity of `private` and `public` directives.

site/src/sphinx/server-http-file.rst
```diff
@@ -47,6 +47,28 @@ an ``index.html`` file using the ``autoIndex(boolean)`` method in :api:`HttpFile
...
+
+You can specify additional response headers such as ``cache-control`` and other custom headers.
+
+.. code-block:: java
+
+    HttpFileServiceBuilder fsb =
+            HttpFileServiceBuilder.forFileSystem("/var/lib/www/images");
+
+    // Specify cache control directives.
+    ServerCacheControl cc = new ServerCacheControlBuilder()
```
minwoox: Could you import these two classes?

core/src/main/java/com/linecorp/armeria/server/ServerCacheControlBuilder.java
trustin: Made sure `cachePrivate` has precedence over `cachePublic`, i.e. if both are set, `cachePrivate()` returns `true` and `cachePublic()` returns `false`.

trustin:
Thanks for reviewing :-)

PR #1617, opened 2019-02-25
Reduce the memory footprint of the default Dropwizard `HierarchicalNa…

trustin:
…meMapper`

Motivation:

At some point, Micrometer's `Meter.Id.getTags()` started to return a
copy of its tag list. It was returning an `Iterable<Tag>` without any
additional copying before. We obviously want to avoid making a copy of
the tag list like we did before.

Modifications:

- Use `getTagsAsIterable()` instead of `getTags()`

Result:

- Less memory footprint when converting a `Meter.Id` into a Dropwizard
  meter name.


trustin:
FYI - The change occurs in Micrometer 1.1.0.

PR #1618, opened 2019-02-26
Update the example of CORS configuration for gRPC-web

hyangtack:
Motivation:
For gRPC-web, there is no example about exposing headers with `access-control-expose-headers` header.

Modifications:
- Update the example for configuring CORS.


site/src/sphinx/server-grpc.rst
```diff
@@ -135,7 +136,12 @@ you build a :api:`GrpcService`:
                               .allowRequestMethods(HttpMethod.POST) // Allow POST method.
                               // Allow Content-type and X-GRPC-WEB headers.
                               .allowRequestHeaders(HttpHeaderNames.CONTENT_TYPE,
-                                                   HttpHeaderNames.of("X-GRPC-WEB"));
+                                                   HttpHeaderNames.of("X-GRPC-WEB"))
+                              // Expose trailers of the HTTP response to the client so that it can get
+                              // the their values.
+                              .exposeHeaders(GrpcHeaderNames.GRPC_STATUS,
```
hyangtack: I'm curious if I need to use `HttpHeaderNames` instead of `GrpcHeaderNames` because `GrpcHeaderNames` is in the `internal` package. Please let me know what you think!

anuraaga: I think we can move it to common. I had it in internal since I thought it was an implementation detail of the wire protocol but didn't know about this sort of use case.

hyangtack: Ah, I see. Then, could I move it to common package in this PR? :-)

anuraaga: I think so, yeah

site/src/sphinx/server-grpc.rst
```diff
@@ -135,7 +136,12 @@ you build a :api:`GrpcService`:
                               .allowRequestMethods(HttpMethod.POST) // Allow POST method.
                               // Allow Content-type and X-GRPC-WEB headers.
                               .allowRequestHeaders(HttpHeaderNames.CONTENT_TYPE,
-                                                   HttpHeaderNames.of("X-GRPC-WEB"));
+                                                   HttpHeaderNames.of("X-GRPC-WEB"))
```
trustin: I'm curious if we need to add this to `GrpcHeaderNames`. @anuraaga, what do you think?

site/src/sphinx/server-grpc.rst
```diff
@@ -135,7 +136,12 @@ you build a :api:`GrpcService`:
...
+                              // Expose trailers of the HTTP response to the client so that it can get
+                              // the their values.
```
trustin: 'so that ...' could be removed?

site/src/sphinx/server-grpc.rst
anuraaga: I hadn't actually heard of that header until now, but it's not in the spec and is being removed from the improbable implementation

https://github.com/improbable-eng/grpc-web/issues/325

We can keep it here until `@improbable/grpc-web` updates though but after that we'd want to remove it from here too I think.

Tracking issue is this one I think - https://github.com/improbable-eng/grpc-web/issues/346

trustin: I see. Then let's leave it as it is for now and consider removing it in the future.

trustin:
Thanks, @hyangtack and reviewers.

PR #1620, opened 2019-02-26
Fix a bug where trailers is added to message headers

minwoox:
Motivation:
According to the [HTTP spec](https://tools.ietf.org/html/rfc7540#section-8.1), one HEADERS frame, zero or more DATA frames and
optionally, one HEADERS frame can be sent in request/response. Some of our code violate this rule, so we should fix them.

Modifications:
- Fix `Http(Response|Request)Aggregator` not to aggregate split trailers
- Fix Retrofit client not to add trailers to message headers
  - Currently, there's no way to preserve trailers in `retrofit` client, so just discard it.
- Fix Retrofit client to call callback when only headers is received.
- Refactor `tryWrite()` in `HttpResponseDecoder` and `DecodedHttpRequest` so that we are ready to log trailers.
- Add `drainAll(...)` to `StreamMessage`.

Result:
- Meets spec

Todo:
- Support trailers in Retrofit client after upgrading the dependency.


core/src/test/java/com/linecorp/armeria/client/HttpResponseWrapperTest.java
```diff
@@ -0,0 +1,187 @@
...
+                HttpData.ofUtf8("foo"));
+    }
+
+    @Test
+    public void headersAndTrailers() throws Exception {
+        final DecodedHttpResponse res = new DecodedHttpResponse(CommonPools.workerGroup().next());
+        final HttpResponseWrapper wrapper = httpResponseWrapper(res);
+
+        wrapper.tryWrite(HttpHeaders.of(200));
+        wrapper.tryWrite(HttpHeaders.of(HttpHeaderNames.of("bar"), "baz"));
```
trustin: Probably better making sure `tryWrite()` returns `true` in many places in this test case?

core/src/test/java/com/linecorp/armeria/client/HttpResponseWrapperTest.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private static class TestHttpResponseDecoder extends HttpResponseDecoder {
+        TestHttpResponseDecoder(Channel channel,
+                                InboundTrafficController inboundTrafficController) {
+            super(channel, inboundTrafficController);
+        }
+    }
+
+    private static List<HttpObject> unaggregate(StreamMessage<HttpObject> req) throws Exception {
```
trustin: What do you think about adding a new method to `StreamMessage`?

```java
public interface StreamMessage<T> ... {
    CompletableFuture<List<T>> drainAll();
    CompletableFuture<List<T>> drainAll(EventExecutor executor);
    CompletableFuture<List<T>> drainAll(boolean withPooledObjects);
    CompletableFuture<List<T>> drainAll(EventExecutor executor, boolean withPooledObjects);
}
```

minwoox: That's a good suggestion. Will add them. :)

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
```diff
@@ -126,18 +126,20 @@ protected EventExecutor defaultSubscriberExecutor() {
...
 
     @Override
-    void subscribe(SubscriptionImpl subscription) {
-        final Subscriber<?> subscriber = subscription.subscriber();
+    SubscriptionImpl subscribe(SubscriptionImpl subscription) {
         if (!subscribedUpdater.compareAndSet(this, 0, 1)) {
-            eventLoop.execute(() -> failLateSubscriber(this.subscription, subscriber));
-            return;
+            final SubscriptionImpl oldSubscription = this.subscription;
+            assert oldSubscription != null;
```
minwoox: This can be `false` because at this point the `subscription` couldn't be set.
It's no worry to fix it using `subscriptionUpdater` like we do in other classes (not `subscribedUpdater`), but I just want to make it sure that it's worth because I couldn't find any usage of this.
@trustin, @anuraaga, do we need this `EventLoopStreamMessage`?

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -230,6 +230,59 @@ default boolean isComplete() {
      */
     void subscribe(Subscriber<? super T> subscriber, EventExecutor executor, boolean withPooledObjects);
 
+    /**
+     * Returns a {@link CompletableFuture} that contains the list of {@link T}s which were written to this
+     * {@link StreamMessage}. The {@link CompletableFuture} can be competed exceptionally with one of the
```
trustin: How about:

```
Subscribes to this {@link StreamMessage} and retrieves all elements from it.
The returned {@link CompletableFuture} will be completed exceptionally ...
...

@return the {@link CompletableFuture} which will be completed with the list of the elements retrieved.
```

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -230,6 +230,59 @@ default boolean isComplete() {
...
+     * following exceptions:
+     * <ul>
+     *      <li>{@link IllegalStateException} if other {@link Subscriber} subscribed to this stream already</li>
+     *      <li>{@link AbortedStreamException} if this stream has been {@linkplain #abort() aborted}.</li>
+     * </ul>
+     */
+    CompletableFuture<List<T>> drainAll(EventExecutor executor);
```
trustin: Please add `@param executor`.

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -230,6 +230,59 @@ default boolean isComplete() {
...
+    /**
+     * Returns a {@link CompletableFuture} that contains the list of {@link T}s which were written to this
+     * {@link StreamMessage}. The {@link CompletableFuture} can be competed exceptionally with one of the
+     * following exceptions:
+     * <ul>
+     *      <li>{@link IllegalStateException} if other {@link Subscriber} subscribed to this stream already</li>
+     *      <li>{@link AbortedStreamException} if this stream has been {@linkplain #abort() aborted}.</li>
+     * </ul>
+     *
+     * @param withPooledObjects if {@code true}, receives the pooled {@link ByteBuf} and {@link ByteBufHolder}
```
trustin: Please add `@param executor`.

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
minwoox: Fixed. Thank you!

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
minwoox: Added. Thank you!

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -230,6 +230,66 @@ default boolean isComplete() {
...
+     * The returned {@link CompletableFuture} will be completed exceptionally when one of the
+     * following exceptions is raised:
+     * <ul>
+     *      <li>{@link IllegalStateException} if other {@link Subscriber} subscribed to this stream already</li>
+     *      <li>{@link AbortedStreamException} if this stream has been {@linkplain #abort() aborted}.</li>
+     * </ul>
+     *
+     * @param withPooledObjects if {@code true}, receives the pooled {@link ByteBuf} and {@link ByteBufHolder}
+     *                          as is, without making a copy. If you don't know what this means, use
+     *                          {@link StreamMessage#subscribe(Subscriber)}.
```
hyangtack: StreamMessage#drainAll()?

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -230,6 +230,66 @@ default boolean isComplete() {
...
+     * following exceptions is raised:
+     * <ul>
+     *      <li>{@link IllegalStateException} if other {@link Subscriber} subscribed to this stream already</li>
+     *      <li>{@link AbortedStreamException} if this stream has been {@linkplain #abort() aborted}.</li>
+     * </ul>
+     *
+     * @param executor the executor to use retrieves all elements
+     * @param withPooledObjects if {@code true}, receives the pooled {@link ByteBuf} and {@link ByteBufHolder}
+     *                          as is, without making a copy. If you don't know what this means, use
+     *                          {@link StreamMessage#subscribe(Subscriber)}.
```
hyangtack: StreamMessage#drainAll(EventExecutor)?

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
anuraaga: I think we can go ahead and remove the class now, other performance improvements make it less meaningful from looking at benchmraks.

retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/StreamingCallSubscriberTest.java
```diff
@@ -109,16 +130,18 @@ public void splitHeaders() throws Exception {
...
         subscriber.onComplete();
 
         verify(subscription, times(4)).request(1L);
 
         await().untilAsserted(() -> assertThat(callback.callbackCallingCount).isEqualTo(1));
-        await().untilAsserted(() -> assertThat(callback.response.header("content-type"))
-                .isEqualToIgnoringCase("text/plain; charset=utf-8"));
-        await().untilAsserted(() -> assertThat(callback.response.body().string()).isEqualTo("bar"));
+
+        TimeUnit.SECONDS.sleep(2);
```
hyangtack: nit: Is this necessary?

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -140,6 +142,38 @@ public void subscribe(Subscriber<? super U> subscriber, EventExecutor executor,
...
+
+    @Override
+    public CompletableFuture<List<U>> drainAll(boolean withPooledObjects) {
+        return filterElements(delegate.drainAll(withPooledObjects));
+    }
+
+    @Override
+    public CompletableFuture<List<U>> drainAll(EventExecutor executor, boolean withPooledObjects) {
+        requireNonNull(executor, "executor");
+        return filterElements(delegate.drainAll(executor, withPooledObjects), executor);
```
trustin: Probably better defining a filtering subscriber for efficiency?

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -230,6 +230,66 @@ default boolean isComplete() {
      */
     void subscribe(Subscriber<? super T> subscriber, EventExecutor executor, boolean withPooledObjects);
 
+    /**
+     * Subscribes to this {@link StreamMessage} and retrieves all elements from it.
+     * The returned {@link CompletableFuture} will be completed exceptionally when one of the
+     * following exceptions is raised:
```
trustin: The returned future will fail with other exceptions than the two exceptions below when the stream is terminated with other exceptions via `onError(Throwable)`, so I think we need to rephrase 'when ....' a little bit. How about this:

```
Subscribes to this {@link StreamMessage} and retrieves all elements from it.
The returned {@link CompletableFuture} may be completed exceptionally with the following exceptions:

<ul>
  <li>{@link IllegalStateException} if other {@link Subscriber} subscribed to this stream already.</li>
  <li>{@link AbortedStreamException} if this stream has been {@linkplain #abort() aborted}.</li>
  <li>Other exceptions that occurred due to an error while retrieving the elements.</li>
</ul>
```

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageDrainer.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableList.Builder;
+
+import io.netty.util.ReferenceCountUtil;
+
+final class StreamMessageDrainer<T> implements Subscriber<T> {
```
trustin: You could make this class non-final and add some extension point so that a subclass can add an element selectively for `FilteredStreamMessage`, e.g.

```java
class StreamMessageDrainer implements Subscriber {
    ...
    final add(T t) {
        drained.add(t);
    }
}

class FilteredStreamMessageDrainer extends StreamMessageDrainer {
    @Override
    public void onNext(T t) {
        if (...) {
            add(t);
        } else {
            ReferenceCountUtil.safeRelease(t);
        }
    }
}
```

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageDrainer.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public void onNext(T t) {
+        drained.add(t);
+    }
+
+    @Override
+    public void onError(Throwable t) {
+        future.completeExceptionally(t);
+        drained.build().forEach(ReferenceCountUtil::safeRelease);
```
trustin: - We could release first so that the objects are returned to the pool sooner.
- After draining, we should dereference `drained` so that the objects are GC'd.

core/src/main/java/com/linecorp/armeria/common/stream/EventLoopStreamMessage.java
minwoox: Ok, got it. Will remove it then. Thanks! :bow:

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
minwoox: Got it. Will add the third condition to the Javadoc of `subscribe(...)` as well.

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageDrainer.java
minwoox: I realized that we don't need this because we're already using `FilteringSubscriber`. My initial approach was just wrong. Thank you for letting me wake up!

retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/StreamingCallSubscriberTest.java
minwoox: Just added because there's no way to retrieve trailers now. Will remove after we upgrade Retrofit and are able to get trailers which lets us know the end of stream.

core/src/main/java/com/linecorp/armeria/server/DecodedHttpRequest.java
```diff
@@ -99,13 +102,26 @@ protected EventLoop defaultSubscriberExecutor() {
...
-            ctx.logBuilder().increaseRequestLength((HttpData) obj);
+        if (receivedTrailers) {
+            ReferenceCountUtil.safeRelease(obj);
+            return false;
+        }
+
+        boolean published = false;
+        if (obj instanceof HttpHeaders) { // HTTP trailers.
+            published = super.tryWrite(obj);
+            receivedTrailers = true;
```
trustin: I'm curious if we can just call `super.close()` here rather than having an additional boolean field.

minwoox: That's a good idea. Will fix it.

trustin:
Much appreciated, @minwoox

PR #1623, opened 2019-03-02
Add support for ProtoReflectionService.

### LINKED ISSUE #1621: gRPC server reflection support
https://github.com/grpc/grpc-go/blob/master/Documentation/server-reflection-tutorial.md

anuraaga:
I was originally thinking of forking `ProtoReflectionService` from upstream, but realized it's not hard to just use it as is. This implementation is quite similar to upstream which essentially runs custom logic based on an `instanceof` check.

https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java#L111

Fixes #1621

### DIFF
--- a/grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
+++ b/grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
@@ -99,6 +99,11 @@ import io.grpc.Status;
 import io.grpc.Status.Code;
 import io.grpc.StatusException;
 import io.grpc.StatusRuntimeException;
+import io.grpc.protobuf.services.ProtoReflectionService;
+import io.grpc.reflection.v1alpha.ServerReflectionGrpc;
+import io.grpc.reflection.v1alpha.ServerReflectionGrpc.ServerReflectionStub;
+import io.grpc.reflection.v1alpha.ServerReflectionRequest;
+import io.grpc.reflection.v1alpha.ServerReflectionResponse;
 import io.grpc.stub.ServerCallStreamObserver;
 import io.grpc.stub.StreamObserver;
 import io.netty.util.AsciiString;
@@ -287,6 +292,7 @@ public class GrpcServiceServerTest {
             sb.serviceUnder("/", new GrpcServiceBuilder()
                     .setMaxInboundMessageSizeBytes(MAX_MESSAGE_SIZE)
                     .addService(new UnitTestServiceImpl())
+                    .addService(ProtoReflectionService.newInstance())
                     .enableUnframedRequests(true)
                     .supportedSerializationFormats(GrpcSerializationFormats.values())
                     .build()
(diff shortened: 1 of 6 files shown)


grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
```diff
@@ -284,17 +289,24 @@ protected void configure(ServerBuilder sb) throws Exception {
...
+                            .build(),
+                    service -> service
+                            .decorate(LoggingService.newDecorator())
+                            .decorate((delegate, ctx, req) -> {
+                                ctx.log().addListener(requestLogQueue::add, RequestLogAvailability.COMPLETE);
+                                return delegate.serve(ctx, req);
+                            }));
+            sb.service(
+                    new GrpcServiceBuilder()
+                            .addService(ProtoReflectionService.newInstance())
```
trustin: Do you think it will be good to have a shortcut like `enableReflection()`?

anuraaga: It's probably ok, but I wonder if it might cause confusion too since it would automatically add a service (e.g., they would see it in DocService and might not know where it came from). With the current way, it means all added services use `addService`.

trustin: OK!

anuraaga:
Yeah - added docs

site/src/sphinx/server-grpc.rst
```diff
@@ -243,6 +243,26 @@ to all your clients before enabling this flag!
...
+    ServerBuilder sb = new ServerBuilder();
+    ...
+    sb.service(new GrpcServiceBuilder().addService(new MyHelloService())
+                                       .addService(ProtoReflectionService.newInstance())
+                                       .build());
+    ...
+    Server server = sb.build();
+    server.start();
+
+For more information, see the official `Link tutorial <https://github.com/grpc/grpc-java/blob/master/documentation/server-reflection-tutorial.md>`_.
```
hyangtack: nit: How about changing from `Link tutorial` to `gRPC Server Reflection Tutorial`?

anuraaga: Thanks - fixed. Realized I don't know rst at all. Would it be possible to migrate the docs to Markdown for easier contributions?

anuraaga: Happened to find https://www.sphinx-doc.org/en/master/usage/markdown.html

trustin: Yeah, but Markdown has never been the first citizen, so you'd better learn from other .rst pages in this project and ask us at Slack. :wink:

site/src/sphinx/server-grpc.rst
```diff
@@ -243,6 +243,26 @@ to all your clients before enabling this flag!
 
 See more details at :ref:`client-grpc`.
 
+Server Reflection
+=================
```
trustin: Needs to use `----------------` because this is a subsection.

anuraaga: Oops - guess Exception Propagation was wrong too

trustin: Yeah, but I fixed it in other PR. :-)

trustin:
Thanks, @anuraaga !

PR #1624, opened 2019-03-02
Fix a bug where ByteBuf is released before storing it for content preview

minwoox:
Motivation:
The HTTP response content is released after writing the socket in `ChannelOutboundBuffer`.
If we want to log the content, we should do it before we write it to the socket.

Modifications:
- Call `increase(Response|Request)Length` before writing to channel.

Result:
- Fewer bugs.


minwoox:
Let me add some tests!

trustin:
@kshgizmo, please review once @minwoox adds a test case. :stuck_out_tongue_winking_eye:

hyeonjames:
thanks for fixing my mistake !

minwoox:
@hyeonjames No problem. It's our mistake. 😄 

@kshgizmo Please take a look at it.

trustin:
Thanks, @minwoox and reviewers.

PR #1627, opened 2019-03-04
Fix the API for adding decorated `ServiceWithPathMappings` and update…

trustin:
… documentation

Motivation:

We recently received a good question about how to decorate a
`GrpcService`, which is a `ServiceWithPathMappings`.

I found there is not enough explanation about such a question and it is
currently not easy to specify multiple decorators for a
`ServiceWithPathMappings`.

Modifications:

- Allow specifying multiple decorators when registering a
  `ServiceWithPathMappings`.
- Add some sections about decorating `ServiceWithPathMappings` and
  `GrpcService`.

Result:

- Easier to decorate a `ServiceWithPathMappings` such as `GrpcService`.
- Better documentation.

### DIFF
--- a/site/src/sphinx/server-decorator.rst
+++ b/site/src/sphinx/server-decorator.rst
@@ -141,8 +141,8 @@ Decorating ``ServiceWithPathMappings``
 --------------------------------------
 
 :api:`ServiceWithPathMappings` is a special variant of :api:`Service` which allows a user to register multiple
-routes for a single service. It has a method called ``pathMappings()`` which returns a ``Set<PathMapping>``
-so that you do not have to specify path mappings when registering your service:
+routes for a single service. It has a method called ``pathMappings()`` which returns a ``Set`` of
+:apiplural:`PathMapping` so that you do not have to specify path mappings when registering your service:
 
 .. code-block:: java
 
--- a/site/src/sphinx/server-decorator.rst
+++ b/site/src/sphinx/server-decorator.rst
@@ -51,6 +51,12 @@ If your decorator is expected to be reusable, it is recommended to define a new 
 
 .. code-block:: java
 
+    import com.linecorp.armeria.common.HttpRequest;
+    import com.linecorp.armeria.common.HttpResponse;
+    import com.linecorp.armeria.server.Service;
+    import com.linecorp.armeria.server.ServiceRequestContext;
+    import com.linecorp.armeria.server.SimpleDecoratingService;
+
     public class AuthService extends SimpleDecoratingService<HttpRequest, HttpResponse> {
         public AuthService(Service<HttpRequest, HttpResponse> delegate) {
             super(delegate);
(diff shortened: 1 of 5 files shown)


trustin:
@perlun Looking forward to your feed back on the documentation.

site/src/sphinx/server-decorator.rst
```diff
@@ -135,6 +135,45 @@ a certain type from a server:
...
+.. _server-decorator-service-with-path-mappings:
+
+Decorating ``ServiceWithPathMappings``
+--------------------------------------
+
+:api:`ServiceWithPathMappings` is a special variant of :api:`Service` which allows a user to register multiple
+routes for a single service easily. Because it requires a special treatment unlike usual :apiplural:`Service`,
+the two following restrictions apply:
+
+- You cannot specify path mappings when registering a :api:`ServiceWithPathMappings`.
```
anuraaga: Might want to clarify that actually they can, but it won't use the `ServiceWithPathMappings` in such a case.

site/src/sphinx/server-grpc.rst
```diff
@@ -93,6 +93,30 @@ a :api:`GrpcServiceBuilder` and add it to the :api:`ServerBuilder`:
     We bound the :api:`GrpcService` without specifying any path mappings. It is because :api:`GrpcService`
     implements :api:`ServiceWithPathMappings`, which dynamically provides path mappings by itself.
 
+.. _server-grpc-decorator:
+
+Decorating a ``GrpcService``
+----------------------------
+
+Unlike a usual Armeria :api:`Service`, :api:`GrpcService` implements a special interface called
+:api:`ServiceWithPathMappings`. You cannot decorate a :api:`GrpcService` using ``Service.decorate()`` but
```
anuraaga: Ditto, `cannot` is a bit confusing since it is actually possible, but only when using an explicit path mapping.

site/src/sphinx/server-decorator.rst
trustin: Ah, yeah. Let me rephrase it a little bit.

trustin: Updated. PTAL again :-)

site/src/sphinx/server-decorator.rst
```diff
@@ -135,6 +135,70 @@ a certain type from a server:
...
+Decorating ``ServiceWithPathMappings``
+--------------------------------------
+
+:api:`ServiceWithPathMappings` is a special variant of :api:`Service` which allows a user to register multiple
+routes for a single service. It has a method called ``pathMappings()`` which returns a ``Set`` of
+:apiplural:`PathMapping` so that you do not have to specify path mappings when registering your service:
+
+.. code-block:: java
+
+    public class MyServiceWithPathMappings implements ServiceWithPathMappings<HttpRequest, HttpResponse> {
```
minwoox: Seems like we missed some import statements throughout this document.
Such as for `ServiceWithPathMappings`, `ServiceRequestContext` and `PathMapping`.

trustin: Thanks for the heads up. Added.

minwoox:
LGTM one more. 😄

trustin:
Thanks for reviewing.

PR #1635, opened 2019-03-07
Reduce the search space of AnnotationUtil.getAnnotations() and find()

### LINKED ISSUE #1632: StackOverflowError exception with annotatedService in Kotlin
Hi.
I am using armeria with kotlin lang. When I create an annotatedService a StackOverflowError exception occur in server start-up.
```
class PikaRest {

    @Get("/{id}")
    fun redirect(ctx: ServiceRequestContext, @Param("id") id: String): HttpResponse {

    }

}
```

### LINKED ISSUE #1633: Stack overflow in armeria 0.80 (works in 0.79) with AnnotatedServices in Kotlin
In #1569 the behaviour of annotations was modified to allow for recursive annotations(?). Unfortunately, there is a bug in it causing AnnotatedServices written in Kotlin to stack overflow on server init. The following is a short test
```kotlin
fun startTestWebserver() =
   ServerBuilder()
       .annotatedService("/", MiniServer)
       .apply {
           port(12341, SessionProtocol.HTTP)
       }
       .build()

object MiniServer {
   @Post("/test/{pathParam}")
   @Consumes("application/json")
   fun test(@Param("pathParam") pathParam: String?, doodle: String) = HttpResponse.of(HttpStatus.NO_CONTENT)
}
```

In kotlin there are a couple of annotations similar but with extensions to the java annotations explicitly excluded via `BUILT_IN_META_ANNOTATIONS`. Unless there are strong reasons for it, a cache might be better for avoiding stack overflows rather than expanding the black list.

trustin:
Motivation:

There is no cyclic reference detection implemented in `AnnotationUtil`,
which means an annotation like the following can trigger a
`StackOverflowError`:

    @Cyclic
    @Retention(RUNTIME)
    @interface Cyclic {}

This was not a problem before because we blacklisted well known JDK
classes that has cyclic references, but it does not work for other
classes.

Modifications:

- Added cyclic dependency detection in the traversal logic.
- Added a thread-local blacklist of the annotation types with cyclic
  dependencies.

Result:

- Annotation scanning is faster.
- Annotation scanning does not trigger a `StackOverflowException`.
- Fixes #1632
- Fixes #1633

/cc @iEAmi @tobias-

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
@@ -35,7 +35,6 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.WeakHashMap;
 import java.util.function.Predicate;
 
 import javax.annotation.Nullable;
@@ -46,6 +45,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.MapMaker;
 
 /**
  * A utility class which helps to get annotations from an {@link AnnotatedElement}.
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
```diff
@@ -70,11 +77,17 @@
     }
 
     /**
-     * Built-in meta annotations defined in {@code java.lang.annotation} package.
+     * A thread-local hash set of annotation classes that contain cyclic references.
      */
-    private static final Set<Class<? extends Annotation>> BUILT_IN_META_ANNOTATIONS =
-            ImmutableSet.of(Documented.class, Inherited.class, Native.class,
-                            Retention.class, Repeatable.class, Target.class);
+    private static final ThreadLocal<Set<Class<? extends Annotation>>> knownCyclicAnnotationTypes =
```
trustin: Chose to use a thread-local variable instead of a global one to avoid class loader leak.

trustin: Actually I have to use `WeakHashMap` to avoid class loader leak. Not enough sleep :-)

trustin:
Replaced `IdentityHashMap` with `WeakHashMap` to fix the class loader leak. Will merge once build finishes.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
anuraaga: Mostly out of curiosity, if using a weak map, do you still need to use threadlocal?

trustin: I thought `WeakHashMap` is not thread safe, no?

trustin: .. or did you mean using a concurrent version of weak key hash map?

anuraaga: Yeah was comparing to using synchronizedMap with a weak hash map.

tobias-:
Tested with simple Kotlin class and works.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotationUtil.java
trustin: Yeah, switched to `MapMaker`.

trustin:
Thanks for reviewing.

PR #1641, opened 2019-03-08
Support non-JSON Spring Actuator responses

### LINKED ISSUE #1637: Actuator endpoint accidentally json quotes plain text
See https://github.com/line/armeria/pull/1578/files#r263519826 for context

trustin:
Motivation:

Spring Actuator responses are not always JSON. It can be a plaintext or
even an octet stream.

Modifications:

- Respect the negotiated media type.
- Added support for a plaintext response represented as a `CharSequence`.
- Added support for a binary response represented as a `Resource`.
- Miscellaneous:
  - `msg` -> `req` or `res`
  - Added `final` wherever applicable.
  - Added missing `spring.factories` to `armeria-spring-boot-actuator-autoconfigure`.
  - Made all Spring Boot examples use our actuator implementation.
  - Fixed a packaging issue which made more than one `Server` 
    instantiated. (The fix originally written by @hyangtack)

Result:

- Fixes #1637
- A user can also retrieve the heap dump via the Actuator service.
- A user does not need to import `ArmeriaSpringActuatorAutoConfiguration`
  manually anymore.
- `Server` startup does not fail anymore when a user pulls both
  `armeria-spring-boot-actuator-starter` and
  `armeria-spring-boot-webflux-starter`.

### DIFF
--- a/examples/spring-boot-minimal/build.gradle
+++ b/examples/spring-boot-minimal/build.gradle
@@ -20,6 +20,8 @@ dependencies {
     compileOnly 'org.springframework.boot:spring-boot-configuration-processor'
     annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
 
+    runtime project(':spring:boot-actuator-starter')
+
     testCompile 'junit:junit'
     testCompile 'net.javacrumbs.json-unit:json-unit-fluent'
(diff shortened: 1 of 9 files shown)


adriancole:
looks like the right fix and fast!

trustin:
Thanks, reviewers :-)

PR #1656, opened 2019-03-12
Add `SystemInfo` utility class which provides the system information

trustin:
Motivation:

It'd be nice for Armeria to provide a reliable way to retrieve the
system information that requires some hack depending on OS and Java
version, such as:

- Java version
- local hostname
- current PID
- microsecond-precision time

Modifications:

- Added `SystemInfo` which provides the information mentioned in the
  'Motivation' section.

Result:

- Convenience.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
@@ -1,0 +1,342 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.common.util;
+
+import static com.google.common.base.MoreObjects.firstNonNull;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.net.IDN;
+import java.net.InetAddress;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Clock;
+import java.time.Instant;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+
+import javax.annotation.Nullable;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Ascii;
+
+/**
+ * Provides utilities for accessing the information about the current system and process.
+ */
+public final class SystemInfo {
+
+    private static final Logger logger = LoggerFactory.getLogger(SystemInfo.class);
+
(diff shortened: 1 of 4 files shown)


trustin:
@anuraaga IIRC you requested a utility class for getting the local hostname long time ago, so it'd be nice if you have a suggestion for a better class name, etc.

trustin:
A reminder to self: On Linux, you can check the symlink `/proc/self`:

```
$ ls -l /proc/self
lrwxrwxrwx 1 root root 0 Mar 12 09:37 /proc/self -> 7628
```

core/src/main/java/com/linecorp/armeria/common/util/MoreSystem.java
```diff
@@ -0,0 +1,224 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Ascii;
+
+import io.netty.util.internal.PlatformDependent;
+
+/**
+ * Provides utilities for accessing the information about the current system and process.
+ */
+public final class MoreSystem {
```
anuraaga: I guess this is following the `More` convention with the `System` class. It does read quite strangely though. Do you see it being more than just providing system information? If not, could just be `SystemInformation` or similar.

core/src/main/java/com/linecorp/armeria/common/util/MoreSystem.java
```diff
@@ -0,0 +1,224 @@
...
+    private static final class Pid {
+
+        static final int PID;
+
+        static {
+            int pid = -1;
+            // Try ProcessHandle.pid() if Java 9+.
+            if (PlatformDependent.javaVersion() >= 9) {
+                try {
+                    final Class<?> handleClass = Class.forName("java.lang.ProcessHandle", true,
```
anuraaga: What do you think about using class loading instead of reflection? I guess it might complicate the build though

core/src/main/java/com/linecorp/armeria/common/util/MoreSystem.java
trustin: Asked the naming question at Twitter and got the following:

- `Os`
- `PosixUtil`
- `LocalServerInfo`
- `SystemHosts`
- `CurrentSystem`
- `LocalSystem`
- `Sys`

Perhaps, `SystemInfo`?

core/src/main/java/com/linecorp/armeria/common/util/MoreSystem.java
trustin: Yeah, `ProcessHandle` is not available in Java 8 so it will complicate the build, especially that we pass `--release 8` to the compiler universally. Let me keep this because it simplifies the build. :-)

core/src/main/java/com/linecorp/armeria/common/util/MoreSystem.java
anuraaga: +1 to `SystemInfo`

trustin:
- Renamed to `SystemInfo`
- Added `javaVersion()` and `currentTimeMicros()`

I feel like `currentTimeMicros()` might be better hosted somewhere else, but perhaps it's just fine to keep it here.

trustin:
Bonus: On Linux, hostname is retrieved from `/proc/sys/kernel/hostname` which means we now retrieves the hostname even on a system without the `hostname` command. I guess it will make @anuraaga happy. :wink:

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -0,0 +1,341 @@
...
+
+    private static final class Hostname {
+
+        private static final Pattern HOSTNAME_PATTERN = Pattern.compile(
+                "^(?:[-_a-zA-Z0-9]|[-_a-zA-Z0-9][-_.a-zA-Z0-9]*[-_a-zA-Z0-9])$");
+
+        static final String HOSTNAME;
+
+        static {
+            // Try /proc/sys/kernel/hostname' on Linux.
```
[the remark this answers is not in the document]
trustin: Fixed :-)

trustin:
Thanks, @minwoox @hyangtack !

PR #1658, opened 2019-03-15
Fix ByteBuf leak from Spring WebFlux integration

hyangtack:
Motivation:
 `ArmeriaServerHttpResponse`, especially its super class, `AbstractServerHttpResponse`, uses `ChannelSendOperator#WriteBarrier` when writing a response.
The `WriteBarrier` internally caches the first published object, but it does nothing when the cached object is discarded, which might cause `ByteBuf` leak.

Modifications:
- Do not inherit `AbstractServerHttpResponse` anymore, which causes mis-ordering of invoking `onNext` of `PublisherBasedHttpResponse`.
- Add `HttpResponseProcessor` between `PublisherBasedHttpResponse` and the content publisher.
  - It always requests one element to the publisher when subscribing, and then caches the first published object.
  - When a Spring controller returns `Mono`, it always received `request(unbounded)` signal internally, so we need to have a reference to the published object in order to release it when it is discarded.
- Add test for checking `ByteBuf` leak.

Result:
- No more `ByteBuf` leak.

### DIFF
--- a/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
+++ b/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
@@ -23,7 +23,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
@@ -72,17 +72,27 @@ import reactor.core.publisher.Mono;
 final class ArmeriaServerHttpResponse implements ServerHttpResponse {
     private static final Logger logger = LoggerFactory.getLogger(ArmeriaServerHttpResponse.class);
 
-    /**
-     * COMMITTING -> COMMITTED is the period after doCommit is called but before
-     * the response status and headers have been applied to the underlying
-     * response during which time pre-commit actions can still make changes to
-     * the response status and headers.
-     */
     private enum State {
-        NEW, COMMITTING, COMMITTED
+        /**
+         * The initial state.
+         */
+        NEW,
+        /**
+         * Started to prepare an {@link com.linecorp.armeria.common.HttpHeaders} to be emitted.
+         */
+        COMMITTING,
+        /**
+         * The {@link com.linecorp.armeria.common.HttpHeaders} has been ready and not allowed for
+         * modification anymore.
+         */
+        COMMITTED
     }
 
-    private final AtomicReference<State> state = new AtomicReference<>(State.NEW);
+    private static final AtomicReferenceFieldUpdater<ArmeriaServerHttpResponse, State> stateUpdater =
+            AtomicReferenceFieldUpdater.newUpdater(ArmeriaServerHttpResponse.class, State.class, "state");
+
+    private volatile State state = State.NEW;
+
     private final HttpHeaders headers = new HttpHeaders();
     private final MultiValueMap<String, ResponseCookie> cookies = new LinkedMultiValueMap<>();
     private final List<Supplier<? extends Mono<Void>>> commitActions = new ArrayList<>(4);
@@ -108,7 +118,7 @@ final class ArmeriaServerHttpResponse implements ServerHttpResponse {
 
     @Override
     public boolean setStatusCode(@Nullable HttpStatus status) {
-        if (state.get() != State.COMMITTED) {
+        if (state != State.COMMITTED) {
             if (status != null) {
                 armeriaHeaders.status(status.value());
             }
@@ -126,14 +136,14 @@ final class ArmeriaServerHttpResponse implements ServerHttpResponse {
 
     @Override
     public MultiValueMap<String, ResponseCookie> getCookies() {
-        return state.get() == State.COMMITTED ? CollectionUtils.unmodifiableMultiValueMap(cookies)
-                                              : cookies;
+        return state == State.COMMITTED ? CollectionUtils.unmodifiableMultiValueMap(cookies)
+                                        : cookies;
     }
 
     @Override
     public void addCookie(ResponseCookie cookie) {
         requireNonNull(cookie, "cookie");
-        checkState(state.get() != State.COMMITTED,
+        checkState(state != State.COMMITTED,
                    "Can't add the cookie %s because the HTTP response has already been committed",
                    cookie);
         getCookies().add(cookie.getName(), cookie);
@@ -151,7 +161,7 @@ final class ArmeriaServerHttpResponse implements ServerHttpResponse {
 
     @Override
     public boolean isCommitted() {
-        return state.get() != State.NEW;
+        return state != State.NEW;
     }
 
     @Override
@@ -176,7 +186,7 @@ final class ArmeriaServerHttpResponse implements ServerHttpResponse {
     }
 
     private Mono<Void> doCommit(@Nullable Supplier<? extends Mono<Void>> writeAction) {
-        if (!state.compareAndSet(State.NEW, State.COMMITTING)) {
+        if (!stateUpdater.compareAndSet(this, State.NEW, State.COMMITTING)) {
             return Mono.empty();
         }
 
@@ -198,7 +208,7 @@ final class ArmeriaServerHttpResponse implements ServerHttpResponse {
                 armeriaHeaders.add(HttpHeaderNames.SET_COOKIE, cookieValues);
             }
 
-            state.set(State.COMMITTED);
+            stateUpdater.set(this, State.COMMITTED);
         }));
 
         if (writeAction != null) {
@@ -357,7 +367,7 @@ final class ArmeriaServerHttpResponse implements ServerHttpResponse {
             subscriber().onSubscribe(new Subscription() {
                 @Override
                 public void request(long n) {
-                    if (!isValidDemand(n)) {
+                    if (!validateDemand(n)) {
                         cancel();
                         return;
                     }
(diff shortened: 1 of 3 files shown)


spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -97,12 +105,13 @@
...
         return Mono.defer(() -> {
-            final HttpResponse response = HttpResponse.of(
-                    Flux.concat(Mono.just(headers), publisher.map(factoryWrapper::toHttpData))
-                        // Publish the response stream on the event loop in order to avoid the possibility of
-                        // calling subscription.request() from multiple threads while publishing messages
-                        // with onNext signals or starting the subscription with onSubscribe signal.
-                        .publishOn(Schedulers.fromExecutor(ctx.eventLoop())));
+            final HttpResponse response = HttpResponse.of(new HttpResponsePublisher(
+                    headers, publisher.map(factoryWrapper::toHttpData)
+                                      // HttpResponsePublisher is not a thread-safe, so we make the upstream
```
trustin: a thread-safe -> thread-safe

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -193,4 +202,157 @@ static Cookie toNettyCookie(ResponseCookie resCookie) {
         cookie.setPath(resCookie.getPath());
         return cookie;
     }
+
+    /**
+     * The {@link ChannelSendOperator.WriteBarrier} caches the first published object but it does nothing
+     * when the object is discarded, which might cause {@link ByteBuf} leak. This {@link Publisher}
```
trustin: - the object is discarded -> it is discarded

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -193,4 +202,157 @@ static Cookie toNettyCookie(ResponseCookie resCookie) {
...
+     * and {@link Subscriber} will doing with similar behavior but will release the cached object when
```
trustin: will doing with similar behavior -> behaves similarly

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -193,4 +202,157 @@ static Cookie toNettyCookie(ResponseCookie resCookie) {
...
+                    public void cancel() {
+                        // Do nothing because the upstream has already completed.
+                    }
+                });
+            }
+
+            private void releaseFirstContent() {
+                if (state != State.FIRST_CONTENT_SENT &&
+                    firstContent instanceof ReferenceCounted) {
+                    logger.debug("Release the cached content: {}", firstContent);
```
trustin: Release ... -> Releasing the first cached content?

spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/ByteBufLeakTest.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        // The buffer allocation might be done after closing the socket. So waiting a little before
+        // checking the allocated buffers.
+        Thread.sleep(3000);
+
+        await().untilAsserted(this::makeSureReleasedAllAllocatedBuffers);
+    }
+
+    private void makeSureReleasedAllAllocatedBuffers() {
```
trustin: - How about doing `await().untilAsserted(...)` inside this method, because this method is always used with `untilAsserted()`?
- How about renaming: `ensureAllBuffersAreReleased`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -193,4 +202,157 @@ static Cookie toNettyCookie(ResponseCookie resCookie) {
...
+                this.headers = headers;
+                this.subscriber = subscriber;
+            }
+
+            @Override
+            public void onSubscribe(Subscription s) {
+                assert upstreamSubscription == null;
+                upstreamSubscription = s;
+                // To get the cached object from the upstream which is ChannelSendOperator#WriteBarrier,
+                // request 1 object to the publisher before finishing subscribing.
```
trustin: finishing subscribing -> the subscription is finished?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -193,4 +202,157 @@ static Cookie toNettyCookie(ResponseCookie resCookie) {
...
+                                subscriber.onComplete();
+                            } else {
+                                subscriber.onError(cause);
+                            }
+                        }
+                    }
+
+                    @Override
+                    public void cancel() {
+                        // Do nothing because the upstream has already completed.
```
minwoox: has already been?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -193,4 +203,167 @@ static Cookie toNettyCookie(ResponseCookie resCookie) {
...
+            subscriber().onNext(o);
+        }
+
+        private void finishSubscribing(Subscription s) {
+            assert firstContent != null;
+            subscriber().onSubscribe(new Subscription() {
+                @Override
+                public void request(long n) {
+                    if (!isValidDemand(n)) {
+                        return;
```
minwoox: `releaseFirstContent();` here?

hyangtack: Oops! Thanks! 👍

spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/ByteBufLeakTest.java
```diff
@@ -0,0 +1,156 @@
...
+                        outWriter.print("GET /empty HTTP/1.1\r\n\r\n");
+                        break;
+                }
+                outWriter.flush();
+            }
+        }
+
+        // The buffer allocation might be done after closing the socket. So waiting a little before
+        // checking the allocated buffers.
+        Thread.sleep(3000);
```
trustin: If the number of allocated buffers is deterministic, we could use `await()` instead? e.g. `await().untilAsserted(() -> assertThat(allocatedBuffers).hasSize(...));`

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -193,4 +274,236 @@ static Cookie toNettyCookie(ResponseCookie resCookie) {
...
+
+        private void releaseFirstContent() {
+            if (state != PublishingState.FIRST_CONTENT_SENT &&
+                firstContent instanceof ReferenceCounted) {
+                logger.debug("Releasing the first cached content: {}", firstContent);
+                ReferenceCountUtil.safeRelease(firstContent);
+            }
+        }
+
+        private boolean isValidDemand(long n) {
```
trustin: How about renaming to `validateDemand()` because this method has side effect?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
trustin: and maybe a test case for this case?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -15,126 +15,204 @@
...
+     * COMMITTING -> COMMITTED is the period after doCommit is called but before
+     * the response status and headers have been applied to the underlying
+     * response during which time pre-commit actions can still make changes to
+     * the response status and headers.
+     */
+    private enum State {
+        NEW, COMMITTING, COMMITTED
+    }
+
+    private final AtomicReference<State> state = new AtomicReference<>(State.NEW);
```
trustin: Consider using `AtomicReferenceUpdater` for less indirection and memory footprint.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -15,126 +15,214 @@
...
+                    getCookies().values().stream()
+                                .flatMap(Collection::stream)
+                                .map(ArmeriaServerHttpResponse::toNettyCookie)
+                                .map(ServerCookieEncoder.LAX::encode)
+                                .collect(toImmutableList());
+            if (!cookieValues.isEmpty()) {
+                armeriaHeaders.add(HttpHeaderNames.SET_COOKIE, cookieValues);
+            }
+
+            stateUpdater.set(this, State.COMMITTED);
```
minwoox: just `state = State.COMMITTED`?

hyangtack:
Thanks, @trustin and @minwoox !

PR #1665, opened 2019-03-20
Create event-loop threads that make Project Reactor happy.

trustin:
Motivation:

Project Reactor determines whether a thread is non-blocking or not by
checking if the thread implements a special tag interface `NonBlocking`.
Because our event loop threads do not implement it, Project Reactor's
`Schedulers.isInNonBlockingThread()` will return `false`.

Modification:

- Added `reactor.core.scheduler.NonBlocking` interface to `armeria-core`.
- Added `EventLoopThreadFactory` that creates an `EventLoopThread`.
- Added `EventLoopThread` which implements `NonBlocking`.
- Updated the existing code to use `EventLoopThreadFactory` instead of
  `DefaultThreadFactory` wherever possible.

Result:

- Project Reactor and its users are happy.
- We are a little bit unhappy due to ugliness of the fix.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/EventLoopThread.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/EventLoopThread.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/linecorp/armeria/common/util/EventLoopThread.java
```diff
@@ -0,0 +1,29 @@
...
+ * under the License.
+ */
+package com.linecorp.armeria.common.util;
+
+import io.netty.util.concurrent.FastThreadLocalThread;
+import reactor.core.scheduler.NonBlocking;
+
+/**
+ * A {@link FastThreadLocalThread} that implements {@link NonBlocking} tag interface to make Project Reactor
+ * happy.
```
minwoox: 😄

PR #1666, opened 2019-03-20
Add `grpc-service` example

hyangtack:
Motivation:
We don't have an example for gRPC service. So it'd be useful for our user if we have it.

Modifications:
- Added `grpc-service` in the example module.


hyangtack:
I also changed `update-examples.sh` in armeria-example repository.
https://gist.github.com/hyangtack/995bf99079f97b3fa7719bd4b65cd63f

examples/grpc-service/src/main/proto/hello.proto
```diff
@@ -0,0 +1,20 @@
...
+  rpc BidiHello (stream HelloRequest) returns (stream HelloReply) {}
+}
+
+message HelloRequest {
+  string name = 1;
+}
+
+message HelloReply {
+  string message = 1;
+}
```
trustin: newline?

examples/grpc-service/src/main/java/example/armeria/grpc/Main.java
```diff
@@ -0,0 +1,49 @@
...
+                .https(httpsPort)
+                .tlsSelfSigned()
+                .service(new GrpcServiceBuilder()
+                                 .addService(new HelloServiceImpl())
+                                 // See https://github.com/grpc/grpc-java/blob/master/documentation/server-reflection-tutorial.md
+                                 .addService(ProtoReflectionService.newInstance())
+                                 .supportedSerializationFormats(GrpcSerializationFormats.values())
+                                 .enableUnframedRequests(true)
+                                 .build())
+                // You can access the documentation service on http://127.0.0.1:8080/docs.
```
trustin: on -> at?

examples/README.md
```diff
@@ -4,6 +4,10 @@
   - Learn how to write an HTTP service using annotations.
   - See [Annotated services](https://line.github.io/armeria/server-annotated-service.html).
 
+- `grpc-service`
```
trustin: How about just `grpc`?

examples/README.md
```diff
@@ -4,6 +4,10 @@
...
+  - Learn how to write a gRPC service with Armeria gRPC module.
+  - See [GrpcService](https://line.github.io/armeria/server-grpc.html).
```
trustin: How about:

```md
See [Running a gRPC service](...) and [Calling a gRPC service](...).
```

examples/README.md
hyangtack: It caused a build failure because this module depends on `:grpc` module. Gradle thought it's a cyclic dependency. :-(

examples/README.md
hyangtack: Oops! Will fix. :-)

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+     * Sends an {@link HelloReply} immediately when receiving a request.
+     */
+    @Override
+    public void hello(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
+        responseObserver.onNext(buildReply(toMessage(request.getName())));
+        responseObserver.onCompleted();
+    }
+
+    /**
+     * Sends an {@link HelloReply} 3 seconds later after receiving a request.
```
minwoox: nit: could remove `later`?

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    @Override
+    public void lazyHello(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
+        // Respond 3 seconds later.
+        Mono.delay(Duration.ofSeconds(3))
```
anuraaga: I don't think it's typical for gRPC services to use reactor idioms like this and feel it makes the example hard to understand. Do you think it makes more sense to implement with armeria patterns (which also teaches RequestContext which is always very important)?

```java
RequestContext.current().contextAwareEventLoop().schedule(...)
```

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            .subscribe(unused -> responseObserver.onNext(buildReply(toMessage(request.getName()))),
+                       responseObserver::onError, responseObserver::onCompleted);
+    }
+
+    /**
+     * Sends 5 {@link HelloReply} responses when receiving a request.
+     */
+    @Override
+    public void lotsOfReplies(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
+        Flux.interval(Duration.ofSeconds(1))
```
anuraaga: This would be less elegant without the reactor, but I might still lean towards making the example use as little non-gRPC / non-armeria concepts as possible to make it easier to read.

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
hyangtack: Actually we started to write this example for a user who uses gRPC and Spring WebFlux together, and we want to show how to use gRPC with Project reactor. 

But I agree with your opinion about teaching how to use `RequestContext`, so how about changing this `lazyHello` example to use `RequestContext.current().contextAwareEventLoop().schedule(...)` and leaving the `lotsOfReplies` example as it is? Make sense?

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+import example.armeria.grpc.Hello.HelloReply;
+import example.armeria.grpc.Hello.HelloRequest;
+import example.armeria.grpc.HelloServiceGrpc.HelloServiceImplBase;
+import io.grpc.stub.StreamObserver;
+import reactor.core.publisher.Flux;
+
+public class HelloServiceImpl extends HelloServiceImplBase {
+
+    /**
+     * Sends an {@link HelloReply} immediately when receiving a request.
```
trustin: a HelloReply

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+        responseObserver.onNext(buildReply(toMessage(request.getName())));
+        responseObserver.onCompleted();
+    }
+
+    /**
+     * Sends an {@link HelloReply} 3 seconds after receiving a request.
+     */
+    @Override
+    public void lazyHello(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
+        // You can use the event loop for scheduling.
```
trustin: for scheduling a task.

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+            responseObserver.onNext(buildReply(toMessage(request.getName())));
+            responseObserver.onCompleted();
+        }, 3, TimeUnit.SECONDS);
+    }
+
+    /**
+     * Sends an {@link HelloReply} using {@code blockingTaskExecutor}.
+     *
+     * @see <a href="https://line.github.io/armeria/server-grpc.html#blocking-service-implementation">
+     *      Blocking service implementation</a>
```
trustin: How about breaking after `Blocking` so that no space is added between `>` and `Blocking`?

```
@see <a href="https://line.github.io/armeria/server-grpc.html#blocking-service-implementation">Blocking
     service implementation</a>
```

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+     */
+    @Override
+    public void blockingHello(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
+        // Unlike upstream gRPC-java, Armeria does not run service logic in a separate thread pool by default.
```
trustin: java -> Java

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+        // Thus this method will run in the event loop, which means that you can suffer the performance
```
trustin: Thus this -> Therefore, this

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+        // degradation if you call a blocking API in this method. In this case, you have the following options:
+        //
+        // 1. Call a blocking API in the blockingTaskExecutor provided by Armeria.
+        // 2. Set GrpcServiceBuilder.useBlockingTaskExecutor(true) when building your GrpcService.
+        // 3. Call a blocking API in the separate thread pool you managed.
```
trustin: you managed -> you manage

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+        //
+        // You can see the option 1 in this example.
```
trustin: In this example, we chose the option 1:

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
anuraaga: I think the callbacks may or may not be called with context mounted, which isn't good practice. Should this example mention the pattern to make sure the `subscribe` callbacks are called with the context aware loop? I think it's `subscribeOn` but I'm not super-familiar with these APIs.

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+                // Do nothing.
+            }
+            responseObserver.onNext(buildReply(toMessage(request.getName())));
+            responseObserver.onCompleted();
+        });
+    }
+
+    /**
+     * Sends 5 {@link HelloReply} responses when receiving a request.
+     */
```
trustin: How about adding `@see #lazyHello(HelloRequest, StreamObserver)` ?

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+    @Override
+    public void lotsOfReplies(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
+        // You can also write this code without Reactor like 'lazyHello' example.
```
trustin: like we did in the 'lazyHello()' method above?

examples/grpc-service/src/main/java/example/armeria/grpc/Main.java
```diff
@@ -0,0 +1,52 @@
...
+                                 .addService(ProtoReflectionService.newInstance())
+                                 .supportedSerializationFormats(GrpcSerializationFormats.values())
+                                 .enableUnframedRequests(true)
+                                 // You can set useBlockingTaskExecutor(true) in order to execute all gRPC
+                                 // methods in the blockingTaskExecutor thread pool.
+                                 // .useBlockingTaskExecutor(true)
+                                 .build())
+                // You can access the documentation service at http://127.0.0.1:8080/docs.
+                // See https://line.github.io/armeria/server-docservice.html for more information.
+                .serviceUnder("/docs", new DocService())
```
trustin: How about adding an example request?

examples/grpc-service/src/main/java/example/armeria/grpc/Main.java
```diff
@@ -0,0 +1,52 @@
...
+    public static void main(String[] args) throws Exception {
+        final Server server = newServer(8080, 8443);
+
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            server.stop().join();
+            logger.info("Server has been stopped.");
+        }));
+
+        server.start().join();
+        logger.info("Server has been started.");
```
trustin: How about printing the URL to the docservice page, so that a user can play with it quickly? (Maybe we should do the same for other examples with docservice.)

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
hyangtack: Thanks! Good point. Will make a fix for that. :-)

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
hyangtack: Oops! Thanks!

examples/grpc-service/src/main/java/example/armeria/grpc/Main.java
hyangtack: I'm not sure but it seems that `GrpcDocServicePlugin` doesn't support it. @anuraaga Am I right?

anuraaga: I remember adding support (changing some type signatures to Object to make it happen) but admittedly have not been using it in production even though I should 😣

If it's not working I'll look into it

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+        Flux.interval(Duration.ofSeconds(1))
+            .take(5)
+            .map(index -> "Hello, " + request.getName() + "! (sequence: " + (index + 1) + ')')
+            // You can make your Flux/Mono publish the signals in the RequestContext-aware executor.
+            .publishOn(Schedulers.fromExecutor(RequestContext.current().contextAwareExecutor()))
+            .subscribe(message -> {
+                           // Ensure this callback is executed in the RequestContext-aware executor.
+                           final ServiceRequestContext ctx = RequestContext.current();
+                           responseObserver.onNext(buildReply(message));
```
minwoox: Probably better to add `assert ctx != null;`?

examples/grpc-service/src/main/java/example/armeria/grpc/Main.java
hyangtack: `exampleRequestForMethod` works so I added an example request for 3 unary methods. Thanks @minwoox :-)

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
hyangtack: I thought it but `RequestContext.current()` returns non-null context or throws an `IllegalStateException`. So the assertion seems meaningless. 🤔

minwoox: Yeah, right. :) I just thought that the readers might be suspicious because we pulled it and did nothing. But the comment explains well, so probably not need it. 😄

hyangtack: Yes, I wanted to do something with it but there was nothing to do.. 😸

examples/grpc-service/src/main/java/example/armeria/grpc/HelloServiceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+    @Override
+    public void lotsOfReplies(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
+        // You can also write this code without Reactor like 'lazyHello' example.
+        Flux.interval(Duration.ofSeconds(1))
+            .take(5)
+            .map(index -> "Hello, " + request.getName() + "! (sequence: " + (index + 1) + ')')
+            // You can make your Flux/Mono publish the signals in the RequestContext-aware executor.
+            .publishOn(Schedulers.fromExecutor(RequestContext.current().contextAwareExecutor()))
+            .subscribe(message -> {
+                           // Ensure this callback is executed in the RequestContext-aware executor.
```
anuraaga: `Ensure` sounds like this command is required to run on the executor 

`Confirm this call back is being executed on...`

hyangtack: Fixed. Thanks!

PR #1667, opened 2019-03-20
Show RequestBody form only when the HTTP method is PUT or POST in Doc…

minwoox:
…Service

Motivation:

Currently, documentation service shows request body form regardless of the HTTP method.
We should show it only when the HTTP method is PUT or POST.

Modifications:

- Show request body form only when the HTTP method is POST or PUT.
- Split `MethodPage` into several `SFC`s and `DebugPage`.
  - All the states in `MethodPage` are now moved into `DebugPage`.

Result:

- The request body form is no longer shown when the HTTP method is not POST or PUT.

Todos:

- More refactoring (divide `DebugPage` into `RpcDebugPage` and `HttpDebugPage`?)

### DIFF
--- a/docs-client/src/lib/transports/grpc-unframed.ts
+++ b/docs-client/src/lib/transports/grpc-unframed.ts
@@ -28,9 +28,13 @@ export default class GrpcUnframedTransport extends Transport {
 
   protected async doSend(
     method: Method,
-    bodyJson: string,
     headers: { [name: string]: string },
+    bodyJson?: string,
   ): Promise<string> {
+    if (!bodyJson) {
+      throw new Error('gRPC request should have body');
+    }
+
     const endpoint = method.endpoints.find((ep) =>
       ep.availableMimeTypes.includes(GRPC_UNFRAMED_MIME_TYPE),
     );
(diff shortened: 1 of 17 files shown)


docs-client/src/lib/transports/grpc-unframed.ts
```diff
@@ -28,9 +28,13 @@ export default class GrpcUnframedTransport extends Transport {
 
   protected async doSend(
     method: Method,
-    bodyJson: string,
     headers: { [name: string]: string },
+    bodyJson?: string,
   ): Promise<string> {
+    if (!bodyJson) {
+      throw new Error('gRPC request should have body');
```
trustin: - gRPC request -> A gRPC request
- should -> must
- missing period

docs-client/src/lib/transports/thrift.ts
```diff
@@ -27,9 +27,13 @@ export default class ThriftTransport extends Transport {
 
   protected async doSend(
     method: Method,
-    bodyJson: string,
     headers: { [name: string]: string },
+    bodyJson?: string,
   ): Promise<string> {
+    if (!bodyJson) {
+      throw new Error('Thrift request should have body');
```
trustin: A Thrift request must have body.

minwoox:
Thanks for reviewing!

PR #1675, opened 2019-03-22
Add start/stop parameters to StartStopSupport

trustin:
Motivation:

A user sometimes wants to pass an argument when starting or stopping
something.

Modifications:

- Added two more type parameters to `StartStopSupport`.
- Added `StartStopSupport.start(arg, failIfStarted)`
- Added `StartStopSupport.stop(arg)`
- Modified the parameters of `StartStop.doStart()`, `doStop()` and
  listener notification methods.

Result:

- A user can pass a startup/shutdown argument when using
  `StartStopSupport`.
- `StartStopSupport` has breaking changes.


trustin:
Thanks for reviewing. :smile:

PR #1688, opened 2019-03-28
Add InvalidResponseHeadersException

trustin:
Motivation:

When a Thrift client throws an `InvalidResponseException`, a user need
to know the details about the response, such as HTTP status code.

Modifications:

- Added `InvalidResponseHeadersException` which extends
  `InvalidResponseException`
- Throw `InvalidResponseHeadersException` instead of
  `InvalidResponseException` where possible.

Result:

A user have enough information to determine the status of a Thrift
endpoint.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/InvalidResponseHeadersException.java
+++ b/core/src/main/java/com/linecorp/armeria/client/InvalidResponseHeadersException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 LINE Corporation
+ * Copyright 2019 LINE Corporation
  *
  * LINE Corporation licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
--- a/core/src/main/java/com/linecorp/armeria/client/InvalidResponseHeadersException.java
+++ b/core/src/main/java/com/linecorp/armeria/client/InvalidResponseHeadersException.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2015 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.client;
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/linecorp/armeria/client/InvalidResponseHeadersException.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright 2015 LINE Corporation
```
hyangtack: nit: 2015->2019 :-)

linna-li:
nit(in PR description): Throw `InvalidResponseHeadersException` instead of `InvalidResponseException` where possible.

trustin:
@linna-li Thanks! Updates the PR description.

thrift/src/test/java/com/linecorp/armeria/server/thrift/ThriftSerializationFormatsTest.java
```diff
@@ -122,8 +122,8 @@ public void notDefault() throws Exception {
     public void notAllowed() throws Exception {
         final HelloService.Iface client =
                 Clients.newClient(server.uri(TEXT, "/hellobinaryonly"), HelloService.Iface.class);
-        thrown.expect(InvalidResponseException.class);
-        thrown.expectMessage("415 Unsupported Media Type");
+        thrown.expect(InvalidResponseHeadersException.class);
+        thrown.expectMessage(":status=415");
```
linna-li: is it needed to add unit test case for server error?
like `status is (503) `?

trustin: Sure. Let me add one.

trustin: Added a test case to `ThriftOverHttpClientTest`

linna-li: Thanks for the unit test! LGTM

core/src/main/java/com/linecorp/armeria/client/InvalidResponseHeadersException.java
```diff
@@ -0,0 +1,66 @@
...
+import static java.util.Objects.requireNonNull;
+
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.common.HttpHeaders;
+
+/**
+ * An {@link InvalidResponseException} raised when a client received a response with invalid headers.
+ */
+public class InvalidResponseHeadersException extends InvalidResponseException {
```
minwoox: How about naming as `InvalidResponseStatusException` because this is raised only when the status is not `OK`?

trustin: Perhaps this exception could be used somewhere else? I deliberately made it broadly applicable.

minwoox:
Thanks, @trustin!

PR #1690, opened 2019-03-29
Ensure to close HttpResponseWrapper on the channel event loop

hyangtack:
Motivation:
There's chance to close `HttpResponseWrapper` on an event loop which is different from the channel event loop.
It is definitely not intended action, and it can cause `IllegalReferenceCountException` due to double-free if a user configured a content previewer.

Modifications:
- Used channel event loop when closing the `HttpResponseWrapper`.

Result:
- No more `IllegalReferenceCountException`.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/client/HttpResponseDecoderTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/HttpResponseDecoderTest.java
@@ -86,18 +86,12 @@ public class HttpResponseDecoderTest {
                 // which means that we have more chance to reproduce the bug if two threads are racing
                 // for notifying RESPONSE_END to listeners.
                 .contentPreview(100)
-                // In order to use a different thread to send a request.
+                // In order to use a different thread to to subscribe to the response.
                 .decorator(new RetryingHttpClientBuilder(strategy).maxTotalAttempts(2).newDecorator())
                 .decorator((delegate, ctx, req) -> {
                     final AtomicReference<Thread> responseStartedThread = new AtomicReference<>();
                     ctx.log().addListener(log -> {
                         responseStartedThread.set(Thread.currentThread());
-                        if (!ctx.eventLoop().inEventLoop()) {
-                            // Actually we don't expect this situation, but it is prepared for the case.
-                            logger.error("{} Response started in another thread: {} != {}",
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/linecorp/armeria/client/HttpResponseDecoderTest.java
```diff
@@ -0,0 +1,128 @@
...
+    public void confirmResponseStartAndEndInTheSameThread() throws InterruptedException {
+        final AtomicBoolean failed = new AtomicBoolean();
+        final RetryStrategy strategy =
+                (ctx, cause) -> CompletableFuture.completedFuture(Backoff.withoutDelay());
+        final HttpClient client = new HttpClientBuilder(protocol.uriText() + "://127.0.0.1:" + rule.httpPort())
+                // This increases the execution duration of 'endResponse0' of the DefaultRequestLog,
+                // which means that we have more chance to reproduce the bug if two threads are racing
+                // for notifying RESPONSE_END to listeners.
+                .contentPreview(100)
+                // In order to use a different thread to send a request.
```
minwoox: nit: to send a request -> to subscribe to the response.

core/src/test/java/com/linecorp/armeria/client/HttpResponseDecoderTest.java
```diff
@@ -0,0 +1,128 @@
...
+                .decorator(new RetryingHttpClientBuilder(strategy).maxTotalAttempts(2).newDecorator())
+                .decorator((delegate, ctx, req) -> {
+                    final AtomicReference<Thread> responseStartedThread = new AtomicReference<>();
+                    ctx.log().addListener(log -> {
+                        responseStartedThread.set(Thread.currentThread());
+                        if (!ctx.eventLoop().inEventLoop()) {
```
minwoox: I think this can cause this test failure when a connection is not established by the VM conditions or etc.(In that case, this is executed by the thread who calls `execute()`.)
So I think it's perfectly fine to remove this and only check the thread is same or not. (Still, in that case, the thread which starts end ends the response will be the same.)
WDYT?

hyangtack: Thanks! I misunderstood what you explained offline before. 👍

PR #1697, opened 2019-04-04
Fix a bug where `DnsEndpointGroup` does not respect search domains

### LINKED ISSUE #1694: DNS query failed in k8s by service name
Hi,  the DnsAddressEndpointGroup can not resolve successfully by the k8s service name, and must use a fully qualified domain name.
e.g. 

```java
//query failed
new DnsAddressEndpointGroupBuilder("service-a").port(8080).build()
//only a fully qualified domain name work
new DnsAddressEndpointGroupBuilder("service-a.infra.svc.cluster.local").port(8080).build()
````

trustin:
Motivation:

When using `Dns*EndpointGroup` in a system with search domains in
`/etc/resolv.conf`, the resolver should try all search domains until the
host name is resolved.

However, it treats the specified host name as a full domain name by
appending a dot (.) to the specified host name.

Modifications:

- Introduce a new `DnsQuestion` implementation called
  `DnsQuestionWithoutTrailingDot` which does not append a dot at the end
  of the host name.

Result:

- `Dns*EndpointGroup` respects the search domains in `/etc/resolv.conf`.
- Fixes #1694

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
@@ -32,7 +32,6 @@ import com.linecorp.armeria.common.CommonPools;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
 import io.netty.channel.EventLoop;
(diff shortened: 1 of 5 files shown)


anuraaga:
Just wondering, should we file an issue to Netty to upstream something along these lines?

PR #1699, opened 2019-04-05
Make CORS policy to be applied to the specific paths

### LINKED ISSUE #1698: Add PathMapping to CorsPolicy
Currently, `CorsService` finds a CORS policy by matching an `Origin` header value. So a user has to configure different `CorsService`s if he or she wants to apply different policies for different paths even the `Origin` domains are the same.

`@CorsDecorator` can handle this kind of situation because it can be specified on each methods in an annotated service, but using the annotation will be worked only if a user knows what origins are allowed at the compile time, which means that it's not suitable for configuring the origins dynamically. Also, it's only for the annotated services.

I think, if the `CorsPolicy` has a `PathMapping` property, `CorsService` will be able to find a policy by using an `Origin` header value and the request path. Therefore, a user can apply CORS policies for different paths more convenient way.

hyangtack:
Motivation:
Currently, there is no way to apply a CORS policy to the specific paths. So it is hard for a user to configure different policies for different paths when the `Origin` domains are the same.
`@CorsDecorator` can handle this kind of situation because it can be specified on each methods in an annotated service, but using the annotation will be worked only if a user knows what origins are allowed at the compile time, which means that it's not suitable for configuring the origins dynamically. Also, it's only for the annotated services.
If the `CorsPolicy` has a `PathMapping` property, `CorsService` will be able to find a policy by using an `Origin` header value and the request path. Therefore, a user can apply CORS policies for different paths more convenient way.

Modifications:
- Added `boolean hasPathPatternOnly()` method to `PathMapping` interface, which describes whether the mapping has only path patterns as its condition.
- Added `pathMapping()` methods to `AbstractCorsPolicyBuilder` and `CorsServiceBuilder`.
  - Note that only the `PathMapping` that `PathMapping.hasPathPatternOnly()` returns `true` can be allowed.
- `CorsServiceBuilder` no longer keeps `ChainedCorsPolicyBuilder` list. It will be immediately built when its `and()` method is called. It's because the order of policies became important. The first configured path mapping must be evaluated first.
- Added `pathPatterns` property to `@CorsDecorator`.
- Updated `server-cors.rst`.

Result:
- A user can apply a CORS policy to the specific paths.
- Closes #1698.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
@@ -131,7 +131,9 @@ abstract class AbstractCorsPolicyBuilder<B extends AbstractCorsPolicyBuilder> {
      *
      * @param pathMapping the {@link PathMapping} that this policy is supposed to be applied to
      * @return {@code this} to support method chaining.
-     * @throws IllegalArgumentException if the {@link PathMapping} has conditions beyond the path pattern
+     * @throws IllegalArgumentException if the {@link PathMapping} has conditions beyond the path pattern,
+     *                                  i.e. the {@link PathMapping} created by
+     *                                  {@link PathMapping#withHttpHeaderInfo(Set, List, List)}
      */
     public B pathMapping(PathMapping pathMapping) {
         requireNonNull(pathMapping, "pathMapping");
--- a/core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
@@ -86,9 +86,8 @@ abstract class AbstractCorsPolicyBuilder<B extends AbstractCorsPolicyBuilder> {
     }
 
     B setConfig(CorsDecorator corsDecorator) {
-        if (corsDecorator.pathPatterns().length > 0) {
-            Arrays.stream(corsDecorator.pathPatterns()).forEach(this::pathMapping);
-        }
+        Arrays.stream(corsDecorator.pathPatterns()).forEach(this::pathMapping);
+
         if (corsDecorator.credentialsAllowed()) {
             allowCredentials();
         }
--- a/core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
@@ -120,6 +120,7 @@ abstract class AbstractCorsPolicyBuilder<B extends AbstractCorsPolicyBuilder> {
      *
      * @param pathPattern the path pattern that this policy is supposed to be applied to
      * @return {@code this} to support method chaining.
+     * @throws IllegalArgumentException if the path pattern is not valid
      */
     public B pathMapping(String pathPattern) {
         return pathMapping(PathMapping.of(pathPattern));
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/linecorp/armeria/server/cors/CorsConfig.java
```diff
@@ -48,25 +50,23 @@
     private final boolean enabled;
     private final boolean anyOriginSupported;
     private final boolean shortCircuit;
-    private final Set<CorsPolicy> policies;
+    private final List<CorsPolicy> policies;
```
hyangtack: Changed `Set` to `List` in order to visit policies in the configured order. :-)

adriancole:
thank you

site/src/sphinx/server-cors.rst
```diff
@@ -98,6 +98,29 @@ You can also directly add a :api:`CorsPolicy` created by a :api:`CorsPolicyBuild
...
+                              // or '/message/internal/web/api/'.
+                              .pathMapping(PathMapping.of("prefix:/message/web/api/"))
+                              // A shortcut of 'PathMapping.of(...)'.
+                              .pathMapping("prefix:/message/internal/web/api/")
+                              .allowRequestMethods(HttpMethod.POST, HttpMethod.GET)
+                              .newDecorator()));
+
+.. note::
+
+    Please refer to the :api:`PathMapping` in order to learn how to specify a path pattern you interest.
```
trustin: 'you interest' could be removed?

trustin:
> I presume that path mapping is something we can refine to method expressions?

What do you mean by 'method expressions'?

adriancole:
I mean for example POST against our api is not something the UI should do, but javascript tracers do it.

So, GET /**, but not POST

adriancole:
incidentally, nevermind on this request as I think we didn't do anything specific about POST vs GET.. memory was bad! https://github.com/openzipkin/zipkin/commit/65a2db7672f9aedf219e09eba325760e5c8dc9b9

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -82,6 +86,9 @@ B self() {
     }
 
     B setConfig(CorsDecorator corsDecorator) {
+        if (corsDecorator.pathPatterns().length > 0) {
+            Arrays.stream(corsDecorator.pathPatterns()).forEach(this::pathMapping);
```
anuraaga: Think the if and stream can be combined into a single for loop

adriancole:
ps I just had to extend our CORS mappings to include "/" -> allow  POST with http header "X-GRPC-WEB" for.. grpc web

trustin:
Great job, @hyangtack. Thanks for reviewing, reviewers.

adriancole:
thank you contributors!

adriancole:
PS I think for most concerns I don't need to worry about differentiation on GET as these requests are [simple](https://medium.com/@f2004392/cors-preflight-request-options-9d05b9248e5a) so won't initiate CORS flows.

We will ideally use this to compose two different CORS aspects: normal json POST (in one configuration) and grpc POST (in another)

I'm not sure if this feature provides for composition, but if it doesn't I'll open another ticket.

hyangtack:
@adriancole I'm not sure I understood correctly, you can now apply different CORS policies with one CorsService instance based on the request path. For example:

- `prefix:/some_json_api`: allowing `POST` method and a custom header `x-json-blah`
- `regex:\/com.linecorp.service.grpc.[^/]+\/.+`: allowing `POST` method and a custom header `x-cors-blah`

If we need to use the request headers in the preflight request, such as Accept, User-Agent, etc., for deciding a CORS policy, we need to make more changes.

adriancole:
sorry I mean cors policy specified independently from two armeria service configurators.. if it will compose. maybe I just need to try it.

trustin:
> sorry I mean cors policy specified independently from two armeria service configurators.. if it will compose. maybe I just need to try it.

IIUC this is not possible. You need to decorate only once. Do you think it will be useful to allow decorating more than once although it's less optimal?

adriancole:
> sorry I mean cors policy specified independently from two armeria service
> configurators.. if it will compose. maybe I just need to try it.
>
> IIUC this is not possible. You need to decorate only once. Do you think it
> will be useful to allow decorating more than once although it's less
> optimal?
>

for us in zipkin it is ok as we can fake it. main thing is it feels a break
in modularity for one place to need special knowledge of the combined cors
policy of otherwise individual components

PR #1706, opened 2019-04-09
Don't use address for service name

tacigar:
In most cases, address can not be used in place of service name.
The namespace of service name is polluted.
If `remoteEndpoint.serviceName` is not set, it should not be set to span.

Related PR: #835 #1235

### DIFF
--- a/zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
+++ b/zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
@@ -25,8 +25,6 @@ import java.util.function.Function;
 
 import javax.annotation.Nullable;
 
-import com.google.common.base.Strings;
-
 import com.linecorp.armeria.client.Client;
 import com.linecorp.armeria.client.ClientRequestContext;
 import com.linecorp.armeria.client.SimpleDecoratingClient;
@@ -125,7 +123,6 @@ public class HttpTracingClient extends SimpleDecoratingClient<HttpRequest, HttpR
     }
 
     private void setRemoteEndpoint(Span span, RequestLog log) {
-
         final SocketAddress remoteAddress = log.context().remoteAddress();
         final InetAddress address;
         final int port;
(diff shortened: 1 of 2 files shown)


zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -137,22 +134,7 @@ private void setRemoteEndpoint(Span span, RequestLog log) {
             address = null;
             port = 0;
         }
-
-        final String remoteServiceName;
-        if (this.remoteServiceName != null) {
-            remoteServiceName = this.remoteServiceName;
-        } else {
-            final String authority = log.requestHeaders().authority();
```
anuraaga: What do you think of still following back to authority?

trustin: Yes, authority should be OK in this case. What do you think, @tacigar ?

tacigar: Ummm...
I'm not sure but I think host name (& port) should not be used for service name too...
@adriancole What do you think?

adriancole: this conversation has happened in the past. there's basically two ways about it.

* some feel they have clean hosts (probably port part of authority is still dubious)
* some have not clean hosts

usually in zipkin don't default service names to dns values as life usually isn't clean. However, this is inconvenient to places that have clean hosts (like cluster scoped, not instance scoped).

this boils down to what causes the least damage vs highest convenience. I personally prefer to keep defaults towards causing least damage as it is harder to clean up later vs ask someone to opt-in.

adriancole:
ps this change once merged will move some of the pollution problem at least off the primary service list https://github.com/openzipkin/zipkin/pull/2484

that doesn't mean don't do this :D

zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
tacigar: Thank you for your valuable comments 🙇

huydx:
@trustin @tacigar is LINE member, does he still need to sign ICLA?

syleeeee:
@huydx , @tacigar Yes, Armeria belongs to LINE org, but still is an open source. As you can see ICLA doesn't require your employment status. So this PR will be merged after signing. 
Thank you for your first pull request. 👍

tacigar:
Done 🙇

trustin:
Thanks @tacigar and reviewers. :-)

tacigar:
Thank you~ :)

PR #1707, opened 2019-04-10
Add `boolean shutdownOnStop` parameter to `ServerBuilder.blockingTaskExecutor()`

### LINKED ISSUE #1685: Add `boolean shutdownOnStop` parameter to `ServerBuilder.blockingTaskExecutor()`
When a user specifies a custom blocking task executor using `ServerBuilder.blockingTaskExecutor(Executor)`, we have to know if a user wants the executor to shut down when the server stops. If so (i.e. specified `true`), we could wait until the executor is terminated before the stop process is complete.

We need to deprecate the old `ServerBuilder.blockingTaskExecutor(Executor)` method in favor of `blockingTaskExecutor(Executor, boolean)`.

geminiKim:
Hi :)
I added `ServerBuilder.blockingTaskExecutor(Executor, boolean[shutdownOnStop])`, but I don't know how to using `shutdownOnStop`.

Modifications:
- Deprecate `ServerBuilder.blockingTaskExecutor(Executor)`
- Add `ServerBuilder.blockingTaskExecutor(Executor, boolean[shutdownOnStop])`

Result:
- Fix #1685

\- This is my first PR at Armenia that might be insufficient, Please advise the insufficient part.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Server.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Server.java
@@ -469,12 +469,12 @@ public final class Server implements AutoCloseable {
                     while (!executor.isTerminated()) {
                         try {
                             executor.awaitTermination(1, TimeUnit.DAYS);
-                        } catch (InterruptedException e) {
-                            Thread.currentThread().interrupt();
+                        } catch (InterruptedException ignore) {
+                            // Do nothing.
                         }
                     }
                 } catch (Exception e) {
-                    logger.warn("Failed to shutdown the {}:", executor, e);
+                    logger.warn("Failed to shutdown the blockingTaskExecutor: {}", executor, e);
                 }
             }
 
--- a/core/src/main/java/com/linecorp/armeria/server/Server.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Server.java
@@ -457,11 +457,11 @@ public final class Server implements AutoCloseable {
         private void finishDoStop(CompletableFuture<Void> future) {
             if (config.shutdownBlockingTaskExecutorOnStop()) {
                 final ExecutorService executor;
-                if (config.blockingTaskExecutor() instanceof InterminableExecutorService) {
-                    executor =
-                            ((InterminableExecutorService) config.blockingTaskExecutor()).getExecutorService();
+                final ExecutorService blockingTaskExecutor = config.blockingTaskExecutor();
+                if (blockingTaskExecutor instanceof InterminableExecutorService) {
+                    executor = ((InterminableExecutorService) blockingTaskExecutor).getExecutorService();
                 } else {
-                    executor = config.blockingTaskExecutor();
+                    executor = blockingTaskExecutor;
                 }
 
                 try {
--- a/core/src/main/java/com/linecorp/armeria/server/Server.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Server.java
@@ -456,17 +456,25 @@ public final class Server implements AutoCloseable {
 
         private void finishDoStop(CompletableFuture<Void> future) {
             if (config.shutdownBlockingTaskExecutorOnStop()) {
+                final ExecutorService executor;
                 if (config.blockingTaskExecutor() instanceof InterminableExecutorService) {
-                    final ExecutorService service =
+                    executor =
                             ((InterminableExecutorService) config.blockingTaskExecutor()).getExecutorService();
-                    try {
-                        service.shutdown();
-                        while (!service.isTerminated()) {
-                            service.awaitTermination(500, TimeUnit.MILLISECONDS);
+                } else {
+                    executor = config.blockingTaskExecutor();
+                }
+
+                try {
+                    executor.shutdown();
+                    while (!executor.isTerminated()) {
+                        try {
+                            executor.awaitTermination(1, TimeUnit.DAYS);
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
                         }
-                    } catch (Exception e) {
-                        logger.warn("Failed to shutdown the {}:", service.getClass().getSimpleName(), e);
                     }
+                } catch (Exception e) {
+                    logger.warn("Failed to shutdown the {}:", executor, e);
                 }
             }
 
--- a/core/src/main/java/com/linecorp/armeria/server/Server.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Server.java
@@ -455,8 +455,21 @@ public final class Server implements AutoCloseable {
         }
 
         private void finishDoStop(CompletableFuture<Void> future) {
-            // TODO(trustin): Add shutdownBlockingTaskExecutorOnStop
-            // TODO(trustin): Count the pending blocking tasks and wait until it goes zero.
+            if (config.shutdownBlockingTaskExecutorOnStop()) {
+                if (config.blockingTaskExecutor() instanceof InterminableExecutorService) {
+                    final ExecutorService service =
+                            ((InterminableExecutorService) config.blockingTaskExecutor()).getExecutorService();
+                    try {
+                        service.shutdown();
+                        while (!service.isTerminated()) {
+                            service.awaitTermination(500, TimeUnit.MILLISECONDS);
+                        }
+                    } catch (Exception e) {
+                        logger.warn("Failed to shutdown the {}:", service.getClass().getSimpleName(), e);
+                    }
+                }
+            }
+
             if (!config.shutdownAccessLogWriterOnStop()) {
                 future.complete(null);
                 return;
--- a/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
@@ -652,7 +652,7 @@ public final class ServerBuilder {
      */
     @Deprecated
     public ServerBuilder blockingTaskExecutor(Executor blockingTaskExecutor) {
-        this.blockingTaskExecutor = requireNonNull(blockingTaskExecutor, "blockingTaskExecutor");
+        blockingTaskExecutor(blockingTaskExecutor, false);
         return this;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
@@ -652,8 +652,7 @@ public final class ServerBuilder {
      */
     @Deprecated
     public ServerBuilder blockingTaskExecutor(Executor blockingTaskExecutor) {
-        blockingTaskExecutor(blockingTaskExecutor, false);
-        return this;
+        return blockingTaskExecutor(blockingTaskExecutor, false);
     }
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
@@ -177,6 +177,7 @@ public final class ServerBuilder {
     private Duration gracefulShutdownQuietPeriod = DEFAULT_GRACEFUL_SHUTDOWN_QUIET_PERIOD;
     private Duration gracefulShutdownTimeout = DEFAULT_GRACEFUL_SHUTDOWN_TIMEOUT;
     private Executor blockingTaskExecutor = CommonPools.blockingTaskExecutor();
+    private boolean shutdownBlockingTaskExecutorOnStop;
     private MeterRegistry meterRegistry = Metrics.globalRegistry;
     private String serviceLoggerPrefix = DEFAULT_SERVICE_LOGGER_PREFIX;
     private AccessLogWriter accessLogWriter = AccessLogWriter.disabled();
@@ -645,9 +646,25 @@ public final class ServerBuilder {
     /**
      * Sets the {@link Executor} dedicated to the execution of blocking tasks or invocations.
      * If not set, {@linkplain CommonPools#blockingTaskExecutor() the common pool} is used.
+     *
+     * @deprecated Use {@link #blockingTaskExecutor(Executor, boolean)}.
+     *
      */
+    @Deprecated
     public ServerBuilder blockingTaskExecutor(Executor blockingTaskExecutor) {
         this.blockingTaskExecutor = requireNonNull(blockingTaskExecutor, "blockingTaskExecutor");
+        return this;
+    }
+
+    /**
+     * Sets the {@link Executor} dedicated to the execution of blocking tasks or invocations.
+     * If not set, {@linkplain CommonPools#blockingTaskExecutor() the common pool} is used.
+     *
+     * @param shutdownOnStop whether to shut down the {@link Executor} when the {@link Server} stops
+     */
+    public ServerBuilder blockingTaskExecutor(Executor blockingTaskExecutor, boolean shutdownOnStop) {
+        this.blockingTaskExecutor = requireNonNull(blockingTaskExecutor, "blockingTaskExecutor");
+        shutdownBlockingTaskExecutorOnStop = shutdownOnStop;
         return this;
     }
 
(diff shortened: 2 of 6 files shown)


hyangtack:
Thank you for your first contribution to Armeria!  👏 

As described in #1685, the new parameter `shutdownBlockingTaskExecutorOnStop` you added will be used to specify whether a user wants the blocking task executor to shut down when the server stop.

So, you need to fix [here](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/Server.java#L458). Also, you can refer to [here](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/Server.java#L420) how `shutdownWorkerGroupOnStop` does, which is a similar work what you have to do.

Please leave a message if you have a further question! 😃

geminiKim:
@hyangtack I finished work, but not sure if it is correct. Please give feedback after the review. 
Thank you :)

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -455,8 +455,16 @@ private void doStop(CompletableFuture<Void> future,
         }
 
         private void finishDoStop(CompletableFuture<Void> future) {
-            // TODO(trustin): Add shutdownBlockingTaskExecutorOnStop
-            // TODO(trustin): Count the pending blocking tasks and wait until it goes zero.
+            if (config.shutdownBlockingTaskExecutorOnStop()) {
+                while (gracefulShutdownSupport.pendingResponses() != 0) {
```
hyangtack: I think we need to check whether the `blockingTaskExecutor` has a pending job from here and then shutdown it. 

Checking `pendingResponses` seems okay for the requests received by the server, but we need to consider another jobs which are not belonging to the request/response. 
For example, assume that there's a service which:
- submits a new background task to the `blockingTaskExecutor` when a request is received 
- sends a response immediately after the task is submitted

In this case, there's no `pendingResponses` but the `blockingTaskExecutor` may be executing some job.
So, I think you need to update this `if` block. Also, would you try to add a test for this new feature? You can refer to the following file for writing a test.
https://github.com/line/armeria/blob/master/core/src/test/java/com/linecorp/armeria/server/ServerTest.java

Does it make sense? Please let us know if you have a further question. 😃

trustin:
@geminiKim Gentle ping. Any progress? 😉

core/src/main/java/com/linecorp/armeria/server/Server.java
geminiKim: You are all right, but I can not find the method of the interface for that task.

One idea is to graceful shutdown using `shutdown`, `shutdownNow` and `awaitTermination` of `ExecutorService`,  but it is constrained by this code:
https://github.com/line/armeria/blob/738b4f5c683b07fd620d7a8a5026aeb6a410fb36/core/src/main/java/com/linecorp/armeria/server/ServerConfig.java#L156

`blockingTaskExecutor` are wrapped with `InterminableExecutorService` or `ExecutorBasedExecutorService` at `ServerConfig`, and wrapping class implement this code:
```
@Override
public void shutdown() {
    throw new UnsupportedOperationException();
}

@Override
public List<Runnable> shutdownNow() {
    throw new UnsupportedOperationException();
}
```
How can i check pending jobs?

Adding new features is difficult for me, because i am not familiar with Armenia yet. but I can keep trying wtih thanks to your advice. 
Thank you :D

trustin: That's a good question. How about this:

- Add a package-local method that returns the original executor to `InterminableExecutorService`.
  - If `instanceof InterminableExecutorService`, unwrap it to use the original `ExecutorService`.
- As a safety measure, catch the `UnsupportedOperationException` raised by `shutdown()` and `awaitTermination()` and log the exception.

Makes sense?

geminiKim:
I'm added some code (Thanks to trustin for giving a detailed guideline) and cleanup commit-line.

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -455,8 +455,21 @@ private void doStop(CompletableFuture<Void> future,
         }
 
         private void finishDoStop(CompletableFuture<Void> future) {
-            // TODO(trustin): Add shutdownBlockingTaskExecutorOnStop
-            // TODO(trustin): Count the pending blocking tasks and wait until it goes zero.
+            if (config.shutdownBlockingTaskExecutorOnStop()) {
+                if (config.blockingTaskExecutor() instanceof InterminableExecutorService) {
+                    final ExecutorService service =
+                            ((InterminableExecutorService) config.blockingTaskExecutor()).getExecutorService();
```
trustin: This has to be:

```java
final ExecutorService executor;
if (config.blockingTaskExecutor() instanceof InterminableExecutorService) {
    executor = config.blockingTaskExecutor().getExecutorService();
} else {
    executor = config.blockingTaskExecutor();
}

try {
    service.shutdown();
    ...
```

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -455,8 +455,21 @@ private void doStop(CompletableFuture<Void> future,
...
+                    try {
+                        service.shutdown();
+                        while (!service.isTerminated()) {
+                            service.awaitTermination(500, TimeUnit.MILLISECONDS);
```
trustin: - You need to catch `InterruptedException` and swallow it because server shutdown should not be interrupted.
- Perhaps we can wait for more than 500ms? No need to wake up at all until termination. How about 1 day?

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -455,8 +455,21 @@ private void doStop(CompletableFuture<Void> future,
...
+                        }
+                    } catch (Exception e) {
+                        logger.warn("Failed to shutdown the {}:", service.getClass().getSimpleName(), e);
```
trustin: For anonymous class, `getSimpleName()` will not give useful information. How about just passing `service`?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -645,12 +646,28 @@ public ServerBuilder gracefulShutdownTimeout(Duration quietPeriod, Duration time
     /**
      * Sets the {@link Executor} dedicated to the execution of blocking tasks or invocations.
      * If not set, {@linkplain CommonPools#blockingTaskExecutor() the common pool} is used.
+     *
+     * @deprecated Use {@link #blockingTaskExecutor(Executor, boolean)}.
+     *
      */
+    @Deprecated
     public ServerBuilder blockingTaskExecutor(Executor blockingTaskExecutor) {
         this.blockingTaskExecutor = requireNonNull(blockingTaskExecutor, "blockingTaskExecutor");
```
trustin: Could delegate to `blockingTaskExecutor(Executor, boolean)`.

core/src/test/java/com/linecorp/armeria/server/ServerTest.java
```diff
@@ -358,6 +358,60 @@ public void customStartStopExecutor() {
         threads.forEach(t -> assertThat(t.getName()).startsWith(prefix));
     }
 
+    @Test
+    public void gracefulShutdownBlockingTaskExecutor() {
+        final Queue<Thread> threads = new LinkedTransferQueue<>();
```
trustin: This seems unused.

core/src/test/java/com/linecorp/armeria/server/ServerTest.java
```diff
@@ -358,6 +358,60 @@ public void customStartStopExecutor() {
...
+
+        threads.add(server.stop().thenApply(unused -> Thread.currentThread()).join());
+
+        assertThat(server.config().blockingTaskExecutor().isShutdown()).isTrue();
+        assertThat(server.config().blockingTaskExecutor().isTerminated()).isTrue();
+    }
+
+    @Test
+    public void notGracefulShutdownBlockingTaskExecutor() {
+        final Queue<Thread> threads = new LinkedTransferQueue<>();
```
trustin: This seems unused.

core/src/test/java/com/linecorp/armeria/server/ServerTest.java
```diff
@@ -358,6 +358,60 @@ public void customStartStopExecutor() {
...
+                Thread.sleep(processDelayMillis * 2);
+            } catch (InterruptedException ignored) {
+                // Ignored
+            }
+        });
+
+        threads.add(server.stop().thenApply(unused -> Thread.currentThread()).join());
+
+        assertThat(server.config().blockingTaskExecutor().isShutdown()).isFalse();
+        assertThat(server.config().blockingTaskExecutor().isTerminated()).isFalse();
```
trustin: At the end of the test, you need to shut down the executor.

thrift/src/test/java/com/linecorp/armeria/server/thrift/ThriftServiceTest.java
```diff
@@ -636,7 +636,8 @@ private static void invoke0(THttpService service, HttpData content,
                 ServiceRequestContextBuilder.of(req)
                                             .eventLoop(eventLoop.get())
                                             .serverConfigurator(builder -> {
-                                                builder.blockingTaskExecutor(ImmediateEventExecutor.INSTANCE);
+                                                builder.blockingTaskExecutor(ImmediateEventExecutor.INSTANCE,
+                                                                             true);
```
trustin: `false` is probably better, although it will not make much difference. `ImmediateEventExecutor` never terminates.

geminiKim:
Updated the code on comments

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -456,17 +456,25 @@ private void doStop(CompletableFuture<Void> future,
...
+                    executor = config.blockingTaskExecutor();
+                }
+
+                try {
+                    executor.shutdown();
+                    while (!executor.isTerminated()) {
+                        try {
+                            executor.awaitTermination(1, TimeUnit.DAYS);
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
```
trustin: This will make the loop very busy because the next `awaitTermination` will wake up immediately and throw `InterruptedException`. Although not perfect, I'd recommend just swallowing it.

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -456,17 +456,25 @@ private void doStop(CompletableFuture<Void> future,
...
                         }
-                    } catch (Exception e) {
-                        logger.warn("Failed to shutdown the {}:", service.getClass().getSimpleName(), e);
                     }
+                } catch (Exception e) {
+                    logger.warn("Failed to shutdown the {}:", executor, e);
```
trustin: `"Failed to shutdown the blockingTaskExecutor: {}"`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -652,7 +652,7 @@ public ServerBuilder gracefulShutdownTimeout(Duration quietPeriod, Duration time
      */
     @Deprecated
     public ServerBuilder blockingTaskExecutor(Executor blockingTaskExecutor) {
-        this.blockingTaskExecutor = requireNonNull(blockingTaskExecutor, "blockingTaskExecutor");
+        blockingTaskExecutor(blockingTaskExecutor, false);
         return this;
```
trustin: Could be simplified to:

    return blockingTaskExecutor(blockingTaskExecutor, false);

geminiKim:
Updated the code on comments 😄

core/src/test/java/com/linecorp/armeria/server/ServerTest.java
```diff
@@ -358,6 +360,57 @@ public void customStartStopExecutor() {
...
+    @Test
+    public void gracefulShutdownBlockingTaskExecutor() {
+        final ExecutorService executor = Executors.newSingleThreadExecutor();
+
+        final Server server = new ServerBuilder()
+                .blockingTaskExecutor(executor, true)
+                .service("/", (ctx, req) -> HttpResponse.of(200))
+                .build();
+
+        server.start().thenApply(unused -> Thread.currentThread()).join();
```
hyangtack: nit: `.thenApply(unused -> Thread.currentThread())` looks not necessary. Would you remove it?

geminiKim: That is my mistake 😮, Thank you for thoughtful comments

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -177,6 +177,7 @@
     private Duration gracefulShutdownQuietPeriod = DEFAULT_GRACEFUL_SHUTDOWN_QUIET_PERIOD;
     private Duration gracefulShutdownTimeout = DEFAULT_GRACEFUL_SHUTDOWN_TIMEOUT;
     private Executor blockingTaskExecutor = CommonPools.blockingTaskExecutor();
+    private boolean shutdownBlockingTaskExecutorOnStop;
```
minwoox: I wonder if we need to add this to `Flags`, WDYT? /cc @trustin

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -455,8 +455,29 @@ private void doStop(CompletableFuture<Void> future,
         }
 
         private void finishDoStop(CompletableFuture<Void> future) {
-            // TODO(trustin): Add shutdownBlockingTaskExecutorOnStop
-            // TODO(trustin): Count the pending blocking tasks and wait until it goes zero.
+            if (config.shutdownBlockingTaskExecutorOnStop()) {
+                final ExecutorService executor;
+                if (config.blockingTaskExecutor() instanceof InterminableExecutorService) {
```
minwoox: nit: You could extract `config.blockingTaskExecutor()` as a variable and use it below.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -645,9 +646,24 @@ public ServerBuilder gracefulShutdownTimeout(Duration quietPeriod, Duration time
     /**
      * Sets the {@link Executor} dedicated to the execution of blocking tasks or invocations.
      * If not set, {@linkplain CommonPools#blockingTaskExecutor() the common pool} is used.
+     *
+     * @deprecated Use {@link #blockingTaskExecutor(Executor, boolean)}.
+     *
      */
+    @Deprecated
```
minwoox: I'm wondering if it's a good idea to deprecate this. There should be some users who do not want to care about `shutdownOnStop`. How about turning this back from deprecated? /cc @trustin

trustin: Just like other similar builder methods:

- `ServerBuilder.workerGroup()`
- `ServerBuilder.accessLogger()`

I think we do not really know which behavior a user prefers when it comes down to life cycle. People who do not care could make an informed decision of specifying `false` in my opinion.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: I'm not sure I like the idea. See my comment at: https://github.com/line/armeria/pull/1707#discussion_r277188573

core/src/main/java/com/linecorp/armeria/server/Server.java
trustin: Indeed. @geminiKim, would you mind cleaning this up a little bit?

trustin:
Thanks a lot, @geminiKim! Hoping to see you in the next pull request. :wink: We also look forward to your feature/improvement requests and bug reports!

geminiKim:
Thank you @trustin and all reviewer.
I think I need to know more about Armeria for the good contribution.
This PR was a lot of shortcomings this time, but I will do better next time, Thanks a lot again😃

PR #1708, opened 2019-04-10
Fix JSON parse on header in docs-client

### LINKED ISSUE #1622: Replace quotes from OSX notes app with quotation mark
When I used armeria doc service, it supports to utilize REST API.
It only understands quotation mark " (U+0022) as string quotation. However, MAC OS X has different quotations like left double quotation “ (U+201C) if I make the example in notes app.

So, If there is no constraint rule, it is very helpful to support that armeria can automatically replace those quotation marks as ascii quotation mark.

JunoJunho:
Motivation:
When I try to use docs-client to simply test my API, I found that there
is a bug validation on header is not properly working.
I found that there is an error in onSubmit method. Armeria invokes
JSON.parse method to validate and parse user input header, but it does
not passed to conditional branch.

Modification:
I correct parsedHeader in checking branch onSubmit method.

------
I think this PR can resolve #1622 issue.

### DIFF
--- a/docs-client/src/containers/MethodPage/index.tsx
+++ b/docs-client/src/containers/MethodPage/index.tsx
@@ -513,7 +513,7 @@ export default class MethodPage extends React.PureComponent<Props, State> {
     if (headers) {
       try {
         const parsedHeaders = JSON.parse(headers);
-        if (typeof params !== 'object') {
+        if (typeof parsedHeaders !== 'object') {
           this.setState({
             debugResponse: `HTTP headers must be a JSON object.\nYou entered: ${typeof parsedHeaders}`,
           });


JunoJunho:
While I sending this PR, I know that, had code been fixed?

minwoox:
Thank you for this.
But, I think I have fixed this bug in #1667 so from 0.83.0 the bug seems gone. Could you check it out with the latest version, please?

JunoJunho:
@minwoox It seems fixed, but I will.

minwoox:
Thanks a lot @JunoJunho and we really appreciate your enthusiasm for finding and fixing a bug.

trustin:
@JunoJunho Could you fix the conflict?

trustin:
As @minwoox said, this issue has been fixed already, so let me close this. Thanks a lot for trying to help us, @JunoJunho. Let us look forward to your next pull request. :wink:

PR #1712, opened 2019-04-12
Clean up RetryStrategy and RetryStrategyWithContent Javadoc

trustin:
- Updated the Javadoc of `RetryStrategy` and `RetryStrategyWithContent`
  so that it is more user-centric.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
@@ -24,7 +24,6 @@ import java.util.function.BiFunction;
 import javax.annotation.Nullable;
 
 import com.linecorp.armeria.client.ClientRequestContext;
-import com.linecorp.armeria.client.ResponseTimeoutException;
 import com.linecorp.armeria.client.UnprocessedRequestException;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpStatus;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
```diff
@@ -136,11 +133,6 @@ static RetryStrategy onStatus(
      * @param ctx the {@link ClientRequestContext} of this request
      * @param cause the {@link Throwable} which is raised while sending a request. {@code null} it there's no
      *              exception.
-     *
-     * @see RetryingClientBuilder#responseTimeoutMillisForEachAttempt(long)
-     *
-     * @see <a href="https://line.github.io/armeria/advanced-retry.html#per-attempt-timeout">Per-attempt
-     *      timeout</a>
      */
     CompletionStage<Backoff> shouldRetry(ClientRequestContext ctx, @Nullable Throwable cause);
```
minwoox: So we don't have mention about `Exceptions.peel()`?

trustin: No, we don't. I found all the examples in `client-retry.rst` also use the exceptions as they are. I think it's really up to a user to use `Exceptions.peel()` or not.

trustin:
image

core/src/main/java/com/linecorp/armeria/client/retry/RetryStrategy.java
trustin: I added a tip about using `Exceptions.peel()` in `client-retry.rst` instead.

site/src/sphinx/client-retry.rst
```diff
@@ -152,32 +158,63 @@ using :api:`RetryingHttpClientBuilder`:
...
+                                                    @Nullable Throwable cause) {
+            if (cause != null) {
+                if (cause instanceof ResponseTimeoutException ||
+                    cause instanceof UnprocessedRequestException) {
+                    // The response timed out or the request has not been handled
+                    // by the server.
+                    return CompletableFuture.completedFuture(backoff);
+                }
+
+                Throwable peeled = Exceptions.peel(cause);
```
minwoox: How about moving this right under the `if (cause != null) {`? or Might we remove the `if` condition which checks the exception instance because it's already described in the above example?

PR #1714, opened 2019-04-14
Allow SettableHealthChecker to have chained invocation.

anuraaga:
Currently, it can sometimes be inconvenient that a `SettableHealthChecker` can't have `setHealthy` called fluently. This is mainly because the default is `false` though - it might be good to consider whether the default should be `true`, in which case this change becomes less useful.


hyangtack:
> This is mainly because the default is false though - it might be good to consider whether the default should be true, in which case this change becomes less useful.
@anuraaga How about adding a constructor which can set the initial `isHealthy` value as well?

trustin:
How about changing the default value to `true`, adding a constructor with an initial value and changing the return type of `setHealthy()`?

trustin:
@anuraaga Gentle ping. Hoping to release 0.84.0 today. :wink:

trustin:
Could you fix the test failures? :wink:

anuraaga:
Doh! Sorry will fix em up tomorrow

anuraaga:
Travis seems to be having trouble downloading openjdk

trustin:
Thanks, @anuraaga !

PR #1716, opened 2019-04-15
Add HTTP content encoding property to ArmeriaSettings

hyangtack:
Motivation:
Currently, Spring framework provides `server.compression` property to configure HTTP content encoding,
but `ArmeriaSettings` doesn't provide it. So a user cannot configure it without fixing their code.

Modifications:
- Added `Compression` class to `ArmeriaSettings`, which is similar to the Spring's one.
  - We could not reuse the Spring's class because we need to support Spring Boot 1.x and 2.x.
- Moved `ArmeriaReactiveWebServerFactory#contentEncodingDecorator` to `ArmeriaConfigurationUtil` for sharing the method with `spring-boot-autoconfigure`.

Result:
- A user can configure HTTP content encoding with a Spring property file.

### DIFF
--- a/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java
+++ b/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java
@@ -16,6 +16,8 @@
 package com.linecorp.armeria.internal.spring;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.requireNonNull;
 
 import java.io.IOException;
@@ -28,6 +30,7 @@ import java.net.UnknownHostException;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.HashMap;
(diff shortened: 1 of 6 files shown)


spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
```diff
@@ -150,6 +155,93 @@ public Port setProtocol(SessionProtocol protocol) {
         }
     }
 
+    /**
+     * Configurations for the HTTP content encoding.
+     */
+    public static class Compression {
```
hyangtack: Picked `Compression` because this is similar property to the Spring's `Compression` class. Please let me know if you think another name is proper. :-)

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java
```diff
@@ -367,10 +389,10 @@ public static void configureTls(ServerBuilder sb, Ssl ssl,
...
                     .trustManager(getTrustManagerFactory(ssl, trustStoreSupplier));
             final SslProvider sslProvider = ssl.getProvider() == null ?
-                    Flags.useOpenSsl() ?
-                            SslProvider.OPENSSL
-                            : SslProvider.JDK
-                    : ssl.getProvider();
+                                            Flags.useOpenSsl() ?
+                                            SslProvider.OPENSSL
+                                                               : SslProvider.JDK
+                                                                      : ssl.getProvider();
```
minwoox: How about using `if`? it's hard for me to recognize when there are two ternaries.
```java
final SslProvider sslProvider;
final SslProvider provider = ssl.getProvider();
if (provider != null) {
    sslProvider = provider;
} else {
    sslProvider = Flags.useOpenSsl() ? SslProvider.OPENSSL : SslProvider.JDK;
}
```

trustin:
Thanks, @hyangtack !

Issue #1703, opened 2019-04-08
Allow unary grpc without google's library


adriancole:
In zipkin server, we are often sensitive to version trees, and google's grpc library has some concerns around this including protobuf, guava, and recently things like opencensus also.

I'd love to be able to directly support two endpoints with no google grpc library dependency:
* grpc inbound unary call of storing a list of spans https://github.com/openzipkin/zipkin/pull/2328
* grpc outbound call to stackdriver https://github.com/openzipkin/zipkin-gcp/blob/master/storage-stackdriver/src/main/java/zipkin2/storage/stackdriver/StackdriverSpanConsumer.java

In the future, we can consider things like streaming etc, but it would be sufficient to cover the above, even only the inbound call.

The ideal integration would be either a hand-coded or armeria annotation markup grpc listener. Possibly this could be implemented by a new dependency "armeria-grpc-protocol" which doesn't include any google grpc deps and can allow you to receive a byte buffer of the upload and process as you see fit.

trustin:
Do you expect this not to depend on `protobuf`?

adriancole:
I'd love to not be able to, but I understand the method metadata may not be straightforward. For example, I personally only need 1 endpoint so if have a way to declare the method metadata directly would be able to opt-out of protobuf library dep.

trustin:
> doesn't include any google grpc deps and can allow you to receive a byte buffer of the upload and process as you see fit.

I'm not sure I understood correctly, but isn't it already possible? What is different from handling an HTTP request stream as it is?

adriancole:
@trustin sorry.. to be more specific, there are a number of types with dependencies on grpc library in armeria-grpc. Even if we excluded the grpc library dependency, there's no signature of `GrpcServiceBuilder` we could use (they accept io.grpc.BindableService). Perhaps a learning test could help if there's a way we can achieve the goal that I'm not thinking of..

anuraaga:
Specifically, the idea is that logic related to headers, framer, and deframer, could be extracted for advanced users that want to use the gRPC wire protocol directly without the gRPC stubs. I think this isn't too invasive though depending on how far we go with it can possibly lead to some overkill too.

trustin:
I see. Sounds like a good idea, although I'm not sure if we need to publish a dedicated artifact for this. Perhaps we could make it part of `armeria-grpc` and let an advanced user exclude `grpc-*` dependencies?

adriancole:
[...]

PR #1720, opened 2019-04-17
Extract gRPC wire protocol core classes to a protocol package.

anuraaga:
For #1703 

This is the first of a series of PRs to create a package that allows users to have simple gRPC functionality without dependencies on the gRPC It is almost purely moving classes around, with some bridge classes between armeria-specific classes for `Status` and compression and the gRPC ones for normal users.

After this PR, I will add some simple helper methods for doing unary server / client invocation based on these classes for users like @adriancole Want to make sure this isn't on the entirely wrong path in the first place for now.


grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaStatusException.java
```diff
@@ -0,0 +1,45 @@
...
+     * Construct a {@link ArmeriaStatusException} for the given gRPC status code and message.
+     */
+    public ArmeriaStatusException(int code, String message) {
+        super(message);
+        this.code = code;
+    }
+    /**
+     * Construct a {@link ArmeriaStatusException} for the given gRPC status code, message, and cause.
+     */
+
```
trustin: This empty line could be moved before `/**`.

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaStatusException.java
```diff
@@ -0,0 +1,45 @@
...
+ * under the License.
+ */
+
+package com.linecorp.armeria.common.grpc.protocol;
+
+/**
+ * A {@link Exception} that contains enough information to convert it to a gRPC status.
+ */
+public class ArmeriaStatusException extends RuntimeException {
+
```
trustin: `serialVersionUID` is missing.

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaStatusException.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Construct a {@link ArmeriaStatusException} for the given gRPC status code, message, and cause.
+     */
+
+    public ArmeriaStatusException(int code, String message, Throwable cause) {
+        super(message, cause);
+        this.code = code;
+    }
+
+    public int getCode() {
```
trustin: Missing Javadoc

grpc/src/main/java/com/linecorp/armeria/internal/grpc/HttpStreamReader.java
```diff
@@ -67,7 +67,7 @@
     private boolean sawLeadingHeaders;
 
     public HttpStreamReader(DecompressorRegistry decompressorRegistry,
-                            ArmeriaMessageDeframer deframer,
+                            com.linecorp.armeria.common.grpc.protocol.ArmeriaMessageDeframer deframer,
```
trustin: Could use `import`?

grpc/src/main/java/com/linecorp/armeria/internal/grpc/StatusExceptionConverter.java
```diff
@@ -0,0 +1,36 @@
...
+import io.grpc.StatusRuntimeException;
+
+public final class StatusExceptionConverter {
+
+    public static StatusRuntimeException toGrpc(ArmeriaStatusException armeria) {
+        StatusRuntimeException converted = Status.fromCodeValue(armeria.getCode())
+                                                 .withDescription(armeria.getMessage())
+                                                 .withCause(armeria.getCause())
+                                                 .asRuntimeException();
+        converted.setStackTrace(armeria.getStackTrace());
```
trustin: I'm curious if we fill the stack trace twice - 1) at `ArmeriaStatusException.<init>` and 2) `asRuntimeException()` here. If so, can we avoid the 2nd one?

anuraaga: Yeah we do... I almost could except that the constructor we'd need on `StatusRuntimeException`, which accepts a boolean for filling the trace, is not `public`. Since this is only in error cases, maybe it's uncommon enough not to worry about the performance, but what do you think?

Other options
- Create an exception that extends `StatusRuntimeException` with a constructor we can use
- Create the `Status` manually and set the cause to the `ArmeriaStatusException`

In both cases, the behavior of the exception would be slightly different than upstream. Quite small difference but lean towards keeping them the same.

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaMessageDeframer.java
```diff
@@ -187,6 +184,9 @@ public int hashCode() {
     private boolean inDelivery;
     private boolean startedDeframing;
 
+    /**
+     * Construct a {@link ArmeriaMessageDeframer} for reading messages out of a gRPC request or response.
```
minwoox: nit: a -> an

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaMessageDeframer.java
```diff
@@ -294,7 +294,10 @@ public boolean isClosed() {
         return unprocessed == null;
     }
 
-    public ArmeriaMessageDeframer decompressor(Decompressor decompressor) {
+    /**
+     * Set the {@link Decompressor} for this deframer.
```
minwoox: nit: Set -> Sets

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaMessageFramer.java
```diff
@@ -90,6 +87,9 @@
     private Compressor compressor;
     private boolean closed;
 
+    /**
+     * Constructs a {@link ArmeriaMessageFramer} to write messages to a gRPC request or response.
```
minwoox: nit: a -> an

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaStatusException.java
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.common.grpc.protocol;
+
+/**
+ * A {@link Exception} that contains enough information to convert it to a gRPC status.
```
minwoox: nit: A -> An

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaStatusException.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public class ArmeriaStatusException extends RuntimeException {
+
+    public static final long serialVersionUID = -8370257107063108923L;
+
+    private final int code;
+
+    /**
+     * Construct a {@link ArmeriaStatusException} for the given gRPC status code and message.
```
minwoox: nit: Construct a -> Constructs an

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/StatusCodes.java
```diff
@@ -0,0 +1,51 @@
...
+final class StatusCodes {
+
+    /**
+     * Some resource has been exhausted, perhaps a per-user quota, or
+     * perhaps the entire file system is out of space.
+     */
+    static final int RESOURCE_EXHAUSTED = 8;
+
+    /**
+     * Internal errors.  Means some invariants expected by underlying
```
minwoox: nit: two spaces before `Means` and `If`

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ForwardingCompressor.java
```diff
@@ -0,0 +1,56 @@
...
+import io.grpc.Codec.Identity;
+
+/**
+ * A {@link Compressor} that forwards to a {@link io.grpc.Compressor}.
+ */
+public final class ForwardingCompressor implements Compressor {
+
+    @Nullable
+    public static Compressor forGrpc(io.grpc.Compressor delegate) {
+        if (delegate == Identity.NONE) {
```
minwoox: `requireNonNull(delegate, "delegate")`?

grpc/src/main/java/com/linecorp/armeria/internal/grpc/ForwardingDecompressor.java
```diff
@@ -0,0 +1,56 @@
...
+import io.grpc.Codec.Identity;
+
+/**
+ * A {@link Decompressor} that forwards to a {@link io.grpc.Decompressor}.
+ */
+public final class ForwardingDecompressor implements Decompressor {
+
+    @Nullable
+    public static ForwardingDecompressor forGrpc(io.grpc.Decompressor delegate) {
+        if (delegate == Identity.NONE) {
```
minwoox: ditto

grpc/src/main/java/com/linecorp/armeria/internal/grpc/StatusExceptionConverter.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.linecorp.armeria.common.grpc.protocol.ArmeriaStatusException;
+
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
+
+public final class StatusExceptionConverter {
+
+    public static StatusRuntimeException toGrpc(ArmeriaStatusException armeria) {
+        StatusRuntimeException converted = Status.fromCodeValue(armeria.getCode())
```
minwoox: nit: `requireNonNull(armeria, "armeria")`? and could be final?

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/StatusCodes.java
anuraaga: Interestingly I just copied this from upstream ;)

anuraaga:
Thanks, updated

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaStatusException.java
```diff
@@ -0,0 +1,50 @@
...
+ * An {@link Exception} that contains enough information to convert it to a gRPC status.
+ */
+public class ArmeriaStatusException extends RuntimeException {
+
+    public static final long serialVersionUID = -8370257107063108923L;
+
+    private final int code;
+
+    /**
+     * Construct an {@link ArmeriaStatusException} for the given gRPC status code and message.
```
hyangtack: nit: Construct -> Constructs

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaStatusException.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public ArmeriaStatusException(int code, String message) {
+        super(message);
+        this.code = code;
+    }
+
+    /**
+     * Construct an {@link ArmeriaStatusException} for the given gRPC status code, message, and cause.
```
hyangtack: nit: Construct -> Constructs
message, and -> message and (removing `,`)

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaStatusException.java
trustin: Yeah, seems like a bug in Checkstyle?

grpc/src/main/java/com/linecorp/armeria/internal/grpc/StatusExceptionConverter.java
trustin: OK. Let's leave it as it is for now. Could you add something lile:

```
XXX(anuraaga): We fill stack trace twice. Can we avoid the 2nd `fillInStackTrace()`?
```

..?

grpc/src/main/java/com/linecorp/armeria/internal/grpc/StatusExceptionConverter.java
```diff
@@ -0,0 +1,46 @@
...
+        requireNonNull(armeria, "armeria");
+        final StatusRuntimeException converted = Status.fromCodeValue(armeria.getCode())
+                                                       .withDescription(armeria.getMessage())
+                                                       .withCause(armeria.getCause())
+                                                       .asRuntimeException();
+        // NB(anuraaga): We end up filling the stacktrace of the exception twice, which is unfortunate since
+        // generating a stacktrace is expensive. Ideally, we could call the StatusRuntimeException with the
+        // toggle for filling in the stacktrace, but it is not a public API. Other approaches could be to
+        // subclass StatusRuntimeException (which will probably just cause upstream to declare the class final
+        // :P) or include ArmeriaStatusException in the cause chain of Status that we return from armeria. Both
```
hyangtack: nit: armeria -> Armeria 😜

anuraaga: Heh I always do this. I wonder if the IntelliJ settings.jar could be configured someway so lowercase shows a red line :)

trustin:
Thanks, @anuraaga. :-)

PR #1723, opened 2019-04-18
UnframedGrpcService now sends a response even there is no content to …

### LINKED ISSUE #1721: No response from UnframedGrpcService if onNext call is missing
Assume that we have a following implementation for `Hello` method:
```
    @Override
    public void hello(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
        // Note that 'responseObserver.onNext()' is not called.
        responseObserver.onCompleted();
    }
```

If a user calls `Hello` method from the `DocService`, the server does not respond anything.

/cc @anuraaga

hyangtack:
…respond

Modifications:
- Made the future completed correctly with `HttpResponse` even if no message was read by `ArmeriaMessageDeframer`.

Result:
- Closes #1721

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/server/grpc/UnframedGrpcService.java
+++ b/grpc/src/main/java/com/linecorp/armeria/server/grpc/UnframedGrpcService.java
@@ -256,8 +256,7 @@ class UnframedGrpcService extends SimpleDecoratingService<HttpRequest, HttpRespo
                 new Listener() {
                     @Override
                     public void messageRead(ByteBufOrStream message) {
-                        // We know there is only one message in total, so don't bother with checking endOfStream
-                        // We also know that we don't support compression, so this is always a ByteBuffer.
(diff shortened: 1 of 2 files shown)


grpc/src/main/java/com/linecorp/armeria/server/grpc/UnframedGrpcService.java
```diff
@@ -264,7 +264,11 @@ public void messageRead(ByteBufOrStream message) {
                     }
 
                     @Override
-                    public void endOfStream() {}
+                    public void endOfStream() {
+                        if (!res.isDone()) {
```
anuraaga: Can you update the comment above or remove it? It was wrong since the reality was 0-1 message, not exactly 1.

hyangtack: Thanks! Updated.

trustin:
Thanks, @hyangtack !

PR #1726, opened 2019-04-18
Make `RedirectService` preserve the query string by default

### LINKED ISSUE #1725: RedirectService doesn't preserve query parameters
I noticed `RedirectService` by default redirects without preserving query parameters. There is the `locationFunction` that could be used by users to preserve query parameters, but I guess this might be common enough that we provide such a location function that preserves the query parameters, or even just always preserve them. What do you think?

trustin:
Motivation:

When redirecting, a user would want to preserve the query string in the
original request URI, by appending it to the end of the new location.

Modifications:

- Make `RedirectService` preserve the query string by default.
  - Note: the query string is not appended if the new location already
    contains a query string.
- Add `preserveQueryString` parameter to the constructors of
  `RedirectService` so that a user can go back to the old behavior.
- Update the class documentation of `RedirectService`.

Result:

- Fixes #1725
- Less surprising behavior.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/RedirectService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/RedirectService.java
@@ -274,9 +274,7 @@ public class RedirectService extends AbstractHttpService {
             return location;
         }
 
-        final StringBuilder buf = new StringBuilder(location.length() + query.length() + 1);
-        buf.append(location).append('?').append(query);
-        return buf.toString();
+        return location + '?' + query;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/server/RedirectService.java
```diff
@@ -128,16 +247,36 @@ public RedirectService(HttpStatus redirectStatus,
...
+
+    private static String appendQueryString(ServiceRequestContext ctx, String location) {
+        final String query = ctx.query();
+        if (query == null || location.lastIndexOf('?') >= 0) {
+            // The request URI does not have a query string or
+            // The new location includes a query string already.
+            return location;
+        }
+
+        final StringBuilder buf = new StringBuilder(location.length() + query.length() + 1);
```
anuraaga: Since there isn't any conditional logic, concatenation should be at least as fast as this.

trustin: Oh, does simple concatenation perform as well as or better than this?

anuraaga: Yeah in Java 8 it should be the same bytecode AFAIU. In future Java, since they can customize concatenation bytecode using invokedynamic it may get even better

adriancole:
cool

trustin:
Thanks for reviewing. 😄

PR #1727, opened 2019-04-19
Extract grpc protocol to separate artifact and add AbstractUnaryGrpcService.

anuraaga:
For #1703 

Tried extracting a package within the armeria-grpc artifact that only contains wire format logic, but it is fragile due to accidental dependencies and classpath magic like service loaders, which we use. So now we extract to a separate artifact entirely.

`AbstractUnaryGrpcService` allows defining unary gRPC services that are provided the direct bytes of a single gRPC message and return the bytes of a single gRPC message, and the base class takes cafe of the gRPC wire protocol. As this is an advanced use case, it doesn't attempt to implement everything in gRPC but should hopefully provide enough for most armeria-grpc users that are concerned about having dependencies on stub libraries.

@adriancole can you check whether this works for you or provide any suggestions?


grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/AbstractUnaryGrpcService.java
```diff
@@ -0,0 +1,122 @@
...
+        return HttpResponse.from(responseFuture);
+    }
+
+    private CompletableFuture<ByteBuf> deframeMessage(HttpData framed, ByteBufAllocator alloc) {
+        final CompletableFuture<ByteBuf> deframed = new CompletableFuture<>();
+        try (ArmeriaMessageDeframer deframer = new ArmeriaMessageDeframer(
+                new Listener() {
+                    @Override
+                    public void messageRead(ByteBufOrStream message) {
+                        // Compression not supported.
```
adriancole: question. does grpc compress by default?

anuraaga: Nope, users have to explicitly enable it. And even if a client happened to enabled it, since we don't return any supported encodings in the response headers, they will fallback to sending uncompressed messages anyways.

anuraaga: Nope, users have to explicitly enable it. And even if a client happened to enabled it, since we don't return any supported encodings in the response headers, they will fallback to sending uncompressed messages anyways.

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/AbstractUnaryGrpcService.java
```diff
@@ -0,0 +1,122 @@
...
+     * Returns an unframed response message to return to the client, given an unframed request message. It is
+     * expected that the implementation has the logic to know how to parse the request and serialize a response
+     * into {@code byte[]}. The returned {@code byte[]} will be framed and returned to the client.
+     */
+    protected abstract CompletableFuture<byte[]> handleMessage(byte[] message);
+
+    @Override
+    protected final HttpResponse doPost(ServiceRequestContext ctx, HttpRequest req) {
+        final CompletableFuture<HttpResponse> responseFuture =
+                req.aggregateWithPooledObjects(ctx.contextAwareEventLoop(), ctx.alloc())
```
hyangtack: Just curious. We have `GrpcServiceBuilder#useBlockingTaskExecutor`, don't we need to respect the flag?

anuraaga: This service doesn't use `GrpcServiceBuilder` - it's meant to implement the protocol without gRPC :) As it's a very advanced API, I want to keep its feature set / API surface small so lean towards not adding knobs like that to this class. How does that sound?

hyangtack: Ah, I see. This class is definitely not related to the `GrpcServiceBuilder`. I mis-thought about that. Thank you for explanation.

adriancole:
wonder if you need to rebase this in order to get the gradle + JITPACK goodness or not..

adriancole:
at any rate I can just use snapshot to test :P

anuraaga:
Jitpack is set up by the depending project (e.g, zipkin), don't think we'd need to do anything here for you to try it

adriancole:
@anuraaga sorry.. didn't tag you on this reference https://github.com/line/gradle-scripts/pull/55

grpc/src/test/java/com/linecorp/armeria/common/grpc/protocol/AbstractUnaryGrpcServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+import com.linecorp.armeria.grpc.testing.Messages.Payload;
+import com.linecorp.armeria.grpc.testing.Messages.SimpleRequest;
+import com.linecorp.armeria.grpc.testing.Messages.SimpleResponse;
+import com.linecorp.armeria.grpc.testing.TestServiceGrpc.TestServiceBlockingStub;
+import com.linecorp.armeria.server.ServerBuilder;
+import com.linecorp.armeria.testing.server.ServerRule;
+
+public class AbstractUnaryGrpcServiceTest {
+
+    // This service only depends on protobuf. Users can use a custom decoder / encoder to avoid even that.
```
adriancole: maybe add an example with an empty response? I'm not sure if my code is messed up or not :) 

```
io.grpc.StatusRuntimeException: UNKNOWN: HTTP status code 200
invalid content-type: null
headers: Metadata(:status=200,grpc-encoding=identity)
DATA-----------------------------

**Comment by adriancole** [comment]
looks like still a dep on grpc's utilities, used for the following header

```
    /**
     * {@code "armeria.grpc.ThrowableProto-bin"}.
     */
    public static final AsciiString ARMERIA_GRPC_THROWABLEPROTO_BIN =
            HttpHeaderNames.of(ProtoUtils.keyForProto(ThrowableProto.getDefaultInstance()).name());
```



```
Caused by: java.lang.ClassNotFoundException: io.grpc.protobuf.ProtoUtils
	at java.net.URLClassLoader.findClass(URLClassLoader.java:471) ~[?:?]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:588) ~[?:?]
	at org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:93) ~[zipkin-server-2.13.0-SNAPSHOT-exec.jar:?]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:521) ~[?:?]
	at com.linecorp.armeria.common.grpc.GrpcHeaderNames.<clinit>(GrpcHeaderNames.java:52) ~[armeria-grpc-0.84.0-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.common.grpc.protocol.AbstractUnaryGrpcService.<clinit>(AbstractUnaryGrpcService.java:47) ~[armeria-grpc-0.84.0-SNAPSHOT.jar!/:?]
```

File: grpc/src/test/java/com/linecorp/armeria/common/grpc/protocol/AbstractUnaryGrpcServiceTest.java (hunk 3 above)
Comment (trustin) [reply grpc/src/test/java/com/linecorp/armeria/common/grpc/protocol/AbstractUnaryGrpcServiceTest.java]: @anuraaga @adriancole Please let me know if this PR is ready to be merged.

**Comment by adriancole** [comment]
testing by inlining "armeria.grpc.ThrowableProto-bin" vs using code to generate it

**Comment by adriancole** [comment]
right now I am building with a small patch. I still have the exception above, just trying to figure out what to do about it.

```diff
diff --git a/grpc/src/main/java/com/linecorp/armeria/common/grpc/GrpcHeaderNames.java b/grpc/src/main/java/com/linecorp/armeria/common/grpc/GrpcHeaderNames.java
index b14f5442..0c0bb342 100644
--- a/grpc/src/main/java/com/linecorp/armeria/common/grpc/GrpcHeaderNames.java
+++ b/grpc/src/main/java/com/linecorp/armeria/common/grpc/GrpcHeaderNames.java
@@ -17,8 +17,6 @@
 package com.linecorp.armeria.common.grpc;
 
 import com.linecorp.armeria.common.HttpHeaderNames;
-
-import io.grpc.protobuf.ProtoUtils;
 import io.netty.util.AsciiString;
 
 /**
@@ -49,7 +47,7 @@ public final class GrpcHeaderNames {
      * {@code "armeria.grpc.ThrowableProto-bin"}.
      */
     public static final AsciiString ARMERIA_GRPC_THROWABLEPROTO_BIN =
-            HttpHeaderNames.of(ProtoUtils.keyForProto(ThrowableProto.getDefaultInstance()).name());
+            HttpHeaderNames.of("armeria.grpc.ThrowableProto-bin");
 
     private GrpcHeaderNames() {}
 }
```

adriancole:
this also crashes the server.. looking into it now.

```
java.lang.NoClassDefFoundError: io/grpc/ServerServiceDefinition
	at com.linecorp.armeria.server.grpc.GrpcDocServicePlugin.loadDocStrings(GrpcDocServicePlugin.java:195) ~[armeria-grpc-0.84.0-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.server.docs.DocService.lambda$addDocStrings$4(DocService.java:181) ~[armeria-0.84.0-SNAPSHOT.jar!/:?]
	at java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:271) ~[?:?]
	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474) ~[?:?]
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913) ~[?:?]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578) ~[?:?]
	at com.linecorp.armeria.server.docs.DocService.addDocStrings(DocService.java:183) ~[armeria-0.84.0-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.server.docs.DocService.access$200(DocService.java:80) ~[armeria-0.84.0-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.server.docs.DocService$1.serverStarting(DocService.java:161) ~[armeria-0.84.0-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.server.Server$ServerStartStopSupport.notifyStarting(Server.java:494) ~[armeria-0.84.0-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.server.Server$ServerStartStopSupport.notifyStarting(Server.java:266) ~[armeria-0.84.0-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.common.util.StartStopSupport.notifyListeners(StartStopSupport.java:342) ~[armeria-0.84.0-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.common.util.StartStopSupport.lambda$start$3(StartStopSupport.java:171) ~[armeria-0.84.0-SNAPSHOT.jar!/:?]
	at io.netty.util.concurrent.GlobalEventExecutor$TaskRunner.run(GlobalEventExecutor.java:248) [netty-common-4.1.34.Final.jar!/:4.1.34.Final]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-common-4.1.34.Final.jar!/:4.1.34.Final]
	at java.lang.Thread.run(Thread.java:834) [?:?]
Caused by: java.lang.ClassNotFoundException: io.grpc.ServerServiceDefinition
	at java.net.URLClassLoader.findClass(URLClassLoader.java:471) ~[?:?]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:588) ~[?:?]
	at org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:93) ~[zipkin-server-2.13.0-SNAPSHOT-exec.jar:?]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:521) ~[?:?]
	... 19 more
```

grpc/src/test/java/com/linecorp/armeria/common/grpc/protocol/AbstractUnaryGrpcServiceTest.java
anuraaga: Not ready - I didn't follow my package-info comment and depended on armeria-grpc's GrpcHeaderNames, which isn't armeria-core ;) Will fix it probably by inlining and moving the class.

anuraaga:
Ah I think the doc service stuff will get triggered as long as armeria-grpc is on the classpath, using service loader.

This would be a reason to extract a separate artifact for `grpc-protocol`. Otherwise, I can just add a try / catch there which feels yucky but might be fine. @trustin any thoughts?

adriancole:
I've a try/catch in progress.. just testing now

```diff
--- a/core/src/main/java/com/linecorp/armeria/server/docs/DocService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/docs/DocService.java
@@ -32,6 +32,7 @@ import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
@@ -178,8 +179,14 @@ public class DocService extends AbstractCompositeService<HttpRequest, HttpRespon
     private static ServiceSpecification addDocStrings(ServiceSpecification spec, List<ServiceConfig> services) {
         final Map<String, String> docStrings =
                 plugins.stream()
-                       .flatMap(plugin -> plugin.loadDocStrings(findSupportedServices(plugin, services))
-                                                .entrySet().stream())
+                       .flatMap(plugin -> {
+                           try { // Guard when optional classpath dependencies are unmet
+                               return plugin.loadDocStrings(findSupportedServices(plugin, services))
+                                   .entrySet().stream();
+                           } catch (NoClassDefFoundError e) {
+                               return Stream.empty();
+                           }
+                       })
                        .collect(toImmutableMap(Entry::getKey, Entry::getValue, (a, b) -> a));
 
         return new ServiceSpecification(
```

adriancole:
ok with the two diffs I posted the server starts. the response is different, which I'm looking into now. Seems like a missing content type header.

before
```bash
$ curl -X POST localhost:9411/zipkin.proto3.SpanService/Report -H'Content-Type: application/grpc' -v
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 9411 (#0)
> POST /zipkin.proto3.SpanService/Report HTTP/1.1
> Host: localhost:9411
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Type: application/grpc
> 
< HTTP/1.1 200 OK
< content-type: application/grpc+proto
< grpc-status: 13
< grpc-message: Half-closed without a request
< content-length: 0
< 
* Connection #0 to host localhost left intact
```

now (plus the two patches)
```bash
$ curl -X POST localhost:9411/zipkin.proto3.SpanService/Report -H'Content-Type: application/grpc' -v
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 9411 (#0)
> POST /zipkin.proto3.SpanService/Report HTTP/1.1
> Host: localhost:9411
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Type: application/grpc
> 
< HTTP/1.1 200 OK
< grpc-encoding: identity
< transfer-encoding: chunked
< 
* Connection #0 to host localhost left intact
```

grpc/src/main/java/com/linecorp/armeria/common/grpc/protocol/AbstractUnaryGrpcService.java
```diff
@@ -0,0 +1,122 @@
...
+                       } finally {
+                           requestMessage.release();
+                       }
+                       return handleMessage(requestMessageBytes);
+                   })
+                   .thenApply(responseMessage -> {
+                       final ArmeriaMessageFramer framer = new ArmeriaMessageFramer(
+                               ctx.alloc(), Integer.MAX_VALUE);
+                       final HttpData framed = framer.writePayload(Unpooled.wrappedBuffer(responseMessage));
+                       return HttpResponse.of(
```
adriancole: this is where my empty response gets.. though the content is different and that trips up gRPC client (might be my mistake)

adriancole:
ps hold a tick as I think the path I'm looking at isn't actually required..

adriancole:
here's my entire code now.. same thing with the difference in curl result (also with grpc client fail). Not sure if there's a smoking gun here.

```proto
service SpanService {
  rpc Report(zipkin.proto3.ListOfSpans) returns (google.protobuf.Empty) {}
}
```

```java
@ConditionalOnProperty(name = "zipkin.collector.grpc.enabled") // disabled by default
final class ZipkinGrpcCollector {
  @Bean ArmeriaServerConfigurator grpcCollectorConfigurator(StorageComponent storage,
    CollectorSampler sampler, CollectorMetrics metrics) {
    CollectorMetrics grpcMetrics = metrics.forTransport("grpc");
    Collector collector = Collector.newBuilder(getClass())
      .storage(storage)
      .sampler(sampler)
      .metrics(grpcMetrics)
      .build();

    return sb ->
      sb.service("/zipkin.proto3.SpanService/Report", new SpanService(collector, grpcMetrics));
  }

  static final class SpanService extends AbstractUnaryGrpcService {

    final Collector collector;
    final CollectorMetrics metrics;

    SpanService(Collector collector, CollectorMetrics metrics) {
      this.collector = collector;
      this.metrics = metrics;
    }

    @Override protected CompletableFuture<byte[]> handleMessage(byte[] bytes) {
      metrics.incrementMessages();
      CompletableFutureCallback result = new CompletableFutureCallback();
      collector.acceptSpans(bytes, SpanBytesDecoder.PROTO3, result);
      return result;
    }
  }

  static final class CompletableFutureCallback extends CompletableFuture<byte[]>
    implements Callback<Void> {

    @Override public void onSuccess(Void value) {
      complete(new byte[0] /* empty result */);
    }

    @Override public void onError(Throwable t) {
      completeExceptionally(t);
    }
  }
}
```

adriancole:
it seems the thing grpc's client dislikes most is the missing content-type in the response

```
io.grpc.StatusRuntimeException: UNKNOWN: HTTP status code 200
invalid content-type: null
headers: Metadata(:status=200,grpc-encoding=identity)
DATA-----------------------------

anuraaga:
Yeah will add that in, that was not intended. Will also update the test to use upstream client instead of armeria client to catch that, since it's more important to test that anyways. Thanks for checking out the code!

adriancole:
actually there's a bit tougher issue trying to exclude protobuf (which reduces the jar cost of this integration to near nothing). serviceloader pukes on .hasNext()

trustin:
@anuraaga wrote:
> This would be a reason to extract a separate artifact for `grpc-protocol`. Otherwise, I can just add a try / catch there which feels yucky but might be fine. @trustin any thoughts?

+1 for a separate module, because otherwise it will be fairly easy to break this again and again.

anuraaga:
Ready for another look.

Extracted a module, which solves doc service issue and ensures no more accidental usage of gRPC dependencies. And fixed content type header, upstream client went from not working to working in a unit test. 

@adriancole think you should be good. I merged master so jitpack might work too

adriancole:
trying now thanks!

adriancole:
Screen Shot 2019-04-22 at 8 19 16 PM

the anticipation! cc @bsideup

adriancole:
dunno why it adds in `0.4.0.Final-` to the stuff.. 

https://jitpack.io/com/github/anuraaga/armeria/unary-grpc-service-armeria-0.4.0.Final-g792dcc8-1703/build.log

adriancole:
ignore the jitpack related comments.. will follow-up on these elsewhere

adriancole:
shipit! this works https://github.com/apache/incubator-zipkin/pull/2328/files

trustin:
I think we can merge this and do a follow-up on `armeria-grpc-protocol` later. Sounds good, @anuraaga and @adriancole ?

anuraaga:
Yeah I'm good but just to clarify, what do you mean by follow up? Do you mean the name of the artifact?

trustin:
@anuraaga That's because I didn't know you finished the separation already. That was fast. :+1:

trustin:
Could you update the PR description and commit message finally so I can merge this PR?

anuraaga:
Updated description :)

trustin:
@hyangtack Could you take another look since it was updated since your last review?

trustin:
Thanks, @anuraaga and @adriancole !

Issue #1728, opened 2019-04-19
armeria-thrift does not handle service method throwing exception


edgao:
With this thrift file: (compiled using thrift 0.11.0)
```thrift
exception TestException {
  1: string foo;
}

service TestService {
  string ping() throws (1: TestException te);
}
```

and this main method: (using armeria 0.83.0)
```java
  public static void main(String[] args) throws Exception {
    Server server = new ServerBuilder()
        .verboseResponses(true)
        .http(SERVICE_PORT)
        .service(
            "/" + SERVICE_LOCATION,
            THttpService.ofFormats(
                (TestService.Iface)() -> "pong",
                ThriftSerializationFormats.COMPACT
            )
        ).build();
    CompletableFuture<Void> future = server.start();
    future.join();
  }
```

Armeria throws this exception:
```text
Exception in thread "main" java.lang.IllegalArgumentException: failed to retrieve function metadata: com.liveramp.partner_stats_service.TestService$Iface.ping()
	at com.linecorp.armeria.internal.thrift.ThriftServiceMetadata.registerFunction(ThriftServiceMetadata.java:190)
	at com.linecorp.armeria.internal.thrift.ThriftServiceMetadata.lambda$init$1(ThriftServiceMetadata.java:94)
	at java.util.HashMap.forEach(HashMap.java:1289)
	at java.util.Collections$UnmodifiableMap.forEach(Collections.java:1505)
	at com.linecorp.armeria.internal.thrift.ThriftServiceMetadata.init(ThriftServiceMetadata.java:93)
	at com.linecorp.armeria.internal.thrift.ThriftServiceMetadata.init(ThriftServiceMetadata.java:72)
	at com.linecorp.armeria.internal.thrift.ThriftServiceMetadata.<init>(ThriftServiceMetadata.java:60)
	at com.linecorp.armeria.server.thrift.ThriftServiceEntry.<init>(ThriftServiceEntry.java:49)
	at com.linecorp.armeria.internal.shaded.guava.collect.CollectCollectors.lambda$toImmutableMap$1(CollectCollectors.java:61)
	at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.util.Iterator.forEachRemaining(Iterator.java:116)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
[...]

PR #1729, opened 2019-04-19
Use `Class.getDeclaredField()` instead of `getField()` when fetching Thrift metadata.

edgao:
fixes https://github.com/line/armeria/issues/1728 ?

Was able to test this locally and confirm that it works.

### DIFF
--- a/thrift/src/test/java/com/linecorp/armeria/server/thrift/ThriftServiceTest.java
+++ b/thrift/src/test/java/com/linecorp/armeria/server/thrift/ThriftServiceTest.java
@@ -52,8 +52,12 @@ import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.SerializationFormat;
 import com.linecorp.armeria.common.thrift.ThriftProtocolFactories;
 import com.linecorp.armeria.common.thrift.ThriftSerializationFormats;
+import com.linecorp.armeria.common.thrift.text.Response;
+import com.linecorp.armeria.common.thrift.text.RpcDebugService;
 import com.linecorp.armeria.common.util.CompletionActions;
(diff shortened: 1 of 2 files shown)


edgao:
Update: Turns out this is because we use the `java:private-members` option when generating thrift code, so the unit test I added doesn't actually do anything. Specifically, without this option set, the `TestService` class contains this internal class:

```java
public static class ping_result implements org.apache.thrift.TBase<ping_result, ping_result._Fields>, java.io.Serializable, Cloneable, Comparable<ping_result>   {
    // <snip>
    public TestException te; // required
    // <snip>
}
```

But with `private-members` enabled, `te` is a private field, and so `getFields` does not return it.

thrift/src/test/java/com/linecorp/armeria/server/thrift/ThriftServiceTest.java
```diff
@@ -600,6 +604,19 @@ public void testMultipleInheritance() throws Exception {
         assertThat(client2.recv_sort()).containsExactly(NAME_A, NAME_B, NAME_C);
     }
 
+    @Test
+    public void testServiceThrowingException() {
```
trustin: Could you revert this part now that we know this does not reproduce the problem?

edgao: Done.

trustin:
@hyangtack Please review. :-)

trustin:
Thanks again, @edgao !

edgao:
thanks for the super quick turnaround! just curious - do you know when the next official release will be? (not a blocker for us since we've built a patched jar)

Issue #1567, opened 2019-01-30
Immutable HttpHeaders and Cookies


trustin:
`HttpHeaders` in Armeria is mutable. Although it is convenient to use, it sometimes leads us to a surprising behavior when it's modified by an unexpected place.

To address such issues, I'd like to propose to make `HttpHeaders` immutable, which means modifying an `HttpHeaders` will result in a new `HttpHeaders` instance. The following is an example of the API I have in mind:

```java
// Mutable headers. Behaves like current HttpHeaders do.
MutableHeaders mh1 = MutableHeaders.of().add("name", "value");

// Immutable headers. Does not have methods like 'set()'.
Headers ih1 = ImmutableHeaders.of("name", "value", ...);

// Conversion to immutable headers.
Headers ih2 = mh1.toImmutable();

// Conversion to mutable headers.
MutableHeaders mh2 = ih2.toMutable();
assertThat(mh2).isEqualTo(mh1);
assertThat(mh2).isNotSameAs(mh1); 

// Mutation of immutable headers.
Headers ih3 = ih2.mutate((MutableHeaders h) -> {
    h.set("name", "new-value");
    ...
});

// One-time mutation.
Headers ih4 = ih2.mutator().set("name", "new-value");
```

Thoughts? I'd like to get as much feed back (e.g. what could be a problem when we do this?) as possible because this will be a big breaking change. To be honest, I may even fail to refactor Armeria itself because Armeria itself relies on mutability of `HttpHeaders` in some places.

Note that I removed `Http` from the type names, which may need an additional discussion as well.

southernkasaist:
How about using `builder` way like lombok style?
e.g., 
```java
Headers headers = ImmutableHeaders.builder().add("name1", "value1").build();
Headers other = headers.toBuilder().add("name2", "value2").set("name1", "value3").build();
```

tobias-:
1. Please litter the methods in ImmutableHeaders with JSR-305 annotations such as `@javax.annotation.CheckReturnValue`as it really helps avoiding trivial mistakes.
2. Please consider doing it "Kotlin-style" where the Kotlin List (comparable to ImmutableHeaders) doesn't even have the add methods, but the MutableList (comparable to MutableHeaders) does. In Kotlin, if you need to mutate a List, you will have to make it mutable by doing `fooList.toMutableList()` before doing add. Unless a method/parameter is explicitly MutableList, it will be a List, which will not even have those methods.

trustin:
[...]

PR #1731, opened 2019-04-20
Immutable `HttpHeaders`

trustin:
Related: #1567
Motivation:

`HttpHeaders` is mutable. It is currently very easy for everyone to update the request or response headers a lot later than expected, causing various hard-to-debug problems.

Modifications:

- Replaced the current `HttpHeaders` API and implementations with the new one.
  - `HttpHeaders` has some type hierarchy now.
    - `HttpHeaders` as a generic HTTP header container.
      - `HttpHeaders` does not have short methods like `path()`, `method()` and `status()`.
        `RequestHeaders` and `ResponseHeaders` have them.
    - `RequestHeaders` as a request-specific HTTP header container.
    - `ResponseHeaders` as a response-specific HTTP header container.
  - There are builders now:
    - `HttpHeadersBuilder`
    - `RequestHeadersBuilder`
    - `ResponseHeadersBuilder`
- Changed the overall API related with specifying HTTP headers.
  - We now accept `CharSequence` instead of `AsciiString` as a header name, so that a user can specify a `String` as a header name.
  - We now accept `Iterable<? extends Map.Entry<...>>` instead of `Headers<...>` so that a user can specify wider range of types.
  - A header name is converted to an `AsciiString` internally.
  - Deprecated `HttpHeaders.EMPTY_HEADERS` in favor of `HttpHeaders.of()`.
- Added `RequestContext.updateRequest()` so that a decorator author can update the request when filtering a request.

Result:

- We provide more secure API for representing HTTP headers.


core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
```diff
@@ -340,9 +338,9 @@ public B addHttpHeader(CharSequence name, Object value) {
     /**
      * Adds the specified HTTP headers.
      */
-    public B addHttpHeaders(Headers<AsciiString, String, ?> httpHeaders) {
+    public B addHttpHeaders(Iterable<? extends Entry<? extends CharSequence, ?>> httpHeaders) {
```
huydx: Is there any rationale to move from AsciiString, String to CharSequence ? is that just loosen the spec that HTTP header only able to be US-ASCII?

trustin: Internally, when you specify a `CharSequence` as a name, it will be automatically converted to an `AsciiString` using `HttpHeaderNames.of()`. The change is purely for user's convenience:

```java
// Before:
HttpHeaders headers = HttpHeaders.of(HttpHeaderNames.of("x-foo", "bar");
// After:
HttpHeaders headers = HttpHeaders.of("x-foo", "bar");
```

Unlike the names, the values are stored as-is in `CharSequence`, and converted to a `String` lazily. This change is based on the assumption that a user does not read all header values in reality, hence reducing the cost of unnecessary `CharSequence/AsciiString` to `String` conversion.

trustin:
I think this pull request is ready except the low test coverage. Please start to review or take a deep breath to get yourself prepared for this giant pull request. I'll update the PR description and add test cases :wink:

core/src/main/java/com/linecorp/armeria/common/RequestHeadersBuilder.java
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * Creates a new empty builder.
+     */
+    public RequestHeadersBuilder() {
+        this(DEFAULT_SIZE_HINT);
+    }
+
+    /**
+     * Creates a new empty builder with an hint about the size of the resulting {@link RequestHeaders}.
```
minwoox: nit: an -> a

core/src/main/java/com/linecorp/armeria/common/ResponseHeadersBuilder.java
```diff
@@ -0,0 +1,250 @@
...
+
+    /**
+     * Creates a new empty builder.
+     */
+    public ResponseHeadersBuilder() {
+        this(DEFAULT_SIZE_HINT);
+    }
+
+    /**
+     * Creates a new empty builder with an hint about the size of the resulting {@link ResponseHeaders}.
```
minwoox: nit: an -> a

core/src/main/java/com/linecorp/armeria/common/util/TextFormatter.java
```diff
@@ -183,7 +183,7 @@ public static void appendEpochMillis(StringBuilder buf, long timeMillis) {
      *
      * @return the human readable string representation of the given epoch time
      */
-    public static StringBuilder epochMicro(long timeMicros) {
+    public static StringBuilder epochMicros(long timeMicros) {
```
minwoox: Oops, thanks!

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -489,8 +515,18 @@ public static long parseCacheControlSeconds(@Nullable String value) {
     /**
      * Converts the specified Netty HTTP/2 into Armeria HTTP/2 headers.
      */
-    public static HttpHeaders toArmeria(Http2Headers headers, boolean endOfStream) {
-        final HttpHeaders converted = new DefaultHttpHeaders(false, headers.size(), endOfStream);
+    public static HttpHeaders toArmeria(Http2Headers headers, boolean request, boolean endOfStream) {
+        final HttpHeadersBuilder converted;
+        if (request) {
+            converted = headers.contains(HttpHeaderNames.METHOD) ? new RequestHeadersBuilder(headers.size())
```
minwoox: nit: `requireNonNull(headers, "headers");`

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -434,82 +439,102 @@ public void setMaxRequestLength(long maxRequestLength) {
...
         requireNonNull(name, "name");
-        return createAdditionalHeadersIfAbsent().remove(name);
+        final HttpHeaders additionalResponseHeaders = this.additionalResponseHeaders;
+        if (additionalResponseHeaders == null || additionalResponseHeaders.isEmpty()) {
+            return false;
+        }
+
+        final HttpHeadersBuilder builder = createAdditionalHeadersIfAbsent().toBuilder();
+        final boolean removed = builder.remove(name);
+        this.additionalResponseHeaders = builder.build();
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -434,82 +439,102 @@ public void setMaxRequestLength(long maxRequestLength) {
...
         requireNonNull(name, "name");
-        return createAdditionalTrailersIfAbsent().remove(name);
+        final HttpHeaders additionalResponseTrailers = this.additionalResponseTrailers;
+        if (additionalResponseTrailers == null || additionalResponseTrailers.isEmpty()) {
+            return false;
+        }
+
+        final HttpHeadersBuilder builder = createAdditionalTrailersIfAbsent().toBuilder();
+        final boolean removed = builder.remove(name);
+        this.additionalResponseTrailers = builder.build();
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/server/auth/OAuth2TokenExtractor.java
```diff
@@ -27,28 +27,30 @@
...
-        this.header = header;
+    OAuth2TokenExtractor(CharSequence header) {
+        this.header = HttpHeaderNames.of(header);
     }
 
     @Nullable
     @Override
-    public OAuth2Token apply(HttpHeaders headers) {
+    public OAuth2Token apply(RequestHeaders headers) {
         final String authorization = headers.get(header);
```
minwoox: should check `null` for `headers`?

core/src/main/java/com/linecorp/armeria/server/cors/CorsConfig.java
```diff
@@ -118,14 +116,18 @@ public boolean isShortCircuit() {
      * Returns the policy for the specified {@code origin}.
      *
      * @return {@link CorsPolicy} which allows the {@code origin},
-     *         {@code null} if the {@code origin} is not allowed in any policy.
+     *         {@code null} if the {@code origin} is {@code null} or not allowed in any policy.
      */
     @Nullable
-    public CorsPolicy getPolicy(String origin, PathMappingContext pathMappingContext) {
-        requireNonNull(origin, "origin");
+    public CorsPolicy getPolicy(@Nullable String origin, PathMappingContext pathMappingContext) {
```
minwoox: Seems like `null` check for `pathMappingContext` is missing.

core/src/main/java/com/linecorp/armeria/server/docs/ServiceSpecification.java
```diff
@@ -41,7 +40,7 @@
  */
 public final class ServiceSpecification {
 
-    private static ServiceSpecification emptyServiceSpecification =
+    private static final ServiceSpecification emptyServiceSpecification =
```
minwoox: Oops, thanks!

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1664 @@
...
+     * @param headers The headers used to {@link #set(CharSequence, CharSequence)} values in this instance
+     * @return {@code this}
+     */
+    public HttpHeadersBase set(
+            Iterable<? extends Entry<? extends CharSequence, ? extends CharSequence>> headers) {
+        if (headers == this) {
+            return this;
+        }
+
+        for (Entry<? extends CharSequence, ? extends CharSequence> e : headers) {
```
minwoox: Seems like we need `null` check for `headers`

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1664 @@
...
+        }
+
+        if (!addFast(headers)) {
+            addSlow(headers);
+        }
+        return this;
+    }
+
+    /**
+     * Sets a new header. Any existing headers with this name are removed. Before the {@code value} is added,
```
minwoox: nit: with this name -> with the specified name?

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1664 @@
...
+        }
+
+        return this;
+    }
+
+    /**
+     * Retains all current headers but calls {@link #setObject(CharSequence, Object)} for each entry in
+     * the specified {@code headers}.
+     *
+     * @param headers The headers used to set the values in this instance
```
minwoox: nit: The -> the

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1664 @@
...
+     * @return {@code this}
+     */
+    public HttpHeadersBase setObject(Iterable<? extends Entry<? extends CharSequence, ?>> headers) {
+        if (headers == this) {
```
minwoox: Seems like we need `null` check for `headers`

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1664 @@
...
+        }
+
+        return this;
+    }
+
+    /**
+     * Retains all current headers but calls {@link #set(CharSequence, CharSequence)} for each entry in
+     * the specified {@code headers}.
+     *
+     * @param headers The headers used to {@link #set(CharSequence, CharSequence)} values in this instance
```
minwoox: nit: The -> the

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
trustin: It's an internal use only API, so I'd like to skip null check here.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1664 @@
...
+     * for (CharSequence value : values) {
+     *     headers.add(name, value);
+     * }
+     * }</pre>
+     *
+     * @param name the header name
+     * @param values the values of the header
+     * @return {@code this}
+     */
+    public HttpHeadersBase add(CharSequence name, CharSequence... values) {
```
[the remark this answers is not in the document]
trustin: Yes.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -257,46 +256,51 @@ public void setMaxResponseLength(long maxResponseLength) {
...
         final HttpHeaders additionalRequestHeaders = this.additionalRequestHeaders;
-        if (additionalRequestHeaders == null) {
+        if (additionalRequestHeaders == null || additionalRequestHeaders.isEmpty()) {
             return false;
         }
-        return additionalRequestHeaders.remove(name);
+
+        final HttpHeadersBuilder builder = additionalRequestHeaders.toBuilder();
+        final boolean removed = builder.remove(name);
+        this.additionalRequestHeaders = builder.build();
```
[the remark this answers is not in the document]
trustin: Updated `*HeadersBuilder` so that `build()` returns the original headers if no changes were made.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
trustin: Updated `*HeadersBuilder` so that `build()` returns the original headers if no changes were made.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
trustin: Updated `*HeadersBuilder` so that `build()` returns the original headers if no changes were made.

core/src/main/java/com/linecorp/armeria/common/HeaderOverridingHttpRequest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    @Override
+    @Nullable
+    public MediaType contentType() {
+        return headers.contentType();
+    }
+
+    @Override
+    public CompletableFuture<AggregatedHttpMessage> aggregate() {
+        return delegate.aggregate().thenApply(this::replaceHeaders);
```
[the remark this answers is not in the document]
minwoox: Yeah, there's nothing wrong with this, I just misunderstood the behavior. Thank s for letting me know.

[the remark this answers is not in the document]
trustin: No worries. :smile:

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1257 @@
...
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (!(o instanceof HttpHeaderGetters)) {
+            return false;
+        }
+
+        final HttpHeaderGetters h2 = (HttpHeaderGetters) o;
```
hyangtack: nit: `h2`->`h`?

core/src/main/java/com/linecorp/armeria/client/Clients.java
```diff
@@ -288,10 +285,46 @@ private static ClientBuilderParams builderParams(Object client) {
...
-    public static SafeCloseable withHttpHeader(AsciiString name, String value) {
+    public static SafeCloseable withHttpHeader(CharSequence name, CharSequence value) {
+        requireNonNull(name, "name");
+        requireNonNull(value, "value");
+        return withHttpHeaders(headers -> headers.toBuilder().set(name, value).build());
+    }
+
+    /**
+     * Sets the specified HTTP header in a thread-local variable so that the header is sent by the client call
+     * made from the current thread. Use the `try-with-resources` block with the returned
```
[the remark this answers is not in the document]
trustin: Oops! :sweat_smile:

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1257 @@
...
+        @Override
+        public String toString() {
+            if (key == null) {
+                return "<HEAD>";
+            }
+
+            assert value != null;
+            return new StringBuilder(key.length() + value.length() + 1)
+                    .append(key)
+                    .append(value)
```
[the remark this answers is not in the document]
trustin: :sweat_smile:

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: Renamed to `that`.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBuilder.java
```diff
@@ -15,212 +15,881 @@
...
+     * @param values the value of the header
+     * @return {@code this}
+     */
+    public HttpHeadersBuilder set(CharSequence name, Iterable<? extends CharSequence> values) {
+        setters().set(name, values);
+        return this;
+    }
+
+    /**
+     * Sets a header with the specified name and values. Any existing headers with the specified  name are
```
minwoox: nit: two spaces before `name are`

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBuilder.java
```diff
@@ -15,212 +15,881 @@
...
+    public HttpHeadersBuilder setObject(Iterable<? extends Entry<? extends CharSequence, ?>> headers) {
+        setters().setObject(headers);
+        return this;
+    }
+
+    /**
+     * Sets a header with the specified {@code name} to {@code value}. This will remove all previous values
+     * associated with {@code name}.
+     *
+     * @param name The name to modify
```
minwoox: nit: The -> the? Could you also change the following Javadoc, please?

core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodingClient.java
```diff
@@ -83,7 +83,14 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
             // Client specified encoding, so we don't do anything automatically.
             return delegate().execute(ctx, req);
         }
-        req.headers().set(HttpHeaderNames.ACCEPT_ENCODING, acceptEncodingHeader);
+
+        req = HttpRequest.of(req, req.headers().toBuilder()
+                                     .set(HttpHeaderNames.ACCEPT_ENCODING, acceptEncodingHeader)
+                                     .build());
+        if (ctx.request() instanceof HttpRequest) {
```
anuraaga: Is it more appropriate to compare `ctx.request()` to the original `req`? This instanceof seems like magic (took a couple minutes to understand it).

anuraaga: Alternatively, it would still be clearer if `ctx` had an `updateRequest` method which contained this `instanceof` since it seems like an implementation detail of that class, not here.

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
```diff
@@ -0,0 +1,310 @@
...
+import java.util.stream.Stream;
+
+import javax.annotation.Nullable;
+
+import com.google.common.collect.Streams;
+
+import io.netty.util.AsciiString;
+
+/**
+ * Provides the getter methods to {@link HttpHeaders} and {@link HttpHeadersBuilder}.
```
anuraaga: Would users be using this interface (it's public and in common)? If so I think the javadoc could have more detail on when / how it's used.

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
```diff
@@ -0,0 +1,310 @@
...
+ */
+public interface HttpHeaderGetters extends Iterable<Entry<AsciiString, String>> {
+
+    /**
+     * Tells whether the headers will be the last frame in an HTTP/2 stream.
+     */
+    boolean isEndOfStream();
+
+    /**
+     * Returns the value of the {@code "content-type"} header.
```
anuraaga: Maybe `Returns the parsed {@code content-type} header.`

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
```diff
@@ -0,0 +1,310 @@
...
+
+    /**
+     * Returns the value of a header with the specified {@code name}. If there are more than one value for
+     * the specified {@code name}, the first value in insertion order is returned.
+     *
+     * @param name the name of the header to retrieve
+     * @param defaultValue the default value
+     * @return the first header value or {@code defaultValue} if there is no such header
+     */
+    String get(CharSequence name, CharSequence defaultValue);
```
anuraaga: Since the return type is `String` I'd expect `defaultValue` to also be `String`.

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
```diff
@@ -0,0 +1,310 @@
...
+
+    /**
+     * Returns {@code true} if this headers does not contain any entries.
+     */
+    boolean isEmpty();
+
+    /**
+     * Returns a {@link Set} of all header names. The returned {@link Set} cannot be modified.
+     */
+    Set<AsciiString> names();
```
anuraaga: Should the iteration methods return `CharSequence` instead of `AsciiString`? All the other methods take `CharSequence` as the type of the key.

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
```diff
@@ -13,227 +13,157 @@
...
     }
 
     /**
-     * Returns new HTTP headers with a single entry.
+     * Returns a new {@link HttpHeaders} with the specified header. The value is converted into
+     * a {@link String} via {@link HttpHeadersBase#addObject(CharSequence, Object)}.
      */
-    static HttpHeaders of(AsciiString name, String value) {
-        return new DefaultHttpHeaders().add(name, value);
+    static HttpHeaders of(CharSequence name, Object value) {
```
anuraaga: Might be overthinking, but I always get nervous when the same method name has parameters that inherit from each other (I think many lint tools would complain about it too).

anuraaga: Also personally, I think it's extremely rare to use non-string header values in most user code, so the helpers might not be used much to warrant the potential confusion.

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
```diff
@@ -13,227 +13,157 @@
...
     }
 
     /**
-     * Returns the immutable view of this headers.
+     * Returns a new {@link HttpHeaders} with the specified headers. The values are converted into
+     * {@link String}s via {@link HttpHeadersBase#addObject(CharSequence, Object)}.
      */
-    default HttpHeaders asImmutable() {
-        return new ImmutableHttpHeaders(this);
+    static HttpHeaders of(CharSequence name1, Object value1,
```
anuraaga: How about also `CharSequence... nameValues` and `checkArgument(nameValues.length % 2 == 0)`?

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
```diff
@@ -13,227 +13,157 @@
...
+     * builder = toBuilder();
+     * mutator.accept(builder);
+     * return builder.build();
+     * }</pre>
+     *
+     * @see #toBuilder()
      */
-    default HttpHeaders toMutable() {
-        return this;
+    default HttpHeaders mutate(Consumer<HttpHeadersBuilder> mutator) {
```
anuraaga: Possibly `withMutations`?

Anyways just for reference, I think immutablejs provides a good immutable API

https://immutable-js.github.io/immutable-js/docs/#/Map/withMutations

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1214 @@
...
+        entries = new HeaderEntry[findNextPositivePowerOfTwo(max(2, min(sizeHint, 128)))];
+        hashMask = (byte) (entries.length - 1);
+        head = firstNonPseudo = new HeaderEntry();
+    }
+
+    /**
+     * Creates an immutable shallow copy or a mutable deep copy of the specified {@link HttpHeadersBase}.
+     * <ul>
+     *   <li>If the {@code immutable} parameter is {@code true}, both the specified {@link HttpHeadersBase} and
+     *       this {@link HttpHeadersBase} become immutable.</li>
```
anuraaga: Since this is a private class, it's probably ok but I was quite surprised to see the parameter become immutable

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1214 @@
...
+class HttpHeadersBase implements HttpHeaderGetters {
+
+    static final int DEFAULT_SIZE_HINT = 16;
+
+    /**
+     * Constant used to seed the hash code generation. Could be anything but this was borrowed from murmur3.
+     */
+    static final int HASH_CODE_SEED = 0xc2b2ae35;
+
+    private final HeaderEntry[] entries;
```
anuraaga: It could be an interesting future optimization if we can use something similar to an `EnumSet` when it's applicable, with just one each of commonly known header names. It should be very common.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1214 @@
...
+            }
+        }
+    }
+
+    final void setObject(CharSequence name, Object value) {
+        requireNonNull(value, "value");
+        set(name, fromObject(value));
+    }
+
+    final void setObject(CharSequence name, Iterable<?> values) {
```
anuraaga: Similar to the above comment, personal preference would be to use a different method name like `All` similar to the netty API to make the overload more precise.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1214 @@
...
+
+        return value;
+    }
+
+    private void ensureMutability() {
+        checkState(!immutable, "cannot be mutated anymore");
+    }
+
+    private static boolean keyEquals(@Nullable AsciiString a, CharSequence b) {
+        return a != null && (a == b || a.contentEqualsIgnoreCase(b));
```
anuraaga: Do you think `a == b` should be upstreamed into `AsciiString.contentEqualsIgnoreCase` itself?

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
trustin: I chose `CharSequence` because the setter methods accept `CharSequence` as a value. Internally, a `CharSequence` value is kept as it is and is converted to a `String` only when required. `defaultValue` is similar - when a header exists, no conversion is made.

After thinking a little bit more, I started to think this kind of optimization might be of dubious value. What do you think? Let me change all value types to `String` if you agree.

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
trustin: It is merely to accept both `AsciiString` and `String`. A header name is internally converted to `AsciiString` using `HttpHeaderNames.of()`. So, here, we accept the widest type and expose the narrowest type. I think a user will expect `AsciiString` or `String` rather than `CharSequence` because it lacks useful methods like comparison.

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
trustin: > Might be overthinking, but I always get nervous when the same method name has parameters that inherit from each other (I think many lint tools would complain about it too).

You mean it can be confusing to have `of()` methods in both `HttpHeaders` and its subtypes, or to have many parameters names `name` and `value`?

> Also personally, I think it's extremely rare to use non-string header values in most user code, so the helpers might not be used much to warrant the potential confusion.

I find our test cases use `Integer` or `MediaType` as a header value pretty often, so I think it's pretty useful?

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
trustin: Sounds good to me. Thanks for the reference.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: Yeah, let me add a comment.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: Yes. Please feel free to send a pull request if I don't beat you. :-)

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
trustin: I'd like to hide it from the public API by making it package-local but Javadoc doesn't seem to like it. Let me try a little bit harder to hide it before giving up. :-)

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
anuraaga: Sounds good

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
anuraaga: I was thinking about the fact that two methods with the same name take similar arguments 

`of(CharSequence, CharSequence)`
`of(CharSequence, Object)`

I personally almost always control click in intellij just to make sure the correct method was called or get nervous 

I think it relates to this point of effective java

http://thefinestartist.com/effective-java/41

trustin: I see. At worst it will be some performance loss due to using `addObject()` instead of `add()` and the end result will be same. Are you still concerned? Do you have a suggestion for a better name?

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
trustin: If we change the value type to `String`, do we perform an automatic conversion to `String` for the values? (which is fine by me.)

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: Agreed. Let me introduce a named factory method or something similar.

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
anuraaga: `ofObjects` would be clear and not ambiguous. I don't feel so strongly though if it seems obtuse

core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodingClient.java
trustin: `RequestContext` already has a method `setRequest()` which raises an exception when the original request is of a different type than that of the given one. What do you think about adding `trySetRequest()`? Perhaps we could add some general explanation about when to use it rather than `setRequest()` to the Javadoc of `trySetRequest()`.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: I personally like the current names because the parameter types are clearly different from each other - `CharSequence` vs `CharSequence...` vs `Iterable<CharSequence>` vs a single parameter `Iterable<Entry>`, but I'd like to make sure if there's a better option.

It's fairly easy for `setObject` because we can rename them to `setObjects`, but I have no idea about what to do about `set` because `sets()` doesn't make sense and neither `setAll()` nor `setAllObjects()` seem to explain what it does exactly. Any suggestions for better and consistent names?

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
trustin: Done.

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
trustin: Let me leave them as they are, then. It seems OK to me.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1205 @@
...
+
+    final long getTimeMillisAndRemove(CharSequence name, long defaultValue) {
+        final Long v = getTimeMillisAndRemove(name);
+        return v != null ? v : defaultValue;
+    }
+
+    final void add(CharSequence name, String value) {
+        final AsciiString normalizedName = normalizeName(name);
+        requireNonNull(value, "value");
+        checkNotNull(value, "value");
```
minwoox: Could remove this?

core/src/main/java/com/linecorp/armeria/common/RequestHeaders.java
```diff
@@ -0,0 +1,191 @@
...
+     *
+     * @throws IllegalArgumentException if the specified {@link HttpHeaders} does not have
+     *                                  {@code ":method"} or {@code ":path"} header.
+     */
+    static RequestHeaders of(HttpHeaders headers) {
+        if (headers instanceof RequestHeaders) {
+            return (RequestHeaders) headers;
+        }
+
+        checkArgument(headers.contains(HttpHeaderNames.METHOD), ":method header does not exist.");
```
minwoox: nit: null check for `headers`?

core/src/main/java/com/linecorp/armeria/common/ResponseHeaders.java
```diff
@@ -0,0 +1,194 @@
...
+     *
+     * @throws IllegalArgumentException if the specified {@link HttpHeaders} does not have
+     *                                  {@code ":status"} header.
+     */
+    static ResponseHeaders of(HttpHeaders headers) {
+        if (headers instanceof ResponseHeaders) {
+            return (ResponseHeaders) headers;
+        }
+
+        checkArgument(headers.contains(HttpHeaderNames.STATUS), ":status header does not exist.");
```
minwoox: nit: null check for `headers`?

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: Just removed the `immutable` field and made sure it is impossible to call the setter methods. As a bonus, a user can continue building after calling `build()`:

```java
HttpHeadersBuilder b = HttpHeaders.builder();
b.add("a", "b");
HttpHeaders h1 = b.build();
b.add("c", "d");
HttpHeaders h2 = b.build();

assert h1.toString().equals("{a=b}");
assert h2.toString().equals("{a=b, c=d}");
```

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: Oops :-D

core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodingClient.java
trustin: Renamed `setRequest()` to `updateRequest()` and made it return `boolean` instead of raising an exception on type mismatch.

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
trustin: Now hidden from the public API.

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -221,12 +222,14 @@ public boolean tryWrite(HttpObject o) {
                     logBuilder.startResponse();
 
                     assert o instanceof HttpHeaders;
-                    final HttpHeaders headers = (HttpHeaders) o;
-                    final HttpStatus status = headers.status();
 
-                    if (status != null && status.codeClass() != HttpStatusClass.INFORMATIONAL) {
-                        state = State.WAIT_DATA_OR_TRAILERS;
-                        logBuilder.responseHeaders(headers);
+                    if (o instanceof ResponseHeaders) {
```
anuraaga: Should we throw an exception if it's `RequestHeaders` here? I guess it's an armeria programming bug?

core/src/main/java/com/linecorp/armeria/common/DefaultRequestHeadersBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+        return getters != null ? getters.authority() : null;
+    }
+
+    @Override
+    public RequestHeadersBuilder authority(String authority) {
+        setters().authority(authority);
+        return this;
+    }
+
+    // Override the return type of the chaining methods in the superclass.
```
anuraaga: Not sure if it's strictly better, but a lot of these type of classes accept a template parameter in the base class with the subclass type, so the base class can cast to the template parameter type without having to do so in all children.

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
```diff
@@ -13,227 +13,156 @@
...
     }
 
     /**
-     * Returns new HTTP headers with a single entry.
+     * Returns a new {@link HttpHeaders} with the specified header. The value is converted into
+     * a {@link String} via {@link HttpHeadersBase#addObject(CharSequence, Object)}.
      */
-    static HttpHeaders of(AsciiString name, String value) {
-        return new DefaultHttpHeaders().add(name, value);
+    static HttpHeaders of(CharSequence name, Object value) {
```
anuraaga: What do you think about keeping the methods of type `AsciiString` and `@Deprecated` them?

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
anuraaga: Ah sorry I might have looked at an old version or just got confused, I thought the overrides were using `CharSequence` for the `value`. If not, then this is already best.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
anuraaga: `setAllObjects` is pretty clear to me, especially making clear that the value is iterated on. The semantics are pretty different between `setObject`, which converts the object to a string, and the iteration ones, which doesn't convert the value to a string but actually iterates on it.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1152 @@
...
+    /**
+     * Constant used to seed the hash code generation. Could be anything but this was borrowed from murmur3.
+     */
+    static final int HASH_CODE_SEED = 0xc2b2ae35;
+
+    // XXX(anuraaga): It could be an interesting future optimization if we can use something similar
+    //                to an EnumSet when it's applicable, with just one each of commonly known header names.
+    //                It should be very common.
+    @VisibleForTesting
+    final HeaderEntry[] entries;
```
anuraaga: Just wondering, did you consider guava's `Multimap`? I think this implementation is similar to a `LinkedListMultimap`. It might still be better to hand-implement, just want to make sure you're aware of it :)

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -115,6 +101,28 @@
         return null;
     }
 
+    /**
+     * Returns the {@link Request} associated with this context.
+     */
+    <T extends Request> T request();
+
+    /**
+     * Replaces the {@link Request} associated with this context with the specified one. This method is useful
```
anuraaga: Thanks!

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1152 @@
...
+        }
+
+        final HttpHeaderGetters that = (HttpHeaderGetters) o;
+        if (isEndOfStream() != that.isEndOfStream() ||
+            size() != that.size()) {
+            return false;
+        }
+
+        for (AsciiString name : names()) {
+            if (!getAll(name).equals(that.getAll(name))) {
```
hyangtack: Just curious. If we have two HttpHeaderBase instances as follows:
```
HttpHeaderBase headers1 = new HttpHeaderBase(16);
headers1.add("name", "value1");
headers1.add("name", "value2");
headers1.add("name", "value3");

HttpHeaderBase headers2 = new HttpHeaderBase(16);
headers2.add("name", "value3");
headers2.add("name", "value2");
headers2.add("name", "value1");
```
Should `headers1.equals(headers2)` return `true`? If so, the current code seems returning `false`.

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
trustin: Would it be worth doing that given that a user will have to recompile?

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: I knew about `Multimap` but I didn't use it because:

- We have to do a case-insensitive search for any `CharSequence` keys (both `AsciiString` and `String`)
- We cannot expose Guava API in our public API because we shade Guava.
  - Or extra allocation cost if we wrap a `MultiMap`.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
```diff
@@ -0,0 +1,1152 @@
...
+
+    private final class HeaderEntry implements Map.Entry<AsciiString, String> {
+
+        final int hash;
+        @Nullable
+        final AsciiString key;
+        @Nullable
+        final String value;
+        @Nullable
+        private String valueAsString;
```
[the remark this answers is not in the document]
trustin: Oops. Not really. Just a cruft from refactoring. :-)

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: I think it should return `false` because the header value order matters when the header name is same. Added test cases to `HttpHeadersBase`.

core/src/main/java/com/linecorp/armeria/common/DefaultRequestHeadersBuilder.java
trustin: That's a good idea. Fixed by adding `AbstractHttpHeadersBuilder`. I didn't touch the interfaces however because it looks prettier that way when Javadoc is rendered.

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
trustin: Yeah, now throwing `IllegalStateException`.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: I was actually asking about these methods:

- `set(Iterable<Map.Entry<>>)` and `setObject(Iterable<Map.Entry<>>)`
- `setIfAbsent(Iterable<Map.Entry<>>)`
- `add(Iterable<Map.Entry<>>)` and `addObject(Iterable<Map.Entry<>>)`

Do you think we can leave them as they are and just rename:

- `set/add(CharSequence name, String... values)` to `set/addAll()`
- `set/addObject(CharSequence name, Object... values)` to `set/addAllObjects()`

..?

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
anuraaga: Ah you're right, while I wanted to avoid having to recompile, it means picking the right method requires a cast which isn't good.

`NoSuchMethodError` at runtime is the scariest thing that can happen when upgrading a dependency, but guess there isn't much we can do here.

On the bright side, I guess most common uses of this class use the status or method variants and these might not be used so much in the wild.

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
trustin: Modified again to just use assertions.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
anuraaga: Yeah that seems fine

core/src/main/java/com/linecorp/armeria/common/RequestHeaders.java
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * Creates a new empty builder.
+     */
+    static RequestHeadersBuilder builder() {
+        return new DefaultRequestHeadersBuilder();
+    }
+
+    /**
+     * Creates a new builder with the specified {@link HttpMethod} and {@code path}.
```
anuraaga: I think there is inconsistency in javadoc among the classes e.g., `{@code ":path"}`, `{@code path}`. Would personally lean towards `{@code :path}` or `{@code path}` since quotes seems redundant when using at code.

core/src/main/java/com/linecorp/armeria/common/ResponseHeadersBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+     * Returns a newly created {@link ResponseHeaders} with the entries in this builder.
+     * Note that any further modification of this builder is prohibited after this method is invoked.
+     *
+     * @throws IllegalStateException if this builder does not have {@code ":status"} header set.
+     */
+    @Override
+    ResponseHeaders build();
+
+    /**
+     * Sets the the {@code ":status"} header.
```
anuraaga:
```suggestion
     * Sets the {@code ":status"} header.
```

core/src/main/java/com/linecorp/armeria/common/ResponseHeadersBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    ResponseHeadersBuilder status(int statusCode);
+
+    /**
+     * Sets the the {@code ":status"} header.
```
anuraaga:
```suggestion
     * Sets the {@code ":status"} header.
```

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -489,8 +515,19 @@ public static long parseCacheControlSeconds(@Nullable String value) {
     /**
      * Converts the specified Netty HTTP/2 into Armeria HTTP/2 headers.
      */
-    public static HttpHeaders toArmeria(Http2Headers headers, boolean endOfStream) {
-        final HttpHeaders converted = new DefaultHttpHeaders(false, headers.size(), endOfStream);
+    public static HttpHeaders toArmeria(Http2Headers headers, boolean request, boolean endOfStream) {
+        final HttpHeadersBuilder converted;
+        if (request) {
+            converted = headers.contains(HttpHeaderNames.METHOD) ? RequestHeaders.builder()
```
anuraaga: Might have missed it, but I don't think there are new unit tests for these conditionals. Should we add?

core/src/main/java/com/linecorp/armeria/common/RequestHeaders.java
trustin: `":path"` refers to the HTTP/2 `:path` pseudo header and `path` refers to the parameter `path`.

trustin: So, we wrap header names with double quotes and do not for parameter names, so I think `":path"` and `path` are consistent.

anuraaga: Makes sense - I got confused since they're both together in this PR, but when reading documentation as normal it wouldn't be an issue :)

core/src/main/java/com/linecorp/armeria/common/ResponseHeadersBuilder.java
trustin: Thanks!

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
trustin: Added a test case. :smile:

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
trustin: I tried renaming but I somehow don't like it.. I'd like to keep it as it is unless you are strong about this one. :pray:

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBuilder.java
```diff
@@ -0,0 +1,509 @@
...
+     *
+     * @param name the name of the header to retrieve
+     * @return a {@link List} of header values or an empty {@link List} if no values are found.
+     */
+    List<String> getAllAndRemove(CharSequence name);
+
+    /**
+     * Removes all the headers with the specified name and returns the header value which was added first.
+     *
+     * @param name the name of the header to search
```
anuraaga: Above uses `retrieve` here users `search`. I think using `retrieve` or `get` throughout is better, `search` feels like a more heavyweight operation

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBuilder.java
```diff
@@ -0,0 +1,509 @@
...
+     * Adds all header names and values of the specified {@code headers}.
+     *
+     * @return {@code this}
+     * @throws IllegalArgumentException if {@code headers == this}.
+     */
+    HttpHeadersBuilder add(
+            Iterable<? extends Entry<? extends CharSequence, String>> headers);
+
+    /**
+     * Adds a new header. Before the {@code value} is added, it's converted to a {@link String}.
```
anuraaga: Can you link to the string conversion logic? I think it can help users understand what will happen

core/src/main/java/com/linecorp/armeria/common/RequestHeaders.java
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Immutable HTTP/2 headers for an {@link HttpRequest}.
+ *
+ * @see ResponseHeaders
+ */
+public interface RequestHeaders extends HttpHeaders, RequestHeaderGetters {
+
+    /**
+     * Creates a new empty builder.
```
anuraaga: Guess this and below should be `Returns` for consistency with the other ones

core/src/main/java/com/linecorp/armeria/common/ResponseHeaders.java
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * Immutable HTTP/2 headers for an {@link HttpResponse}.
+ *
+ * @see RequestHeaders
+ */
+public interface ResponseHeaders extends HttpHeaders, ResponseHeaderGetters {
+
+    /**
+     * Creates a new empty builder.
```
anuraaga: `Returns` throughout

core/src/main/java/com/linecorp/armeria/common/ResponseHeadersBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * Builds a {@link ResponseHeaders}.
+ *
+ * @see ResponseHeaders#builder()
+ * @see ResponseHeaders#toBuilder()
+ */
+public interface ResponseHeadersBuilder extends HttpHeadersBuilder, ResponseHeaderGetters {
+    /**
+     * Returns a newly created {@link ResponseHeaders} with the entries in this builder.
+     * Note that any further modification of this builder is prohibited after this method is invoked.
```
anuraaga: Ditto

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
```diff
@@ -0,0 +1,309 @@
...
+
+import io.netty.util.AsciiString;
+
+/**
+ * Provides the getter methods to {@link HttpHeaders} and {@link HttpHeadersBuilder}.
+ */
+interface HttpHeaderGetters extends Iterable<Entry<AsciiString, String>> {
+
+    /**
+     * Tells whether the headers will be the last frame in an HTTP/2 stream.
```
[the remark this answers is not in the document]
trustin: Thanks.

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     * Returns {@code true} if a header with the {@code name} exists, {@code false} otherwise.
+     *
+     * @param name the header name
+     */
+    boolean contains(CharSequence name);
+
+    /**
+     * Returns {@code true} if a header with the {@code name} and {@code value} exists.
+     *  @param name the header name
```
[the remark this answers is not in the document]
trustin: Oops! :sweat_smile:

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBuilder.java
```diff
@@ -0,0 +1,509 @@
...
+/**
+ * Builds an {@link HttpHeaders}.
+ *
+ * @see HttpHeaders#builder()
+ * @see HttpHeaders#toBuilder()
+ */
+public interface HttpHeadersBuilder extends HttpHeaderGetters {
+    /**
+     * Returns a newly created {@link HttpHeaders} with the entries in this builder.
+     * Note that any further modification of this builder is prohibited after this method is invoked.
```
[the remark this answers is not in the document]
trustin: That restriction has been removed some time ago. Forgot to update the Javadoc. :sweat_smile:

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBuilder.java
trustin: All updated to a simplified version:

```
     * @param name the header name
```

core/src/main/java/com/linecorp/armeria/common/RequestHeaders.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/HttpHeaders.java
```diff
@@ -13,227 +13,156 @@
...
      */
-    static HttpHeaders of(HttpMethod method, String path) {
-        return new DefaultHttpHeaders().method(method).path(path);
-    }
+    @Deprecated
+    HttpHeaders EMPTY_HEADERS = of();
 
     /**
-     * Returns new HTTP response headers.
+     * Returns an empty singleton {@link HttpHeaders}.
```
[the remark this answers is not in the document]
trustin: Removed 'singleton'

core/src/main/java/com/linecorp/armeria/common/HttpHeaderGetters.java
```diff
@@ -0,0 +1,309 @@
...
+     */
+    boolean isEmpty();
+
+    /**
+     * Returns a {@link Set} of all header names. The returned {@link Set} cannot be modified.
+     */
+    Set<AsciiString> names();
+
+    /**
+     * Returns an iterator that yields all header entries. The iteration order is as follows:
```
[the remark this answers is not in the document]
trustin: Agreed and fixed.

core/src/main/java/com/linecorp/armeria/common/HttpHeadersBuilder.java
anuraaga: By the way, I think this can just be `@see StringValueConverter` - lucky that's in `common` :D

https://github.com/line/armeria/pull/1731/files#diff-71e30b4cf4ff10735fc468087527be95

core/src/main/java/com/linecorp/armeria/server/file/AbstractHttpFile.java
```diff
@@ -141,24 +144,26 @@ protected final String generateEntityTag(HttpFileAttributes attrs) {
...
             return null;
         }
 
         // TODO(trustin): Cache the headers (sans the 'date' header') if attrs did not change.
         final String etag = generateEntityTag(attrs);
-        final HttpHeaders headers = HttpHeaders.of(HttpStatus.OK);
-        headers.set(HttpHeaderNames.CONTENT_LENGTH, Long.toString(attrs.length()));
+        final ResponseHeadersBuilder headers =
+                ResponseHeaders.builder(HttpStatus.OK)
+                               .set(HttpHeaderNames.CONTENT_LENGTH, Long.toString(attrs.length()));
```
hyangtack: nit: set -> setLong?

core/src/main/java/com/linecorp/armeria/common/HttpHeadersJsonDeserializer.java
```diff
@@ -37,6 +37,8 @@
  */
 public final class HttpHeadersJsonDeserializer extends StdDeserializer<HttpHeaders> {
 
+    // FIXME(trustin): Add Request/ResponseHeadersJsonDeserializer.
```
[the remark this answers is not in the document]
trustin: - Extracted deserialization logic from `HttpHeadersJsonDeserializer` to `AbstractHttpHeadersJsonDeserializer`.
- Added `RequestHeadersJsonDeserializer`, `ResponseHeadersJsonDeserializer` and their tests.

anuraaga:
Thanks a lot!

core/src/main/java/com/linecorp/armeria/common/AbstractHttpHeadersJsonDeserializer.java
```diff
@@ -0,0 +1,101 @@
...
+            return headers;
+        } catch (IllegalStateException e) {
+            return ctx.reportInputMismatch(handledType(),
+                                           firstNonNull(e.getMessage(), "Required headers are missing."));
+        }
+    }
+
+    private void addHeader(DeserializationContext ctx, HttpHeadersBuilder builder,
+                           AsciiString name, JsonNode valueNode) throws JsonMappingException {
+        if (!valueNode.isTextual()) {
```
hyangtack: nit: How about changing here as follows for code readability?
```
if (valueNode.isTextual()) {
  builder.add(name, valueNode.asText());
} else {
  ctx.reportInputMismatch(...)
}
```

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -77,32 +78,35 @@ static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
...
     static HttpResponse of(HttpStatus status) {
         requireNonNull(status, "status");
-        if (status.codeClass() == HttpStatusClass.INFORMATIONAL) {
-            final HttpResponseWriter res = streaming();
-            res.write(HttpHeaders.of(status));
-            return res;
-        } else if (isContentAlwaysEmpty(status)) {
-            return new OneElementFixedHttpResponse(HttpHeaders.of(status));
+        checkArgument(status.codeClass() != HttpStatusClass.INFORMATIONAL,
+                      "statue: %s (expected: a non-1xx status");
```
minwoox: nit: statue -> status

trustin: Doh! Fixed. :-)

core/src/main/java/com/linecorp/armeria/common/AbstractHttpHeadersJsonDeserializer.java
trustin: Updated :smile:

trustin:
Thank you, reviewers! It was very helpful. :-)

PR #1736, opened 2019-04-23
Support JUnit 5

delegacy:
Motivation:

Some users want to run their unit tests with JUnit 5,
but they might feel bad when they use `ServerRule` (of JUnit 4) and run
their tests with `junit-vintage-engine`.

Modifications:

- Add `ServerExtension` and apply it to some tests
- Add `SelfSignedCertificateExtension` and apply it to some tests
- Add `EventLoopExtension` and apply it to some tests
- Add `EventLoopGroupExtension`


delegacy:
If this change is acceptable, I want to continue the similar for other rules (`SelfSignedCertificateRule`, ...)
- Add classes for JUnit 5 and apply them to some existing tests

Some concerns what I have for now are:

- How to publish `testing` jars (one jar both for JUnit 4 and JUnit 5 vs. separated jars each)
- How to migrate all existing unit tests to JUnit 5

dependencies.yml
```diff
@@ -233,6 +233,18 @@ junit:
     javadocs:
     - https://junit.org/junit4/javadoc/4.12/
 
+org.junit.jupiter:
+  junit-jupiter-api:
+    version: &JUNIT_JUPITER_VERSION '5.4.2'
```
trustin: Could you add Javadoc links?

gradle/scripts/lib/java.gradle
```diff
@@ -108,6 +108,8 @@ configure(projectsWithFlags('java')) {
                 dependsOn: [tasks.checkstyleMain, tasks.checkstyleTest])
 
         test {
+            useJUnitPlatform()
```
trustin: I think this has to be done in `/build.gradle` for all Java projects instead of here, because `gradle/scripts` is shared with other projects.

trustin:
> How to publish testing jars (one jar both for JUnit 4 and JUnit 5 vs. separated jars each)

How about splitting `armeria-testing` to:

- `armeria-testing` for the common classes that are agnostic to testing frameworks
- `armeria-testing-junit` for JUnit 5
- `armeria-testing-junit4` for JUnit 4

..?

> How to migrate all existing unit tests to JUnit 5

I don't have a good answer for this but perhaps time will tell? :sweat_smile:

delegacy:
> How about splitting armeria-testing to:

How about this?
`armeria-testing-common` for common, `armeria-testing-junit` for JUnit 5, `armeria-testing-junit4` for JUnit 4.

> I don't have a good answer for this but perhaps time will tell? 😅

Then, I will apply `Extension`s to a few test classes just for verification of new implementation.

delegacy:
> How about moving all delegate classes to `.internal` packages? They don't need to be part of the public API.

Do you mean moving delegate classes into `testing-internal` subproject?

trustin:
> Do you mean moving delegate classes into `testing-internal` subproject?

No. We can't do that because `testing-internal` is not published. I meant just renaming packages, i.e. `c.l.a.testing.common` to `c.l.a.testing.internal`.

delegacy:
> No. We can't do that because `testing-internal` is not published. I meant just renaming packages, i.e. `c.l.a.testing.common` to `c.l.a.testing.internal`.

Done. Ready for a next round.

trustin:
How about rename packages like the following:

- `...{junit,junit4}.{common,server}` -> `{common,server}.{junit,junit4}`

..? I want to make sure our package structure is consistent, so it'd be nice the new package structure resembles that of the core:

- `...client.{circuitbreaker,encoding,...}
- `...common.{logging,metric,...}
- `...internal.{annotation,crypto,...}
- `...server.{annotation,auth,...}

I guess this is the last comment from me. Great job!

trustin:
I realized that the previous package structure was better and asked him to revert the last package name changes. Sorry about the confusion!

delegacy:
No problem. Please take a look again.

testing-internal/build.gradle
```diff
@@ -1,6 +1,8 @@
 dependencies {
-    compile project(':testing')
+    compile project(':testing:junit')
+    compile project(':testing:junit4')
 
+    compile 'org.junit.jupiter:junit-jupiter-api'
```
hyangtack: nit: `junit-jupiter-api` and `junit` seem unnecessary. How about removing them if so?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/SelfSignedCertificateExtension.java
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Creates a new instance.
+     *
+     * @deprecated Use {@link #SelfSignedCertificateExtension(TemporalAccessor, TemporalAccessor)}.
+     *
+     * @param notBefore {@link Certificate} is not valid before this time
+     * @param notAfter {@link Certificate} is not valid after this time
+     */
+    @Deprecated
```
hyangtack: Do we really need to leave deprecated constructors? This class is newly added for junit5, so it maybe okay to remove them. What do you think? @delegacy @line/cdtf

trustin: +1 for removal

delegacy:
To reviewers, addressed the comments.

Travis CI build keeps failing, but it seems not related with my changes.
Please take a look again.

trustin:
Great job, @delegacy! :+1:

delegacy:
Thanks all!

PR #1737, opened 2019-04-23
Provide a way to configure per-service settings fluently

### LINKED ISSUE #1619: Provide a way to configure content previewing to the specific service
Currently, a user can configure content previewing to the whole server using `ServerBuilder`. A user might want to enable content previewing to only the HTTP services when he or she has Thrift and HTTP services together.

### LINKED ISSUE #1649: More per-service settings with fluent route mapping.
As shown in #1619, we are getting increasing needs for specifying different settings for different services. To address that, we need to provide a way to build a service configuration more fluently.

Here's the initial idea for the new API for registering a service fluently:

```java
sb.route().get("/greet/{name}")
          .consumes(MediaType.JSON)
          .produces(MediaType.JSON_UTF_8)
          .contentPreview(..)
          .verboseResponses(true)
          .responseTimeout(...)
          .service(myService) // Terminates service construction
  .route().path("glob:/foo/*/baz")
          .methods(HttpMethod.GET, HttpMethod.POST)
          .service(myService2)
  .route().path("prefix:/alice")
          .service(myService3) // Unspecified methods means 'all methods'.
  .route().get("/a/b")
          .post("/a/b/c")
          .service(myService4) // A service is bound to two path/method pairs.
  ...
```

minwoox:
Motivations:

- Per-service setting
  - It will be nice if we have an API to configure per-service settings fluently. See #1649, #1619
- Removing `VirtualHost` setters in `ServerBuilder`
  - We could set up a `VirtualHost` in two ways using:
    - `ChainedVirtualHostBuilder` that provides fluent APIs.
    - The setter which takes a `VirtualHost` already built as an argument.
  - Our users can choose to suit their taste for building `VirtualHost`s, however, it made the code more complex.

Modifications:
- Per-service setting
  - Add `AbstractRouteBuilder` and it's children which are `RouteBuilder` and `VirtualHostRouteBuilder`.
  - Add package-local `ServiceConfigBuilder` and change the constructor of `ServiceConfig` as package-local.
    - You have to use `ServerBuilder` to create `ServiceConfig` in tests.
- Removing `VirtualHost` setters in `ServerBuilder`
  - Breaking changes
    - Remove setters in `ServerBuilder` which are `defaultVirtualHost(virtualHost)` and `virtualHost(virtualHost)`.
    - Remove `VirtualHostBuilder` and integrate `AbstractVirtualHostBuilder` and `ChainedVirtualHostBuilder` into one:
      - Name the integrated one as `VirtualHostBuilder`.
  - Deprecated
    - `ServerBuilder.defaultRequestTimeoutMillis()`. Use `ServerBuilder.requestTimeoutMillis()`.
    - `ServerBuilder.defaultMaxRequestLength()`. Use `ServerBuilder.maxRequestLength()`.
    - `ClientBuilder.defaultResponseTimeoutMillis()`. Use `ServerBuilder.responseTimeoutMillis()`.
    - `ClientBuilder.defaultMaxResponseLength()`. Use `ServerBuilder.maxResponseLength()`.

Result:
- Close #1649 and close #1619
- Hopefully, the less complex code in `ServerBuilder`.
- You can now configure per-service settings fluently.

To-dos:
- Add documentation.
- Add another building method with `ServiceWithPathMappings` to `RoutBuilder`. (I'm not sure if it's a good idea though.)
- Add `RouteBuilder.apply(Consumer<RouteBuilder>)`( which might be used, for example, to set the `POST` method and `produceTypes` for Thrift automatically).


core/src/main/java/com/linecorp/armeria/server/PathMappingContext.java
```diff
@@ -68,12 +69,11 @@
...
-     * Returns a list of {@link MediaType} that is ordered by client-side preferences.
-     * It only includes the media types supported by the virtual host. {@code null} would be returned
-     * if the virtual host does not support media type negotiation.
+     * Returns a list of {@link MediaType}s that are specified in {@link HttpHeaderNames#ACCEPT} in the order
+     * of client-side preferences. If the client does not send the header, this will contain only
+     * {@link MediaType#ANY_TYPE}.
      */
-    @Nullable
-    List<MediaType> produceTypes();
+    List<MediaType> acceptTypes();
```
hyangtack: I think that it's good try to change this name from a `Request`'s point of view. Don't we need to consider changing `consumeType` to something like `contentType` as well?

minwoox: That's a good suggestion. I was sure that `acceptTypes` is better naming because I changed the logic which purely extracts the content types in the `Accept` header but was not sure that I could change `consumeType`. Let me rename it. Thanks!

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -146,6 +147,12 @@ public PathMappingResult apply(PathMappingContext mappingCtx) {
         return PathMappingResult.empty();
     }
 
+    // TODO(minwoox) Move this method into MediaType?
+    private static boolean isAnyType(MediaType contentType) {
+        // Ignores the quality factor.
```
hyangtack: nit: `Ignores all parameters including the quality factor`? Because there's a method `MediaType.belongsTo()` which ignores the quality factor only.

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -146,6 +147,12 @@ public PathMappingResult apply(PathMappingContext mappingCtx) {
         return PathMappingResult.empty();
     }
 
+    // TODO(minwoox) Move this method into MediaType?
```
trustin: If performance does not matter, you could use `contentType.is(MediaType.ANY_TYPE)`.

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * A builder class for building a {@link Service} fluently. This class can only be created through
+ * {@link ServerBuilder#route()}.
+ *
+ * <p>Call {@link #service(Service)} to build the {@link Service} and return to the {@link ServerBuilder}.
+ *
+ * <pre>{@code
+ * ServerBuilder sb = new ServerBuilder();
+ * sb.route().get("/foo/bar")                      // Configure the first service.
+ *   .consumes(JSON, PLAIN_TEXT_UTF_8)
```
trustin: Perhaps we could indent at `.get()` to highlight the level?

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+ *   .produces(JSON_UTF_8, PLAIN_TEXT_UTF_8)
+ *   .requestTimeoutMillis(10)
+ *   .maxRequestLength(8192)
+ *   .verboseResponses(true)
+ *   .contentPreview(500)
+ *   .service((ctx, req) -> HttpResponse.of(OK))   // Return to the ServerBuilder.
+ *   .http(8080)                                   // ServerBuilder
```
trustin: I understand you put this to demonstrate that we are at `ServerBuilder`, but it seems impractical to me. How about removing this line?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -858,20 +768,22 @@ public ServerBuilder sslContext(
...
             SessionProtocol protocol,
             File keyCertChainFile, File keyFile, String keyPassword) throws SSLException {
-
-        defaultVirtualHostBuilderUpdated();
         defaultVirtualHostBuilder.sslContext(protocol, keyCertChainFile, keyFile, keyPassword);
         return this;
     }
 
+    /**
+     * Returns a {@link RouteBuilder} which is for building a {@link Service} fluently.
```
trustin: `... for binding a {@link Service} fluently.` ?

core/src/test/java/com/linecorp/armeria/server/PathMappingTest.java
```diff
@@ -67,6 +72,26 @@ public void successfulOf() {
...
+        assertThat(pathMappingWithInfo.supportedMethods()).containsExactly(HttpMethod.GET);
+
+        assertThatThrownBy(
+                () -> pathMappingWithInfo.withHttpHeaderInfo(ImmutableSet.of(HttpMethod.GET),
+                                                             ImmutableList.of(), ImmutableList.of()))
+                .isInstanceOf(IllegalArgumentException.class);
+
+        assertThatThrownBy(
+                () -> m.withHttpHeaderInfo(ImmutableSet.of(),
+                                                             ImmutableList.of(), ImmutableList.of()))
```
trustin: nit: could be merged into a single line.

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,522 @@
...
+    }
+
+    /**
+     * Sets the path pattern and {@link HttpMethod#GET} that the {@linkplain #service(Service) service} is
+     * bound to and supports.
+     *
+     * @throws IllegalArgumentException if the specified path pattern is invalid
+     */
+    public B get(String pathPattern) {
+        addPathMapping(pathPattern, EnumSet.of(GET));
```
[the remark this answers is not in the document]
minwoox: It's only used one time so probably not worth it. Let me leave it as it is. :-)

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
minwoox: I needed to check if the `contentType` is the `ANY_TYPE`.
`contentType.is(MediaType.ANY_TYPE)` returns `true` for all content type. :-)

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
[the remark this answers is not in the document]
hyangtack: Ok, it makes sense. :-)

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,522 @@
...
+     * The {@link Service} is found when the path of a request matches the {@link PathMapping} and the
+     * {@link HttpMethod} of the request is one of the {@link HttpMethod}s set by
+     * {@link #methods(HttpMethod...)}.
+     *
+     * @throws IllegalArgumentException if the {@link PathMapping} is wrapped by
+     *                                  {@link PathMapping#withHttpHeaderInfo(Set, List, List)}
+     */
+    public B pathMapping(PathMapping pathMapping) {
+        if (pathMapping instanceof HttpHeaderPathMapping) {
+            throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
minwoox: Fixed, thanks!

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
minwoox: Fixed, thanks!

core/src/test/java/com/linecorp/armeria/server/PathMappingTest.java
minwoox: Oops!

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+/**
+ * A builder class for binding a {@link Service} fluently.
+ *
+ * @param <B> the self type
+ * @param <V> {@link ServerBuilder} or {@link VirtualHostBuilder} that you create this builder from
+ *
+ * @see RouteBuilder
+ * @see VirtualHostRouteBuilder
+ */
+public abstract class AbstractRouteBuilder<B extends AbstractRouteBuilder<B, V>, V> {
```
trustin: I'm curious if we can make this class package-local. Perhaps it's possible if the Javadocs of the subclasses are rendered correctly?

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+    @Nullable
+    private Function<Service<HttpRequest, HttpResponse>, Service<HttpRequest, HttpResponse>> decorator;
+
+    @SuppressWarnings("unchecked")
+    final B self() {
+        return (B) this;
+    }
+
+    /**
+     * Sets the path pattern that the {@linkplain #service(Service) service} is bound to. The {@link Service}
```
trustin: - the service -> `a {@link Service}` ?
- is bound to -> will be bound to ?

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+    public B consumes(MediaType... consumeTypes) {
+        consumes(ImmutableList.copyOf(requireNonNull(consumeTypes, "consumeTypes")));
+        return self();
+    }
+
+    /**
+     * Sets {@link MediaType}s that the {@link Service} consumes. If not set,
+     * the {@linkplain #service(Service) service} accepts all media types.
+     */
+    public B consumes(List<MediaType> consumeTypes) {
```
trustin: `Iterable<MediaType>`?

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+        } else {
+            this.decorator = castDecorator;
+        }
+
+        return self();
+    }
+
+    /**
+     * Sets the {@link Service} and returns the {@link V} that you create this
+     * {@link B} from.
```
trustin: that this builder was created from?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -52,8 +52,7 @@
  * @see ChainedCorsPolicyBuilder
  * @see CorsPolicyBuilder
  */
-@SuppressWarnings("rawtypes")
-abstract class AbstractCorsPolicyBuilder<B extends AbstractCorsPolicyBuilder> {
+abstract class AbstractCorsPolicyBuilder<B extends AbstractCorsPolicyBuilder<B>> {
```
trustin: Thanks :-)

trustin:
How about updating the examples in `server-basics.rst` to leverage the new API? I guess we could use the new API for all non-RPC service examples.

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -66,10 +66,13 @@
  */
 public abstract class AbstractRouteBuilder<B extends AbstractRouteBuilder<B, V>, V> {
 
+    private static final EnumSet<HttpMethod> defaultMethods =
+            EnumSet.of(OPTIONS, GET, HEAD, POST, PUT, PATCH, DELETE, TRACE, CONNECT);
```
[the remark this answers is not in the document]
minwoox: Fixed, Thanks!

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
minwoox: Fixed, thanks!

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -125,16 +128,17 @@ public B pathUnder(String prefix) {
...
-            throw new IllegalArgumentException(
-                    "cannot set a PathMapping which is wrapped by PathMapping.withHttpHeaderInfo(): " +
-                    pathMapping);
-        }
-        defaultPathMapping = requireNonNull(pathMapping, "pathMapping");
+        requireNonNull(pathMapping, "pathMapping");
+        checkArgument(pathMapping.hasPathPatternOnly(),
+                      "pathMapping: %s " +
+                      "(expected: the path mapping which has only the path patterns as its condition)",
+                      pathMapping.getClass().getSimpleName());
```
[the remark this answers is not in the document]
minwoox: Oops~

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
minwoox: 😄

core/src/main/java/com/linecorp/armeria/server/VirtualHostRouteBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.function.Function;
+
+import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+
+/**
+ * A builder class for binding a {@link Service} fluently. This class can only be created through
```
trustin: created -> instantiated

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.function.Function;
+
+import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+
+/**
+ * A builder class for binding a {@link Service} fluently. This class can only be created through
```
trustin: created -> instantiated

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -16,44 +16,830 @@
...
- * <h2>Example</h2>
- * <pre>{@code
- * VirtualHostBuilder vhb = new VirtualHostBuilder("*.example.com");
- * vhb.service("/foo", new FooService())
- *    .serviceUnder("/bar/", new BarService())
- *    .service(PathMapping.ofRegex("^/baz/.*", new BazService());
  *
- * VirtualHost vh = vhb.build();
- * }</pre>
+ * <p>This class can only be created through the {@link ServerBuilder#withDefaultVirtualHost()} or
```
trustin: created -> instantiated

core/src/main/java/com/linecorp/armeria/server/VirtualHostRouteBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+ * {@link VirtualHostBuilder#route()}.
+ *
+ * <p>Call {@link #service(Service)} to build the {@link Service} and return to the {@link VirtualHostBuilder}.
+ *
+ * <pre>{@code
+ * ServerBuilder sb = new ServerBuilder();
+ * sb.withVirtualHost("example.com")
+ *   .route().get("/foo/bar")                              // Configure the first service.
```
trustin: Could you update this example to show how to configure multiple virtual hosts?

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,509 @@
...
+    /**
+     * Sets the {@link Service} and returns the {@link B} that this builder was created from.
+     *
+     * @throws IllegalStateException if the path that the {@link Service} will be bound to is not specified
+     *
+     * @see #path(String)
+     * @see #pathUnder(String)
+     * @see #pathMapping(PathMapping)
+     */
+    public abstract B service(Service<HttpRequest, HttpResponse> service);
```
trustin: I guess you can just return `Object` and remove the type parameter `B` completely.

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -57,6 +58,7 @@
                           List<MediaType> consumeTypes, List<MediaType> produceTypes) {
         this.pathStringMapping = requireNonNull(pathStringMapping, "pathStringMapping");
         this.supportedMethods = requireNonNull(supportedMethods, "supportedMethods");
+        checkArgument(!supportedMethods.isEmpty(), "supportedMethod can't be empty");
```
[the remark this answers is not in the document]
minwoox: Oops, fixed :-)

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,490 @@
...
+    /**
+     * Sets the {@link HttpMethod}s that a {@link Service} will support.
+     * If not set, all {@link HttpMethod}s except {@link HttpMethod#UNKNOWN} are set.
+     *
+     * @see #path(String)
+     * @see #pathUnder(String)
+     * @see #pathMapping(PathMapping)
+     */
+    public AbstractRouteBuilder methods(HttpMethod... methods) {
+        methods(EnumSet.copyOf(Arrays.asList(methods)));
```
trustin: Could be:

```java
    return methods(ImmutableSet.copyOf(requireNonNull(methods, "methods")));
```

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,490 @@
...
+
+    /**
+     * Sets the {@link HttpMethod}s that a {@link Service} will support. If not set,
+     * all {@link HttpMethod}s except {@link HttpMethod#UNKNOWN} are set.
+     *
+     * @see #path(String)
+     * @see #pathUnder(String)
+     * @see #pathMapping(PathMapping)
+     */
+    public AbstractRouteBuilder methods(Set<HttpMethod> methods) {
```
trustin: `Iterable<HttpMethod>`

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -57,6 +58,7 @@
                           List<MediaType> consumeTypes, List<MediaType> produceTypes) {
         this.pathStringMapping = requireNonNull(pathStringMapping, "pathStringMapping");
         this.supportedMethods = requireNonNull(supportedMethods, "supportedMethods");
+        checkArgument(!supportedMethods.isEmpty(), "supportedMethods can't be empty");
         this.consumeTypes = requireNonNull(consumeTypes, "consumeTypes");
         this.produceTypes = requireNonNull(produceTypes, "produceTypes");
```
trustin: How about making sure `supportedMethods`, `consumeTypes` and `produceTypes` are all immutable, like using `ImmutableList.copyOf()` or `ImmutableSet.copyOf()`? I guess it will be harder to make sure immutability otherwise.

core/src/test/java/com/linecorp/armeria/server/VirtualHostRouterBuilderTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.List;
+
+import org.junit.Test;
+
+import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+
+public class VirtualHostRouterBuilderTest {
```
trustin: `RouteBuilderTest` (r)

minwoox: Oooooops~

core/src/main/java/com/linecorp/armeria/server/HttpHeaderPathMapping.java
```diff
@@ -57,10 +58,10 @@
     HttpHeaderPathMapping(PathMapping pathStringMapping, Set<HttpMethod> supportedMethods,
                           List<MediaType> consumeTypes, List<MediaType> produceTypes) {
         this.pathStringMapping = requireNonNull(pathStringMapping, "pathStringMapping");
-        this.supportedMethods = requireNonNull(supportedMethods, "supportedMethods");
+        this.supportedMethods = ImmutableSet.copyOf(requireNonNull(supportedMethods, "supportedMethods"));
```
trustin: There has to be a way to make sure that the resulting `Set` is `ImmutableEnumSet`. How about converting `supportedMethods` to `EnumSet` first? e.g.

```java
this.supportedMethods = ImmutableSet.copyOf(EnumSet.copyOf(requireNonNull(supportedMethods, "supportedMethods")));
```

trustin: Perhaps we could stop sweating about keeping `AbstractRouteBuilder.supportedMethods` is an `EnumSet` or an `ImmutableEnumSet` once we make this change here.

core/src/main/java/com/linecorp/armeria/server/AbstractRouteBuilder.java
```diff
@@ -0,0 +1,488 @@
...
+/**
+ * A builder class for binding a {@link Service} fluently.
+ *
+ * @see RouteBuilder
+ * @see VirtualHostRouteBuilder
+ */
+abstract class AbstractRouteBuilder {
+
+    private static final Set<HttpMethod> defaultMethods =
+            ImmutableSet.copyOf(EnumSet.of(OPTIONS, GET, HEAD, POST, PUT, PATCH, DELETE, TRACE, CONNECT));
```
trustin: `ImmutableSet.copyOf(EnumSet.allOf(HttpMethod.class))`?

trustin:
Please fix the compilation errors and address the last comment, then we're good to go!

trustin:
@anuraaga Any comments just in case you are interested? :-)

core/src/main/java/com/linecorp/armeria/common/HttpMethod.java
```diff
@@ -103,6 +108,14 @@
      */
     UNKNOWN;
 
+    private static final Set<HttpMethod> knownMethods;
+
+    static {
+        final Set<HttpMethod> allMethods = EnumSet.allOf(HttpMethod.class);
+        allMethods.remove(UNKNOWN);
+        knownMethods = ImmutableSet.copyOf(allMethods);
```
anuraaga: Sets.immutableEnumSet

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * A builder class for binding a {@link Service} fluently. This class can only be instantiated through
+ * {@link ServerBuilder#route()}.
+ *
+ * <p>Call {@link #service(Service)} to build the {@link Service} and return to the {@link ServerBuilder}.
+ *
+ * <pre>{@code
+ * ServerBuilder sb = new ServerBuilder();
+ * sb.route().get("/foo/bar")                              // Configure the first service.
```
anuraaga: Instead of `route` returning the builder what do you think about it taking a parameter `Consumer<Route builder>`?

 I think it's similarly small syntax but will work with most code formatting (current style won't work for e.g. google Java format I think) and it is a bit magical that `service` switches back to a `ServerBuilder`.

core/src/main/java/com/linecorp/armeria/common/HttpMethod.java
minwoox: Thank, will fix in my next PR. :-)

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
minwoox: I thought to return the `ServerBuilder` was ok since we are already doing that in the `VirtualHostBuilder`. 
I was actually going to add a method something like `RouteBuilder.apply(Consumer<RouteBuilder>)` later, but you think it's better to add the `Consumer` to `route()` so that the `RouteBuilder` does not return `ServerBuilder`?
And the style that you mentioned is that the `.consumes()` is aligned with `.get()` in the example?

trustin:
:tada: Nice job, @minwoox!

minwoox:
Thanks a lot, for reviewing. 😉  Will send the follow-up PR soon.

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
anuraaga: For the style, I'm specifically referring to `route` below which lines up with the above `route`. Automatic code formatters would align `route` with `service` above it I think. I think with the suggested `apply`, it would be similar. Having route take a consumer would avoid it since the parameter would be indented and all calls to route would be aligned. 

I don't know if a consumer API has to be the only one but it would be much easier for me since I use automatic Google format for whitespace

minwoox: @anuraaga, thanks for the point!
It was not a good idea that I made the example which violates our code style to just make it look prettier. I will fix the example like:
```java
sb.route()
  .get("/foo/bar")
  .consumes(JSON, PLAIN_TEXT_UTF_8)
  ...
```

I thought about your suggestion, and it was great to have `ServerBuilder.route(Consumer<RouteBuilder>)`. But I thought it again, and just a little bit worried about having `routeBuilder` in the code.
The code might be like:
```java
sb.route(routeBuilder -> {
    routeBuilder.consumes(JSON, PLAIN_TEXT_UTF_8)
                .produces(JSON_UTF_8, PLAIN_TEXT_UTF_8)
                .requestTimeoutMillis(10)
                .....
});
```
So I think it's probably better to keep as it is.
And it's weird that `RouteBuilder.service(service)` returns the `ServerBuilder`  which the `RouteBuilder` instantiated from, as you pointed out, so perhaps, if I change the name to `RouteBuilder.build(service)`, it could be better I think.

The API `RouteBuilder.apply(Consumer<RouteBuilder>)` which I was going to add was to provide convenient APIs for gRPC and Thrift services which handles `Content-type` and `Accept` headers.
Because the `RouteBuilder` wraps a `PathMapping` with the method `withHttpHeadersInfo()`, I thought it would provide useful methods which set the `HttpMethods`, `consumeTypes` and `produceTypes` on behalf of users, but it turned out to be not.
@trustin and I had a discussion about this, and we have decided that we'd better change current `PathMapping`.
A new wrapper (maybe `Route`? in that case I should rename `RouteBuilder` to something else) will have the `PathMapping`, and optionally `HttpMethods`, `consumeTypes` and `produceTypes`. And the wrapper will handle the request by given condition.
so, inside `RouteBuilder`, `PathMapping` is not always wrapped with `withHttpHeadersInfo`, `RouteBuilder.apply(Consumer<RouteBuilder>)` doesn't need anymore. :-)

anuraaga: Finally back from Golden Week :)

Ah I didn't even notice I was tricked by the manual code formatting ;) Yeah if it's

```
sb.route()
    .get()
    .service()
    .route()
    .get()
    .service()
```

then it's consistent in most code styles. Personally I feel like the addition of a variable to have indentation per route is a win, but it's up to you. I find this sort of pattern is easy to write, but very hard to read - and usually we'd want to prioritize code readability?

trustin: I guess breaking the sentence will help readability quite a bit..?

```java
sb.route()
  .get()
  .service();

sb.route()
  .get()
  .service();
```

I still think the version that uses lambda expression makes the code too verbose. It'd be nice if we have a DSL for Kotlin later though.

anuraaga: Ah that's a good point, breaking between routes seems like a good option :)

Just for reference, though, I believe `Consumer<RouteBuilder>` results in this kotlin code

```kotlin
sb.route({
  get("/foo")
  consumes(...)
  etc(...)
})
```

so it might still be nice for that DSL

trustin: Yeah, @minwoox and I talked about adding `ServerBuilder.route(Consumer<RouteBuilder>)` as an overridden method, but we were not sure if it's a good idea to have two `route()` methods with different return types..  Should we have it with different name, like `withRoute(Consumer<RouteBuilder>)`, or would it just complicate the API?

anuraaga: Using `with` seems fine to me - it's consistent with `withMutations` in the other PR too :) It'd be good to consolidate on that throughout the codebase, almost always use `withXXX` when accepting a `Consumer`.

trustin: What do you think @minwoox?

minwoox: Yeah, it sounds a good idea. 😄

PR #1741, opened 2019-04-25
`RoutingTrie` can now handle a trailing slash correctly

hyangtack:
Motivation:
- #1735 made `DefaultPathMapping` generate a correct path pattern with a trailing slash,
  but the `RoutingTrie` hasn't still handled the trailing slash correctly.

Modifications:
- Made `RoutingTrie` handle the trailing slash correctly.

Result:
- A service can be correctly mapped to the path with a trailing slash.


trustin:
:+1:

Issue #489, opened 2017-04-05
Allow using wildcards with RequestContextExportingAppender


trustin:
A user has to specify all individual property names. For example:

```xml
    <export>elapsed_nanos</export>
    <export>local.host</export>
    <export>local.port</export>
    <export>remote.host</export>
    <export>remote.port</export>
    <export>req.authority</export>
    <export>req.content_length</export>
    <export>req.direction</export>
    <export>req.method</export>
    <export>req.path</export>
    <export>req.rpc_method</export>
    <export>req.rpc_params</export>
    <export>res.content_length</export>
    <export>res.rpc_result</export>
    <export>res.status_code</export>
    <export>scheme</export>
    <export>tls.cipher</export>
    <export>tls.proto</export>
    <export>tls.session_id</export>
    <export>req.http_headers.user-agent</export>
    <export>req.http_headers.x-forwarded-for</export>
```

A user may want to use a wildcard character to simplify this:

```xml
    <export>*</export> <!-- All built-ins -->
    <export>req.http_headers.user-agent</export>
    <export>req.http_headers.x-forwarded-for</export>
```

.. or ..

```xml
    <export>*</export> <!-- All built-ins -->
    <export>-tls.*</export> <!-- except the TLS-related ones -->
```

masonshin:
I'm working on it.

PR #1742, opened 2019-04-25
Allow using wildcards with RequestContextExportingAppender

masonshin:
For #489

I applied a wildcard `*` pattern about builtin properties. We can also adopt a wildcard about `headers` or `attrs`. But it can be caused some overhead to by logging.
I also specified to document that the wildcard applies only to the `builtin` property.

### DIFF
--- a/logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java
+++ b/logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java
@@ -15,10 +15,12 @@
  */
 package com.linecorp.armeria.common.logback;
 
-import java.util.ArrayList;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.regex.Pattern;
 
 import javax.net.ssl.SSLSession;
@@ -165,13 +167,10 @@ public enum BuiltInProperty {
     static List<BuiltInProperty> findByMdcKeyPattern(String mdcKeyPattern) {
         final Pattern pattern = Pattern.compile(
                 ("\\Q" + mdcKeyPattern + "\\E").replaceAll(WILDCARD_REGEX, "\\\\E.*\\\\Q"));
-        final List<BuiltInProperty> matchedBuiltInProperties = new ArrayList<>();
-        mdcKeyToEnum.forEach((mdcKey, property) -> {
-            if (pattern.matcher(mdcKey).matches()) {
-                matchedBuiltInProperties.add(property);
-            }
-        });
-        return matchedBuiltInProperties;
+        return mdcKeyToEnum.entrySet().stream()
+                           .filter(e -> pattern.matcher(e.getKey()).matches())
+                           .map(Entry::getValue)
+                           .collect(toImmutableList());
     }
 
     final String mdcKey;
--- a/logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java
+++ b/logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java
@@ -15,9 +15,11 @@
  */
 package com.linecorp.armeria.common.logback;
 
+import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java
```diff
@@ -157,8 +161,16 @@
         mdcKeyToEnum = builder.build();
     }
 
-    static Optional<BuiltInProperty> findByMdcKey(String mdcKey) {
-        return Optional.ofNullable(mdcKeyToEnum.get(mdcKey));
+    static List<BuiltInProperty> findByMdcKeyPattern(String mdcKeyPattern) {
+        final Pattern pattern = Pattern.compile(
+                ("\\Q" + mdcKeyPattern + "\\E").replace(WILDCARD_STR, "\\E.*\\Q"));
```
trustin: Perhaps we should use `replaceAll()` to handle the case where there are more than one asterisk in the pattern?

logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java
```diff
@@ -157,8 +161,16 @@
...
+        final List<BuiltInProperty> matchedBuiltInProperties = new ArrayList<>();
+        mdcKeyToEnum.forEach((mdcKey, property) -> {
+            if (pattern.matcher(mdcKey).matches()) {
+                matchedBuiltInProperties.add(property);
```
trustin: How about using the stream API?

```java
mdcKeyToEnum.entrySet().stream()
            .filter(e -> pattern.matcher(e.getKey()).matches())
            .map(Entry::getValue)
            .collect(toImmutableList());
```

trustin:
Gentle ping, @masonshin :wink:

logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java
masonshin: `String.replace` replaces matched targets, so it is working well as expected. But it seems to be good to avoid confusion. I will change it to `replaceAll`.

logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java
masonshin: :sunglasses:

masonshin:
@trustin Sorry too late update. PTAL once again. :)

logback/src/main/java/com/linecorp/armeria/common/logback/RequestContextExporterBuilder.java
```diff
@@ -117,9 +117,13 @@ private static AsciiString toHeaderName(CharSequence name) {
...
-        if (opt.isPresent()) {
-            builtIns.add(opt.get());
+        final List<BuiltInProperty> builtInPropertyList = BuiltInProperty.findByMdcKeyPattern(mdcKey);
+        if (!builtInPropertyList.isEmpty()) {
+            builtIns.addAll(builtInPropertyList);
+            return;
+        }
+
+        if (mdcKey.contains(BuiltInProperty.WILDCARD_STR)) {
             return;
```
minwoox: Just wondering if we need to log or throw an exception when the key contains `*` and there are no properties found for that.

masonshin: Since the document tells explicitly that it only supports built-in properties, it is better to tell it clearly if no matches are found.

minwoox: @masonshin Could you revert it, please? 🙇 
Have thought about it again and we can skip this case silently because it means including the properties matched by the pattern. (i.e there could be no match and we can skip to the next one.)

masonshin: I agree. Let me revert it.

trustin:
Thank you, @masonshin!

PR #1743, opened 2019-04-25
Provide a way to create a client with an Endpoint rather than with a URI

SeungukHan:
Motivation:
   - Current HttpClient only allow string or URI object for argument,
     but Endpoint is commonly use for handling URI. For convenience,
     HttpClient can accept Endpoint as argument.

   Modification:
   - Add Endpoint argument to HttpClient creation methods.

   Result:
   - HttpClient can be created with Endpoint instead of string or URI.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
@@ -22,6 +22,8 @@ import java.net.URI;
 import java.util.Optional;
 import java.util.Set;
 
+import javax.annotation.Nullable;
+
 import com.linecorp.armeria.common.Scheme;
 
 /**
@@ -42,29 +44,12 @@ public abstract class AbstractClientFactory implements ClientFactory {
         return newClient(URI.create(uri), clientType, options);
     }
 
-    /**
-     * Creates a new client that connects to the specified {@link URI} using the default
-     * {@link ClientFactory}.
-     *
-     * @param uri the URI of the server endpoint
-     * @param clientType the type of the new client
-     * @param options the {@link ClientOptionValue}s
-     */
     @Override
     public final <T> T newClient(URI uri, Class<T> clientType, ClientOptionValue<?>... options) {
         requireNonNull(options, "options");
         return newClient(uri, clientType, ClientOptions.of(options));
     }
 
-    /**
-     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme} using
-     * the default {@link ClientFactory}.
-     *
-     * @param scheme the {@link Scheme} for the {@code endpoint}
-     * @param endpoint the server {@link Endpoint}
-     * @param clientType the type of the new client
-     * @param options the {@link ClientOptionValue}s
-     */
     @Override
     public final <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType,
                                  ClientOptionValue<?>... options) {
@@ -74,15 +59,6 @@ public abstract class AbstractClientFactory implements ClientFactory {
         return newClient(scheme, endpoint, clientType, ClientOptions.of(options));
     }
 
-    /**
-     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme} using
-     * the default {@link ClientFactory}.
-     *
-     * @param scheme the {@link Scheme} for the {@code endpoint}
-     * @param endpoint the server {@link Endpoint}
-     * @param clientType the type of the new client
-     * @param options the {@link ClientOptions}
-     */
     @Override
     public final <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType, ClientOptions options) {
         requireNonNull(scheme, "scheme");
@@ -91,22 +67,11 @@ public abstract class AbstractClientFactory implements ClientFactory {
         return newClient(scheme, endpoint, "", clientType, options);
     }
 
-    /**
-     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme}
-     * and {@code path} using the default {@link ClientFactory}.
-     *
-     * @param scheme the {@link Scheme} for the {@code endpoint}
-     * @param endpoint the server {@link Endpoint}
-     * @param path the service {@code path}
-     * @param clientType the type of the new client
-     * @param options the {@link ClientOptionValue}s
-     */
     @Override
-    public final <T> T newClient(Scheme scheme, Endpoint endpoint, String path, Class<T> clientType,
+    public final <T> T newClient(Scheme scheme, Endpoint endpoint, @Nullable String path, Class<T> clientType,
                                  ClientOptionValue<?>... options) {
         requireNonNull(scheme, "scheme");
         requireNonNull(endpoint, "endpoint");
-        requireNonNull(path, "path");
         requireNonNull(options, "options");
         return newClient(scheme, endpoint, path, clientType, ClientOptions.of(options));
     }
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
@@ -111,10 +111,10 @@ public abstract class AbstractClientFactory implements ClientFactory {
     }
 
     /**
-     * Makes sure the scheme of the specified {@link Scheme} is supported by this {@link ClientFactory}.
+     * Makes sure the specified {@link Scheme} is supported by this {@link ClientFactory}.
      *
      * @param scheme the {@link Scheme} of the server endpoint
-     * @return the supported {@link Scheme}
+     * @return the {@link Scheme}
      *
      * @throws IllegalArgumentException if the {@link Scheme} is not supported by this {@link ClientFactory}
      */
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
@@ -75,6 +75,43 @@ public abstract class AbstractClientFactory implements ClientFactory {
     }
 
     /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme} using
+     * the default {@link ClientFactory}.
+     *
+     * @param scheme the {@link Scheme} for the {@code endpoint}
+     * @param endpoint the server {@link Endpoint}
+     * @param clientType the type of the new client
+     * @param options the {@link ClientOptions}
+     */
+    @Override
+    public final <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType, ClientOptions options) {
+        requireNonNull(scheme, "scheme");
+        requireNonNull(endpoint, "endpoint");
+        requireNonNull(options, "options");
+        return newClient(scheme, endpoint, "", clientType, options);
+    }
+
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme}
+     * and {@code path} using the default {@link ClientFactory}.
+     *
+     * @param scheme the {@link Scheme} for the {@code endpoint}
+     * @param endpoint the server {@link Endpoint}
+     * @param path the service {@code path}
+     * @param clientType the type of the new client
+     * @param options the {@link ClientOptionValue}s
+     */
+    @Override
+    public final <T> T newClient(Scheme scheme, Endpoint endpoint, String path, Class<T> clientType,
+                                 ClientOptionValue<?>... options) {
+        requireNonNull(scheme, "scheme");
+        requireNonNull(endpoint, "endpoint");
+        requireNonNull(path, "path");
+        requireNonNull(options, "options");
+        return newClient(scheme, endpoint, path, clientType, ClientOptions.of(options));
+    }
+
+    /**
      * Makes sure the scheme of the specified {@link URI} is supported by this {@link ClientFactory}.
      *
      * @param uri the {@link URI} of the server endpoint
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
@@ -42,10 +42,36 @@ public abstract class AbstractClientFactory implements ClientFactory {
         return newClient(URI.create(uri), clientType, options);
     }
 
+    /**
+     * Creates a new client that connects to the specified {@link URI} using the default
+     * {@link ClientFactory}.
+     *
+     * @param uri the URI of the server endpoint
+     * @param clientType the type of the new client
+     * @param options the {@link ClientOptionValue}s
+     */
     @Override
     public final <T> T newClient(URI uri, Class<T> clientType, ClientOptionValue<?>... options) {
         requireNonNull(options, "options");
         return newClient(uri, clientType, ClientOptions.of(options));
+    }
+
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme} using
+     * the default {@link ClientFactory}.
+     *
+     * @param scheme the {@link Scheme} for the {@code endpoint}
+     * @param endpoint the server {@link Endpoint}
+     * @param clientType the type of the new client
+     * @param options the {@link ClientOptionValue}s
+     */
+    @Override
+    public final <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType,
+                                 ClientOptionValue<?>... options) {
+        requireNonNull(scheme, "scheme");
+        requireNonNull(endpoint, "endpoint");
+        requireNonNull(options, "options");
+        return newClient(scheme, endpoint, clientType, ClientOptions.of(options));
     }
 
     /**
@@ -82,6 +108,26 @@ public abstract class AbstractClientFactory implements ClientFactory {
         }
 
         return parsedScheme;
+    }
+
+    /**
+     * Makes sure the scheme of the specified {@link Scheme} is supported by this {@link ClientFactory}.
+     *
+     * @param scheme the {@link Scheme} of the server endpoint
+     * @return the supported {@link Scheme}
+     *
+     * @throws IllegalArgumentException if the {@link Scheme} is not supported by this {@link ClientFactory}
+     */
+    protected final Scheme validateScheme(Scheme scheme) {
+        requireNonNull(scheme, "scheme");
+
+        final Set<Scheme> supportedSchemes = supportedSchemes();
+        if (!supportedSchemes.contains(scheme)) {
+            throw new IllegalArgumentException(
+                    "Unsupported scheme: " + scheme + " (expected: " + supportedSchemes + ')');
+        }
+
+        return scheme;
     }
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
@@ -34,6 +34,7 @@ public abstract class AbstractClientFactory implements ClientFactory {
     @Override
     public final <T> T newClient(String uri, Class<T> clientType, ClientOptionValue<?>... options) {
         requireNonNull(uri, "uri");
+        requireNonNull(clientType, "clientType");
         requireNonNull(options, "options");
         return newClient(URI.create(uri), clientType, ClientOptions.of(options));
     }
@@ -41,11 +42,15 @@ public abstract class AbstractClientFactory implements ClientFactory {
     @Override
     public final <T> T newClient(String uri, Class<T> clientType, ClientOptions options) {
         requireNonNull(uri, "uri");
+        requireNonNull(clientType, "clientType");
+        requireNonNull(options, "options");
         return newClient(URI.create(uri), clientType, options);
     }
 
     @Override
     public final <T> T newClient(URI uri, Class<T> clientType, ClientOptionValue<?>... options) {
+        requireNonNull(uri, "uri");
+        requireNonNull(clientType, "clientType");
         requireNonNull(options, "options");
         return newClient(uri, clientType, ClientOptions.of(options));
     }
@@ -55,6 +60,7 @@ public abstract class AbstractClientFactory implements ClientFactory {
                                  ClientOptionValue<?>... options) {
         requireNonNull(scheme, "scheme");
         requireNonNull(endpoint, "endpoint");
+        requireNonNull(clientType, "clientType");
         requireNonNull(options, "options");
         return newClient(scheme, endpoint, clientType, ClientOptions.of(options));
     }
@@ -63,8 +69,9 @@ public abstract class AbstractClientFactory implements ClientFactory {
     public final <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType, ClientOptions options) {
         requireNonNull(scheme, "scheme");
         requireNonNull(endpoint, "endpoint");
+        requireNonNull(clientType, "clientType");
         requireNonNull(options, "options");
-        return newClient(scheme, endpoint, "", clientType, options);
+        return newClient(scheme, endpoint, null, clientType, options);
     }
 
     @Override
@@ -72,6 +79,7 @@ public abstract class AbstractClientFactory implements ClientFactory {
                                  ClientOptionValue<?>... options) {
         requireNonNull(scheme, "scheme");
         requireNonNull(endpoint, "endpoint");
+        requireNonNull(clientType, "clientType");
         requireNonNull(options, "options");
         return newClient(scheme, endpoint, path, clientType, ClientOptions.of(options));
     }
--- a/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
@@ -19,6 +19,12 @@ import static java.util.Objects.requireNonNull;
 
 import java.net.URI;
 
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.common.Scheme;
+import com.linecorp.armeria.common.SerializationFormat;
+import com.linecorp.armeria.common.SessionProtocol;
+
 /**
  * Creates a new client that connects to the specified {@link URI} using the builder pattern. Use the factory
  * methods in {@link Clients} if you do not have many options to override. If you are creating an
@@ -49,7 +55,12 @@ import java.net.URI;
  */
 public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuilder> {
 
-    private final URI uri;
+    @Nullable
+    private URI uri;
+    @Nullable
+    private Scheme scheme;
+    @Nullable
+    private Endpoint endpoint;
     private ClientFactory factory = ClientFactory.DEFAULT;
 
     /**
@@ -60,18 +71,44 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuil
     }
 
     /**
-     * Creates a new {@link ClientBuilder} that builds the client that connects to
-     * the specified {@link Endpoint} with {@code scheme}.
-     */
-    public ClientBuilder(String scheme, Endpoint endpoint) {
-        this(requireNonNull(endpoint, "endpoint").toURI(scheme));
-    }
-
-    /**
      * Creates a new {@link ClientBuilder} that builds the client that connects to the specified {@link URI}.
      */
     public ClientBuilder(URI uri) {
         this.uri = requireNonNull(uri, "uri");
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@link SessionProtocol} and the {@link SerializationFormat}.
+     */
+    public ClientBuilder(SessionProtocol protocol, SerializationFormat format, Endpoint endpoint) {
+        this(Scheme.of(format, protocol), requireNonNull(endpoint, "endpoint"));
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@link SessionProtocol}.
+     */
+    public ClientBuilder(SessionProtocol protocol, Endpoint endpoint) {
+        this(requireNonNull(protocol, "protocol"), SerializationFormat.NONE,
+             requireNonNull(endpoint, "endpoint"));
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@code Scheme}.
+     */
+    public ClientBuilder(String scheme, Endpoint endpoint) {
+        this(Scheme.parse(scheme), requireNonNull(endpoint, "endpoint"));
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@link Scheme}.
+     */
+    public ClientBuilder(Scheme scheme, Endpoint endpoint) {
+        this.scheme = requireNonNull(scheme, "scheme");
+        this.endpoint = requireNonNull(endpoint, "endpoint");
     }
 
     /**
@@ -92,6 +129,11 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuil
      */
     public <T> T build(Class<T> clientType) {
         requireNonNull(clientType, "clientType");
-        return factory.newClient(uri, clientType, buildOptions());
+
+        if (uri != null) {
+            return factory.newClient(uri, clientType, buildOptions());
+        } else {
+            return factory.newClient(scheme, endpoint, clientType, buildOptions());
+        }
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
@@ -61,6 +61,8 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuil
     private final Scheme scheme;
     @Nullable
     private final Endpoint endpoint;
+    @Nullable
+    private final String path;
     private ClientFactory factory = ClientFactory.DEFAULT;
 
     /**
@@ -74,7 +76,7 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuil
      * Creates a new {@link ClientBuilder} that builds the client that connects to the specified {@link URI}.
      */
     public ClientBuilder(URI uri) {
-        this(requireNonNull(uri), null, null);
+        this(requireNonNull(uri, "uri"), null, null, null);
     }
 
     /**
@@ -107,13 +109,48 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuil
      * {@link Endpoint} with the {@link Scheme}.
      */
     public ClientBuilder(Scheme scheme, Endpoint endpoint) {
-        this(null, requireNonNull(scheme, "scheme"), requireNonNull(endpoint, "endpoint"));
+        this(null, requireNonNull(scheme, "scheme"), requireNonNull(endpoint, "endpoint"), "");
     }
 
-    private ClientBuilder(URI uri, Scheme scheme, Endpoint endpoint) {
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@link SessionProtocol}, {@link SerializationFormat}, and {@code path}.
+     */
+    public ClientBuilder(SessionProtocol protocol, SerializationFormat format, Endpoint endpoint, String path) {
+        this(Scheme.of(format, protocol), requireNonNull(endpoint, "endpoint"), requireNonNull(path, "path"));
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the HTTP client that connects to the specified
+     * {@link Endpoint} with the {@link SessionProtocol} and {@code path}.
+     */
+    public ClientBuilder(SessionProtocol protocol, Endpoint endpoint, String path) {
+        this(requireNonNull(protocol, "protocol"), SerializationFormat.NONE,
+             requireNonNull(endpoint, "endpoint"), requireNonNull(path, "path"));
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@code scheme} and {@code path}.
+     */
+    public ClientBuilder(String scheme, Endpoint endpoint, String path) {
+        this(Scheme.parse(scheme), requireNonNull(endpoint, "endpoint"), requireNonNull(path, "path"));
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@link Scheme} and {@code path}.
+     */
+    public ClientBuilder(Scheme scheme, Endpoint endpoint, String path) {
+        this(null, requireNonNull(scheme, "scheme"), requireNonNull(endpoint, "endpoint"),
+             requireNonNull(path, "path"));
+    }
+
+    private ClientBuilder(URI uri, Scheme scheme, Endpoint endpoint, String path) {
         this.uri = uri;
         this.scheme = scheme;
         this.endpoint = endpoint;
+        this.path = path;
     }
 
     /**
@@ -138,7 +175,7 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuil
         if (uri != null) {
             return factory.newClient(uri, clientType, buildOptions());
         } else {
-            return factory.newClient(scheme, endpoint, clientType, buildOptions());
+            return factory.newClient(scheme, endpoint, path, clientType, buildOptions());
         }
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
@@ -56,11 +56,11 @@ import com.linecorp.armeria.common.SessionProtocol;
 public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuilder> {
 
     @Nullable
-    private URI uri;
+    private final URI uri;
     @Nullable
-    private Scheme scheme;
+    private final Scheme scheme;
     @Nullable
-    private Endpoint endpoint;
+    private final Endpoint endpoint;
     private ClientFactory factory = ClientFactory.DEFAULT;
 
     /**
@@ -74,7 +74,7 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuil
      * Creates a new {@link ClientBuilder} that builds the client that connects to the specified {@link URI}.
      */
     public ClientBuilder(URI uri) {
-        this.uri = requireNonNull(uri, "uri");
+        this(requireNonNull(uri), null, null);
     }
 
     /**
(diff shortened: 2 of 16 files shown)


core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +139,52 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
         return new HttpClientBuilder(uri).factory(factory).options(options).build();
     }
 
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} using the default
+     * {@link ClientFactory}.
+     *
+     * @param endpoint the URI of the server endpoint {@link Endpoint}
```
minwoox: `endpoint` is redundant.
`the server endpoint {@link Endpoint}` -> `the server {@link Endpoint}`

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +139,52 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+     */
+    static HttpClient of(Endpoint endpoint, ClientOptionValue<?>... options) {
+        return of(ClientFactory.DEFAULT, endpoint.toURI(), options);
+    }
+
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} using the default
+     * {@link ClientFactory}.
+     *
+     * @param endpoint the URI of the server endpoint {@link Endpoint}
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +139,52 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+    static HttpClient of(Endpoint endpoint, ClientOptions options) {
+        return of(ClientFactory.DEFAULT, endpoint.toURI(), options);
+    }
+
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} using an alternative
+     * {@link ClientFactory}.
+     *
+     * @param factory an alternative {@link ClientFactory}
+     * @param endpoint the server endpoint {@link Endpoint}
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +139,52 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+    static HttpClient of(ClientFactory factory, Endpoint endpoint, ClientOptionValue<?>... options) {
+        return new HttpClientBuilder(endpoint.toURI()).factory(factory).options(options).build();
+    }
+
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} using an alternative
+     * {@link ClientFactory}.
+     *
+     * @param factory an alternative {@link ClientFactory}
+     * @param endpoint the server endpoint {@link Endpoint}
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +139,52 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
 
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} using the default
+     * {@link ClientFactory}.
+     *
+     * @param endpoint the URI of the server endpoint {@link Endpoint}
+     * @param options the {@link ClientOptionValue}s
+     */
+    static HttpClient of(Endpoint endpoint, ClientOptionValue<?>... options) {
+        return of(ClientFactory.DEFAULT, endpoint.toURI(), options);
```
minwoox: `requireNonNull(endpoint, "endpoint")`
Should check if the `endpoint` is `null` before calling its method.

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +139,52 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} using the default
+     * {@link ClientFactory}.
+     *
+     * @param endpoint the URI of the server endpoint {@link Endpoint}
+     * @param options the {@link ClientOptions}
+     */
+    static HttpClient of(Endpoint endpoint, ClientOptions options) {
+        return of(ClientFactory.DEFAULT, endpoint.toURI(), options);
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +139,52 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} using an alternative
+     * {@link ClientFactory}.
+     *
+     * @param factory an alternative {@link ClientFactory}
+     * @param endpoint the server endpoint {@link Endpoint}
+     * @param options the {@link ClientOptionValue}s
+     */
+    static HttpClient of(ClientFactory factory, Endpoint endpoint, ClientOptionValue<?>... options) {
+        return new HttpClientBuilder(endpoint.toURI()).factory(factory).options(options).build();
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +139,52 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} using an alternative
+     * {@link ClientFactory}.
+     *
+     * @param factory an alternative {@link ClientFactory}
+     * @param endpoint the server endpoint {@link Endpoint}
+     * @param options the {@link ClientOptions}
+     */
+    static HttpClient of(ClientFactory factory, Endpoint endpoint, ClientOptions options) {
+        return new HttpClientBuilder(endpoint.toURI()).factory(factory).options(options).build();
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +433,32 @@ public String authority() {
         return this.authority = authority;
     }
 
+    /**
+     * Converts this endpoint into a URI string.
+     *
+     * @return the URI string
+     */
+    public String toURI() {
```
minwoox: How about using `authority()` method? If you do that, you do not have to consider when the `endpoint` is a group.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +433,32 @@ public String authority() {
...
+                authority = '[' + ipAddr + ']';
+            }
+        } else {
+            if (NetUtil.isValidIpV4Address(ipAddr)) {
+                authority = ipAddr + ':' + port;
+            } else {
+                authority = '[' + ipAddr + "]:" + port;
+            }
+        }
+        return "http://" + authority;
```
minwoox: Is there any chance that the client uses `https`?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
SeungukHan: Current `authority()` return `domain name` when `domain name` and `ipAddr` are set.
But what I expect to `toURI()` is return `ipAddr` under that condition.

minwoox: When an `Endpoint` has a domain name and IP address, we should use 
IP address + port to connect to the remote host and the domain name as the authority header. So if we just use `ipAddr` the domain name vanishes. 
@Trustin and I have talked about this and he said it's probably better to add `newClient(String scheme, Endpoint endpoint, ClientOptions)` to `ClientFactory` so that we can still use the value.

minwoox:
Could you also add the methods to `Clients` as well, please? ^^*

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
SeungukHan: Great! I understand.
I'll update it. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -59,6 +59,14 @@ public ClientBuilder(String uri) {
         this(URI.create(requireNonNull(uri, "uri")));
     }
 
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to
+     * the specified {@link Endpoint} with {@code scheme}.
+     */
+    public ClientBuilder(String scheme, Endpoint endpoint) {
```
trustin: Could you also add the following variants:

- `ClientBuilder(Scheme scheme, Endpoint endpoint)`
- `ClientBuilder(SessionProtocol protocol, SerializationFormat format, Endpoint endpoint)`
  - Shortcut of `ClientBuilder(Scheme.of(format, protocol), endpoint)`
- `ClientBuilder(SessionProtocol protocol, Endpoint endpoint)`
  - Shortcut of `ClientBuilder(protocol, SerializationFormat.NONE, endpoint)`
- `ClientBuilder(String scheme, Endpoint endpoint)` could also be a shortcut of `ClientBuilder(Scheme.parse(scheme), endpoint)`.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -418,21 +419,43 @@ public String authority() {
...
 
         return this.authority = authority;
     }
 
+    /**
+     * Converts this endpoint into a URI.
+     *
+     * @return the URI
+     */
+    public URI toURI(String scheme) {
```
trustin: - `toUri()`
- Could you also add the variants similar to those mentioned in `ClientBuilder`?
  - One thing I'm not sure is whether `Endpoint.toUri(SessionProtocol)` should return a URI that starts with `none+`. Perhaps we should not? :thinking: 
- When a scheme is a `String`, we need to URI-encode it or make sure it does not contain anything weird, e.g. what would happen if `scheme` is `http://www.badguys.com/`?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -418,21 +419,43 @@ public String authority() {
...
+        return URI.create(scheme + "://" + authority);
+    }
+
+    private String buildAuthority(String host) {
+        if (NetUtil.isValidIpV6Address(host)) {
+            host = '[' + host + ']';
+        }
+
+        if (port != 0) {
+            host = host + ':' + port;
```
trustin: Could you optimize this a little bit so that concatenation is not done in two-fold? e.g.

```java
if (port != 0) {
    if (isValidIpV6Address(host)) {
        return '[' + host + "]:" + port;
    } else {
        return host + ':' + port
    }
else {
    ...
}
```

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +141,61 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
 
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} with {@code scheme}
+     * using the default {@link ClientFactory}.
+     *
+     * @param scheme a scheme of URI
+     * @param endpoint the server {@link Endpoint}
+     * @param options the {@link ClientOptionValue}s
+     */
+    static HttpClient of(String scheme, Endpoint endpoint, ClientOptionValue<?>... options) {
```
trustin: `String scheme` should be replaced with `SessionProtocol protocol`, because the serialization format is always `none` for `HttpClient`.

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +141,61 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} with {@code scheme}
+     * using the default {@link ClientFactory}.
+     *
+     * @param scheme a scheme of URI
+     * @param endpoint the server {@link Endpoint}
+     * @param options the {@link ClientOptions}
+     */
+    static HttpClient of(String scheme, Endpoint endpoint, ClientOptions options) {
```
trustin: `String scheme` should be replaced with `SessionProtocol protocol`, because the serialization format is always `none` for `HttpClient`.

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +141,61 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} with {@code scheme}
+     * using an alternative {@link ClientFactory}.
+     *
+     * @param factory an alternative {@link ClientFactory}
+     * @param scheme a scheme of URI
+     * @param endpoint the server {@link Endpoint}
+     * @param options the {@link ClientOptionValue}s
+     */
+    static HttpClient of(ClientFactory factory, String scheme, Endpoint endpoint,
```
trustin: `String scheme` should be replaced with `SessionProtocol protocol`, because the serialization format is always `none` for `HttpClient`.

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -139,6 +141,61 @@ static HttpClient of(ClientFactory factory, URI uri, ClientOptions options) {
...
+    /**
+     * Creates a new HTTP client that connects to the specified {@link Endpoint} with {@code scheme}
+     * using an alternative {@link ClientFactory}.
+     *
+     * @param factory an alternative {@link ClientFactory}
+     * @param scheme a scheme of URI
+     * @param endpoint the server {@link Endpoint}
+     * @param options the {@link ClientOptions}
+     */
+    static HttpClient of(ClientFactory factory, String scheme, Endpoint endpoint, ClientOptions options) {
```
trustin: `String scheme` should be replaced with `SessionProtocol protocol`, because the serialization format is always `none` for `HttpClient`.

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -45,6 +45,16 @@ public HttpClientBuilder(String uri) {
         this(URI.create(requireNonNull(uri, "uri")));
     }
 
+    /**
+     * Creates a new instance.
+     *
+     * @throws IllegalArgumentException if the {@code scheme} is not one of the fields
+     *                                  in {@link SessionProtocol}
+     */
+    public HttpClientBuilder(String scheme, Endpoint endpoint) {
```
trustin: `String scheme` should be replaced with `SessionProtocol protocol`, because the serialization format is always `none` for `HttpClient`.

trustin:
@SeungukHan Gentle ping 😉

SeungukHan:
Oh! I am on the job!
I think the next update can be pushed until tomorrow!
Sorry for late TT

trustin:
No pressure. Please take your time. :smile:

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -418,21 +422,77 @@ public String authority() {
...
+        if (isGroup()) {
+            authority = "group:" + groupName;
+        } else {
+            authority = buildAuthority(hasIpAddr() ? ipAddr() : host(), port);
+        }
+
+        return URI.create(scheme.uriText() + "://" + authority);
+    }
+
+    private static String buildAuthority(String host, int port) {
```
minwoox: Question: Is there any reason that you didn't use `Endpoint.authority()` method?

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -56,6 +64,17 @@ public HttpClientBuilder(URI uri) {
         this.uri = URI.create(SerializationFormat.NONE + "+" + uri);
     }
 
+    /**
+     * Creates a new instance.
+     *
+     * @throws IllegalArgumentException if the {@link Scheme} is not one of the fields
+     *                                  in {@link SessionProtocol}
+     */
+    public HttpClientBuilder(Scheme scheme, Endpoint endpoint) {
```
minwoox: `scheme` should be replaced with `SessionProtocol` protocol, because the serialization format is always none for `HttpClient`.

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -56,6 +64,17 @@ public HttpClientBuilder(URI uri) {
...
+        this.scheme = requireNonNull(scheme, "scheme");
```
minwoox: After converting a `SessionProtocol` to a `Scheme` by adding `none+`, we still should validate the scheme.

core/src/test/java/com/linecorp/armeria/client/EndpointTest.java
```diff
@@ -33,14 +33,16 @@ public void parse() {
         assertThat(foo.ipAddr()).isNull();
         assertThat(foo.ipFamily()).isNull();
         assertThat(foo.hasIpAddr()).isFalse();
+        assertThat(foo.toUri("none+http").toString()).isEqualTo("none+http://foo");
```
minwoox: Could you add a test case that @trustin mentioned which takes `http://www.badguys.com/` as a parameter?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -418,21 +422,77 @@ public String authority() {
...
+     * @return the URI
+     */
+    public URI toUri(Scheme scheme) {
+        requireNonNull(scheme, "scheme");
+
+        final String authority;
+        if (isGroup()) {
+            authority = "group:" + groupName;
+        } else {
+            authority = buildAuthority(hasIpAddr() ? ipAddr() : host(), port);
```
hyangtack: Is it okay to use `ipAddr` instead of `host` from here? What if the scheme is `https`?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
SeungukHan: Thanks, now I understand the existing `authority` method is enough for creating `URI` object.

core/src/main/java/com/linecorp/armeria/client/Clients.java
```diff
@@ -59,23 +62,23 @@
...
         return newClient(ClientFactory.DEFAULT, uri, clientType, options);
     }
 
     /**
-     * Creates a new client that connects to the specified {@code uri} using an alternative
+     * Creates a new client that connects to the specified {@code uri} using the specified
      * {@link ClientFactory}.
      *
-     * @param factory an alternative {@link ClientFactory}
+     * @param factory the specified {@link ClientFactory}
```
hyangtack: I'm sorry but I think `an alternative {@link ClientFactory}` would be better from here. I wanted to rewrite the method description only. :-)

SeungukHan:
I am going to add some test codes for fill code coverage.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
SeungukHan: I change to use `authority()` method when creating `URI`.

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -618,4 +622,72 @@ void testEscapedPathParam() throws Exception {
 
         assertEquals("routed", response.contentUtf8());
     }
+
+    @Test
+    void givenHttpClient_thenBuildClient() throws Exception {
+        final URI uri = URI.create(server.httpUri("/"));
+        final Endpoint endpoint = Endpoint.of(uri.getHost()).withDefaultPort(uri.getPort());
```
hyangtack: nit: How about extracting this as a new method, like `newEndpoint()`?

SeungukHan:
The `URI` which is used for creating `Client` contains `path` value too.
In order to `Client` having `path` can be generated with `Endpoint` and related paramters, I add `path` parameter for that kind of factory method.

minwoox:
> The URI which is used for creating Client contains path value too.
For Client having path can be generated with Endpoint and related paramters, I add path parameter for that kind of factory method.

I think we don't have to have the variants which take `path` as a parameter.
It makes API more complicated and a user can insert a path to the `HttpRequest`.
So it's probably better not to provide those APIs. What do you think?

SeungukHan:
I totally agree to a `path` parameter makes APIs more complicated.
But, there are some case where a client need base path.
https://github.com/line/armeria/blob/a8ed1ebaa97158b88dcae3c531959e9de95e33f3/thrift/src/test/java/com/linecorp/armeria/server/thrift/ThriftSerializationFormatsTest.java#L59-L61

If there is a workaround solution please let me know. I really hope to remove the `path` parameter.

minwoox:
> But, there are some case where a client need base path.

Then, we should let the user use `uri`? Any thought on this @hyangtack?
Anyway, we can still remove the `path` in `HttpClientBuilder`.

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -42,12 +42,75 @@
         return newClient(URI.create(uri), clientType, options);
     }
 
+    /**
+     * Creates a new client that connects to the specified {@link URI} using the default
+     * {@link ClientFactory}.
```
trustin: This Javadoc block is unnecessary and incorrect. Note that a method description is inherited from the method it overrides/implements.

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -42,12 +42,75 @@
...
+     */
     @Override
     public final <T> T newClient(URI uri, Class<T> clientType, ClientOptionValue<?>... options) {
         requireNonNull(options, "options");
         return newClient(uri, clientType, ClientOptions.of(options));
     }
 
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme} using
+     * the default {@link ClientFactory}.
```
trustin: This Javadoc block is unnecessary and incorrect. Note that a method description is inherited from the method it overrides/implements.

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -42,12 +42,75 @@
...
+                                 ClientOptionValue<?>... options) {
+        requireNonNull(scheme, "scheme");
+        requireNonNull(endpoint, "endpoint");
+        requireNonNull(options, "options");
+        return newClient(scheme, endpoint, clientType, ClientOptions.of(options));
+    }
+
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme} using
+     * the default {@link ClientFactory}.
```
trustin: This Javadoc block is unnecessary and incorrect. Note that a method description is inherited from the method it overrides/implements.

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -42,12 +42,75 @@
...
+    public final <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType, ClientOptions options) {
+        requireNonNull(scheme, "scheme");
+        requireNonNull(endpoint, "endpoint");
+        requireNonNull(options, "options");
+        return newClient(scheme, endpoint, "", clientType, options);
+    }
+
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme}
+     * and {@code path} using the default {@link ClientFactory}.
```
trustin: This Javadoc block is unnecessary and incorrect. Note that a method description is inherited from the method it overrides/implements.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -63,7 +76,81 @@ public ClientBuilder(String uri) {
...
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@link Scheme} and {@code path}.
+     */
+    public ClientBuilder(Scheme scheme, Endpoint endpoint, String path) {
+        this(null, requireNonNull(scheme, "scheme"), requireNonNull(endpoint, "endpoint"),
+             requireNonNull(path, "path"));
+    }
+
+    private ClientBuilder(URI uri, Scheme scheme, Endpoint endpoint, String path) {
```
trustin: - You need to annotate the parameters with `@Nullable`.
- Could you add an assertion like:

  ```java
  assert uri != null || scheme != null && endpoint != null && path != null;
  ```

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -149,6 +149,53 @@ static void disableShutdownHook() {
      */
     <T> T newClient(URI uri, Class<T> clientType, ClientOptions options);
 
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme} using
+     * the default {@link ClientFactory}.
```
trustin: `using the default {@link ClientFactory}` is incorrect and unnecessary.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -149,6 +149,53 @@ static void disableShutdownHook() {
...
+     * @param scheme the {@link Scheme} for the {@code endpoint}
+     * @param endpoint the server {@link Endpoint}
+     * @param clientType the type of the new client
+     * @param options the {@link ClientOptionValue}s
+     */
+    <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType, ClientOptionValue<?>... options);
+
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme} using
+     * the default {@link ClientFactory}.
```
trustin: `using the default {@link ClientFactory}` is incorrect and unnecessary.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -149,6 +149,53 @@ static void disableShutdownHook() {
...
+     * @param scheme the {@link Scheme} for the {@code endpoint}
+     * @param endpoint the server {@link Endpoint}
+     * @param clientType the type of the new client
+     * @param options the {@link ClientOptions}
+     */
+    <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType, ClientOptions options);
+
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme}
+     * and {@code path} using the default {@link ClientFactory}.
```
trustin: `using the default {@link ClientFactory}` is incorrect and unnecessary.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -149,6 +149,53 @@ static void disableShutdownHook() {
...
+     * @param path the service {@code path}
+     * @param clientType the type of the new client
+     * @param options the {@link ClientOptionValue}s
+     */
+    <T> T newClient(Scheme scheme, Endpoint endpoint, String path, Class<T> clientType,
+                    ClientOptionValue<?>... options);
+
+    /**
+     * Creates a new client that connects to the specified {@link Endpoint} with the {@link Scheme}
+     * and {@code path} using the default {@link ClientFactory}.
```
trustin: `using the default {@link ClientFactory}` is incorrect and unnecessary.

trustin:
Looking again, I started to think we are adding too many factory methods, which may increase the cognitive load to our users. How about converting the optional parameters such as `SerializationFormat` and `path` into a builder method? e.g.

```java
// Case 1:
new ClientBuilder(protocol, endpoint)
    .serializationFormat(...)
    .path(...)
    .build(...);
// Case 2:
new ClientBuilder(scheme, endpoint)
    .path(...)
    .build(...);
// Case 3:
new ClientBuilder(scheme, endpoint)
    .serializationFormat(...); // Triggers an IllegalStateException.
```

By making this change, we could remove all the factory constructors which require `SerializationFormat` and/or `path` as parameters.

SeungukHan:
Yes, I also think current `Factory` and `static factory method` are too complicated.
Let me make `builder` get optional paramters.
Thanks!! :)

trustin:
Please ping us when it's ready for another round!

SeungukHan:
Oops! I see! :D

SeungukHan:
Oh, I think it is ready for next round, PTAL~~

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -48,6 +50,32 @@
         return newClient(uri, clientType, ClientOptions.of(options));
     }
 
+    @Override
+    public final <T> T newClient(Scheme scheme, Endpoint endpoint, Class<T> clientType,
```
hyangtack: It's not a part of this PR, but don't we need to do `null`-check for `clientType` as well? Would you please add it? :-)

SeungukHan: Sure~ I added null check for `clientType` at here.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -63,7 +81,41 @@ public ClientBuilder(String uri) {
...
+    }
+
+    private ClientBuilder(@Nullable URI uri, @Nullable Scheme scheme, @Nullable SessionProtocol protocol,
+                          @Nullable Endpoint endpoint) {
+        this.uri = uri;
+        this.scheme = scheme;
+        this.protocol = protocol;
+        this.endpoint = endpoint;
+
+        checkArguments();
```
minwoox: Seems that `check'Arguments'()` throws `IllegalStateException` not `IllegalArgumentExcpetion`.
I think we do not have to call it here and just check the state in `build()`

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -74,6 +126,29 @@ public ClientBuilder factory(ClientFactory factory) {
...
+    /**
+     * Sets the {@link SerializationFormat} of the client. The default is {@link SerializationFormat#NONE}.
+     */
+    public ClientBuilder serializationFormat(SerializationFormat format) {
+        ensureEndpoint();
+        if (scheme != null) {
+            throw new IllegalStateException("scheme is already given");
+        }
+
+        this.format = requireNonNull(format, "serializationFormat");
```
minwoox: nit: "serializationFormat" -> "format"

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -84,6 +159,38 @@ public ClientBuilder factory(ClientFactory factory) {
...
+        }
+
+        if (endpoint == null) {
+            throw new IllegalStateException("both uri and endpoint are not given");
+        }
+
+        if (scheme == null && protocol == null) {
+            throw new IllegalStateException("both scheme and protocol are not given");
+        }
+    }
```
minwoox: I think we do not need this method since the fields are all final and we already did null check in the constructor. And there's no case that the condition violates this states.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -84,6 +159,38 @@ public ClientBuilder factory(ClientFactory factory) {
...
+            return factory.newClient(uri, clientType, buildOptions());
+        } else if (path != null) {
+            return factory.newClient(scheme(), endpoint, path, clientType, buildOptions());
+        } else {
+            return factory.newClient(scheme(), endpoint, clientType, buildOptions());
+        }
+    }
+
+    private Scheme scheme() {
+        return scheme == null ? Scheme.of(format, requireNonNull(protocol)) : scheme;
```
minwoox: Probably we can remove the null check for the `protocol` because when the `scheme` is `null`, the `protocol` is always not `null`?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +438,96 @@ public String authority() {
         return this.authority = authority;
     }
 
+    /**
+     * Converts this endpoint into an URI using the {@code scheme}.
```
minwoox: nit: an URI -> a URI
The article depends on the pronunciation of the following word.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
SeungukHan: Yes, I think so too.
Thanks!

core/src/main/java/com/linecorp/armeria/client/Clients.java
SeungukHan: Oh I didn't apply a change...!
Sorry, let me apply it!

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
SeungukHan: I apply it to other comments too.
Thanks!!

core/src/test/java/com/linecorp/armeria/client/EndpointTest.java
SeungukHan: I add this test case at `toUri()` method.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -63,7 +81,41 @@ public ClientBuilder(String uri) {
      * Creates a new {@link ClientBuilder} that builds the client that connects to the specified {@link URI}.
      */
     public ClientBuilder(URI uri) {
-        this.uri = requireNonNull(uri, "uri");
+        this(requireNonNull(uri, "uri"), null, null, null);
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@code Scheme}.
```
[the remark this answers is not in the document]
SeungukHan: Applied~

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
SeungukHan: Yes, I think we do not need `requireNonNull` at here.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
SeungukHan: Let me remove this checking method.
I think constructors with required parameters are enough.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -66,6 +77,40 @@ public ClientBuilder(URI uri) {
...
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@link SessionProtocol} and the {@link SerializationFormat}.
+     */
+    public ClientBuilder(SessionProtocol protocol, SerializationFormat format, Endpoint endpoint) {
+        this(Scheme.of(format, protocol), requireNonNull(endpoint, "endpoint"));
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
```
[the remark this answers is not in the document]
SeungukHan: Now `ClientBuilder` get `SerializationFormat` as parameter, so we can not conclude the builder always build `HttpClient`.

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
hyangtack: Thank you! 👍

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -84,6 +157,23 @@ public ClientBuilder factory(ClientFactory factory) {
...
+        }
+    }
+
+    private Scheme scheme() {
+        return scheme == null ? Scheme.of(format, protocol) : scheme;
+    }
+
+    private void ensureEndpoint() {
+        if (endpoint == null) {
+            throw new IllegalStateException("endpoint is not given");
```
trustin: How about:

```java
if (endpoint == null) {
    throw new IllegalStateException(
            getClass().getSimpleName() + " must be created with an " + Endpoint.class.getSimpleName() +
            " to call this method.");
}
```

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -63,7 +81,39 @@ public ClientBuilder(String uri) {
...
+    public ClientBuilder(Scheme scheme, Endpoint endpoint) {
+        this(null, requireNonNull(scheme, "scheme"), null, requireNonNull(endpoint, "endpoint"));
+    }
+
+    /**
+     * Creates a new {@link ClientBuilder} that builds the client that connects to the specified
+     * {@link Endpoint} with the {@link SessionProtocol}.
+     */
+    public ClientBuilder(SessionProtocol protocol, Endpoint endpoint) {
+        this(null, null, requireNonNull(protocol, "sessionProtocol"), requireNonNull(endpoint, "endpoint"));
```
hyangtack: nit: `"sessionProtocol"` -> `"protocol"`

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -53,7 +63,24 @@ public HttpClientBuilder(String uri) {
...
+    }
+
+    /**
+     * Creates a new instance.
+     *
+     * @throws IllegalArgumentException if the {@code sessionProtocol} is not one of the fields
+     *                                  in {@link SessionProtocol}
+     */
+    public HttpClientBuilder(SessionProtocol sessionProtocol, Endpoint endpoint) {
+        validateScheme(requireNonNull(sessionProtocol).uriText());
```
hyangtack: nit: `requireNonNull(sessionProtocol)` -> `requireNonNull(sessionProtocol, "sessionProtocol")`

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +438,96 @@ public String authority() {
...
+
+    /**
+     * Converts this endpoint into a URI using the {@link SessionProtocol} and {@link SerializationFormat}.
+     *
+     * @param sessionProtocol the {@link SessionProtocol} for {@link URI}.
+     * @param serializationFormat the {@link SerializationFormat} for {@link URI}
+     *
+     * @return the URI
+     */
+    public URI toUri(SessionProtocol sessionProtocol, SerializationFormat serializationFormat) {
```
trustin: How about removing `SerializationFormat` parameter and making this method to a shortcut of `toUri(sessionProtocol, null)`?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +438,96 @@ public String authority() {
...
+     * Converts this endpoint into a URI using the {@link SessionProtocol}, {@link SerializationFormat}
+     * and {@code path}.
+     *
+     * @param sessionProtocol the {@link SessionProtocol} for {@link URI}.
+     * @param serializationFormat the {@link SerializationFormat} for {@link URI}
+     * @param path the {@code path} for {@link URI}.
+     *
+     * @return the URI
+     */
+    public URI toUri(SessionProtocol sessionProtocol, SerializationFormat serializationFormat, String path) {
```
trustin: - How about removing `SerializationFormat` parameter and making this method to a shortcut of `toUri(Scheme.of(NONE, sessionProtocol), path)`?
- `path` could be `@Nullable`.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +438,96 @@ public String authority() {
...
+     * Converts this endpoint into a URI using the {@code scheme}.
+     *
+     * @param scheme the {@code scheme} for {@link URI}.
+     *
+     * @return the URI
+     */
+    public URI toUri(String scheme) {
+        requireNonNull(scheme, "scheme");
+
+        return toUri(Scheme.parse(scheme));
```
trustin: How about calling `toUri(scheme, null)`?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +438,96 @@ public String authority() {
...
+     * @param scheme the {@code scheme} for {@link URI}.
+     * @param path the {@code path} for {@link URI}.
+     *
+     * @return the URI
+     */
+    public URI toUri(String scheme, String path) {
+        requireNonNull(scheme, "scheme");
+        requireNonNull(path, "path");
+
+        return toUri(Scheme.parse(scheme), path);
```
trustin: This will fail if a user specified `"http"` rather than `"none+http"`. Is this an intended behavior? How about calling `new URI(scheme, authority(), path, ...)` directly here?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +438,96 @@ public String authority() {
...
+
+    /**
+     * Converts this endpoint into a URI using the {@link Scheme} and the {@code path}.
+     *
+     * @param scheme the {@link Scheme} for {@link URI}.
+     * @param path the {@code path} for {@link URI}.
+     *
+     * @return the URI
+     */
+    public URI toUri(Scheme scheme, @Nullable String path) {
```
trustin: Question: What happens if a user specifies an empty `path` instead of `null`? Should we convert it to `null`?

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -433,6 +438,96 @@ public String authority() {
...
+
+    /**
+     * Converts this endpoint into a URI using the {@code scheme} and {@code path}.
+     *
+     * @param scheme the {@code scheme} for {@link URI}.
+     * @param path the {@code path} for {@link URI}.
+     *
+     * @return the URI
+     */
+    public URI toUri(String scheme, String path) {
```
trustin: `path` could be `@Nullable`.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
SeungukHan: Empty `path` is allowed for creating `URI`.
Both URIs created by `empty` and `null` as `path` are same.
I check it with test code.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
SeungukHan: About `http://www.badguys.com`, I find that current URI constructor does not validate `scheme` part correctly, so that `http://www.badguys.com` as scheme is not filtered well.
I can not help but adding `Matcher` for validating `scheme` as described in https://tools.ietf.org/html/rfc3986#section-3.1.

trustin:
Thanks for your patience. Great job! :+1:

Issue #1584, opened 2019-02-12
Provide a way to get request body in service method.


hyeonjames:
Currently, Armeria provides a `StringRequestConverterFunction`, `ByteArrayRequestConverterFunction` by default to get the request body.

However, it is only for specific types

A user may want to get it for other types, too ( e,g. `application/json`, `application/xml` etc )

tmdgus0118:
It'd be better if it supports `byte[]`, `ByteBuf`, `char[]` etc

jongmin92:
I'm working on it.

PR #1745, opened 2019-04-25
Enable to convert request body to expected result type regardless of content-type

jongmin92:
**Related-Issue:** https://github.com/line/armeria/issues/1584

**Motivation:**

Currently, it is possible to convert the request body to `byte[]`, `HttpData`, `String`, `CharSequence` only for a specific `content-type`.
If the converters can convert the body to `byte[]`, `HttpData`, `String`, `CharSequence` regardless of `content-type`, a user can easily convert the request body to specific result type what they want.

**Modification:**

- In `JacksonRequestConverterFunction`, pass the request to the next `RequestConverterFunction` if it failed to deserialize its content.
- In `ByteArrayRequestConverterFunction` and `StringRequestConverterFunction`, the request body can be converted to `expectedResultType` regardless of `content-type`.

**Result:**

A user can convert the request body to `byte[]`, `HttpData`, `String` and `CharSequence` regardless of `content-type`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
@@ -43,7 +43,7 @@ public class ByteArrayRequestConverterFunction implements RequestConverterFuncti
             if (array.length == length) {
                 return array;
             } else {
-                return Arrays.copyOf(array, length);
+                return Arrays.copyOfRange(array, request.content().offset(), length);
             }
         }
         if (expectedResultType == HttpData.class) {
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
@@ -16,9 +16,10 @@
 
 package com.linecorp.armeria.server.annotation;
 
+import java.util.Arrays;
+
 import com.linecorp.armeria.common.AggregatedHttpMessage;
 import com.linecorp.armeria.common.HttpData;
-import com.linecorp.armeria.common.MediaType;
 import com.linecorp.armeria.server.ServiceRequestContext;
 
 /**
@@ -36,17 +37,17 @@ public class ByteArrayRequestConverterFunction implements RequestConverterFuncti
     @Override
     public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage request,
                                  Class<?> expectedResultType) throws Exception {
-        final MediaType mediaType = request.contentType();
-        if (mediaType == null ||
-            mediaType.is(MediaType.OCTET_STREAM) ||
-            mediaType.is(MediaType.APPLICATION_BINARY)) {
-
-            if (expectedResultType == byte[].class) {
-                return request.content().array();
+        if (expectedResultType == byte[].class) {
+            final byte[] array = request.content().array();
+            final int length = request.content().length();
+            if (array.length == length) {
+                return array;
+            } else {
+                return Arrays.copyOf(array, length);
             }
-            if (expectedResultType == HttpData.class) {
-                return request.content();
-            }
+        }
+        if (expectedResultType == HttpData.class) {
+            return request.content();
         }
         return RequestConverterFunction.fallthrough();
     }
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
@@ -37,17 +37,18 @@ public class ByteArrayRequestConverterFunction implements RequestConverterFuncti
     @Override
     public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage request,
                                  Class<?> expectedResultType) throws Exception {
+        final HttpData content = request.content();
         if (expectedResultType == byte[].class) {
-            final byte[] array = request.content().array();
-            final int length = request.content().length();
+            final byte[] array = content.array();
+            final int length = content.length();
             if (array.length == length) {
                 return array;
             } else {
-                return Arrays.copyOfRange(array, request.content().offset(), length);
+                return Arrays.copyOfRange(array, content.offset(), length);
             }
         }
         if (expectedResultType == HttpData.class) {
-            return request.content();
+            return content;
         }
         return RequestConverterFunction.fallthrough();
     }
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
@@ -76,7 +76,8 @@ public class JacksonRequestConverterFunction implements RequestConverterFunction
                 } catch (JsonProcessingException e) {
                     if (expectedResultType == byte[].class ||
                         expectedResultType == HttpData.class ||
-                        expectedResultType == String.class) {
+                        expectedResultType == String.class ||
+                        expectedResultType == CharSequence.class) {
                         return RequestConverterFunction.fallthrough();
                     }
 
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
@@ -75,7 +75,8 @@ public class JacksonRequestConverterFunction implements RequestConverterFunction
                     return reader.readValue(content);
                 } catch (JsonProcessingException e) {
                     if (expectedResultType == byte[].class ||
-                        expectedResultType == HttpData.class) {
+                        expectedResultType == HttpData.class ||
+                        expectedResultType == String.class) {
                         return RequestConverterFunction.fallthrough();
                     }
 
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
@@ -65,23 +65,20 @@ public class JacksonRequestConverterFunction implements RequestConverterFunction
     @Nullable
     public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage request,
                                  Class<?> expectedResultType) throws Exception {
-
         final MediaType contentType = request.contentType();
         if (contentType != null && (contentType.is(MediaType.JSON) ||
                                     contentType.subtype().endsWith("+json"))) {
-            if (expectedResultType == byte[].class) {
-                return request.content().array();
-            }
-            if (expectedResultType == HttpData.class) {
-                return request.content();
-            }
-
             final ObjectReader reader = readers.computeIfAbsent(expectedResultType, mapper::readerFor);
             if (reader != null) {
                 final String content = request.content(contentType.charset().orElse(StandardCharsets.UTF_8));
                 try {
                     return reader.readValue(content);
                 } catch (JsonProcessingException e) {
+                    if (expectedResultType == byte[].class ||
+                        expectedResultType == HttpData.class) {
+                        return RequestConverterFunction.fallthrough();
+                    }
+
                     throw new IllegalArgumentException("failed to parse a JSON document: " + e, e);
                 }
             }
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
@@ -29,6 +29,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 
 import com.linecorp.armeria.common.AggregatedHttpMessage;
+import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.MediaType;
 import com.linecorp.armeria.server.ServiceRequestContext;
 
@@ -68,8 +69,11 @@ public class JacksonRequestConverterFunction implements RequestConverterFunction
         final MediaType contentType = request.contentType();
         if (contentType != null && (contentType.is(MediaType.JSON) ||
                                     contentType.subtype().endsWith("+json"))) {
-            if (expectedResultType == String.class) {
-                return request.content(contentType.charset().orElse(StandardCharsets.UTF_8));
+            if (expectedResultType == byte[].class) {
+                return request.content().array();
+            }
+            if (expectedResultType == HttpData.class) {
+                return request.content();
             }
 
             final ObjectReader reader = readers.computeIfAbsent(expectedResultType, mapper::readerFor);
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -68,6 +68,10 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
         final MediaType contentType = request.contentType();
         if (contentType != null && (contentType.is(MediaType.JSON) ||
                                     contentType.subtype().endsWith("+json"))) {
+            if (expectedResultType == String.class) {
```
hyangtack: In [JSON specifications](https://tools.ietf.org/html/rfc7159.html#page-4), a JSON string value, such as `"something"`, can also be a JSON text. So this might make a wrong result that a user doesn't expect.
So how about trying to convert the content to a JSON object and passing it to the next `RequestConverterFunction` if it failed? 
In this way, you might inject the content as a byte array as well. (as described in the original issue #1584) Also, it may handle XML document, too. 
Please let me know if my comment is unclear. 😃

trustin: How would you distinguish between the following two cases when the request body is:

```json
"value"
```

- Case 1: A user expects `"\"value\"", which is an as-is conversion of the request body.
- Case 2: A user expects `"value"`, which is a conversion of the JSON string.

Given the ambiguity, perhaps we should not handle `String.class` specially but only `byte[].class` and `HttpData.class`?

trustin:
Gentle ping, @jongmin92 :wink:

jongmin92:
Thank for the ping!
I will apply the comments by this week (__)

jongmin92:
Thanks for detailed comments!
I applied comments.
But I don't ensure whether I applied comments well.

I have a question to @hyangtack 
Is it ok the way that tries to convert the content to a JSON object and passing it to the next RequestConverterFunction if it failed?

I think like that. First, To do return as `byte[]` or `HttpData`, I should check `expectedResultType`. But As I understand your comment, Don't I need to check `expectedResultType`.

If I implement based on your comment, I will add `RequestConverterFunction.fallthrough();` instead of [this line](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java#L77) to pass content to the next RequestConverterFunction.

Is it sure I understand it well?

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -68,6 +69,13 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
         final MediaType contentType = request.contentType();
         if (contentType != null && (contentType.is(MediaType.JSON) ||
                                     contentType.subtype().endsWith("+json"))) {
+            if (expectedResultType == byte[].class) {
+                return request.content().array();
```
trustin: You need to make a copy of the array if the length does not match:

```java
final byte[] array = request.content().array();
final int length = request.content().length();
if (array.length == length) {
    return array;
} else {
    return Arrays.copyOf(array, length);
}
```

@hyangtack I'm curious if we need to try to release the `HttpData` here. What do you think?

hyangtack: When the content needs to be converted with `RequestConverterFunction`s, the `AnnotatedHttpService` aggregates the request from [here](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpService.java#L223) like the below: 
```
        final CompletableFuture<AggregatedHttpMessage> f =
                aggregationRequired(aggregationStrategy, req) ? req.aggregate()
                                                              : CompletableFuture.completedFuture(null);
```
So it seems not necessary to release the `HttpData` because the aggregator already did it.

hyangtack:
@jongmin92 Sorry for the late response. :-)

> Is it ok the way that tries to convert the content to a JSON object and passing it to the next RequestConverterFunction if it failed?

I think it is okay. And it should be done because we need to convert `"text"` as a JSON string `text`.

> I think like that. First, To do return as byte[] or HttpData, I should check expectedResultType. But As I understand your comment, Don't I need to check expectedResultType?

You're right. We need to check `expectedResultType`.
How about handling this as the following way?

- Call `RequestConverterFunction.fallthrough()` if the JSON conversion has failed and the `expectedResultType` is one of `HttpData`, `byte[] or `String`.
  - For the other types, throw an exception like as it is.
- Make `StringRequestConverterFunction` convert the content to `String` regardless of the `content-type`.
- Make `ByteArrayRequestConverterFunction` convert the content to `HttpData` or `byte[]` regardless of the `content-type`.

In this way, we can convert the content as the `expectedResultType` regardless of the `content-type`, so it seems closer to the original issue.

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
jongmin92: Thanks for comments. (__)
I have two questions.
1. Does it mean I don't need to handle when `expectedResultType` is the `HttpData`?
2. I don't know the reason why should I copy of the array if the length does not match.
More basically, I don't know well why happen the situation that the length does not match.

jongmin92:
@hyangtack 
Thanks for the comment :)
I understand it!
I will try to apply this comment ASAP!

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
hyangtack: > I don't know the reason why should I copy of the array if the length does not match.
More basically, I don't know well why happen the situation that the length does not match.

You may refer to [here](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/HttpData.java#L234-L247) about why we need to check the length.
Actually, the array always has the same length with the `content-length` because the aggregator aggregates the contents as a single byte array from [here](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/HttpMessageAggregator.java#L139-L146).

But we need to follow the API specifications, so it would be better to check the length. The current implementation of `ByteArrayRequestConverterFunction` also breaks the rule, so please fix it in your PR. :-)

> Does it mean I don't need to handle when expectedResultType is the HttpData?

For `HttpData`, we just need to assign the object reference to the parameter. So I think we don't need to check the length.

jongmin92:
@hyangtack 
I applied the comments!
Please check once again (__)

core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
```diff
@@ -36,18 +37,18 @@
...
-            }
-            if (expectedResultType == HttpData.class) {
-                return request.content();
+        if (expectedResultType == byte[].class) {
+            final byte[] array = request.content().array();
+            final int length = request.content().length();
+            if (array.length == length) {
+                return array;
+            } else {
+                return Arrays.copyOf(array, length);
```
hyangtack: I think that we need to call `Arrays.copyOfRange(array, request.content.offset(), length)` because its start offset may or may not be `0`.

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -74,6 +74,11 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
                 try {
                     return reader.readValue(content);
                 } catch (JsonProcessingException e) {
+                    if (expectedResultType == byte[].class ||
+                        expectedResultType == HttpData.class) {
```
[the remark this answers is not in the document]
jongmin92: Oh, I agree with [trustin's comment](https://github.com/line/armeria/pull/1745#discussion_r278795682).
Do we need to handle `String` too?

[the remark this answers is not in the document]
hyangtack: Ah, I mean that we need to handle a content like `{"a": 1}`. It's not a JSON string, it's an object. So it would be failed to convert the content into a string in this converter. Also, it might be handled by the next `StringRequestConverterFunction` as converting it into a raw string `{"a": 1}`.

For the case that @trustin explained:
If the content is a JSON string like `"value"` and the `expectedResultType` is `String.class`, the conversion would be succeeded. So I think we don't need to consider the case from here.

Does it make sense? Please feel free to leave any question. :-)

core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java
```diff
@@ -74,6 +74,12 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
                 try {
                     return reader.readValue(content);
                 } catch (JsonProcessingException e) {
+                    if (expectedResultType == byte[].class ||
+                        expectedResultType == HttpData.class ||
+                        expectedResultType == String.class) {
```
hyangtack: Oh, we missed `CharSequence.class`. 😅 It is also handled by `StringRequestConverterFunction`, so it should be checked from here, too.
And, would you please add the test about that type, too?

jongmin92:
Thank you for leading me!
Is It also ok these changes?

https://github.com/line/armeria/pull/1745/files#diff-015e919db6e649f771b70d81f440d80cR525
```
-        // StringRequestConverterFunction will work for the content type of any of 'text'.
+        // StringRequestConverterFunction will work for regardless of the content type.
```

https://github.com/line/armeria/pull/1745/files#diff-015e919db6e649f771b70d81f440d80cR532
```
-        // ByteArrayRequestConverterFunction will work for the content type of 'application/octet-stream',
-        // 'application/binary' or none.
+        // ByteArrayRequestConverterFunction will work for regardless of the content type.
```

core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
```diff
@@ -36,18 +37,18 @@
...
-            mediaType.is(MediaType.OCTET_STREAM) ||
-            mediaType.is(MediaType.APPLICATION_BINARY)) {
-
-            if (expectedResultType == byte[].class) {
-                return request.content().array();
-            }
-            if (expectedResultType == HttpData.class) {
-                return request.content();
+        if (expectedResultType == byte[].class) {
+            final byte[] array = request.content().array();
```
minwoox: How about extracting `request.content()` as a variable and reuse it?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -522,15 +522,14 @@ after your request converters, so you don't have to specify any :api:`@RequestCo
         @Post("/hello2")
         public HttpResponse hello2(MyJsonRequest body) { ... }
 
-        // StringRequestConverterFunction will work for the content type of any of 'text'.
+        // StringRequestConverterFunction will work for regardless of the content type.
```
minwoox: work for -> work

site/src/sphinx/server-annotated-service.rst
```diff
@@ -522,15 +522,14 @@ after your request converters, so you don't have to specify any :api:`@RequestCo
...
         @Post("/hello3")
         public HttpResponse hello3(String body) { ... }
 
         @Post("/hello4")
         public HttpResponse hello4(CharSequence body) { ... }
 
-        // ByteArrayRequestConverterFunction will work for the content type of 'application/octet-stream',
-        // 'application/binary' or none.
+        // ByteArrayRequestConverterFunction will work for regardless of the content type.
```

core/src/test/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunctionTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    static final String JSON_TEXT = "{\"key\": \"value\"}";
+
+    @Test
+    public void jsonTextToByteArray() throws Exception {
+        when(req.contentType()).thenReturn(MediaType.JSON);
+        when(req.content()).thenReturn(HttpData.ofUtf8(JSON_TEXT));
+
+        final Object result = function.convertRequest(ctx, req, byte[].class);
+        Assert.assertTrue(result instanceof byte[]);
```
minwoox: Could you use AssertJ instead of TestNG?
`assertThat(result).isInstanceOf(byte[].class)`

core/src/test/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunctionTest.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Test
+    public void jsonTextToHttpData() throws Exception {
+        when(req.contentType()).thenReturn(MediaType.JSON);
+        when(req.content()).thenReturn(HttpData.ofUtf8(JSON_TEXT));
+
+        final Object result = function.convertRequest(ctx, req, HttpData.class);
+        Assert.assertTrue(result instanceof HttpData);
```
minwoox: ditto

core/src/test/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunctionTest.java
```diff
@@ -0,0 +1,81 @@
...
+        function.convertRequest(ctx, req, String.class);
+    }
+
+    @Test
+    public void jsonTextToJsonRequest() throws Exception {
+        when(req.contentType()).thenReturn(MediaType.JSON);
+        when(req.content(StandardCharsets.UTF_8)).thenReturn(JSON_TEXT);
+
+        final Object result = function.convertRequest(ctx, req, JsonRequest.class);
+        Assert.assertTrue(result instanceof JsonRequest);
```
minwoox: ditto

core/src/test/java/com/linecorp/armeria/server/annotation/StringRequestConverterFunctionTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    static final String JSON_TEXT = "{\"a\": 1}";
+
+    @Test
+    public void jsonTextToString() throws Exception {
+        when(req.contentType()).thenReturn(MediaType.JSON);
+        when(req.content(ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET)).thenReturn(JSON_TEXT);
+
+        final Object result = function.convertRequest(ctx, req, String.class);
+        Assert.assertTrue(result instanceof String);
```
minwoox: ditto

core/src/test/java/com/linecorp/armeria/server/annotation/StringRequestConverterFunctionTest.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Test
+    public void jsonTextToCharSequence() throws Exception {
+        when(req.contentType()).thenReturn(MediaType.JSON);
+        when(req.content(ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET)).thenReturn(JSON_TEXT);
+
+        final Object result = function.convertRequest(ctx, req, CharSequence.class);
+        Assert.assertTrue(result instanceof CharSequence);
```
minwoox: ditto

hyangtack:
@jongmin92 Would you update the PR description for the recent changes? 😃

jongmin92:
@hyangtack 
I just updated the PR title and description. Could you check it is alright? 😃

hyangtack:
@jongmin92 Thank you! It's time to merge this PR! 🎉

hyangtack:
Thank you for your first contribution, @jongmin92 ! 👍

jongmin92:
Thanks for leading me! I will try to handle other issues!

PR #1746, opened 2019-04-26
Improve the overall look of the web site

trustin:
Motivation:

It's always better looking nicer.

Modifications:

- Typography
  - Increased the overall font size and line height a little bit for
    readability.
  - Changed the heading font from Source Sans Pro to Merriweather
    because it is more distinguished from body text.
  - Changed the body font from Source Sans Pro to Lato because Source
    Sans Pro is relatively smaller and denser than other sans-serif
    fonts, making the CSS customization harder.
    - Lato is the default font used by the Sphinx theme we are using,
      which means one less thing to worry.
  - Changed the monospace font from Inconsolata to Hack because:
    - Inconsolata looks thin when paired with Lato.
    - Hack is my favorite font. :-)
- Layout
  - Increased the width of the content pane from 800px to 960px.
  - Made the content pane resizable by dragging its left or right edge.
- Badges
  - Replaced the Slack badge with a custom badge from Shields.io,
    simplifying `add_badges.js`.
  - Used `<object>` tag instead of `` tag so that the badges are
    unaffected by our CSS and the hyperlinks embedded in the `.svg`
    images are preserved.
  - Added more badges:
    - GitHub stars
    - Twitter
    - Commit activity
    - Good first issues
- Dependencies
  - gh-fork-ribbon 0.1.1 -> 0.2.2
  - highlightjs-line-numbers.js 2.6.0 -> 2.7.0
  - sphinx-gradle-plugin 2.3.1 -> 2.4.0
- Miscellaneous
  - `0px` -> `0`
  - Cleaned up code block and literal styles
  - Fixed a bug where wrong fonts were being used in the search bar and
    prev/next buttons.

Result:

- Looking better.

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -234,7 +234,7 @@ junit:
     - https://junit.org/junit4/javadoc/4.12/
 
 kr.motd.gradle:
-  sphinx-gradle-plugin: { version: '2.3.1' }
+  sphinx-gradle-plugin: { version: '2.4.0' }
 
 # We do not depend on log4j. We just need this to stop dependency-management-plugin from
 # complaining about its bad POM.
--- a/site/src/sphinx/_static/add_badges.js
+++ b/site/src/sphinx/_static/add_badges.js
@@ -1,24 +1,19 @@
-// Append the Slack invitation and project badges at the end of the menu (sidenav).
-function addSlackInvitation(parent) {
-  var li = document.createElement('li');
-  li.className = 'toctree-l1';
-  var div = document.createElement('div');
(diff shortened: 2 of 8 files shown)


trustin:
New badges:
image

trustin:
### Before:
before

### After:
after

### Badges:
image

trustin:
See https://imgur.com/a/UDUSZFJ for resizable content.

trustin:
ezgif-1-810fa24f6930

trustin:
Thanks for reviewing :-)

PR #1747, opened 2019-04-28
Introduce 'Copy as a curl command' button to a debug request window

### LINKED ISSUE #196: DocService: letting a user copy a debug request as a cURL command
It would be useful if `DocService` provides a way to copy the current debug request as a cURL command.

wooyeong:
Motivation:

- Rarely, I want to copy a debug request to a curl command to put
additional headers to the request. Even if I can copy it from the dev
tools provided by browsers, opening them is annoying.

Modifications:

- Introduced a 'Copy as a curl command' button
  - It copies the converted command to the clipboard, but not intuitive...
- Added some methods to support curl export in `Transport` class
  - Each three transports has its own MIME type and body format to make
  a request in text

Result:

- Close #196

For your reference,

<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

### DIFF
--- a/docs-client/src/containers/MethodPage/DebugPage.tsx
+++ b/docs-client/src/containers/MethodPage/DebugPage.tsx
@@ -285,9 +285,7 @@ class DebugPage extends React.PureComponent<Props, State> {
   private onExport = () => {
     const escapeSingleQuote = (text: string) => text.replace(/'/g, `'\\''`);
 
-    const endpointPath = this.state.endpointPath; // annotated service only
     const additionalHeaders = this.state.additionalHeaders;
-    const queries = this.state.additionalQueries;
     const requestBody = this.state.requestBody;
 
     try {
@@ -315,7 +313,10 @@ class DebugPage extends React.PureComponent<Props, State> {
       }
 
       const httpMethod = method.httpMethod;
-      const [path, body] = transport.curlPathAndBody(
+      const endpoint = transport.findDebugMimeTypeEndpoint(method);
+      const path = endpoint.pathMapping;
+      const body = transport.getCurlBody(
+        endpoint,
         method,
         escapeSingleQuote(requestBody),
       );
@@ -323,10 +324,12 @@ class DebugPage extends React.PureComponent<Props, State> {
 
       if (this.props.isAnnotatedHttpService) {
         if (this.props.exactPathMapping) {
+          const queries = this.state.additionalQueries;
           uri =
             `'${host}${escapeSingleQuote(path)}` +
             `${queries.length > 0 ? `?${escapeSingleQuote(queries)}` : ''}'`;
         } else {
+          const endpointPath = this.state.endpointPath;
           this.validateEndpointPath(endpointPath);
           uri = `'${host}${escapeSingleQuote(endpointPath)}'`;
         }
--- a/docs-client/src/containers/MethodPage/DebugPage.tsx
+++ b/docs-client/src/containers/MethodPage/DebugPage.tsx
@@ -291,7 +291,7 @@ class DebugPage extends React.PureComponent<Props, State> {
       }
 
       const httpMethod = method.httpMethod;
-      const [path, body] = transport.curlPathAndBody(method, requestBody);
+      const [path, body] = transport.curlPathAndBody(method, requestBody.replace(/'/g, `'\\''`));
       let uri;
 
       if (this.props.isAnnotatedHttpService) {
--- a/docs-client/src/containers/MethodPage/DebugPage.tsx
+++ b/docs-client/src/containers/MethodPage/DebugPage.tsx
@@ -339,14 +339,14 @@ class DebugPage extends React.PureComponent<Props, State> {
         headers[docServiceDebug] = 'true';
       }
 
-      const header = Object.keys(headers)
+      const headerOptions = Object.keys(headers)
         .map((title) => {
-          return `-H '${title}:${headers[title]}'`;
+          return `-H '${title}: ${headers[title]}'`;
         })
         .join(' ');
 
       const curlCommand =
-        `curl -X${httpMethod} ${header} ${uri}` +
+        `curl -X${httpMethod} ${headerOptions} ${uri}` +
         `${this.props.useRequestBody ? ` -d '${body}'` : ''}`;
 
       DebugPage.copyTextToClipboard(curlCommand);
--- a/docs-client/src/containers/MethodPage/DebugPage.tsx
+++ b/docs-client/src/containers/MethodPage/DebugPage.tsx
@@ -17,10 +17,10 @@
 import Button from '@material-ui/core/Button';
 import Grid from '@material-ui/core/Grid';
 import IconButton from '@material-ui/core/IconButton';
-import Snackbar from "@material-ui/core/Snackbar";
+import Snackbar from '@material-ui/core/Snackbar';
 import Tooltip from '@material-ui/core/Tooltip';
 import Typography from '@material-ui/core/Typography';
-import CloseIcon from "@material-ui/icons/Close";
+import CloseIcon from '@material-ui/icons/Close';
 import DeleteSweepIcon from '@material-ui/icons/DeleteSweep';
 import FileCopyIcon from '@material-ui/icons/FileCopy';
 import React, { ChangeEvent } from 'react';
@@ -31,7 +31,7 @@ import githubGist from 'react-syntax-highlighter/styles/hljs/github-gist';
 import jsonMinify from 'jsonminify';
 import { RouteComponentProps } from 'react-router';
 import Section from '../../components/Section';
-import { docServiceDebug } from "../../lib/header-provider";
+import { docServiceDebug } from '../../lib/header-provider';
 import jsonPrettify from '../../lib/json-prettify';
 import { Method } from '../../lib/specification';
 import { TRANSPORTS } from '../../lib/transports';
@@ -176,16 +176,16 @@ class DebugPage extends React.PureComponent<Props, State> {
           <Grid item xs={12} sm={6}>
             <Tooltip title="Copy response">
               <IconButton
-                  onClick={this.onCopy}
-                  disabled={this.state.debugResponse.length == 0}
+                onClick={this.onCopy}
+                disabled={this.state.debugResponse.length === 0}
               >
                 <FileCopyIcon />
               </IconButton>
             </Tooltip>
             <Tooltip title="Clear response">
               <IconButton
-                  onClick={this.onClear}
-                  disabled={this.state.debugResponse.length == 0}
+                onClick={this.onClear}
+                disabled={this.state.debugResponse.length === 0}
               >
                 <DeleteSweepIcon />
               </IconButton>
@@ -200,18 +200,15 @@ class DebugPage extends React.PureComponent<Props, State> {
           </Grid>
         </Grid>
         <Snackbar
-            open={this.state.snackbarOpen}
-            message={this.state.snackbarMessage}
-            autoHideDuration={3000}
-            onClose={this.onSnackbarDismiss}
-            action={[
-              <IconButton
-                  color="inherit"
-                  onClick={this.onSnackbarDismiss}
-              >
-                <CloseIcon />
-              </IconButton>,
-            ]}
+          open={this.state.snackbarOpen}
+          message={this.state.snackbarMessage}
+          autoHideDuration={3000}
+          onClose={this.onSnackbarDismiss}
+          action={
+            <IconButton color="inherit" onClick={this.onSnackbarDismiss}>
+              <CloseIcon />
+            </IconButton>
+          }
         />
       </Section>
     );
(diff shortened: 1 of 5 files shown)


trustin:
This is pretty cool! Just a few comments:

- How about renaming the button to `Copy as a cURL command` and making it copy directly to the clipboard instead of using the right side pane?
  - When the command is copied to clipboard, we could show some notification using a [Snackbar](https://material-ui.com/demos/snackbars/).
- Could you handle the case where the request body contains `'`? You could use heredoc as explained in: https://stackoverflow.com/a/23930212

wooyeong:
Okay, I'll change the label and copying it directly w/o touching the panel on the right, and will attempt to use of `Snackbar` (actually I thought about it yesterday). Also, will take a look at the body part you mentioned as well.

By the way, is `cURL` right? according to the [official repository](https://github.com/curl/curl) and [website](https://curl.haxx.se/), I think they usually write it down as `curl` or `Curl` nowadays?

trustin:
> By the way, is cURL right?

Let's use the official notation, then. :-)

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -237,14 +273,74 @@ class DebugPage extends React.PureComponent<Props, State> {
...
     const response = this.state.debugResponse;
-    const textArea = document.createElement('textarea');
-    textArea.style.opacity = '0.0';
-    textArea.value = response;
-    document.body.appendChild(textArea);
-    textArea.focus();
-    textArea.select();
-    document.execCommand('copy');
-    document.body.removeChild(textArea);
+    if (response.length > 0) {
```
wooyeong: An empty string cannot be copied to the clipboard, at least on my browser.

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -321,6 +419,19 @@ class DebugPage extends React.PureComponent<Props, State> {
     this.executeRequest(params);
   };
 
+  private onSnackbarOpen = (text: string) => {
```
minwoox: Could be just `showSnackbar`?

docs-client/src/lib/transports/transport.ts
```diff
@@ -47,8 +47,37 @@ export default abstract class Transport {
     return this.doSend(method, filledHeaders, bodyJson, endpointPath, queries);
   }
 
+  public curlPathAndBody(method: Method, body: string): [string, string] {
+    const endpoint = this.findDebugMimeTypeEndpoint(method);
+    if (!endpoint) {
```
minwoox: Probably no need to check this because `findDebugMimeTypeEndpoint` do not return `undefined`?

docs-client/src/lib/transports/transport.ts
```diff
@@ -47,8 +47,37 @@ export default abstract class Transport {
...
+
+    return [this.getCurlPath(endpoint), this.getCurlBody(body, endpoint, method)];
+  }
+
+  protected findDebugMimeTypeEndpoint(method: Method): Endpoint {
+    const endpoint = method.endpoints.find((ep) =>
+        ep.availableMimeTypes.includes(this.getDebugMimeType()),
+    );
+    if (!endpoint) {
+      throw new Error('Endpoint does not support debug transport.');
```
minwoox: How about `Endpoint does not support debug transport. MimeType: ${this.getDebugMimeType()}`?

minwoox: Also, Could you remove the duplicate code in grpc-unframed.ts, please?

docs-client/src/containers/MethodPage/DebugPage.tsx
trustin: Or disable the button?

wooyeong: I like @trustin's idea. 😆

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -171,6 +193,20 @@ class DebugPage extends React.PureComponent<Props, State> {
...
           </Grid>
         </Grid>
+        <Snackbar
+            open={this.state.snackbarOpen}
+            message={this.state.snackbarMessage}
+            autoHideDuration={3000}
+            onClose={this.onSnackbarDismiss}
+            action={[
+              <IconButton
+                  color="inherit"
```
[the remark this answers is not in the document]
wooyeong: W/o this, color doesn't look good.

<kbd></kbd>

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -237,14 +273,76 @@ class DebugPage extends React.PureComponent<Props, State> {
...
+        throw new Error("This method doesn't have a debug transport.");
+      }
+
+      const httpMethod = method.httpMethod;
+      const [path, body] = transport.curlPathAndBody(method, escapeSingleQuote(requestBody));
+      let uri;
+
+      if (this.props.isAnnotatedHttpService) {
+        if (this.props.exactPathMapping) {
+          uri = `'${host}${escapeSingleQuote(path)}?${escapeSingleQuote(queries)}'`;
```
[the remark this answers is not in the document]
wooyeong: Updated!

docs-client/src/containers/MethodPage/DebugPage.tsx
wooyeong: Changed.

docs-client/src/lib/transports/transport.ts
wooyeong: 👍

docs-client/src/lib/transports/transport.ts
wooyeong: Also, removed from `annotated-http.ts`

wooyeong:
I wonder if there were some regressions when I apply checkstyle. Could you kindly take a look please?

docs-client/src/containers/MethodPage/DebugPage.tsx
minwoox: Yup. 😆

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -76,6 +81,18 @@ class DebugPage extends React.PureComponent<Props, State> {
       );
     }
   }
+
+  private static copyTextToClipboard(text: string) {
+    const textArea = document.createElement('textarea');
+    textArea.style.opacity = '0.0';
```
trustin: Not an expert on this topic, but could this be `display = 'none'`?

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -237,14 +276,86 @@ class DebugPage extends React.PureComponent<Props, State> {
...
+      }
+
+      headers['content-type'] = transport.getDebugMimeType();
+      if (process.env.WEBPACK_SERVE === 'true') {
+        headers[docServiceDebug] = 'true';
+      }
+
+      const header = Object.keys(headers)
+        .map((title) => {
+          return `-H '${title}:${headers[title]}'`;
```
trustin: I'm curious it will look prettier if we insert a space after a colon (`:`).

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -237,14 +276,86 @@ class DebugPage extends React.PureComponent<Props, State> {
...
+      } else {
+        uri = `'${host}${escapeSingleQuote(path)}'`;
+      }
+
+      headers['content-type'] = transport.getDebugMimeType();
+      if (process.env.WEBPACK_SERVE === 'true') {
+        headers[docServiceDebug] = 'true';
+      }
+
+      const header = Object.keys(headers)
```
trustin: `headerOpt` ?

docs-client/src/lib/transports/annotated-http.ts
```diff
@@ -25,21 +25,22 @@ export default class AnnotatedHttpTransport extends Transport {
     return mimeType === ANNOTATED_HTTP_MIME_TYPE;
   }
 
+  public getDebugMimeType(): string {
+    return ANNOTATED_HTTP_MIME_TYPE;
+  }
+
+  protected getCurlPath(endpoint: Endpoint): string {
+    return endpoint.pathMapping.substring('exact:'.length);
```
minwoox: How about removing this method and just handling it in `DebugPage`?
I think this can cause misunderstanding to the reader because this only handles an exact path mapping.

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -237,14 +276,86 @@ class DebugPage extends React.PureComponent<Props, State> {
...
+    if (response.length > 0) {
+      DebugPage.copyTextToClipboard(response);
+      this.showSnackbar('The response has been copied to the clipboard.');
+    }
+  };
+
+  private onExport = () => {
+    const escapeSingleQuote = (text: string) => text.replace(/'/g, `'\\''`);
+
+    const endpointPath = this.state.endpointPath; // annotated service only
```
minwoox: This line can be removed and the `endpointPath` can be declared where it needs (line 330).

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -237,14 +276,86 @@ class DebugPage extends React.PureComponent<Props, State> {
...
+    const additionalHeaders = this.state.additionalHeaders;
+    const queries = this.state.additionalQueries;
```
minwoox: ditto

docs-client/src/lib/transports/transport.ts
```diff
@@ -47,7 +51,37 @@ export default abstract class Transport {
...
+  }
+
+  protected getCurlPath(endpoint: Endpoint): string {
+    return endpoint.pathMapping;
+  }
+
+  protected getCurlBody(
+    body: string,
+    _endpoint: Endpoint,
+    _method: Method,
```
minwoox: nit: Could remove underscores before the parameters?

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -237,14 +276,86 @@ class DebugPage extends React.PureComponent<Props, State> {
...
+        uri = `'${host}${escapeSingleQuote(path)}'`;
+      }
+
+      headers['content-type'] = transport.getDebugMimeType();
+      if (process.env.WEBPACK_SERVE === 'true') {
+        headers[docServiceDebug] = 'true';
+      }
+
+      const header = Object.keys(headers)
+        .map((title) => {
```
hyangtack: nit: How about renaming `title` to `name` for better readability?

docs-client/src/containers/MethodPage/DebugPage.tsx
wooyeong: Having a space makes it look better.

docs-client/src/containers/MethodPage/DebugPage.tsx
wooyeong: Changed to `headerOptions`.

docs-client/src/lib/transports/annotated-http.ts
wooyeong: Removed `getCurlPath`, and I think it becomes simpler.

docs-client/src/lib/transports/transport.ts
wooyeong: Since they are unused method parameters (an overridden method uses), they need to start w/ an underscore.

docs-client/src/containers/MethodPage/DebugPage.tsx
wooyeong: Changed to `name`.

docs-client/src/containers/MethodPage/DebugPage.tsx
wooyeong: Thanks :)

docs-client/src/containers/MethodPage/DebugPage.tsx
wooyeong: I've tried `display = 'none'` and `visibility = 'hidden'`, neither works well on chrome.
I guess dom's physical existence and visibility affect the behavior, but also I'm not an expert :)

docs-client/src/lib/transports/transport.ts
minwoox: Forgot that ESLInt has that rule. 😆

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -237,14 +276,89 @@ class DebugPage extends React.PureComponent<Props, State> {
...
+        method,
+        escapeSingleQuote(requestBody),
+      );
+      let uri;
+
+      if (this.props.isAnnotatedHttpService) {
+        if (this.props.exactPathMapping) {
+          const queries = this.state.additionalQueries;
+          uri =
+            `'${host}${escapeSingleQuote(path)}` +
```
minwoox: Should remove `exact:` from the path.

wooyeong: Thanks. 😭

minwoox:
Please, fix the check style error. :-)

wooyeong:
I should be familiar w/ `'` when I write code in typescript 😅. Sorry!

hyangtack:
Nice work, @wooyeong ! Thanks for your patience.

hyangtack:
Thank you for your first contribution, @wooyeong ! 👍

wooyeong:
Thank you for reviewing this 🙇

PR #1748, opened 2019-04-30
Add UnaryGrpcClient.

### LINKED ISSUE #1703: Allow unary grpc without google's library

anuraaga:
Fixes #1703 

This is the client version of #1727. Advanced users can use `UnaryGrpcClient` to issue gRPC-framed requests to a gRPC server given they handle serialization / deserialization of messages themselves, either by depending on protobuf (without needing to depend on grpc) or reimplementing the logic themselves.

As this is an advanced API, I think it's good to keep it small since it won't be as well tested as normal APIs. I've only exposed the most flexible APIs without providing any extra helpers as a result.

/cc @adriancole This should finish the ungrpc'ing of zipkin

### DIFF
--- a/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
+++ b/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.common.grpc.protocol;
+
+import java.util.concurrent.CompletableFuture;
+
+import com.linecorp.armeria.client.Client;
+import com.linecorp.armeria.client.ClientDecoration;
+import com.linecorp.armeria.client.ClientOption;
+import com.linecorp.armeria.client.ClientRequestContext;
+import com.linecorp.armeria.client.Clients;
+import com.linecorp.armeria.client.HttpClient;
+import com.linecorp.armeria.client.SimpleDecoratingClient;
+import com.linecorp.armeria.common.HttpData;
+import com.linecorp.armeria.common.HttpHeaderNames;
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.grpc.protocol.ArmeriaMessageDeframer.ByteBufOrStream;
+import com.linecorp.armeria.common.grpc.protocol.ArmeriaMessageDeframer.Listener;
+import com.linecorp.armeria.unsafe.ByteBufHttpData;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
+import io.netty.buffer.Unpooled;
+
+/**
+ * A {@link UnaryGrpcClient} can be used to make requests go a gRPC server without depending on gRPC stubs.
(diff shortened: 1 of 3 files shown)


adriancole:
cool! you thinking this for stackdriver?

anuraaga:
Yeah

adriancole:
looking forward to it

grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
```diff
@@ -0,0 +1,176 @@
...
+                           final HttpStatus status = msg.status();
+                           // Response always has status.
+                           assert status != null;
+
+                           if (!status.equals(HttpStatus.OK) || msg.content().isEmpty()) {
+                               // Nothing to deframe.
+                               return CompletableFuture.completedFuture(HttpResponse.of(msg));
+                           }
+
+                           CompletableFuture<HttpResponse> responseFuture = new CompletableFuture<>();
```
minwoox: nit: final?

anuraaga:
Thanks, updated

grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
```diff
@@ -0,0 +1,176 @@
...
+ * normal gRPC stubs or file a feature request.
+ */
+public class UnaryGrpcClient {
+
+    private final HttpClient httpClient;
+
+    /**
+     * Constructs a {@link UnaryGrpcClient} for the given {@link HttpClient}.
+     */
+    public UnaryGrpcClient(HttpClient httpClient) {
```
trustin: I find this is not consistent with how other Armeria clients are instantiated. For example, I can imagine the following user code:

```java
UnaryGrpcClient client = new ClientBuilder("graw+https://...")
    ...
    .build(UnaryGrpcClient.class);

client.execute(path, payload);
```

Side note: Not an immediate goal, but a unary gRPC service/client could make use of `RpcRequest` and `RpcResponse`?

grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
```diff
@@ -0,0 +1,176 @@
...
+                httpClient,
+                ClientOption.DECORATION.newValue(
+                        ClientDecoration.of(HttpRequest.class, HttpResponse.class, GrpcFramingDecorator::new)
+                ));
+    }
+
+    /**
+     * Executes a unary gRPC client request. The given {@code payload} will be framed and sent to the path at
+     * {@code uri}. {@code uri} should be the method's URI, which is always of the format
+     * /:package-name.:service-name/:method. For example, for the proto package {@code armeria.protocol}, the
```
trustin: `{@code /:package-name.:service-name/:method}`

grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
anuraaga: Ah cool idea, will try it out since it doesn't look that tough!

I did think a bit about `RpcRequest/Response` but it seemed like it'd make the implementation quite complicated and I don't expect many users of the class so keeping it simple might be best.

anuraaga: I realized it might not work well since we map `gproto` to the `application/grpc` media type since many gRPC upstream servers do not work with `grpc+proto`, at least at the time of writing that. `graw` would need to be the same content type, which I don't think we allow with `SerializationFormatProvider`. Do you think it's worth investigating it more?

trustin: Ah, that's a problem. Let's add some TODO/FIXME and fix it later then. Thanks for giving a try!

trustin:
Thanks a lot @anuraaga and @adriancole for a great idea and execution.

PR #1753, opened 2019-05-08
Add options to customize the JSON marshalling behavior of gRPC service/client

### LINKED ISSUE #1696: Support for setting preservingProtoFieldNames in GrpcService when both protobuf and json wire formats are used with the GRPC service
Currently the GrpcService initialisation via the builder doesn't allow the option of preserving Proto field names in the output. If we enable the default constructor of the GrpcService to allow the option of setting preservingProtoFieldNames then the same can be passed onto GrpcJsonUtil. This would be helpful if we don't need to modify every Proto file to include json_name option.

anuraaga:
…RPC service / client.

Fixes #1696


trustin:
Thank you, @anuraaga !

PR #1755, opened 2019-05-08
Do not set SO_KEEPALIVE socket option by default

trustin:
Motivation:

There's not really a need to set the `SO_KEEPALIVE` socket option in a
modern socket programming, because the keepalive interval is way longer
than what users usually want and it is often difficult to configure the
interval.

Modifications:

- Remove the code that sets `SO_KEEPALIVE` channel option on the client
  side.

Result:

We don't waste time setting an additional channel option.


trustin:
Thanks for reviewing :-)

PR #1761, opened 2019-05-10
Add withRoute(Consumer) and withVirtualHost(Consumer)

minwoox:
Motivation:
We introduced a new style for a method which takes a `Consumer` as a parameter and starts the name with `with`. (e.g `withMutations`)
It's probably good to consolidate on that throughout the codebase.

Modification:
- Add `withDefaultVirtualHost(Consumer)` and `withVirtualHost(Consumer)` to `ServerBuilder`.
- Add `withRoute(Consumer)` to `ServerBuilder`.
- Rename `RouteBuilder.service(Service)` to `build(Service)`.
- Rename `RouteBuilder` to `ServiceBindingBuilder`.
  - Will use `RouteBuilder` to build a `Route` which might wrap a `PathMapping` in the future.
- Deprecated
  - `withDefaultVirtualHost()` -> `defaultVirtualHost()` in `ServerBuilder`.
  - `withVirtualHost(...)` -> `virtualHost(...)` in `ServerBuilder`.

Result:
- Consistency


core/src/main/java/com/linecorp/armeria/server/ServiceBindingBuilder.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private final ServerBuilder serverBuilder;
+
+    ServiceBindingBuilder(ServerBuilder serverBuilder) {
+        this.serverBuilder = requireNonNull(serverBuilder, "serverBuilder");
+    }
+
+    @Override
+    public ServiceBindingBuilder path(String pathPattern) {
+        return (ServiceBindingBuilder) super.path(pathPattern);
```
anuraaga: Can we add a template parameter to the abstract class instead of having to add these in children?

minwoox: I couldn't use it because of the Javadoc API reference. 😰 
The abstract parent class is package-private so this child inherits the Javadoc from the parent. In the rendered API reference page, the return type is the template parameter (e.g `T` not `ServiceBindingBuilder`).
And I couldn't add a description like `@param <T> the self type` to the child  Javadoc.
We consider using self-describing template parameter such as `SELF` instead of `T` like AssertJ does. But probably it's not a common, so we chose this painful way. 😢

trustin:
Thanks @minwoox and reviewers.

PR #1762, opened 2019-05-10
Fix `http.url` Zipkin span tag and add `http.protocol` and `http.serfmt`

trustin:
Motivation:

The Zipkin span tag `http.url` has to be a valid addressable URL that
starts with `http://` or `https://`. We currently use `Scheme.uriText()`
which may be surprising to users.

Modifications:

- Make sure `http.url` starts with `http://` or `https://` for HTTP(S)
  endpoints.
- Add `http.protocol` and `http.serfmt`.

Result:

Less surprising behavior.

### DIFF
--- a/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanTags.java
+++ b/zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanTags.java
@@ -83,16 +83,23 @@ public final class SpanTags {
 
     private static String generateUrl(Scheme scheme, String authority, String path, @Nullable String query) {
         final SessionProtocol sessionProtocol = scheme.sessionProtocol();
-        final StringBuilder uriBuilder = new StringBuilder();
+        final String uriScheme;
         if (SessionProtocol.httpValues().contains(sessionProtocol)) {
-            uriBuilder.append("http://");
+            uriScheme = "http://";
         } else if (SessionProtocol.httpsValues().contains(sessionProtocol)) {
-            uriBuilder.append("https://");
+            uriScheme = "https://";
         } else {
-            uriBuilder.append(sessionProtocol.uriText()).append("://");
+            uriScheme = sessionProtocol.uriText() + "://";
         }
 
-        uriBuilder.append(authority).append(path);
+        final StringBuilder uriBuilder = new StringBuilder(
+                uriScheme.length() + authority.length() + path.length() +
+                (query != null ? query.length() + 1 : 0));
+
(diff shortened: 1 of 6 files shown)


zipkin/src/test/java/com/linecorp/armeria/client/tracing/HttpTracingClientTest.java
```diff
@@ -108,12 +106,13 @@ public void shouldSubmitSpanWhenSampled() throws Exception {
...
-                "http.path", "/hello/armeria",
-                "http.status_code", "200",
-                "http.url", "none+h2c://foo.com/hello/armeria"));
+        assertThat(span.tags()).containsEntry("http.host", "foo.com")
+                               .containsEntry("http.method", "POST")
+                               .containsEntry("http.path", "/hello/armeria")
+                               .containsEntry("http.status_code", "200")
+                               .containsEntry("http.url", "http://foo.com/hello/armeria")
+                               .containsEntry("http.protocol", "h2c")
+                               .containsEntry("http.serfmt", "none");
```
adriancole: as this creates index load, possibly add when not none?

trustin: Updated.

zipkin/src/main/java/com/linecorp/armeria/internal/tracing/SpanTags.java
```diff
@@ -73,6 +81,32 @@ public static void addTags(Span span, RequestLog log) {
         }
     }
 
+    private static String generateUrl(Scheme scheme, String authority, String path, @Nullable String query) {
+        final SessionProtocol sessionProtocol = scheme.sessionProtocol();
+        final String uriScheme;
+        if (SessionProtocol.httpValues().contains(sessionProtocol)) {
+            uriScheme = "http://";
+        } else if (SessionProtocol.httpsValues().contains(sessionProtocol)) {
+            uriScheme = "https://";
```
adriancole: is this one a pain to test? I noticed it wasn't in the test case, but not terribly fussed

trustin: I can add one for sure.

trustin: Added `SpanTagsTest` :smile:

trustin:
Thanks, reviewers :+1:

PR #1764, opened 2019-05-13
Add new endpoint selection strategy which increases weight gradually

imasahiro:


core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,135 @@
...
+     *   <li>if endpoint weights are 3,5,7, then select result is abcabcabcbcbcbb abcabcabcbcbcbb ...</li>
+     * </ul>
+     */
+    private final class SlowStartAwareWeightedRoundRobinSelector implements EndpointSelector {
+        private final EndpointGroup endpointGroup;
+        private final AtomicInteger sequence = new AtomicInteger();
+        private volatile EndpointsAndWeights endpointsAndWeights;
+        private final BiFunction<Endpoint, Duration, Integer> endpointWeighter;
+
+        private final Duration slowStartInterval = Duration.ofSeconds(10);
```
imasahiro: Will make following fields configurable after finish PoC impl

trustin: I think we should also make the number of steps configurable because a user may not want to adjust the weight too often (e.g. every millisecond)

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,135 @@
...
+        private final int maxWeight = 1000;
```
trustin: I'm not sure if we need to set this because the target weight of each endpoint can be determined by the original weight value of the underlying `EndpointGroup`. What do you think?

core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointSelectionStrategy.java
```diff
@@ -34,9 +38,19 @@
...
+            SlowStartAwareEndpointSelectionStrategy::new;
+
     /**
      * Creates a new {@link EndpointSelector} that selects an {@link Endpoint} from the specified
      * {@link EndpointGroup}.
      */
     EndpointSelector newSelector(EndpointGroup endpointGroup);
+
+    @FunctionalInterface
+    interface EndpointWeighter extends BiFunction<Endpoint, Duration, Integer> {
```
trustin: We could use `long` and `int` instead of `Duration` and `Integer` to reduce GC pressure during slow-start.

core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointSelectionStrategy.java
```diff
@@ -34,9 +38,19 @@
      */
     EndpointSelectionStrategy WEIGHTED_ROUND_ROBIN = new WeightedRoundRobinStrategy();
 
+    /**
+     * Weighted round-robin strategy that increases weight gradually.
+     */
+    Function<EndpointWeighter, EndpointSelectionStrategy> SLOW_START_AWARE_WEIGHTED_ROUND_ROBIN =
+            SlowStartAwareEndpointSelectionStrategy::new;
```
trustin: I guess a user can just call the constructor directly. What is this field useful for?

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,135 @@
...
+            final int currentSequence = sequence.getAndIncrement();
+            return endpointsAndWeights.selectEndpoint(currentSequence);
+        }
+
+        private void scheduleUpdateRequestWeight(List<Endpoint> currentlyEndpoints,
+                                                 List<Endpoint> previousEndpoints) {
+            if (slowStartInterval.isZero()) {
+                return;
+            }
+            if (updateEndpointWeightFuture != null && !updateEndpointWeightFuture.isDone()) {
```
trustin: No need to check if `isDone()` because `cancel()` does the check.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,135 @@
...
+                updateEndpointWeightFuture.cancel(true);
+            }
+            final Set<Endpoint> newEndpoints = Sets.difference(ImmutableSet.copyOf(previousEndpoints),
+                                                               ImmutableSet.copyOf(currentlyEndpoints));
+            final int step = Math.max((int) slowStartInterval.toMillis() / maxWeight, 1);
+
+            updateEndpointWeightFuture = clientFactory.eventLoopGroup()
+                                                      .schedule(() -> updateEndpointWeight(currentlyEndpoints,
+                                                                                           newEndpoints, step),
```
trustin: I think each endpoint must be tracked individually because otherwise the weight of the previously newly added endpoints will jump to the target weight immediately, e.g.

1. A new endpoint A with weight 1000 is added.
2. Slow-start for the endpoint A starts.
3. Another new endpoint B with weight 1000 is added.
4. Slow-start for the endpoint B starts, but the weight of the endpoint A jumps straight up to 1000.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        private void updateEndpointWeight(List<Endpoint> endpoints, Set<Endpoint> newEndpoints, int step) {
+            endpointsAndWeights = new EndpointsAndWeights(
+                    endpoints.stream()
+                             .map(e -> {
+                                 if (e.weight() >= maxWeight || !newEndpoints.contains(e)) {
+                                     return e;
+                                 }
+                                 return e.withWeight(Math.min(e.weight() + step, maxWeight));
```
trustin: The weight of each endpoint has to increase proportionally. For example, if we increase the weight for 10 seconds, two endpoints with the weight of 500 and 1000 should have the weight of 250 and 500 at 5-second point.

trustin:
The direction seems OK to me. Thanks for the initiative!

trustin:
Gentle ping :wink:

trustin:
@imasahiro Ready for a review?

imasahiro:
not yet, please give me more time to address all comments

imasahiro:
I think now ready for review.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+
+import com.linecorp.armeria.client.ClientFactory;
+import com.linecorp.armeria.client.ClientRequestContext;
+import com.linecorp.armeria.client.Endpoint;
+import com.linecorp.armeria.client.endpoint.WeightedRoundRobinStrategy.EndpointsAndWeights;
+
+import io.netty.util.concurrent.ScheduledFuture;
+
+/**
+ * A weighted round-robin strategy that lamp up the weight gradually from zero to its normal weight value.
```
trustin: lamp -> lamps

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+                                                   int numberOfSteps) {
+        this.endpointWeighter = endpointWeighter;
+        this.clientFactory = clientFactory;
+        this.slowStartInterval = requireNonNull(slowStartInterval, "slowStartInterval");
+        checkArgument(numberOfSteps > 0, "numberOfSteps(%s) > 0");
+        this.numberOfSteps = numberOfSteps;
+    }
+
+    @Override
+    public SlowStartAwareWeightedRoundRobinSelector newSelector(EndpointGroup endpointGroup) {
```
trustin: Could return `EndpointSelector` because `SlowStartAwareWeightedRoundRobinSelector` is private.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+public final class SlowStartAwareEndpointSelectionStrategy implements EndpointSelectionStrategy {
+    private final EndpointWeighter endpointWeighter;
+    private final ClientFactory clientFactory;
+    private final Duration slowStartInterval;
+    private final int numberOfSteps;
+
+    /**
+     * Creates a {@link SlowStartAwareEndpointSelectionStrategy}.
+     */
+    public SlowStartAwareEndpointSelectionStrategy(EndpointWeighter endpointWeighter,
```
trustin: - Could we use static factory methods instead of constructors?
- Could accept `scheduledExecutorService` instead of `ClientFactory`
- How about accepting the total duration instead of interval? I.e. “I want to ramp up the weight for 10 seconds.”
- `of(...)` could require essential properties only, e.g. `ScheduledExecutorService` and `slowStartDuration`
- `builder()` could allow a user to specify optional properties, e.g. number of steps, weighter

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+        int stepUp() {
+            return ++step;
+        }
+    }
+
+    /**
+     * Controls an {@link Endpoint} weight to ramp up a request load to the {@link Endpoint}.
+     */
+    @FunctionalInterface
+    public interface EndpointWeighter {
```
trustin: - Could we make this top-level?
- `EndpointWeightTransition`? Any better names?

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+        EndpointWeighter DEFAULT = (e, step, maxStep) -> (int) (e.weight() * (1.0 * step / maxStep * 100));
```
trustin: - Shouldn’t this be `(double) e.weight() * step / maxStep`? Why do wee multiply by 100?
- `linear()`?

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+
+        /**
+         * Computes an {@link Endpoint} weight based on original {@link Endpoint} weight, current/max steps.
+         */
+        int compute(Endpoint e, int step, int maxStep);
```
trustin: - e -> endpoint
- step -> currentSteps
- maxStep -> totalSteps

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+            if (endpointAndStep == null) {
+                return e;
+            }
+            final int originalWeight = e.weight();
+            final int newWeight = endpointWeighter.compute(e, endpointAndStep.stepUp(), numberOfSteps);
+            if (originalWeight <= newWeight) {
+                currentWeights.remove(e);
+                return e;
+            }
+            return e.withWeight(newWeight);
```
trustin: Could we also handle the case where the new weight is negative?

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        private Endpoint updateWeight(Endpoint e) {
+            final EndpointAndStep endpointAndStep = currentWeights.get(e);
+            if (endpointAndStep == null) {
+                return e;
+            }
+            final int originalWeight = e.weight();
+            final int newWeight = endpointWeighter.compute(e, endpointAndStep.stepUp(), numberOfSteps);
+            if (originalWeight <= newWeight) {
```
trustin: `newWeight >= originalWeight` for better readabilitiy?

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+        checkArgument(numberOfSteps > 0, "numberOfSteps: %s (expected: > 0)", numberOfSteps);
+        this.numberOfSteps = numberOfSteps;
+        return this;
+    }
+
+    /**
+     * Creates a {@link SlowStartAwareEndpointSelectionStrategy}.
+     */
+    public SlowStartAwareEndpointSelectionStrategy build() {
+        final Duration interval;
```
trustin: `interval` -> `duration`

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+        return this;
+    }
+
+    /**
+     * Sets an intervals during slow start.
+     */
+    public SlowStartAwareEndpointSelectionStrategyBuilder slowStartInterval(Duration slowStartInterval) {
+        if (totalSlowStartDuration != null) {
+            throw new IllegalArgumentException("totalSlowStartDuration is already given");
+        }
```
trustin: `checkState()`?

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+        this.executorService = requireNonNull(executorService, "executorService");
+        return this;
+    }
+
+    /**
+     * Sets a total duration of slow start.
+     */
+    public SlowStartAwareEndpointSelectionStrategyBuilder totalSlowStartDuration(
+            Duration totalSlowStartDuration) {
+        this.totalSlowStartDuration = requireNonNull(totalSlowStartDuration,
```
trustin: Could raise an `IllegalStateException` if `slowStartInterval` is given already?

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+    private EndpointWeightTransition endpointWeightTransition;
+    private ScheduledExecutorService executorService;
+    private Duration totalSlowStartDuration;
+    private Duration slowStartInterval;
+    private int numberOfSteps;
+
+    /**
+     * Sets an {@link EndpointWeightTransition}.
+     */
+    public SlowStartAwareEndpointSelectionStrategyBuilder endpointWeighter(
```
trustin: `weightTransition()`?

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+ * A builder for creating a new {@link SlowStartAwareEndpointSelectionStrategy} that ramps up the weight
+ * gradually.
+ */
+public final class SlowStartAwareEndpointSelectionStrategyBuilder {
+    private EndpointWeightTransition endpointWeightTransition;
+    private ScheduledExecutorService executorService;
+    private Duration totalSlowStartDuration;
+    private Duration slowStartInterval;
+    private int numberOfSteps;
+
```
trustin: Could make the default constructor package-private.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -24,42 +24,63 @@
...
+        return new SlowStartAwareEndpointSelectionStrategyBuilder();
+    }
+
+    /**
+     * Creates a {@link SlowStartAwareEndpointSelectionStrategy}.
+     */
+    public static EndpointSelectionStrategy of(EndpointWeightTransition endpointWeightTransition,
+                                               ScheduledExecutorService executorService,
+                                               Duration slowStartInterval,
+                                               int numberOfSteps) {
```
trustin: How about reducing the number of parameters of this method by introducing some default values? e.g. `weightTransition` could default to `linear()`. `numerOfSteps` could be `20`.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -126,12 +146,12 @@ private Endpoint updateWeight(Endpoint e) {
...
             final int originalWeight = e.weight();
-            final int newWeight = endpointWeighter.compute(e, endpointAndStep.stepUp(), numberOfSteps);
-            if (originalWeight <= newWeight) {
+            final int newWeight = endpointWeightTransition.compute(e, endpointAndStep.stepUp(), numberOfSteps);
+            if (newWeight >= originalWeight) {
                 currentWeights.remove(e);
                 return e;
             }
-            return e.withWeight(newWeight);
+            return e.withWeight(newWeight >= 0 ? newWeight : 0);
```
trustin: `Math.max(0, newWeight)`

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        @Override
+        public Endpoint select(ClientRequestContext ctx) {
+            final int currentSequence = sequence.getAndIncrement();
+            return endpointsAndWeights.selectEndpoint(currentSequence);
+        }
+
+        private void scheduleUpdateRequestWeight(List<Endpoint> newEndpoints) {
+            if (updateEndpointWeightFuture != null && !updateEndpointWeightFuture.cancel(false)) {
```
anuraaga: I think this method can be called concurrently, since `addListener` notifications can happen in quick succession and don't have any fixed thread model IIUC.

Simplest solution is probably to make this method `synchronized`.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * An {@link EndpointSelector} that is similar to {@link WeightedRoundRobinStrategy} but it
+     * increases each {@link Endpoint} weight gradually.
+     */
+    private final class SlowStartAwareWeightedRoundRobinSelector implements EndpointSelector {
+        private final EndpointGroup endpointGroup;
+        private final AtomicInteger sequence = new AtomicInteger();
+        private volatile EndpointsAndWeights endpointsAndWeights;
```
anuraaga: Think it's good to group `final` and `volatile`

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,175 @@
...
+                             .map(this::updateWeight)
+                             .collect(toImmutableList()));
+        }
+
+        private Endpoint updateWeight(Endpoint e) {
+            final EndpointAndStep endpointAndStep = currentWeights.get(e);
+            if (endpointAndStep == null) {
+                return e;
+            }
+            final int originalWeight = e.weight();
```
anuraaga: `definedWeight` - `original` implies it was previously this weight but it's actually our goal.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,175 @@
...
+        public Endpoint select(ClientRequestContext ctx) {
+            final int currentSequence = sequence.getAndIncrement();
+            return endpointsAndWeights.selectEndpoint(currentSequence);
+        }
+
+        private void scheduleUpdateRequestWeight(List<Endpoint> newEndpoints) {
+            if (updateEndpointWeightFuture != null && !updateEndpointWeightFuture.cancel(false)) {
+                updateEndpointWeightFuture = null;
+            }
+            newEndpoints.forEach(e -> currentWeights.put(e, new EndpointAndStep(e)));
```
anuraaga: Shouldn't we `computeIfAbsent` instead of `put` and remove keys not in `newEndpoints`? I think if an endpoint keeps on going down and up, this method will be called continuously and all endpoints will have near-zero weight forever, even though we just want that bad endpoint to have near-zero weight.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,175 @@
...
+
+        private Endpoint updateWeight(Endpoint e) {
+            final EndpointAndStep endpointAndStep = currentWeights.get(e);
+            if (endpointAndStep == null) {
+                return e;
+            }
+            final int originalWeight = e.weight();
+            final int newWeight = endpointWeightTransition.compute(e, endpointAndStep.stepUp(), numberOfSteps);
+            if (newWeight >= originalWeight) {
+                currentWeights.remove(e);
```
anuraaga: Based on my above comment, I think we want to keep the state that an endpoint has reached the defined weight and not reset it unless it goes away completely. I guess `EndpointAndStep` can have a flag marking that and instead of checking `null` above check the flag.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,175 @@
...
+                                                        0,
+                                                        slowStartInterval.toMillis(),
+                                                        TimeUnit.MILLISECONDS);
+        }
+
+        private void updateEndpointWeight(List<Endpoint> endpoints) {
+            endpointsAndWeights = new EndpointsAndWeights(
+                    endpoints.stream()
+                             .map(this::updateWeight)
+                             .collect(toImmutableList()));
```
anuraaga: We should add logic to cancel the updater when all the weights have finished ramping, it's wasteful to run this forever in an environment with relatively stable endpoints.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,175 @@
...
+    private final int numberOfSteps;
+
+    /**
+     * Creates a {@link SlowStartAwareEndpointSelectionStrategy}.
+     */
+    SlowStartAwareEndpointSelectionStrategy(EndpointWeightTransition endpointWeightTransition,
+                                            ScheduledExecutorService executorService,
+                                            Duration slowStartInterval,
+                                            int numberOfSteps) {
+        this.endpointWeightTransition = endpointWeightTransition;
```
anuraaga: I think these are also `requireNonNull`

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+import java.time.Duration;
+import java.util.concurrent.ScheduledExecutorService;
+
+/**
+ * A builder for creating a new {@link SlowStartAwareEndpointSelectionStrategy} that ramps up the weight
+ * gradually.
+ */
+public final class SlowStartAwareEndpointSelectionStrategyBuilder {
+    private EndpointWeightTransition endpointWeightTransition;
+    private ScheduledExecutorService executorService;
```
anuraaga: As @trustin mentioned these could have defaults. The service could default to common pools too.

core/src/main/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,175 @@
...
+
+        @Override
+        public EndpointSelectionStrategy strategy() {
+            return SlowStartAwareEndpointSelectionStrategy.this;
+        }
+
+        @Override
+        public Endpoint select(ClientRequestContext ctx) {
+            final int currentSequence = sequence.getAndIncrement();
+            return endpointsAndWeights.selectEndpoint(currentSequence);
```
anuraaga: I wonder if it's simple to mark all rampup is complete and delegate to normal endpoint selection at that point until next call to `scheduleUpdateRequestWeight`. I'd like to use this class with equal defined weights to handle rolling restart, so don't need to spend computation on weight calculation after the rampup is done.

core/src/test/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategyTest.java
```diff
@@ -0,0 +1,139 @@
...
+                                                           Duration.ofSeconds(1),
+                                                           10);
+
+        final TestDynamicEndpointGroup endpointGroup = new TestDynamicEndpointGroup();
+        EndpointGroupRegistry.register("dynamic", endpointGroup, strategy);
+        endpointGroup.updateEndpoints(ImmutableList.of(Endpoint.of("127.0.0.1", 1000)));
+
+        final EndpointSelector selector = EndpointGroupRegistry.getNodeSelector("dynamic");
+        // Schedule updateEndpointWeight.
+        TimeUnit.SECONDS.sleep(1);
```
anuraaga: It might be good to have a test that uses a real executor to make sure it is used, but I think making the `updateWeight` method `@VisibleForTesting` and testing a complex series of transitions without `await()` and sleeps would be easier to reason about.

core/src/test/java/com/linecorp/armeria/client/endpoint/SlowStartAwareEndpointSelectionStrategyTest.java
```diff
@@ -0,0 +1,139 @@
...
+                   assertThat(e.port()).isEqualTo(1111);
+                   assertThat(e.weight()).isGreaterThanOrEqualTo(1);
+               });
+        await().atMost(Duration.ofSeconds(1))
+               .untilAsserted(() -> assertThat(selector.select(ctx).weight()).isGreaterThanOrEqualTo(10));
+        await().atMost(Duration.ofSeconds(1))
+               .untilAsserted(() -> assertThat(selector.select(ctx).weight()).isGreaterThanOrEqualTo(20));
+        await().atMost(Duration.ofSeconds(1))
+               .untilAsserted(() -> assertThat(selector.select(ctx).weight()).isGreaterThanOrEqualTo(30));
+        // Set known endpoint but the weight continue to rump up with previous weight.
```
anuraaga: You should use a group with two endpoints and change one of them to verify. As I mentioned above, I don't think this is actually implemented yet. The test passes because `Endpoint.equals` doesn't take weight into account so the listener isn't notified (endpoints didn't change).

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/Endpoint.java#L646

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java#L100

core/src/main/java/com/linecorp/armeria/client/endpoint/GradualStartEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+     * Creates a {@link GradualStartEndpointSelectionStrategyBuilder}.
+     */
+    public static GradualStartEndpointSelectionStrategyBuilder builder() {
+        return new GradualStartEndpointSelectionStrategyBuilder();
+    }
+
+    /**
+     * Creates a {@link GradualStartEndpointSelectionStrategy}.
+     */
+    public static EndpointSelectionStrategy of(Duration slowStartInterval) {
```
trustin: I think total duration is better as a parameter in this case because the number of steps is not clear here.

core/src/main/java/com/linecorp/armeria/client/endpoint/GradualStartEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * A builder for creating a new {@link GradualStartEndpointSelectionStrategy} that ramps up the weight
+ * gradually.
+ */
+public final class GradualStartEndpointSelectionStrategyBuilder {
+    private EndpointWeightTransition endpointWeightTransition;
+    private ScheduledExecutorService executorService;
+    private Duration totalSlowStartDuration;
+    private Duration slowStartInterval;
+    private int numberOfSteps;
```
trustin: Could we set some defaults here and simplify `GradualStartEndpointSelectionStrategy.of()`?

core/src/test/java/com/linecorp/armeria/client/endpoint/WeightedRoundRobinStrategyTest.java
```diff
@@ -257,7 +257,7 @@ public void selectFromDynamicEndpointGroup() {
         assertThat(selector.select(ctx)).isEqualTo(Endpoint.of("127.0.0.1", 1111).withWeight(1));
     }
 
-    private static final class TestDynamicEndpointGroup extends DynamicEndpointGroup {
+    static final class TestDynamicEndpointGroup extends DynamicEndpointGroup {
```
trustin: Could be moved top-level?

core/src/test/java/com/linecorp/armeria/client/endpoint/GradualStartEndpointSelectionStrategyTest.java
```diff
@@ -0,0 +1,135 @@
...
+                   assertThat(e.port()).isEqualTo(1111);
+                   assertThat(e.weight()).isGreaterThanOrEqualTo(1);
+               });
+        await().atMost(Duration.ofSeconds(1))
+               .untilAsserted(() -> assertThat(selector.select(ctx).weight()).isGreaterThanOrEqualTo(10));
+        await().atMost(Duration.ofSeconds(1))
+               .untilAsserted(() -> assertThat(selector.select(ctx).weight()).isGreaterThanOrEqualTo(20));
+        await().atMost(Duration.ofSeconds(1))
+               .untilAsserted(() -> assertThat(selector.select(ctx).weight()).isGreaterThanOrEqualTo(30));
+        // Set known endpoint but the weight continue to rump up with previous weight.
```
trustin: - to rump -> ramping
- the weight continue -> the weight should continue

What would happen if the new endpoint has the same host and port but with different weight? Could you add a test case for it as well?

core/src/main/java/com/linecorp/armeria/client/endpoint/GradualStartEndpointSelectionStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+     * Creates a {@link GradualStartEndpointSelectionStrategy}.
+     */
+    GradualStartEndpointSelectionStrategy(EndpointWeightTransition endpointWeightTransition,
+                                          ScheduledExecutorService executorService,
+                                          Duration slowStartInterval,
+                                          int numberOfSteps) {
+        this.endpointWeightTransition = requireNonNull(endpointWeightTransition, "endpointWeightTransition");
+        this.executorService = requireNonNull(executorService, "executorService");
+        this.slowStartInterval = requireNonNull(slowStartInterval, "slowStartInterval");
+        checkArgument(numberOfSteps > 0, "numberOfSteps(%s) > 0");
```
trustin: `"numberOfSteps: %s (expected: > 0)", numberOfSteps`

core/src/main/java/com/linecorp/armeria/client/endpoint/GradualStartEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+        this.executorService = requireNonNull(executorService, "executorService");
+        return this;
+    }
+
+    /**
+     * Sets a total duration of slow start.
+     */
+    public GradualStartEndpointSelectionStrategyBuilder totalSlowStartDuration(
+            Duration totalSlowStartDuration) {
+        checkState(slowStartInterval == null, "slowStartInterval is already given");
```
trustin: nit: Missing period in a sentence

core/src/main/java/com/linecorp/armeria/client/endpoint/GradualStartEndpointSelectionStrategyBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+        this.totalSlowStartDuration = requireNonNull(totalSlowStartDuration,
+                                                     "totalSlowStartDuration");
+        return this;
+    }
+
+    /**
+     * Sets an intervals during slow start.
+     */
+    public GradualStartEndpointSelectionStrategyBuilder slowStartInterval(Duration slowStartInterval) {
+        checkState(totalSlowStartDuration == null, "totalSlowStartDuration is already given");
```
trustin: nit: Missing period in a sentence

trustin:
This PR may need some changes if #2235 is merged.

trustin:
Closing in favor of #3217. Thanks a lot for trying, @imasahiro !

PR #1766, opened 2019-05-14
Add AbstractUnsafeUnaryGrpcService for really advanced and unsafe users.

anuraaga:
Users like zipkin can use access to unsafe buffers to maximize performance, and it's not so hard to support it so may as well.

/cc @adriancole


hyangtack:
Thanks, @anuraaga !

PR #1770, opened 2019-05-15
Return pooled buffers from HTTP decoded responses when requested.

anuraaga:
Currently, HTTP decoding always returns unpooled data, even if a user requested pooled objects. This allows HTTP decoding to return pooled objects if any of the inputs were pooled, which indicates the user requested pooled objects.

Also added a minor optimization for the common case where calls to get the decoded output only return one buffer.

Came up in https://github.com/apache/incubator-zipkin/pull/2595

/cc @adriancole

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/encoding/GzipStreamDecoderFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/encoding/GzipStreamDecoderFactory.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.client.encoding;
 
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.handler.codec.compression.ZlibWrapper;
 
 /**
@@ -29,7 +30,7 @@ public class GzipStreamDecoderFactory implements StreamDecoderFactory {
     }
 
     @Override
-    public StreamDecoder newDecoder() {
-        return new ZlibStreamDecoder(ZlibWrapper.GZIP);
+    public StreamDecoder newDecoder(ByteBufAllocator alloc) {
+        return new ZlibStreamDecoder(ZlibWrapper.GZIP, alloc);
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodedResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodedResponse.java
@@ -32,20 +32,25 @@ import com.linecorp.armeria.common.HttpObject;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.internal.ArmeriaHttpUtil;
 
+import io.netty.buffer.ByteBufAllocator;
+
 /**
  * A {@link FilteredHttpResponse} that applies HTTP decoding to {@link HttpObject}s as they are published.
  */
 class HttpDecodedResponse extends FilteredHttpResponse {
 
     private final Map<String, StreamDecoderFactory> availableDecoders;
+    private final ByteBufAllocator alloc;
 
     @Nullable
     private StreamDecoder responseDecoder;
     private boolean headersReceived;
 
-    HttpDecodedResponse(HttpResponse delegate, Map<String, StreamDecoderFactory> availableDecoders) {
+    HttpDecodedResponse(HttpResponse delegate, Map<String, StreamDecoderFactory> availableDecoders,
+                        ByteBufAllocator alloc) {
         super(delegate);
         this.availableDecoders = availableDecoders;
+        this.alloc = alloc;
     }
 
     @Override
@@ -78,7 +83,7 @@ class HttpDecodedResponse extends FilteredHttpResponse {
                 // If the server returned an encoding we don't support (shouldn't happen since we set
                 // Accept-Encoding), decoding will be skipped which is ok.
                 if (decoderFactory != null) {
-                    responseDecoder = decoderFactory.newDecoder();
+                    responseDecoder = decoderFactory.newDecoder(alloc);
                 }
             }
 
--- a/core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodedResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodedResponse.java
@@ -48,7 +48,7 @@ class HttpDecodedResponse extends FilteredHttpResponse {
 
     HttpDecodedResponse(HttpResponse delegate, Map<String, StreamDecoderFactory> availableDecoders,
                         ByteBufAllocator alloc) {
-        super(delegate);
+        super(delegate, false);
         this.availableDecoders = availableDecoders;
         this.alloc = alloc;
     }
--- a/core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodedResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodedResponse.java
@@ -48,7 +48,7 @@ class HttpDecodedResponse extends FilteredHttpResponse {
 
     HttpDecodedResponse(HttpResponse delegate, Map<String, StreamDecoderFactory> availableDecoders,
                         ByteBufAllocator alloc) {
-        super(delegate, false);
+        super(delegate, true);
         this.availableDecoders = availableDecoders;
         this.alloc = alloc;
     }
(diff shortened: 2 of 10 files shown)


core/src/main/java/com/linecorp/armeria/client/encoding/ZlibStreamDecoder.java
```diff
@@ -34,34 +35,41 @@
...
 
     private final EmbeddedChannel decoder;
+    private final ByteBufAllocator alloc;
 
-    ZlibStreamDecoder(ZlibWrapper zlibWrapper) {
+    private boolean withPooledObjects;
+
+    ZlibStreamDecoder(ZlibWrapper zlibWrapper, ByteBufAllocator alloc) {
         decoder = new EmbeddedChannel(false, ZlibCodecFactory.newZlibDecoder(zlibWrapper));
+        this.alloc = alloc;
```
hyangtack: Just curious. It seems not used. Did you intend it?

anuraaga: Oops, thanks for noticing! Was originally allocating buffers but ended up using the embedded channel's buffer instead, so I need to pass the alloc to it :)

core/src/main/java/com/linecorp/armeria/client/encoding/HttpDecodedResponse.java
```diff
@@ -32,20 +32,25 @@
...
+    private final ByteBufAllocator alloc;
 
     @Nullable
     private StreamDecoder responseDecoder;
     private boolean headersReceived;
 
-    HttpDecodedResponse(HttpResponse delegate, Map<String, StreamDecoderFactory> availableDecoders) {
+    HttpDecodedResponse(HttpResponse delegate, Map<String, StreamDecoderFactory> availableDecoders,
+                        ByteBufAllocator alloc) {
         super(delegate);
```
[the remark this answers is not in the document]
anuraaga: Thanks for finding this! I was happy to test with a unit test, but I guess I need to add a higher level test too ;) Let me try to fix this.

anuraaga:
It ended up being more complicated than I was expecting. One finding I had is it's not appropriate to check `instanceof ByteBufHolder` to know if it's pooled or not - that may have content that is `UnpooledByteBuf`. I realized that it's not the job of a publisher in our model to worry about pooled or not, it can publish all pooled objects if it wants, and it's the job of a subscriber to unpool as needed, so I changed the logic to always return pooled objects from the decoder, and `FilterSubscriber` now unpools as appropriate.

core/src/test/java/com/linecorp/armeria/client/encoding/HttpDecodedResponseTest.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final Map<String, StreamDecoderFactory> DECODERS =
+            ImmutableMap.of("gzip", new GzipStreamDecoderFactory());
+
+    private static final ResponseHeaders RESPONSE_HEADERS =
+            ResponseHeaders.of(HttpStatus.OK, HttpHeaderNames.CONTENT_ENCODING, "gzip");
+
+    private static final byte[] PAYLOAD;
+
+    static {
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
```
hyangtack: nit: Please add `final` to the immutable variables in this class. 😆

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -66,7 +66,7 @@ protected FilteredStreamMessage(StreamMessage<T> delegate) {
     protected FilteredStreamMessage(StreamMessage<T> delegate, boolean withPooledObjects) {
         requireNonNull(delegate, "delegate");
         this.delegate = delegate;
-        this.withPooledObjects = withPooledObjects;
+        this.filterSupportsPooledObjects = withPooledObjects;
```
hyangtack: nit: How about changing the parameter name as well?

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -66,7 +66,7 @@ protected FilteredStreamMessage(StreamMessage<T> delegate) {
     protected FilteredStreamMessage(StreamMessage<T> delegate, boolean withPooledObjects) {
         requireNonNull(delegate, "delegate");
         this.delegate = delegate;
```
hyangtack: nit: Not a part of your PR, but please change here to `this.delegate = requireNonNull(delegate, "delegate");` for simplicity. 🙇

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
anuraaga: I thought about it, but what I felt is that for users of this constructor (subclasses), they care about withPooledObjects or not so it seems like a clear name, but for implementors of this class, it's confusing to have both that and the subscriber one and I renamed it for the sake of this implementation only. I'd lean towards keeping the API parameter as is.

hyangtack: I agree. Thank you for explanation!

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -190,10 +191,14 @@ public void onSubscribe(Subscription s) {
         @Override
         public void onNext(T o) {
             ReferenceCountUtil.touch(o);
-            if (!withPooledObjects) {
+            if (!filterSupportsPooledObjects) {
```
hyangtack: I'm confusing, is this parameter really necessary for `FilteredStreamMessage`? As you said [this comment](https://github.com/line/armeria/pull/1770#issuecomment-493003539), this is not a final subscriber, so it seems unnecessary to care whether the buffer is pooled or not from here. Let me think about this more tomorrow. 🤔

anuraaga: For our current implementations, nope but if a user implemented their own filtering response, it may not be aware of pooled objects, so the input to `filter` would be leaked. I think that's why we have the default that the filter does not support pooled objects, and an advanced user would have to opt-in with this parameter.

minwoox: Yeah, I agree with you on that. :)

hyangtack: Basically I also agree with @anuraaga's comment. 👍 Then, don't we need to handle the case which might copy the buffer twice?

I'm not sure, but is it okay for here?
```
if (!filterSupportsPooledObjects || !subscribedWithPooledObjects) {
    o = PooledObjects.toUnpooled(o);
}
```
, or like this?
```
if (!subscribedWithPooledObjects &&
    (o != filtered || filterSupportsPooledObjects)) {
    filtered = PooledObjects.toUnpooled(filtered);
}
```
What do you think? @minwoox @anuraaga

minwoox: Ah that's a good point! We shouldn't call toUnpooled if we already did.
So it could be:
```java
if (!filterSupportsPooledObjects) {
    o = PooledObjects.toUnpooled(o);
}
U filtered = filter(o);
if (filterSupportsPooledObjects && !subscribedWithPooledObjects) {
    filtered = PooledObjects.toUnpooled(filtered);
}
```
?

anuraaga: Calling `toUnpooled` on an unpooled object unnecessarily does a couple of extra instance checks but seems safe. I thought about that conditional but don't think there is any hard contract that just because an implementation didn't pass `withPooledObjects`, meaning it's input is never pooled, it wasn't allowed to publish pooled objects. If we wanted to make this our behavior, we should update the javadoc of the parameter to explicitly state this (it currently only talks about input). The overhead for unpooling an unpooled one isn't so bad so we could keep it simpler or add that javadoc - what do you think?

hyangtack: Ok. Let's leave it as it is. Thank you!

minwoox: That makes sense. Thank!

core/src/main/java/com/linecorp/armeria/client/encoding/GzipStreamDecoderFactory.java
```diff
@@ -29,7 +30,7 @@ public String encodingHeaderValue() {
     }
 
     @Override
-    public StreamDecoder newDecoder() {
```
adriancole: this probably doesn't impact many folks, but this is an api break I think

minwoox: Ah, thanks! Added it in the release note.

Issue #1711, opened 2019-04-11
TLS Protocols & Ciphers with Spring Boot


gquintana:
This is related to issue #1659 

HTTPS is working properly until I set armeria.ssl.enabled-protocols or armeria.ssl.ciphers

Here is my spring boot configuration
```
server:
  port: 9411
  ssl:
    enabled: true
    key-alias: "myhost.mycompany.fr"
    # ciphers: "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,TLS_DHE_RSA_WITH_AES_256_GCM_SHA384,TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_256_CBC_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDH_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDH_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDH_ECDSA_WITH_AES_256_CBC_SHA384,TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDH_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDH_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDH_RSA_WITH_AES_256_CBC_SHA384,TLS_ECDH_RSA_WITH_AES_128_CBC_SHA256,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_256_CBC_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA256"
    #enabled-protocols: "TLSv1.2"
    key-store: "/opt/zipkin-2.12.8/keystore.jks"
    key-store-password: "changeme"
[...]

PR #1777, opened 2019-05-16
Add a flag that can dump available ciphers on startup.

anuraaga:
Currently it is very difficult to know what ciphers are supported by a given armeria-based server that uses OpenSSL (JDK users can look at JDK's documentation) because OpenSSL may be boring, or system, depending on configuration, and the version always changes. This PR adds a flag to dump information about ciphers so users can see the exact ciphers that would be used in their environment.

I tried also using `getEnabledProtocols` to see what protocols are enabled by default, but it always seems to return SSLV2 and TLS1, 1.1, even though when I use `testssl.sh`, it always only shows TLS1.2, so for now I haven't tried logging that since I don't know how to get the right value.

`-Dcom.linecorp.armeria.dumpOpenSslInfo=true`

For #1711

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -18,6 +18,7 @@ package com.linecorp.armeria.common;
 
 import java.io.IOException;
 import java.nio.channels.ClosedChannelException;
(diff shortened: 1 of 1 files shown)


minwoox:
Thanks, @anuraaga!

PR #1779, opened 2019-05-17
Consolidate SSL logic and enable TLSv1.3 by default.

anuraaga:
While working on other SSL stuff, I noticed that the logic for SSL configuration is in two places and a bit confusing. And it's not too hard to enable TLSv1.3, so figured may as well turn it on since it seems to be getting more widely used now. FWIU, openssl intends users to automatically get it enabled when updating their dependency version, so figure similar for Armeria should be acceptable, but let me know if we should flag this. For reference, https://wiki.openssl.org/index.php/TLS1.3

No rush on this PR :)

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
@@ -42,7 +42,6 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Ascii;
 
-import com.linecorp.armeria.common.Flags;
 import com.linecorp.armeria.common.HttpObject;
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.logging.RequestLogBuilder;
@@ -50,6 +49,7 @@ import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.internal.ChannelUtil;
 import com.linecorp.armeria.internal.Http1ClientCodec;
 import com.linecorp.armeria.internal.ReadSuppressingHandler;
+import com.linecorp.armeria.internal.SslContextUtil;
 import com.linecorp.armeria.internal.TrafficLoggingHandler;
 
 import io.netty.buffer.ByteBuf;
@@ -87,17 +87,13 @@ import io.netty.handler.codec.http2.Http2ConnectionEncoder;
 import io.netty.handler.codec.http2.Http2Exception;
 import io.netty.handler.codec.http2.Http2FrameReader;
 import io.netty.handler.codec.http2.Http2FrameWriter;
-import io.netty.handler.codec.http2.Http2SecurityUtil;
 import io.netty.handler.codec.http2.Http2Settings;
 import io.netty.handler.flush.FlushConsolidationHandler;
-import io.netty.handler.ssl.ApplicationProtocolConfig;
 import io.netty.handler.ssl.ApplicationProtocolNames;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.ssl.SslHandshakeCompletionEvent;
-import io.netty.handler.ssl.SslProvider;
-import io.netty.handler.ssl.SupportedCipherSuiteFilter;
 import io.netty.util.AsciiString;
 import io.netty.util.ReferenceCountUtil;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
```diff
@@ -141,24 +137,9 @@
...
-                builder.sslProvider(
-                        Flags.useOpenSsl() ? SslProvider.OPENSSL : SslProvider.JDK);
+                SslContextUtil.configureDefaults(builder, httpPreference == HttpPreference.HTTP1_REQUIRED);
                 clientFactory.sslContextCustomizer().accept(builder);
 
-                if (httpPreference == HttpPreference.HTTP2_REQUIRED ||
-                    httpPreference == HttpPreference.HTTP2_PREFERRED) {
-
-                    builder.ciphers(Http2SecurityUtil.CIPHERS, SupportedCipherSuiteFilter.INSTANCE)
-                           .applicationProtocolConfig(new ApplicationProtocolConfig(
```
trustin: Can we make sure these two calls (ciphers and applicationProtocolConfig), or at least applicationProtocolConfig call, are made after the customizer is applied? Otherwise we would end up violating HTTP/2 specification IIUC.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -254,10 +251,7 @@ private static SslContext buildSslContext(
             return BouncyCastleKeyFactoryProvider.call(() -> {
                 final SslContextBuilder builder = builderSupplier.get();
 
-                builder.sslProvider(Flags.useOpenSsl() ? SslProvider.OPENSSL : SslProvider.JDK);
-                builder.ciphers(Http2SecurityUtil.CIPHERS, SupportedCipherSuiteFilter.INSTANCE);
-                builder.applicationProtocolConfig(HTTPS_ALPN_CFG);
```
trustin: This has to be called after the customizer.

core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
anuraaga: Ah thanks - I was leaning towards allowing users to have maximize ability to customize, but makes sense not to let them break HTTP/2 compatibility.

core/src/main/java/com/linecorp/armeria/internal/SslContextUtil.java
```diff
@@ -0,0 +1,83 @@
...
+                                         boolean forceHttp1,
+                                         Consumer<? super SslContextBuilder> userCustomizer) {
+        sslContext.sslProvider(Flags.useOpenSsl() ? SslProvider.OPENSSL : SslProvider.JDK)
+                  .protocols(DEFAULT_PROTOCOLS.toArray(new String[0]));
+
+        // We call user customization logic before setting ciphers and ALPN to make sure they don't break
+        // compatibility with HTTP/2.
+        userCustomizer.accept(sslContext);
+
+        sslContext.ciphers(DEFAULT_CIPHERS, SupportedCipherSuiteFilter.INSTANCE);
```
trustin: I'm not completely sure, but I guess a user might want to customize the cipher suite to some extent. Maybe better moving this before .accept()?

trustin: According to the specification, it seems to define the blacklist and some required ciphers: https://httpwg.org/specs/rfc7540.html#rfc.section.9.2

trustin: .. which means we could allow setting cipher suite and add some validation logic on our side.

anuraaga: Cool went with allowing cipher customization but validating HTTP/2 compatibility.

core/src/main/java/com/linecorp/armeria/internal/SslContextUtil.java
```diff
@@ -0,0 +1,405 @@
...
+        final SslContextBuilder builder = sslContextSupplier.get();
+
+        builder.sslProvider(Flags.useOpenSsl() ? SslProvider.OPENSSL : SslProvider.JDK)
+                  .protocols(DEFAULT_PROTOCOLS.toArray(new String[0]));
+
+        builder.ciphers(DEFAULT_CIPHERS, SupportedCipherSuiteFilter.INSTANCE);
+
+        userCustomizer.accept(builder);
+
+        // We call user customization logic before setting ALPN to make sure they don't break
```
hyangtack: Seems better to move this comment the above? (L81)

anuraaga: I think this location is clearer since it explains why we're setting ALPN here, the customizer call actually isn't the main point here. But changed the verb to `called` in case it makes it more natural.

trustin:
Thanks, @anuraaga

anuraaga:
The default protocols and ciphers work with TLS1.2 and 1.3, so don't think users should need to do anything special in the new version - some insecure ciphers were disabled, and I don't think we'd want to add documentation on how to add back insecure ciphers. 

Looking at Slack seems like the TLS failure you're looking at repros in the old version too, so guess we don't need to do anything here.

adriancole:
yeah at this moment I think my inability to configure SSL in a way that works unrelated to this change :P

adriancole:
Just to loop back. I confirm both TLS 1.2 and 1.3 work using curl out-of-box.

For the curious, [here's a gist](https://gist.github.com/adriancole/f07bd093dffc51a2b8a87d179505dde5) for making a build of curl on OS/X that can use TLS 1.3.

Here's a setup for an armeria spring boot which you can test with:

```bash
# assuming you generate the key like this
keytool -genkeypair -alias your_alias -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore your_keystore.p12 -validity 3650

# enter the following overrides, note the password is key-store-password not key-password
java -jar your_boot_app.jar --armeria.ssl.key-store=your_keystore.p12 --armeria.ssl.key-store-type=PKCS12 --armeria.ssl.key-store-password=123123 --armeria.ssl.key-alias=your_alias  --armeria.ssl.enabled=true --armeria.ports[0].port=9411 --armeria.ports[0].protocols[0]=https
```

PR #1782, opened 2019-05-21
Fix additional trailers are not added for trailers-only responses and add some advanced gRPC unit tests.

anuraaga:
Currently, trailers are not added when the response is trailers-only. This means a user's call to `addAdditionalResponseTrailer` may be ignored based on whether the response has a stream of objects or only the trailers, which can be hard for gRPC users where both situations occur.

Also added a unit test demonstrating adding a protobuf to response trailers and using a server interceptor to replace exceptions.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -172,7 +172,7 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                 final HttpHeaders additionalHeaders = reqCtx.additionalResponseHeaders();
                 final HttpHeaders additionalTrailers = reqCtx.additionalResponseTrailers();
 
-                ResponseHeadersBuilder newHeaders = fillAdditionalHeaders(headers, additionalHeaders);
+                final ResponseHeadersBuilder newHeaders = fillAdditionalHeaders(headers, additionalHeaders);
                 if (endOfStream && !additionalTrailers.isEmpty()) {
                     newHeaders.setIfAbsent(additionalTrailers);
                 }
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -170,7 +170,13 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                 }
 
                 final HttpHeaders additionalHeaders = reqCtx.additionalResponseHeaders();
-                final ResponseHeadersBuilder newHeaders = fillAdditionalHeaders(headers, additionalHeaders);
+                final HttpHeaders additionalTrailers = reqCtx.additionalResponseTrailers();
+
+                ResponseHeadersBuilder newHeaders = fillAdditionalHeaders(headers, additionalHeaders);
+                if (endOfStream && !additionalTrailers.isEmpty()) {
+                    newHeaders.setIfAbsent(additionalTrailers);
+                }
+
                 if (newHeaders.contains(HttpHeaderNames.CONTENT_LENGTH) &&
                     !reqCtx.additionalResponseTrailers().isEmpty()) {
                     newHeaders.remove(HttpHeaderNames.CONTENT_LENGTH);
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -169,11 +169,6 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                     break;
                 }
 
-                final HttpHeaders additionalHeaders = reqCtx.additionalResponseHeaders();
-                final HttpHeaders additionalTrailers = reqCtx.additionalResponseTrailers();
-
-                final ResponseHeadersBuilder newHeaders = fillAdditionalHeaders(headers, additionalHeaders);
-
                 if (req.method() == HttpMethod.HEAD) {
                     // HEAD responses always close the stream with the initial headers, even if not explicitly
                     // set.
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -170,7 +170,13 @@ public void onNext(HttpObject o) {
...
 
                 final HttpHeaders additionalHeaders = reqCtx.additionalResponseHeaders();
+                final HttpHeaders additionalTrailers = reqCtx.additionalResponseTrailers();
+
                 final ResponseHeadersBuilder newHeaders = fillAdditionalHeaders(headers, additionalHeaders);
+                if (endOfStream && !additionalTrailers.isEmpty()) {
+                    newHeaders.setIfAbsent(additionalTrailers);
+                }
+
                 if (newHeaders.contains(HttpHeaderNames.CONTENT_LENGTH) &&
```
anuraaga: I'm a bit unsure what this logic is doing - does anyone know why does the presence of additional trailers affect the content length header?

minwoox: Yeah the logic was added in https://github.com/line/armeria/pull/1566
It's because curl does not get HTTP trailers when Content-length exists. (i.e Should send the response with chunked transfer encoding.)

anuraaga: Ah - it's because trailers are only defined for chunked encoding, and armeria only applies chunked encoding if content-length isn't set. Added a comment to make that clearer.

minwoox: Yeah, thanks for the comment! 😄

grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
```diff
@@ -285,8 +302,47 @@ public void streamClientCancelsBeforeResponseClosedCancels(
...
+                                                                  .build()
+                                                                  .toByteArray()));
+            responseObserver.onError(new IllegalStateException("This error should have metadata"));
+        }
     }
 
+    private static final ServerInterceptor REPLACE_EXCEPTION = new ServerInterceptor() {
+        @Override
+        public <REQ, RESP> Listener<REQ> interceptCall(ServerCall<REQ, RESP> call, Metadata headers,
+                                                          ServerCallHandler<REQ, RESP> next) {
```
minwoox: nit: indentation?

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -170,9 +170,18 @@ public void onNext(HttpObject o) {
                 }
 
                 final HttpHeaders additionalHeaders = reqCtx.additionalResponseHeaders();
+                final HttpHeaders additionalTrailers = reqCtx.additionalResponseTrailers();
+
                 final ResponseHeadersBuilder newHeaders = fillAdditionalHeaders(headers, additionalHeaders);
+                if (endOfStream && !additionalTrailers.isEmpty()) {
```
hyangtack: Just curious. Would it be better to do this after [this code block](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java#L183-L201)? Because `endOfStream` flag can be updated in that code block.

anuraaga: Ah thanks for noticing - moved to after that block

hyangtack: Seems that the L172-175 can also be moved to here. :-)

grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
```diff
@@ -285,8 +302,47 @@ public void streamClientCancelsBeforeResponseClosedCancels(
...
+        @Override
+        public void errorReplaceException(SimpleRequest request,
+                                          StreamObserver<SimpleResponse> responseObserver) {
+            responseObserver.onError(new IllegalStateException("This error should be replaced"));
+        }
+
+        @Override
+        public void errorAdditionalMetadata(SimpleRequest request,
+                                            StreamObserver<SimpleResponse> responseObserver) {
+            ServiceRequestContext ctx = RequestContext.current();
```
hyangtack: nit: `final`

grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
```diff
@@ -1009,6 +1067,28 @@ public void onCompleted() {}
...
+        assertThatThrownBy(() -> blockingClient.errorReplaceException(SimpleRequest.getDefaultInstance()))
+                .isInstanceOf(StatusRuntimeException.class)
+                .hasMessage("UNKNOWN: Error was replaced");
+
+        requestLogQueue.take();
+    }
+
+    @Test
+    public void errorAdditionalMetadata() throws Exception {
+        Throwable t = catchThrowable(
```
hyangtack: nit: `final`

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -169,36 +169,46 @@ public void onNext(HttpObject o) {
...
+                if (req.method() == HttpMethod.HEAD) {
+                    // HEAD responses always close the stream with the initial headers, even if not explicitly
+                    // set.
+                    endOfStream = true;
+                } else {
+                    final int statusCode = status.code();
+                    switch (statusCode) {
+                        case 204:
+                        case 205:
+                        case 304:
```
trustin: Curious if we can reuse `ArmeriaHttpUtil.isContentAlwaysEmpty(int)`. It has a check for informational class statuses, though.

anuraaga: Good idea I think it makes sense - using the helper now. Also touched the helper a little for better readability / less duplication.

trustin:
Thanks, @anuraaga

PR #1783, opened 2019-05-21
Introduce `InvalidSamlRequestException` to distinguish invalid request errors…

### LINKED ISSUE #1780: Throw another exception when SAML module failed to find `SAMLResponse` parameter
... in order to handle it differently when sending an HTTP response for login failure.

- Related to https://github.com/line/centraldogma/issues/379
- https://github.com/line/armeria/blob/master/saml/src/main/java/com/linecorp/armeria/server/saml/SamlService.java#L215

hyangtack:
… from server errors

Motivation:
Currently, there is no way to distinguish between invalid request errors and server errors in the `SamlSingleSignOnHandler`.

Modifications:
- Added `InvalidSamlRequestException` for invalid request errors.
- It will be thrown when the request has an error.

Result:
- Closes #1780 
- A user can distinguish between invalid request errors and server errors in the `SamlSingleSignOnHandler`.

### DIFF
--- a/saml/src/main/java/com/linecorp/armeria/server/saml/InvalidSamlRequestException.java
+++ b/saml/src/main/java/com/linecorp/armeria/server/saml/InvalidSamlRequestException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 LINE Corporation
+ * Copyright 2019 LINE Corporation
  *
  * LINE Corporation licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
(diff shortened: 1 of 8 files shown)


saml/src/main/java/com/linecorp/armeria/server/saml/InvalidSamlRequestException.java
```diff
@@ -0,0 +1,52 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.saml;
+
+import javax.annotation.Nullable;
+
+/**
+ * Indicates that a SAML request is not valid so that its problem should be propagated to the sender.
```
minwoox: nit: probably we can remove the clause after `not valid`?

hyangtack: Yeah, it'd be better to remove that. Thanks!

saml/src/main/java/com/linecorp/armeria/server/saml/InvalidSamlRequestException.java
```diff
@@ -0,0 +1,52 @@
...
+    public InvalidSamlRequestException(@Nullable String message, @Nullable Throwable cause) {
+        super(message, cause);
+    }
+
+    /**
+     * Creates a new instance with the specified {@code cause}.
+     */
+    public InvalidSamlRequestException(@Nullable Throwable cause) {
+        super(cause);
+    }
```
trustin: Could also override the `protected` constructor in `SamlException`?

saml/src/main/java/com/linecorp/armeria/server/saml/SamlAssertionConsumerFunction.java
```diff
@@ -120,27 +120,27 @@ private SamlIdentityProviderConfig resolveIdpConfig(Issuer issuer) {
                 return config;
             }
         }
-        throw new SamlException("failed to find identity provider from configuration " +
-                                issuer.getValue());
+        throw new InvalidSamlRequestException("failed to find identity provider from configuration " +
```
trustin: `... configuration: "` (colon)

saml/src/main/java/com/linecorp/armeria/server/saml/SamlException.java
```diff
@@ -20,7 +20,7 @@
 import com.linecorp.armeria.common.Flags;
 
 /**
- * A checked {@link Exception} which is raised from SAML services.
+ * Indicates that an error occurs while processing a SAML request.
```
trustin: occurred?

PR #1785, opened 2019-05-23
Make gRPC client built without path work correctly

SeungukHan:
Motivation:
 - Current gRPC client that built without path makes a request without
 root path(slash).
   For example, when gRPC client call "hello()" method, the request path
   will be "{scheme}://{authority}hello".
   Root path(slash) need to be located at the next of "{authority}".

Modification:
 - When URI does not have any path value, URI having root path(slash)
    will be passed for DefaultClientBuilderParams

Result:
 - gRPC client built without path can send request correctly.

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactory.java
+++ b/grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactory.java
@@ -29,6 +29,7 @@ import java.util.function.Consumer;
 
 import org.curioswitch.common.protobuf.json.MessageMarshaller;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 
 import com.linecorp.armeria.client.Client;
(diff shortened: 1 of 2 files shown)


grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactory.java
```diff
@@ -180,4 +183,13 @@
             throw new Error(e); // Should never happen.
         }
     }
+
+    private static URI rootPathUri(URI uri) {
+        try {
+            return new URI(uri.getScheme(), uri.getRawAuthority(), "/", uri.getRawQuery(),
+                           uri.getRawFragment());
```
minwoox: `"#" + uri.getRawFragment()`?

grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactory.java
```diff
@@ -180,4 +183,13 @@
             throw new Error(e); // Should never happen.
         }
     }
+
+    private static URI rootPathUri(URI uri) {
+        try {
+            return new URI(uri.getScheme(), uri.getRawAuthority(), "/", uri.getRawQuery(),
```
minwoox: `"?" + uri.getRawQuery()`?

trustin:
Thanks, @SeungukHan!

PR #1787, opened 2019-05-23
Reload PropertiesEndpointGroup if the watched properties file is updated

jrhee17:
Purpose:
- Give option to have PropertiesEndpointGroup reload automatically if the watched properties file is updated

Modification:
- Add a constructor in `PropertiesEndpointGroup` which receives a `filePath`. Allow `endpoints` dynamic update by inheriting `DynamicEndpointGroup`.
- Add a `FileWatcherRegistry` class which keeps track of each file watch event.
- Add a `RestartableThread` class which runs a new `Runnable` every time it is restarted. 
- Add a `FileWatcherRunnable` class which is responsible for detecting file changes in background and calling the appropriate callback method.

Result:
- When a `PropertiesEndpointGroup` is initialized specifying the properties path, the `endpointGroup` will reload automatically on changes to the properties file.


core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+        startFutureIfPossible();
+    }
+
+    /**
+     * Stops watching a properties file corresponding to the resourceUrl.
+     * @param resourceUrl url to stop watching
+     */
+    void deregister(@Nonnull URL resourceUrl) {
+        final PropertiesFileWatcherContext context = ctxRegistry.remove(resourceUrl.getFile());
+        context.key.cancel();
```
hyangtack: Don't we need to `null` check to avoid NPE? (It seems possible to be called more than once )

anuraaga:
@imasahiro I remember you were working on something related to file watching - is there any context we've forgotten about since then? :)

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+                WatchKey key;
+                while ((key = watchService.take()) != null) {
+                    for (WatchEvent<?> event : key.pollEvents()) {
+                        @SuppressWarnings("unchecked")
+                        final Path watchedPath = ((Path) key.watchable())
+                                .resolve(((WatchEvent<Path>) event).context());
+                        final String watchedPathFile = watchedPath.toFile().getAbsolutePath();
+
+                        if (event.kind().equals(ENTRY_MODIFY) || event.kind().equals(ENTRY_CREATE)) {
+                            if (ctxRegistry.containsKey(watchedPathFile)) {
```
anuraaga: There are certain environments where this simple check doesn't actually work from what I've seen. The file might be a symlink, or the directory, not the file might change. I've had better luck iterating, and comparing the real path using `startsWith`

https://github.com/curioswitch/curiostack/blob/master/common/server/framework/src/main/java/org/curioswitch/common/server/framework/files/FileWatcher.java#L140

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+final class PropertiesFileWatcherRegistry {
+    private static final Logger logger = LoggerFactory.getLogger(PropertiesFileWatcherRegistry.class);
+
+    @Nullable
+    private CompletableFuture<Void> future;
+
+    private final Map<String, PropertiesFileWatcherContext> ctxRegistry = new ConcurrentHashMap<>();
+    private final WatchService watchService;
+    private final ExecutorService eventLoop =
```
anuraaga: We use `EventLoop` throughout the codebase with a very specific meaning (Netty event loop), it would be better to call this `executor` to prevent confusion.

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,223 @@
...
+
+    private class PropertiesFileWatcherRunnable implements Runnable {
+        @Override
+        public void run() {
+            try {
+                WatchKey key;
+                while ((key = watchService.take()) != null) {
+                    for (WatchEvent<?> event : key.pollEvents()) {
+                        @SuppressWarnings("unchecked")
+                        final Path watchedPath = getRealPath(
```
anuraaga: Think you should catch and log the exception here and `continue` - it's possible `getRealPath` will throw if the file was modified then immediately deleted. Fine to skip such an event.

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroup.java
```diff
@@ -85,13 +93,42 @@ public static PropertiesEndpointGroup of(ClassLoader classLoader, String resourc
...
+                resourceUrl,
                 requireNonNull(endpointKeyPrefix, "endpointKeyPrefix"),
                 defaultPort));
     }
 
+    /**
+     * Creates a new {@link EndpointGroup} instance that loads the host names (or IP address) and the port
+     * numbers of the {@link Endpoint} from the {@code resourceName} resource file. The resource file must
+     * contain at least one property whose name starts with {@code endpointKeyPrefix}:. Reloading property files
+     * for changes is {@code false} by default.
```
trustin: `Reloading property files ... by default.` could be removed? If you want to emphasize that the other factory methods do not reload, you could write something there, not here.

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroup.java
```diff
@@ -36,9 +40,11 @@
  * A {@link Properties} backed {@link EndpointGroup}. The list of {@link Endpoint}s are loaded from the
  * {@link Properties}.
  */
-public final class PropertiesEndpointGroup implements EndpointGroup {
+public final class PropertiesEndpointGroup extends DynamicEndpointGroup {
+    private static final PropertiesFileWatcherRegistry registry = new PropertiesFileWatcherRegistry();
 
-    // TODO(ide) Reload the endpoint list if the file is updated.
+    @Nullable
+    private Runnable closeCallback;
```
trustin: - This field has to be moved after the static methods.
- This field could be `final` if it is set with `null` when unused.

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroup.java
```diff
@@ -181,14 +247,11 @@ private static void validateDefaultPort(int defaultPort) {
...
-
-    private PropertiesEndpointGroup(List<Endpoint> endpoints) {
-        this.endpoints = endpoints;
-    }
-
     @Override
-    public List<Endpoint> endpoints() {
-        return endpoints;
+    public void close() {
+        super.close();
```
trustin: Maybe we should wrap with `try-finally` and call `closeCallback` in the `finally` block?

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,215 @@
...
+ */
+final class PropertiesFileWatcherRegistry implements AutoCloseable {
+    private static final Logger logger = LoggerFactory.getLogger(PropertiesFileWatcherRegistry.class);
+
+    @Nullable
+    private CompletableFuture<Void> future;
+    private final Map<Path, PropertiesFileWatcherContext> ctxRegistry = new ConcurrentHashMap<>();
+    private final WatchService watchService;
+    private final ExecutorService executor =
+            Executors.newSingleThreadExecutor(new ThreadFactoryBuilder().setDaemon(true).build());
```
trustin: - Could you set the thread name?
- Could we make the executor is started lazily and shut down when there are no watches anymore, so that this class is more friendly to multi-classloader environment? For example, someone might be using Armeria as a part of a Java EE webapp and want Armeria to clean up everything correctly when the web application is reloaded.

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Registers a {@code path} of the properties file to be watched.
+     * @param filePath path to be watched
+     * @param reloader callback to be called on a file change event
+     */
+    synchronized void register(Path filePath, Runnable reloader) {
+        final Path realFilePath = getRealPath(filePath);
+        checkArgument(!ctxRegistry.containsKey(realFilePath),
+                      "file is already watched: %s", realFilePath);
```
trustin: What happens if a user creates a two `PropertiesEndpointGroup` instances with the same paths? Although not very efficient, we would still allow such a case.

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,215 @@
...
+                        if (event.kind().equals(ENTRY_MODIFY) || event.kind().equals(ENTRY_CREATE)) {
+                            if (ctxRegistry.keySet().contains(realPath)) {
+                                try {
+                                    ctxRegistry.get(realPath).reloader.run();
+                                } catch (Exception e) {
+                                    logger.warn("unexpected error from listener: {} ", realPath, e);
+                                }
+                            }
+                        } else if (event.kind().equals(OVERFLOW)) {
+                            logger.info("failed to reload file: {}", realPath);
```
trustin: I'm not sure which is correct, but maybe we can try reloading in this case as well?

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,215 @@
...
+    private final WatchService watchService;
+    private final ExecutorService executor =
+            Executors.newSingleThreadExecutor(new ThreadFactoryBuilder().setDaemon(true).build());
+
+    /**
+     * Create a registry using the default file system.
+     */
+    PropertiesFileWatcherRegistry() {
+        try {
+            watchService = FileSystems.getDefault().newWatchService();
```
trustin: If I understood the NIO.2 API correctly, you have to use the `FileSystem` retrieved from `Path`:

```java
final Path path = Paths.get(uri);
watchService = path.getFileSystem().newWatchService();
```

This means, we may have more than one `WatchService` implementations depending on how a user configured the JRE, i.e. this class needs some refactoring to have different `PropertiesFileWatchRegistry`s for different `FileSystem`s.

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -107,4 +119,119 @@ public void illegalDefaultPort() {
...
+        PrintWriter printWriter = new PrintWriter(file);
+        Properties props = new Properties();
+        props.setProperty("serverA.hosts.0", "127.0.0.1:8080");
+        props.store(printWriter, "");
+        printWriter.close();
+
+        final PropertiesEndpointGroup endpointGroupA = PropertiesEndpointGroup.of(
+                classLoader, file.getName(), "serverA.hosts", 80, true);
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> endpointGroupA.endpoints().size() == 1);
```
trustin: - If there's no good reason to override the default timeout, please remove `atMost()` calls.
- You could use `untilAsserted()` instead of `until()`:

  ```java
  await().untilAsserted(() -> assertThat(endpointGroupA.endpoints()).hasSize(1));
  ```

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -107,4 +119,119 @@ public void illegalDefaultPort() {
...
+
+        // Update resource
+        printWriter = new PrintWriter(file);
+        props = new Properties();
+        props.setProperty("serverA.hosts.0", "127.0.0.1:8080");
+        props.setProperty("serverA.hosts.1", "127.0.0.1:8081");
+        props.store(printWriter, "");
+        printWriter.close();
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> endpointGroupA.endpoints().size() == 2);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -107,4 +119,119 @@ public void illegalDefaultPort() {
...
+        PrintWriter printWriter = new PrintWriter(file);
+        Properties props = new Properties();
+        props.setProperty("serverA.hosts.0", "127.0.0.1:8080");
+        props.store(printWriter, "");
+        printWriter.close();
+
+        final PropertiesEndpointGroup endpointGroupA = PropertiesEndpointGroup.of(
+                classLoader, file.getName(), "serverA.hosts", 80, true);
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> endpointGroupA.endpoints().size() == 1);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -107,4 +119,119 @@ public void illegalDefaultPort() {
...
+
+        endpointGroupA.close();
+
+        final PropertiesEndpointGroup endpointGroupB = PropertiesEndpointGroup.of(
+                classLoader, file.getName(), "serverB.hosts", 80, true);
+        await().atMost(20, TimeUnit.SECONDS).until(() -> endpointGroupB.endpoints().isEmpty());
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -107,4 +119,119 @@ public void illegalDefaultPort() {
...
+
+        // Update resource
+        printWriter = new PrintWriter(file);
+        props = new Properties();
+        props.setProperty("serverB.hosts.0", "127.0.0.1:8080");
+        props.setProperty("serverB.hosts.1", "127.0.0.1:8081");
+        props.store(printWriter, "");
+        printWriter.close();
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> endpointGroupB.endpoints().size() == 2);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -107,4 +119,119 @@ public void illegalDefaultPort() {
...
+        props.setProperty("serverA.hosts.1", "127.0.0.1:8081");
+        props.store(printWriter, "");
+        printWriter.close();
+
+        final PropertiesEndpointGroup propertiesEndpointGroup = PropertiesEndpointGroup.of(
+                classLoader, file.getName(), "serverA.hosts", 80, true);
+        final StaticEndpointGroup staticEndpointGroup = new StaticEndpointGroup(Endpoint.of("127.0.0.1", 8081));
+        final EndpointGroup endpointGroup = propertiesEndpointGroup.orElse(staticEndpointGroup);
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> endpointGroup.endpoints().size() == 2);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -107,4 +119,119 @@ public void illegalDefaultPort() {
...
+
+        printWriter = new PrintWriter(file);
+        props = new Properties();
+        props.store(printWriter, "");
+        printWriter.close();
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> endpointGroup.endpoints().size() == 1);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -107,4 +119,119 @@ public void illegalDefaultPort() {
...
+
+        printWriter = new PrintWriter(file);
+        props = new Properties();
+        props.setProperty("serverA.hosts.0", "127.0.0.1:8080");
+        props.setProperty("serverA.hosts.1", "127.0.0.1:8081");
+        props.setProperty("serverA.hosts.2", "127.0.0.1:8082");
+        props.store(printWriter, "");
+        printWriter.close();
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> endpointGroup.endpoints().size() == 3);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistryTest.java
```diff
@@ -0,0 +1,111 @@
...
+                // do nothing
+            }
+            throw new RuntimeException();
+        });
+
+        PrintWriter printWriter = new PrintWriter(file);
+        printWriter.print(1);
+        printWriter.close();
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> val.get() == 1);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistryTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+        assertThat(propertiesFileWatcherRegistry.isRunning()).isTrue();
+
+        printWriter = new PrintWriter(file);
+        printWriter.print(2);
+        printWriter.close();
+
+        await().atMost(20, TimeUnit.SECONDS).until(() -> val.get() == 2);
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroup.java
```diff
@@ -85,13 +93,42 @@ public static PropertiesEndpointGroup of(ClassLoader classLoader, String resourc
...
+     * @param resourceName the name of the resource where the list of {@link Endpoint}s is loaded from
+     * @param endpointKeyPrefix the property name prefix
+     * @param defaultPort the default port number to use
+     * @param reloadable whether to watch and reload resource.
+     *
+     * @throws IllegalArgumentException if failed to load any hosts from the specified resource file
+     */
+    public static PropertiesEndpointGroup of(ClassLoader classLoader, String resourceName,
+                                             String endpointKeyPrefix, int defaultPort,
+                                             boolean reloadable) {
```
trustin: Maybe we can just require `Path` instead of `ClassLoader` and `resourceName` to differentiate this from other factory methods that do not reload? What do you think?

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroup.java
```diff
@@ -141,20 +138,78 @@ public static PropertiesEndpointGroup of(Properties properties, String endpointK
...
         } catch (IOException e) {
-            throw new IllegalArgumentException("failed to load: " + resourceName, e);
+            throw new IllegalArgumentException("failed to load: " + resourceUrl.getFile(), e);
+        }
+    }
+
+    private static List<Endpoint> loadEndpoints(Path path, String endpointKeyPrefix, int defaultPort) {
+        final URL resourceUrl;
+        try {
+            resourceUrl = path.toUri().toURL();
```
trustin: We could stop converting a `Path` to a `URL` but just use `Files.newInputStream(Path, ...)`.

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
jrhee17: I'm still thinking about use-cases for this & how I can do this efficiently 🤔. Just leaving a comment to let you know I'm not disregarding this comment:)

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
jrhee17: I was kind of thinking as long as a submitted `thread` (or `future`) isn't running, there should be minimal overhead from just instantiating `ExecutorService`. (The `future` is currently stopped when no paths are registered). Let me know if I am thinking about this wrong 🤔

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -42,7 +42,7 @@
 
     @BeforeClass
     public static void before() {
-        Awaitility.setDefaultTimeout(20, TimeUnit.SECONDS);
+        Awaitility.setDefaultTimeout(1, TimeUnit.MINUTES);
```
trustin: You have to revert this back in `@AfterClass`.

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistryTest.java
```diff
@@ -37,7 +37,7 @@
 
     @BeforeClass
     public static void before() {
-        Awaitility.setDefaultTimeout(20, TimeUnit.SECONDS);
+        Awaitility.setDefaultTimeout(1, TimeUnit.MINUTES);
```
trustin: You have to revert this back in `@AfterClass`.

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
```diff
@@ -42,7 +43,12 @@
 
     @BeforeClass
     public static void before() {
-        Awaitility.setDefaultTimeout(1, TimeUnit.MINUTES);
+        Awaitility.setDefaultTimeout(1, TimeUnit.SECONDS);
```
trustin: Is this intended? It seems to me that you don't need to adjust the timeout at all if you are decreasing the timeout.

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,233 @@
...
+
+        @Nullable
+        private CompletableFuture<Void> future;
+        private final ExecutorService executor;
+        private final Supplier<Runnable> runnableSupplier;
+
+        private RestartableFuture(String namePrefix, Supplier<Runnable> runnableSupplier) {
+            executor = Executors.newSingleThreadExecutor(
+                    new ThreadFactoryBuilder().setDaemon(true)
+                                              .setNameFormat(namePrefix).build());
```
trustin: When is this executor's thread terminated? I don't see any `shutdown` or `awaitTermination` calls. Even if the `Runnable` is finished running, the executor thread will stay alive. This causes resource leak in a container environment where an app can be reloaded.

core/src/test/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroupTest.java
jrhee17: whoops! fixed

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
jrhee17: Ah you're right -- there will be idle threads.
When I tried starting and terminating the `ExecutorService`, it seemed to take too long (5s for each start-stop iteration) so I changed to using `Thread`s which seemed must faster.
Let me know if you have any further concerns

core/src/main/java/com/linecorp/armeria/client/endpoint/RestartableThread.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Stops the current running thread if available.
+     */
+    void stop() {
+        checkState(thread != null, "tried to stop null thread");
+        if (isRunning()) {
+            thread.interrupt();
+            try {
+                thread.join(AWAIT_TERMINATE_MILLIS);
```
trustin: Is there a case where this thread is not terminated quickly? How about making sure it's terminated?

```java
while (thread.isAlive()) {
    thread.interrupt();
    try {
        thread.join(1000);
    } catch (InterruptedException e) {
        interrupted = true;
    }
}
thread = null;

if (interrupted) {
    Thread.currentThread().interrupt();
}
```

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+import java.nio.file.WatchService;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import com.linecorp.armeria.client.endpoint.FileWatcherRunnable.FileWatcherContext;
+
+/**
+ * Wraps a {@code WatchService} and allows paths to be registered.
```
hyangtack: nit: `@code` -> `@link`?

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+        if (!existsDirWatcher) {
+            removedCtx.cancel();
+        }
+        if (ctxRegistry.isEmpty()) {
+            restartableThread.stop();
+        }
+    }
+
+    /**
+     * Check if future for {@code WatchService} is running.
```
hyangtack: nit: `Returns whether the watching thread is running.`?

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+     * @return true if future is running
+     */
+    @VisibleForTesting
+    boolean isRunning() {
+        return restartableThread.isRunning();
+    }
+
+    /**
+     * Close watch service, future, and clear registry.
+     * @throws Exception may be thrown if {@code WatchService} is already closed
```
hyangtack: From the javadoc for WatchService.close():
```
     * If this watch service is already closed then invoking this method
     * has no effect.
```
So this statement seems necessary to be checked again?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
```diff
@@ -0,0 +1,138 @@
...
+            while ((key = watchService.take()) != null) {
+                for (WatchEvent<?> event : key.pollEvents()) {
+                    @SuppressWarnings("unchecked")
+                    final Path watchedPath = ((Path) key.watchable())
+                            .resolve(((WatchEvent<Path>) event).context());
+                    final Path realFilePath;
+                    try {
+                        realFilePath = watchedPath.toRealPath();
+                    } catch (IOException e) {
+                        logger.warn("skipping -- unable to get real path for {}", watchedPath);
```
hyangtack: nit: How about passing `e`? `, watchedPath);` -> `, watchedPath, e);`

core/src/main/java/com/linecorp/armeria/client/endpoint/RestartableThread.java
jrhee17: good idea! done:)

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
jrhee17: didn't know about `@link`:( done!

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesFileWatcherRegistry.java
jrhee17: good point:)

jrhee17:
Sorry for making big changes frequently after each review:(
**Updates**
- modify `WatchService` logic to not cast `Object` to `Path` incorrectly for `OVERFLOW` type. (this means we won't be able to reload on `OVERFLOW` since we don't know the target directory)
- allow multiple `FileSystem`s to be registered
- move new files to `properties` package
- introduce a "registration key"

jrhee17:
In retrospect, it may be better to leave `PropertiesEndpointGroup` in the `endpoints` package to avoid breaking changes (and since there aren't many relevant files). Let me move those files back to the original package tonight :P

trustin:
> Sorry for making big changes frequently after each review:(

No worries. It's always happy to see a pull request evolves in a right direction. :+1:

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,228 @@
...
+            }
+        });
+        fileSystemWatchServiceMap.clear();
+    }
+
+    /**
+     * Key used to register/unregister for file change events.
+     */
+    static final class FileWatchRegisterKey {
+        private final Path filePath;
```
trustin: Question: Did you wrap `Path` with `FileWatchRegisterKey` for identity comparison?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
```diff
@@ -0,0 +1,144 @@
...
+                        try {
+                            realFilePath = watchedPath.toRealPath();
+                        } catch (IOException e) {
+                            logger.warn("skipping -- unable to get real path for {}", watchedPath, e);
+                            continue;
+                        }
+
+                        runCallback(realFilePath);
+                    } else {
+                        if (event.kind().equals(OVERFLOW)) {
```
trustin: Could be simplified to `} else if (...) {...} else {...}`

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
```diff
@@ -0,0 +1,144 @@
...
+                        } else {
+                            logger.debug("ignoring unexpected event type: {}", event.kind().name());
+                        }
+                    }
+                }
+
+                final boolean reset = key.reset();
+                if (!reset) {
+                    logger.warn("aborting file watch due to unexpected error");
+                    break;
```
trustin: Question: Is this an error that prevents us from continue watching? What happens if we do not break the loop here?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+
+    private void runCallback(Path filePath) {
+        ctxSupplier.get().stream().filter(
+                ctx -> filePath.startsWith(ctx.getDirPath())).forEach(ctx -> {
+            try {
+                ctx.runCallback();
+            } catch (Exception e) {
+                logger.warn("unexpected error from listener: {} ", filePath, e);
```
trustin: Could you capitalize the first character of all log messages in this PR? e.g. `unexpected error` -> `Unexpected error`

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
```diff
@@ -0,0 +1,144 @@
...
+            this.watchKey = watchKey;
+            this.callback = callback;
+            this.dirPath = dirPath;
+        }
+
+        private void runCallback() {
+            callback.run();
+        }
+
+        Path getDirPath() {
```
trustin: nit: `dirPath()`?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
```diff
@@ -0,0 +1,144 @@
...
+            return dirPath;
+        }
+
+        void cancel() {
+            watchKey.cancel();
+        }
+    }
```
trustin: How about overriding `toString()` for easier debugging later?

```java
@Override
public String toString() {
    return MoreObjects.toStringHelper(this).add("dirPath", dirPath)....toString();
```

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,228 @@
...
+            this.filePath = filePath;
+        }
+
+        /**
+         * Returns the file path associated with the current key. The path is used to unregister
+         * from the {@link WatchService}.
+         *
+         * @return the path associated with the current key
+         */
+        Path getFilePath() {
```
trustin: nit: `filePath()`?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,228 @@
...
+            return filePath;
+        }
+    }
```
trustin: How about overriding `toString()` for easier debugging later?

```java
@Override
public String toString() {
    return MoreObjects.toStringHelper(this).add("filePath", filePath).toString();
```

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
jrhee17: done! I didn't apply capitalization to exception messages for now.. but seems like the convention is mixed within the project. would you prefer exception messages be capitalized as well?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
jrhee17: Yes. This was done to allow duplicate path registration. Let me know if I am missing anything (different design pattern, explicitly override to show identity comparison, etc.)

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
jrhee17: Good point. I wasn't able to create a test case but there was a race condition I was able to reproduce by adding a breakpoint and debugging. 
1. `watchService.take()` returns a `WatchKey`
2. `key.cancel()` is called 
3. `key.reset()` is called and returns false, breaking the loop
This would prevent losing directory watches for other "innocent" paths.

The javadoc imply that `invalid` keys don't take part in further watch events, so I agree it would be better to remove the `break` statement here

jrhee17:
**Updates**
- removed `getRealPath` in `FileWatcherRunnable` since we already use the real path when registering  `Path` to `WatchService`
javadoc for `WatchKey.watchable()`:
> Returns the object for which this watch key was created
- I also realized since all events have access to `WatchKey.watchable()`, we can reload files on `OVERFLOW` events
- Decided to pass `context` over a `supplier` for better code readability

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
trustin: For exception message, please capitalize only when it's a full sentence. No need to capitalize otherwise.

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -74,10 +76,10 @@
         void register(FileWatchRegisterKey watchRegisterKey, Runnable callback) {
             final Path dirPath;
             try {
-                dirPath = watchRegisterKey.getFilePath().toRealPath().getParent();
+                dirPath = watchRegisterKey.filePath().toRealPath().getParent();
             } catch (IOException e) {
                 throw new IllegalArgumentException("failed to locate file " +
```
trustin: For example, this doesn't have to be capitalized because it's not a complete sentence.

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
```diff
@@ -19,92 +19,80 @@
...
-                            logger.warn("ignoring deleted file: {}", watchedPath);
-                            continue;
-                        }
-
-                        final Path realFilePath;
-                        try {
-                            realFilePath = watchedPath.toRealPath();
-                        } catch (IOException e) {
-                            logger.warn("skipping -- unable to get real path for {}", watchedPath, e);
+                            logger.warn("Ignoring deleted file: {}", filePath);
```
trustin: a deleted file

core/src/test/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnableTest.java
```diff
@@ -19,18 +19,27 @@
...
 
     @Test
     public void testPropertyFileWatcherRunnableExitsOnInterrupt() throws InterruptedException {
         final WatchService watchService = mock(WatchService.class);
-        final FileWatcherRunnable fileWatcherRunnable = new FileWatcherRunnable(watchService, ArrayList::new);
+        final FileWatcherRunnable fileWatcherRunnable = new FileWatcherRunnable(watchService, mock(
+                FileWatchServiceContext.class));
+        when(watchService.take()).then(invocation -> {
+            while (!Thread.currentThread().isInterrupted()) {
+                // do nothing
```
trustin: nit: How about adding `Thread.yield()` at least?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,239 @@
...
+
+            final WatchKey watchKey;
+            try {
+                watchKey = dirPath.register(watchService,
+                                            ENTRY_CREATE,
+                                            ENTRY_MODIFY,
+                                            ENTRY_DELETE,
+                                            OVERFLOW);
+            } catch (IOException e) {
+                throw new IllegalArgumentException("failed to watch file " + watchRegisterKey.filePath(), e);
```
trustin: a file

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,239 @@
...
+         * @param callback function invoked on file change
+         *
+         * @throws IllegalArgumentException if failed to locate file or failed to start watching
+         */
+        void register(FileWatchRegisterKey watchRegisterKey, Runnable callback) {
+            final Path dirPath;
+            try {
+                dirPath = watchRegisterKey.filePath().toRealPath().getParent();
+            } catch (IOException e) {
+                throw new IllegalArgumentException("failed to locate file " +
```
trustin: a file

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
trustin: I see. How about adding some comment about identity comparison in the class Javadoc?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRunnable.java
```diff
@@ -0,0 +1,141 @@
...
+                    if (event.kind().type() == Path.class) {
+                        @SuppressWarnings("unchecked")
+                        final Path filePath = dirPath.resolve(((WatchEvent<Path>) event).context());
+                        if (event.kind().equals(ENTRY_DELETE)) {
+                            logger.warn("Ignoring deleted file: {}", filePath);
+                            continue;
+                        }
+                        runCallback(filePath);
+                    } else if (event.kind().equals(OVERFLOW)) {
+                        logger.debug("Watch events may have been lost");
```
trustin: missing preiod, because it's a full sentence.

core/src/test/java/com/linecorp/armeria/client/endpoint/RestartableThreadTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Rule
+    public final TestName testName = new TestName();
+
+    @Test
+    public void testRestartableThreadRestartBehavior() {
+        final RestartableThread restartableThread =
+                new RestartableThread(testName.getMethodName(), () -> () -> {
+                    while (!Thread.currentThread().isInterrupted()) {
+                        // do nothing
```
trustin: Ditto - `Thread.yield()`?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,239 @@
...
+         * isn't running yet, this method starts a thread to start watching the {@link FileSystem}.
+         * @param watchRegisterKey key that contains the path to be watched
+         * @param callback function invoked on file change
+         *
+         * @throws IllegalArgumentException if failed to locate file or failed to start watching
+         */
+        void register(FileWatchRegisterKey watchRegisterKey, Runnable callback) {
+            final Path dirPath;
+            try {
+                dirPath = watchRegisterKey.filePath().toRealPath().getParent();
```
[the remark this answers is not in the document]
jrhee17: hmmm.. I think it depends on what the user wants -- the user might want to watch the "real file" disregarding symlinks, or intentionally watch symlink directories that may change. I'm starting to think it might actually be better to delegate this decision to the user and not manipulate the input `path`. (especially since I'm not 100% sure what the use-cases for this class will be)
wdyt?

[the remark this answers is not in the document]
anuraaga: Yeah, it'd be simple for the user to call `toRealPath` if they wanted that behavior, so not modifying the input SGTM

jrhee17:
**Updates**
- applied comments regarding logging style
- modified to not use real paths
- renamed `FileWatchServiceContext` to `FileSystemWatchContext` for better readability

trustin:
@imasahiro Any comments?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,234 @@
...
+         * When changes are detected, the {@code callback} is invoked. If the {@link WatchService}
+         * isn't running yet, this method starts a thread to start watching the {@link FileSystem}.
+         * @param watchRegisterKey key that contains the path to be watched
+         * @param callback function invoked on file change
+         *
+         * @throws IllegalArgumentException if failed to locate file or failed to start watching
+         */
+        void register(FileWatchRegisterKey watchRegisterKey, Runnable callback) {
+            final Path dirPath = watchRegisterKey.filePath().getParent();
+            checkArgument(dirPath != null, "no parent directory for input path: %s",
```
imasahiro: nit. We usualy use `requireNonNull` for null check?

trustin: I think it has to be `IAE` in this case because the user parameter (`filePath`) is not null but its property (`dirPath`) is.

jrhee17: Originally, I thought it would be better to throw an `IllegalArgumentException` over an NPE to let the user know the input is wrong. If this isn't necessary, I'm fine with sticking to convention:) Any thoughts?

jrhee17: (It does seem like other input parameters throw NPE though...)
https://github.com/line/armeria/blob/fd6f91f373c03862cbb2a622fd70d8b85c11305c/core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroup.java#L126
`requireNonNull` might be better as it is consistent with other fields validations

core/src/main/java/com/linecorp/armeria/client/endpoint/PropertiesEndpointGroup.java
```diff
@@ -181,14 +248,35 @@ private static void validateDefaultPort(int defaultPort) {
...
+                requireNonNull(endpointKeyPrefix, "endpointKeyPrefix"),
+                defaultPort));
+        watchRegisterKey = registry.register(path, () ->
+                setEndpoints(loadEndpoints(path, endpointKeyPrefix, defaultPort)));
+    }
+
+    /**
+     * Unregisters the current {@link PropertiesEndpointGroup} from the {@link FileWatcherRegistry}.
+     * Invoking this method will attempt to release any unused resources used to continuously
+     * watch properties files.
```
trustin: Just realized that this refers to `FileWatcherRegistry` which is not a public class, exposing the implementation detail of this class. How about just removing the Javadoc entirely, given that it behaves just like a usual `Closeable`?

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
jrhee17: oh wait... let me remove the fix for this. I missed @trustin 's comment:(

core/src/main/java/com/linecorp/armeria/client/endpoint/RestartableThread.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param name name of thread
+     * @param runnableSupplier supplies the runnable for the thread to use on each restart
+     */
+    RestartableThread(String name, Supplier<Runnable> runnableSupplier) {
+        this.name = name;
+        this.runnableSupplier = runnableSupplier;
+    }
+
+    /**
+     * Starts thread with runnable if not running yet.
```
minwoox: nit: a thread?

core/src/main/java/com/linecorp/armeria/client/endpoint/RestartableThread.java
```diff
@@ -0,0 +1,87 @@
...
+
+            thread = null;
+            if (interrupted) {
+                Thread.currentThread().interrupt();
+            }
+        }
+    }
+
+    /**
+     * Whether thread is currently running.
```
minwoox: a thread?

core/src/test/java/com/linecorp/armeria/client/endpoint/RestartableThreadTest.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestName;
+
+public class RestartableThreadTest {
+
+    @Rule
+    public final TestName testName = new TestName();
```
minwoox: nit: we can use `TestInfo` and convert this test using JUnit 5 if you want.

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,239 @@
...
+
+        void close() throws IOException {
+            currWatchEventMap.clear();
+            restartableThread.stop();
+            watchService.close();
+        }
+    }
+
+    private final Map<FileSystem, FileWatchServiceContext> fileSystemWatchServiceMap
+            = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
jrhee17: yes it can since all modifications to this map is in `syncrhonized` calls. 👍

core/src/main/java/com/linecorp/armeria/client/endpoint/FileWatcherRegistry.java
```diff
@@ -0,0 +1,239 @@
...
+                watchKey = dirPath.register(watchService,
+                                            ENTRY_CREATE,
+                                            ENTRY_MODIFY,
+                                            ENTRY_DELETE,
+                                            OVERFLOW);
+            } catch (IOException e) {
+                throw new IllegalArgumentException("failed to watch file " + watchRegisterKey.filePath(), e);
+            }
+            currWatchEventMap.put(watchRegisterKey, new FileWatchEvent(watchKey, callback, dirPath));
+            if (!currWatchEventMap.isEmpty()) {
```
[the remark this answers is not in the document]
jrhee17: you're right, we probably don't need this check.
(Let me know if you also think the method name should be changed to something more appropriate like `startIfPossible`... I think as-is is fine though since a single thread can't start multiple times)

core/src/test/java/com/linecorp/armeria/client/endpoint/RestartableThreadTest.java
jrhee17: 👍 although I would have to rebase and I'm a little concerned bc the comments in this PR won't be trackable anymore + it's difficult for reviewers to see change.   If you think rebasing is fine, I can implement in this PR. let me know what you think:)

minwoox: Yeah, it's fine as it is. :-) We were planning to convert all the tests using JUnit 5 eventually, but we can change it later.

trustin:
BOOM! 💥 Thanks again for your hard work and we look forward seeing you again in a new pull request. 😉 👏

PR #1789, opened 2019-05-24
Revamp Route API

minwoox:
Motivation:
`PathMapping` should be used to map only the path from a request as its name implies.
However, it checks HTTP methods, content-type and accepts headers from a request to
find a service as well. Also, the content negotiation was somewhat unclear so we have to make it clear.

Modifications:
- Add `Route` which replaces the role of `PathMapping`
  - `Route` now has `PathMapping` and the list of HTTP methods, consume types and produce types.
  - The mapped service is found when an HTTP request meets all of the following conditions:
    - The path of the request matches the `PathMapping`
    - The HTTP method of the request is one of the methods in `Route`, if not empty
    - The `content-type` of the request is one of the `consumes` in `Route`, if not empty
    - The `accept` header of the request is one of the `produces` in `Route`, if not empty
- Add `RoutingContext`, `RoutingResult` and `Routed`
- Breaking changes
  - `ServiceWithPathMappings` -> `ServiceWithRoutes`
  - `PathMapping` is package-private
  - `PathMappingResult` -> `RoutingResult`.
  - `PathMapped` -> `Routed`
- Content negotiation
  - If the `Route` does not specify `consumes` and `produces`, the `Route` only checks the path no matter the request has content-type and accept headers or not. (The server responded 406 before.)

Result:
- You should now use `Route` to map an HTTP request on a service.

Feedback appreciated:
- I hope to rename `RouteElement`. It's just a container which has `Route`, `RouteResult` and the found service. I wanted to name it as `RouteResult`, but I already used it. If you have any suggestion, please let me know.
  - Renamed to `Routed`, thanks @trustin!


core/src/main/java/com/linecorp/armeria/server/GlobPathMapping.java
```diff
@@ -166,119 +162,119 @@ static PatternAndParamCount globToRegex(String glob) {
             }
 
             switch (asterisks) {
-            case 1:
```
minwoox: The changes below are just aligning.

core/src/main/java/com/linecorp/armeria/server/RouteElement.java
```diff
@@ -27,75 +27,73 @@
  *
  * @param <T> the type of the mapped value
  */
-public final class PathMapped<T> {
+public final class RouteElement<T> {
```
trustin: `Routed`? e.g. `Routed<ServiceConfig>`

core/src/main/java/com/linecorp/armeria/server/RouteResult.java
```diff
@@ -0,0 +1,171 @@
...
+import javax.annotation.Nullable;
+
+import com.google.common.base.MoreObjects;
+
+import com.linecorp.armeria.common.MediaType;
+
+/**
+ * The result returned by {@link Route#apply(RoutingContext)}.
+ */
+public class RouteResult {
```
trustin: - `RoutingResult`?
- `final`?

thrift/src/test/java/com/linecorp/armeria/it/metric/DropwizardMetricsIntegrationTest.java
```diff
@@ -94,6 +94,8 @@ public void normal() throws Exception {
             assertThat(dropwizardRegistry.getMeters()
                                          .get(clientMetricNameWithStatusAndResult("requests", 200, "failure"))
                                          .getCount()).isEqualTo(3L);
+            System.err.println(dropwizardRegistry.getMeters());
+            System.err.println(serverMetricNameWithStatusAndResult("requests", 200, "failure"));
```
trustin: Better using a logger

core/src/main/java/com/linecorp/armeria/server/RouteResult.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * Returns a new builder.
+     */
+    public static RouteResultBuilder builder() {
+        return new RouteResultBuilder();
+    }
+
+    private final PathMappingResult pathMappingResult;
```
trustin: I'm curious if we can merge `PathMappingResult(Builder)` into `RoutingResult(Builder)` somehow, because there's always 1-to-1 cardinality for each other. For example, `PathMapping.apply()` could return a `RoutingResultBuilder`? Not sure if it is a very good idea though. Thoughts?

core/src/main/java/com/linecorp/armeria/server/AbstractPathMapping.java
```diff
@@ -16,36 +16,36 @@
...
+import static com.linecorp.armeria.internal.RouteUtil.UNKNOWN_LOGGER_NAME;
+import static com.linecorp.armeria.internal.RouteUtil.ensureAbsolutePath;
 
 import java.util.Optional;
 
 /**
- * A skeletal {@link PathMapping} implementation. Implement {@link #doApply(PathMappingContext)}.
+ * A skeletal {@link PathMapping} implementation. Implement {@link #doApply(RoutingContext)}.
  */
 public abstract class AbstractPathMapping implements PathMapping {
```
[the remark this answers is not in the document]
minwoox: Oops forgot to remove that. 😅

core/src/main/java/com/linecorp/armeria/server/RouteElement.java
minwoox: Will use that. Thanks!

core/src/main/java/com/linecorp/armeria/server/RouteResult.java
minwoox: Will Change it, thanks!

thrift/src/test/java/com/linecorp/armeria/it/metric/DropwizardMetricsIntegrationTest.java
minwoox: Oops, these are shouldn't be here. 😅

core/src/main/java/com/linecorp/armeria/server/AbstractServiceBindingBuilder.java
```diff
@@ -93,7 +93,7 @@
      * @throws IllegalArgumentException if the specified path pattern is invalid
      */
     public AbstractServiceBindingBuilder path(String pathPattern) {
-        defaultPathMapping = PathMapping.of(requireNonNull(pathPattern, "pathPattern"));
+        defaultRouteBuilder = Route.builder().path(pathPattern);
```
hyangtack: nit: `requireNonNull(pathPattern, "pathPattern")`?

core/src/main/java/com/linecorp/armeria/server/AbstractPathMapping.java
```diff
@@ -16,36 +16,36 @@
...
      * {@inheritDoc} This method performs sanity checks on the specified {@code path} and calls
-     * {@link #doApply(PathMappingContext)}.
+     * {@link #doApply(RoutingContext)}.
      */
     @Override
-    public final PathMappingResult apply(PathMappingContext mappingCtx) {
-        ensureAbsolutePath(mappingCtx.path(), "path");
-        return doApply(mappingCtx);
+    public final PathMappingResult apply(RoutingContext routingCtx) {
+        ensureAbsolutePath(routingCtx.path(), "path");
```
hyangtack: nit: `requireNonNull(routingCtx, "routingCtx")`?

core/src/main/java/com/linecorp/armeria/server/CompositeRouter.java
```diff
@@ -35,27 +36,32 @@
...
-            final PathMapped<I> result = delegate.find(mappingCtx);
+            final RouteElement<I> result = delegate.find(routingCtx);
             if (result.isPresent()) {
                 return resultMapper.apply(result);
             }
         }
-        mappingCtx.delayedThrowable().ifPresent(Exceptions::throwUnsafely);
-        return PathMapped.empty();
+        routingCtx.delayedThrowable().ifPresent(Exceptions::throwUnsafely);
+        if (routingCtx.isCorsPreflight()) {
```
hyangtack: Just curious. Would it be better to send `403 forbidden` for every failed preflight request? I mean, should we check this before throwing delayed throwable? 🤔

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -121,27 +121,8 @@ B setConfig(CorsDecorator corsDecorator) {
...
-     */
-    public B pathMapping(PathMapping pathMapping) {
-        requireNonNull(pathMapping, "pathMapping");
-        if (!pathMapping.hasPathPatternOnly()) {
-            throw new IllegalArgumentException(
-                    "pathMapping: " + pathMapping.getClass().getSimpleName() +
-                    " (expected: the path mapping which has only the path patterns as its condition)");
-        }
-        pathMappings.add(pathMapping);
+    public B route(String pathPattern) {
```
hyangtack: Don't necessary to add `route(Route route)`?

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -133,21 +131,8 @@ public CorsServiceBuilder addPolicy(CorsPolicy policy) {
...
-     * Adds a {@link PathMapping} that this policy is supposed to be applied to.
-     *
-     * @param pathMapping the {@link PathMapping} that this policy is supposed to be applied to
-     * @throws IllegalArgumentException if the {@link PathMapping} has conditions beyond the path pattern,
-     *                                  i.e. the {@link PathMapping} created by
-     *                                  {@link PathMapping#withHttpHeaderInfo(Set, List, List)}
-     */
-    public CorsServiceBuilder pathMapping(PathMapping pathMapping) {
-        firstPolicyBuilder.pathMapping(pathMapping);
+    public CorsServiceBuilder route(String pathPattern) {
```
hyangtack: Ditto - Don't necessary to add `route(Route route)`?

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
minwoox: I didn't add the API intentionally because the user should use just the path. In the previous implementation, we were calling `pathMapping.hasPathPatternOnly()` to check whether the `pathMapping` has conditions other than that. So I thought just these APIs enough for the users because they do not have to specify HTTP methods, consumes and produces.
But I'm not so strong on this and if we still need the API I can add it.

core/src/main/java/com/linecorp/armeria/server/CompositeRouter.java
minwoox: Yeah, let me change it. :-)

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
trustin: +1 for not accepting `Route`

hyangtack: I mis-thought why I made the limitation which requires path pattern only. Yes, you are right. It's still not useful after this changes. I mis-thought this change can overcome the limitation . 😅

core/src/main/java/com/linecorp/armeria/server/AbstractServiceBindingBuilder.java
minwoox: Forgot to put it back. 😅

core/src/main/java/com/linecorp/armeria/server/RoutingResultBuilder.java
```diff
@@ -46,65 +46,65 @@
...
 
     /**
-     * Returns a newly-created {@link RouteResult}.
+     * Returns a newly-created {@link RoutingResult}.
      */
-    public RouteResult build() {
+    public RoutingResult build() {
         if (pathMappingResult != null) {
-            return new RouteResult(pathMappingResult, score, negotiatedResponseMediaType);
+            return new RoutingResult(pathMappingResult, score, negotiatedResponseMediaType);
```
trustin: How about spreading the fields of `PathMappingResult`? i.e.

```java
new RoutingResult(path, query, pathParams, score, negotiatedResponseMediaType);
```

`RoutingResult` and `RoutingResultBuilder` will not need to keep `PathMappingResult` but just the three fields from it for reduced indirection.

core/src/main/java/com/linecorp/armeria/server/RouteResult.java
trustin: This is not solved yet, right?

minwoox: Didn't have time to solve it yet :)

minwoox: Yeah, it seems fine that `PathMapping.apply()` returns a `RoutingResultBuilder`. Let me remove the `PathMappingResult`. :-)

core/src/main/java/com/linecorp/armeria/server/PathMapping.java
```diff
@@ -15,166 +15,26 @@
...
-     * @param mappingCtx a context to find the {@link Service}.
+     * @param routingCtx a context to find the {@link Service}.
      *
-     * @return a non-empty {@link PathMappingResult} if the specified {@code path} matches this mapping.
-     *         {@link PathMappingResult#empty()} if not matches.
+     * @return a settable {@link RoutingResultBuilder} if the specified {@code path} matches this mapping.
+     *         {@link RoutingResult#immutableBuilder()} if not matches.
      */
-    PathMappingResult apply(PathMappingContext mappingCtx);
+    RoutingResultBuilder apply(RoutingContext routingCtx);
```
hyangtack: Just curious. For not-matched case, would `@Nullable` and returning `null` be better choice?

minwoox: Ah, yeah, I don't know why I couldn't think of it. 😅 Will change it. Thanks!

trustin:
Any comments, @anuraaga ?

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,351 @@
...
+     *   <li>{@code regex:^/files/(?<filePath>.*)$} (regular expression)</li>
+     * </ul>
+     *
+     * @throws IllegalArgumentException if the specified path pattern is invalid
+     */
+    public RouteBuilder path(String pathPattern) {
+        return pathMapping(getPathMapping(pathPattern));
+    }
+
+    private static PathMapping getPathMapping(String pathPattern) {
```
anuraaga: Class would probably be more readable if all the public methods are on top.

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,351 @@
...
+        methods(ImmutableList.copyOf(requireNonNull(methods, "methods")));
+        return this;
+    }
+
+    /**
+     * Sets {@link HttpMethod}s that this {@link Route} supports. If not set, the mapped {@link Service}
+     * accepts any {@link HttpMethod}s.
+     */
+    public RouteBuilder methods(Iterable<HttpMethod> methods) {
+        this.methods = ImmutableSet.copyOf(Sets.immutableEnumSet(requireNonNull(methods, "methods")));
```
anuraaga: Don't need `ImmutableSet.copyOf`

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,351 @@
...
+    private PathMapping pathMapping;
+
+    private Set<HttpMethod> methods = ImmutableSet.of();
+
+    private List<MediaType> consumes = ImmutableList.of();
+
+    private List<MediaType> produces = ImmutableList.of();
+
+    /**
+     * Sets the specified {@code pathPattern} of the {@link Route}. e.g.
```
anuraaga: I think these can all be formatted something like `Sets the {@link Route} to match the specified {@code pathPattern}`

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,351 @@
...
+        return new GlobPathMapping(glob);
+    }
+
+    RouteBuilder pathMapping(PathMapping pathMapping) {
+        this.pathMapping = pathMapping;
+        return this;
+    }
+
+    /**
+     * Sets the specified {@code exactPath} of the {@link Route}. The mapped {@link Service} is found when a
```
anuraaga: `Sets the {@link Route} to match the specified exact path`

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,351 @@
...
+     */
+    public RouteBuilder methods(Iterable<HttpMethod> methods) {
+        this.methods = ImmutableSet.copyOf(Sets.immutableEnumSet(requireNonNull(methods, "methods")));
+        return this;
+    }
+
+    /**
+     * Sets {@link MediaType}s that this {@link Route} consumes. If not set, the the mapped {@link Service}
+     * accepts {@link HttpRequest}s that have any {@link HttpHeaderNames#CONTENT_TYPE}.
+     * In order to get this work, {@link #methods(Iterable)} should be set.
```
anuraaga: `must be set` throughout

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,351 @@
...
+        ensureUniqueMediaTypes(produceTypes, "produceTypes");
+        produces = ImmutableList.copyOf(produceTypes);
+        return this;
+    }
+
+    /**
+     * Returns a newly-created {@link Route} based on the properties of this builder.
+     */
+    public Route build() {
+        checkState(pathMapping != null, "Should set a path before calling this.");
```
anuraaga: Must

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -0,0 +1,351 @@
...
+        return Objects.hash(pathMapping, methods, consumes, produces);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (o == null || getClass() != o.getClass()) {
```
anuraaga: Think we always do `if (!(o instanceof RouteBuilder))`

core/src/main/java/com/linecorp/armeria/server/Route.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Set;
+
+import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.MediaType;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
+
+/**
+ * Matches the {@link RoutingContext} and extracts the path parameters from it.
```
anuraaga: I think this class is more semantic than just matching / extracting. Maybe something like `{@link Route} maps from an incoming request to a {@link Service} based on its path, method, content type, and accepted types.`

core/src/main/java/com/linecorp/armeria/server/Route.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Returns the trie path of this {@link Route} if this has a trie matching condition,
+     * {@link Optional#empty()} otherwise. The trie path does not exist when the {@link Route} is created
+     * using a glob or a regex expression.
+     *
+     * @see RouteBuilder#glob(String)
+     * @see RouteBuilder#regex(String)
+     */
+    Optional<String> triePath();
```
anuraaga: A lot of these methods look pretty complicated - is it appropriate for them to be in the user-facing API?

core/src/main/java/com/linecorp/armeria/server/Route.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Returns the regular expression of this {@link Route} if this is created with the regular expression or
+     * glob pattern. Please note that this regex does not include the {@code pathPrefix} if this is
+     * created with {@link RouteBuilder#pathWithPrefix(String, String)}.
+     * You should call {@link #prefix()} to retrieve that.
+     */
+    Optional<String> regex();
+
+    /**
+     * Returns the complexity of this {@link Route}. It would be increased if this path mapping has
```
anuraaga: `A higher complexity indicates more expensive computation for route matching, usually due to additional number of checks.`

core/src/main/java/com/linecorp/armeria/server/Route.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    int complexity();
+
+    /**
+     * Returns the {@link Set} of {@link HttpMethod}s that this {@link Route} supports.
+     */
+    Set<HttpMethod> methods();
+
+    /**
+     * Returns the {@link List} of {@link MediaType}s that this {@link Route} consumes.
```
anuraaga: I guess it might just be for symmetry, but just wondering does this need to be a `List` instead of `Set`? I think requests only have one content type so this also can only have one possible match per request (unlike produces)

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
minwoox: Thanks! didn't notice that they are copied inside.

core/src/main/java/com/linecorp/armeria/server/Route.java
minwoox: They are needed in DocService which is in the different package. Let me see what I can do. 🤔

core/src/main/java/com/linecorp/armeria/server/Route.java
minwoox: The `HttpHeaderPathMapping` which is sort of the ancestor of this class, had the types in a `List`. But yeah, it should be a `Set` because this indicates the content types that the mapped service accepts.
And the produces should be a `Set` as well. Let me make changes. Thanks!

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
minwoox: Copied from another class without much thinking. Let me change all those classes to reduce the same mistake. 😉

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
minwoox: I prefer these sort of private methods are located right under the method where they are used. But at the bottom is fine too. Let me move it down.

anuraaga: Yeah it does make the code implementation easier to read - but I think a lot of people read the class as documentation (in lieu of reading the javadoc) so having all the publics together helps for that :)

minwoox: Ah, yeah didn't think of that. :-) It could be an obstacle especially when the private method is too long.

core/src/main/java/com/linecorp/armeria/server/Route.java
minwoox: The methods are used in various other packages, so can't just hide them. I'm thinking to add a new internal interface or move the `PathMapping` to internal package so we can just expose it. But it takes a little time and will make this  PR bigger so let me do that following PR. 😉

anuraaga: Cool - thanks for thinking about it.

minwoox:
@anuraaga All fixed. PTAL 🙇

trustin:
Thanks a lot, @minwoox and reviewers. :smile:

minwoox:
Thanks, reviewers!

PR #1790, opened 2019-05-24
Marshal headers <-> gRPC Metadata.

anuraaga:
Now incoming headers are copied into `Metadata` and outgoing `Metadata` is copied into headers. Since `Metadata` is supported now, I have stopped using `EMPTY_METADATA` which was questionable to begin with, since it's mutable so a singleton shouldn't be used. Tests have replaced some code that used to use armeria for manipulating headers with ones that use `Metadata` (armeria versions are very well tested in `core` already so don't really need them here I guess).

Will leave this WIP until hearing back on https://github.com/grpc/grpc-java/issues/5788

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/internal/grpc/MetadataUtil.java
+++ b/grpc/src/main/java/com/linecorp/armeria/internal/grpc/MetadataUtil.java
@@ -18,6 +18,7 @@ package com.linecorp.armeria.internal.grpc;
 
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.Map.Entry;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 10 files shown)


anuraaga:
I think it's ok, but we do have to take into concern that when upstream makes a stable API, this one we use here will probably go away, so a user would have a confusing runtime error if they upgraded to the latest `grpc-java` before we get a chance to catch up.

Also enabled all the interop tests :)

trustin:
> I think it's ok, but we do have to take into concern that when upstream makes a stable API, this one we use here will probably go away, so a user would have a confusing runtime error if they upgraded to the latest `grpc-java` before we get a chance to catch up.

Agreed. I think we can keep eyes on the gRPC release announcement and catch up quickly in this case.

> Also enabled all the interop tests :) 

Thanks!

grpc/src/main/java/com/linecorp/armeria/internal/grpc/MetadataUtil.java
```diff
@@ -0,0 +1,152 @@
...
+        for (int i = 0; i < serializedMetadata.length; i += 2) {
+            final AsciiString name = new AsciiString(serializedMetadata[i], false);
+
+            final byte[] valueBytes = serializedMetadata[i + 1];
+
+            final String value;
+            if (isBinary(name)) {
+                value = BASE64_ENCODING_OMIT_PADDING.encode(valueBytes);
+            } else {
+                if (!isGrpcAscii(valueBytes)) {
```
hyangtack: nit: How about changing here as follows for better readability?
```
if (isBinary()) {
  ...
} else if (isGrpcAscii()) {
  ... 
} else {
  ...
  continue;
}
```

trustin:
Thanks, @anuraaga !

PR #1795, opened 2019-05-27
Add AggregatedHttpRequest/Response

### LINKED ISSUE #1702: Consider separate messages for AggregatedHttpMessage request / response
When using an `AggregatedHttpMessage` that is a response, `status` is always set. It's a bit tedious that `status` is `@Nullable` because in the request case, it wouldn't be populated, but in the response case it always is, meaning having to always do a no-op null check to satisfy nullness checks.

minwoox:
Motivation:
We recently introduced `RequestHeaders` and `ResponseHeaders`.
To use APIs that those headers provide from the headers in `AggregatedHttpMessage`,
we have to call `RequestHeaders.of(AggregatedHttpMessage.headers())` which is very cumbersome.
Also, `AggregatedHttpMessage` has getters related a request(`path()`, `method()`, etc.) and a response(`status()`).
If we split `AggregatedHttpMessage`, the subclasses can have setters only what they need so we can reduce the validity check and `Nullable` annotations.

Modifications:
- Add `AggregatedHttpRequest` and `AggregatedHttpResponse` which inherit `AggregatedHttpMessage`.
- Hide `AggregatedHttpMessage` from public.
  - I didn't find any case that a user has to use `AggregatedHttpMessage`, please let me know if there's a case.
- Breaking Changes
  - `HttpRequest.aggregate()` now returns `CompletableFuture<AggregatedHttpRequest>`.
  - `HttpResponse.aggregate()` now returns `CompletableFuture<AggregatedHttpResponse>`.
  - `convertRequest()` in `RequestConverterFunction` now takes `AggregatedHttpRequest`.
- Deprecated
  - `trailingHeaders` -> `trailers` in `AggregatedHttpRequest/Response` and `HttpResult`

Result:
- You can use `AggregatedHttpRequest` and `AggregatedHttpResponse`
- Close #1702


core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
```diff
@@ -246,8 +242,8 @@ static HttpRequest of(RequestHeaders headers, HttpData content, HttpHeaders trai
      * {@code objs} must not contain {@link RequestHeaders}.
      */
     static HttpRequest of(RequestHeaders headers, HttpObject... objs) {
-        if (Arrays.stream(objs).anyMatch(obj -> obj instanceof HttpHeaders)) {
-            throw new IllegalArgumentException("objs contains Headers, which is not allowed.");
+        if (Arrays.stream(objs).anyMatch(obj -> obj instanceof RequestHeaders)) {
```
trustin: `!(obj instanceof HttpData)`?

core/src/main/java/com/linecorp/armeria/common/ResponseHeaders.java
```diff
@@ -179,6 +179,10 @@ static ResponseHeaders of(HttpHeaders headers) {
         }
 
         requireNonNull(headers, "headers");
+        // From the section 8.1.2.4 of RFC 7540:
+        //// For HTTP/2 responses, a single :status pseudo-header field is defined that carries the HTTP status
+        //// code field (see [RFC7231], Section 6). This pseudo-header field MUST be included in all responses;
+        //// otherwise, the response is malformed (Section 8.1.2.6).
```
trustin: Thanks :smile:

core/src/test/java/com/linecorp/armeria/server/ServerTest.java
```diff
@@ -432,7 +432,7 @@ private static void testSimple(
             assertThat(in.readLine()).isEqualTo(expectedStatusLine);
             // Read till the end of the connection.
             final List<String> headers = new ArrayList<>();
-            for (;;) {
+            for (;;)  {
```
trustin: !

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
minwoox: The last element of `objs` could be `HttpHeaders` which is an HTTP trailers.
Do we need to change the API to prohibit it?

core/src/test/java/com/linecorp/armeria/server/ServerTest.java
minwoox: 😅

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
minwoox: Or check the last element separately?

core/src/main/java/com/linecorp/armeria/common/AbstractAggregatedHttpMessage.java
```diff
@@ -0,0 +1,38 @@
...
+        this.trailingHeaders = trailingHeaders;
+    }
+
+    @Override
+    public HttpData content() {
+        return content;
+    }
+
+    @Override
+    public HttpHeaders trailingHeaders() {
```
anuraaga: This is probably a good time to deprecate `trailingHeaders` in favor of `trailers`, which we have consolidated on in other APIs (e.g., `additionalResponseTrailer`). What do you think? If so, let's rename it in all the parameters throughout the PR too.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpService.java
```diff
@@ -188,7 +189,7 @@ private static void warnIfHttpResponseArgumentExists(Type returnType, Parameteri
             } else if (arg instanceof Class) {
                 final Class<?> clazz = (Class<?>) arg;
                 if (HttpResponse.class.isAssignableFrom(clazz) ||
-                    AggregatedHttpMessage.class.isAssignableFrom(clazz)) {
+                    AggregatedHttpRequest.class.isAssignableFrom(clazz)) {
```
hyangtack: Seems to be `AggregatedHttpResponse`?

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpRequest.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /**
+     * Creates a new HTTP request.
+     *
+     * @param method the HTTP method of the request
+     * @param path the path of the request
+     * @param mediaType the {@link MediaType} of the request content
+     * @param content the content of the request
+     * @param trailingHeaders the trailing HTTP headers
```
anuraaga: Trailers can only be in the response, so we don't need them in the API for `AggregatedHttpRequest`.

Another benefit of splitting the classes :D

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -0,0 +1,247 @@
...
+     *
+     * @param headers the HTTP headers
+     * @param content the content of the HTTP response
+     * @param trailingHeaders the trailing HTTP headers
+     */
+    static AggregatedHttpResponse of(ResponseHeaders headers, HttpData content, HttpHeaders trailingHeaders) {
+        requireNonNull(headers, "headers");
+        requireNonNull(content, "content");
+        requireNonNull(trailingHeaders, "trailingHeaders");
+        return of(Collections.emptyList(), headers, content, trailingHeaders);
```
anuraaga: We usually use `ImmutableList.of()` for empty list so guava can optimize for it sometimes.

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -0,0 +1,247 @@
...
+    /**
+     * Returns the {@link ResponseHeaders}.
+     */
+    @Override
+    ResponseHeaders headers();
+
+    /**
+     * Returns the informational class (1xx) HTTP headers.
+     */
+    List<ResponseHeaders> informationals();
```
anuraaga: I guess the old API may have been like this too, but wondering is it better to separate informational like this instead of aggregating it all into a single `ResponseHeaders`?

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
anuraaga: I think the check will be needed for response, but for request we can prohibit it using @trustin's check since they're not allowed as mentioned above.

minwoox: Ah right, I was totally misunderstanding. :-)

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpService.java
minwoox: Oops, thanks for noticing this. 😅

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpRequest.java
hyangtack: Isn't it possible to have trailers in the request over HTTP/2 connection?

https://httpwg.org/specs/rfc7540.html#HttpSequence

anuraaga: Ah thanks! Then we do need them. I mistakenly thought HTTP/2 had the same semantics as HTTP/1 for trailers orz

core/src/main/java/com/linecorp/armeria/common/AbstractAggregatedHttpMessage.java
minwoox: I prefer `trailers` to `trailingHeaders`. :-) If there's no one who agin this, I will rename it.

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
minwoox: Ah just copied from the old source without much thinking. Thanks!

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
minwoox: informational headers could have a status so we still need two separate fields for that?

anuraaga: Ah cool then that makes sense.

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
trustin: Hmm... Why don't we just accept `HttpData...` at the first place? (and update Javadoc)

core/src/main/java/com/linecorp/armeria/common/AbstractAggregatedHttpMessage.java
trustin: :heavy_plus_sign: :one:

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
minwoox: Yeah I was making the API deprecated and adding a new one. :-)

minwoox: Realize that does not have to deprecate the old one and add new one becuase users would not notice the difference.

minwoox:
All fixed. PTAL!

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpRequest.java
```diff
@@ -158,18 +158,18 @@ static AggregatedHttpRequest of(HttpMethod method, String path, MediaType mediaT
      * @param path the path of the request
      * @param mediaType the {@link MediaType} of the request content
      * @param content the content of the request
-     * @param trailingHeaders the trailing HTTP headers
+     * @param trailers the trailing HTTP headers
      */
     static AggregatedHttpRequest of(HttpMethod method, String path, MediaType mediaType,
-                                    HttpData content, HttpHeaders trailingHeaders) {
+                                    HttpData content, HttpHeaders trailers) {
```
minwoox: Didn't change all the parameters and variables names to `trailers`, but just a few classes related `HttpRequest` and `HttpResponse`.

minwoox: Changed my mind. Rename all parameters and variables. :-)

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -211,22 +210,22 @@ static AggregatedHttpResponse of(ResponseHeaders headers, HttpData content, Http
      * @param informationals the informational class (1xx) HTTP headers
      * @param headers the HTTP headers
      * @param content the content of the HTTP response
-     * @param trailingHeaders the trailing HTTP headers
+     * @param trailers the trailing HTTP headers
```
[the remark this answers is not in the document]
minwoox: Thought that it's okay as it is, but I will make changes for that. :-)

site/src/sphinx/server-annotated-service.rst
```diff
@@ -801,7 +801,7 @@ more response types which can be used in the annotated service.
 - :api:`HttpResult`
 
   - It contains the :api:`HttpHeaders` and the object which can be converted into HTTP response body by
-    response converters. A user can customize the HTTP status and headers including the trailing headers,
+    response converters. A user can customize the HTTP status and headers including the trailers,
     with this type.
```
trustin: not relevant to this PR but could you fix the indentation?

.appveyor.yml
```diff
@@ -7,7 +7,7 @@ build:
   verbosity: detailed
 
 build_script:
-  - gradlew.bat --no-daemon --stacktrace --warning-mode=all checkstyle check build
+  - gradlew.bat clean --no-daemon --stacktrace --warning-mode=all checkstyle check build
```
trustin: `clean` could come after all options.

minwoox: Thanks! This was not the cause, I undo it. :-)

site/src/sphinx/server-annotated-service.rst
minwoox: Moved to the above line. :-)

minwoox:
Thanks for reviewing. 😉

PR #1797, opened 2019-05-28
Rename HttpData.of to HttpData.wrap, add copyOf, and deprecate offset.

### LINKED ISSUE #1771: Remove HttpData.offset / length?
Currently, to access the array of an `HttpData` safely, you need to make sure you call of `array`, `offset`, and `length`. But it's very easy to forget the latter two.

I don't think we ever slice an array when making `HttpData`. Does it make sense to remove the offset / length methods to make the API simpler?

anuraaga:
Currently, we allow creating a `HttpData` as a slice of a `byte[]`. However, in practice we almost never use it, only when constructing an `HttpData` after decoding a `CharSequence` (not `String`) which is very rare usage, so I've changed it to copy out the slice. We previously also did so when retrieving the result of Thrift serialization, but replaced that with using `ByteBufHttpData`. So instead, we can change the slicing APIs to use `ByteBuf` and remove the need for users to worry about `offset` / `length` when accessing the data array.

Without the ability to slice `HttpData`, a user can access the array of an `HttpData` for processing by simply calling `.array()`, where it has been easy to forget to call `.offset()` and `.length()` as is currently required. It's especially problematic because it still usually works, but couldn't be guaranteed to work in all situations.

Fixes #1771


core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -104,7 +83,10 @@ static HttpData of(Charset charset, CharSequence text) {
 
         final CharBuffer cb = CharBuffer.wrap(text);
         final ByteBuffer buf = charset.encode(cb);
-        return of(buf.array(), buf.arrayOffset(), buf.remaining());
+
+        byte[] bytes = new byte[buf.remaining()];
```
hyangtack: nit: `final`

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -235,7 +235,7 @@ default boolean isEmpty() {
      *
      * @param charset the {@link Charset} to use for decoding this data
      *
-     * @return the decoded {@link String}
+     * @return the psdecoded {@link String}
```
hyangtack: Seems typo. 😅

anuraaga: Oops😅

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
```diff
@@ -276,22 +257,6 @@ static AggregatedHttpMessage of(HttpStatus status, MediaType mediaType, byte[] c
...
-    /**
-     * Creates a new HTTP response of the specified {@link HttpStatus}.
-     *
-     * @param mediaType the {@link MediaType} of the response content
-     * @param content the content of the response
-     * @param offset the start offset of {@code content}
-     * @param length the length of {@code content}
-     */
-    static AggregatedHttpMessage of(
-            HttpStatus status, MediaType mediaType, byte[] content, int offset, int length) {
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/common/DefaultHttpData.java
```diff
@@ -43,22 +39,15 @@ public DefaultHttpData(byte[] data, int offset, int length, boolean endOfStream)
         return data;
     }
 
-    @Override
-    public int offset() {
```
trustin: Could deprecate and always return 0

core/src/main/java/com/linecorp/armeria/common/DefaultHttpData.java
```diff
@@ -43,22 +39,15 @@ public DefaultHttpData(byte[] data, int offset, int length, boolean endOfStream)
...
+        return data.length;
     }
 
     @Override
     @SuppressWarnings("ImplicitArrayToString")
     public String toString() {
         return MoreObjects.toStringHelper(this)
-                          .add("offset", offset)
-                          .add("length", length)
                           .add("array", data.toString()).toString();
```
trustin: This will yield a string like `[B@...`. How about just doing `.add("length", data.length)`?

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -104,7 +83,10 @@ static HttpData of(Charset charset, CharSequence text) {
 
         final CharBuffer cb = CharBuffer.wrap(text);
         final ByteBuffer buf = charset.encode(cb);
-        return of(buf.array(), buf.arrayOffset(), buf.remaining());
+
+        final byte[] bytes = new byte[buf.remaining()];
+        buf.get(bytes);
```
trustin: We could avoid making a copy when `buf.arrayOffset()` is `0` and `buf.array().length` is `buf.remaining()`.

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -236,11 +218,6 @@ static HttpData ofAscii(String format, Object... args) {
      */
     byte[] array();
 
-    /**
-     * Returns the start offset of the {@link #array()}.
-     */
-    int offset();
```
trustin: Could deprecate and always return `0`.

core/src/main/java/com/linecorp/armeria/server/annotation/ByteArrayRequestConverterFunction.java
```diff
@@ -36,16 +34,10 @@
      */
     @Override
     public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage request,
-                                 Class<?> expectedResultType) throws Exception {
+                                 Class<?> expectedResultType) {
```
trustin: We should not remove the `throws` clause because a user may extend this class and override this method.

trustin:
I wonder if it's a good idea to remove the overloaded methods that accept `offset` and `length` or not. It introduces ambiguity that forces a user to guess if the specified array will be copied or not, because no copy may be made if offset is 0 and length is same with the length of the array. However, it can be convenient in cases where you want to avoid unnecessary array copy:

```java
if (offset == 0 && length == data.length) {
    return HttpData.of(data);
} else {
    return HttpData.of(Arrays.copyOfRange(data, offset, offset + length));
}
```

trustin:
Perhaps we could rename all factory methods of `HttpData` for clarity:

- `of(byte[])` and `of(ByteBuf)` to `wrap()` or `wrapperOf()`
- `of(byte[], int offset, int length)` to `wrapOrCopy(byte[], int, int)` or `wrapperOrCopyOf(...)`

Thoughts?

anuraaga:
Yeah I was definitely thinking of renaming `of(byte[])` to `wrap`, seems like a good idea.

`wrapOrCopy` seems confusing since having a method that may reflect mutations in the input or may not seems unpredictable. In that case I would consider having `copyOf(byte[])`, `copyOf(byte[], int, int)` and `wrap(byte[], int, int)` which is implemented as `return new ByteBufHttpData(Unpooled.wrappedBuffer(bytes, offset, size), false)`. If we expose these on `Aggregated*` too though it'd be a whole lot of APIs.

trustin:
> `wrapOrCopy` seems confusing since having a method that may reflect mutations in the input or may not seems unpredictable. In that case I would consider having `copyOf(byte[])`, `copyOf(byte[], int, int)` and `wrap(byte[], int, int)` which is implemented as `return new ByteBufHttpData(Unpooled.wrappedBuffer(bytes, offset, size), false)`.

+1

> If we expose these on `Aggregated*` too though it'd be a whole lot of APIs.

Agreed. Let's leave them removed in `Aggregated*` and let users use `HttpData` when they want to slice the array.

anuraaga:
Tried changing to the proposed API let me know if it looks ok

anuraaga:
Also I'm wondering if the ByteBuf versions should be prefixed with unsafe or be in a different package

core/src/main/java/com/linecorp/armeria/client/Http1ResponseDecoder.java
```diff
@@ -192,7 +192,7 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                                 fail(ctx, ContentTooLargeException.get());
                                 return;
                             } else {
-                                res.write(HttpData.of(data));
+                                res.write(HttpData.copyOf(data));
```
trustin: I'm curious if we can use `wrap()` here.

core/src/main/java/com/linecorp/armeria/common/DefaultHttpData.java
trustin: How about removing `.add("array", ...)`?

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -62,18 +62,13 @@ public boolean isEndOfStream() {
 
     @Override
     public byte[] array() {
-        if (buf.hasArray()) {
+        if (buf.hasArray() && buf.arrayOffset() == 0) {
```
trustin: `&& buf.array().length == length` ?

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -62,18 +62,13 @@ public boolean isEndOfStream() {
...
             return buf.array();
         } else {
             return ByteBufUtil.getBytes(buf);
```
trustin: I'm curious if we have to cache the array. What do you think?

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -79,7 +83,89 @@ static HttpData of(byte[] data, int offset, int length) {
...
+        requireNonNull(data);
+        if (offset < 0 || length < 0 || offset > data.length - length) {
+            throw new ArrayIndexOutOfBoundsException(
+                    "offset: " + offset + ", length: " + length + ", data.length: " + data.length);
+        }
+        if (length == 0) {
+            return EMPTY_DATA;
+        }
+
+        return new DefaultHttpData(Arrays.copyOfRange(data, offset, length), false);
```
trustin: length -> offset+length?

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -122,21 +212,15 @@ static HttpData of(Charset charset, String text) {
...
-        return of(text.getBytes(charset));
+        return wrap(text.getBytes(charset));
     }
 
     /**
-     * Converts the specified Netty {@link ByteBuf} into an {@link HttpData}. Unlike {@link #of(byte[])}, this
-     * method makes a copy of the {@link ByteBuf}.
-     *
-     * @return a new {@link HttpData}. {@link #EMPTY_DATA} if the readable bytes of {@code buf} is 0.
+     * @deprecated use {@link #copyOf(ByteBuf)}.
```
trustin: use -> Use

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -237,9 +321,12 @@ static HttpData ofAscii(String format, Object... args) {
     byte[] array();
 
     /**
-     * Returns the start offset of the {@link #array()}.
+     * @deprecated the offset of {@link HttpData} is always {@code 0}.
```
trustin: the -> The

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -79,7 +83,89 @@ static HttpData of(byte[] data, int offset, int length) {
...
+        requireNonNull(data, "data");
+        if (!data.isReadable()) {
+            return EMPTY_DATA;
+        }
+
+        return wrap(ByteBufUtil.getBytes(data));
+    }
+
+    /**
+     * @deprecated use {@link #wrap(byte[])}.
```
trustin: use -> Use

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -79,7 +83,89 @@ static HttpData of(byte[] data, int offset, int length) {
...
+     */
+    @Deprecated
+    static HttpData of(byte[] data) {
+        return wrap(data);
+    }
+
+    /**
+     * @deprecated use {@link #wrap(byte[], int, int)}.
```
trustin: use -> Use

trustin:
Presses a wrong button. I actually requested changes. :wink:

core/src/main/java/com/linecorp/armeria/client/Http1ResponseDecoder.java
anuraaga: Yeah noticed this - didn't want to think too much about it but changed to `wrap(retain)` since seems ok.

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpRequest.java
```diff
@@ -113,26 +113,7 @@ static AggregatedHttpRequest of(HttpMethod method, String path, MediaType mediaT
...
-     * @param length the length of {@code content}
-     */
-    static AggregatedHttpRequest of(
-            HttpMethod method, String path, MediaType mediaType, byte[] content, int offset, int length) {
-        requireNonNull(method, "method");
-        requireNonNull(path, "path");
-        requireNonNull(mediaType, "mediaType");
-        requireNonNull(content, "content");
-        return of(method, path, mediaType, HttpData.of(content, offset, length));
+        return of(method, path, mediaType, HttpData.wrap(content));
```
anuraaga: I'm wondering is this method supposed to wrap or copy or be deleted. The javadoc doesn't mention the potential danger of wrapping like `HttpData` javadoc always has. Should we update the javadoc?

core/src/main/java/com/linecorp/armeria/common/DefaultHttpData.java
anuraaga: Ah I see, thanks

core/src/main/java/com/linecorp/armeria/common/HttpData.java
anuraaga: Oof - added tests

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
anuraaga: I think it's uncommon enough to call it twice that it's ok without. If I'm not mistaken adding a fourth field to this class will tip it into being a larger object size so memory usage of all apps, even if they don't use this method, will go up.

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
anuraaga: Oof - added tests

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpRequest.java
trustin: That'd be nice. How about mentioning that `byte[]` is wrapped via `HttpData.wrap()`?

trustin: Thanks!

anuraaga:
I found a gotcha with `wrap(byte[], int, int)` when implemented using an unpooled bytebuffer - previously it was possible to publish the same `HttpData` created that way multiple times, but now it isn't since the first subscriber would read all the bytes. Do we want to support this?

trustin:
> I found a gotcha with `wrap(byte[], int, int)` when implemented using an unpooled bytebuffer - previously it was possible to publish the same `HttpData` created that way multiple times, but now it isn't since the first subscriber would read all the bytes. Do we want to support this?

That's a good point. Perhaps we should introduce another `HttpData` implementation dedicated to `wrap(byte[], int, int)` which does not involve a `ByteBuf`?

anuraaga:
Tried adding a separate implementation, how does it look?

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -83,7 +82,11 @@ static HttpData wrap(byte[] data, int offset, int length) {
             return EMPTY_DATA;
         }
 
-        return new ByteBufHttpData(Unpooled.wrappedBuffer(data, offset, length), false);
+        if (data.length == length) {
```
trustin: `offset == 0 &&` ?

core/src/test/java/com/linecorp/armeria/common/HttpDataTest.java
```diff
@@ -66,6 +66,14 @@ void wrapRangeNoOffset() {
             assertThat(data.array()).containsExactly(1, 5);
         }
 
+        @Test
+        void wrapRangeFull() {
+            final HttpData data = HttpData.wrap(payload, 0, 4);
```
trustin: Could update the test case to make sure non-zero offset works.

trustin:
@anuraaga Looks good. Left just some comments.

core/src/main/java/com/linecorp/armeria/common/HttpData.java
anuraaga: I don't think it's possible because of the index check above

core/src/test/java/com/linecorp/armeria/common/HttpDataTest.java
anuraaga: `wrapRange` above has non-zero offset is there another case I missed?

core/src/main/java/com/linecorp/armeria/common/HttpData.java
trustin: Ah yeah. :sweat_smile:

trustin:
Thanks, @anuraaga!

PR #1799, opened 2019-05-29
Add document about per service configuration

minwoox:
Following #1737

### DIFF
--- a/site/src/sphinx/server-basics.rst
+++ b/site/src/sphinx/server-basics.rst
@@ -154,8 +154,8 @@ for a specific service, you can use fluent API:
     ServerBuilder sb = new ServerBuilder();
     sb.route()                                   // Configure the service.
       .post("/foo/bar")                          // Matched when the path is "/foo/bar" and the method is POST.
-      .consumes(JSON)                            // Matched when the "content-type" header is "application/json".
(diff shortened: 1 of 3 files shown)


site/src/sphinx/server-basics.rst
```diff
@@ -143,6 +143,36 @@ You might be interested in decorating a service using other decorators, for exam
...
+
+The examples above are just mapping the path of an HTTP request on a service. If you want to set configuration
+for a specific service, you can use fluent API:
+
+.. code-block:: java
+
+    ServerBuilder sb = new ServerBuilder();
+    sb.route()                                   // Configure the service.
+      .post("/foo/bar")                          // Matched when the path is "/foo/bar" and the method is POST.
+      .consumes(JSON)                            // Matched when the "content-type" header is "application/json".
```
hyangtack: How about adding `import` or specifying `MediaType.JSON` from here?

minwoox: Will use `MediaType.JSON` and add the import statement because we still need it. Thanks! :-)

minwoox: Changed. Didn't add the import statement because we already did that in the previous example. :-)

PR #1801, opened 2019-05-29
Initialize JUnit5 extensions Before/AfterAll instead of Each.

anuraaga:
I noticed that the junit5 extension starts / stops the server around every test method instead of the class. We almost always use `@ClassRule` with the old `ServerRule` since this is fairly resource intensive. Also I think it's better to guarantee these resources are set up before user's `@BeforeEach` type of callbacks. Does it make sense to change these to `All`?


trustin:
+1 for `All`. Is there a way to make this configurable by the way?

anuraaga:
It's just code, so we would need to add a constructor parameter to control the behavior. As some already have parameters, I would lean towards switching to a builder pattern for these though if we did so. Do you think it's worth it?

anuraaga:
Ah also I guess we can have a `protected boolean` method instead. What do you think?

trustin:
> Ah also I guess we can have a `protected boolean` method instead. What do you think?

SGTM

anuraaga:
Added a base class so extensions by default run around class but a method can be overridden to have them run around method instead.

trustin:
Thanks, @anuraaga

Issue #1601, opened 2019-02-17
Can't run single test of Parameterized test when delegating builds to Gradle in IntelliJ


anuraaga:
I'm not sure if this is a bug in Gradle or IntelliJ or something in the armeria configuration, but when I set IntelliJ to delegate tests to Gradle with Settings > Build, Execution, Deployment > Build Tools > Gradle > Runner, I can't run a single test out of a parameterized one.

For example, if I run `HttpServerStreamingTest.testTooLargeContentToNonExistentService` by pressing the green button next to the method, I get

```
Execution failed for task ':core:test'.
> No tests found for given includes: [**/HttpServerStreamingTest**](exclude rules) [com.linecorp.armeria.server.HttpServerStreamingTest.testTooLargeContentToNonExistentService](filter.includeTestsMatching)
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Get more help at https://help.gradle.org
```

anuraaga:
It's fine to workaround by using IntelliJ runner for tests (a bit annoying since for some reason Gradle runner is a global setting, not project-level one...) but I thought this might become more important after reading that IntelliJ 2019.1 will default to using the Gradle runner

https://github.com/tbroyer/gradle-apt-plugin/commit/b96813adfaa59731ead938f8bd20b64314f0c661

trustin:
I wonder if this is an IntelliJ IDEA issue. Didn't try the Gradle runner for quite a while because IntelliJ runner starts much faster..

PR #1802, opened 2019-05-29
Add a utility for executing junit5 tests with timeout and convert a parameterized test to it

anuraaga:
…arameterized test to it.

Was curious if junit5 has the same problem as #1601 and it doesn't! So I want to migrate all the parameterized tests to junit5 (or at least the gRPC ones since I use them a lot and it's annoying that I can't use Gradle runner because of them).

I found that the biggest missing piece from junit5 is global timeouts. 5.5 will have it, but that will probably take a while to be released. Also, I don't think they'll support disable-on-debug in their first release either. So wondering if it makes sense to add this utility for running a lambda with a timeout using the current junit5 style.

### DIFF
--- a/testing-internal/src/main/java/com/linecorp/armeria/testing/internal/TestUtil.java
+++ b/testing-internal/src/main/java/com/linecorp/armeria/testing/internal/TestUtil.java
@@ -34,7 +34,7 @@ public final class TestUtil {
     }
 
     /**
-     * Executes {@code r}, timing it out if not done by the passing of {@code timeout} has passed. The timeout
+     * Executes {@code r}, timing it out if not done by the passing of {@code timeout}. The timeout
      * is not enabled if in an IDE debug mode.
      */
     public static void withTimeout(Executable r, Duration timeout) {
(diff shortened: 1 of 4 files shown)


testing-internal/src/main/java/com/linecorp/armeria/testing/internal/TestUtil.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Executes {@code r}, timing it out if not done by the passing of {@code timeout}. The timeout
+     * is not enabled if in an IDE debug mode.
+     */
+    public static void withTimeout(Executable r, Duration timeout) {
+        if (isDebugging()) {
+            try {
+                r.execute();
+            } catch (Throwable t) {
+                throw new Error(t);
```
trustin: Maybe we can use `Exceptions.throwUnsafely()`?

anuraaga: Ah yeah that's better

PR #1803, opened 2019-05-29
Fix project evaluation dependencies

trustin:
Related: https://github.com/line/gradle-scripts/pull/61
Motivation:

When configure-on-demand is enabled, inter-project dependencies are
calculated correctly only when task dependency is specified as a task
path string or `Project.evaluationDependsOn()` must be used. See:

- https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:configuration_on_demand

> The task dependencies declared via task path are supported and cause
> relevant projects to be configured.
> Example: someTask.dependsOn(":someOtherProject:someOtherTask")

Modifications:

- Add `Project.evaluationDependsOn()` calls where necessary

Result:

- Fixes a bug where a certain project is not configured when it has to
  be, causing a weird compilation error due to unconfigured dependencies.

### DIFF
--- a/site/build.gradle
+++ b/site/build.gradle
@@ -18,13 +18,14 @@ buildscript {
 apply plugin: 'base'
 apply plugin: 'kr.motd.sphinx'
 
+def aggregatedProjects = projectsWithFlags('java', 'publish') - projectsWithFlags('no_aggregation')
+aggregatedProjects.each { evaluationDependsOn it.path }
+
 sphinx {
     group = 'Documentation'
     description = 'Generates the Sphinx web site.'
     sourceDirectory "${project.projectDir}/src/sphinx"
 }
-
-def aggregatedProjects = projectsWithFlags('java', 'publish') - projectsWithFlags('no_aggregation')
 
 task javadoc(type: Javadoc,
              group: 'Documentation',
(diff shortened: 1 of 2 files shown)


anuraaga:
Hmm I was using configure on demand in large projects for a long time and never ran into this.

Anyways it seems that Gradle team is basically depreciating the feature because configuration is very fast now anyways

https://guides.gradle.org/migrating-build-logic-from-groovy-to-kotlin/#before_you_start_migrating

While it says `shouldn't`, I've found that kotlin scripts just don't work with it on at all.

So I'd recommend

- Remove configure on demand
- Enable parallel which has a bigger impact
- Either go through the scripts to make sure they use Gradle's newer lazy APIs (register, named, etc) or migrate to Kotlin where the default DSL is lazy

trustin:
Yeah they seem to want to remove it but couldn't because at some cases it's useful: https://github.com/gradle/gradle/issues/2868 Let me revisit this PR tomorrow.

anuraaga:
Cool - just an anecdote but I've stopped using it in a couple of hugish projects to migrate the scripts to Kotlin and it's been just fine. Gradle's way faster than four years ago :D

trustin:
> Gradle's newer lazy APIs (register, named, etc)

Never knew about them. Do you know any good resources about this? :bow:

anuraaga:
Yup - think this is the main doc for it

https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

anuraaga:
In case it's useful, for reference the bug for Kotlin + CoD (not call of duty)

https://github.com/gradle/gradle/issues/4823

It reminds me that I think if going with this PR's approach, `evaluationDependsOn` might be a little more idiomatic than having to use task.path everywhere. Gradle API's are good at working no matter what (String, Task, TaskProvider, etc) is given to them so this change to force using the path feels surprising to me.

trustin:
Updated to use `evaluationDependsOn()` instead of error-prone `Task.path`

PR #1805, opened 2019-05-31
Preserve percent-encoded character after parsing URI by PathAndQuery

hyangtack:
Motivation:
Currently, percent-encoded characters only for ampersand(`&`), semi-colon(`;`) and equal(`=`) are preserved after parsing URI by `PathAndQuery`. So, if a request has a percent-encoded `#`(`%23`), it is decoded as a raw `#` character. Due to this behavior, `QueryStringDecoder` cannot parse parameters correctly if a query string has a `#`, because `#` is a fragment identifier and the decoder ignores everything after the identifier.

Modifications:
- Every percent-encoded characters for reserved characters are preserved after parsing URI by `PathAndQuery`.
- Reserved characters. (from `https://tools.ietf.org/html/rfc3986#section-2.2`)
  - `:` / `/` / `?` / `#` / `[` / `]` / `@`
  - `!` / `$` / `&` / `'` / `(` / `)` / `*` / `+` / `,` / `;` / `=`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
@@ -178,7 +178,7 @@ public final class PathAndQuery {
 
     @Override
     public String toString() {
-        return MoreObjects.toStringHelper(this)
+        return MoreObjects.toStringHelper(this).omitNullValues()
                           .add("path", path)
                           .add("query", query)
                           .toString();
@@ -351,7 +351,7 @@ public final class PathAndQuery {
             if (isPath) {
                 return false;
             } else if (cp != 0x0A && cp != 0x0D && cp != 0x09) {
-                // .. except 0x0A (LF), (byte) 0x0D (CR) and 0x09 (TAB) because they are used in a form.
+                // .. except 0x0A (LF), 0x0D (CR) and 0x09 (TAB) because they are used in a form.
                 return false;
             }
         }
@@ -538,7 +538,7 @@ public final class PathAndQuery {
 
     /**
      * Reserved characters which require percent-encoding. These values are only used for constructing
-     * {@link #RAW_CHAR_TO_MARKER} map and {@link #MARKER_TO_PERCENT_ENCODED_CHAR} map.
+     * {@link #RAW_CHAR_TO_MARKER} and {@link #MARKER_TO_PERCENT_ENCODED_CHAR} mapping tables.
      *
      * @see <a href="https://tools.ietf.org/html/rfc3986#section-2.2">RFC 3986, section 2.2</a>
      */
--- a/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
@@ -20,6 +20,7 @@ import static io.netty.util.internal.StringUtil.decodeHexNibble;
 import static java.util.Objects.requireNonNull;
 
 import java.util.BitSet;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
@@ -29,6 +30,7 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableMap.Builder;
 
 import com.linecorp.armeria.common.Flags;
 import com.linecorp.armeria.common.metric.MeterIdPrefix;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -49,6 +51,11 @@
     private static final BitSet ALLOWED_PATH_CHARS = new BitSet();
     private static final BitSet ALLOWED_QUERY_CHARS = new BitSet();
 
+    private static final int PERCENT_ENCODING_MARKER = 0xFF;
+
+    private static final Map<Integer, Integer> RAW_CHAR_TO_MARKER;
```
trustin: Could be represented as `int[256]` for better performance?

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -49,6 +51,11 @@
...
+    private static final Map<Integer, String> MARKER_TO_PERCENT_ENCODED_CHAR;
```
trustin: Could be represented as `String[]` for better performance?

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -269,14 +263,15 @@ private static Bytes decodePercentsAndEncodeToUtf8(String value, int start, int
                     }
                 } else {
                     // If query:
-                    final Integer marker = RAW_CHAR_TO_MARKER.get(decoded);
-                    if (marker != null) {
+                    assert decoded < 256;
```
trustin: I guess this isn't necessary because `AIIOBE` will be raised anyway if that ever happens.

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -269,14 +263,15 @@ private static Bytes decodePercentsAndEncodeToUtf8(String value, int start, int
...
+                    final int marker = RAW_CHAR_TO_MARKER[decoded];
+                    if (marker > 0) {
```
trustin: nit: `!=` ?

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -53,8 +51,8 @@
 
     private static final int PERCENT_ENCODING_MARKER = 0xFF;
 
-    private static final Map<Integer, Integer> RAW_CHAR_TO_MARKER;
-    private static final Map<Integer, String> MARKER_TO_PERCENT_ENCODED_CHAR;
+    private static final int[] RAW_CHAR_TO_MARKER = new int[256];
```
trustin: Realized this can be `byte[]` which is more compact.

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
hyangtack: I see. Removed. :-)

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -546,4 +535,42 @@ int nextCodePoint() {
             return c1;
         }
     }
+
+    /**
+     * Reserved characters which require percent-encoding. These values are only used for constructing
+     * {@link #RAW_CHAR_TO_MARKER} map and {@link #MARKER_TO_PERCENT_ENCODED_CHAR} map.
```
minwoox: nit; not map anymore?

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -365,7 +351,7 @@ private static boolean appendOneByte(Bytes buf, int cp, boolean wasSlash, boolea
             if (isPath) {
                 return false;
             } else if (cp != 0x0A && cp != 0x0D && cp != 0x09) {
-                // .. except 0x0A (LF), 0x0D (CR) and 0x09 (TAB) because they are used in a form.
+                // .. except 0x0A (LF), (byte) 0x0D (CR) and 0x09 (TAB) because they are used in a form.
```
minwoox: Is that a reason that you put `(byte)` only in front of `0x0D`?

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
hyangtack: Oops! Thanks!

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -185,9 +179,9 @@ public int hashCode() {
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this)
-                .add("path", path)
-                .add("query", query)
-                .toString();
+                          .add("path", path)
+                          .add("query", query)
```
minwoox: Is it better to omit `null` value?

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
hyangtack: Replacement failure. 😅

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
hyangtack: Good catch. Added! :-)

trustin:
Thanks, @hyangtack :-)

PR #1806, opened 2019-05-31
Migrate from Parameterized to JUnit 5.

### LINKED ISSUE #1601: Can't run single test of Parameterized test when delegating builds to Gradle in IntelliJ

anuraaga:
junit5 parameters are fundamentally different than junit4 in that they are passed to test methods themselves instead of the test constructor. This allows having parameterized tests and non-parameterized ones in the same file (this PR removes a few instances where we ran the exact same test multiple times since it didn't use the parameter of the test class) though it does create a little more boilerplate in many instances.

The most complicated migration was for `StreamMessage`, which I also simplified by removing the inheritance, which doesn't help much since we removed `EventLoopStreamMessage`.

On the flip side, I could use parameters in the junit5 version of the spring test :-)

Fixes #1601


testing-internal/src/main/java/com/linecorp/armeria/testing/internal/UniqueProtocolsProvider.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+
+import com.linecorp.armeria.common.SessionProtocol;
+
+/**
+ * A {@link ArgumentsProvider} that provides all the unique {@link SessionProtocol} Armeria supports, which is
```
minwoox: nit: A -> an

trustin:
Thanks, @anuraaga :-)

Issue #1781, opened 2019-05-21
Does documentation for RetryClient loss some detail?


JunoJunho:
https://line.github.io/armeria/client-retry.html

In above link, I do not get any difference between two code chunks except import statements.

If you have any other intention, please tell me and let me fix.

minwoox:
Hello @JunoJunho, thanks for the report.
Unfortunately, they are different examples. The first one uses `ClientBuilder` so the `build()` method has a parameter `HttpClient.class`. Meanwhile, the second one uses `HttpClientBuilder` so the `build()` method does not take any parameter.

minwoox:
But the scheme part of the URI is different when using `ClientBuilder` and `HttpClientBuilder`. The scheme should start with `none+http(s)` for `ClientBuilder` and `http(s)` for `HttpClientBuilder`. So I think it's better to update those examples. Would you mind sending a PR for that?

JunoJunho:
Ah, there is a difference. I totally did not notice that.

Firstly, I though that if there is a difference in that topic, then it should be related with `RetryClient`. What you said is difference between `ClientBuilder` and `HttpClientBuilder`. I am not sure still about on it.

Anyway, I will happily prepare for that PR.

trustin:
Much appreciated (in advance :stuck_out_tongue_winking_eye:), @JunoJunho!

JunoJunho:
We can end this issue!

trustin:
Nice work, @JunoJunho :rocket:

PR #1807, opened 2019-06-02
Update clinet-retry.rst

JunoJunho:
Motivation:
- (To me,) It is hard to understand difference between first and second example in retry-client.rst
- Related to #1781 

Modification:
 - Add uri for ClientBuilder and HttpClientBuilder to understand difference easily (thanks to suggestion @minwoox 

Results:
- Update document in retry-client.rst

I don't know it can give little help for someone like me.

### DIFF
--- a/site/src/sphinx/client-retry.rst
+++ b/site/src/sphinx/client-retry.rst
@@ -18,7 +18,7 @@ So, let's find out what we can do with :api:`RetryingClient`.
 ``RetryingClient``
 ------------------
 
-You can just use the ``decorator()`` method in :api:`HttpClientBuilder` to build a :api:`RetryingHttpClient`:
+You can just use the ``decorator()`` method in :api:`ClientBuilder` or :api:`HttpClientBuilder` to build a :api:`RetryingHttpClient`. For example:
 
 .. code-block:: java
 
--- a/site/src/sphinx/client-retry.rst
+++ b/site/src/sphinx/client-retry.rst
@@ -31,20 +31,20 @@ You can just use the ``decorator()`` method in :api:`ClientBuilder` to build a :
     import com.linecorp.armeria.common.HttpResponse;
 
     RetryStrategy strategy = RetryStrategy.onServerErrorStatus();
-    HttpClient client = new ClientBuilder(...)
+    HttpClient client = new ClientBuilder("none+http://example.com/hello")
             .decorator(RetryingHttpClient.newDecorator(strategy))
             .build(HttpClient.class);
 
     final AggregatedHttpResponse res = client.execute(...).aggregate().join();
 
-or even simply,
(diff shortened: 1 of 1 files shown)


site/src/sphinx/client-retry.rst
```diff
@@ -31,20 +31,20 @@ You can just use the ``decorator()`` method in :api:`ClientBuilder` to build a :
     import com.linecorp.armeria.common.HttpResponse;
 
     RetryStrategy strategy = RetryStrategy.onServerErrorStatus();
-    HttpClient client = new ClientBuilder(...)
+    HttpClient client = new ClientBuilder("none+http://example.com/hello")
             .decorator(RetryingHttpClient.newDecorator(strategy))
             .build(HttpClient.class);
```
trustin: How about using `HttpClientBuilder`?

```java
import com.linecorp.armeria.client.HttpClientBuilder;

HttpClient client = new HttpClientbuilder("http://example.com/hello")
        .decorator(RetryingHttpClient.newDecorator(strategy))
        .build();
```

trustin: Never mind. It has to be left as it is.

minwoox:
I realized that people tend to use the first example of what they encounter and didn't read all the examples thoroughly.
So I think we can just remove the `ClientBuilder` example.
Sorry about that I changed my mind. 😉

JunoJunho:
Hmm, then, @minwoox , you think `HttpClientBuilder` example is more precise for this document?
Retry document covers `RetryingHttpClient` and `RetryingRpcClient`. I think your suggestion is meaningful.

I will try to change document for that if one of other two reviews agree :)

hyangtack:
+1 for removing the `ClientBuilder` example. :-)

trustin:
Yes, please.

JunoJunho:
I update doc.

gentle ping~

site/src/sphinx/client-retry.rst
```diff
@@ -18,33 +18,18 @@ So, let's find out what we can do with :api:`RetryingClient`.
 ``RetryingClient``
 ------------------
 
-You can just use the ``decorator()`` method in :api:`ClientBuilder` to build a :api:`RetryingHttpClient`:
+You can just use the ``decorator()`` method in :api:`HttpClientBuilder` to build a :api:`RetryingHttpClient`:
```
trustin: How about:

```rst
... in :api:`ClientBuilder` or :api:`HttpClientBuilder` ...
```

.. just in case it's unclear?

JunoJunho: Hmm,, let me modify your intention on commit.

trustin:
It somehow ended up with something different than the original intention of this pull request. Thank you so much for your patience, @JunoJunho !

PR #1808, opened 2019-06-03
Add SubscriptionOption

### LINKED ISSUE #1700: Add a StreamMessage subscription option that notifies cancellation to onError()
When a subscriber of a `StreamMessage` wants to get notified when a subscription has been cancelled, it has to add a callback to `StreamMessage.completionFuture()`. It is error-prone and cumbersome:

```java
StreamMessage m = ...;
MySubscriber s = ...;
m.subscribe(s);
m.completionFuture().handle(s);

class MySubscriber implements Subscriber<...>, BiFunction<...> {
    ...
    public void onComplete() { /* no-op */ }
    public void onError(Throwable cause) { /* no-op */ }
    public Void apply(result, cause) {
        if (cause == null) { /* handle success */ }
        else { /* handle failure or cancellation */ }
    }
}
```

Instead, we could provide an option that makes `StreamMessage` to notify a cancellation event to `onError()` so that a user doesn't have to add additional callback:

```java
StreamMessage m = ...;
MySubscriber s = ...;
m.subscribe(s, /* some option */);

class MySubscriber implements Subscriber<...>{
    ...
    public void onComplete() { /* handle success */ }
    public void onError(Throwable cause) { /* handle failure or cancellation */ }
}
```

Currently, `subscribe()` method is already crowded with overloading, so we may want to do something like this:

```java
m.subscribe(s, SubscriptionOption.NOTIFY_CANCELLATION,
            SubscriptionOption.WITH_POOLED_OBJECTS);

// or

m.subscribe(s, executor, SubscriptionOption.NOTIFY_CANCELLATION,
            SubscriptionOption.WITH_POOLED_OBJECTS);
```

Alternatively, we could even make `Executor` an option:

```java
m.subscribe(s, ImmutableMap.of(
    SubscriptionOption.NOTIFY_CANCELLATION, true,
    SubscriptionOption.WITH_POOLED_OBJECTS, true,
    SubscriptionOption.EXECUTOR, executor));
```

I think I prefer the one that does not use a `Map`, although the latter is more flexible. Thoughts? /cc @anuraaga

minwoox:
Motivation:
Currently, `Subscriber.onError(cause)` is not invoked when a `StreamMessage` is canceled.
This is because the subscriber is able to handle the situation after canceling the `StreamMessage`
so that it does not have to wait for `Subscriber.onError(cause)`.
However, if we provide an option which calls `Subscriber.onError(cause)` with the `CancelledSusbscriptionException`,
the user might set the ending logic in the `onError` method.

Modifications:
- Add `SubscriptionOption` to specify when calling `StreamMessage.subscribe()`.
  - `WITH_POOLED_OBJECTS`: to receive pooled `ByteBuf` and `ByteBufHolder`.
  - `NOTIFY_CANCELLATION`: to get notified by `Subscriber.onError(cause)` when canceling.
- Deprecated
  - `StreamMessage.subscribe(..., withPooledObject)` -> `StreamMessage.subscribe(..., SubscriptionOption.WITH_POOLED_OBJECTS)`

Result:
- You can now get notified in `Subscriber.onError(cause)` when you cancel the `StreamMessage`.
- Close #1700

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
@@ -16,8 +16,13 @@
 
 package com.linecorp.armeria.common.stream;
 
+import static com.linecorp.armeria.common.stream.StreamMessageUtil.containsNotifyCancellation;
+import static com.linecorp.armeria.common.stream.StreamMessageUtil.containsWithPooledObjects;
+import static com.linecorp.armeria.common.stream.SubscriptionOption.NOTIFY_CANCELLATION;
+import static com.linecorp.armeria.common.stream.SubscriptionOption.WITH_POOLED_OBJECTS;
 import static java.util.Objects.requireNonNull;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
@@ -44,6 +49,8 @@ public abstract class FilteredStreamMessage<T, U> implements StreamMessage<U> {
 
     private static final Logger logger = LoggerFactory.getLogger(FilteredStreamMessage.class);
 
+    private static final SubscriptionOption[] EMPTY_OPTIONS = new SubscriptionOption[0];
+
     private final StreamMessage<T> delegate;
     private final boolean filterSupportsPooledObjects;
 
@@ -65,7 +72,7 @@ public abstract class FilteredStreamMessage<T, U> implements StreamMessage<U> {
      */
     protected FilteredStreamMessage(StreamMessage<T> delegate, boolean withPooledObjects) {
         this.delegate = requireNonNull(delegate, "delegate");
-        this.filterSupportsPooledObjects = withPooledObjects;
+        filterSupportsPooledObjects = withPooledObjects;
     }
 
     /**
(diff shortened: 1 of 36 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -191,9 +204,6 @@ public void onSubscribe(Subscription s) {
         @Override
         public void onNext(T o) {
             ReferenceCountUtil.touch(o);
-            if (!filterSupportsPooledObjects) {
```
minwoox: @anuraaga I found out that the object is always `false` when `subscribedWithPooledObjects` is `false` even though `filterSupportsPooledObjects` is `true`. So I changed to use `filterSupportsPooledObjects` when subcribing the `delegate`. PTAL 🙇

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -270,12 +239,12 @@ default boolean isComplete() {
...
-     * @param withPooledObjects if {@code true}, receives the pooled {@link ByteBuf} and {@link ByteBufHolder}
-     *                          as is, without making a copy. If you don't know what this means, use
-     *                          {@link StreamMessage#drainAll()}.
+     * @param options {@link SubscriptionOption}s to subscribe with. Note that
+     *                {@link SubscriptionOption#NOTIFY_CANCELLATION} is ineffective because there's no
+     *                cancelling while draining all elements.
      * @return the {@link CompletableFuture} which will be completed with the list of the elements retrieved.
      */
-    CompletableFuture<List<T>> drainAll(boolean withPooledObjects);
+    CompletableFuture<List<T>> drainAll(SubscriptionOption... options);
```
minwoox: I changed this to take the `SubscriptionOption`s even though I don't have to do because we don't call `Subscription.cancel()` inside.
I thought it would be better for consistency with the `subscribe` methods and we might add more options to `SubscipriptionOption` later. So I changed it, but if you think we don't have to, just let me know. I will undo it. :-)

anuraaga:
Is it possible to keep the old methods and deprecate them?

minwoox:
> Is it possible to keep the old methods and deprecate them?

I thought there's a reason that I removed all the old methods, but I couldn't think of it now. 😅  Will add them back.

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
```diff
@@ -116,53 +123,60 @@ public boolean isEmpty() {
...
-        delegate.subscribe(new FilteringSubscriber(subscriber, withPooledObjects), executor, withPooledObjects);
+        requireNonNull(options, "options");
+        delegate.subscribe(new FilteringSubscriber(subscriber, containsWithPooledObjects(options)),
+                           executor, filteringSubscriptionOptions(options).toArray(EMPTY_OPTIONS));
     }
 
-    @Override
-    public CompletableFuture<List<U>> drainAll() {
-        return drainAll(false);
+    private ArrayList<SubscriptionOption> filteringSubscriptionOptions(SubscriptionOption... options) {
```
trustin: Could just return `SubscriptionOption[]`? All callers convert it to an array.

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageUtil.java
```diff
@@ -28,5 +33,13 @@ static Throwable abortedOrLate(Subscriber<?> oldSubscriber) {
         return new IllegalStateException("subscribed by other subscriber already");
     }
 
+    static boolean containsWithPooledObjects(SubscriptionOption... options) {
+        return Stream.of(options).anyMatch(option -> option == WITH_POOLED_OBJECTS);
+    }
+
+    static boolean containsNotifyCancellation(SubscriptionOption... options) {
+        return Stream.of(options).anyMatch(option -> option == NOTIFY_CANCELLATION);
```
trustin: Could you optimize these two methods using plain foreach loops?

core/src/main/java/com/linecorp/armeria/common/stream/SubscriptionOption.java
```diff
@@ -0,0 +1,41 @@
...
+package com.linecorp.armeria.common.stream;
+
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
+
+/**
+ * Options used when subscribing a {@link StreamMessage}.
```
trustin: - subscribing to
- How about adding `@see StreamMessage#subscribe(...., SubscriptionOption...)`?

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
anuraaga: I think normally we'd add an `Iterable` override to `subscribe` instead of having callers always pass in an array. Would that make more sense? Since we can use an `EnumSet`, delegating things to an `Iterable` should be quite cheap too.

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
anuraaga: Ah this seems more natural, thanks

core/src/test/java/com/linecorp/armeria/common/stream/AbstractStreamMessageTest.java
```diff
@@ -17,6 +17,8 @@
 package com.linecorp.armeria.common.stream;
```
anuraaga: I'm selfishly hoping #1806 gets merged first XD

minwoox: No problem at all. :-) I will remove some duplicated code after #1806 is merged.

core/src/main/java/com/linecorp/armeria/common/stream/FilteredStreamMessage.java
minwoox: I didn't add those variants to reduce the number of public APIs. But I can add them if we need.

anuraaga: Yeah it could be overkill - Java APIs like `Files` usually don't have an iterator. It mainly came to mind since things like optimizing `containsX` below becomes much easier if using `EnumSet`. If it's possible to have it only in the internal API that would work too, but no worries if it seems too complicated.

trustin: It's gonna be just a two-element array, so I guess delegating things to an `Iterable` is probably an overkill?

minwoox: Changed to return `SubscriptionOption[]`, thanks!

> `containsX` below becomes much easier if using `EnumSet`

Yeah, that made me almost to add the `Iterable` version. :-)

> If it's possible to have it only in the internal API that would work too, but no worries if it seems too complicated.

Yeah, it seems complicated as you said. :-) I will keep that in my mind, and add it when it's necessary.

minwoox:
All fixed.
@anuraaga Add the old methods back and make them deprecated.
@trustin Add `SubsriptionOptionTest` to integrate the related tests and remove the duplicated code in `FileteredStreamMessageTest`.
PTAL. 🙇

core/src/test/java/com/linecorp/armeria/common/stream/FilteredStreamMessageTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+import com.linecorp.armeria.unsafe.ByteBufHttpData;
+
+import io.netty.buffer.ByteBufHolder;
+
+class FilteredStreamMessageTest {
+
+    @Test
+    void withPooledObjects() {
+        boolean filterSupportsPooledObjects = true;
```
anuraaga: I think it's better to either use `@ParameterizedTest` or split into four tests.

minwoox: Ah, yeah, will use `@ParameterizedTest`. Thanks!

minwoox:
Ah thanks, I didn't have to deprecate the `drailAll()`.
But I'm curious if we could just leave the methods which take varargs as a parameter and get rid of the all the methods which didn't take them like we did in `ClientFactory.newClient(...)`

trustin:
> But I'm curious if we could just leave the methods which take varargs as a parameter and get rid of the all the methods which didn't take them like we did in `ClientFactory.newClient(...)`

Performance-wise, it'd be nice to have a version without varargs so that a zero-arg call does not create an empty array unnecessarily.

minwoox:
Thanks a lot for reviewing. 😉

Issue #653, opened 2017-06-30
Implement the go-to form for DocService debug page


trustin:
Sometimes, a Server can serve a large number of methods. It would be useful if we add a go-to form with auto-completion, just like the 'go-to class' features in a modern IDE.

PR #1812, opened 2019-06-04
Add docs proxy prefix in WEBPACK_SERVE mode of docs-client

ikhoon:
Motivation
* Current Armeria server is able to set prefix of `DocService`,
  but docs-client with `WEBPACK_SERVE` mode can't set prefix to proxy to DocService.

Modification
* Add `PROXY_DOC_PREFIX` evironment variable with default value of '/docs'.
* Enable proxy logs when `WEBPACK_SERV` is true for better debug experience.
* Add rewrite rule in proxy to set prefix path.

Result
* We can control proxy doc prefix using the evironment variable `PROXY_DOC_PREFIX` in WEBPACK_SERVE mode.

### DIFF
--- a/docs-client/README.md
+++ b/docs-client/README.md
@@ -18,21 +18,25 @@ $ ./gradlew :docs-client:yarn_run_start --no-daemon
 ```
 
 This will usually not be useful since without a server running, the client does not have any spec it can render.
-You can have server calls proxied to a running Armeria server by specifying the `PROXY_PORT` environment
-variable, e.g.,
+You can have server calls proxied to a running Armeria server by specifying environment variables.
+
+* `PROXY_PORT`: Armeria server port (default: 8080)
+* `PROXY_DOC_PREFIX`: Armeria server DocService path prefix (default: '/docs')
+
+e.g.
 
 ```bash
-$ PROXY_PORT=51234 yarn run start
+$ PROXY_PORT=51234 PROXY_DOC_PREFIX='/documentation' yarn run start
 ```
 
 or with Gradle
 
 ```bash
-$ PROXY_PORT=51234 ./gradlew :docs-client:yarn_run_start --no-daemon
+$ PROXY_PORT=51234  PROXY_DOC_PREFIX='/documentation' ./gradlew :docs-client:yarn_run_start --no-daemon
 ```
 
 Replacing the port of a docs page in the running server with `3000` will use the dev server to render while
-proxying all server calls to the actual Armeria server.
+proxying all server calls to the actual Armeria server port and docs path.
 
 ## Updating licenses
 
--- a/docs-client/scripts/serve.ts
+++ b/docs-client/scripts/serve.ts
@@ -42,10 +42,13 @@ async function historyFallback(ctx: any, next: any) {
 }
 
 const proxyPort = process.env.PROXY_PORT || '8080';
+const proxyDocPrefix = process.env.PROXY_DOC_PREFIX || '/docs';
 
 const proxier = proxy('/', {
   target: `http://127.0.0.1:${proxyPort}`,
   changeOrigin: true,
+  logs: process.env.WEBPACK_SERVE,
+  rewrite: (path: string) => `${proxyDocPrefix}${path}`,
 });
 
 async function proxyToApi(ctx: any, next: any) {
--- a/docs-client/scripts/serve.ts
+++ b/docs-client/scripts/serve.ts
@@ -47,7 +47,7 @@ const proxyDocPrefix = process.env.PROXY_DOC_PREFIX || '/docs';
 const proxier = proxy('/', {
   target: `http://127.0.0.1:${proxyPort}`,
   changeOrigin: true,
-  logs: process.env.WEBPACK_SERVE,
+  logs: !!process.env.WEBPACK_SERVE,
   rewrite: (path: string) => `${proxyDocPrefix}${path}`,
 });
 


docs-client/scripts/serve.ts
```diff
@@ -42,10 +42,13 @@ async function historyFallback(ctx: any, next: any) {
 }
 
 const proxyPort = process.env.PROXY_PORT || '8080';
+const proxyDocPrefix = process.env.PROXY_DOC_PREFIX || '/docs';
 
 const proxier = proxy('/', {
   target: `http://127.0.0.1:${proxyPort}`,
   changeOrigin: true,
+  logs: process.env.WEBPACK_SERVE,
```
ikhoon: I wonder it is good or not. 🤔

ikhoon:
While I'm looking around Armeria features, I found docs-client using `WEBPACK_SERVE` doesn't work without code modification.
Because It failed to find `specification.json`. So I made this PR. 

Please let me know if it needs any correction. :-)

docs-client/scripts/serve.ts
anuraaga: !!process.env.WEBPACK_SERVE would make it explicitly boolean for clarity.

Though this is only used in development so `true` would be fine too

ikhoon:
> Thanks! This LGTM but I'm just curious how you are using the dev server. 

I have been looking around issues with [`good-first-issue`](https://github.com/line/armeria/issues?q=is%3Aopen+is%3Aissue+label%3Agood-first-issue)which I can contribute. :-) 
I saw one issue, [Implement the go-to form for DocService debug page](https://github.com/line/armeria/issues/653), no one assigned yet. To solve the issue I'm setting up the development everionment. That's why I'm using the dev server. 😀

> It was designed for development of the docs page in Armeria and most of our test cases use `/docs` so the fixed path hasn't been an issue. Can I ask how are you using it so that it's at another prefix?

It could be my mistake, the proxy has routed to wrong urls which don't have prefix `/docs` every time I run `yarn run start`.
```
2019-06-05T00:55:53.054Z - GET /specification.json proxy to -> http://127.0.0.1:8080/specification.json
```
Let me know if any wrong with me :-)

docs-client/scripts/serve.ts
ikhoon: Thanks! it looks better. 👍

anuraaga:
Ah thanks I'm remembering a little better now. Let me look a little at it, I think there's something wrong with the client that it's requesting from the wrong URL (it would affect production too, not just development mode if it's requesting the specification from the wrong place).

anuraaga:
Ah - I wonder if the README was confusing. When it says `Replacing the port of a docs page in the running server with 3000` it really means it. You should access the dev server at `http://localhost:3000/docs/`, not `http://localhost:3000/`. Does that work for you?

If it works, I don't think we should add the rewriting since we want to verify the docs client works in a way closer to production (`/docs/`) instead of using rewriting. But we should extend the README if it was confusing.

trustin:
How about just redirecting / to `/docs/` and updating `README.md`?

docs-client/README.md
```diff
@@ -18,21 +18,25 @@ $ ./gradlew :docs-client:yarn_run_start --no-daemon
...
+You can have server calls proxied to a running Armeria server by specifying environment variables.
+
+* `PROXY_PORT`: Armeria server port (default: 8080)
+* `PROXY_DOC_PREFIX`: Armeria server DocService path prefix (default: '/docs')
+
+e.g.
 
 ```bash
-$ PROXY_PORT=51234 yarn run start
+$ PROXY_PORT=51234 PROXY_DOC_PREFIX='/documentation' yarn run start
```
trustin: - `'` seems redundant.
- How about `/other_docs` ?

docs-client/README.md
```diff
@@ -18,21 +18,25 @@ $ ./gradlew :docs-client:yarn_run_start --no-daemon
...
 ```
 
 or with Gradle
 
 ```bash
-$ PROXY_PORT=51234 ./gradlew :docs-client:yarn_run_start --no-daemon
+$ PROXY_PORT=51234  PROXY_DOC_PREFIX='/documentation' ./gradlew :docs-client:yarn_run_start --no-daemon
```
trustin: - `'` seems redundant.
- How about `/other_docs` ?

docs-client/README.md
```diff
@@ -18,21 +18,25 @@ $ ./gradlew :docs-client:yarn_run_start --no-daemon
 ```
 
 This will usually not be useful since without a server running, the client does not have any spec it can render.
-You can have server calls proxied to a running Armeria server by specifying the `PROXY_PORT` environment
-variable, e.g.,
+You can have server calls proxied to a running Armeria server by specifying environment variables.
+
+* `PROXY_PORT`: Armeria server port (default: 8080)
```
minwoox: How about renaming `PROXY_PORT` to `ARMERIA_PORT`? I realized that I was a little bit confused whether the `PROXY_PORT` means the port of Armeria server or the proxy server. /cc @anuraaga

anuraaga:
Think we didn't settle on a good behavior :)

I think the ideal one would be "redirect `/` to `/docs/` if `/` is 404". It might be too complicated to implement to be worth it though.

I don't think we should rewrite requests - Armeria doc service doesn't really need it, so it seems fragile to do it during development-only.

Also I'm planning on migrating this to `webpack-dev-server` (there was a kerfuffle where it was deprecated, `webpack-serve` emerged, `webpack-dev-server` got undeprecated, and `webpack-serve` got deprecated...) so this whole file would probably go away. While doing that, I can keep in mind the `/docs` without trailing slash shows a confusing page in mind and try to handle it - how does that sound? In the meantime, I think we just have to temporarily add one sentence to the README - "Note, the URL used when accessing the dev server must end with a slash".

ikhoon:
@anuraaga Thanks for your kind feedback. :-)

> I think the ideal one would be "redirect / to /docs/ if / is 404". It might be too complicated to implement to be worth it though.

I totally agreed with your opinion. We don't have to spend much time on this because it is a development feature and not a critical issue.

> Also I'm planning on migrating this to webpack-dev-server (there was a kerfuffle where it was deprecated, webpack-serve emerged, webpack-dev-server got undeprecated, and webpack-serve got deprecated...) 

I'm about to close this issue and support your job. I hope `webpack-dev-server` will be stable and undeprecated!

PR #1815, opened 2019-06-06
Implement the go-to form for DocService debug page

### LINKED ISSUE #653: Implement the go-to form for DocService debug page

ikhoon:
Fixes #653

Motivation
* In a server which serves a large number of methods, it takes time to go methods in debug page.

Modification
* Add [react-select](https://github.com/JedWatson/react-select)@3.0.4 dependency on docs-client for auto selection
* Convert docs specification into suggestion model
* Put the `go-to` input on the rightmost side of the `AppBar`

Result
* We can access any of the methods, structs, enums, and exceptions quickly
by using auto-complete or up-down key navigation.

### DIFF
--- a/docs-client/src/components/GotoSelect/index.tsx
+++ b/docs-client/src/components/GotoSelect/index.tsx
@@ -255,9 +255,7 @@ function makeSuggestions(
           .map((method) => {
             return {
               label: `${service.name}#${method.name}|${method.httpMethod}`,
-              value: `/methods/${service.name}/${method.name}/${
-                method.httpMethod
-              }`,
+              value: `/methods/${service.name}/${method.name}/${method.httpMethod}`,
             };
           })
           .filter(predicateWithLimit);
--- a/docs-client/src/components/GotoSelect/index.tsx
+++ b/docs-client/src/components/GotoSelect/index.tsx
@@ -98,7 +98,6 @@ const styles = (theme: Theme) =>
       marginTop: theme.spacing.unit,
       left: 0,
       right: 0,
-      // width: '125%',
       flex: 1,
     },
     divider: {
@@ -313,10 +312,8 @@ class GotoSelect extends React.Component<GotoSelectProps> {
             classes={classes}
             styles={selectStyles}
             inputId="go-to-select"
-            // placeholder="Go to ..."
             options={makeSuggestions(specification)}
             components={components}
-            // value={this.state.single}
             onChange={handleSelection}
           />
         </NoSsr>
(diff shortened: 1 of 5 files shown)


ikhoon:
## Here is a screenshot for the feature :-) 

<image src="https://user-images.githubusercontent.com/1866157/59041505-b6dd7100-88b3-11e9-87d5-edc826751620.gif" width=800 />

docs-client/src/components/GotoSelect/index.tsx
```diff
@@ -0,0 +1,328 @@
...
+};
+
+function makeSuggestions(specification: Specification): GroupType[] {
+  const suggestions: GroupType[] = [];
+
+  if (specification.getServices().length > 0) {
+    suggestions.push({
+      label: 'Services',
+      options: specification.getServices().flatMap((service) =>
+        service.methods.map((method) => {
```
trustin: How about limiting the number of items shown in the suggestions, just in case we have too many of them?

docs-client/src/components/GotoSelect/index.tsx
```diff
@@ -0,0 +1,328 @@
...
+      marginLeft: `${theme.spacing.unit}px`,
+      left: 2,
+    },
+    paper: {
+      position: 'absolute',
+      zIndex: 1,
+      marginTop: theme.spacing.unit,
+      left: 0,
+      right: 0,
+      // width: '125%',
```
trustin: Could be removed?

ikhoon: Sure! I missed it :-)

trustin:
@ikhoon Could you also run `yarn licenses generate-disclaimer --prod > ../licenses/web-licenses.txt` to update the disclaimer?

docs-client/src/components/GotoSelect/index.tsx
ikhoon: It seems good to limit the numbers of showing items. I didn't catch it. :-)
But I think users got confused when the items which exist are not shown just after typing.

I guess the memory footprint of suggestions is almost similar to specifications. But the items which are shown on screen have an impact on memory usage and rendering time.

What about the logic below?
* To limit the initial number of items around 100 for up|down key navigation.
* To reload items until 20 when filtering the items(typing in the control).

trustin: Sounds good to me.

trustin:
Could you rebase? I've just updated the front-end dependencies. (#1819)

ikhoon:
I have dealt with 
* Rebase master branch
* Update disclaimer
* Implement to limit default to 100, filtered things to 20.

trustin:
Thanks, @ikhoon and welcome to the team! 🚀

PR #1817, opened 2019-06-07
Fix TSLint errors

trustin:


trustin:
Seems like there's something else that's causing a build error on my machine.

PR #1819, opened 2019-06-07
Update frontend dependencies

trustin:
- Migrate from `@babel/polyfills` to `core-js` as advised by Babel 7.4
- Fixed TSLint errors

### DIFF
--- a/docs-client/webpack.config.ts
+++ b/docs-client/webpack.config.ts
@@ -48,7 +48,9 @@ const config: Configuration = {
                     modules: false,
                     useBuiltIns: 'entry',
                     corejs: 3,
-                    targets: { browsers: ['last 2 versions'] },
+                    targets: {
+                      browsers: ['>1%', 'not ie 11', 'not op_mini all'],
+                    },
                   },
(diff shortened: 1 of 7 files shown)


docs-client/webpack.config.ts
```diff
@@ -47,6 +47,7 @@ const config: Configuration = {
                   {
                     modules: false,
                     useBuiltIns: 'entry',
+                    corejs: 3,
                     targets: { browsers: ['last 2 versions'] },
```
anuraaga: Thanks - while we're here, how about updating this browserlist query to this one

https://browserl.ist/?q=>1%25%2C+not+ie+11%2C+not+op_mini+all

I found that I've been using `last 2 versions` which is fairly outdated practice, it includes all browsers, including IE10 which has no support for ES6 at all and everything gets polyfilled! I think our jar will get quite a bit smaller.

trustin: Updated. Down-sized from 1.22 MiB to 1.15 MiB.

PR #1828, opened 2019-06-11
Add unit test that uses gRPC Context.

anuraaga:
While looking into gRPC context, I noticed the implementation is very similar to Armeria's, using a thread-local. So didn't find any reason not to work, and at least this example seems to be working ok. Let's add the test for now and follow up on what use cases don't work.

/cc @andrewoma


trustin:
Thanks, @anuraaga !

PR #1829, opened 2019-06-11
Add a main class to `:examples:annotated-http-service`

### LINKED ISSUE #1647: `:examples:annotated-http-service` does not have a main class.
I'd be nice to have a `Main` class so that a user can play with it easily, like other examples.

hyangtack:
Motivation:
It'd be nice to have a `Main` class so that a user can play with it easily, like other examples.

Modifications:
- Replaced `ServerFactory` with `Main` class.
- Added `DocService` to the server.

Result:
- Closes #1647


hyangtack:
Thanks, reviewers!

PR #1830, opened 2019-06-11
RequestMetricSupport: use Micrometer dotted notation in metric names

perlun:
The [Micrometer naming conventions](http://micrometer.io/docs/concepts#_naming_meters) suggest that dotted notation (`foo.bar.baz`) should be used when naming meters. This has the advantage of letting different Micrometer monitoring implementations be able to convert it to the naming convention which is native for it (e.g. JMX, Prometheus etc).

Hence, the suggested change.

The PR changes the Dropwizard metric names slightly:

- `requestDuration` -> `request.duration`
- ` requestLength` -> `request.length`
- `responseDuration` -> `response.duration`
- `responseLength` -> `response.length`
- `totalDuration` -> `total.duration`

Prometheus metric names are unaffected by the change.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/metric/MoreNamingConventions.java
+++ b/core/src/main/java/com/linecorp/armeria/common/metric/MoreNamingConventions.java
@@ -37,6 +37,7 @@ import io.micrometer.prometheus.PrometheusNamingConvention;
 /**
  * Provides commonly-used {@link NamingConvention}s.
  */
+@Deprecated
 public final class MoreNamingConventions {
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
@@ -66,7 +66,7 @@ public final class RequestMetricSupport {
         final RequestContext ctx = log.context();
         final MeterRegistry registry = ctx.meterRegistry();
         final MeterIdPrefix activeRequestsId = meterIdPrefixFunction.activeRequestPrefix(registry, log)
-                                                                    .append("activeRequests");
+                                                                    .append("active.requests");
 
         final ActiveRequestMetrics activeRequestMetrics = MicrometerUtil.register(
                 registry, activeRequestsId, ActiveRequestMetrics.class,
@@ -231,15 +231,15 @@ public final class RequestMetricSupport {
             failure = parent.counter(requests, idPrefix.tags("result", "failure"));
 
             requestDuration = newTimer(
-                    parent, idPrefix.name("requestDuration"), idPrefix.tags());
+                    parent, idPrefix.name("request.duration"), idPrefix.tags());
             requestLength = newDistributionSummary(
-                    parent, idPrefix.name("requestLength"), idPrefix.tags());
+                    parent, idPrefix.name("request.length"), idPrefix.tags());
             responseDuration = newTimer(
-                    parent, idPrefix.name("responseDuration"), idPrefix.tags());
+                    parent, idPrefix.name("response.duration"), idPrefix.tags());
             responseLength = newDistributionSummary(
-                    parent, idPrefix.name("responseLength"), idPrefix.tags());
+                    parent, idPrefix.name("response.length"), idPrefix.tags());
             totalDuration = newTimer(
-                    parent, idPrefix.name("totalDuration"), idPrefix.tags());
+                    parent, idPrefix.name("total.duration"), idPrefix.tags());
         }
 
         @Override
@@ -305,11 +305,11 @@ public final class RequestMetricSupport {
             this.idPrefix = idPrefix;
 
             connectionAcquisitionDuration = newTimer(
-                    parent, idPrefix.name("connectionAcquisitionDuration"), idPrefix.tags());
-            dnsResolutionDuration = newTimer(parent, idPrefix.name("dnsResolutionDuration"), idPrefix.tags());
-            socketConnectDuration = newTimer(parent, idPrefix.name("socketConnectDuration"), idPrefix.tags());
+                    parent, idPrefix.name("connection.acquisition.duration"), idPrefix.tags());
+            dnsResolutionDuration = newTimer(parent, idPrefix.name("dns.resolution.duration"), idPrefix.tags());
+            socketConnectDuration = newTimer(parent, idPrefix.name("socket.connect.duration"), idPrefix.tags());
             pendingAcquisitionDuration = newTimer(
-                    parent, idPrefix.name("pendingAcquisitionDuration"), idPrefix.tags());
+                    parent, idPrefix.name("pending.acquisition.duration"), idPrefix.tags());
 
             final String timeouts = idPrefix.name("timeouts");
             writeTimeouts = parent.counter(timeouts, idPrefix.tags("cause", "WriteTimeoutException"));
@@ -323,7 +323,7 @@ public final class RequestMetricSupport {
                 return actualRequests;
             }
 
-            final Counter counter = parent.counter(idPrefix.name("actualRequests"), idPrefix.tags());
+            final Counter counter = parent.counter(idPrefix.name("actual.requests"), idPrefix.tags());
             if (actualRequestsUpdater.compareAndSet(this, null, counter)) {
                 return counter;
             }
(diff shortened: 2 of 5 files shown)


core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -231,15 +231,15 @@ private RequestMetricSupport() {}
...
+                    parent, idPrefix.name("request.duration"), idPrefix.tags());
             requestLength = newDistributionSummary(
-                    parent, idPrefix.name("requestLength"), idPrefix.tags());
+                    parent, idPrefix.name("request.size"), idPrefix.tags());
             responseDuration = newTimer(
-                    parent, idPrefix.name("responseDuration"), idPrefix.tags());
+                    parent, idPrefix.name("response.duration"), idPrefix.tags());
             responseLength = newDistributionSummary(
-                    parent, idPrefix.name("responseLength"), idPrefix.tags());
+                    parent, idPrefix.name("response.size"), idPrefix.tags());
```
perlun: `response.length` vs `response.size` - we discussed this internally in our team and this was the suggested approach. If you feel the `response.length` form is better, please let me know; I am willing to change this part of the PR if required.

trustin:
I agree that we need to follow the Micrometer convention but we also need to make sure that the resulting Prometheus meter names exported by `PrometheusMeterRegistry` abide by the [Prometheus convention](https://prometheus.io/docs/practices/naming/), i.e. I'm happy as long as [`PrometheusMetricsIntegrationTest`](https://github.com/line/armeria/blob/master/thrift/src/test/java/com/linecorp/armeria/it/metric/PrometheusMetricsIntegrationTest.java) passes. What do you think, @perlun ?

trustin:
If we abide by the Micrometer convention, we do not need to apply the custom naming convention in `PrometheusRegistries.configureRegistry()` and `DropwizardRegistries.configureRegistry()`. Could you update these classes as well?

perlun:
> I agree that we need to follow the Micrometer convention but we also need to make sure that the resulting Prometheus meter names exported by `PrometheusMeterRegistry` abide by the [Prometheus convention](https://prometheus.io/docs/practices/naming/), i.e. I'm happy as long as [`PrometheusMetricsIntegrationTest`](https://github.com/line/armeria/blob/master/thrift/src/test/java/com/linecorp/armeria/it/metric/PrometheusMetricsIntegrationTest.java) passes. What do you think, @perlun ?

Makes sense - this was because of the `request.size` vs `request.length`. I reverted that part and the Prometheus integration test now seems to be passing locally for me.

> If we abide by the Micrometer convention, we do not need to apply the custom naming convention in PrometheusRegistries.configureRegistry() and DropwizardRegistries.configureRegistry(). Could you update these classes as well?

Sure. I changed this to use the plain `identity()` method instead and updated the tests accordingly. Given this change, we might be able to drop the `MoreNamingConventions` class altogether - I haven't done this yet, but let me know if you feel it's the right way to go.

anuraaga:
Just to confirm, and it might help to add to the PR description, does the integration tests passing mean we keep the same exported metrics for Promethean and dropwizard? I'm a bit surprised that going from camel case to dots would be the same but if so that's great 😊

core/src/main/java/com/linecorp/armeria/common/metric/MoreNamingConventions.java
```diff
@@ -18,53 +18,29 @@
...
 import io.micrometer.core.instrument.config.NamingConvention;
-import io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry;
-import io.micrometer.prometheus.PrometheusMeterRegistry;
-import io.micrometer.prometheus.PrometheusNamingConvention;
 
 /**
  * Provides commonly-used {@link NamingConvention}s.
  */
 public final class MoreNamingConventions {
+    private static final NamingConvention IDENTITY_NAMING_CONVENTION = ( name, type, baseUnit ) -> name;
```
trustin: How about:

- Reverting all the changes in this class
- Deprecating this class
- Remove any references to the methods in this class at all in other places

..? It's because we do not even need to touch `MeterRegistry.config().namingConvention()` if we are going to use the default naming convention.

perlun: Thanks for the suggestion. Revert the `MoreNamingConventions` and `MoreNamingConventionsTest` changes.

perlun:
@anuraaga 

> Just to confirm, and it might help to add to the PR description, does the integration tests passing mean we keep the same exported metrics for Prometheus and dropwizard? I'm a bit surprised that going from camel case to dots would be the same but if so that's great blush

Thanks. The Dropwizard names are slightly changed; the `request.duration` etc. form goes all the way through there.

The Prometheus names: I'm not sure these are unchanged or not. :expressionless: The `PrometheusMetricsIntegrationTest` passes both _with_ the `prometheus()` naming convention being configured and without it... So I guess there is some tests that should be added here to ensure that we are not breaking things (at least not without knowing it). If you know how it should be done, I think a separate PR from you adding these tests would be highly appreciated since I feel pretty deep in the water already... :slightly_smiling_face:

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
trustin: I'm not sure which is better to be honest. Perhaps we need to do some quick research on what others use.

core/src/main/java/com/linecorp/armeria/common/metric/MoreNamingConventions.java
```diff
@@ -37,6 +37,7 @@
 /**
  * Provides commonly-used {@link NamingConvention}s.
  */
+@Deprecated
```
trustin: How about also adding `@deprecated Use Micrometer's default {@link NamingConvention}.` in the Javadoc?

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
trustin: Uh, never mind. I realized you already reverted back to `length`.

anuraaga:
I think the prometheus integration tests themselves are fine and it means that indeed, the micrometer default metric convention against `a.b.c` is the same as the current one that parses `lowerCamel` used by Armeria, which happens to also sanitize all foreign characters like `.` into `_`. So prometheus users shouldn't notice anything from this change for default metrics. If they have custom camel case metrics, though, those will be exported in a different way if we don't add the `prometheus()` naming convention. 

It looks like the Dropwizard will export differently for default metrics too.

I tend to be quite wary when changing metric names - they can cause large problems for unrelated teams. In practice, it's often very difficult to even simply migrate the graph of a old name to a new one and people accept the loss of historical data on metric name changes.

I'm wondering whether we should try our best to leave our current behavior for now with the new one as opt-in, and give users a couple of versions to migrate their dashboards or give feedback on whether it would be feasible for them. @trustin what do you think?

trustin:
@anuraaga How about defining a System property flag which allows a user from opting-in to the new naming convention, then? We could add one in `Flags` and modify `MoreNamingConventions` to return different naming conventions based on it.

I guess this is beyond the original proposal from @perlun, so I'm not sure he is interested in making this change, though. Shall I take over this pull request, @perlun, or will you take a stab on this?

anuraaga:
+1 to having a `Flags` for it :)

perlun:
@anuraaga 

> I tend to be quite wary when changing metric names - they can cause large problems for unrelated teams. In practice, it's often very difficult to even simply migrate the graph of a old name to a new one and people accept the loss of historical data on metric name changes.

Very valid concerns, I think it's good that you're thinking about this perspective.

@trustin 

> I guess this is beyond the original proposal from @perlun, so I'm not sure he is interested in making this change, though. Shall I take over this pull request, @perlun, or will you take a stab on this?

Yeah, I won't mind. The important part for us is to get the proper Micrometer-naming-compliant names in place, and we already did a `MeterFilter` that patches it locally for us. :wink: So it's not urgent, I was just trying to help fixing it upstream but I'm quite certain that you will do a lot better job at getting it incorporated in the way you see fit best into the overall architecture.

Feel free to use whatever parts of my PR that you like. :+1:

perlun:
@trustin Any progress on this?

trustin:
@perlun Unfortunately not. I'll leave a comment when I start work on it. Meanwhile, please feel free to keep touching.

trustin:
I didn't forget about this one. Hoping to make sure this come true before 1.0.

trustin:
Sorry for not picking this up yet. Good news is I will work on this soon. :bowing_man:

trustin:
Superseded by https://github.com/line/armeria/pull/2367

perlun:
Greatly appreciated @trustin, thanks! :bowing_man:

PR #1832, opened 2019-06-12
Remove exactPath, triePath, prefix and regex method in Route api

minwoox:
Motivation:
As @anuraaga suggested, we can hide those four methods because the customers do not use them and they can give more complexity.

Modifications:
- Remove `exactPath`, `triePath`, `prefix` and `regex` in `Route`.
- Add `RoutePathType` which represents the path which was specified when a `Route` is created.
- Add `pathType()` and `paths()` to `Route` so that we are able to get the paths which we were getting using those four methods.
- Fail fast when a user does not specify a prefix path with `CompositeService`.

Result:
- Fewer exposed APIs


core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpDocServicePlugin.java
```diff
@@ -174,41 +167,55 @@ private static void addMethodInfo(Map<Class<?>, Set<MethodInfo>> methodInfos,
...
 
         builder.availableMimeTypes(availableMimeTypes(route));
         return builder.build();
     }
 
-    private static String endpointPathMapping(Route route) {
-        final Optional<String> exactPath = route.exactPath();
-        if (exactPath.isPresent()) {
-            return EXACT + exactPath.get();
+    private static String normalizedPathParam(Route route) {
```
trustin: `normalizePathParams()` ? (no D, pluralize 'param')

core/src/main/java/com/linecorp/armeria/server/PathMapping.java
```diff
@@ -57,33 +57,27 @@
...
      */
-    Optional<String> exactPath();
+    RoutePathType pathType();
 
     /**
-     * Returns the prefix of this {@link PathMapping} if this is a {@link PrefixPathMapping} or a
-     * {@link PathMappingWithPrefix}, {@link Optional#empty()} otherwise.
+     * Returns the list of paths that this {@link Route} has. The paths are different according to the value
+     * of {@link #pathType()}. If the path type has a {@linkplain RoutePathType#hasTriePath() trie path},
+     * the first element is the path that represents the type and the second element is the trie path.
```
trustin: this method will return a two-element list whose first element is ... and the second element is ...

core/src/main/java/com/linecorp/armeria/server/PathParamMapping.java
```diff
@@ -41,7 +41,7 @@
  *   <li>A set of path parameters declared in the path pattern</li>
  * </ul>
  */
-final class DefaultPathMapping extends AbstractPathMapping {
+final class PathParamMapping extends AbstractPathMapping {
```
trustin: How about `ParameterizedPathMapping`?

core/src/main/java/com/linecorp/armeria/server/RoutePathType.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * The prefix path type. e.g, "/", "/foo/"
+     */
+    PREFIX(true),
+
+    /**
+     * The path which contains path parameters. e.g, "/:", "/foo/:/bar/:"
+     */
+    PATH_PARAM(true),
```
trustin: `PARAMETERIZED`? Please feel free to suggest a better name.

core/src/main/java/com/linecorp/armeria/server/RegexWithPrefixMapping.java
```diff
@@ -20,24 +20,29 @@
...
-import java.util.Optional;
+import java.util.List;
 import java.util.Set;
 
 import javax.annotation.Nullable;
 
-final class PathMappingWithPrefix extends AbstractPathMapping {
+import com.google.common.collect.ImmutableList;
+
+final class RegexWithPrefixMapping extends AbstractPathMapping {
```
trustin: `RegexPathMappingWithPrefix`?

core/src/main/java/com/linecorp/armeria/server/RoutePathType.java
minwoox: I can't think of any better name. `PARAMETERIZED` is fine for me. Thanks!

minwoox:
All fixed. PTAL 🙇

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpDocServicePlugin.java
```diff
@@ -174,41 +167,55 @@ private static void addMethodInfo(Map<Class<?>, Set<MethodInfo>> methodInfos,
...
-            return null;
-        }
-
-        final EndpointInfoBuilder builder = new EndpointInfoBuilder(hostnamePattern, endpointPathMapping);
-        if (endpointPathMapping.startsWith(REGEX) && route.prefix().isPresent()) {
-            // PathMappingWithPrefix
-            builder.regexPathPrefix(PREFIX + route.prefix().get());
+        final EndpointInfoBuilder builder;
+        final RoutePathType pathType = route.pathType();
+        final List<String> path = route.paths();
```
anuraaga: Perhaps `pathComponents`

core/src/main/java/com/linecorp/armeria/server/CatchAllPathMapping.java
```diff
@@ -18,17 +18,17 @@
...
+import java.util.List;
 import java.util.Set;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 
 final class CatchAllPathMapping extends AbstractPathMapping {
 
-    static final CatchAllPathMapping INSTANCE = new CatchAllPathMapping();
+    static final CatchAllPathMapping instance = new CatchAllPathMapping();
```
anuraaga: Just to check, is this intentional? I think we've so far used all-caps for `INSTANCE`.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpDocServicePlugin.java
minwoox: Sounds better. :-)

core/src/main/java/com/linecorp/armeria/server/CatchAllPathMapping.java
minwoox: Yeah, we have used that. But it's technically not a constant. It's not public, so I changed it without much worry. Throughout the whole our code, we use two different conventions for the static final references. I want to change them all, but some of them are public so I couldn't.

minwoox: Do you think we need to keep the variable using underscores with uppercase? 🤔

minwoox: Changed it back to `INSTANCE`. :-)

trustin:
Thanks, @minwoox !

minwoox:
Thanks for reviewing. :wink:

PR #1833, opened 2019-06-12
Hide `CacheControlBuilder` from the public API

trustin:
Motivation:

We moved `Client/ServerCacheControl*` into the `common` package in #1816.
This means we can make `protected` methods package-local.

Modifications:

- Make `CacheControlBuilder` package-local
- Make all `protected` methods package-local
- Rename `ArmeriaHttpUtil.parseCacheControl()` to `parseDirectives()`
  for future reuse
- Rename `ArmeriaHttpUtil.parseCacheControlSeconds()` to
  `parseDirectiveValueAsSeconds()` for future reuse.
- Remove type parameter from `CacheControlBuilder` for cleaner Javadoc
  generation

Result:

- Leaner public API which hides the stuff which was unnecessarily
  visible from users.


minwoox:
LGTM once more!

PR #1840, opened 2019-06-14
Add brave module which supersedes zipkin module

### LINKED ISSUE #1792: Add HttpTracingClient/ServiceBuilder
A user currently has to build `Tracing` by him/herself and set `currentTraceContext` manually. We could add `HttpTracingClient/ServiceBuilder` which builds a `Tracing` for a user:

```java
HttpTracingService.builder()
                  .localServiceName("myService")
                  .spanReporter(...)
                  .build();

// Alternatively
HttpTracingService.builder()
                  .tracing(Tracing.newBuilder()...build())
                  .build();
```

minwoox:
Motivation:
As @anuraaga comments, it's not Zipkin, but Brave we're interfacing.

Modifications:
- Move `internal.tracing` package in `zipkin` module to `internal.brave` in `brave` module.
- Deprecation
  - `armeria-brave` supersedes `armeria-zipkin`.
  - `BraveClient` and `BraveService` supersedes `TracingHttpClient` and `TracingHttpService`
  - `RequestContextCurrentTraceContext` in `brave` supersedes the same class in `zipkin` 
- Fix a bug where `decorateScope` is not called.
- Miscellaneous
  - Add missing `equals` in `DefaultAggregatedHttpRequest`
  - Update Brave from 5.6.5 -> 5.6.6

Result:
- `armeria-brave` supersedes `armeria-zipkin`.
- A user can now decorate a scope when it's created.
- Close #1792 anyway.
- Brave 5.6.5 -> 5.6.6


zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -55,32 +55,54 @@
  * <p>This decorator puts trace data into HTTP headers. The specifications of header names and its values
  * correspond to <a href="http://zipkin.io/">Zipkin</a>.
  */
-public class HttpTracingClient extends SimpleDecoratingClient<HttpRequest, HttpResponse> {
+public final class HttpTracingClient extends SimpleDecoratingClient<HttpRequest, HttpResponse> {
```
minwoox: Should I revert this?

trustin: I think it's fine.

zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -55,32 +55,54 @@
...
+     * Returns a new builder.
+     */
+    public static HttpTracingClientBuilder builder() {
+        return new HttpTracingClientBuilder();
+    }
+
+    /**
+     * Creates a new tracing {@link Client} decorator using the specified {@link Tracing} instance.
+     */
+    public static Function<Client<HttpRequest, HttpResponse>, HttpTracingClient> decorator(Tracing tracing) {
```
trustin: IIRC we did not decide to use `decorator()` instead of `newDecorator()` yet.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/AbstractTracingBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+        return (B) this;
+    }
+
+    /**
+     * Sets the {@link Tracing}.
+     *
+     * @throws IllegalStateException if other setters for building a new {@link Tracing} are already invoked
+     */
+    public B tracing(Tracing tracing) {
+        checkState(tracingBuilder == null, "tracingBuilder is already set.");
```
trustin: Could we rephrase the exception message from user's standpoint? I don't think a user will know what this really means.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/AbstractTracingBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Sets the {@link CurrentTraceContext}. If this is not set and {@link #tracing(Tracing)} is not invoked,
+     * {@link RequestContextCurrentTraceContext} will be set by default.
+     *
+     * @throws IllegalStateException if {@link #tracing(Tracing)} is already set
+     *
+     * @see Builder#currentTraceContext(CurrentTraceContext)
+     */
+    public B currentTraceContext(CurrentTraceContext currentTraceContext) {
```
trustin: Don't we need to hide this method from users because otherwise a user can specify other implementation than `RequestContextCurrentTraceContext`? We could add a delegated `addScopeDecorator()` instead.

zipkin/src/main/java/com/linecorp/armeria/server/tracing/TracingService.java
```diff
@@ -0,0 +1,113 @@
...
+     * Returns a new builder.
+     */
+    public static TracingServiceBuilder builder() {
+        return new TracingServiceBuilder();
+    }
+
+    /**
+     * Creates a new tracing {@link Service} decorator using the specified {@link Tracing} instance.
+     */
+    public static Function<Service<HttpRequest, HttpResponse>, TracingService> decorator(Tracing tracing) {
```
trustin: `newDecorator()` at the moment.

zipkin/src/main/java/com/linecorp/armeria/server/tracing/TracingService.java
```diff
@@ -0,0 +1,113 @@
...
+        ensureScopeUsesRequestContext(tracing);
```
trustin: How about adding `requireNonNull()` on `tracing` in `ensureScopeUsesRequestContext()`?

zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -55,32 +55,54 @@
...
  */
-public class HttpTracingClient extends SimpleDecoratingClient<HttpRequest, HttpResponse> {
+public final class HttpTracingClient extends SimpleDecoratingClient<HttpRequest, HttpResponse> {
 
     private static final Logger logger = LoggerFactory.getLogger(HttpTracingClient.class);
 
+    /**
+     * Returns a new builder.
+     */
+    public static HttpTracingClientBuilder builder() {
```
trustin: Shouldn't we make the same naming change for the client side?

zipkin/src/main/java/com/linecorp/armeria/common/tracing/AbstractTracingBuilder.java
```diff
@@ -56,10 +60,14 @@ private B self() {
...
      * Sets the {@link Tracing}.
      *
-     * @throws IllegalStateException if other setters for building a new {@link Tracing} are already invoked
+     * @throws IllegalStateException if other setters(e.g., localServiceName, spanReporter, etc.) for
+     *                               building a new {@link Tracing} are already invoked
      */
     public B tracing(Tracing tracing) {
-        checkState(tracingBuilder == null, "tracingBuilder is already set.");
+        if (tracingBuilder != null) {
+            throw new IllegalStateException("cannot set tracing after other setters(e.g., localServiceName, " +
```
trustin: nit: space between `setters` and `(`

minwoox: Fixed! 😱

minwoox:
@anuraaga I agree with you on that. Probably it's better not to have these builders because there's no way to share the same `Tracing`.

zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -54,32 +54,44 @@
  *
  * <p>This decorator puts trace data into HTTP headers. The specifications of header names and its values
  * correspond to <a href="http://zipkin.io/">Zipkin</a>.
+ *
+ * @deprecated Use {@link TracingClient}.
  */
-public class HttpTracingClient extends SimpleDecoratingClient<HttpRequest, HttpResponse> {
+@Deprecated
+public final class HttpTracingClient extends SimpleDecoratingClient<HttpRequest, HttpResponse> {
```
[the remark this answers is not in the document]
minwoox: That was my first thought, but I chose this way because it's easier to remove and it will make removing deprecated methods and class simpler.

core/src/main/java/com/linecorp/armeria/common/DefaultAggregatedHttpRequest.java
```diff
@@ -51,4 +54,43 @@ public String scheme() {
...
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (!(obj instanceof DefaultAggregatedHttpRequest)) {
```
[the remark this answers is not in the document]
minwoox: You mean with `AggregatedHttpRequest`? yeah that makes sense. Thanks :-)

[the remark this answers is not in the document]
anuraaga: Ah yeah that's what I meant :)

minwoox:
So let me remove the builders and leave the bug fix and renaming things 😉 Thanks!

adriancole:
I agree: multiple tracing instances is usually a problem and we wouldnt
want to guide folks towards this as it unleashes all sorts of questions
such as whose threadlocal is this anyway? :)

ikhoon:
@minwoox `codecov/patch` shows the coverage of diff(63.2%) is lower than the target one.

zipkin/src/main/java/com/linecorp/armeria/client/tracing/TracingClient.java
```diff
@@ -0,0 +1,169 @@
...
+ * correspond to <a href="http://zipkin.io/">Zipkin</a>.
+ */
+public final class TracingClient extends SimpleDecoratingClient<HttpRequest, HttpResponse> {
+
+    private static final Logger logger = LoggerFactory.getLogger(TracingClient.class);
+
+    /**
+     * Creates a new tracing {@link Client} decorator using the specified {@link Tracing} instance.
+     */
+    public static Function<Client<HttpRequest, HttpResponse>, TracingClient> newDecorator(Tracing tracing) {
```
adriancole: note we still have an issue to accept `HttpTracing`.. even if we aren't implementing anything with that, it might make sense to use this as an arg to the decorator (internally calling httpTracing.tracing()) rather than go back and overload/deprecate later

zipkin/src/main/java/com/linecorp/armeria/client/tracing/TracingClient.java
```diff
@@ -0,0 +1,169 @@
...
+        return newDecorator(tracing, null);
+    }
+
+    /**
+     * Creates a new tracing {@link Client} decorator using the specified {@link Tracing} instance
+     * and the remote service name.
+     */
+    public static Function<Client<HttpRequest, HttpResponse>, TracingClient> newDecorator(
+            Tracing tracing, @Nullable String remoteServiceName) {
```
adriancole: if this had HttpTracing, the removeServiceName arg would be redundant and unnecessary as that's a part of HttpTracing

trustin:
How about moving to `tracing.zipkin` (or just to `zipkin`) and renaming to `ZipkinClient/Service` by taking advantage of this breaking change, then? We may need to add `tracing.opentracing` or other vendor-specific implementations later.

adriancole:
I think opentracing is dead now, so probably if something later it would be opentelemetry which is the latest brand name.

Zipkin could work, or more specifically brave as brave supports backends besides zipkin (eg stackdriver and aws X-Ray)

minwoox:
Thanks @trustin and @adriancole for the idea.
> I think opentracing is dead now, so probably if something later it would be opentelemetry which is the latest brand name.

> Zipkin could work, or more specifically brave as brave supports backends besides zipkin (eg stackdriver and aws X-Ray)
So do you recommend moving to `tracing.brave` and name them to `ZipkinClient/Service`? Right?

adriancole:
> So do you recommend moving to tracing.brave and name them to ZipkinClient/Service? Right?

In brave, we don't double-brand. For example, we just use the package name to indicate it is brave related. The entrypoint types are typically named TracingXXX where XXX is what's being decorated.

```java
package brave.servlet;

--snip--

public final class TracingFilter implements Filter { 
```

The fact you might have a couple things with prefix Tracing shouldn't really matter as it would be a mistake to use two at the same time almost always. The double-instrumentation problem is something you will more routinely encounter once you decide to change to make different plugins possible. I think it will cost a lot of work, honestly, as many things don't have the features brave has especially relating to context propagation decoration.

minwoox:
@adriancole Thanks for the comment. So you think it's not a good idea that uses `brave` in the package and `Zipkin` in the class name together, right?
Then we should use the original names which are `TracingClient/Service`? and the package will be `com.linecorp.armeria.client.tracing.brave`? or `com.linecorp.armeria.client.brave`? which `tracing` is removed?

adriancole:
@minwoox if your goal is to exchange components I would put the "brave" part higher up in the type hierarchy as maybe you recall we have some shared components between client and server. It kindof depends on what your normal packaging model is for 3rd party support..

ex. `com.linecorp.armeria.brave.client` vs `com.linecorp.armeria.client.brave`

It seems I suppose ok to keep discussing this on this issue, though it is a much more huge change than what's being done here.

anuraaga:
Not a strong feeling but my opinion is it's ok to duplicate the name of a concept in both package and class in Armeria since package is generally pointless to users, it's automatically handled by the IDE, but does help with maintainers that are abstracting many different implementations, which is what Armeria does, by having separate folders. If we took no doublespeak too literally, we would rename `server.ServerBuilder` to `server.Builder` which seems obviously bad.

But I did realize that it does seem to make sense to rename all of our uses of `zipkin` with `brave` since that is what we are interfacing with so is more precise.

minwoox:
> It seems I suppose ok to keep discussing this on this issue, though it is a much more huge change than what's being done here.

Yeah, I think it's ok. It happens all the time. 🤣 

>  I would put the "brave" part higher up in the type hierarchy...

I really like your idea that put it higher up. I just ran my eyes through the other packages in Zipkin and there are not many things to share when we introduce vendor-specific implementations.
However, that made me think that perhaps we need to put `Zipkin` or `Brave` in the class name otherwise they would have the same class names when we introduce a new one:
- `com.linecorp.armeria.brave.client.TracingClient`
- `com.linecorp.armeria.minwooxvendor.client.TracingClient`

As @anuraaga said, we do have duplicate names and it's okay to duplicate. Some of the languages such as Golang highly recommend not to use the duplicated name because they use [the package and its contents' name together](https://blog.golang.org/package-names) in their code. But we don't do that and it might reduce the chances that a user got confused. 😄

minwoox:
> But I did realize that it does seem to make sense to rename all of our uses of zipkin with brave since that is what we are interfacing with so is more precise.

@anuraaga So we should change the artifact ID to `armeria-brave`?

anuraaga:
Technically yes it means changing to `armeria-brave`. ') But it's so breaking that I wouldn't do it without keeping `armeria-zipkin` as an alias for a couple of versions.

I hadn't thought of a more collective `armeria-tracing` though. My personal preference is to keep Armeria's focus on micrometer and brave and other implementations are out-of-tree, similar to gRPC-opencensus + out-of-tree (e.g. brave-grpc).

minwoox:
> Technically yes it means changing to armeria-brave. ') But it's so breaking that I wouldn't do it without keeping armeria-zipkin as an alias for a couple of versions.

Yes, definitely 😄 

> I hadn't thought of a more collective armeria-tracing though. My personal preference is to keep Armeria's focus on micrometer and brave and other implementations are out-of-tree, similar to gRPC-opencensus + out-of-tree (e.g. brave-grpc).

Agreed. 😄

minwoox:
@anuraaga Add brave module. PTAL 🙇

minwoox:
> @minwoox but codecov/patch shows the coverage of diff(63.2%) is lower than the target one.

@ikhoon Increased the coverage. 😉

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -0,0 +1,68 @@
...
+import com.linecorp.armeria.common.brave.RequestContextCurrentTraceContext;
+
+import brave.Tracing;
+import brave.http.HttpTracing;
+
+/**
+ * Decorates a {@link Client} to trace outbound {@link HttpRequest}s using
+ * <a href="https://github.com/openzipkin/brave">Brave</a>.
+ *
+ * <p>This decorator puts trace data into HTTP headers. The specifications of header names and its values
```
anuraaga: Think we can remove this, or otherwise make it more generic. I think the first version managed header keys in our code, but for a while we use `Tracing.injector` so the user can configure this to be anything, not necessarily the zipkin propagation.

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -0,0 +1,68 @@
...
+ * correspond to <a href="http://zipkin.io/">Zipkin</a>.
+ */
+public final class BraveClient extends HttpTracingClient {
+
+    private static final Logger logger = LoggerFactory.getLogger(BraveClient.class);
+
+    /**
+     * Creates a new tracing {@link Client} decorator using the specified {@link Tracing} instance.
```
anuraaga: `{@link HttpTracing}`

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -0,0 +1,68 @@
...
+import brave.http.HttpTracing;
+
+/**
+ * Decorates a {@link Client} to trace outbound {@link HttpRequest}s using
+ * <a href="https://github.com/openzipkin/brave">Brave</a>.
+ *
+ * <p>This decorator puts trace data into HTTP headers. The specifications of header names and its values
+ * correspond to <a href="http://zipkin.io/">Zipkin</a>.
+ */
+public final class BraveClient extends HttpTracingClient {
```
anuraaga: I think we might have talked a bit about this, but now even moreso, it seems like `HttpTracingClient` should extend `BraveClient`. Even for just a deprecation period, it seems weird for the `brave` artifact to have to depend on a deprecated `zipkin` artifact.

brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * Use this singleton when building a {@link Tracing} instance for use with
+     * {@link BraveService} or {@link BraveClient}.
+     *
+     * <p>If you need to customize the context, use {@link #builder()} instead.
+     *
+     * @see Tracing.Builder#currentTraceContext(CurrentTraceContext)
+     */
+    public static final CurrentTraceContext DEFAULT = new RequestContextCurrentTraceContext(new Builder());
```
anuraaga: Should we go ahead and add a method `default()` instead of making the instance itself public?

brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
```diff
@@ -0,0 +1,311 @@
...
+    /**
+     * Use this to ensure the trace context propagates to children.
+     *
+     * <p>Ex.
+     * <pre>{@code
+     *  // Ensure the trace context propagates to children
+     * ctx.onChild(RequestContextCurrentTraceContext::copy);
+     * }</pre>
+     */
+    public static void copy(RequestContext src, RequestContext dst) {
```
anuraaga: Should a user ever need to invoke this?

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
```diff
@@ -0,0 +1,57 @@
...
+import com.linecorp.armeria.server.tracing.HttpTracingService;
+
+import brave.Tracing;
+import brave.http.HttpTracing;
+
+/**
+ * Decorates a {@link Service} to trace inbound {@link HttpRequest}s using
+ * <a href="https://github.com/openzipkin/brave">Brave</a>.
+ *
+ * <p>This decorator retrieves trace data from HTTP headers. The specifications of header names and its values
```
anuraaga: Ditto

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -0,0 +1,68 @@
...
+                        e.getMessage());
+        }
+        return delegate -> new BraveClient(delegate, httpTracing);
+    }
+
+    /**
+     * Creates a new instance.
+     */
+    private BraveClient(Client<HttpRequest, HttpResponse> delegate, HttpTracing httpTracing) {
+        super(delegate, httpTracing.tracing(), httpTracing.serverName(),
```
anuraaga: Since we aren't migrating to use the parser and other fields of `HttpTracing`, and it'd be a pretty big change, I think we should stick to accepting `Tracing` for now like before. I think even long-term, it's reasonable to expose an API that takes `Tracing` and one that takes `HttpTracing` later.

If we did stick with accepting `HttpTracing`, we need to add to the javadoc that we currently do not respect all the fields in it

brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
minwoox: This is used in `BraveClient` and `BraveService` so that this should be public. Do you want me to this into the internal package?

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
minwoox: Yes it looks very weird that the `brave` artifact depends on `zipkin` artifact. :-)
But, I did it to reduce duplicate code in `BraveClient` and `BraveService`.
A user used to use three public classes which are `HttpTracingClient`, `HttpTracingService` and `RequestContextCurrentTraceContext`.
So I added `BraveClient`, `BraveService` and `RequestContextCurrentTraceContext` (different package) which correspond to them.

If I want to make `BraveClient` to extend `HttpTracingClient` or vice versa, I had to add this creepy constructor:
```
protected HttpTracingService(Service<HttpRequest, HttpResponse> delegate, Tracing tracing,
                             BiConsumer<RequestContext, RequestContext> traceContextPropagator) {...}
```
because `RequestContextCurrentTraceContext.copy()` method is used in the `execute()`.
I didn't want to add this constuctor to `brave` module, so I chose to add it to `zipkin`. That's the reason.

Perhaps I just copy the code from `HttpTracingClient` and `HttpTracingService` to the new module like I did in the first place.
What do you think?

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
minwoox: I agree. :-) Let me remove the `HttpTracing` for now. Will just leave a `TODO` comment for that. 😉

brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
minwoox: Sounds good. Let me do that. :-)

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
anuraaga: Ah thanks for the details, I understand the issue now. I might have a slight preference to copying, and we just delete the zipkin folder once deprecation is done, but am ok with this too.

brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
anuraaga: Ah then yeah think it should be in internal package.

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
minwoox: I prefer copying too. :-) Let me do that. 😄 Thanks!

brave/build.gradle
```diff
@@ -0,0 +1,5 @@
+dependencies {
+    testCompile project(':thrift')
+
+    compile 'io.zipkin.brave:brave'
```
trustin: nit: `compile` could come before `testCompile`.

brave/src/main/java/com/linecorp/armeria/client/brave/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Distributed tracing clients based on <a href="https://github.com/openzipkin/brave">Brave</a>,
+ * a Java tracing library compatible with <a href="http://zipkin.io/">Zipkin</a>.
```
trustin: Add/remove this to/from all `package-info.java` files for consistency.

minwoox:
@anuraaga Fixed. PTAL 🙇

core/src/main/java/com/linecorp/armeria/common/DefaultRpcRequest.java
```diff
@@ -120,11 +120,11 @@ public boolean equals(Object obj) {
             return true;
         }
 
-        if (!(obj instanceof DefaultRpcRequest)) {
+        if (!(obj instanceof RpcRequest)) {
             return false;
         }
 
-        final DefaultRpcRequest that = (DefaultRpcRequest) obj;
+        final RpcRequest that = (RpcRequest) obj;
```
trustin: I think we should check the exact type here because a user may implement a new subtype of `RpcRequest` which has an additional property. In that case, `equals()` will not be supposed to return `true`.

core/src/main/java/com/linecorp/armeria/common/sse/DefaultServerSentEvent.java
```diff
@@ -96,16 +96,16 @@ public boolean equals(Object obj) {
             return true;
         }
 
-        if (!(obj instanceof DefaultServerSentEvent)) {
+        if (!(obj instanceof ServerSentEvent)) {
```
trustin: Ditto

zipkin/build.gradle
```diff
@@ -1,6 +1,5 @@
 dependencies {
     testCompile project(':thrift')
 
-    // Zipkin
-    compile 'io.zipkin.brave:brave'
+    compile project(':brave')
```
trustin: nit: `compile` should come first.

zipkin/src/main/java/com/linecorp/armeria/common/tracing/RequestContextCurrentTraceContext.java
```diff
@@ -44,14 +46,17 @@
  * <p>This {@link CurrentTraceContext} stores/loads the trace context into/from a
  * {@link RequestContext}'s attribute so that there's no need for thread local variables
  * which can lead to unpredictable behavior in asynchronous programming.
+ *
+ * @deprecated Use the same class in the `armeria-brave` dependency.
```
trustin: the same class -> `{@link ...}` with FQCN

zipkin/src/main/java/com/linecorp/armeria/client/tracing/HttpTracingClient.java
```diff
@@ -54,25 +56,33 @@
  *
  * <p>This decorator puts trace data into HTTP headers. The specifications of header names and its values
  * correspond to <a href="http://zipkin.io/">Zipkin</a>.
+ *
+ * @deprecated Use {@code BraveClient} in the `armeria-brave` dependency.
```
trustin: We can now use `{@link ..}` everywhere that `:zipkin` depends on `:brave`.

minwoox:
@trustin All fixed. PTAL 🙇

minwoox:
Thanks for reviewing. 😉

PR #1841, opened 2019-06-17
Export blocking task executor metrics

### LINKED ISSUE #1683: Export the metrics about `blockingTaskExecutor`.
We could expose the metrics about `ServerConfig.blockingTaskExecutor()` stats to Micrometer. Could use: https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java

imasahiro:
Closes #1683

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
@@ -39,6 +39,9 @@ import com.linecorp.armeria.internal.ConnectionLimitingHandler;
 import com.linecorp.armeria.server.logging.AccessLogWriter;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics;
+import io.micrometer.core.instrument.internal.TimedExecutor;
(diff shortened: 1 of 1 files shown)


trustin:
I'm curious what would happen if a user stops a server and then restarts it? Would be meters double-registered?

imasahiro:
`MeterRegistry.registerMeterIfNecessary` is called internally when binding this metrics, the metrics is registered at once.

imasahiro:
But hm.., I realized that if we have a multiple `Server` instance, multiple `Server` instances share this metrics. Should we avoid this?  Or it would be acceptable to user to face such kind of situation?

trustin:
> But hm.., I realized that if we have a multiple `Server` instance, multiple `Server` instances share this metrics. Should we avoid this?  Or it would be acceptable to user to face such kind of situation?

In that case, a user will have to specify a different `MeterRegistry` or we need to allow specifying a different meter ID prefix than `armeria.` So, I think there's nothing much we can do in this pull request. We'll have to fix it separately.

trustin:
Thanks, @imasahiro!

PR #1845, opened 2019-06-17
Provide MeterIdPrefixFunction for Retrofit

### LINKED ISSUE #1591: Add method/class name to metric for Retrofit
When using retrofit, a user may want to get class and method name in metrics. 
(I guess RpcRequest/Response already shows class/method name but HttpRequest/Response doesn't )

kojilin:
Motivation:
Currently user can't add tag with retrofit interface and method name.
Make ArmeriaCallFactory set Invocation to RequestLog for metrics and monitoring.

Modifications:
- Add `InvocationAttribute` to let user can fetch Invocation.
- Add default implementation of `MeterIdPrefixFunction` for retrofit. Which has the same tags as `MeterIdPrefixFunction.ofDefault`.
  - `RetrofitMeterIdPrefixFunctionBuilder` provides `ofType` and `withServiceTag` to use with `com.linecorp.armeria.spring.MeterIdPrefixFunctionFactory`.

Result:
- Close #1591


retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/InvocationAttribute.java
```diff
@@ -0,0 +1,57 @@
...
+        if (invocation == null) {
+            return;
+        }
+        log.attr(RETROFIT_INVOCATION).set(invocation);
+    }
+
+    /**
+     * Get {@code Invocation} from {@code RequestLog}.
+     */
+    public static Optional<Invocation> getInvocation(RequestLog log) {
```
trustin: How about just returning `Invocation` and annotating it with `@Nullable`?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/metric/RetrofitMeterIdPrefixFunctionBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+     * Creates a {@link RetrofitMeterIdPrefixFunctionBuilder} with {@code name}.
+     */
+    public static RetrofitMeterIdPrefixFunctionBuilder ofName(String name) {
+        return new RetrofitMeterIdPrefixFunctionBuilder(requireNonNull(name, "name"));
+    }
+
+    /**
+     * Creates a {@link RetrofitMeterIdPrefixFunctionBuilder} with name starts with "armeria.".
+     */
+    public static RetrofitMeterIdPrefixFunctionBuilder ofType(String type) {
```
trustin: How about:

- merging `ofName()` and `ofType()` to `builder(String... names)`
- and moving it to `RetrofitMeterIdPrefixFunction`?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/metric/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Metric utilities for <a href="https://square.github.io/retrofit/">Retrofit2</a> adapter for armeria.
+ */
+@NonNullByDefault
+package com.linecorp.armeria.client.retrofit2.metric;
```
trustin: Would it be worth adding a dedicated package for these classes?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/InvocationAttribute.java
```diff
@@ -0,0 +1,57 @@
...
+import io.netty.util.AttributeKey;
+import retrofit2.Invocation;
+
+/**
+ * Armeria Retrofit automatically adds an invocation to {@link RequestLog}. You can retrieve
+ * the invocation by {@link InvocationAttribute#getInvocation(RequestLog)} for metrics and monitoring.
+ *
+ * @see Invocation
+ */
+public final class InvocationAttribute {
```
trustin: I'm not sure I like the name, but I have a better idea either. Any other suggestions?

kojilin: Maybe `InvocationUtil` or move to other existing class ?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/InvocationUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.common.logging.RequestLog;
+
+import io.netty.util.AttributeKey;
+import retrofit2.Invocation;
+
+/**
+ * Armeria Retrofit automatically adds an invocation to {@link RequestLog}. You can retrieve
+ * the invocation by {@link InvocationUtil#getInvocation(RequestLog)} for metrics and monitoring.
```
trustin: How about rephrasing this:

```
Retrieves a Retrofit {@link Invocation} associated with a {@link RequestLog}.
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/InvocationUtil.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+        log.attr(RETROFIT_INVOCATION).set(invocation);
+    }
+
+    /**
+     * Get {@code Invocation} from {@code RequestLog}.
+     */
+    @Nullable
+    public static Invocation getInvocation(RequestLog log) {
+        return log.attr(RETROFIT_INVOCATION).get();
```
trustin: How about avoiding unnecessary object allocation?

```
if (log.hasAttr(RETROFIT_INVOCATION)) {
    return log.attr(RETROFIT_INVOCATION).get();
} else {
    return null;
}
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,137 @@
...
+
+        private RetrofitMeterIdPrefixFunctionBuilder(String name) {
+            this.name = name;
+        }
+
+        /**
+         * Make tag of {@link RetrofitMeterIdPrefixFunction} contains Retrofit service interface name.
+         */
+        public RetrofitMeterIdPrefixFunctionBuilder withServiceTag(String tagName, String defaultServiceName) {
+            this.tagName = tagName;
```
minwoox: nit: `requireNonNull(tagName, "tagName");`?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,137 @@
...
+            this.defaultServiceName = defaultServiceName;
```
minwoox: nit: `requireNonNull(defaultServiceName, "defaultServiceName")`?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,137 @@
...
+        // Add the 'httpStatus' tag.
+        final HttpStatus status;
+        if (log.isAvailable(RequestLogAvailability.RESPONSE_HEADERS)) {
+            status = log.status();
+        } else {
+            status = HttpStatus.UNKNOWN;
+        }
+        tags.add(Tag.of("httpStatus", status.codeAsText()));
+
+        return new MeterIdPrefix(name, tags);
```
minwoox: Not a strong opinion, but how about extracting this method to other class like `MeterIdPrefixUtil` so we can remove the duplicate code in `MeterIdPrefixFunction`?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
minwoox: How about renaming to `serviceTagName`?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/InvocationUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import io.netty.util.AttributeKey;
+import retrofit2.Invocation;
+
+/**
+ * Armeria Retrofit automatically adds an invocation to {@link RequestLog}. You can retrieve
+ * the invocation by {@link InvocationUtil#getInvocation(RequestLog)} for metrics and monitoring.
+ *
+ * @see Invocation
+ */
+public final class InvocationUtil {
```
minwoox: > I'm not sure I like the name, but I have a better idea either. Any other suggestions?

How about removing this class and moving the logic into the existing classes?
We can place `RETROFIT_INVOCATION` into `ArmeriaCallFactory` I think. `getInvocation` is used only once and `setInvocation` is used twice but we can just use it only once by removing the duplicate code. If we do that, we don't even have to use try with resources when the `invocation` is `null`.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public MeterIdPrefix activeRequestPrefix(MeterRegistry registry, RequestLog log) {
+        return new MeterIdPrefix(name, buildTags(log));
+    }
+
+    @Override
+    public MeterIdPrefix apply(MeterRegistry registry, RequestLog log) {
+        final List<Tag> tags = buildTags(log);
```
minwoox: nit: `requireNonNull(log, "log")`

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
kojilin: I try to extract as static method, curious do we prefer it as default method?

minwoox: Thanks! I think the static method is fine. 😄

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -71,26 +71,12 @@ static MeterIdPrefixFunction ofDefault(String name) {
...
-                tags.add(Tag.of("httpStatus", status.codeAsText()));
-
-                return new MeterIdPrefix(name, tags);
+                final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(4);
+                buildTags(tagListBuilder, log);
+                return new MeterIdPrefix(name, tagListBuilder.build());
             }
 
-            private List<Tag> buildTags(RequestLog log) {
+            private void buildTags(ImmutableList.Builder<Tag> tagListBuilder, RequestLog log) {
```
trustin: How about moving `buildTags()` after `apply()` because it's used by both `activeRequestPrefix()` and `apply()`?

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -104,19 +90,42 @@ public MeterIdPrefix apply(MeterRegistry registry, RequestLog log) {
...
+                final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(4);
+                buildTags(tagListBuilder, log);
+                appendHttpStatusTag(tagListBuilder, log);
+                return new MeterIdPrefix(name, tagListBuilder.build());
             }
         };
     }
 
+    /**
+     * Append {@link HttpStatus} to {@link Tag}.
```
trustin: `Appends`

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -104,19 +90,42 @@ public MeterIdPrefix apply(MeterRegistry registry, RequestLog log) {
...
+     */
+    static void appendHttpStatusTag(ImmutableList.Builder<Tag> tagListBuilder, RequestLog log) {
```
trustin: Could you move this to an internal utility class like `armeria.internal.metric.RequestMetricSupport`? It exposes Guava type in the public API.

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -129,7 +138,7 @@ public MeterIdPrefix apply(MeterRegistry registry, RequestLog log) {
      * a {@link MeterIdPrefix} using response properties that's not always available when the active request
      * counter is increased, such as HTTP status.
      */
-    default MeterIdPrefix activeRequestPrefix(MeterRegistry registry,  RequestLog log) {
+    default MeterIdPrefix activeRequestPrefix(MeterRegistry registry, RequestLog log) {
```
trustin: Thanks!

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactory.java
```diff
@@ -48,12 +53,28 @@
...
 final class ArmeriaCallFactory implements Factory {
 
+    private static final AttributeKey<Invocation> RETROFIT_INVOCATION =
+            AttributeKey.valueOf(ArmeriaCallFactory.class, "RETROFIT_INVOCATION");
+
+    /**
+     * Retrieves a Retrofit {@link Invocation} associated with a {@link RequestLog}.
+     */
+    @Nullable
+    public static Invocation getInvocation(RequestLog log) {
```
trustin: I think a user will not be able to access `Invocation` when he or she wants to write his or her own `MeterIdPrefixFunction`. How about reviving `InvocationUtil`?

minwoox: Couldn't think of that the user would use. Sorry about the mess. 😢

kojilin: I forget ArmeriaCallFactory is not public..0rz..

PR #1846, opened 2019-06-18
Make health check client retry interval more configurable to ba…

imasahiro:
…healthcheck request among servers

Motivation:
- About HealthCheckedEndpointGroup, it causes a little bit heavy load when 1. about 1000 clients send a healthcheck request to a server at the same time and 2. when a backend server is not warmed up yet and not JIT-compiled well yet.

Changes
- Introduce `Backoff` to HealthCheckedEndpointGroup to give a chance to balance the request among HealthCheckedEndpointGroup users

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
@@ -42,7 +42,7 @@ import io.micrometer.core.instrument.binder.MeterBinder;
  * An {@link EndpointGroup} decorator that only provides healthy {@link Endpoint}s.
  */
 public abstract class HealthCheckedEndpointGroup extends DynamicEndpointGroup {
-    static final Backoff DEFAULT_HEALTHCHECK_RETRY_BACKOFF = Backoff.fixed(3000).withJitter(0.1);
+    static final Backoff DEFAULT_HEALTHCHECK_RETRY_BACKOFF = Backoff.fixed(3000).withJitter(0.2);
 
     private final ClientFactory clientFactory;
     private final EndpointGroup delegate;
@@ -61,7 +61,8 @@ public abstract class HealthCheckedEndpointGroup extends DynamicEndpointGroup {
                                          EndpointGroup delegate,
                                          Duration retryInterval) {
         this(clientFactory, delegate,
-             Backoff.fixed(requireNonNull(retryInterval, "retryInterval").toMillis()));
+             Backoff.fixed(requireNonNull(retryInterval, "retryInterval").toMillis())
+                    .withJitter(0.2));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
```diff
@@ -41,24 +42,38 @@
  * An {@link EndpointGroup} decorator that only provides healthy {@link Endpoint}s.
  */
 public abstract class HealthCheckedEndpointGroup extends DynamicEndpointGroup {
-    static final Duration DEFAULT_HEALTHCHECK_RETRY_INTERVAL = Duration.ofSeconds(3);
+    static final Backoff DEFAULT_HEALTHCHECK_RETRY_BACKOFF = Backoff.fixed(3000).withJitter(0.1);
```
trustin: We use the default jitter rate of 0.2 for other retries. Could we use the same value?

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
```diff
@@ -41,24 +42,38 @@
...
      * A subclass being initialized with this constructor must call {@link #init()} before start being used.
+     *
+     * @deprecated Use {@link #HealthCheckedEndpointGroup(ClientFactory, EndpointGroup, Backoff)}.
      */
+    @Deprecated
     protected HealthCheckedEndpointGroup(ClientFactory clientFactory,
                                          EndpointGroup delegate,
                                          Duration retryInterval) {
+        this(clientFactory, delegate,
+             Backoff.fixed(requireNonNull(retryInterval, "retryInterval").toMillis()));
```
trustin: I'm curious if we should add jitter by default here.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
imasahiro: Updated

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupBuilder.java
```diff
@@ -76,12 +77,22 @@ public HttpHealthCheckedEndpointGroupBuilder protocol(SessionProtocol protocol)
...
      * Sets the interval between health check requests. Must be positive.
+     * @deprecated Use {@link #retryBackoff(Backoff)}.
      */
+    @Deprecated
     public HttpHealthCheckedEndpointGroupBuilder retryInterval(Duration retryInterval) {
         requireNonNull(retryInterval, "retryInterval");
         checkArgument(!retryInterval.isNegative() && !retryInterval.isZero(),
                       "retryInterval: %s (expected > 0)", retryInterval);
-        this.retryInterval = retryInterval;
+        retryBackoff = Backoff.fixed(retryInterval.toMillis());
```
minwoox: jitter here as well?

trustin:
Thanks, @imasahiro !

PR #1848, opened 2019-06-18
Make sure all safely retriable exceptions are wrapped by UnprocessedR…

### LINKED ISSUE #1653: Change to return UnprocessedRequestException in RetryingClient
When `RetryingClient` sets the timeout, it completes `HttpResponse` with `ResponseTimeoutException` if it passes the timeout deadline.
It makes sense if `RetryingClient` reaches that point after several retries, but it wouldn't make sense when it couldn't even connect to the endpoint and reached the timeout due to heavy jobs in decorator, etc.

ikhoon:
…equestException

Motivation:
`UnprocessedRequestException` means that the request has not been processed by the server.
If we wrap socket connect failures and more with UnprocessedRequestException,
user can safely retry the request without worrying about the idempotency of the request.

Modification:
* Add `GoAwayReceivedException` which raised when a server sent an HTTP/2 GOAWAY.
* Add `RefusedStreamException` which raised when a server set MAX_Current_STREAM to 0.
  see httpwg.org/specs/rfc7540.html#SETTINGS_MAX_CONCURRENT_STREAMS
* Remove singleton `INSTANCE` of `UnprocessedRequestException` and add a new constructor
  which accepts `Throwable` to make sure `UnprocessedRequestExecepion` always has a cause.
* Change `ResponseTimeoutException.get()` to
  `new UnprocessedRequestException(RequestTimeoutException.get())` in ConcurrencyLimitingClient.
* Change `UnprocessedRequestException.get()` to `NullPointerException`
  when sessionProtocol is null in `HttpClientDelegate#doExecute`.

Result:
* We can get more retriable exceptions.

Fixes #1653

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
@@ -22,10 +22,9 @@ import java.util.Map.Entry;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.RequestContextWrapper;
-import com.linecorp.armeria.server.ServiceRequestContext;
 
 /**
- * Wraps an existing {@link ServiceRequestContext}.
+ * Wraps an existing {@link ClientRequestContext}.
  */
 public class ClientRequestContextWrapper
         extends RequestContextWrapper<ClientRequestContext> implements ClientRequestContext {
--- a/core/src/main/java/com/linecorp/armeria/client/GoAwayReceivedException.java
+++ b/core/src/main/java/com/linecorp/armeria/client/GoAwayReceivedException.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.client;
+
+/**
+ * A {@link RuntimeException} raised when a server sent an
+ * <a href="https://httpwg.org/specs/rfc7540.html#GOAWAY">HTTP/2 GOAWAY frame</a> with
+ * the {@code lastStreamId} less then the stream ID of the request.
+ */
+public final class GoAwayReceivedException extends RuntimeException {
+
+    private static final long serialVersionUID = -7167601309699030853L;
+
+    private static final GoAwayReceivedException INSTANCE = new GoAwayReceivedException();
+
+    /**
+     * Returns a singleton {@link GoAwayReceivedException}.
+     */
+    public static GoAwayReceivedException get() {
+        return INSTANCE;
+    }
+
+    private GoAwayReceivedException() {
+        super(null, null, false, false);
+    }
+}
--- a/core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
@@ -41,6 +41,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.MoreObjects;
 
+import com.linecorp.armeria.common.ClosedSessionException;
 import com.linecorp.armeria.common.SessionProtocol;
 
 import io.netty.bootstrap.Bootstrap;
@@ -391,7 +392,8 @@ final class HttpChannelPool implements AutoCloseable {
                 final SessionProtocol protocol = getProtocolIfHealthy(channel);
                 if (closed || protocol == null) {
                     channel.close();
-                    promise.completeExceptionally(UnprocessedRequestException.get());
+                    promise.completeExceptionally(
+                            new UnprocessedRequestException(ClosedSessionException.get()));
                     return;
                 }
 
@@ -421,7 +423,8 @@ final class HttpChannelPool implements AutoCloseable {
                 } else {
                     // Server set MAX_CONCURRENT_STREAMS to 0, which means we can't send anything.
                     channel.close();
-                    promise.completeExceptionally(UnprocessedRequestException.get());
+                    promise.completeExceptionally(
+                            new UnprocessedRequestException(RefusedStreamException.get()));
                 }
 
                 channel.closeFuture().addListener(f -> {
@@ -452,10 +455,10 @@ final class HttpChannelPool implements AutoCloseable {
                     }
                 });
             } else {
-                promise.completeExceptionally(future.cause());
+                promise.completeExceptionally(new UnprocessedRequestException(future.cause()));
             }
         } catch (Exception e) {
-            promise.completeExceptionally(e);
+            promise.completeExceptionally(new UnprocessedRequestException(e));
         }
     }
 
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
@@ -202,11 +202,13 @@ final class HttpClientDelegate implements Client<HttpRequest, HttpResponse> {
             final HttpSession session = HttpSession.get(channel);
             res.init(session.inboundTrafficController());
             final SessionProtocol sessionProtocol = session.protocol();
+
+            // Should never reach here.
             if (sessionProtocol == null) {
                 needsRelease = false;
                 try {
                     // TODO(minwoox): Make a test that handles this case
-                    final UnprocessedRequestException cause = UnprocessedRequestException.get();
+                    final NullPointerException cause = new NullPointerException("sessionProtocol");
                     handleEarlyRequestException(ctx, req, cause);
                     res.close(cause);
                 } finally {
--- a/core/src/main/java/com/linecorp/armeria/client/RefusedStreamException.java
+++ b/core/src/main/java/com/linecorp/armeria/client/RefusedStreamException.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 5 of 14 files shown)


core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -202,11 +202,13 @@ private void doExecute(PooledChannel pooledChannel, ClientRequestContext ctx,
...
             res.init(session.inboundTrafficController());
             final SessionProtocol sessionProtocol = session.protocol();
+
+            // Should never reach here.
             if (sessionProtocol == null) {
                 needsRelease = false;
                 try {
                     // TODO(minwoox): Make a test that handles this case
-                    final UnprocessedRequestException cause = UnprocessedRequestException.get();
+                    final NullPointerException cause = new NullPointerException("sessionProtocol");
```
ikhoon: I changed this into NullPointerException because it is abnormal behavior and not a retriable case.

core/src/main/java/com/linecorp/armeria/client/RefusedStreamException.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.client;
+
+/**
+ * A {@link RuntimeException} raised when a server set MAX_CONCURRENT_STREAMS to 0,
```
trustin: `HTTP/2 <a href="https://httpwg.org/specs/rfc7540.html#SETTINGS_MAX_CONCURRENT_STREAMS">{@code MAX_CONCURRENT_STREAMS}</a>`

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
```diff
@@ -29,19 +29,26 @@
...
-
     /**
-     * Returns an {@link UnprocessedRequestException} which may be a singleton or a new instance, depending on
-     * whether {@linkplain Flags#verboseExceptions() the verbose exception mode} is enabled.
+     * Creates a new instance with the specified {@code message} and {@code cause}.
      */
-    public static UnprocessedRequestException get() {
-        return Flags.verboseExceptions() ? new UnprocessedRequestException() : INSTANCE;
+    public UnprocessedRequestException(@Nullable String message, @Nullable Throwable cause) {
+        super(message, cause);
```
trustin: Set `false` to `writableStackTrace` if `Flags.verboseExceptions()` is `false`.

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
```diff
@@ -29,19 +29,26 @@
...
     }
 
-    private UnprocessedRequestException() {}
+    /**
+     * Creates a new instance with the specified {@code cause}.
+     */
+    public UnprocessedRequestException(@Nullable Throwable cause) {
+        super(cause);
```
trustin: Set `false` to `writableStackTrace` if `Flags.verboseExceptions()` is `false`.

core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingExceedException.java
```diff
@@ -0,0 +1,39 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.client.limit;
+
+/**
+ * A {@link RuntimeException} raised when {@link ConcurrencyLimitingClient#numActiveRequests()} exceeds
+ * the configured {@code maxConcurrency}.
```
trustin: It's impossible to exceed the max concurrency with `ConcurrencyLimitingClient`. The exception is raised only when the client was not able to send a request until it reaches at the timeout. Could you rename this class and rephrase the Javadoc?

core/src/main/java/com/linecorp/armeria/internal/InboundTrafficController.java
```diff
@@ -82,7 +82,7 @@ public void inc(int numProducedBytes) {
     public void dec(int numConsumedBytes) {
         final int oldValue = getAndAdd(-numConsumedBytes);
         if (oldValue > lowWatermark && oldValue - numConsumedBytes <= lowWatermark) {
-            // Just went below high watermark
+            // Just went below low watermark
```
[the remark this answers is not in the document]
trustin: Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
```diff
@@ -22,10 +22,9 @@
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.RequestContextWrapper;
-import com.linecorp.armeria.server.ServiceRequestContext;
 
 /**
- * Wraps an existing {@link ServiceRequestContext}.
+ * Wraps an existing {@link ClientRequestContext}.
```
trustin: Thanks!

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -391,7 +392,8 @@ private void notifyConnect(SessionProtocol desiredProtocol, PoolKey key, Future<
                 final SessionProtocol protocol = getProtocolIfHealthy(channel);
                 if (closed || protocol == null) {
                     channel.close();
-                    promise.completeExceptionally(UnprocessedRequestException.get());
+                    promise.completeExceptionally(
+                            new UnprocessedRequestException(ClosedSessionException.get()));
```
trustin: There are two more `completeExceptionally()` methods below in this method you need to take care of.

ikhoon: Thanks! I wrapped below exception with UnprocessedRequestException.  :-)

core/src/main/java/com/linecorp/armeria/client/RefusedStreamException.java
ikhoon: I updated the link.

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
ikhoon: I overrided `fillInStackTrace` for verbose Exception. Is it workaround?
```java
@Override
public Throwable fillInStackTrace() {
    if (Flags.verboseExceptions()) {
        super.fillInStackTrace();
    }
    return this;
}
```

core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingExceedException.java
ikhoon: The `ConcurrencyLimitingExceedException` is changed into `PendingRequestTimeoutException`. What do you think about this name? 🧐

trustin: I like the new name. :+1:

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
```diff
@@ -29,19 +34,31 @@
...
-    public static UnprocessedRequestException get() {
-        return Flags.verboseExceptions() ? new UnprocessedRequestException() : INSTANCE;
+    public UnprocessedRequestException(Throwable cause) {
+        super(requireNonNull(cause, "cause"));
     }
 
-    private UnprocessedRequestException() {}
+    @Nonnull
+    @Override
+    public synchronized Throwable getCause() {
```
trustin: `synchronized` is unnecessary because `super.getCause()` is already synchronized, i.e.

```java
public Throwable getCause() {
    synchronized (this) {
        // return super.getCause() is actually:
        synchronized (this) {
            return (cause==this ? null : cause)
        }
    }
}
```

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
trustin: I think it's good as it is. The use of 4-arg constructor is mainly useful when we want to prevent overriding the stack trace or adding suppressed exceptions, i.e. when we want to make an exception a stateless singleton. For `UnprocessedRequestException`, that's not the case, so it's just enough to override `fillInStackTrace()`.

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
```diff
@@ -29,19 +34,31 @@
...
 
     /**
-     * Returns an {@link UnprocessedRequestException} which may be a singleton or a new instance, depending on
-     * whether {@linkplain Flags#verboseExceptions() the verbose exception mode} is enabled.
+     * Creates a new instance with the specified {@code cause}.
      */
-    public static UnprocessedRequestException get() {
-        return Flags.verboseExceptions() ? new UnprocessedRequestException() : INSTANCE;
+    public UnprocessedRequestException(Throwable cause) {
+        super(requireNonNull(cause, "cause"));
```
trustin: How about passing the string representation of cause as a message?

    super(requireNonNull(cause, "cause").toString(), cause);

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
ikhoon: @trustin Thanks for your kind explanation. 🙏

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
ikhoon: I didn't think of that. Thanks.

core/src/main/java/com/linecorp/armeria/client/UnprocessedRequestException.java
ikhoon: It looks better than the null message. Thanks for all of your nice and gentle feedback. 😁

core/src/main/java/com/linecorp/armeria/client/GoAwayReceivedException.java
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.client;
+
+/**
+ * A {@link RuntimeException} raised when a server sent an HTTP/2 GOAWAY frame with
```
trustin: One last nit: How about adding hyperlink to `HTTP/2 GOAWAY frame`?

trustin:
@ikhoon Could you update the PR description?

ikhoon:
@trustin Sure, I will update the PR about the changes soon.

ikhoon:
The squash commit message was updated and PR description was also updated with that.

core/src/main/java/com/linecorp/armeria/client/GoAwayReceivedException.java
ikhoon: Thanks for addressing that! After I saw your other review about javadoc. I have the same idea.

trustin:
@minwoox Any other comments?

Issue #1851, opened 2019-06-24
What's the minimum Java version requirement of Armeria? (was: Why armeria require java11?)


Lsk08:
Hi all,
  I am wondering why armeria require java11?
  Is there a must or just want something new?

trustin:
Armeria requires Java 8. Java 11 is required only when you build Armeria itself.

trustin:
Let me know if you cannot run Armeria on Java 8. We will fix it.

PR #1853, opened 2019-06-24
Add 'requirements' to README.md and web site

trustin:
Related: #1851

Motivation:

Some users misinterpreted our `README.md` and thought our minimum
requirement is Java 11, which is not correct.

Modifications:

- Make it clear that our minimum requirement is Java 8.

Result:

- Hopefully less confusion


trustin:
Thanks for reviewing.

PR #1857, opened 2019-06-26
Ensure a user set the endpoint path in DocService when the path is not…

### LINKED ISSUE #1856: DocService annotated http paths are mangled if not exact
This line mangles all annotated http paths that aren't exact.
https://github.com/line/armeria/commit/35416c2b77fd020ee37c4c42d582b142400bbefd#diff-3b29757cfa7d6be5d3ca726a154114b1R54

will chop off the first 6 characters of the path mangling it badly. While the path may not be usable anyway, it will not give a unexpected 404 or 405 error message.

@minwoox If you're doing https://github.com/line/armeria/issues/1855 anyway, this might be an easy fix?

minwoox:
… an exact path type

Motivation:
When the path of s service in `DocService` is not an exact path type, a user should set the endpoint path
and we must validate it.

Modifications:
- Ensure a user set the endpoint path when the path is not an exact type.
- Rename `PROXY_PORT` to `ARMERIA_PORT` for clarity.

Result:
- Close #1856

### DIFF
--- a/docs-client/src/containers/MethodPage/DebugPage.tsx
+++ b/docs-client/src/containers/MethodPage/DebugPage.tsx
@@ -396,10 +396,8 @@ class DebugPage extends React.PureComponent<Props, State> {
             params.set('queries', queries);
           }
         } else {
-          if (endpointPath) {
-            this.validateEndpointPath(endpointPath);
-            params.set('endpoint_path', endpointPath);
-          }
+          this.validateEndpointPath(endpointPath);
+          params.set('endpoint_path', endpointPath);
         }
       }
 
@@ -449,6 +447,9 @@ class DebugPage extends React.PureComponent<Props, State> {
   };
 
   private validateEndpointPath(endpointPath: string) {
+    if (!endpointPath) {
+      throw new Error(`You should set the ENDPOINT PATH.`);
+    }
     const method = this.props.method;
     const endpoint = method.endpoints[0];
     const regexPathPrefix = endpoint.regexPathPrefix;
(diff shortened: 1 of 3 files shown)


docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -449,6 +447,9 @@ class DebugPage extends React.PureComponent<Props, State> {
   };
 
   private validateEndpointPath(endpointPath: string) {
+    if (!endpointPath) {
+      throw new Error(`You should set the ENDPOINT PATH.`);
```
ikhoon: nit: How about just using single(`'`) quotes here? :-)

minwoox: Oops~ fixed. Thank you. 😉

docs-client/src/containers/MethodPage/DebugPage.tsx
```diff
@@ -449,6 +447,9 @@ class DebugPage extends React.PureComponent<Props, State> {
   };
 
   private validateEndpointPath(endpointPath: string) {
+    if (!endpointPath) {
+      throw new Error('You must set the ENDPOINT PATH.');
```
trustin: `specify the endpoint path`? (set -> specify, no reason to capitalize)

minwoox: Fixed, thanks!

trustin:
Thanks, @minwoox!

minwoox:
Thanks for reviewing!

PR #1859, opened 2019-06-26
Ignore `CancelledSubscriptionException` and `AbortedStreamException`

### LINKED ISSUE #1847: Webflux: `CancelledSubscriptionException` for a request but `ServerHttpResponse` already committed (200 OK)
Perhaps we should suppress `CancelledSubscriptionException` when a response has been committed already.

trustin:
.. while handling a WebFlux response.

Motivation:

When handling a `HEAD` request, Webflux can get the response from a `GET`
request handler and cancel the body subscription immediately. In that
case, `PublisherBasedHttpResponse` will fail its `completionFuture` with
a `CancelledSubscriptionException`, which will be propagated to WebFlux
as an unhandled exception.

Modifications:

- Filter out `CancelledSubscriptionException` and
  `AbortedStreamException` so that WebFlux does not write an ERROR log
  about an unhandled exception.
- Do not log at DEBUG level about a cancelled stream when the current
  request method is `HEAD` because it's expected.

Result:

- Less noisy error handling
- Fixes #1847


trustin:
Thanks for reviewing!

PR #1863, opened 2019-06-28
Add missing dependency for `javax.annotation.Generated`

trustin:
Motivation:

In some configuration, the `java.xml.ws.annotation` module, which
contains `javax.annotation.Generated`, is not added by default. As a
result, a user will get the 'cannot find symbol' compilation error.

Modifications:

- Add an explicit dependency to `jakarta.annotation:jakarta.annotation-api`
- Exclude the `javax.annotation:javax.annotation-api` which has been
  superceded by `jakarta.annotation-api`

Result:

- No more compilation errors

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -353,6 +353,8 @@ org.checkerframework:
 org.curioswitch.curiostack:
   protobuf-jackson:
     version: '1.0.0'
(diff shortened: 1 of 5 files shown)


minwoox:
Thanks @trustin!

PR #1867, opened 2019-06-28
Handle `HttpStatus/ResponseException` in `THttpService`

### LINKED ISSUE #1839: Make `THttpService` respect `HttpStatusException` and `HttpResponseException`
When a Thrift service fails with an `HttpStatusException` or an `HttpResponseException`. `THttpService` wraps it with a `TApplicationException` and sends a `200 OK` response. Instead, we could make `THttpService` respect such exceptions and send a proper HTTP error response.

This can be useful when a service wants to send a timeout response such as `503 Service Unavailable`, because otherwise a client will have to handle two different timeouts in two different ways:

- `503 Service Unavailable` (when Thrift service method failed to produce a response before the request times out.)
- `200 OK` with a Thrift-level timeout exception (when Thrift service method produces an error response that tells the request has timed out.)

trustin:
Motivation:

When a Thrift service fails with an `HttpStatusException` or an
`HttpResponseException`. `THttpService` wraps it with a
`TApplicationException` and sends a `200 OK` response. Instead, we could
make `THttpService` respect such exceptions and send a proper HTTP error
response.

This can be useful when a service wants to send a timeout response such
as `503 Service Unavailable`, because otherwise a client will have to
handle two different timeouts in two different ways:

- `503 Service Unavailable` (when Thrift service method failed to
  produce a response before the request times out.)
- `200 OK` with a Thrift-level timeout exception (when Thrift service
  method produces an error response that tells the request has timed out.)

Modifications:

- Make `THttpService` handle `HttpStatusException` and
  `HttpResponseException` rather than encoding it into a
  `TApplicationException`.

Result:

- Closes #1839

### DIFF
--- a/thrift/src/main/java/com/linecorp/armeria/server/thrift/THttpService.java
+++ b/thrift/src/main/java/com/linecorp/armeria/server/thrift/THttpService.java
@@ -67,6 +67,8 @@ import com.linecorp.armeria.internal.thrift.TByteBufTransport;
 import com.linecorp.armeria.internal.thrift.ThriftFieldAccess;
 import com.linecorp.armeria.internal.thrift.ThriftFunction;
 import com.linecorp.armeria.server.AbstractHttpService;
+import com.linecorp.armeria.server.HttpResponseException;
+import com.linecorp.armeria.server.HttpStatusException;
 import com.linecorp.armeria.server.Service;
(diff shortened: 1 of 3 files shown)


trustin:
Thanks for reviewing :-)

Issue #1822, opened 2019-06-07
Allow repeating `@Path` annotations


trustin:
A user wants to specify multiple path pattens for a single service method in an annotated service. For example:

```java
@Get
@Post
@Path("/v1/some_api")
@Path("/v2/some_api")
```

Alternatively, we could consider allowing repeating `@METHOD` annotations, e.g.

```java
@Post("/v1/some_api")
@Post("/v2/some_api")
```

jrhee17:
mind if I work on this? :D

minwoox:
👍

jrhee17:
I feel like the first proposal and second proposal both are reasonable and can be worked on in separate iterations. I'll start with working on the first proposal (allowing multiple paths, but not multiple `@METHOD` annotations)

trustin:
OK for supporting both patterns.

minwoox:
Closed by #1870

PR #1870, opened 2019-06-29
Support multiple Path annotations

jrhee17:
Motivation:
A user might want to specify multiple paths for annotated http services. 

Modification:
- Allow repeatable annotations for `@Path` annotation.
- Different behavior in path declaration (described below)
- One `AnnotatedHttpService` is created for each `(httpMethod, pathMapping)`
    - It is difficult to accommodate multiple pathMappings using a single `AnnotatedHttpService` because `AnnotatedValueResolver` may be different
    - It may be possible to optimize this by serving multiple  `httpMethod`s that have the same `pathMapping` using a single `AnnotatedHttpService` at the cost of code complexiy
- Merge `MethodInfo` endpoints with same (`httpMethod`, `methodName`)
- Known issue: overloaded methods for `AnnotatedHttpService` are not displayed in docs. I'm thinking this can be tackled in a separate PR.

#### Breaking change in path declaration behavior
**AS-IS**
```
@Get("/")
@Post
public void method1() {} // accepts both Get("/") and Post("/")
```
**TO-BE**
```
@Get("/")
@Post
public void method1() {} // exception thrown. Path declared for Get not applied to Post
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -18,7 +18,6 @@ package com.linecorp.armeria.internal.annotation;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static com.google.common.collect.Sets.toImmutableEnumSet;
 import static com.linecorp.armeria.internal.ArmeriaHttpUtil.concatPaths;
 import static com.linecorp.armeria.internal.RouteUtil.ensureAbsolutePath;
 import static com.linecorp.armeria.internal.annotation.AnnotatedValueResolver.toRequestObjectResolvers;
@@ -42,11 +41,11 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.EnumMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -211,8 +210,9 @@ public final class AnnotatedHttpServiceFactory {
 
         final List<Method> methods = requestMappingMethods(object);
         return methods.stream()
-                      .map((Method method) -> create(pathPrefix, object, method, exceptionHandlerFunctions,
-                                                     requestConverterFunctions, responseConverterFunctions))
+                      .flatMap((Method method) ->
+                                       create(pathPrefix, object, method, exceptionHandlerFunctions,
+                                              requestConverterFunctions, responseConverterFunctions).stream())
                       .collect(toImmutableList());
     }
 
@@ -255,35 +255,39 @@ public final class AnnotatedHttpServiceFactory {
     }
 
     /**
-     * Returns an {@link AnnotatedHttpService} instance defined to {@code method} of {@code object} using
-     * {@link Path} annotation.
+     * Returns a list of {@link AnnotatedHttpService} instances. A single {@link AnnotatedHttpService} is
+     * created per each {@link Route} associated with the {@code method}.
      */
     @VisibleForTesting
-    static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Method method,
-                                              List<ExceptionHandlerFunction> baseExceptionHandlers,
-                                              List<RequestConverterFunction> baseRequestConverters,
-                                              List<ResponseConverterFunction> baseResponseConverters) {
+    static List<AnnotatedHttpServiceElement> create(String pathPrefix, Object object, Method method,
+                                                    List<ExceptionHandlerFunction> baseExceptionHandlers,
+                                                    List<RequestConverterFunction> baseRequestConverters,
+                                                    List<ResponseConverterFunction> baseResponseConverters) {
 
         final Set<Annotation> methodAnnotations = httpMethodAnnotations(method);
         if (methodAnnotations.isEmpty()) {
             throw new IllegalArgumentException("HTTP Method specification is missing: " + method.getName());
         }
 
-        final Set<HttpMethod> methods = toHttpMethods(methodAnnotations);
-        if (methods.isEmpty()) {
-            throw new IllegalArgumentException(method.getDeclaringClass().getName() + '#' + method.getName() +
-                                               " must have an HTTP method annotation.");
-        }
         final Class<?> clazz = object.getClass();
-        final String pattern = findPattern(method, methodAnnotations);
+        final Map<HttpMethod, Set<String>> httpMethodPatternsMap = getHttpMethodPatternsMap(method,
+                                                                                            methodAnnotations);
         final String computedPathPrefix = computePathPrefix(clazz, pathPrefix);
+        final Set<MediaType> consumableMediaTypes = consumableMediaTypes(method, clazz);
+        final Set<MediaType> producibleMediaTypes = producibleMediaTypes(method, clazz);
 
-        final Route route = Route.builder()
-                                 .path(computedPathPrefix, pattern)
-                                 .methods(methods)
-                                 .consumes(consumableMediaTypes(method, clazz))
-                                 .produces(producibleMediaTypes(method, clazz))
-                                 .build();
+        final Set<Route> routes = httpMethodPatternsMap.entrySet().stream().flatMap(
+                pattern -> {
+                    final HttpMethod httpMethod = pattern.getKey();
+                    final Set<String> pathMappings = pattern.getValue();
+                    return pathMappings.stream().map(
+                            pathMapping -> Route.builder()
+                                                .path(computedPathPrefix, pathMapping)
+                                                .methods(httpMethod)
+                                                .consumes(consumableMediaTypes)
+                                                .produces(producibleMediaTypes)
+                                                .build());
+                }).collect(Collectors.toSet());
 
         final List<ExceptionHandlerFunction> eh =
                 getAnnotatedInstances(method, clazz, ExceptionHandler.class, ExceptionHandlerFunction.class)
@@ -294,41 +298,6 @@ public final class AnnotatedHttpServiceFactory {
         final List<ResponseConverterFunction> res =
                 getAnnotatedInstances(method, clazz, ResponseConverter.class, ResponseConverterFunction.class)
                         .addAll(baseResponseConverters).build();
-
-        List<AnnotatedValueResolver> resolvers;
-        try {
-            resolvers = AnnotatedValueResolver.ofServiceMethod(method, route.paramNames(),
-                                                               toRequestObjectResolvers(req));
-        } catch (NoParameterException ignored) {
-            // Allow no parameter like below:
-            //
-            // @Get("/")
-            // public String method1() { ... }
-            //
-            resolvers = ImmutableList.of();
-        }
-
-        final Set<String> expectedParamNames = route.paramNames();
-        final Set<String> requiredParamNames =
-                resolvers.stream()
-                         .filter(AnnotatedValueResolver::isPathVariable)
-                         .map(AnnotatedValueResolver::httpElementName)
-                         .collect(Collectors.toSet());
-
-        if (!expectedParamNames.containsAll(requiredParamNames)) {
-            final Set<String> missing = Sets.difference(requiredParamNames, expectedParamNames);
-            throw new IllegalArgumentException("cannot find path variables: " + missing);
-        }
-
-        // Warn unused path variables only if there's no '@RequestObject' annotation.
-        if (resolvers.stream().noneMatch(r -> r.annotationType() == RequestObject.class) &&
-            !requiredParamNames.containsAll(expectedParamNames)) {
-            final Set<String> missing = Sets.difference(expectedParamNames, requiredParamNames);
-            logger.warn("Some path variables of the method '" + method.getName() +
-                        "' of the class '" + clazz.getName() +
-                        "' do not have their corresponding parameters annotated with @Param. " +
-                        "They would not be automatically injected: " + missing);
-        }
 
         final Optional<HttpStatus> defaultResponseStatus = findFirst(method, StatusCode.class)
                 .map(code -> {
@@ -358,14 +327,30 @@ public final class AnnotatedHttpServiceFactory {
                         defaultHeaders.status().code(), methodAlias);
         }
 
-        // A CORS preflight request can be received because we handle it specially. The following
-        // decorator will prevent the service from an unexpected request which has OPTIONS method.
-        final Function<Service<HttpRequest, HttpResponse>,
-                ? extends Service<HttpRequest, HttpResponse>> initialDecorator;
-        if (methods.contains(HttpMethod.OPTIONS)) {
-            initialDecorator = Function.identity();
+        final ResponseHeaders responseHeaders = defaultHeaders.build();
+        final HttpHeaders responseTrailers = defaultTrailers.build();
+
+        return routes.stream().map(route -> {
+            final List<AnnotatedValueResolver> resolvers = getAnnotatedValueResolvers(req, route, method,
+                                                                                      clazz);
+            return new AnnotatedHttpServiceElement(
+                    route,
+                    new AnnotatedHttpService(object, method, resolvers, eh, res, route, responseHeaders,
+                                             responseTrailers),
+                    decorator(method, clazz, getInitialDecorator(route.methods())));
+        }).collect(Collectors.toList());
+    }
+
+    /**
+     * A CORS preflight request can be received because we handle it specially. The following
+     * decorator will prevent the service from an unexpected request which has OPTIONS method.
+     */
+    private static Function<Service<HttpRequest, HttpResponse>, ? extends Service<HttpRequest, HttpResponse>>
+            getInitialDecorator(Set<HttpMethod> httpMethods) {
+        if (httpMethods.contains(HttpMethod.OPTIONS)) {
+            return Function.identity();
         } else {
-            initialDecorator = delegate -> new SimpleDecoratingHttpService(delegate) {
+            return delegate -> new SimpleDecoratingHttpService(delegate) {
                 @Override
                 public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
                     if (req.method() == HttpMethod.OPTIONS) {
@@ -376,11 +361,47 @@ public final class AnnotatedHttpServiceFactory {
                 }
             };
         }
-        return new AnnotatedHttpServiceElement(route, new AnnotatedHttpService(object, method, resolvers,
-                                                                               eh, res, route,
-                                                                               defaultHeaders.build(),
-                                                                               defaultTrailers.build()),
-                                               decorator(method, clazz, initialDecorator));
+    }
+
+    private static List<AnnotatedValueResolver> getAnnotatedValueResolvers(List<RequestConverterFunction> req,
+                                                                           Route route, Method method,
+                                                                           Class<?> clazz) {
+        final Set<String> expectedParamNames = route.paramNames();
+        List<AnnotatedValueResolver> resolvers;
+        try {
+            resolvers = AnnotatedValueResolver.ofServiceMethod(method, expectedParamNames,
+                                                               toRequestObjectResolvers(req));
+        } catch (NoParameterException ignored) {
+            // Allow no parameter like below:
+            //
+            // @Get("/")
+            // public String method1() { ... }
+            //
+            resolvers = ImmutableList.of();
+        }
+
+        final Set<String> requiredParamNames =
+                resolvers.stream()
+                         .filter(AnnotatedValueResolver::isPathVariable)
+                         .map(AnnotatedValueResolver::httpElementName)
+                         .collect(Collectors.toSet());
+
+        if (!expectedParamNames.containsAll(requiredParamNames)) {
+            final Set<String> missing = Sets.difference(requiredParamNames, expectedParamNames);
+            throw new IllegalArgumentException("cannot find path variables: " + missing);
+        }
+
+        // Warn unused path variables only if there's no '@RequestObject' annotation.
+        if (resolvers.stream().noneMatch(r -> r.annotationType() == RequestObject.class) &&
+            !requiredParamNames.containsAll(expectedParamNames)) {
+            final Set<String> missing = Sets.difference(expectedParamNames, requiredParamNames);
+            logger.warn("Some path variables of the method '" + method.getName() +
+                        "' of the class '" + clazz.getName() +
+                        "' do not have their corresponding parameters annotated with @Param. " +
+                        "They would not be automatically injected: " + missing);
+        }
+
+        return resolvers;
     }
 
     /**
@@ -426,25 +447,6 @@ public final class AnnotatedHttpServiceFactory {
                 .stream()
                 .filter(annotation -> HTTP_METHOD_MAP.containsKey(annotation.annotationType()))
                 .collect(Collectors.toSet());
-    }
-
-    /**
-     * Returns {@link Set} of {@link HttpMethod}s mapped to HTTP method annotations.
-     *
-     * @see Options
-     * @see Get
-     * @see Head
-     * @see Post
-     * @see Put
-     * @see Patch
-     * @see Delete
-     * @see Trace
-     */
-    private static Set<HttpMethod> toHttpMethods(Set<Annotation> annotations) {
-        return annotations.stream()
-                          .map(annotation -> HTTP_METHOD_MAP.get(annotation.annotationType()))
-                          .filter(Objects::nonNull)
-                          .collect(toImmutableEnumSet());
     }
 
     /**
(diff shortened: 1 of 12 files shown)


trustin:
Test looks good to me. 👍

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -509,23 +514,23 @@ private static int order(Method method) {
...
         for (Annotation a : methodAnnotations) {
             final String p = (String) invokeValueMethod(a);
             if (DefaultValues.isUnspecified(p)) {
                 continue;
             }
-            checkArgument(pattern == null,
-                          "Only one path can be specified. (" + pattern + ", " + p + ')');
-            pattern = p;
+            checkArgument(patterns.isEmpty(),
+                          "Only one path can be specified. (" + patterns + ", " + p + ')');
```
kojilin: Just think if error message is hard to understand. it's more like `You can specify path through `Path` or  Http method annotation?

trustin:
How's it going? Just let us know if you have any questions working on this. :wink:

trustin:
Please let us know once this pull request is ready for reviews.

jrhee17:
I saw your comment late :P   Yep:) I'll remove the WIP label and add a comment when ready

ikhoon:
Fixes: #1822 (just for link)

trustin:
@jrhee17 Thanks a lot for coming back! Is it now ready for reviews?

jrhee17:
Hi :D .  I think overall this is good to go, but let me clean up code/add some documentation/add tests before asking for review.   I'll leave a comment when ready

jrhee17:
I think this PR is (finally) ready for review now~ :-)

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -358,14 +327,30 @@ static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Meth
...
+
+    /**
+     * A CORS preflight request can be received because we handle it specially. The following
+     * decorator will prevent the service from an unexpected request which has OPTIONS method.
+     */
+    private static Function<Service<HttpRequest, HttpResponse>, ? extends Service<HttpRequest, HttpResponse>>
+            getInitialDecorator(Set<HttpMethod> httpMethods) {
+        if (httpMethods.contains(HttpMethod.OPTIONS)) {
+            return Function.identity();
         } else {
```
trustin: This `else` block could be unwrapped.

core/src/main/java/com/linecorp/armeria/server/docs/ServiceInfo.java
```diff
@@ -96,10 +122,39 @@ public String name() {
...
 
+    /**
+     * Duplicate {@link MethodInfo} with matching (name, httpMethod) are merged.
+     * Merge targets are endpoints since {@code exampleHttpHeaders}, {@code exampleRequests},
+     * and {@code docString} cannot be different for a single method.
+     */
+    @VisibleForTesting
+    static Iterable<MethodInfo> mergeEndpoints(Iterable<MethodInfo> methodInfos) {
+        final Map<Pair<String, HttpMethod>, MethodInfo> methodInfoMap = new HashMap<>();
+        for (MethodInfo methodInfo: methodInfos) {
```
trustin: Missing space between `methodInfo` and `:`

core/src/main/java/com/linecorp/armeria/server/docs/ServiceInfo.java
```diff
@@ -96,10 +122,39 @@ public String name() {
...
+            final Pair<String, HttpMethod> pair = new Pair<>(methodInfo.name(), methodInfo.httpMethod());
```
trustin: Maybe we can just use `List<Object>` as a key and remove `Pair`?

core/src/main/java/com/linecorp/armeria/server/docs/ServiceInfo.java
```diff
@@ -96,10 +122,39 @@ public String name() {
...
+                                          value.parameters(), value.exceptionTypeSignatures(),
+                                          endpointInfos, value.exampleHttpHeaders(),
+                                          value.exampleRequests(), value.httpMethod(),
+                                          value.docString());
+                }
+            });
+        }
+        return methodInfoMap.values().stream()
+                            .sorted(comparing(MethodInfo::name).thenComparing(MethodInfo::httpMethod))
+                            .collect(Collectors.toList());
```
trustin: `ImmutableList.toImmutableList()`?

core/src/main/java/com/linecorp/armeria/server/docs/ServiceInfo.java
```diff
@@ -96,10 +122,39 @@ public String name() {
...
         return methods;
     }
 
+    /**
+     * Duplicate {@link MethodInfo} with matching (name, httpMethod) are merged.
+     * Merge targets are endpoints since {@code exampleHttpHeaders}, {@code exampleRequests},
+     * and {@code docString} cannot be different for a single method.
+     */
+    @VisibleForTesting
+    static Iterable<MethodInfo> mergeEndpoints(Iterable<MethodInfo> methodInfos) {
```
trustin: Could return `List<MethodInfo>` - Be as specific as possible when returning.

core/src/main/java/com/linecorp/armeria/server/docs/ServiceInfo.java
```diff
@@ -96,10 +122,39 @@ public String name() {
...
     @JsonProperty
-    public Set<MethodInfo> methods() {
+    public List<MethodInfo> methods() {
         return methods;
     }
 
+    /**
+     * Duplicate {@link MethodInfo} with matching (name, httpMethod) are merged.
+     * Merge targets are endpoints since {@code exampleHttpHeaders}, {@code exampleRequests},
+     * and {@code docString} cannot be different for a single method.
```
trustin: A little bit of rephrasing:

```
Merges the {@link MethodInfo}s with the same method name and {@link HttpMethod} pair
into a single {@link MethodInfo}. Note that only the {@link EndpointInfo}s are merged
because the {@link MethodsInfo}s being merged always have the same
{@code exampleHttpHeaders} and {@code exampleRequests}.
```

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceBuilderTest.java
```diff
@@ -68,12 +67,13 @@ public void root() {}
...
-            @Post("/")
-            public void root() {}
-        });
+        // TODO: Check if we need to keep supporting this
+        //        Server.builder().annotatedService(new Object() {
+        //            @Options
+        //            @Get
+        //            @Post("/")
+        //            public void root() {}
+        //        });
```
trustin: I think it's fine to stop supporting it and just raise an exception. How about adding  the test cases for it?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
```diff
@@ -181,6 +182,33 @@ public void testCreateAnnotatedServiceElementWithoutExplicitPathOnMethod() {
...
+        final List<AnnotatedHttpServiceElement> annotatedHttpServiceElements = getMethods(
+                MultiPathSingleMappingService.class, HttpResponse.class).flatMap(method -> {
+            final List<AnnotatedHttpServiceElement> annotatedHttpServices = create(
+                    "/", serviceObject, method, Lists.emptyList(), Lists.emptyList(), Lists.emptyList());
+            return annotatedHttpServices.stream();
+        }).collect(Collectors.toList());
+
+        annotatedHttpServiceElements.forEach(element -> {
+            assertThat(element.route().methods()).hasSize(1);
+        });
```
trustin: Could you add more detailed assertions than just checking the size only?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceTest.java
```diff
@@ -672,6 +675,46 @@ public String headerWithoutValue(RequestContext ctx,
...
+        @Path("/multiGetWithParam2/{param}")
+        public String multiGetWithParam(RequestContext ctx, @Param String param) {
+            return param;
+        }
+
+        @Get
+        @Path("/multiGetDiffParam1/{param1}")
+        @Path("/multiGetDiffParam2/{param2}")
+        public String multiGetDiffParam(RequestContext ctx, @Param String param1, @Param String param2) {
+            return "multiGetDiffParam";
```
trustin: Could you return a string made from `param1` and `param2` and make sure the both path patterns are handled correctly?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceBuilderTest.java
```diff
@@ -191,6 +191,59 @@ public void root(@Header("a") ArrayList<String> a,
...
+            @Post
+            @Path("/a")
+            @Path("/b")
+            public void root(BeanB b) {}
+        });
+
+        Server.builder().annotatedService(new Object() {
+            @Path("/")
+            @Get("/")
+            public void root() {}
```
trustin: What happens in the following case?

```java
@Get("/a")
@Path("/b")
```

Does it have the same effect as:

```java
@Get
@Path("/a")
@Path("/b")
```

..?

Then, what happens in the following case?

```java
@Get("/a")
@Post("/b")
@Path("/c")
```

Is it same as:

```java
// Binding 1
@Get
@Path("/a")
@Path("/c")
// Binding 2
@Post
@Path("/b")
@Path("/c")
```

..?

Could you add the test cases for these cases in:

- `AnnotatedHttpServiceBuilderTest`
- `AnnotatedHttpServiceTest`

Alternatively, we could consider always rejecting the case where a method annotation with a path is used with a `Path` annotation. Which do you think is better, @minwoox @ikhoon @jrhee17 ?

ikhoon: To reduce complication, I prefer to using a "/my/path" only one annotation. 😀  
- `@Get` and `@Path("/b")` 👍
- `@Get("/a")` and `@Path()` 👍
- `@Get("/a")` and `@Path("/b")` 👎

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -358,14 +327,30 @@ static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Meth
...
+
+        return routes.stream().map(route -> {
+            final List<AnnotatedValueResolver> resolvers = getAnnotatedValueResolvers(req, route, method,
+                                                                                      clazz);
+            return new AnnotatedHttpServiceElement(
+                    route,
+                    new AnnotatedHttpService(object, method, resolvers, eh, res, route, responseHeaders,
+                                             responseTrailers),
+                    decorator(method, clazz, getInitialDecorator(route.methods())));
+        }).collect(Collectors.toList());
```
ikhoon: One more `ImmutableList.toImmutableList()`?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -255,35 +255,39 @@ private static HttpStatus defaultResponseStatus(Optional<HttpStatus> defaultResp
...
+                    final HttpMethod httpMethod = pattern.getKey();
+                    final Set<String> pathMappings = pattern.getValue();
+                    return pathMappings.stream().map(
+                            pathMapping -> Route.builder()
+                                                .path(computedPathPrefix, pathMapping)
+                                                .methods(httpMethod)
+                                                .consumes(consumableMediaTypes)
+                                                .produces(producibleMediaTypes)
+                                                .build());
+                }).collect(Collectors.toSet());
```
ikhoon: `ImmutableSet.toImmutableSet()`?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceBuilderTest.java
minwoox: > Alternatively, we could consider always rejecting the case where a method annotation with a path is used with a Path annotation

That's a great idea. That will make everything simple. To clarify:
```java
@Get("/a")   // Exception
@Path("/b")  // @Get should be alone or should not have "/a" when it's used with @Path
```
```java
@Get("/a")   // Exception
@Path        // Path should always have a path string
```
```java
@Get         // OK
@Path("/a")
@Path("/b")
```
```java
@Get         // OK
@Post
@Path("/c")
```
One thing, we have to consider is when only `@Path` is specified:
```java
@Path("/a")
```
Do we have to throw an exception or should allow all HTTP methods.
I slighly lean toward allowing all HTTP methods because we do that in route builder.

jrhee17: > Could you add the test cases for these cases

roger that :D

>  consider always rejecting the case where a method annotation with a path is used with a Path annotation

+1 for me as well as it is visually easier to understand.
I was initially concerned there might be cases where users might want only a certain http method to have a path. In the long term, I think allowing multiple httpMethod patterns can help with this. (i.e. allowing multiple `@Get`, `@Post`, etc...)

> Do we have to throw an exception or should allow all HTTP methods.

+1 as it would also be consistent with other frameworks (like spring)

trustin: > Do we have to throw an exception or should allow all HTTP methods. I slighly lean toward allowing all HTTP methods because we do that in route builder.
> +1 as it would also be consistent with other frameworks (like spring)

+1 for which one? I'm not sure which is better to be honest :-)

jrhee17: just to clarify

```
@Get("/a")        // also OK
@Post("/b") 
@Path
```

> +1 for which one? I'm not sure which is better to be honest :-)

I meant the latter (allowing all http methods).

Looking at this again with a fresh mind, it does feel a little odd that the behavior of `@Path` changes depending on the existence of other annotations that aren't explicitly related.
🤔 🤔 🤔 🤔 🤔

minwoox:
```java
@Get("/a")        // Not OK!
@Post("/b") 
@Path
```
I think this is not ok @jrhee17!
`@Path` should always have a path string and the above example does not make any different from:
```java
@Get("/a")  
@Post("/b") 
```

So I think we should raise an exception. WDYT?

jrhee17: oops.  I meant the following. sry for the confusion :(
```
@Get("/a")  
@Post("/b") 
```

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceBuilderTest.java
```diff
@@ -244,18 +290,6 @@ public void root(@RequestObject HttpData value) {}
 
     @Test
     public void failedOf() {
-        assertThatThrownBy(() -> Server.builder().annotatedService(new Object() {
-            @Path("/")
-            @Get("/")
-            public void root() {}
-        })).isInstanceOf(IllegalArgumentException.class);
```
trustin: Yep, this has to fail.

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
```diff
@@ -181,6 +186,70 @@ public void testCreateAnnotatedServiceElementWithoutExplicitPathOnMethod() {
         });
     }
 
+    @Test
+    public void testMultiPathSuccessGetMapping() {
+        final List<AnnotatedHttpServiceElement> getServiceElements =
+                getServiceElements(new MultiPathSuccessService(), "getMapping", HttpMethod.GET);
+        assertThat(getServiceElements).hasSize(1);
+        assertThat(getServiceElements.get(0).route().methods()).isEqualTo(ImmutableSet.of(HttpMethod.GET));
+        assertThat(getServiceElements.get(0).route().paths()).isEqualTo(ImmutableList.of("/get", "/get"));
```
trustin: - `Route` implements `equals()`, so we could do the following for the most assertions in this test case?
   ```java
   assertThat(getServiceElements).containsOnly(
           Route.builder().get("/").build());
   ```
- Could you explain why two same paths (`/get`) are returned?

core/src/test/java/com/linecorp/armeria/server/docs/ServiceInfoTest.java
```diff
@@ -0,0 +1,83 @@
...
+                ImmutableList.copyOf(ServiceInfo.mergeEndpoints(inputMethodInfos));
+
+        assertEquals(4, collectMethods.size());
+
+        final Function<MethodInfo, Set<String>> getPaths = methodInfo ->
+                methodInfo.endpoints().stream().map(EndpointInfo::pathMapping).collect(Collectors.toSet());
+        assertEquals(ImmutableSet.of("exact:/bar1", "exact:/bar2"), getPaths.apply(collectMethods.get(0)));
+        assertEquals(ImmutableSet.of("exact:/bar3", "exact:/bar4"), getPaths.apply(collectMethods.get(1)));
+        assertEquals(ImmutableSet.of("exact:/foo1", "exact:/foo2"), getPaths.apply(collectMethods.get(2)));
+        assertEquals(ImmutableSet.of("exact:/foo3", "exact:/foo4"), getPaths.apply(collectMethods.get(3)));
```
trustin: Could you use `assertThat(...).containsExactly(...)`?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceBuilderTest.java
jrhee17: yep:)  the test case has been moved.

I'm haven't added any special logic for the following case yet. (I don't think we reached a clear consensus)
```java
@Path("/")
public void pathByItself() {}
```

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
jrhee17: > Route implements equals(), so we could do the following for the most assertions in this test case?

Looks much cleaner this way:) Thanks

> Could you explain why two same paths (/get) are returned?

I was thinking because of this!

https://github.com/line/armeria/blob/84d3b36239acc3276ef16baf7c002e2fc9cd8405/core/src/main/java/com/linecorp/armeria/server/ExactPathMapping.java#L40

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceBuilderTest.java
minwoox: Had a chat with @trustin and let's raise an exception in this case!

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceBuilderTest.java
```diff
@@ -191,6 +183,52 @@ public void root(@Header("a") ArrayList<String> a,
...
+            @Path("/a/{param1}")
+            @Path("/b/{param2}")
+            public void root(String param1, String param2) {}
+        });
+
+        // Multiple paths with same value
+        Server.builder().annotatedService(new Object() {
+            @Get
+            @Path("/a")
+            @Path("/a")
```
minwoox: Not sure that this is ok or not. Isn't it better to raise an exception in this case?

ikhoon: I'm not sure too :-) But I think throwing an exception is too strict because most people expect the same result `@Path("/a")`?

minwoox: Yes, it's too strict. I have thought about the case when a user wants to use this same annotation and there was none. If he/she does this, I thought he/she wants to use a different path but didn't notice it. So maybe raising an exception would be better. But It's fine as it is though.

ikhoon: I understand what you meant. which one is better? 🧐

minwoox: What do you think @jrhee17? If you think this is too much or hard to implement, I think we can just leave this as it is. 😄

jrhee17: Behavior-wise, I think warn/throwing an exception both are fine. I really have no pref. for this.

I think we should consider the following though:

1) If we stick with detecting strictly equal `ExactPathMapping`s, I think this can be do-able. If we consider, cases like `@Path("/path/{name}")`, `@Path("/path/{name2}")`, or bring `RegexPathMapping` into the mix, things might become complicated :(
2) Different methods with the same `PathMapping`. For consistent behavior, I feel like the following should also warn/throw. Let me know what you guys think :)

```java
    @Get("/duplicatePath")
    public HttpResponse test1() { return HttpResponse.of(200); }

    @Get("/duplicatePath")
    public HttpResponse test2() { return HttpResponse.of(200); }
```

minwoox: I just realized that we have [rejectedRouteHandler](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java#L190) for this. So let's leave as it is. 😄

jrhee17: Oooh  I didn't know about this.   Let me fix the current impl. so that `rejectedRouteHandler` can take care of duplicate routes. 🙇

core/src/test/java/com/linecorp/armeria/server/docs/ServiceInfoTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private static MethodInfo createMethodInfo(String methodName, HttpMethod method,
+                                               String endpointPathMapping) {
+        final EndpointInfo endpoint = new EndpointInfoBuilder("*", endpointPathMapping)
+                .availableMimeTypes(MediaType.JSON_UTF_8).build();
+        return new MethodInfo(methodName, TypeSignature.ofBase("T"), ImmutableList.of(), ImmutableList.of(),
+                              ImmutableList.of(endpoint), method, null);
+    }
+
+    @Test
+    public void testCollectMethodsWithDifferentMethods() {
```
minwoox: Could remove `public` from all methods in this class.

site/src/sphinx/server-annotated-service.rst
```diff
@@ -106,7 +106,8 @@ a path and use the HTTP method annotations without a path to map multiple HTTP m
         @Put
         @Delete
         @Path("/hello")
-        public HttpResponse hello() { ... }
+        @Path("/hi")
+        public HttpResponse greeting() { ... }
```
minwoox: 👍

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceTest.java
```diff
@@ -672,6 +675,52 @@ public String headerWithoutValue(RequestContext ctx,
...
+        @Path("/pathSameParam1/{param}")
+        @Path("/pathSameParam2/{param}")
+        public String pathSameParam(RequestContext ctx, @Param String param) {
+            return param;
+        }
+
+        @Get
+        @Path("/pathDiffParam1/{param1}")
+        @Path("/pathDiffParam2/{param2}")
+        public String pathDiffParam(RequestContext ctx, @Param String param1, @Param String param2) {
```
minwoox: How about adding a test case which has different path types?
```java
@Default("1")
@Get
@Path("/a/b")
@Path("/a/{b}")
public String myMethod(@Param @Default("b") String b) {
}
```

core/src/main/java/com/linecorp/armeria/server/docs/ServiceInfo.java
```diff
@@ -74,11 +80,7 @@ public ServiceInfo(String name,
                        @Nullable String docString) {
 
         this.name = requireNonNull(name, "name");
-
-        requireNonNull(methods, "methods");
-
-        this.methods = ImmutableSortedSet.copyOf(comparing(MethodInfo::name)
-                                                         .thenComparing(MethodInfo::httpMethod), methods);
+        this.methods = ImmutableList.copyOf(mergeEndpoints(requireNonNull(methods)));
```
minwoox: So there's no reason to use list here?

docs-client/src/containers/MethodPage/index.tsx
```diff
@@ -79,14 +79,12 @@ function removeBrackets(headers: string): string {
   return headers.substring(1, length - 1).trim();
 }
 
-function isExactPathMapping(method: Method): boolean {
+function isSingleExactPathMapping(method: Method): boolean {
```
minwoox: We use this value to display different forms in doc service.
If this is `true`, `HttpQueryString` shows up. If not, `EndpointPath` shows up.
When multiple paths are specified, we should display `EndpointPath` to get a whole path from the user.
So this is the right way to achieve the behavior at the moment. But when we implement [this](https://github.com/line/armeria/issues/2167), we should reconsider this behavior. Just left a comment for the remind. 😄

core/src/main/java/com/linecorp/armeria/server/docs/ServiceInfo.java
```diff
@@ -43,7 +49,7 @@
 public final class ServiceInfo {
 
     private final String name;
-    private final Set<MethodInfo> methods;
+    private final List<MethodInfo> methods;
```
minwoox: Question: is this should be `List`?

site/src/sphinx/server-annotated-service.rst
jrhee17: 👋

core/src/main/java/com/linecorp/armeria/server/docs/ServiceInfo.java
jrhee17: you're right. reverted back to an `ImmutableSortedSet`!

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -295,9 +301,76 @@ static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Meth
...
+    /**
+     * A CORS preflight request can be received because we handle it specially. The following
+     * decorator will prevent the service from an unexpected request which has OPTIONS method.
+     */
+    private static Function<Service<HttpRequest, HttpResponse>, ? extends Service<HttpRequest, HttpResponse>>
+            getInitialDecorator(Set<HttpMethod> httpMethods) {
+        if (httpMethods.contains(HttpMethod.OPTIONS)) {
+            return Function.identity();
+        }
+        return delegate -> new SimpleDecoratingHttpService(delegate) {
```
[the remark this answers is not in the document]
jrhee17: Done :D

minwoox:
Thanks a lot, @jrhee17! 🙇 
Hope to see your next work. 😄

PR #1872, opened 2019-07-01
Add `current/mapCurrent/currentOrNull()` to `Service/ClientRequestContext`

### LINKED ISSUE #1869: Add `current()` and `mapCurrent()` to `Client/ServiceRequestContext`
A user often has to downcast `RequestContext`:

```java
ServiceRequestContext ctx = (ServiceRequestContext) RequestContext.current();
```

Because of this, the code often gets pretty verbose:

```java
((ServiceRequestContext) RequestContext.current()).routingContext().hostname();
```

If we has `ServiceRequestContext.current()`, it would have been much nicer:

```java
ServiceRequestContext.current().routingContext().hostname();
```

If the current context is not a `ServiceRequestContext` (or `ClientRequestContext`), we could simply raise an `IllegalStateException`.

trustin:
Motivation:

A user often has to downcast `RequestContext`:

    ServiceRequestContext ctx = (ServiceRequestContext) RequestContext.current();

Because of this, the code often gets pretty verbose:

    ((ServiceRequestContext) RequestContext.current()).routingContext().hostname();

If we have `ServiceRequestContext.current()`, it would have been much nicer:

  ServiceRequestContext.current().routingContext().hostname();

If the current context is not a `ServiceRequestContext` (or `ClientRequestContext`), we could simply raise an `IllegalStateException`.

Modifications:

- Add `currentOrNull()` to `RequestContext`
- Add `current()`, `mapCurrent()` and `currentOrNull()` to `Service/ClientRequestContext`
- Use `Service/ClientRequestContext.current/mapCurrent/currentOrNull()` wherever
  possible.

Result:

- Closes #1869

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -16,11 +16,14 @@
 
 package com.linecorp.armeria.client;
 
+import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
 import java.net.URI;
 import java.time.Duration;
 import java.util.Map.Entry;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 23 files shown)


core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -38,6 +41,43 @@
...
+                   "The current context is not a client-side context: %s", ctx);
+        return (ClientRequestContext) ctx;
+    }
+
+    /**
+     * Maps the context of the {@link Request} that is being handled in the current thread.
+     *
+     * @param mapper the {@link Function} that maps the {@link ClientRequestContext}
+     * @param defaultValueSupplier the {@link Supplier} that provides the value when the context is unavailable
+     *                             in the current thread. If {@code null}, the {@code null} will be returned
```
minwoox: Don't we need to explain here when the `ctx` is not a `ClientRequestContext`?
Or I think, how about raising an exception when it does? :-)

trustin: That's a good idea. Let me raise an exception when the type is not correct.

minwoox:
Still LGTM 👍

trustin:
Thanks for reviewing, @anuraaga, @minwoox and @ikhoon !

PR #1873, opened 2019-07-01
Enable TCP fallback for the default DNS name resolver

trustin:
Motivation:

@normanmaurer kindly suggested me to enable TCP fallback for our default
DNS name resolver, and we definitely should to handle a large DNS
response.

Modifications:

- Enable TCP fallback by calling `DnsNameResolverBuilder.socketChannelType()`

Result:

- Can handle a large DNS response


trustin:
Thanks for reviewing :-)

PR #1878, opened 2019-07-04
New `HealthCheckService` with long-polling support

trustin:
Motivation:

- By adding long-polling support to a health check service, an advanced
  health check client can get notified immediately when a server becomes
  healthy (or unhealthy), rather than polling periodically.
- `ManagedHttpHealthCheckService` provides only a very simple update
  mechanism that's hard to customize.

Modifications:

- Deprecate `*HttpHealthCheckService`
- Add `HealthCheckService` and its builder
- Add `ListenableHealthChecker` to implement notification

Result:

- Long-polling support for health checks, which will greatly reduce the
  delay between the actual moment when the server got unhealthy and when
  the client learns the unhealthiness.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
@@ -35,9 +35,9 @@ import com.linecorp.armeria.server.ServiceRequestContext;
 
 final class DefaultHealthCheckUpdateHandler implements HealthCheckUpdateHandler {
 
-    static final DefaultHealthCheckUpdateHandler INSTANCE = new DefaultHealthCheckUpdateHandler();
+    private static final ObjectMapper mapper = new ObjectMapper();
 
-    private final ObjectMapper mapper = new ObjectMapper();
+    static final DefaultHealthCheckUpdateHandler INSTANCE = new DefaultHealthCheckUpdateHandler();
 
     private DefaultHealthCheckUpdateHandler() {}
 
--- a/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
@@ -15,6 +15,8 @@
  */
 package com.linecorp.armeria.server.healthcheck;
 
+import static java.util.Objects.requireNonNull;
+
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
@@ -42,6 +44,7 @@ final class DefaultHealthCheckUpdateHandler implements HealthCheckUpdateHandler 
 
     @Override
     public CompletionStage<Boolean> handle(ServiceRequestContext ctx, HttpRequest req) throws Exception {
+        requireNonNull(req, "req");
         switch (req.method()) {
             case PUT:
             case POST:
--- a/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
@@ -93,10 +93,9 @@ final class DefaultHealthCheckUpdateHandler implements HealthCheckUpdateHandler 
         if (contentType != null && !contentType.is(MediaType.JSON)) {
             throw HttpStatusException.of(HttpStatus.UNSUPPORTED_MEDIA_TYPE);
         }
+
         final Charset charset = contentType == null ? StandardCharsets.UTF_8
                                                     : contentType.charset().orElse(StandardCharsets.UTF_8);
-
-        final JsonNode node;
         try {
             return mapper.readTree(req.content(charset));
         } catch (IOException e) {
--- a/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
@@ -22,7 +22,6 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.CompletionStage;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.JsonNodeType;
@@ -43,7 +42,8 @@ final class DefaultHealthCheckUpdateHandler implements HealthCheckUpdateHandler 
     private DefaultHealthCheckUpdateHandler() {}
 
     @Override
-    public CompletionStage<Boolean> handle(ServiceRequestContext ctx, HttpRequest req) throws Exception {
+    public CompletionStage<HealthCheckUpdateResult> handle(ServiceRequestContext ctx,
+                                                           HttpRequest req) throws Exception {
         requireNonNull(req, "req");
         switch (req.method()) {
             case PUT:
@@ -56,7 +56,7 @@ final class DefaultHealthCheckUpdateHandler implements HealthCheckUpdateHandler 
         }
     }
 
-    private Boolean handlePut(AggregatedHttpRequest req) {
+    private HealthCheckUpdateResult handlePut(AggregatedHttpRequest req) {
         final JsonNode json = toJsonNode(req);
         if (json.getNodeType() != JsonNodeType.OBJECT) {
             throw HttpStatusException.of(HttpStatus.BAD_REQUEST);
@@ -70,25 +70,30 @@ final class DefaultHealthCheckUpdateHandler implements HealthCheckUpdateHandler 
             throw HttpStatusException.of(HttpStatus.BAD_REQUEST);
         }
 
-        return healthy.booleanValue();
+        return healthy.booleanValue() ? HealthCheckUpdateResult.HEALTHY
+                                      : HealthCheckUpdateResult.UNHEALTHY;
     }
 
-    private Boolean handlePatch(AggregatedHttpRequest req) {
-        final String json;
-        try {
-            json = mapper.writeValueAsString(toJsonNode(req));
-        } catch (JsonProcessingException e) {
-            throw new Error(e); // Never happens.
+    private HealthCheckUpdateResult handlePatch(AggregatedHttpRequest req) {
+        final JsonNode json = toJsonNode(req);
+        if (json.getNodeType() != JsonNodeType.ARRAY ||
+            json.size() != 1) {
+            throw HttpStatusException.of(HttpStatus.BAD_REQUEST);
         }
 
-        switch (json) {
-            case "[{\"op\":\"replace\",\"path\":\"/healthy\",\"value\":true}]":
-                return true;
-            case "[{\"op\":\"replace\",\"path\":\"/healthy\",\"value\":false}]":
-                return false;
-            default:
-                throw HttpStatusException.of(HttpStatus.BAD_REQUEST);
+        final JsonNode patchCommand = json.get(0);
+        final JsonNode op = patchCommand.get("op");
+        final JsonNode path = patchCommand.get("path");
+        final JsonNode value = patchCommand.get("value");
+        if (op == null || path == null || value == null ||
+            !"replace".equals(op.textValue()) ||
+            !"/healthy".equals(path.textValue()) ||
+            !value.isBoolean()) {
+            throw HttpStatusException.of(HttpStatus.BAD_REQUEST);
         }
+
+        return value.booleanValue() ? HealthCheckUpdateResult.HEALTHY
+                                    : HealthCheckUpdateResult.UNHEALTHY;
     }
 
     private JsonNode toJsonNode(AggregatedHttpRequest req) {
@@ -100,7 +105,8 @@ final class DefaultHealthCheckUpdateHandler implements HealthCheckUpdateHandler 
         final Charset charset = contentType == null ? StandardCharsets.UTF_8
                                                     : contentType.charset().orElse(StandardCharsets.UTF_8);
         try {
-            return mapper.readTree(req.content(charset));
+            return StandardCharsets.UTF_8.equals(charset) ? mapper.readTree(req.content().array())
+                                                          : mapper.readTree(req.content(charset));
         } catch (IOException e) {
             throw HttpStatusException.of(HttpStatus.BAD_REQUEST);
         }
--- a/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private JsonNode toJsonNode(AggregatedHttpRequest req) {
+        final MediaType contentType = req.contentType();
+        if (contentType != null && !contentType.is(MediaType.JSON)) {
+            throw HttpStatusException.of(HttpStatus.UNSUPPORTED_MEDIA_TYPE);
+        }
+        final Charset charset = contentType == null ? StandardCharsets.UTF_8
+                                                    : contentType.charset().orElse(StandardCharsets.UTF_8);
+
+        final JsonNode node;
```
ikhoon: This looks like a dead code.

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckServiceBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+        this.healthyResponse = requireNonNull(healthyResponse, "healthyResponse");
+        return this;
+    }
+
+    /**
+     * Sets the {@link AggregatedHttpResponse} to send when the {@link Service} is unhealthy. The following
+     * response is sent by default:
+     *
+     * <pre>{@code
+     * HTTP/1.1 200 OK
```
ikhoon: Is it `"503 Service Unavailable"` here?

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckServiceBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+     *                                     timeout specified in the {@code "prefer: wait=<n>"} header.
+     * @return {@code this}
+     * @see #longPolling(long)
+     */
+    public HealthCheckServiceBuilder longPolling(long maxLongPollingTimeoutMillis,
+                                                 float longPollingTimeoutJitterRate) {
+        checkArgument(maxLongPollingTimeoutMillis >= 0,
+                      "maxLongPollingTimeoutMillis: %s (expected: >= 0)", maxLongPollingTimeoutMillis);
+        checkArgument(longPollingTimeoutJitterRate >= 0 && longPollingTimeoutJitterRate <= 1,
+                      "longPollingTimeoutJitterRate: %s (longPollingTimeoutJitterRate: >= 0 && <= 1)",
```
ikhoon: `expected: >= 0 && <= 1`?

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckUpdateHandler.java
```diff
@@ -0,0 +1,46 @@
...
+import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+/**
+ * Handles {@code PUT}, {@code POST} or {@code PATCH} requests sent to {@link HealthCheckService}.
+ */
+@FunctionalInterface
+public interface HealthCheckUpdateHandler {
+    /**
+     * Determines if the healthiness of the {@link Server} needs to be change or not from the given
```
ikhoon: `be change` -> `be changed`

core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
trustin: Oops!

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckServiceBuilder.java
trustin: :sweat_smile:

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckUpdateHandler.java
trustin: Thanks!

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckServiceBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+     * Adds the specified {@link HealthChecker}s that determines the healthiness of the {@link Server}.
+     *
+     * @return {@code this}
+     */
+    public HealthCheckServiceBuilder checkers(HealthChecker... healthCheckers) {
+        return checkers(ImmutableSet.copyOf(requireNonNull(healthCheckers, "healthCheckers")));
+    }
+
+    /**
+     * Adds the specified {@link HealthChecker}s that determines the healthiness of the {@link Server}.
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckServiceBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+                                                                                 "{\"healthy\":false}");
+    private long maxLongPollingTimeoutMillis = DEFAULT_LONG_POLLING_TIMEOUT_MILLIS;
+    private float longPollingTimeoutJitterRate = DEFAULT_LONG_POLLING_TIMEOUT_JITTER_RATE;
+    @Nullable
+    private HealthCheckUpdateHandler updateHandler;
+
+    HealthCheckServiceBuilder() {}
+
+    /**
+     * Adds the specified {@link HealthChecker}s that determines the healthiness of the {@link Server}.
```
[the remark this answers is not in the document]
trustin: Thanks!

core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
```diff
@@ -0,0 +1,106 @@
...
+
+    static final DefaultHealthCheckUpdateHandler INSTANCE = new DefaultHealthCheckUpdateHandler();
+
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    private DefaultHealthCheckUpdateHandler() {}
+
+    @Override
+    public CompletionStage<Boolean> handle(ServiceRequestContext ctx, HttpRequest req) throws Exception {
+        switch (req.method()) {
```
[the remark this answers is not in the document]
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    private Boolean handlePatch(AggregatedHttpRequest req) {
+        final String json;
+        try {
+            json = mapper.writeValueAsString(toJsonNode(req));
+        } catch (JsonProcessingException e) {
+            throw new Error(e); // Never happens.
+        }
+
+        switch (json) {
+            case "[{\"op\":\"replace\",\"path\":\"/healthy\",\"value\":true}]":
```
anuraaga: Isn't it simpler and more performant too to read the values from the JSON node instead of converting back to string and doing string matching?

core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    private JsonNode toJsonNode(AggregatedHttpRequest req) {
+        final MediaType contentType = req.contentType();
+        if (contentType != null && !contentType.is(MediaType.JSON)) {
+            throw HttpStatusException.of(HttpStatus.UNSUPPORTED_MEDIA_TYPE);
+        }
+
+        final Charset charset = contentType == null ? StandardCharsets.UTF_8
+                                                    : contentType.charset().orElse(StandardCharsets.UTF_8);
+        try {
+            return mapper.readTree(req.content(charset));
```
anuraaga: Can consider calling `readTree(byte[])` if the charset is UTF_8 to avoid the copy for the vast majority case.

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
```diff
@@ -0,0 +1,420 @@
...
+ *
+ * <p>All health check responses will contain a {@code "long-polling-supported"} header whose value is
+ * {@code "true"} or {@code "false"}.</p>
+ *
+ * @see HealthCheckServiceBuilder
+ */
+public final class HealthCheckService implements HttpService, TransientService<HttpRequest, HttpResponse> {
+
+    private static final Logger logger = LoggerFactory.getLogger(HealthCheckService.class);
+    private static final AsciiString LONG_POLLING_SUPPORTED = HttpHeaderNames.of("long-polling-supported");
```
anuraaga: Don't think this is a standard header - how about `armeria-long-polling-supported`?

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
```diff
@@ -0,0 +1,420 @@
...
+            case HEAD:
+            case GET:
+                return newResponse(method, isHealthy);
+            case CONNECT:
+            case DELETE:
+            case OPTIONS:
+            case TRACE:
+                return HttpResponse.of(HttpStatus.METHOD_NOT_ALLOWED);
+        }
+
```
anuraaga: Think it helps to add a comment like `// POST or PUT`

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckUpdateHandler.java
```diff
@@ -0,0 +1,46 @@
...
+public interface HealthCheckUpdateHandler {
+    /**
+     * Determines if the healthiness of the {@link Server} needs to be changed or not from the given
+     * {@link HttpRequest}.
+     *
+     * @return A {@link CompletionStage} which is completed with one of the following values:
+     *         <ul>
+     *           <li>{@code true} - the {@link Server} has to be marked as 'healthy'.</li>
+     *           <li>{@code false} - the {@link Server} has to be marked as 'unhealthy'.</li>
+     *           <li>{@code null} - the healthiness of the {@link Server} has to remain as-is.</li>
```
anuraaga: How about using an enum instead of boolean? It's generally good if `null` an `false` have similar semantics for a `Boolean`

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
```diff
@@ -0,0 +1,420 @@
...
+        // TODO(trustin): Optimize this once https://github.com/line/armeria/issues/1835 is resolved.
+        final LongHolder timeoutMillisHolder = new LongHolder();
+        try {
+            ArmeriaHttpUtil.parseDirectives(prefer, (name, value) -> {
+                if ("wait".equals(name)) {
+                    timeoutMillisHolder.value = TimeUnit.SECONDS.toMillis(Long.parseLong(value));
+                }
+            });
+        } catch (NumberFormatException ignored) {
+            // Malformed "wait" value.
```
anuraaga: I think explicitly throwing the bad request exception here is easier to follow the logic.

core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
trustin: :sweat_smile: Was just too lazy. :-)

trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
trustin: Done

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
trustin: Added an assertion.

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckUpdateHandler.java
trustin: Good idea. Replaced with `HealthCheckUpdateResult`.

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
trustin: Done

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
trustin: The header name seems a little too long. I used `armeria-lphc` (long-polling health check) and made it expose the maximum allowed timeout. I'm not sure I like the new name, so please feel free to suggest a better name.

core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
```diff
@@ -0,0 +1,114 @@
...
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.server.HttpStatusException;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+final class DefaultHealthCheckUpdateHandler implements HealthCheckUpdateHandler {
+
+    static final DefaultHealthCheckUpdateHandler INSTANCE = new DefaultHealthCheckUpdateHandler();
+
+    private final ObjectMapper mapper = new ObjectMapper();
```
anuraaga: `static`

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
```diff
@@ -0,0 +1,437 @@
...
+ * <p>All health check responses will contain a {@code "armeria-lphc"} header whose value is the maximum
+ * allowed value of the {@code "Prefer: wait=<seconds>"} header. {@code 0} means long polling has been
+ * disabled. {@code "lphc"} stands for long-polling health check.</p>
+ *
+ * @see HealthCheckServiceBuilder
+ */
+public final class HealthCheckService implements HttpService, TransientService<HttpRequest, HttpResponse> {
+
+    private static final Logger logger = LoggerFactory.getLogger(HealthCheckService.class);
+    private static final AsciiString ARMERIA_LPHC = HttpHeaderNames.of("armeria-lphc");
```
anuraaga: I can't think of anything better other than `armeria-long-polling`, which is shorter than the previous one but lacks health check aspect

core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckServiceBuilder.java
```diff
@@ -0,0 +1,234 @@
...
+     * <pre>{@code
+     * HTTP/1.1 200 OK
+     * Content-Type: application/json; charset=utf-8
+     *
+     * { "healthy": true }
+     * }</pre>
+     *
+     * @return {@code this}
+     */
+    public HealthCheckServiceBuilder healthyResponse(AggregatedHttpResponse healthyResponse) {
```
anuraaga: Just a note, I don't think this will work if the content of the response is `ByteBufHttpData`, since it can only be read once. Since you shouldn't use that class without really knowing what's going on, I think it's ok but just want to make sure.

trustin: Ah, yeah. I just made a copy.

core/src/main/java/com/linecorp/armeria/server/healthcheck/DefaultHealthCheckUpdateHandler.java
trustin: Fixed

trustin:
Thanks for reviewing!

Issue #1810, opened 2019-06-03
String-based headers/content sanitizer implementation


trustin:
It would be convenient if we provide a simple headers/content sanitizer which filters the stringified headers/content. For example, a user had to implement a sanitizer like the following:

```java
   @Nonnull
   public static String changeLogWithoutSensitiveData(@Nonnull Object request) {
       String requestMessage = request.toString();

       if(!isRequireFiltering(requestMessage)) {
           return requestMessage;
       }

       return filterSensitiveData(requestMessage);
   }

   private static boolean isRequireFiltering(@Nonnull String message) {
       if(removeContentNames.stream().noneMatch(message::contains)) {
           return false;
       }

       return true;
   }

   @Nonnull
   private static String filterSensitiveData(@Nonnull String message) {
       Matcher matcher;

       for(Pattern pattern : removeContentRegxs) {
           matcher = pattern.matcher(message);
           message = matcher.replaceAll("");
       }

       return message;
   }
```

We could make the following changes:

- Change the return type of header sanitizers from `HttpHeaders` to `?` so that a user can return any object such as `String`.
- Provide an abstract sanitizer implementation that converts headers or content to a `String` and filters it as instructed by its abstract methods such as `boolean needsSanitization()` and `String sanitize(String)`.
  - Provide a concrete implementation that implements `needsSanitization()` and `sanitize()` to filter the `String` using regular expression(s).
- Add some convenience methods to `LoggingClient/ServiceBuilder` and `RequestLog.toString*()`:

  ```java
  new LoggingClientBuilder()
          .requestContentSanitizer(Pattern.compile(...),
                                   Pattern.compile(...))
          ....
  ```

trustin:
Fixed via #1879

joongsoo:
I doing this issue

amitvc:
@trustin I can implement this. I am new to the project but I am interested in contributing. Can I take this please?

trustin:
@amitvc Hi! That sounds great. How about coming first with some API proposal (some rough API examples) so we can make sure your pull request will be not too far from what we imagined?

Another question out of curiosity: Are you using Armeria or considering to use it?

amitvc:
@trustin Hello! I will send the proposal by today and get some examples. 
I am evaluating Armeria. I am a big fan of netty and have used it before.

amitvc:
[...]

PR #1879, opened 2019-07-05
Relax the output type constraint of sanitizer functions

trustin:
Related: #1810
Motivation:

When writing a header sanitization function, the function must be
written to return an `HttpHeaders`. For some users, it would be simpler
and/or more efficient if we relax the output type constraint of header
sanitizer functions, e.g.

    Function<HttpHeaders, Object> f = headers -> "sanitized!";

Modifications:

- Change the output type to `?` for all sanitizer functions.
- Miscellaneous:
  - Deduplicated sanitization logic in `DefaultRequestLog`.

Result:

- Convenience and efficiency

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
@@ -1072,15 +1072,15 @@ public class DefaultRequestLog implements RequestLog, RequestLogBuilder {
     }
 
     @Override
-    public String toStringRequestOnly(Function<? super HttpHeaders, ? extends HttpHeaders> headersSanitizer,
+    public String toStringRequestOnly(Function<? super HttpHeaders, ?> headersSanitizer,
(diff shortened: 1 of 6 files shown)


trustin:
@imasahiro I think you asked about efficient sanitization of request headers some time ago, right?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1274,7 +1273,7 @@ public String toStringResponseOnly(
         }
 
         if (sanitizedContent != null) {                                       // 17
-            buf.append(", content=").append(contentSanitizer.apply(responseContent));
+            buf.append(", content=").append(sanitizedContent);
```
trustin: Seems like we were sanitizing response content twice by mistake. :sweat_smile:

minwoox: 😱

trustin:
Thanks for reviweing :-)

PR #1886, opened 2019-07-10
Provide a way to set the customized EventLoopScheduler

minwoox:
Motivation:
Currently, connections can be created as many as the number of event loops when using HTTP/2.
A user might want to implement his/her own `EventLoopScheduler` to limit the number of
event loops to handle connections to a certain endpoint.

Modifications:
- Interface `EventLoopScheduler` and add a setter to `ClientFactoryBuilder`
- Add `maxNumEventLoopsPerEndpoint` setters to `ClientFactoryBuilder`
  - A user can restrict the number of event loops to handle connections to `Endpoints` using the default scheduler.
- Extract `EventLoopState` from `DefaultEventLoopScheduler`
  - The implementation of `EventLoopState` is different depending on the number of event loops assigned.
    - `OneEventLoopState` when the `n` is 1.
    - `HeapBasedEventLoopState` for the rest.

Result:
- You can use your own `EventLoopScheduler`.
- You can restrict the number of event loops to handle connections

To-dos:
- CSLB
  - A user can only assign event loops per group, not per endpoint in the group.
- Rebalancing
  - Currently, once event loops are assigned to a certain endpoint, they will handle the connections until the `ClientFactory` is closed.
    This can lead that just a few event loops handle all connections, so we need to rebalance them using `activeRequestCount` or whatsoever.


core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,386 @@
...
+        State(List<EventLoop> eventLoops, int eventLoopSizeLimit, AtomicInteger currentIndex) {
+            this.eventLoops = eventLoops;
+            this.eventLoopSizeLimit = eventLoopSizeLimit;
+            this.currentIndex = currentIndex;
+            if (eventLoops.size() == eventLoopSizeLimit) {
+                init(0);
+            }
+        }
+
+        private void init(int startIndex) {
```
minwoox: Only this logic is added from the previous implementation.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -105,6 +107,10 @@
     private int http1MaxChunkSize = Flags.defaultHttp1MaxChunkSize();
 
     // Armeria-related properties:
+    @Nullable
+    private Function<? super EventLoopGroup, ? extends EventLoopScheduler> eventLoopSchedulerFactory;
+    private int maxEventLoopPerEndpoint;
```
trustin: `maxNumEventLoopsPerEndpoint` ? (`num`, like we have `maxNumConnections` / Use plural for `EventLoop`)

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +138,60 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+    /**
+     * Sets the maximum number of event loops which will be used to handle connections to the specified
+     * {@link Endpoint}.
+     * This method is a shortcut of:
+     * <pre>{@code
+     * String authority = endpoint.authority();
+     * maxEventLoop(authority, maxEventLoop);
+     * }</pre>
+     */
+    public ClientFactoryBuilder maxEventLoop(Endpoint endpoint, int maxEventLoop) {
```
trustin: `maxNumEventLoopsPerEndpoint`? I somehow feel more natural to have `PerEndpoint`..

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +138,60 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+            Function<? super EventLoopGroup, ? extends EventLoopScheduler> eventLoopSchedulerFactory) {
+        checkState(maxEventLoopPerEndpoint == 0 && maxEventLoops.isEmpty(),
+                   "Cannot set eventLoopSchedulerFactory when maxEventLoop per endpoint is specified.");
+        this.eventLoopSchedulerFactory = requireNonNull(eventLoopSchedulerFactory, "eventLoopSchedulerFactory");
+        return this;
+    }
+
+    /**
+     * Sets the maximum number of event loops which will be used to handle connections to all
+     * {@link Endpoint}s except the ones specified by {@link #maxEventLoop(Endpoint, int)} and
```
trustin: and -> or?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +138,60 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+     * maxEventLoop(authority, maxEventLoop);
+     * }</pre>
+     */
+    public ClientFactoryBuilder maxEventLoop(Endpoint endpoint, int maxEventLoop) {
+        requireNonNull(endpoint, "endpoint");
+        return maxEventLoop(endpoint.authority(), maxEventLoop);
+    }
+
+    /**
+     * Sets the maximum number of event loops which will be used to handle connections to the specified
```
trustin: `{@link EventLoop}s` wherever necessary

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,388 @@
...
+
+final class DefaultEventLoopScheduler implements EventLoopScheduler {
+
+    private static final long CLEANUP_INTERVAL_NANOS = Duration.ofMinutes(1).toNanos();
+
+    private final List<EventLoop> eventLoops;
+
+    private final int maxEventLoopPerEndpoint;
+
+    private final Map<String, Integer> maxEventLoops;
```
trustin: Could use `Object2IntMap` and `Object2IntOpenHashMap`

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,388 @@
...
+
+    DefaultEventLoopScheduler(EventLoopGroup eventLoopGroup, int maxEventLoopPerEndpoint,
+                              Map<String, Integer> maxEventLoops) {
+        eventLoops = Streams.stream(eventLoopGroup)
+                            .map(EventLoop.class::cast)
+                            .collect(toImmutableList());
+        checkArgument(maxEventLoopPerEndpoint <= eventLoops.size(),
+                      "maxEventLoopPerEndpoint: %s (expected: <= %s)", maxEventLoopPerEndpoint,
+                      eventLoops.size());
+        if (maxEventLoopPerEndpoint == 0) {
```
trustin: `<=` ?

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    private State state(Endpoint endpoint) {
+        final String authority = endpoint.authority();
+        final State state = states.get(authority);
+        if (state != null) {
+            return state;
+        }
+
+        final Integer maxEventLoop = maxEventLoops.get(authority);
```
trustin: We have to take the default port into account so that `foo.com:80` and `foo.com` are not treated as different endpoints. To fix this:

- Add `SessionProtocol` as an input parameter to `EventLoopScheduler.acquire()` and `ClientFactory.acquireEventLoop()`.
- Handle the cases where:
  - the `Endpoint` or authority string specified in `ClientFactoryBuilder.maxNumEventLoopsPerEndpoint()` has/doesn't have a port number. (If no port number is specified, we should ignore the port number in the `Endpoint` given to `acquire()`.
  - If the `Endpoint` given to `acquire()` does not have a port number, we need to get one from the given `SessionProtocol`.

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,388 @@
...
+        /**
+         * A binary heap of Entry. Ordered by:
+         * <ul>
+         *   <li>{@link Entry#activeRequests()} (lower is better)</li>
+         *   <li>{@link Entry#id()} (lower is better)</li>
+         * </ul>
+         */
+        private final List<Entry> entries = new ArrayList<>();
+        private final List<EventLoop> eventLoops;
+        private final int maxEventLoop;
```
trustin: `maxNumEventLoops`

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,388 @@
...
+
+            if (remove) {
+                i.remove();
+            }
+        }
+
+        lastCleanupTimeNanos = System.nanoTime();
+    }
+
+    private static final class State {
```
trustin: Could we introduce some type hierarchy for this class and provide a special implementation that performs linear search for the case of `maxNumEventLoops <= N` where N is fairly small? A little bit of benchmarking may be necessary to determine N. Consider taking care of this in a follow-up PR. Originally suggested by @anuraaga some time ago.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: Fixed. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: Fixed! :-)

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: Yeah, that sounds more appropriate. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: Fixed. :-)

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
minwoox: Used it. Thanks!

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
minwoox: Fixed. -)

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
minwoox: Will do it in the next PR. (or the next of the next PR 🤣)

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
minwoox: Added `SessionProtocol` to `EventLoopScheduler.acquire()` and change the key of the `states` map from `authority` to `authority` with `isTLS`.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +138,60 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+    public ClientFactoryBuilder maxNumEventLoopsPerEndpoint(Endpoint endpoint, int maxNumEventLoops) {
+        requireNonNull(endpoint, "endpoint");
+        return maxNumEventLoopsPerEndpoint(endpoint.authority(), maxNumEventLoops);
+    }
+
+    /**
+     * Sets the maximum number of {@link EventLoop}s which will be used to handle connections to the specified
+     * {@code authority}.
+     */
+    public ClientFactoryBuilder maxNumEventLoopsPerEndpoint(String authority, int maxNumEventLoops) {
```
minwoox: Could it be better not to provide this API because this takes a string?

trustin: I'm not sure.. Could you tell me why?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +138,60 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+        requireNonNull(authority, "authority");
+        checkState(eventLoopSchedulerFactory == null,
+                   "maxNumEventLoopsPerEndpoint() and eventLoopSchedulerFactory() are mutually exclusive.");
+        checkArgument(maxNumEventLoops > 0,
+                      "maxNumEventLoopsPerEndpoint for '%s': %s (expected: > 0)", authority, maxNumEventLoops);
+        maxNumEventLoopsMap.put(authority, maxNumEventLoops);
```
trustin: - How about checking if `authority` is a valid `Endpoint` expression using `Endpoint.parse()`?
- We could also make `maxNumEventLoopsPerEndpoint(String, int)` a shortcut to `maxNumEventLoopsPerEndpoint(Endpoint.parse(authority), ...)` to avoid double parsing.

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,443 @@
...
+
+    private static final long CLEANUP_INTERVAL_NANOS = Duration.ofMinutes(1).toNanos();
+
+    private final List<EventLoop> eventLoops;
+
+    private final int maxNumEventLoops;
+
+    private final Object2IntMap<String> maxNumEventLoopsMap;
+
+    private final Map<AuthorityWithTLS, State> states = new ConcurrentHashMap<>();
```
trustin: Optimization: Could we split `states` into `tlsStates` and `cleartextStates` and remove the allocation of `AuthorityWithTLS` completely?

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,443 @@
...
+        return acquired;
+    }
+
+    @VisibleForTesting
+    List<Entry> entries(Endpoint endpoint, SessionProtocol sessionProtocol) {
+        return state(endpoint, sessionProtocol).entries();
+    }
+
+    private State state(Endpoint endpoint, SessionProtocol sessionProtocol) {
+        final String authority = endpoint.authority();
```
trustin: Could we normalize the authority by appending the port number if it is same with `SessionProtocol.defaultPort()`?

```java
final String authority;
final String nextTry;
if (endpoint.isGroup()) {
    authority = endpoint.authority();
    nextTry = null;
} else {
    final Endpoint normalizedEndpoint;
    if (endpoint.hasPort()) {
        authority = endpoint.authority();
    } else if (endpoint.isIpAddrOnly() && endpoint.ipFamily() == INET6) {
        authority = '[' + endpoint.host() + "]:" + sessionProtocol.defaultPort();
    } else {
        authority = endpoint.host() + ':' + sessionProtocol.defaultPort();
    }
    nextTry = endpoint.host();
}
```

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,443 @@
...
+    private volatile long lastCleanupTimeNanos = System.nanoTime();
+
+    DefaultEventLoopScheduler(EventLoopGroup eventLoopGroup, int maxNumEventLoops,
+                              Map<String, Integer> maxNumEventLoopsMap) {
+        eventLoops = Streams.stream(eventLoopGroup)
+                            .map(EventLoop.class::cast)
+                            .collect(toImmutableList());
+        checkArgument(maxNumEventLoops <= eventLoops.size(),
+                      "maxNumEventLoopsPerEndpoint: %s (expected: <= %s)", maxNumEventLoops,
+                      eventLoops.size());
```
trustin: How about simply clipping, so that a large value does not fail when deployed into the machines with different number of cores? i.e. `Math.min(maxNumEventLoops, eventLoops.size())`

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,443 @@
...
+        if (maxNumEventLoops <= 0) {
+            this.maxNumEventLoops = eventLoops.size();
```
trustin: What would be the sensible default? `1`? /cc @anuraaga

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -432,21 +445,36 @@ public String authority() {
...
         return this.authority = authority;
     }
 
+    /**
+     * Converts this endpoint into the authority part of a URI removing the port. The return value will be the
+     * same value from {@link #host()} except when the IP address of this endpoint is a valid IP V6 address.
+     * In that case, the value will be {@link #host()} surrounded by {@code []}. If {@link #hasPort()} returns
+     * {@code false}, this value will be the same as {@link #authority()}.
+     */
+    public String authorityWithoutPort() {
```
trustin: With my proposed change, we do not need this method anymore. How about adding `Endpoint.withoutPort()` instead, so that a user can call `Endpoint.withoutPort().authority()`, which is more composable?

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
trustin: Alternatively we could expand `AuthorityWithTLS` into a tuple of (`Endpoint.host()`, `Endpoint.port()`, TLSness) to avoid the cost of string concatenation.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
anuraaga: Alternatively adding `host()` might make sense too, since I think it is the same pattern as the normal Java `URL`

https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#getHost()

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
minwoox: Had a chat with @trustin, and we concluded `1` for `h2`, `h2c`, `HTTP` and `HTTPS` and `eventLoops.size()` for `h1` and `h1c`. :-)

anuraaga: Thanks that sounds even better :)

minwoox: Had a chat with @trustin again, and we decide to use `1` for all session protocols. A user would get confused when he/she sees different behavior when sending requests to "http://a.com" and "h1c://b.com". Please forgive my impatience. 🙇

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
trustin: `host()` is already there actually. :sweat_smile:

anuraaga: Ah should have looked at the code too... Yeah `withoutPort()` seems cleaner then

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,443 @@
...
+    private void cleanup() {
+        if ((++counter & 0xFF) != 0) { // (++counter % 256) != 0
+            return;
+        }
+
+        final long currentTimeNanos = System.nanoTime();
+        if (currentTimeNanos - lastCleanupTimeNanos < CLEANUP_INTERVAL_NANOS) {
+            return;
+        }
+
```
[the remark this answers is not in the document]
minwoox: We need that to prevent multiple threads to enter this section, even though it's okay. :-) Thanks!

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
minwoox: Let me just remove this method because it seems that I don't need this anymore.
Will add later when I need it. Thank!

minwoox:
I think this is ready. PTAL 🙇

core/src/main/java/com/linecorp/armeria/client/UserClient.java
```diff
@@ -140,7 +140,8 @@ protected final O execute(@Nullable EventLoop eventLoop, Endpoint endpoint,
                               I req, BiFunction<ClientRequestContext, Throwable, O> fallback) {
         final DefaultClientRequestContext ctx;
         if (eventLoop == null) {
-            final ReleasableHolder<EventLoop> releasableEventLoop = factory().acquireEventLoop(endpoint);
+            final ReleasableHolder<EventLoop> releasableEventLoop = factory().acquireEventLoop(endpoint,
+                                                                                               sessionProtocol);
```
trustin: How about acquiring an `EventLoop` after endpoint selection occurred, so that event loop is always chosen by a host `Endpoint`? i.e. Modify `DefaultClientRequestContext.init()` to accept a `ClientFactory` and call `acquireEventLoop(...)`.

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,345 @@
...
+                     sessionProtocol);
+    }
+
+    private int maxNumEventLoops(SessionProtocol sessionProtocol) {
+        return isHttp1(sessionProtocol) ? maxNumEventLoopsPerHttp1Endpoint
+                                        : maxNumEventLoopsPerEndpoint;
+    }
+
+    // TODO(minwoox) Use SessionProtocolNegotiationCache and enpoint to bring the protocol version when the
+    //               SessionProtocol is HTTP and HTTPS.
```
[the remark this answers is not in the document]
minwoox: Let me make another PR for this. 😉 
We use `SocketAddress` as a key in `SessionProtocolNegotiationCache ` but I need to use different key because we don't have the `SocketAddress` here. I think that will make this a little bit more complicated, so let me do that in another PR.

core/src/main/java/com/linecorp/armeria/client/EventLoopEntry.java
```diff
@@ -0,0 +1,39 @@
...
+package com.linecorp.armeria.client;
+
+import com.linecorp.armeria.common.util.ReleasableHolder;
+
+import io.netty.channel.EventLoop;
+
+/**
+ * An entry acquired by {@link EventLoopState#acquire()}.
+ */
+interface EventLoopEntry extends ReleasableHolder<EventLoop> {
```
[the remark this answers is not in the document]
minwoox: That's a good idea!

core/src/main/java/com/linecorp/armeria/client/EventLoopState.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import io.netty.channel.EventLoop;
+
+/**
+ * Handles acquiring the {@link EventLoopEntry} for an endpoint.
+ */
+interface EventLoopState {
```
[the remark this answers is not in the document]
minwoox: Yes!

core/src/main/java/com/linecorp/armeria/client/DefaultEventLoopScheduler.java
```diff
@@ -0,0 +1,345 @@
...
+
+    @SuppressWarnings("FieldMayBeFinal")
+    private volatile int acquisitionStartIndex;
+
+    @SuppressWarnings("FieldMayBeFinal")
+    private volatile long lastCleanupTimeNanos = System.nanoTime();
+
+    DefaultEventLoopScheduler(EventLoopGroup eventLoopGroup, int maxNumEventLoopsPerEndpoint,
+                              int maxNumEventLoopsPerHttp1Endpoint,
+                              Map<Endpoint, Integer> maxNumEventLoopsMap) {
```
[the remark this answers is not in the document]
minwoox: Yes let me use that. Thanks!

core/src/main/java/com/linecorp/armeria/client/UserClient.java
minwoox: Yes, I will do that.
As discussed, we do not have to provide a way to set the maximum event loops for a group, will make it take always a host `Endpoint`. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +141,71 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+        checkArgument(maxNumEventLoopsPerEndpoint > 0,
+                      "maxNumEventLoopsPerEndpoint: %s (expected: > 0)", maxNumEventLoopsPerEndpoint);
+        checkState(eventLoopSchedulerFactory == null,
+                   "maxNumEventLoopsPerEndpoint() and eventLoopSchedulerFactory() are mutually exclusive.");
+    }
+
+    /**
+     * Sets the {@link ToIntFunction} which takes an {@link Endpoint} and produces the maximum number of
+     * {@link EventLoop}s which will be used to handle connections to the specified {@link Endpoint}.
+     * The function should return {@code -1} for the {@link Endpoint}s which it doesn't want to handle.
```
trustin: How about allowing 0 and negative values?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +141,71 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+     *         return Integer.MAX_VALUE; // The value will be cut to the size of event loops.
+     *     }
+     *     return -1; // Should return -1 to use the default value.
+     * }
+     * }</pre>
+     */
+    public ClientFactoryBuilder maxNumEventLoopsFunction(ToIntFunction<Endpoint> maxNumEventLoopsFunction) {
+        checkState(eventLoopSchedulerFactory == null,
+                   "maxNumEventLoopsPerEndpoint() and eventLoopSchedulerFactory() are mutually exclusive.");
+        this.maxNumEventLoopsFunction = requireNonNull(maxNumEventLoopsFunction, "maxNumEventLoopsFunction");
```
trustin: How about making this additive? e.g.

```
// This:
cfb.maxNumEventLoopsFunction(e -> {
    if (...) return 3;
    return -1;
});
cfb.maxNumEventLoopsFunction(e -> {
    if (...) return 4;
    return -1;
});
// is effectively same with:
cfg.maxNumEventLoopsFunction(e -> {
    if (...) return 3;
    if (...) return 4;
    return -1;
});
```

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -426,6 +494,13 @@ static String toString(
               .add("useHttp2Preface", useHttp2Preface)
               .add("useHttp1Pipelining", useHttp1Pipelining);
 
+        if (maxNumEventLoopsPerHttp1Endpoint > 0) {
+            helper.add("maxNumEventLoopsPerHttp1Endpoint", maxNumEventLoopsPerHttp1Endpoint);
+        } else if (maxNumEventLoopsPerEndpoint > 0) {
+            helper.add("maxNumEventLoopsPerEndpoint", maxNumEventLoopsPerEndpoint);
+        }
+        helper.add("maxNumEventLoopsFunction", maxNumEventLoopsFunction);
```
trustin: Perhaps we should add this only when the function is not the default?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -381,40 +464,25 @@ public ClientFactory build() {
...
-            long http2MaxHeaderListSize, int http1MaxInitialLineLength, int http1MaxHeaderSize,
-            int http1MaxChunkSize, long idleTimeoutMillis, boolean useHttp2Preface, boolean useHttp1Pipelining,
-            ConnectionPoolListener connectionPoolListener,
-            MeterRegistry meterRegistry) {
-
-        final ToStringHelper helper = MoreObjects.toStringHelper(self).omitNullValues();
+        final ToStringHelper helper = MoreObjects.toStringHelper(this).omitNullValues();
         helper.add("workerGroup", workerGroup + " (shutdownOnClose=" + shutdownWorkerGroupOnClose + ')')
-              .add("socketOptions", socketOptions)
+              .add("eventLoopSchedulerFactory", eventLoopSchedulerFactory)
```
trustin: Perhaps we should add this only when the factory is not the default?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -227,6 +269,9 @@ protected Channel channel() {
 
     @Override
     public EventLoop eventLoop() {
+        if (eventLoop == null) {
+            throw new IllegalStateException("Should call init(endpoint) before invoking this method.");
```
trustin: nit: Could use `checkState()`

core/src/test/java/com/linecorp/armeria/client/MaxNumEventLoopsPerEndpointTest.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /**
+     * eventLoops idx:   [   0,   1,   2,   3,   4,   5,   6]
+     *    - endpointA:       A    A
+     *    - endpointB:                 B    B    B
+     *    - endpointC:       C    C    C              C    C
+     *
+     * <P>The event loop group for endpoints is assigned sequentially. The endpointC needs 5 event loops so it
+     * takes the 3 event loops from the first 3 elements in the list.
```
trustin: The endpointC starts at 5 requiring 5 event loops, so it wraps around at the end of the array and takes the first 3.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: Yeah, the current implementation is actually allowing that. I will update the doc. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: Yes, it was nullable and I changed it to use the default function and I forget about this. 😄

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: `eventLoopSchedulerFactory` is nullable and we check the mutual exclusion in the builder, so I thought this was OK.
But let me check the `null` for clarity. 😉

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
minwoox: Will use it, thanks!

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +141,71 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+     * Sets the {@link ToIntFunction} which takes an {@link Endpoint} and produces the maximum number of
+     * {@link EventLoop}s which will be used to handle connections to the specified {@link Endpoint}.
+     * The function should return {@code -1} for the {@link Endpoint}s which it doesn't want to handle.
+     * For example: <pre>{@code
+     * ToIntFunction<Endpoint> function = endpoint -> {
+     *     if (endpoint.equals(Endpoint.of("foo.com"))) {
+     *         return 5;
+     *     }
+     *     if (endpoint.host().contains("bar.com")) {
+     *         return Integer.MAX_VALUE; // The value will be cut to the size of event loops.
```
trustin: The value will be clamped at the number of event loops.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: Fixed PTAL 🙇

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: Fixed PTAL 🙇

core/src/main/java/com/linecorp/armeria/client/AbstractEventLoopState.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.List;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import io.netty.channel.EventLoop;
+
+abstract class AbstractEventLoopState {
+
+    static AbstractEventLoopState of(List<EventLoop> eventLoops, int maxNumEventLoops,
```
ikhoon: nit: How about making `interface EventLoopState` and extract common method for this, because a factory method which returns `abstract class` is a little strange for me 🙃

core/src/main/java/com/linecorp/armeria/client/AbstractEventLoopEntry.java
```diff
@@ -0,0 +1,54 @@
...
+    public EventLoop get() {
+        return eventLoop;
+    }
+
+    @Override
+    public void release() {
+        parent.release(this);
+    }
+
+    abstract int activeRequests();
```
kojilin: Curious are these/below methods needed to treat as abstract methods? I think it's not invoked outside of state ?

core/src/main/java/com/linecorp/armeria/client/AbstractEventLoopState.java
minwoox: https://github.com/line/armeria/pull/1886/commits/ae50467882dd6d568b3a4ece242eedde2fdb952a#diff-603fc6081c9252d8fcfee38b3fae0719R30
That was the implementation that but as @trustin said, this is package-private and it will make it complicated by introducing another interface, so just removed it. 😄

ikhoon: Sorry, I should've caught review histories. 😅

core/src/main/java/com/linecorp/armeria/client/AbstractEventLoopEntry.java
minwoox: `decrementActiveRequests` and `index` are used in another [class](https://github.com/line/armeria/pull/1886/files/0e69530c861059c85addd0dabf2eed163bed6012#diff-3ca01fccde269ac03030409d2526aa77R102)
and `activeRequests` and `id` are used in the test cases.
We can hide `incrementActiveRequests` and `setIndex` but I thought exposing `decrementActiveRequests` and hiding `incrementActiveRequests` is kinna weird.
So there's only `setIndex` left. To hide it I had to change `List<AbstractEventLoopEntry> entries` to `List<? extends AbstractEventLoopEntry> entries` and changed related things.
But this is package-private so I thought it's not worth it. 😄

trustin: If some methods are only used in tests, let's annotate if `@VisibleForTesting`.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +139,71 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+        checkArgument(maxNumEventLoopsPerEndpoint > 0,
+                      "maxNumEventLoopsPerEndpoint: %s (expected: > 0)", maxNumEventLoopsPerEndpoint);
+        checkState(eventLoopSchedulerFactory == null,
+                   "maxNumEventLoopsPerEndpoint() and eventLoopSchedulerFactory() are mutually exclusive.");
+    }
+
+    /**
+     * Sets the {@link ToIntFunction} which takes an {@link Endpoint} and produces the maximum number of
+     * {@link EventLoop}s which will be used to handle connections to the specified {@link Endpoint}.
+     * The function should return {@code 0} or negative values for the {@link Endpoint}s which it
```
trustin: negative values -> a negative value

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -132,6 +139,71 @@ public ClientFactoryBuilder workerGroup(EventLoopGroup workerGroup, boolean shut
...
+     * doesn't want to handle. For example: <pre>{@code
+     * ToIntFunction<Endpoint> function = endpoint -> {
+     *     if (endpoint.equals(Endpoint.of("foo.com"))) {
+     *         return 5;
+     *     }
+     *     if (endpoint.host().contains("bar.com")) {
+     *         return Integer.MAX_VALUE; // The value will be clamped at the number of event loops.
+     *     }
+     *     return -1; // Should return 0 or negative values to use the default value.
```
trustin: negative values -> a negative value

minwoox: Fixed :-)

core/src/main/java/com/linecorp/armeria/client/AbstractEventLoopEntry.java
minwoox: > If some methods are only used in tests, let's annotate if @VisibleForTesting.

They are used in `AbstractEventLoopState` as well :)
I was talking about when I Changed to using `List<? extends AbstractEventLoopEntry> entries` in `AbstractEventLoopState` and adding `Entry` to the list. In that case, I have to annotate them with `@VisibleForTesting`. But I decided to add `AbstractEventLoopEntry` to the list, so I don't have to. :-)

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -143,8 +173,20 @@ public boolean init(Endpoint endpoint) {
                 runThreadLocalContextCustomizer();
             }
 
+            if (eventLoop == null) {
+                assert factory != null;
+                final ReleasableHolder<EventLoop> releasableEventLoop =
+                        factory.acquireEventLoop(this.endpoint, sessionProtocol());
+                eventLoop = releasableEventLoop.get();
+                log.addListener(log -> releasableEventLoop.release(), RequestLogAvailability.COMPLETE);
```
ikhoon: nit: `log.addListener(log -> ...)` -> `log.addListener(ignored -> ...)` to avoid duplication?

minwoox: Fixed. 😄

minwoox:
Thanks for reviewing!

PR #1889, opened 2019-07-10
Enable message compression on server by default.

### LINKED ISSUE #1885: gRPC server should default to message compression enabled.
I seem to have missed the spec, server should allow message compression by default if the server requests it.

anuraaga:
This matches the default as documented in `ServerCall.setMessageCompression`. I personally don't like the default very much though... I think we would generally want to encourage using stream encoding via `HttpEncodingService` / `HttpDecodingClient` instead of message compression, which will have much better performance.

Fixes #1885


trustin:
Test failure unrelated with this pull request.

minwoox:
> Test failure unrelated with this pull request.

I'm working on it. :-)

trustin:
Thanks, @anuraaga!

PR #1896, opened 2019-07-11
Use OPTIONS instead of HEAD for HTTP/2 upgrade.

### LINKED ISSUE #1895: serviceUnder("/") and useHttp2Preface(false) cause service to be executed twice
Found a strange corner case, it seems that when mounting a service using `serviceUnder("/")` and disabling HTTP/2 preface on a client, the service logic gets executed twice. This is the exact combination that causes the issue - mounting to a different path e.g. `serviceUnder("/foo")` does not have the problem and enabling HTTP/2 preface also doesn't have the problem.

Depending on the logic of the service, being executed twice for a request can be quite problematic.

anuraaga:
This allows the upgrade to happen without user business logic having to worry about it.

Fixes #1895

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
@@ -394,12 +394,12 @@ final class HttpClientPipelineConfigurator extends ChannelDuplexHandler {
         }
 
         /**
-         * Sends the initial upgrade request, which is {@code "HEAD / HTTP/1.1"}.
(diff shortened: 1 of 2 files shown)


minwoox:
As @trusin said, if we changed https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java#L402
to `new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.OPTIONS, "*");`, 
it works.

anuraaga:
Thanks for the tip!

Issue #1084, opened 2018-03-12
EndpointGroup design issues


trustin:
It seems like some users have the following concerns with the current `EndpointGroup` and `EndpointGroupRegistry` API:

- Using a 'string' as a way to specify a target endpoint group makes it hard to build a clear dependency between an `EndpointGroup` and a `Client`.
- There's no clear unified contract for the life cycle of an `EndpointGroup`, which complicates the user application shutdown process.

anuraaga:
While digging into endpoint groups, I came to a feeling that `EndpointGroup` should be tied to `ClientFactory` instead of a static registry. The main reason for this is `HttpHealthCheckedEndpointGroup` is tied to a `ClientFactory` - it doesn't seem to make sense for this to be accessible from different factories through the static registry and is hard for me to reason about. Also, using custom `ClientFactory` is somewhat advanced - I think these users would generally expect registering endpoints to their custom, awesome `ClientFactory`. Normal usage could still use a static method to register to the default `ClientFactory`.

trustin:
One advantage of keeping `EndpointGroup`s in a `ClientFactory` is that they can be closed automatically when their `ClientFactory` is closed.

anuraaga:
In https://github.com/openzipkin/zipkin/issues/2699 I found that even after configuring a http health checked endpoint group with our customized client factory, I was still missing the fact that `DnsEndpointGroup` should be using that factory's event loop group, not common pools, to keep everything together. So not just `HttpHealthCheckedEndpointGroup`, but also `DnsEndpointGroup` and any other type of endpoint group that uses armeria / netty inside it will be easier to set up in a consistent way if a client factory is involved in initialization.

trustin:
@anuraaga wrote:

Realized this idea was proposed in the issue already but didn't think enough until now - I'm not sure this is all that much better since it means creating the client before registering the endpoint group.

```java
HttpClient client = HttpClient.of("https://group:foobar/");
HealthCheckedEndpointGroup foobar = HealthCheckedEndpointGroup.builder(...)
  .options(client.options())
  .build();
EndpointGroupRegistry.register("foobar", foobar, ROUND_ROBIN);
```
Creating an `HttpClient` that references a group that doesn't exist yet seems quite awkward. There's a bit less code but semantically this doesn't seem quite right to me.

Though as mentioned in the issue, it seems hard to solve in a great way so maybe this is still the best we can do.

[...]

PR #1899, opened 2019-07-12
Cancel health check when closing health checked endpoint

anuraaga:
I ran into this while debugging why `awaitInitialEndpoints` in https://github.com/openzipkin/zipkin/pull/2653 often takes seconds to resolve `localhost`. I don't think I found the real culprit yet, but it does look like when closing and re-registering a health checked endpoint group, health checks from the closed one will still continue to be sent forever.

Also wondering if we need to redesign it so `EndpointGroupRegistry` is a field of `ClientFactory` instead of `static`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/OrElseEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/OrElseEndpointGroup.java
@@ -45,7 +45,8 @@ final class OrElseEndpointGroup extends AbstractListenable<List<Endpoint>> imple
 
     @Override
     public void close() {
-        try (EndpointGroup first = this.first; EndpointGroup second = this.second) {
+        try (EndpointGroup first = this.first;
+             EndpointGroup second = this.second) {
             // Just want to ensure closure.
         }
     }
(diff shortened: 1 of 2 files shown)


minwoox:
> Also wondering if we need to redesign it so EndpointGroupRegistry is a field of ClientFactory instead of static.

That's a good idea. It was one of the hardest one to understand when I first looked into Armeria and I think it's unintuitive.

trustin:
Related issue: #1084

trustin:
Not sure if it's a good idea to move the registry under `ClientFactory`, but it's worth exploring.

core/src/main/java/com/linecorp/armeria/client/endpoint/OrElseEndpointGroup.java
```diff
@@ -42,4 +42,10 @@
         }
         return second.endpoints();
     }
+
+    @Override
+    public void close() {
+        first.close();
+        second.close();
```
trustin: Probably better calling `second.close()` in `finally {}`?

anuraaga: Used try / resources

core/src/main/java/com/linecorp/armeria/client/endpoint/OrElseEndpointGroup.java
```diff
@@ -42,4 +42,11 @@
         }
         return second.endpoints();
     }
+
+    @Override
+    public void close() {
+        try (EndpointGroup first = this.first; EndpointGroup second = this.second) {
```
minwoox: Could you fix the check style failure?

trustin:
Thanks, @anuraaga :+1:

PR #1901, opened 2019-07-15
Update jmh-gradle-plugin version to 0.5.0-rc-1

minwoox:
Motivation:
https://github.com/melix/jmh-gradle-plugin says that the minimal plugin version is `0.5.0` when using Gradle `5.5`.


trustin:
Thanks, @minwoox !

PR #1904, opened 2019-07-15
Fix the uri of ArmeriaServerHttpRequest to contain the base url

minwoox:
Motivation:
The uri of `ServerHttpRequest` in Spring Web flux contains the base URL.
Also, a negative value of "max-age" from "set-cookie" means no "max-age" attribute, we shouldn't set it.

Modification:
- Add the base URL when `ArmeriaServerHttpRequest` is created.
- Do not set max-age from set-cookie when the value is negative

Result:
- Spring integration

### DIFF
--- a/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpRequest.java
+++ b/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpRequest.java
@@ -56,9 +56,7 @@ final class ArmeriaServerHttpRequest extends AbstractServerHttpRequest {
     ArmeriaServerHttpRequest(ServiceRequestContext ctx,
                              HttpRequest req,
                              DataBufferFactoryWrapper<?> factoryWrapper) {
-        super(URI.create(requireNonNull(req, "req").path()),
-              null,
-              fromArmeriaHttpHeaders(req.headers()));
+        super(uri(req), null, fromArmeriaHttpHeaders(req.headers()));
         this.ctx = requireNonNull(ctx, "ctx");
         this.req = req;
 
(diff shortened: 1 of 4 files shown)


spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpRequestTest.java
```diff
@@ -39,20 +39,35 @@
...
-                HttpRequest.of(RequestHeaders.of(HttpMethod.POST, "/"),
+                HttpRequest.of(RequestHeaders.builder(HttpMethod.POST, "/")
+                                             .scheme("http")
+                                             .authority("127.0.0.1")
+                                             .build(),
                                Flux.just("a", "b", "c", "d", "e").map(HttpData::ofUtf8));
 
         final ServiceRequestContext ctx = newRequestContext(httpRequest);
         final ArmeriaServerHttpRequest req = request(ctx);
+        System.err.println(req.getURI());
```
trustin: Use a logger or remove it.

trustin:
Special thanks to @jinsankim for reporting this issue and providing useful hints and sample application. :heart:

spring/boot-webflux-autoconfigure/src/test/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpRequestTest.java
minwoox: Oops~ 😱
Thanks!

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpRequest.java
```diff
@@ -68,6 +66,15 @@
                    .publishOn(Schedulers.fromExecutor(ctx.contextAwareExecutor()));
     }
 
+    private static URI uri(HttpRequest req) {
+        final String scheme = req.scheme();
+        final String authority = req.authority();
+        // Server side Armeria HTTP request always has the scheme and authority.
+        assert scheme != null;
+        assert authority != null;
+        return URI.create(scheme + "://" + authority + req.path());
```
ikhoon: How about using [URI constructor](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#URI-java.lang.String-java.lang.String-java.lang.String-java.lang.String-)?
```java
return new URI(scheme, authority, path, null);
```

minwoox: As we discussed, this will ignore the fragment part, so we should use `URI.create()`. Thanks!

minwoox:
Thanks, @trustin and @ikhoon!

Issue #1223, opened 2018-05-23
Consider changing the http tracing code to use Brave's HttpTracing helper


adriancole:
Currently, the tagging policy of Armeria is framework-specific. This is not a problem in "pure armeria" applications per-se, but on the other hand, it doesn't leverage the abstraction work in Brave with regards to span lifecycle and data policy.

Most notably, using the `brave-instrumentation-http` library allow users to plugin standard sampling and tagging policy types. This is particularly helpful when sending data to strict systems such as Amazon X-Ray or Google Stackdriver, as you can ensure the data they need to read makes it.

https://github.com/openzipkin/brave/tree/master/instrumentation/http

Moreover, we have a standard test which verifies some edge cases, and we'd get more coverage for free this way. Here's something similar, for example, how vertx-web is written and tested:

https://github.com/openzipkin/brave/blob/master/instrumentation/vertx-web/src/main/java/brave/vertx/web/TracingRoutingContextHandler.java#L58
https://github.com/openzipkin/brave/blob/master/instrumentation/vertx-web/src/test/java/brave/vertx/web/ITVertxWebTracing.java

I could help a bit with this, but definitely would be awesome if someone else would be interested in doing it.

anuraaga:
I remember looking at that code and avoided it for some reason - can't remember what that reason was though, it might not have been a good one.

adriancole:
PS I'm polishing our self-tracing now and noticed this again.. it would be really useful to have the http pattern in the endpoints. For example, I can guess the root span was a "get /zipkin/traces" endpoint but others might not.

<img width="1024" alt="Screen Shot 2019-04-22 at 9 49 24 AM" src="https://user-images.githubusercontent.com/64215/56480429-f9f6a700-64f4-11e9-84af-a2439bd2d289.png">

here's another related change which might help https://github.com/reactor/reactor-netty/pull/622

adriancole:
one impact of not doing this is it discourages switching because sampling policy is lost. For example, the HttpTracing module includes a server and client sampler object used to mute paths (ex static assets) from being traced, or to avoid originating traces from ad-hoc client use. In zipkin, for example, our self-tracing is going to overrun with low value assets until we add an after-the-fact handler to drop traces related to that. It would be nicer to have a means before-the-fact (normal sampling)

PR #1877, opened 2019-07-03
Add Zipkin integration document

minwoox:

PR #1906, opened 2019-07-16
Changing the http tracing code to use Brave's HttpTracing helper

kojilin:
Try to solve https://github.com/line/armeria/issues/1223 .

Modification:
- Switching to HttpTracing.
- Create armeria's brave http parser/adapter.
- Add ArmeriaHttpTracingBuilder for user.
- Propagate scope even it's no-op.

Need to consider and TODO in this PR:
- ~We can't use RequestLog's timestamp for start and finish.~
- Span is created when calling handleSend/Receive, at that time
 we don't have scheme.
- Basically all use RequestLog, it's easier to get req/res related tag
 value.
- Test for Builder.
- JavaDoc for adapter and parser.

TODO
- [x] http-tests-server
- [x] http-tests-client
- [x] Tests for parser and adapter


kojilin:
cc @adriancole

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.common.logging.RequestLog;
+import com.linecorp.armeria.internal.brave.SpanTags;
+
+import brave.http.HttpClientAdapter;
+
+public class ArmeriaHttpClientAdapter extends HttpClientAdapter<RequestLog, RequestLog> {
+    @Override
+    public String path(RequestLog requestLog) {
```
kojilin: Due to we can't get url when calling send/receive.

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
```diff
@@ -0,0 +1,45 @@
...
+import brave.http.HttpClientParser;
+
+public class ArmeriaHttpClientParser extends HttpClientParser {
+    @Override
+    public <T> void request(HttpAdapter<T, ?> adapter, T req, SpanCustomizer customizer) {
+        super.request(adapter, req, customizer);
+        if (req instanceof RequestLog) {
+            ((RequestLog) req).addListener(log -> {
+                customizer.tag("http.url", adapter.url(req));
+            }, RequestLogAvailability.SCHEME);
```
kojilin: scheme is not available when calling request.

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -57,109 +50,65 @@
 
     private static final Logger logger = LoggerFactory.getLogger(BraveClient.class);
 
-    // TODO(minwoox) Add the variant which takes HttpTracing.
-
-    /**
-     * Creates a new tracing {@link Client} decorator using the specified {@link Tracing} instance.
-     */
-    public static Function<Client<HttpRequest, HttpResponse>, BraveClient> newDecorator(Tracing tracing) {
```
kojilin: Just think if it's better to provide new implementation instead of providing both.

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -57,109 +50,65 @@
...
-
-        // Inject the headers.
         final RequestHeadersBuilder newHeaders = req.headers().toBuilder();
-        injector.inject(span.context(), newHeaders);
+        final Span span = handler.handleSend(injector, newHeaders, ctx.log());
         req = HttpRequest.of(req, newHeaders.build());
         ctx.updateRequest(req);
 
+        // Ensure the trace context propagates to children
+        ctx.onChild(TraceContextUtil::copy);
```
kojilin: even it's no-op?

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceTest.java
```diff
@@ -187,7 +193,6 @@ private static void assertTags(Span span) {
         assertThat(span.tags()).containsEntry("http.host", "foo.com")
                                .containsEntry("http.method", "POST")
                                .containsEntry("http.path", "/hello/trustin")
-                               .containsEntry("http.status_code", "200")
```
kojilin: success call won't have status_code tag.

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
```diff
@@ -23,80 +23,73 @@
...
-        final Span span = contextOrFlags.context() != null ? tracer.joinSpan(contextOrFlags.context())
-                                                           : tracer.nextSpan(contextOrFlags);
-        // For no-op spans, nothing special to do.
-        if (span.isNoop()) {
-            return delegate().serve(ctx, req);
-        }
-
-        final String method = ctx.method().name();
-        span.kind(Kind.SERVER).name(method);
-        ctx.log().addListener(log -> SpanContextUtil.startSpan(span, log),
```
kojilin: `span.start(log.requestStartTimeMicros())` become `span.start()` in `handleReceive`

brave/build.gradle
```diff
@@ -1,5 +1,6 @@
 dependencies {
     compile 'io.zipkin.brave:brave'
+    compile 'io.zipkin.brave:brave-instrumentation-http'
```
adriancole: there is also http-tests which includes the IT which can be used to validate behavior

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
adriancole: maybe add comment saying same

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    @Nullable
+    public String requestHeader(RequestLog requestLog, String name) {
+        return requestLog.requestHeaders().get(name);
+    }
+
+    @Override
+    public Integer statusCode(RequestLog requestLog) {
+        return requestLog.status().code();
```
adriancole: do this instead, assuming code is integer

```java
  @Override @Nullable public Integer statusCode(RequestLog requestLog) {
    int result = statusCodeAsInt(requestLog);
    return result != 0 ? result : null;
  }

  @Override public int statusCodeAsInt(RequestLog requestLog) {
    Status status = requestLog.status(); // adrian doesn't know the actual type here and neither if it can be null
    return status != null ? status.code() : 0;
  }
```

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
adriancole: parser is supplied by end user.. I assume this is a default implementation to match old usage? if so comment

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
adriancole: yes to propagate the unsampled decision and consistent trace ID in log sync

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -57,109 +50,65 @@
...
-            port = socketAddress.getPort();
-        } else {
-            address = null;
-            port = 0;
-        }
-        if (!isNullOrEmpty(remoteServiceName)) {
-            span.remoteServiceName(remoteServiceName);
-        }
-        if (address != null) {
-            span.remoteIpAndPort(address.getHostAddress(), port);
```
adriancole: still need this part as soon as you have resolved IP/port

brave/src/main/java/com/linecorp/armeria/common/brave/ArmeriaHttpTracingBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+ * A helper class for creating a new {@link HttpTracing} instance.
+ * For example,
+ *
+ * <pre>{@code
+ * HttpTracing httpTracing = new ArmeriaHttpTracingBuilder(tracing)
+ *     .build();
+ * }</pre>
+ *
+ */
+public class ArmeriaHttpTracingBuilder {
```
adriancole: maybe we need to chat more about this, I don't think it will work if someone needs to provide an armeria specific tracing object, unless you are just trying to match content. For concerns like tags not being available at a specific point in time, we can play a different game.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    @Nullable
+    public String requestHeader(RequestLog requestLog, String name) {
+        return requestLog.requestHeaders().get(name);
+    }
+
+    @Override
+    public Integer statusCode(RequestLog requestLog) {
+        return requestLog.status().code();
```
adriancole: same thing about status code. also consider if you can provide the httpRoute here.

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
adriancole: you can overwrite the start timestamp, but immediately invoking start.timestamp(newOne)

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -57,109 +50,65 @@
...
-    }
-
     /**
-     * Creates a new tracing {@link Client} decorator using the specified {@link Tracing} instance
-     * and remote service name.
+     * Creates a new tracing {@link Client} decorator using the specified {@link HttpTracing} instance.
      */
     public static Function<Client<HttpRequest, HttpResponse>, BraveClient> newDecorator(
-            Tracing tracing, @Nullable String remoteServiceName) {
+            HttpTracing httpTracing) {
```
anuraaga: I think we can create a default `HttpTracing` from `Tracing` and `String` - even if we don't want that to be part of the public API going forward it would be good to do so with deprecation at least.

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
kojilin: I decide to make all annotation using brave's clock, WDYT? @anuraaga .

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
kojilin: Due to armeria has unknown state,  I check unknown one.

kojilin:
About Route, currently RequestLog can't get routing information, so I think it's better to do in another PR?

adriancole:
ok by me

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
anuraaga: I think we lose timing by using Brave's clock since reasonable time passes from starting the request to calling the decorator (which may be amplified depending on the number and order of decorators). It'd be nice if we can use the more precise Armeria timings IMO.

Start time seems fine to use the overwriting method, and chatting with @adriancole offline he thinks it's reasonable to just copy in the code for `handleFinish` and use our timestamp when finishing since it's a small amount of code. What do you think?

trustin: +1. We hope Brave adds the variants of `start()` and `finish()` that accept the timestamp given by a caller in the future.

kojilin: not sure am i understand correctly and for confirmation.
We implement our own `handleFinish` and not calling handleSend or handleReceive, right?

adriancole: start timestamp easy, just call span.timestamp later.

what @anuraaga means is copy/paste logic inside handleReceive here for now, just use the external timestamp you have (span.finish(thatTime)) vs implicit

kojilin: updated, so the change is correct?

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    @Nullable
+    public String requestHeader(RequestLog requestLog, String name) {
+        return requestLog.requestHeaders().get(name);
+    }
+
+    @Override
+    @Nullable
+    public Integer statusCode(RequestLog requestLog) {
```
adriancole: better to override the integer methods similar to how I said earlier as prevents boxing. in case of UNKNOWN return 0.

adriancole: primitive methods I mean..

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+    public String serializationFormat(RequestLog requestLog) {
+        final SerializationFormat serFmt = requestLog.scheme().serializationFormat();
+        return serFmt == SerializationFormat.NONE ? null : serFmt.uriText();
+    }
+
+    /**
+     * Returns the remote address.
+     */
+    @Nullable
+    public String remoteAddress(RequestLog requestLog) {
```
adriancole: do `HttpServerAdapter.parseClientIpAndPort` instead. not sure if this method is important enough to expose as in HttpClientParser someone can do this manually...

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
```diff
@@ -0,0 +1,73 @@
...
+            customizer.tag("http.host", adapter.authority(requestLog))
+                      .tag("http.url", adapter.url(requestLog))
+                      .tag("http.protocol", adapter.protocol(requestLog));
+
+            final String serFmt = adapter.serializationFormat(requestLog);
+            if (serFmt != null) {
+                customizer.tag("http.serfmt", serFmt);
+            }
+
+            final String raddr = adapter.remoteAddress(requestLog);
```
adriancole: I think better to do remote and local address here manually instead of exposing in the adapter. reason is that it is not usually a good idea to repeat endpoint info already in the span, and exposing methods will encourage that.

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -68,81 +69,90 @@
...
             return delegate().execute(ctx, req);
         }
     }
 
-    private void finishSpan(Span span, RequestLog log) {
-        setRemoteEndpoint(span, log);
-        SpanContextUtil.closeSpan(span, log);
-    }
-
     private void setRemoteEndpoint(Span span, RequestLog log) {
```
adriancole: sorry I told you wrong, if you implement `adapter.parseClientIpAndPort(request, span)` it is done automatically

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -155,11 +165,31 @@ private void setRemoteEndpoint(Span span, RequestLog log) {
...
-        }
         if (address != null) {
             span.remoteIpAndPort(address.getHostAddress(), port);
         }
     }
+
+    /**
+     * Copy from brave.http.HttpHandler#handleFinish(Object, Throwable, Span)
+     */
+    private void handleFinish(RequestLog requestLog, Span span) {
```
adriancole: Maybe cite this issue in comment https://github.com/openzipkin/brave/issues/946

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+    @Nullable
+    public String requestHeader(RequestLog requestLog, String name) {
+        return requestLog.requestHeaders().get(name);
+    }
+
+    @Override
+    @Nullable
+    public Integer statusCode(RequestLog requestLog) {
+        final HttpStatus status = requestLog.status();
+        if (status == HttpStatus.UNKNOWN) {
```
adriancole: same comments about status and address methods here

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientTest.java
```diff
@@ -71,14 +72,15 @@ void tearDown() {
 
     @Test
     void newDecorator_shouldWorkWhenRequestContextCurrentTraceContextNotConfigured() {
-        BraveClient.newDecorator(Tracing.newBuilder().build());
+        BraveClient.newDecorator(HttpTracing.create(Tracing.newBuilder().build()));
     }
 
     @Test
```
adriancole: add test that duration is correct from request log

dependencies.yml
```diff
@@ -214,10 +214,11 @@ io.reactivex.rxjava2:
 
 io.zipkin.brave:
   brave:
-    version: '5.6.6'
+    version: &BRAVE_VERSION '5.6.8'
```
adriancole: we now have 5.6.9 :)

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
kojilin: you mean parseServerIpAndPort?

adriancole: no I mean implement in the client adapter class parseClientIpAndPort() then you do nothing here. if unclear I can contribute to this pull request if you grant me access to your fork

adriancole: oops yeah this is client! sorry I forget both client and server are here. you are right to do it explicitly. my mistake

kojilin: ah you mean server side will handle this.

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientTest.java
```diff
@@ -110,6 +111,10 @@ void shouldSubmitSpanWhenSampled() throws Exception {
 
         assertThat(span.traceId().length()).isEqualTo(16);
 
+        // check duration is correct from request log
```
adriancole: nice

brave/src/test/java/com/linecorp/armeria/it/brave/ArmeriaHttpTracingIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+        assertThat(span.tags())
+                .containsEntry("http.url", url(uri))
+                .containsEntry("context.visible", "true")
+                .containsEntry("request_customizer.is_span", "false")
+                .containsEntry("response_customizer.is_span", "false");
+    }
+
+    @Override
+    public void notFound() {
+        throw new AssumptionViolatedException("Armeria 404 not supported.");
```
adriancole: please cite TODO of the issue about this as 404 should be traceable I think

trustin: There's a limitation in Armeria that you cannot decorate a non-existing path mapping. You can bind a service that returns 404 at / and decorate it as a workaround, though.

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
```diff
@@ -0,0 +1,224 @@
...
+        new DummyRequestContext().makeContextAware(() -> {
+            super.propagatesExtra_unsampledTrace();
+            return null;
+        }).call();
+    }
+
+    @Override
+    @Test
+    public void addsErrorTagOnTransportException() throws Exception {
+        throw new AssumptionViolatedException("Armeria won't record exception in this case.");
```
kojilin: @trustin curious if this is expected behavior.
Client got closed session exception, but we can't get it from RequestLog. Because it will call `onWrapperCompleted`. https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/Http1ResponseDecoder.java#L92

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    @Override
+    @Test
+    public void redirect() throws Exception {
+        throw new AssumptionViolatedException("Armeria does not support client redirect.");
```
kojilin: Am I correct?

trustin: That's correct.

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
trustin: I'm not 100% sure but I guess it's not an expected behavior. Do we have the same problem for HTTP/2? We could investigate this later though because it doesn't seem to belong to this pull request. Let's create a new issue for a follow-up.

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceIntegrationTest.java
```diff
@@ -126,8 +126,9 @@ public void supportsPortableCustomization() throws Exception {
     }
 
     @Override
+    @Test
     public void notFound() {
-        throw new AssumptionViolatedException("Armeria 404 not supported.");
+        throw new AssumptionViolatedException("Armeria cannot decorate a non-existing path mapping.");
```
trustin: non-existent

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
```diff
@@ -0,0 +1,75 @@
...
+ * User can extend this class or implement own {@link HttpClientParser}.
+ */
+public class ArmeriaHttpClientParser extends HttpClientParser {
+    @Override
+    public <T> void response(HttpAdapter<?, T> rawAdapter, T res, Throwable error, SpanCustomizer customizer) {
+        super.response(rawAdapter, res, error, customizer);
+        if (res instanceof RequestLog && rawAdapter instanceof ArmeriaHttpClientAdapter) {
+            final RequestLog requestLog = (RequestLog) res;
+            final ArmeriaHttpClientAdapter adapter = (ArmeriaHttpClientAdapter) rawAdapter;
+            customizer.tag("http.host", adapter.authority(requestLog))
```
ikhoon: nit: Custom tags like `http.xxx` are used multiple times in this PR. How about centralizing them with enum or static value. 😀

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
kojilin: I switch to h2c in this PR and http/2 doesn't have this issue.

kojilin: https://github.com/line/armeria/issues/1914

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public String method(RequestLog requestLog) {
+        return requestLog.method().name();
+    }
+
+    @Override
+    public String url(RequestLog requestLog) {
+        return SpanTags.generateUrl(requestLog);
```
kojilin: better to check SCHEME availability?

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
adriancole: I would probably either make a base class and a top-level for each protocol to test, or a parameterized runner or similar as the behavior is different and important.

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
trustin: Both `SCHEME` and `REQUEST_HEADERS`. What's the expected behavior of this method when it's called too early (Just curious)?

adriancole: all methods return null if not available (in case of the special cased integer responseCode zero)

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -68,98 +66,116 @@
...
+    }
+
+    /**
+     * Creates a new tracing {@link Client} decorator using the specified {@link HttpTracing} instance.
+     */
+    public static Function<Client<HttpRequest, HttpResponse>, BraveClient> newDecorator(
+            HttpTracing httpTracing) {
         try {
-            ensureScopeUsesRequestContext(tracing);
+            ensureScopeUsesRequestContext(httpTracing.tracing());
```
anuraaga: Should we have `httpTracing.toBuilder().clientParser(new ArmeriaHttpClientParser())`?

adriancole: that would overwrite what the user supplied so I think no

adriancole: actually the line mention isnt using http stuff anyway so it shouldn't matter I guess

adriancole:
today I will try this in zipkin itself using jitpack to make a build off this branch (assuming jitpack works on this repo)

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
anuraaga: I guess we'd need to check if the user passed anything first. I didn't mean to specifically point to this line - but if the user didn't set client parser correctly, I think the tracing wouldn't work so we should probably give it a default when not specified.

kojilin: Default implementation of https tracing will use http client/server parser directly (only has few taga I think)

adriancole: yeah if there was a problem due to no data in defaults it would affect everyone. Right now the default parsing is conservative but understood. OTOH, if someone is not passing an HttpTracing (ex they pass Tracing) then it is good to use the armeria parsers because that gives same behavior as before.

adriancole:
sorry today didn't happen. today is the new today :)

adriancole:
trying off https://jitpack.io/#kojilin/armeria

adriancole:
So far so good. I use this branch here. https://github.com/openzipkin/zipkin/pull/2700

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
```diff
@@ -0,0 +1,124 @@
...
+import com.linecorp.armeria.common.RpcRequest;
+import com.linecorp.armeria.common.SerializationFormat;
+import com.linecorp.armeria.common.SessionProtocol;
+import com.linecorp.armeria.common.logging.RequestLog;
+import com.linecorp.armeria.common.logging.RequestLogAvailability;
+import com.linecorp.armeria.internal.brave.SpanTags;
+
+import brave.http.HttpClientAdapter;
+
+public class ArmeriaHttpClientAdapter extends HttpClientAdapter<RequestLog, RequestLog> {
```
trustin: Does it make sense to make this class a singleton and let a user get it via `ArmeriaHttpClientAdapter.get()`?

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
```diff
@@ -0,0 +1,76 @@
...
+ *   <li>http.url</li>
+ *   <li>http.host</li>
+ *   <li>http.protocol</li>
+ *   <li>http.serfmt</li>
+ *   <li>address.remote</li>
+ *   <li>address.local</li>
+ * </ul>
+ * User can extend this class or implement own {@link HttpClientParser}.
+ */
+public class ArmeriaHttpClientParser extends HttpClientParser {
```
trustin: Ditto - Does it make sense to make this class a singleton and let a user get it via `ArmeriaHttpClientParser.get()`?

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
```diff
@@ -0,0 +1,138 @@
...
+import com.linecorp.armeria.common.SerializationFormat;
+import com.linecorp.armeria.common.SessionProtocol;
+import com.linecorp.armeria.common.logging.RequestLog;
+import com.linecorp.armeria.common.logging.RequestLogAvailability;
+import com.linecorp.armeria.internal.brave.SpanTags;
+
+import brave.Span;
+import brave.http.HttpServerAdapter;
+
+public final class ArmeriaHttpServerAdapter extends HttpServerAdapter<RequestLog, RequestLog> {
```
trustin: Ditto - singleton?

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerParser.java
```diff
@@ -0,0 +1,76 @@
...
+ *   <li>http.url</li>
+ *   <li>http.host</li>
+ *   <li>http.protocol</li>
+ *   <li>http.serfmt</li>
+ *   <li>address.remote</li>
+ *   <li>address.local</li>
+ * </ul>
+ * User can extend this class or implement own {@link HttpServerParser}.
+ */
+public class ArmeriaHttpServerParser extends HttpServerParser {
```
trustin: Ditto - singleton?

kojilin:
updated, make some no state class become singleton. I think it's ok due to adapter and parser usually won't be extended.

adriancole:
fwiw usually we don't expose "backwards compat parser" or adapter types anyway. you can consider if you want to remove public on these as end users won't usually need to access them.

trustin:
+1 for hiding them from public API if there's a need for accessing them from outside.

kojilin:
@adrenalinee but if user want to write own parser and get the additional information from adapter. it's better to expose adapter type?

adriancole:
good point @kojilin . The adapter has extra methods for parsing. Probably just add example of why use it in type level javadoc and hide the backwards compat parsers? We want to encourage people supplying their own parsers and not rely on internal types as they aren't portable when used in mixed library setups.

adriancole:
another option is to remove the custom adapter methods. in real life using the custom adapter implies casting. unless the methods exposed are hiding some special details, casting and accessing driectly may be the better choice.

Ex when we expose adapters, usually it is because we have to hack something not easy to reach from the native request/response object. we don't add methods for things easy to reach

kojilin:
Sound like make it private is better, and if user want, they should write own parser and cast req/res to RequestLog and use it directly.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * Returns the method name if {@link RequestLog#requestContent()} is {@link RpcRequest}.
+     */
+    @Nullable
+    public String rpcMethod(RequestLog requestLog) {
+        if (!requestLog.isAvailable(RequestLogAvailability.REQUEST_CONTENT)) {
+            return null;
+        }
+        final Object requestContent = requestLog.requestContent();
```
adriancole: similar comments here in the server call. note that it is probably higher priority to support http.route vs support rpc.method.

We can add rpc.method span name decorator easily on top.  like this https://github.com/openzipkin/brave/blob/master/instrumentation/spring-webmvc/src/main/java/brave/spring/webmvc/SpanCustomizingAsyncHandlerInterceptor.java

adriancole:
wow I didn't realize @kojilin wrote the same thing heh

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
kojilin: you mean in the `BraveService`? Also what do you mean `http.route vs rpc.method`. About http service it's good to use http.route. and rpc.method should be used if it's rpc?

kojilin:
So I make class not public but not sure the `rpc.method` part.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
kojilin: adapter/parser is using order of `route -> methodFromResponse -> method(http)` 
And I want to overwrite it and before span#finish.

adriancole: I think best to not focus too much on rpc.method right now. let's make this work with route first. the order problem you mention deserves its own change. Also remember a user can add a tag with the rpc method regardless. It isn't required to be made a part of the span name.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
```diff
@@ -0,0 +1,148 @@
...
+import com.linecorp.armeria.internal.brave.SpanTags;
+
+import brave.Span;
+import brave.http.HttpServerAdapter;
+
+final class ArmeriaHttpServerAdapter extends HttpServerAdapter<RequestLog, RequestLog> {
+
+    private static final ArmeriaHttpServerAdapter INSTANCE = new ArmeriaHttpServerAdapter();
+
+    public static ArmeriaHttpServerAdapter get() {
```
trustin: nit: `public` could be removed.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerParser.java
```diff
@@ -0,0 +1,86 @@
...
+ *   <li>address.remote</li>
+ *   <li>address.local</li>
+ * </ul>
+ * User can extend this class or implement own {@link HttpServerParser}.
+ */
+final class ArmeriaHttpServerParser extends HttpServerParser {
+
+    private static final ArmeriaHttpServerParser INSTANCE = new ArmeriaHttpServerParser();
+
+    public static ArmeriaHttpServerParser get() {
```
trustin: nit: `public` could be removed.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
trustin: We can use `RequestLog.context().route().meterTag()` as the value of `http.route`. Not sure if we can fill this for the client side though.

adriancole:
ps don't worry about client side route info. it is not implemented anywhere so far

adriancole:
> We can use RequestLog.context().route().meterTag() as the value of
> http.route. Not sure if we can fill this for the client side though.
>
>
does this end up like /things/{bar} ? or is it something different. just
curious if this is a route label or a path expression (route is a path
expression)

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
kojilin: @trustin you mean we will add `route` in `RequestLog#context`?

@adriancole I see, so keep this PR as is?

kojilin: Didn't see the method in the context..

trustin:
> does this end up like /things/{bar} ? or is it something different. just
curious if this is a route label or a path expression (route is a path
expression)

Armeria supports various routing mechanism, from usual `/foo/{bar}` to glob expressions to regular expressions. They all yield different strings for `meterTag()`:

- Default: `<pattern>` e.g. `/foo/{bar}`
- Exact match: `exact:<path>` e.g. `exact:/foo/bar`
- Prefix match: `prefix:<path>` e.g. `prefix:/foo`
- Glob: `glob:<pattern>` e.g. `glob:/foo/*/bar/**`
- Regex: `regex:<pattern>` e.g. `regex:^/([^/]*)`

It can yield longer string when a user combines the path mapping with content negotiation, e.g. `/foo/{bar}, consumes=application/json`.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
adriancole: It would be awesome if this can land with http route as the experience will be much better. ex in zipkin all our server spans are get or post. If that's too hard we can do later. regardless, let's do rpc method later

adriancole:
@trustin the examples you give look good to me. main thing I was curious about is if it adds http method to the thing, and it seems not to.

trustin:
> the examples you give look good to me. main thing I was curious about is if it adds http method to the thing, and it seems not to.

It can, if a user chooses to build a route that accepts certain HTTP methods only, e.g. `/foo/{bar}, methods=GET,POST`. Should we remove the `methods` part?

adriancole:
@trustin I think it would be best to remove methods part and the content type suffixes. We may be wrong in how we deal with http.route semantics, but this would be more consistent and reduce the chance of very strange looking span names (ex `post /foo/{bar}, methods=get,post`)

adriancole:
ex here is the naming conventions, all of which assume route is a path expression
```
    if (!"".equals(route)) return method + " " + route;
    if (statusCode / 100 == 3) return method + " redirected";
    if (statusCode == 404) return method + " not_found";
```

Note: someone else might feel free to make a name based on content type. that's another reason to limit this to path expressions. In the future we may find that your tagging convention is better, and for that matter someone may want to override the span name heuristic to use it.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
kojilin: ah...service request context..let me check if i can add in this PR directly.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
```diff
@@ -88,6 +89,16 @@ public int statusCodeAsInt(RequestLog requestLog) {
...
     }
 
+    @Override
+    public String methodFromResponse(RequestLog requestLog) {
+        return requestLog.method().name();
+    }
+
+    @Override
+    public String route(RequestLog response) {
+        return ((ServiceRequestContext) response.context()).route().meterTag();
```
adriancole: anything in in this call chain (or any others) possible to return null prior to meterTag?

multi-level chains are a common source of bugs, but if the end cannot return null, multi-chain is much better to read

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceIntegrationTest.java
```diff
@@ -123,6 +124,24 @@ public void supportsPortableCustomization() throws Exception {
                 .containsEntry("response_customizer.is_span", "false");
     }
 
+    @Override
+    @Test
```
adriancole: comment why overriding?

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
trustin: - How about guarding it with `instanceof` and returning `null` when the context is not a `ServiceRequestContext`?
- How about adding some TODO like this:
  ```java
  // TODO(kojilin): Do not include `methods` in the returned string.
  ```

kojilin: in the server adapter, it must not be null and `must` be `ServiceRequestContext` I think.

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceIntegrationTest.java
```diff
@@ -123,6 +124,24 @@ public void supportsPortableCustomization() throws Exception {
...
+    public void options() throws Exception {
+        assertThat(HttpClient.of(url("/")).options("/").aggregate().join().status()).isEqualTo(OK);
+        final Span span = takeSpan();
+        assertThat(span.tags()).containsEntry("http.method", "OPTIONS")
+                               .containsEntry("http.path", "/");
+        assertThat(span.name()).isEqualTo("options exact:/");
+    }
+
+    @Override
```
adriancole: comment why overriding? we don't really want to override without calling super unless there's something odd. It is fine to add new methods, but overriding causes questions.

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
kojilin: this is implemented in server adapter, will it be non ServiceRequestContext?

kojilin: Do you think it's possible to use `Route#paths` and `Route#pathType` to make `route`?

e.g.
```
RoutePathType#EXACT || RoutePathType#PREFIX || PARAMETERIZED -> paths.get(1)
REGEX -> paths.get(0),
REGEX_WITH_PREFIX -> paths.get(0) + paths.get(1)

```

trustin: I think you can use it. Good idea. And assertion should be enough for the context type check. :+1:

adriancole: ps I just manually verify route isn't nullable https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/ServiceConfig.java#L93

adriancole:
here's current use of this branch, which shows why limiting to the path part of the meterTag would help

Screenshot 2019-07-24 at 4 09 46 PM

In fact, I would suggest the exact prefix isn't helpful. any variables would be evident in the expression

kojilin:
updated. No README yet, I think I need to create one for it.

minwoox:
> updated. No README yet, I think I need to create one for it. Also line.github.io/armeria/advanced-zipkin.html

It's already there. #1877 😄

adriancole:
@kojilin in other words behavior is same unless you switch to HttpTracing, which is great actually as it derisks upgrading

PS LGTM!

cc @tacigar 
Screenshot 2019-07-24 at 8 23 13 PM
Screenshot 2019-07-24 at 8 23 34 PM

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
```diff
@@ -0,0 +1,173 @@
...
+        final Route route = ((ServiceRequestContext) response.context()).route();
+        switch (route.pathType()) {
+            case EXACT:
+            case PREFIX:
+            case PARAMETERIZED:
+                return route.paths().get(1);
+            case REGEX:
+                return route.paths().get(0);
+            case REGEX_WITH_PREFIX:
+                return route.paths().get(0) + route.paths().get(1);
```
trustin: How about inserting a space between the two components?

core/src/main/java/com/linecorp/armeria/server/Route.java
```diff
@@ -92,8 +92,8 @@ static RouteBuilder builder() {
      *
      * {@link RoutePathType#REGEX} has only one path that represents it. e.g, {@code [ "^/(?<foo>.*)$" ]}
      *
-     * <p>{@link RoutePathType#REGEX_WITH_PREFIX} has two paths. The first one is the prefix and the second
-     * one is the regex. e.g, {@code [ "/bar/", "^/(?<foo>.*)$" ]}
+     * <p>{@link RoutePathType#REGEX_WITH_PREFIX} has two paths. The first one is the regex and the second
+     * one is the path. e.g, {@code [ "/bar/", "^/(?<foo>.*)$" ]}
```
kojilin: cc @minwoox 
https://github.com/line/armeria/blob/master/core/src/test/java/com/linecorp/armeria/server/PathWithPrefixTest.java#L33

core/src/main/java/com/linecorp/armeria/server/Route.java
```diff
@@ -92,8 +92,8 @@ static RouteBuilder builder() {
      *
      * {@link RoutePathType#REGEX} has only one path that represents it. e.g, {@code [ "^/(?<foo>.*)$" ]}
      *
-     * <p>{@link RoutePathType#REGEX_WITH_PREFIX} has two paths. The first one is the prefix and the second
-     * one is the regex. e.g, {@code [ "/bar/", "^/(?<foo>.*)$" ]}
+     * <p>{@link RoutePathType#REGEX_WITH_PREFIX} has two paths. The first one is the regex and the second
+     * one is the path. e.g, {@code [ "^/(?<foo>.*)$", "/bar/" ]}
```
kojilin: cc @minwoox
https://github.com/line/armeria/blob/master/core/src/test/java/com/linecorp/armeria/server/PathWithPrefixTest.java#L33

minwoox: Oops, thanks a lot. ❤️

adriancole: can this be combined into one pattern? I don't quite understand why having two helps. the main thing about http.route is the matched route. Is this saying there are two matched routes?

brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
```diff
@@ -165,4 +165,12 @@ public void route_prefix() {
         when(context.route()).thenReturn(Route.builder().path("exact:/foo").build());
         assertThat(ArmeriaHttpServerAdapter.get().route(requestLog)).isEqualTo("/foo");
     }
+
+    @Test
+    public void route_pathWithPrefix() {
+        final ServiceRequestContext context = mock(ServiceRequestContext.class);
+        when(requestLog.context()).thenReturn(context);
+        when(context.route()).thenReturn(Route.builder().pathWithPrefix("/foo/", "glob:bar").build());
+        assertThat(ArmeriaHttpServerAdapter.get().route(requestLog)).isEqualTo("/foo/ ^/(?:.+/)?bar$");
```
adriancole: Why is this not concatenated? a space here what does this help qualify?

ex this will be `get /foo/ ^/(?:.+/)?bar$` for what seems to be much simpler as `get /foo/**bar$`

what am I missing?

core/src/main/java/com/linecorp/armeria/server/Route.java
trustin: Let's say a user specified `regex:x`, which will match every path that contains x. If we concat the prefix and regex without a space, it will be `/foo/x` which is not distinguishable from `/foo/x`.

brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
trustin: Because a regular expression is complex and we can never know how to simplify an arbitrary regular expression into a glob pattern?

adriancole: I guess the declaration of the route expression with literal `glob:bar` is lost then? it would be much easier to read if it weren't..

core/src/main/java/com/linecorp/armeria/server/Route.java
adriancole: the span name is a grouping concept as well as a user label. I'm not sure if you meant to write /foo/x twice but if you did then the result of the matched route would be the important part. the aim of the http.route is to lower the cardinality of paths.. not really much else.

it may be the case here that flexibility makes these appear more complex than other frameworks but I would sit on the idea a bit and see if there is a way to avail more user friendly route names. this doesnt have to happen in this PR if it is complicated

brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
adriancole: main idea.. and I understand this may not happen here.. is that there is parity between the definition and the result. ex if the definition is a simple prefix with glob, what the user put in could be in the http.route tag. you can consider this not binding comment just I want armeria to appear best!

kojilin: Doesn't have strong opinion here, I think it's for the case that we write like
`sb.annotatedService("/6", new MyAnnotatedService6())` with path setting like `"regex:^/regex/(?<path>.*)$"` or `glob:/glob1/*`. 

https://github.com/line/armeria/blob/master/core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceTest.java#L117

If we want to enhance it, we may need create new method for `PathMapping`?

adriancole: I think metertag is just an implementation choice of getting the http route template. This is one good thing about package private as we can change it :)

fwiw there are other things that have regex even if path templates are more common. ex @shimamoto maintains play zipkin tracing which has some cleaning to make it look less like cat walked on keyboard https://github.com/openzipkin-contrib/play-zipkin-tracing/blob/master/play-zipkin-tracing/play/src/main/scala/brave/play/filter/ZipkinTraceFilter.scala#L53

site/src/sphinx/advanced-zipkin.rst
```diff
@@ -24,6 +24,7 @@ First, you need to create the ``Tracing``:
                              .currentTraceContext(RequestContextCurrentTraceContext.ofDefault())
                              .spanReporter(myReporter)
                              .build();
+    HttpTracing httpTracing = HttpTracing.create(tracing);
```
ikhoon: You did an awesome job!
Finally, could you add `import` in this example?
```java
import brave.http.HttpTracing;
```

brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
kojilin: What do you mean package private ?
Let me check if I can replace the pattern to ** or * in all cases. WDYT? @trustin

kojilin: And it's better to create new method and pre-calculate the route string instead of always `replace` in `adapter#route`, right?

trustin: > Let me check if I can replace the pattern to ** or * in all cases. WDYT?

Wouldn't it cause lots of ambiguity when a user regex/glob mapping extensively? I still prefer leaving the expression as is or doing minimal clean-up only for reduced confusion.

trustin: > I guess the declaration of the route expression with literal glob:bar is lost then? it would be much easier to read if it weren't..

Ah yeah, that's true. I totally missed that the regular expression was derived from a glob expression. Sorry!

Anyways, I think that should be solved by fixing our routing implementation, like adding 'prefix + glob' type.

trustin: Actually, for prefix + glob case, it could be simplified into a single glob expression. Here:

https://github.com/line/armeria/blob/f7e25f9081d41008443bd0896d8a28d1e4e4ca95/core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java#L194-L199

we could prepend `**/` when the glob expression does not start with `/`, e.g. (prefix = `/foo/` and glob = `bar`) could be simplified into `/foo/**/bar`.

trustin: @adriancole, This is what you meant, right? :sweat_smile: 

So, my suggestion is:

- for the case of (prefix + regex), leave the regex as is for less ambiguity.
  - I'm not completely sure `prefix + ' ' + regex` is the best. Any suggestions? Does `'(' + prefix + ", " + regex + ')'` look better?
  - +1 for pre-calculating but we need to discuss about where the precalculated string would be stored, API design, etc.
- for the case of (prefix + glob), let us fix the bug where it is not simplified into a glob, which is not the scope of this pull request.

WDYT?

adriancole: I agree that simplifying like the `/foo/**/bar` will be far more readable.

wrt nesting, all the other frameworks with nested expressions join on '/'. I understand there is certain nuance about what happens with multiple expressions that can yield the same pattern. Moreover, sometimes a request is internally routed multiple times. I do understand that there is some complexity some people want to see somewhere. However, I really do think joining on not slash is the most distracting choice to make as an input to the span name. If you feel strongly about doing space joining, it is ok, just mentioning this would be the only framework doing this. It seems like if someone really needed to know the exact difference between these cases, they could add a tag and/or the controller method.

anyway I think we've spent plenty of think time on this, I appreciate you hearing it out. I think you have the understanding needed to make a good call and both choices are moot for those who avoid this pathWithPrefix method.

minwoox: +1 for `/foo/**/bar`.
And for the case of (prefix + regex), I think `'(' + prefix + ", " + regex + ')'` looks better.

trustin: @adriancole If you are strong about this, then I'm OK with joining on `/`. It is at least better than joining without any separator. :sweat_smile:

minwoox: So when the prefix is `/foo/` and the regex is `^/(?:.+/)?bar$`,
will it be `/foo//^/(?:.+/)?bar$`

trustin: Because the `prefix` already ends with `/`, we can just concat without any separator, like before.

adriancole: mild request is to strip all redundancy here `//^/` otherwise let's get good enough merged :P

minwoox: Yes, it's okay for me. :-)
Thanks for pointing this out anyway. I was feeling somehow the meter tag is unclear and needed to be fixed when I was working on it. But I didn't do it yet. 😄

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapter.java
```diff
@@ -0,0 +1,173 @@
...
+        if (!requestLog.isAvailable(RequestLogAvailability.REQUEST_CONTENT)) {
+            return null;
+        }
+        final Object requestContent = requestLog.requestContent();
+        return requestContent instanceof RpcRequest ? ((RpcRequest) requestContent).method() : null;
+    }
+
+    /**
+     * This sets the client IP:port to the {@link RequestContext#remoteAddress()}
+     * if the {@link HttpServerAdapter#parseClientIpAndPort default parsing} fails.
```
minwoox: nit: link -> linkplain

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerParser.java
```diff
@@ -0,0 +1,86 @@
...
+ * The following tags become available:
+ * <ul>
+ *   <li>http.url</li>
+ *   <li>http.host</li>
+ *   <li>http.protocol</li>
+ *   <li>http.serfmt</li>
+ *   <li>address.remote</li>
+ *   <li>address.local</li>
+ * </ul>
+ * User can extend this class or implement own {@link HttpServerParser}.
```
minwoox: ditto

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerParser.java
```diff
@@ -0,0 +1,86 @@
...
+import com.linecorp.armeria.common.logging.RequestLog;
+import com.linecorp.armeria.internal.brave.SpanTags;
+
+import brave.SpanCustomizer;
+import brave.http.HttpAdapter;
+import brave.http.HttpServerParser;
+
+/**
+ * Default implementation of {@link HttpServerParser}.
+ * This parser add some custom tags and overwrite the name of span if {@link RequestLog#requestContent()}
```
minwoox: ditto

trustin:
Let me merge this pull request first and then send a follow-up pull request quickly, because @kojilin will not be available for a bit while and we want to make this part of today's release.

trustin:
Thanks @kojilin and @adriancole !

adriancole:
This is great. we will use it immediately along with the other awesome in armeria 0.89

PR #1916, opened 2019-07-19
Allow updating `ClientRequestContext.endpoint()` in a decorator

trustin:
Motivation:

- A user sometimes wants to write a decorator which overrides the target
  host specified from the client request URI, which is translated into
  `ClientRequestContext.endpoint()`.
- When a user specifies an endpoint group in a request URI, such as
  `http://group:foo/`, a decorator has no way to know which endpoint
  will actually handle the request, because it is determined after all
  decorators are evaluated. This can be a problem when a decorator has
  to do something based on the target endpoint's property such as host
  name and IP address.

Modifications:

- Add `ClientRequestContext.setEndpoint()` which allows updating the
  target remote endpoint where the request will be sent.
- Add a simple test case which shows how a decorator can select an
  `Endpoint` from an `EndpointGroup` before Armeria does it at the end
  of decorator chain.

Result:

- A user can reroute a request by using a decorator.
- A user can determine which endpoint a request will be sent to when an
  endpoint group is used.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -139,9 +139,15 @@ public interface ClientRequestContext extends RequestContext {
     ClientRequestContext newDerivedContext(Request request);
 
     /**
-     * Returns the remote {@link Endpoint} of the current {@link Request}.
+     * Returns the remote {@link Endpoint} where the current {@link Request} will be sent to.
      */
     Endpoint endpoint();
+
+    /**
+     * Sets the {@link Endpoint} where the current {@link Request} will be sent to.
+     * This method is useful to a decorator that manipulates the target host of the {@link Request}.
+     */
+    void setEndpoint(Endpoint endpoint);
 
     /**
      * Returns the {@link ClientOptions} of the current {@link Request}.
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
@@ -52,6 +52,11 @@ public class ClientRequestContextWrapper
     }
 
     @Override
+    public void setEndpoint(Endpoint endpoint) {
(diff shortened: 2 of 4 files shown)


trustin:
@adriancole PTAL

anuraaga:
The code for selecting a node and updating the endpoint isn't so much, but do you think it'd also make sense for Armeria to select the endpoint before calling decorators to execute the request? It feels like it'd be natural for a request to already have its resolved endpoint when its being executed.

trustin:
> The code for selecting a node and updating the endpoint isn't so much, but do you think it'd also make sense for Armeria to select the endpoint before calling decorators to execute the request? It feels like it'd be natural for a request to already have its resolved endpoint when its being executed.

+1 for moving the endpoint selection code to the beginning of the decorator chain, but I'm curious about its side effect. For example, it may break an existing `EndpointSelectionStrategy` implementation such as `StickyEndpointSelectionStrategy` that relies on a certain `ClientRequestContext` attribute. Do you have any use case related with this?

trustin:
After thinking a little bit more, I now think adding `setEndpoint` is not a good idea because we might want to move both endpoint selection and connection acquisition before the decorator chain so that `EventLoopScheduler` and `HttpChannelPool` can be merged into a single efficient implementation (I think you remember this, right?). If it ever comes true, `setEndpoint()` will have to short-live and be removed. Let me close this pull request and move the endpoint selection logic in the beginning.

`StickyEndpointSelectionStrategy` is usually used by hashing a certain attribute set via `Clients.withContextCustomizer()`, so I think moving the endpoint selection shouldn't be a problem:

```java
try (SafeCloseable ignored = Clients.withContextCustomizer(ctx -> {
    ctx.attr(MY_ATTR).set(someValueForHashing);
}) {
    client.execute(someRequest);
}
```

anuraaga:
Ah that sounds familiar :) And yeah, I agree that for `StickyEndpointSelectionStrategy` it would be very rare to want to set the attribute in a decorator. In the off chance that someone requested it, adding a callback during endpoint selection would take care of it just fine.

PR #1917, opened 2019-07-19
Move Endpoint selection to the beginning of client execution

trustin:
Motivation:

`ClientRequestContext.endpoint()` currently return an `Endpoint` which
may be 1) a host:port endpoint or 2) a group endpoint. Therefore, a
client decorator does not know which host the request will be sent to
exactly; it is determined after all decorators are evaluated. This makes
it hard for client decorators to do the following:

- Injecting a header value derived from the target host name.
- Accounting for the number of requests sent to each host.

Modifications:

- Split the initialization of `DefaultClientRequestContext` into two:
  1) constructor and 2) `init()` method.
- Determine the target host at the context initialization phase.
- Make `ClientRequestContext.endpoint()` nullable, so that the
  decorators can handle the case where endpoint selection has failed.
- Add `ClientRequestContext.endpointSelector()`.
- Add test cases for HTTP, gRPC and Thrift.
- Miscellaneous:
  - Add `EndpointGroup.empty()`

Result:

- Client decorators now knows the exact target host of the request.
- We can make `Endpoint` always refer to a host:port, which will greatly
  simplify both users' code and ours in the future.
- We now have a proper place (`DefaultClientRequestContext.init()`) that
  can be used for `EventLoop` allocation and connection pooling in the
  future.

~It currently has the following issues:~

- ~When `EndpointSelector.select()` raises an exception, the construction of `ClientRequestContext` fails, which means you have no access to its `RequestLog`. We could make `ClientRequestContext` available by deferring `select()` until `ClientRequestContext` is constructed, but the decorators will still not be able to access to the request.~
  - ~The same problem will arise when we move the connection acquisition logic (connection pooling) to the beginning of request execution. The decorators will never see the connection failure.~

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -27,6 +27,7 @@ import java.util.function.Supplier;
 
 import javax.annotation.Nullable;
 
+import com.linecorp.armeria.client.endpoint.EndpointSelector;
 import com.linecorp.armeria.common.ContentTooLargeException;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpRequest;
@@ -34,6 +35,9 @@ import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.RequestContext;
 import com.linecorp.armeria.common.Response;
 import com.linecorp.armeria.common.RpcRequest;
+import com.linecorp.armeria.common.logging.RequestLog;
+
+import io.netty.util.Attribute;
 
 /**
  * Provides information about a {@link Request}, its {@link Response} and its related utilities.
@@ -139,8 +143,28 @@ public interface ClientRequestContext extends RequestContext {
     ClientRequestContext newDerivedContext(Request request);
 
     /**
-     * Returns the remote {@link Endpoint} of the current {@link Request}.
+     * Creates a new derived {@link RequestContext} with the specified {@link Request} and {@link Endpoint}
+     * which the {@link RequestLog} is different from the deriving context.
+     * Note that the references of {@link Attribute}s in the {@link #attrs()} are copied as well.
      */
+    ClientRequestContext newDerivedContext(Request request, Endpoint endpoint);
+
+    /**
+     * Returns the {@link EndpointSelector} used for the current {@link Request}.
+     *
+     * @return the {@link EndpointSelector} if a user specified a group {@link Endpoint}
+     *         {@code null} if a user specified a host {@link Endpoint}.
+     */
+    @Nullable
+    EndpointSelector endpointSelector();
+
+    /**
+     * Returns the remote {@link Endpoint} of the current {@link Request}.
+     *
+     * @return the remote {@link Endpoint}. {@code null} if the {@link Request} has failed
+     *         before its remote {@link Endpoint} is determined.
+     */
+    @Nullable
     Endpoint endpoint();
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -136,23 +136,31 @@ public interface ClientRequestContext extends RequestContext {
         return ClientRequestContextBuilder.of(request, uri).build();
     }
 
+    /**
+     * Creates a new {@link ClientRequestContext} whose properties and {@link Attribute}s are copied from this
+     * {@link ClientRequestContext}, except having its own {@link RequestLog}.
+     */
     @Override
     ClientRequestContext newDerivedContext();
 
+    /**
+     * Creates a new {@link ClientRequestContext} whose properties and {@link Attribute}s are copied from this
+     * {@link ClientRequestContext}, except having a different {@link Request} and its own {@link RequestLog}.
+     */
     @Override
     ClientRequestContext newDerivedContext(Request request);
 
     /**
-     * Creates a new derived {@link RequestContext} with the specified {@link Request} and {@link Endpoint}
-     * which the {@link RequestLog} is different from the deriving context.
-     * Note that the references of {@link Attribute}s in the {@link #attrs()} are copied as well.
+     * Creates a new {@link ClientRequestContext} whose properties and {@link Attribute}s are copied from this
+     * {@link ClientRequestContext}, except having different {@link Request}, {@link Endpoint} and its own
+     * {@link RequestLog}.
      */
     ClientRequestContext newDerivedContext(Request request, Endpoint endpoint);
 
     /**
      * Returns the {@link EndpointSelector} used for the current {@link Request}.
      *
-     * @return the {@link EndpointSelector} if a user specified a group {@link Endpoint}
+     * @return the {@link EndpointSelector} if a user specified a group {@link Endpoint}.
      *         {@code null} if a user specified a host {@link Endpoint}.
      */
     @Nullable
@@ -162,7 +170,7 @@ public interface ClientRequestContext extends RequestContext {
      * Returns the remote {@link Endpoint} of the current {@link Request}.
      *
      * @return the remote {@link Endpoint}. {@code null} if the {@link Request} has failed
-     *         before its remote {@link Endpoint} is determined.
+     *         because its remote {@link Endpoint} couldn't be determined.
      */
     @Nullable
     Endpoint endpoint();
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
@@ -72,14 +72,14 @@ public class DefaultClientRequestContext extends NonWrappingRequestContext imple
     private long responseTimeoutMillis;
     private long maxResponseLength;
 
+    private volatile HttpHeaders additionalRequestHeaders = HttpHeaders.of();
+
     @Nullable
     private String strVal;
 
-    @Nullable
-    private volatile HttpHeaders additionalRequestHeaders;
-
     /**
-     * Creates a new instance.
+     * Creates a new instance. Note that {@link #init(Endpoint)} method must be invoked to finish
+     * the construction of this context.
      *
      * @param sessionProtocol the {@link SessionProtocol} of the invocation
      * @param request the request associated with this context
@@ -312,10 +312,6 @@ public class DefaultClientRequestContext extends NonWrappingRequestContext imple
 
     @Override
     public HttpHeaders additionalRequestHeaders() {
-        final HttpHeaders additionalRequestHeaders = this.additionalRequestHeaders;
-        if (additionalRequestHeaders == null) {
-            return HttpHeaders.of();
-        }
         return additionalRequestHeaders;
     }
 
@@ -323,42 +319,33 @@ public class DefaultClientRequestContext extends NonWrappingRequestContext imple
     public void setAdditionalRequestHeader(CharSequence name, Object value) {
         requireNonNull(name, "name");
         requireNonNull(value, "value");
-        additionalRequestHeaders = createAdditionalHeadersIfAbsent().toBuilder().setObject(name, value).build();
+        additionalRequestHeaders = additionalRequestHeaders.toBuilder().setObject(name, value).build();
     }
 
     @Override
     public void setAdditionalRequestHeaders(Iterable<? extends Entry<? extends CharSequence, ?>> headers) {
         requireNonNull(headers, "headers");
-        additionalRequestHeaders = createAdditionalHeadersIfAbsent().toBuilder().setObject(headers).build();
+        additionalRequestHeaders = additionalRequestHeaders.toBuilder().setObject(headers).build();
     }
 
     @Override
     public void addAdditionalRequestHeader(CharSequence name, Object value) {
         requireNonNull(name, "name");
         requireNonNull(value, "value");
-        additionalRequestHeaders = createAdditionalHeadersIfAbsent().toBuilder().addObject(name, value).build();
+        additionalRequestHeaders = additionalRequestHeaders.toBuilder().addObject(name, value).build();
     }
 
     @Override
     public void addAdditionalRequestHeaders(Iterable<? extends Entry<? extends CharSequence, ?>> headers) {
         requireNonNull(headers, "headers");
-        additionalRequestHeaders = createAdditionalHeadersIfAbsent().toBuilder().addObject(headers).build();
-    }
-
-    private HttpHeaders createAdditionalHeadersIfAbsent() {
-        final HttpHeaders additionalRequestHeaders = this.additionalRequestHeaders;
-        if (additionalRequestHeaders == null) {
-            return this.additionalRequestHeaders = HttpHeaders.of();
-        } else {
-            return additionalRequestHeaders;
-        }
+        additionalRequestHeaders = additionalRequestHeaders.toBuilder().addObject(headers).build();
     }
 
     @Override
     public boolean removeAdditionalRequestHeader(CharSequence name) {
         requireNonNull(name, "name");
         final HttpHeaders additionalRequestHeaders = this.additionalRequestHeaders;
-        if (additionalRequestHeaders == null || additionalRequestHeaders.isEmpty()) {
+        if (additionalRequestHeaders.isEmpty()) {
             return false;
         }
 
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.client;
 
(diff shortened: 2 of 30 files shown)


trustin:
@anuraaga @ikhoon @minwoox Thoughts?

anuraaga:
Does it work for the failure case to set the `Endpoint` of the request to a `FailedEndpoint` or something and continue to execute the request, throwing when about to connect to the `FailedEndpoint`?

trustin:
> Does it work for the failure case to set the `Endpoint` of the request to a `FailedEndpoint` or something and continue to execute the request, throwing when about to connect to the `FailedEndpoint`?

That sounds good to me. We could also do the same when we move the connection pooling to the beginning. Sounds good?

anuraaga:
Yup :)

trustin:
I found it's weird to use `Endpoint` to represent different states that are not very similar to each other - group, host, failure, so I just made `ClientRequestContext.endpoint()` nullable as advised by @minwoox. What do you think? If that's fine, let me add some test cases and remove WIP.

trustin:
That being said, I'd eventually like to make `Endpoint` represent only a host:port, i.e. no group, because `EndpointSelector` can be used instead.

anuraaga:
Yeah that makes sense

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -138,9 +142,29 @@ static ClientRequestContext of(RpcRequest request, URI uri) {
     @Override
     ClientRequestContext newDerivedContext(Request request);
 
+    /**
+     * Creates a new derived {@link RequestContext} with the specified {@link Request} and {@link Endpoint}
+     * which the {@link RequestLog} is different from the deriving context.
```
anuraaga: `with a different {@link RequestLog} than this one.`

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -138,9 +142,29 @@ static ClientRequestContext of(RpcRequest request, URI uri) {
...
+     *         {@code null} if a user specified a host {@link Endpoint}.
+     */
+    @Nullable
+    EndpointSelector endpointSelector();
+
     /**
      * Returns the remote {@link Endpoint} of the current {@link Request}.
+     *
+     * @return the remote {@link Endpoint}. {@code null} if the {@link Request} has failed
+     *         before its remote {@link Endpoint} is determined.
```
anuraaga: Is it better using `before` or `because its remote endpoint couldn't be determined`?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -99,17 +106,46 @@ public DefaultClientRequestContext(
...
 
-    private HttpHeaders createAdditionalHeadersIfAbsent() {
-        final HttpHeaders additionalRequestHeaders = this.additionalRequestHeaders;
-        if (additionalRequestHeaders == null) {
-            return this.additionalRequestHeaders = HttpHeaders.of();
-        } else {
-            return additionalRequestHeaders;
+    /**
+     * Initializes this context with the specified {@link Endpoint}.
+     * This method must be invoked to finish the construction of this context.
```
anuraaga: Maybe also add this to the constructor.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -289,6 +345,15 @@ public void addAdditionalRequestHeaders(Iterable<? extends Entry<? extends CharS
         additionalRequestHeaders = createAdditionalHeadersIfAbsent().toBuilder().addObject(headers).build();
     }
 
+    private HttpHeaders createAdditionalHeadersIfAbsent() {
```
anuraaga: Unrelated to this PR, but is this needed? I think since we have immutable headers now, we should be able to initialize the field to empty instead of null without overhead.

core/src/test/java/com/linecorp/armeria/client/DefaultClientRequestContextTest.java
```diff
@@ -19,7 +19,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
```
anuraaga: I think when migrating to junit5, we should also migrate to the convention of package-private test methods. Maybe it creates too large a diff for this PR though.

grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
```diff
@@ -165,15 +168,16 @@ public void start(Listener<O> responseListener, Metadata metadata) {
...
-            res = httpClient.execute(ctx, req);
-        } catch (Exception e) {
-            close(GrpcStatus.fromThrowable(e), new Metadata());
-            return;
-        }
+
+        final HttpResponse res = initContextAndExecuteWithFallback(
+                httpClient, ctx, endpoint,
+                (unused, cause) -> HttpResponse.ofFailure(GrpcStatus.fromThrowable(cause)
+                                                                    .withCause(cause)
```
anuraaga: Don't need `withCause`

core/src/main/java/com/linecorp/armeria/client/retry/RetryingClient.java
```diff
@@ -250,6 +252,19 @@ protected static int getTotalAttempts(ClientRequestContext ctx) {
...
+    /**
+     * Creates a new derived {@link ClientRequestContext}, replacing the {@link Request} with {@code req}.
+     * If {@link ClientRequestContext#endpointSelector()} exists, a new {@link Endpoint} will be selected.
+     */
+    protected static ClientRequestContext newDerivedContext(ClientRequestContext ctx, Request req) {
+        final EndpointSelector endpointSelector = ctx.endpointSelector();
+        if (endpointSelector != null) {
+            return ctx.newDerivedContext(req, endpointSelector.select(ctx));
+        } else {
+            return ctx.newDerivedContext(req);
```
minwoox: If the resolution is faied, this will throw `IllegalStateExeption` inside `newDerivcedCotext`.
Don't we need to throw `EndpointGroupException` for that?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -138,9 +142,29 @@ static ClientRequestContext of(RpcRequest request, URI uri) {
...
+     * Creates a new derived {@link RequestContext} with the specified {@link Request} and {@link Endpoint}
+     * which the {@link RequestLog} is different from the deriving context.
+     * Note that the references of {@link Attribute}s in the {@link #attrs()} are copied as well.
+     */
+    ClientRequestContext newDerivedContext(Request request, Endpoint endpoint);
+
+    /**
+     * Returns the {@link EndpointSelector} used for the current {@link Request}.
+     *
+     * @return the {@link EndpointSelector} if a user specified a group {@link Endpoint}
```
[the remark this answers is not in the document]
trustin: Fixed

core/src/main/java/com/linecorp/armeria/internal/ClientUtil.java
```diff
@@ -22,34 +22,86 @@
...
+
+        final RequestLogBuilder logBuilder = ctx.logBuilder();
+        if (!ctx.log().isAvailable(RequestLogAvailability.REQUEST_START)) {
+            // An exception is raised even before sending a request,
+            // so end the request with the exception.
+            logBuilder.endRequest(cause);
+
+            final Request req = ctx.request();
+            if (req instanceof StreamMessage) {
+                ((StreamMessage<?>) req).abort();
```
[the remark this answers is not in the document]
trustin: :wink:

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -99,17 +106,46 @@ public DefaultClientRequestContext(
...
+                if (endpointSelector == null) {
+                    throw new EndpointGroupException(
+                            "non-existent " + EndpointGroup.class.getSimpleName() + ": " + groupName);
+                }
+
+                this.endpointSelector = endpointSelector;
+                // Note: thread-local customizer must be run before EndpointSelector.select()
+                //       so that the customizer can inject the attributes which may be required
+                //       by the EndpointSelector.
+                runThreadLocalContextCustomizer();
```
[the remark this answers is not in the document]
trustin: Updated the Javadoc

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
trustin: Rewrote the method Javadoc. Please review again.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
trustin: Changed to `{@code null} if the {@link Request} has failed because its remote {@link Endpoint} couldn't be determined.`

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: Updated the Javadoc for the constructor that requires a subsequent `init()` call.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: Indeed. Updated them all. :+1:

core/src/test/java/com/linecorp/armeria/client/DefaultClientRequestContextTest.java
trustin: Removed the `public` modifiers from this class.

grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -447,104 +415,144 @@ public AccessLogWriter accessLogWriter() {
...
         requireNonNull(value, "value");
-        additionalResponseTrailers = createAdditionalTrailersIfAbsent().toBuilder()
-                                                                       .setObject(name, value).build();
+        for (;;) {
+            final HttpHeaders oldValue = additionalResponseTrailers;
+            final HttpHeaders newValue = oldValue.toBuilder().setObject(name, value).build();
+            if (additionalResponseTrailersUpdater.compareAndSet(this, oldValue, newValue)) {
+                return;
+            }
+        }
```
ikhoon: nit: How about extracting this into private method?
```java
public void setAdditionalResponseTrailer(CharSequence name, Object value) {
    requireNonNull(name, "name");
    requireNonNull(value, "value");

    updateAdditionalResponseTrailer(oldValue -> oldValue.toBuilder().setObject(name, value).build());
}
private void updateAdditionalResponseTrailer(Function<HttpHeaders, HttpHeaders> updater) {
    for (;;) {
        final HttpHeaders oldValue = additionalResponseTrailers;
        final HttpHeaders newValue = updater.apply(oldValue);
        if (additionalResponseTrailersUpdater.compareAndSet(this, oldValue, newValue)) {
           return;
        }
    }
}
```

core/src/test/java/com/linecorp/armeria/client/circuitbreaker/KeyedCircuitBreakerMappingTest.java
```diff
@@ -29,7 +29,6 @@
 public class KeyedCircuitBreakerMappingTest {
     @Test
     public void hostSelector() throws Exception {
```
ikhoon: with removing public? :-)

trustin: Fixed

trustin:
@anuraaga I also hardened the thread safety of `ctx.additional{Request,Response}{Headers,Trailers}()` using CAS. Let me know if you have a concern about this, although I think it will not be that expensive with the level of contention we expect.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -119,16 +156,25 @@ private void runThreadLocalContextCustomizer() {
...
 
-    private DefaultClientRequestContext(DefaultClientRequestContext ctx, Request request) {
+    private DefaultClientRequestContext(DefaultClientRequestContext ctx, Request request, Endpoint endpoint) {
         super(ctx.meterRegistry(), ctx.sessionProtocol(), ctx.method(), ctx.path(), ctx.query(), request);
 
         eventLoop = ctx.eventLoop();
         options = ctx.options();
-        endpoint = ctx.endpoint();
+        endpointSelector = ctx.endpointSelector();
+        this.endpoint = requireNonNull(endpoint, "endpoint");
```
anuraaga: Very minor nit but this looks a little out of place and maybe could go after setting all the `ctx` fields.

core/src/main/java/com/linecorp/armeria/client/endpoint/StaticEndpointGroup.java
```diff
@@ -56,13 +58,6 @@ public StaticEndpointGroup(Iterable<Endpoint> endpoints) {
...
     public String toString() {
-        final StringBuilder buf = new StringBuilder();
-        buf.append("StaticEndpointGroup(");
-        for (Endpoint endpoint : endpoints) {
-            buf.append(endpoint).append(',');
-        }
-        buf.setCharAt(buf.length() - 1, ')');
-
-        return buf.toString();
+        return getClass().getSimpleName() + endpoints;
```
anuraaga: Any reason not to use toStringHelper?

Also, since the class is final I guess instead of `getClass()` could be `StaticEndpointGroup.class`

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: They are all `ctx` fields (`eventLoop`, `options`, `endpointSelector` and `this.endpoint`).

core/src/main/java/com/linecorp/armeria/client/endpoint/StaticEndpointGroup.java
trustin: > Any reason not to use toStringHelper?

Because it will have an extra pair of curly braces? `StaticEndpointGroup{[...]}` vs `StaticEndpointGroup[...]`

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -254,24 +249,25 @@ private ContentPreviewResponse contentPreviewResponse(HttpResponseDuplicator res
...
-            try {
-                final Long later = headers.getTimeMillis(HttpHeaderNames.RETRY_AFTER);
-                millisAfter = later - System.currentTimeMillis();
-            } catch (Exception ignored) {
-                logger.debug("The retryAfter: {}, from the server is neither an HTTP date nor a second.",
-                             value);
+
+            @SuppressWarnings("UseOfObsoleteDateTimeApi")
+            final Date date = DateFormatter.parseHttpDate(value);
+            if (date != null) {
```
ikhoon: I wonder we don't need to catch `IllegalArgumentException`.
IIUC, it can throw `IllegalArgumentException` even though [the documentation](https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/DateFormatter.java#L102) says it returns null if it fails to be parsed.

trustin: Oh, indeed. I fixed everywhere that calls `parseHttpDate()`.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
trustin: Sure. Reduced quite a bit of code duplicates thanks to that. :-)

logback/src/main/java/com/linecorp/armeria/common/logback/RequestContextExporter.java
```diff
@@ -267,7 +267,9 @@ private static void exportAuthority(Map<String, String> out, RequestContext ctx,
         } else {
             final ClientRequestContext cCtx = (ClientRequestContext) ctx;
             final Endpoint endpoint = cCtx.endpoint();
-            if (endpoint.isGroup()) {
+            if (endpoint == null) {
+                authority = "UNKNOWN";
```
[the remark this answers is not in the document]
trustin: Changed to `UNKNOWN`

PR #1919, opened 2019-07-20
Implement ByteBufHttpData.toInputStream

### LINKED ISSUE #1918: Overload ByteBufHttpData.toInputStream to avoid coersion to byte array
`ByteBufHttpData.toInputStream` could return something like `new ByteBufferBackedInputStream(content().nioBuffer())` instead of default which first coerces to a byte array.

anuraaga:
It's fairly common to convert a `HttpData` to an `InputStream`, e.g. when parsing into Jackson, so it's nice to make sure `ByteBufHttpData.toInputStream` is efficient too.

Fixes #1918

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpData.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpData.java
@@ -32,6 +32,7 @@ import java.util.Locale;
 import com.linecorp.armeria.unsafe.ByteBufHttpData;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufUtil;
 import it.unimi.dsi.fastutil.io.FastByteArrayInputStream;
 
@@ -386,6 +387,10 @@ public interface HttpData extends HttpObject {
 
     /**
      * Returns a new {@link InputStream} that is sourced from this data.
+     *
+     * <p>Note, if this {@link HttpData} is pooled (e.g., it is the result of a call to
+     * {@link HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}), then this {@link InputStream} will
+     * release the underlying buffer back to the pool when closed.
      */
     default InputStream toInputStream() {
         return new FastByteArrayInputStream(array());
--- a/core/src/main/java/com/linecorp/armeria/common/HttpData.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpData.java
@@ -34,6 +34,7 @@ import com.linecorp.armeria.unsafe.ByteBufHttpData;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufUtil;
+import io.netty.util.ReferenceCountUtil;
 import it.unimi.dsi.fastutil.io.FastByteArrayInputStream;
 
 /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -386,6 +387,10 @@ default String toStringAscii() {
 
     /**
      * Returns a new {@link InputStream} that is sourced from this data.
+     *
+     * <p>Note, if this {@link HttpData} is pooled (e.g., it is the result of a call to
+     * {@link HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}), then this {@link InputStream} will
+     * release the underlying buffer back to the pool when closed.
```
adriancole: is there impact to calling release twice? or this is mainly to let people know that they are effectively transferring responsibility of the http data to this inputstream and shouldn't use it later

anuraaga: Netty will throw an exception if calling release too many times, meaning the reference count goes negative. The comment is to let users know that they don't need to release, though if they need to keep the buffer anyways they need to first `retain()`. Using pooled objects is definitely more advanced :)

adriancole: thx will keep in mind for when I use this.

trustin: Shouldn't we retain the buffer before creating a `ByteBufInputStream`? I find it unnatural that `HttpData` is released when its stream is closed, because otherwise converting an `HttpData` into an `InputStream` twice might not work:

```java
sendTo(data.toInputStream(), outStreamA);
sendTo(data.toInputStream(), outStreamB);
data.release();
```

anuraaga:
That's a good point - went ahead and called `retainedDuplicate` (just `retain` wouldn't solve the problem)

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -390,7 +390,8 @@ default String toStringAscii() {
      *
      * <p>Note, if this {@link HttpData} is pooled (e.g., it is the result of a call to
      * {@link HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}), then this {@link InputStream} will
-     * release the underlying buffer back to the pool when closed.
+     * increase the reference count of the underlying buffer. Make sure to call {@link InputStream#close()},
+     * usually using a try-with-resources invocation, to release this extra reference.
```
adriancole: very good comment

trustin:
Thanks @anuraaga and reviewers!

adriancole:
I feel this call to have to do bookkeeping in order to handle a potential second use of the inputstream wasn't the right call.

@trustin I get you here, but isn't it often the case someone has a one-shot response? https://github.com/line/armeria/pull/1919#discussion_r305658193

Maybe we can revert this as the bookkeeping needed is very surprising compared to assuming toInputStream can only be invoked once.

Ex. surprising bookkeeping, which would be even more surprising if you didn't know the entire setup of the request: https://github.com/openzipkin/zipkin-aws/pull/135#discussion_r308051795

trustin:
I'm not sure if it's a good idea to make `toInputStream()` transfer the ownership of the underlying buffer. The method name implies it makes a conversion, not any sort of ownership transfer such as 'move'.

anuraaga:
I think I still agree with @trustin's suggestion. It's common to treat a response as one-shot, but we can support multi-shot with this approach and I think the javadoc is very explicit on the needed bookkeeping.

anuraaga:
I filed https://github.com/line/armeria/issues/1936 though which may help either way.

adriancole:
ok let's assume that's the case. how would someone know the HttpData they
are receiving was created via aggregateWithPooledObjects?

don't you think this special knowledge requirement is dangerous or at least
dodgy?

anuraaga:
`toInputStream` and `ReferenceCountUtil` can help write code that is agnostic to the type and and #1936 should help more. For objects that need to be closed, special knowledge is required in any case so I personally still don't see a need to have the input stream release the buffer completely.

adriancole:
agree #1936 helps more

just in case folks don't have experience with other apis, it is not uncommon for closing a stream to close its corresponding body. I'll leave this here and suggest at least making the javadoc more explicit that it is always the case you can open a stream multiple times and you should always close every stream you make. if that isn't the case, add something like a oneShot to clarify.

The notes about reference hunting should be somewhere in a less scary place and obviated if it is made loudly clear that bodies are always repeatable

I would suggest still making a less surprising user api like okhttp

see https://github.com/square/okhttp/blob/d0ea471d183ffc45d061b27b540b5124f10ebefe/okhttp/src/main/java/okhttp3/ResponseBody.kt#L37-L59

PR #1920, opened 2019-07-22
Update @ConsumesJson can consume 'appplication/json' by default.

whhan91:
By RFC7158, Json text is encoded in UTF-8 by default.
However @ConsumesJson can consume only "application/json; charset=utf-8"
not "application/json". Update @ConsumesJson can consume both way.

See https://tools.ietf.org/html/rfc7158#section-8.1

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/annotation/ConsumesJson.java
+++ b/core/src/main/java/com/linecorp/armeria/server/annotation/ConsumesJson.java
@@ -21,10 +21,10 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * An alias for {@code @Consumes("application/json; charset=utf-8")}.
+ * An alias for {@code @Consumes("application/json")}.
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.TYPE, ElementType.METHOD })
-@Consumes("application/json; charset=utf-8")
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/server/annotation/ConsumesJson.java
```diff
@@ -21,10 +21,12 @@
...
 
 /**
- * An alias for {@code @Consumes("application/json; charset=utf-8")}.
+ * An alias for {@code @Consumes("application/json")} and
+ * {@code @Consumes("application/json; charset=utf-8")} which is same meaning.
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.TYPE, ElementType.METHOD })
+@Consumes("application/json")
 @Consumes("application/json; charset=utf-8")
```
trustin: I'm curious if we remove the second annotation (the one with `charset=utf-8`). @whhan91, could you check if everything works fine after removing the second annotation?

minwoox: Have checked it, and we can remove `@Consumes("application/json; charset=utf-8")` and just leave `@Consumes("application/json")`.
Could you update it please?

whhan91: Done!

PR #1924, opened 2019-07-24
Propagate IllegalStateException when calling makeContextAware

minwoox:
Motivation:
In `RequestContext.makeContextAware(future)`, if `pushIfAbsent()` throws the `IllegalStateException` because of pushing
different `RequestContext` to the thread local, it just swallows the exception.

Modification:
- Complete the future with the exception.

Result:
- The exception is propagated to the returned future when calling `makeContextAware()`


babjo:
Thank you!

minwoox:
Thanks for reviewing and reporting!

PR #1925, opened 2019-07-24
Add SimpleDecoratingHttpService

### LINKED ISSUE #1881: Add SimpleDecorating(Http|Rpc)Service
So that a user does not have to add type parameters when implementing `SimpleDecoratingService`.
Also, it will be good if we have `RpcService extends Service<RpcRequest, RpcResponse>` as we have `HttpService extends Service<HttpRequest, HttpResponse>`

minwoox:
Motivation:
A user has to specify type parameter when creating `SimpleDecoratingService`:
```
new SimpleDecoratingService<HttpRequest, HttpResponse>(delegate) {...}
new SimpleDecoratingService<RpcRequest, RpcResponse>(delegate) {...}
```
If we provide `SimpleDecoratingHttpService` and `SimpleDecoratingRpcService`,
it will make it simple.

Modifications:
- Add `SimpleDecoratingHttpService` and `SimpleDecoratingRpcService`
- Add `SimpleDecoratingHttpClient` and `SimpleDecoratingRpcClient`
- Add `RpcService` (we have `HttpService`)
- Use the classes where applicable

Result:
- Close #1881

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/server/ServiceTest.java
+++ b/core/src/test/java/com/linecorp/armeria/server/ServiceTest.java
@@ -79,7 +79,7 @@ public class ServiceTest {
         return (Class<Service<?, ?>>) service.getClass();
     }
 
-    private static final class FooService implements Service<RpcRequest, RpcResponse> {
+    private static final class FooService implements RpcService {
 
         ServiceConfig cfg;
 
@@ -95,8 +95,8 @@ public class ServiceTest {
         }
     }
 
-    public static class FooServiceDecorator extends SimpleDecoratingService<RpcRequest, RpcResponse> {
-        public FooServiceDecorator(Service<RpcRequest, RpcResponse> delegate) {
+    public static class FooServiceDecorator extends SimpleDecoratingRpcService {
+        public FooServiceDecorator(RpcService delegate) {
             super(delegate);
         }
 
@@ -107,8 +107,7 @@ public class ServiceTest {
     }
 
     public static class BadFooServiceDecorator extends FooServiceDecorator {
-        public BadFooServiceDecorator(Service<RpcRequest, RpcResponse> delegate,
-                                      @SuppressWarnings("unused") Object unused) {
+        public BadFooServiceDecorator(RpcService delegate, @SuppressWarnings("unused") Object unused) {
             super(delegate);
         }
     }
--- a/core/src/test/java/com/linecorp/armeria/server/ServiceTest.java
+++ b/core/src/test/java/com/linecorp/armeria/server/ServiceTest.java
@@ -96,7 +96,7 @@ public class ServiceTest {
     }
 
     public static class FooServiceDecorator extends SimpleDecoratingRpcService {
-        public FooServiceDecorator(RpcService delegate) {
+        public FooServiceDecorator(Service<RpcRequest, RpcResponse> delegate) {
             super(delegate);
         }
 
@@ -107,7 +107,8 @@ public class ServiceTest {
     }
 
     public static class BadFooServiceDecorator extends FooServiceDecorator {
-        public BadFooServiceDecorator(RpcService delegate, @SuppressWarnings("unused") Object unused) {
+        public BadFooServiceDecorator(Service<RpcRequest, RpcResponse> delegate,
+                                      @SuppressWarnings("unused") Object unused) {
             super(delegate);
         }
     }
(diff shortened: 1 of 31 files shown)


core/src/main/java/com/linecorp/armeria/client/SimpleDecoratingClient.java
```diff
@@ -18,10 +18,9 @@
 
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Response;
-import com.linecorp.armeria.server.Service;
 
 /**
- * Decorates a {@link Client}. Use {@link DecoratingClient} if your {@link Service} has different
+ * Decorates a {@link Client}. Use {@link DecoratingClient} if your {@link Client} has different
```
trustin: Thanks!

thrift/src/test/java/com/linecorp/armeria/it/thrift/ThriftDynamicTimeoutTest.java
```diff
@@ -127,9 +127,9 @@ void testDisabledTimeout(Function<
         }
     }
 
-    private static final class DynamicTimeoutService extends SimpleDecoratingService<RpcRequest, RpcResponse> {
+    private static final class DynamicTimeoutService extends SimpleDecoratingRpcService {
 
-        DynamicTimeoutService(Service<RpcRequest, RpcResponse> delegate) {
+        DynamicTimeoutService(RpcService delegate) {
```
trustin: Should accept `Service<RpcRequest, RpcResponse>` which is the broadest type.

thrift/src/test/java/com/linecorp/armeria/it/thrift/ThriftDynamicTimeoutTest.java
```diff
@@ -141,10 +141,9 @@ public RpcResponse serve(ServiceRequestContext ctx, RpcRequest req) throws Excep
         }
     }
 
-    private static final class TimeoutDisablingService
-            extends SimpleDecoratingService<RpcRequest, RpcResponse> {
+    private static final class TimeoutDisablingService extends SimpleDecoratingRpcService {
 
-        TimeoutDisablingService(Service<RpcRequest, RpcResponse> delegate) {
+        TimeoutDisablingService(RpcService delegate) {
```
trustin: Should accept `Service<RpcRequest, RpcResponse>` which is the broadest type.

core/src/test/java/com/linecorp/armeria/server/ServiceTest.java
```diff
@@ -107,8 +107,7 @@ public RpcResponse serve(ServiceRequestContext ctx, RpcRequest req) throws Excep
     }
 
     public static class BadFooServiceDecorator extends FooServiceDecorator {
-        public BadFooServiceDecorator(Service<RpcRequest, RpcResponse> delegate,
-                                      @SuppressWarnings("unused") Object unused) {
+        public BadFooServiceDecorator(RpcService delegate, @SuppressWarnings("unused") Object unused) {
```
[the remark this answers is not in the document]
minwoox: Ah, forgot to rever it. I was actually trying to change the constructor to `SimpleDecoratingRpcService(RpcService delegate)`.
But I realized that it couldn't be. So I revert the constructor, but not for this.
Thank you!

minwoox:
like this? https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/annotation/DecoratorFactoryFunction.java#L34
Perhaps we can remove all the `Service<Request, Response>` and make use `HttpService` and `RpcService`?
Some of the `SimpleDecoratingService` such as `RetryingClient` cannot be `SimpleDecorating(Http|Rpc)Service`, but we can perhaps make the subclass to extend `HttpService` or `RpcService`.

anuraaga:
That idea SGTM

minwoox:
Ah, I found out that we need to split some classes. For example
`LoggingService` which extends `SimpleDecoratingService` should be split 
`LoggingHttpService extends SimpleDecoratingHttpService` and
`LoggingRpcService extends SimpleDecoratingRpcService`.
I think it's okay to have them, but not sure it's worth. WDYT?

minwoox:
@anuraaga Anyway I really like the idea, let me try it later when I get some time. 😄

minwoox:
Create #1926

minwoox:
@trustin All fixed. PTAL 🙇

minwoox:
Thanks for reviewing!

PR #1927, opened 2019-07-25
Migrate docs client to webpack-dev-server.

anuraaga:
webpack-dev-server was deprecated, but then undeprecated and the replacement webpack-serve itself is deprecated.

Also adds a redirect to the path with a trailing slash to help with the confusing behavior we saw a while ago.


ikhoon:
@anuraaga Thanks for nice work!
Could you update the license file?
`yarn licenses generate-disclaimer --prod > ../licenses/web-licenses.txt`

anuraaga:
This only affects dev dependencies so the bundle's dependencies, and therefore licenses, aren't changed.

ikhoon:
@anuraaga Ahh, I see. Thank you for migrating  `webpack-dev-server`!

trustin:
Thanks a lot, @anuraaga !

PR #1929, opened 2019-07-26
Simplify a glob path mapping with prefix into a glob

trustin:
Related: #1906
Motivation:

When a user builds a `Route` with `pathWithPrefix()`, the following
route:

    pathWithPrefix("/prefix", "glob:foo")

can be simplified into:

    path("glob:/prefix/**/foo")

only except that the `**` must not be captured.

Modifications:

- Simplify a glob path mapping with prefix into a simple glob path
  mapping, not into `PrefixPathMapping`.
- Add `numGroupsToSkip` option to `GlobPathMapping` so that we can skip
  the `**` group we added implicitly.
- Miscellaneous:
  - Address the comments which were not addressed in #1906
    - Javadoc fixes
    - Removal of redundant `public` modifiers
    - Do not insert between `prefix` and `regex` when generating a span
      name for `PrefixPathMapping`.

Result:

- Much cleaner Brave span names and meter tags for the route with a
  prefix and a glob.
- Potentially better performance for the route with a prefix and a glob.

### DIFF
--- a/brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
+++ b/brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
@@ -48,19 +48,19 @@ class ArmeriaHttpServerAdapterTest {
     private RequestLog requestLog;
 
     @Test
-    public void path() {
+    void path() {
         when(requestLog.path()).thenReturn("/foo");
         assertThat(ArmeriaHttpServerAdapter.get().path(requestLog)).isEqualTo("/foo");
     }
 
     @Test
-    public void method() {
+    void method() {
         when(requestLog.method()).thenReturn(HttpMethod.GET);
         assertThat(ArmeriaHttpServerAdapter.get().method(requestLog)).isEqualTo("GET");
     }
 
     @Test
-    public void url() {
+    void url() {
         when(requestLog.isAvailable(RequestLogAvailability.SCHEME)).thenReturn(true);
         when(requestLog.isAvailable(RequestLogAvailability.REQUEST_HEADERS)).thenReturn(true);
         when(requestLog.scheme()).thenReturn(Scheme.of(SerializationFormat.NONE, SessionProtocol.HTTP));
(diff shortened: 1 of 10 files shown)


trustin:
@adriancole PTAL :wink:

minwoox:
https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/Route.java#L93
Could you update this please?

brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
```diff
@@ -151,26 +151,34 @@ public void parseClientIpAndPort() throws Exception {
...
     }
 
     @Test
-    public void route_pathWithPrefix() {
+    void route_pathWithPrefix_glob() {
         final ServiceRequestContext context = mock(ServiceRequestContext.class);
         when(requestLog.context()).thenReturn(context);
         when(context.route()).thenReturn(Route.builder().pathWithPrefix("/foo/", "glob:bar").build());
-        assertThat(ArmeriaHttpServerAdapter.get().route(requestLog)).isEqualTo("/foo/ ^/(?:.+/)?bar$");
+        assertThat(ArmeriaHttpServerAdapter.get().route(requestLog)).isEqualTo("/foo/**/bar");
```
trustin: @adriancole You will like this, right? :wink:

brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
```diff
@@ -151,26 +151,34 @@ public void parseClientIpAndPort() throws Exception {
...
+    }
+
+    @Test
+    void route_pathWithPrefix_regex() {
+        final ServiceRequestContext context = mock(ServiceRequestContext.class);
+        when(requestLog.context()).thenReturn(context);
+        when(context.route()).thenReturn(Route.builder().pathWithPrefix("/foo/", "regex:(bar|baz)").build());
+        assertThat(ArmeriaHttpServerAdapter.get().route(requestLog)).isEqualTo("/foo/(bar|baz)");
```
trustin: @adriancole and maybe this one, too!

trustin:
> Could you update this please?

@minwoox Done. :sweat_smile:

brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
adriancole: no..  I love it :)

brave/src/test/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerAdapterTest.java
adriancole: I'm already married, but if not.. this expression would be something I'd consider courting

PR #1935, opened 2019-07-29
Add a MockWebServerExtension, inspired by okhttp, which can allow tes…

### LINKED ISSUE #1884: Add something like okhttp's MockWebServer
While working on https://github.com/openzipkin/zipkin/pull/2653/files I realized that okhttp's `MockWebServer` is very convenient when writing clients (not servers, where it isn't really needed). It might be nice if Armeria had a similar testing class. It could just wrap the existing rules and add queues for mocked responses / captured requests similar to that PR.

https://github.com/square/okhttp/tree/master/mockwebserver

anuraaga:
…ting client code by conveniently mocking responses and examining requests.

At least in this version, I don't want to implement every knob that the okhttp version has, but added the delay knob since I've already worked on code that uses it.

Reference:
https://square.github.io/okhttp/4.x/mockwebserver/okhttp3.mockwebserver/-mock-response/

Fixes #1884 

/cc @adriancole


anuraaga:
I didn't implement a JUnit4 version since I don't think it can work well with a `@ClassRule`, as we need to reset state before each test method. JUnit5 is nice by allowing us to only start a single server, but also resetting state each test.

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockResponseBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+     * Sets a time to wait before returning response content. Can be useful to validate client timeout behavior.
+     * This value does not affect the time that headers are returned, to set that use
+     * {@link #headersDelay(Duration)}.
+     */
+    public MockResponseBuilder contentDelay(Duration bodyDelay) {
+        this.bodyDelay = bodyDelay;
+        return this;
+    }
+
+    /**
```
adriancole: rather than repeat "validate client timeout behaviour" I would use an example relevant to the method. ex different text for headers, content and trailers ex one use case. I'm not sure that many people are even aware of trailers existence so that is probably an especially good one to note.

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockWebServerExtension.java
```diff
@@ -0,0 +1,205 @@
...
+     * Enqueues the {@link AggregatedHttpResponse} to return to a client of this {@link MockWebServerExtension}.
+     * Multiple calls will return multiple responses in order.
+     */
+    public MockWebServerExtension enqueue(AggregatedHttpResponse response) {
+        mockResponses.add(MockResponse.of(response));
+        return this;
+    }
+
+    /**
+     * Returns the next {@link RecordedRequest} the server received. Call this method multiple times to retrieve
```
adriancole: I would make this blocking and update docs accordingly. it is much easier to write tests with blocking gets vs having strange awaitility wrapped race conditions

anuraaga: Yup meant to block, fixed.

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockResponseBuilder.java
anuraaga: I ended up with these three methods copying the API from okhttp and adding trailers for symmetry. But I can't think of a use case to separate them so don't know how to do that - instead just removed the knobs to have a single delay which should be plenty.

adriancole:
One use case I'd like to satisfy is health check going red to green on account of time. So, I think any method with delay will help with this. easier to add methods later anyway. thx

ikhoon:
@anuraaga Thanks for your always awesome PRs!
> At least in this version, I don't want to implement every knob that the okhttp version has, but added the delay knob since I've already worked on code that uses it.

I agreed with your opinion. This `MockWebServerExtension` is simple and easy to use based on `Queue`. 👍

Additionally, I was learning of okhttp's mockwebsever, [`Dispatcher`](https://github.com/square/okhttp/tree/master/mockwebserver#dispatcher) also looks good when a user needs more dynamic dispatch. Next time I'd like to add this feature :-)

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockResponseBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+     * Sets a time to wait before returning the response. Can be useful to validate client timeout behavior.
+     */
+    public MockResponseBuilder delay(int amount, TimeUnit timeUnit) {
+        this.delay = Duration.ofMillis(timeUnit.toMillis(amount));
+        return this;
+    }
```
ikhoon: nit; I think, conventionally, we have handled time unit like the following:
```
delay(Duration duration);
delayMillis(long duration);
```

https://github.com/line/armeria/blob/92919ff6b4883e5500124b8b426f44ea1d5457bc/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java#L154-L165

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockWebServerExtension.java
```diff
@@ -0,0 +1,174 @@
...
+
+        configureServer(sb);
+    }
+
+    /**
+     * Configures the {@link ServerBuilder} beyond the defaults of enabling HTTPs and registering a service for
+     * handling enqueued responses. Override this method to configure advanced behavior such as client
+     * authentication and timeouts. Don't call any of the {@code service*} methods, even if you do they will be
+     * ignored as {@link MockWebServerExtension} can only serve responses registered with
+     * {@link MockWebService#enqueue(MockResponse)}.
```
minwoox: `{@link MockWebServerExtension#enqueue(MockResponse)}`?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockWebServerExtension.java
```diff
@@ -0,0 +1,174 @@
...
+                }
+
+                final Duration delay = mockResponse.delay();
+
+                final AggregatedHttpResponse response = mockResponse.response();
+
+                if (delay.isZero()) {
+                    return HttpResponse.of(response);
+                } else {
+                    CompletableFuture<HttpResponse> responseFuture = new CompletableFuture<>();
```
minwoox: nit: final?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockWebServerExtension.java
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * <p>Note, tests in a class using {@link MockWebServerExtension} cannot be run concurrently, i.e., do not set
+ * {@link org.junit.jupiter.api.parallel.Execution} to
+ * {@link org.junit.jupiter.api.parallel.ExecutionMode#CONCURRENT}.
+ *
+ * <p>For example, <pre>{@code
+ *
+ * class MyTest {
+ *
+ * {@literal @}RegisterExtension
```
trustin: - You can use `>` prefix:
  ```
   * <pre>{@code
   * > class MyTest {
   * >     @RegisterExtension
   * >     ...
  ```
- Indentation?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockResponse.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public static MockResponse of(AggregatedHttpResponse response) {
+        return builder(response).build();
+    }
+
+    /**
+     * Returns a {@link MockResponseBuilder} which will return the provided {@link AggregatedHttpResponse} and
+     * can be additionally configured.
+     */
+    public static MockResponseBuilder builder(AggregatedHttpResponse response) {
```
trustin: Could `of()` and `builder()` also accept an `HttpResponse`?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockWebServerExtension.java
```diff
@@ -0,0 +1,174 @@
...
+    public MockWebServerExtension enqueue(MockResponse response) {
+        mockResponses.add(response);
+        return this;
+    }
+
+    /**
+     * Enqueues the {@link AggregatedHttpResponse} to return to a client of this {@link MockWebServerExtension}.
+     * Multiple calls will return multiple responses in order.
+     */
+    public MockWebServerExtension enqueue(AggregatedHttpResponse response) {
```
trustin: Ditto - Could we also accept an `HttpResponse`?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockWebServerExtension.java
```diff
@@ -0,0 +1,174 @@
...
+     * Returns the next {@link RecordedRequest} the server received. Call this method multiple times to retrieve
+     * the requests retrieved, in order. Will block up to 10 seconds waiting for a request.
+     */
+    @Nullable
+    public RecordedRequest takeRequest() {
+        try {
+            return recordedRequests.poll(10, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            logger.warn("Could not take request.", e);
+            Exceptions.throwUnsafely(e);
```
trustin: - This is a double notification - logging and throwing. Let's do only one.
- Wouldn't it be surprising to a user when an `InterruptedException` is raised from a method without a `throws` clause?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockWebServerExtension.java
```diff
@@ -0,0 +1,174 @@
...
+            return null;  // Unreachable.
+        }
+    }
+
+    @Override
+    protected final void configure(ServerBuilder sb) throws Exception {
+        sb.http(0);
+        sb.https(0);
+        sb.tlsSelfSigned();
```
trustin: Not necessarily a problem to solve in this pull request, but I'm curious if we can generate a self-signed certificate for all extensions only once.

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockResponse.java
```diff
@@ -0,0 +1,65 @@
...
+package com.linecorp.armeria.testing.junit.server.mockwebserver;
+
+import java.time.Duration;
+
+import com.linecorp.armeria.common.AggregatedHttpResponse;
+
+/**
+ * A mock response to return to a request made to {@link MockWebServerExtension}.
+ */
+public class MockResponse {
```
trustin: How aboug writing an `HttpResponse` implementation that defers the publication, which may be useful for non-testing scenarios as well, e.g. `HttpResponse.delayed(HttpResponse, Duration)` or `delayed(Supplier<HttpResponse>, Duration)`.

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/RecordedRequest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public RecordedRequest(AggregatedHttpRequest request, ServiceRequestContext context) {
+        this.request = request;
+        this.context = context;
+    }
+
+    /**
+     * The {@link AggregatedHttpRequest} received by the server.
+     */
+    public AggregatedHttpRequest getRequest() {
```
trustin: `request()`?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/RecordedRequest.java
```diff
@@ -0,0 +1,49 @@
...
+        return request;
+    }
+
+    /**
+     * The {@link ServiceRequestContext} created when handling the request. Can be used to, e.g., check whether
+     * the request uses TLS.
+     */
+    public ServiceRequestContext getContext() {
```
trustin: `context()`?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/RecordedRequest.java
```diff
@@ -0,0 +1,49 @@
...
+
+package com.linecorp.armeria.testing.junit.server.mockwebserver;
+
+import com.linecorp.armeria.common.AggregatedHttpRequest;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+/**
+ * A request that has been made to a {@link MockWebServerExtension}.
+ */
+public class RecordedRequest {
```
trustin: `final` and `toString()`

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * A mock server for use when testing clients.
+ */
+@NonNullByDefault
+package com.linecorp.armeria.testing.junit.server.mockwebserver;
```
trustin: How about just `mock`? Because Armeria is for web and the parent package name indicates it's for server-side.

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mockwebserver/MockResponse.java
anuraaga: Good idea - went with that and removed `MockResponse` for now since if we don't have any parameters besides delay, there's no point for it. If we need a test-only interface later can add it back.

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mock/MockWebServerExtension.java
```diff
@@ -54,26 +55,26 @@
...
+ * >       server.enqueue(AggregatedHttpResponse.of(HttpStatus.OK));
+ * >       server.enqueue(AggregatedHttpResponse.of(HttpStatus.FORBIDDEN));
+ * >
+ * >       assertThat(client.get("/").aggregate().join().status()).isEqualTo(HttpStatus.OK);
+ * >       assertThat(client.get("/bad").aggregate().join().status()).isEqualTo(HttpStatus.FORBIDDEN);
+ * >
+ * >       assertThat(server.takeRequest().path()).isEqualTo("/");
+ * >       assertThat(server.takeRequest().path()).isEqualTo("/bad");
+ * >   }
+ * >
```
trustin: This empty line could be removed.

testing/junit/src/test/java/com/linecorp/armeria/testing/junit/server/mock/MockWebServiceExtensionTest.java
```diff
@@ -103,21 +104,15 @@ void normal() {
 
     @Test
     void delay() {
-        server.enqueue(MockResponse.builder(AggregatedHttpResponse.of(HttpStatus.OK))
-                                   .delay(Duration.ofMillis(200))
-                                   .build());
-        server.enqueue(MockResponse.builder(AggregatedHttpResponse.of(HttpStatus.OK))
-                                   .delayMillis(300)
-                                   .build());
+        server.enqueue(HttpResponse.delayed(HttpResponse.of(AggregatedHttpResponse.of(HttpStatus.OK)),
```
trustin: What do you think about adding `AggregatedHttpResponse.delayed()` as well?

trustin: Ugh, actually, `HttpResponse.delayed(AggregatedHttpResponse)`.

anuraaga: Sure

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -78,6 +80,50 @@ static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
...
+        requireNonNull(response, "response");
+        requireNonNull(delay, "delay");
+        return delayed(response, delay, CommonPools.workerGroup().next());
+    }
+
+    /**
+     * Creates a new HTTP response that delegates to the provided {@link HttpResponse}, beginning publishing
+     * after {@code delay} has passed from the provided {@link EventExecutor}.
+     */
+    static HttpResponse delayed(HttpResponse response, Duration delay, EventExecutor executor) {
```
trustin: Could use `ScheduledExecutorService` instead of `EventExecutor`?

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -78,6 +80,50 @@ static HttpResponse from(CompletionStage<? extends HttpResponse> stage) {
...
+        requireNonNull(response, "response");
+        requireNonNull(delay, "delay");
+        return delayed(HttpResponse.of(response), delay);
+    }
+
+    /**
+     * Creates a new HTTP response that delegates to the provided {@link AggregatedHttpResponse}, beginning
+     * publishing after {@code delay} has passed from the provided {@link EventExecutor}.
+     */
+    static HttpResponse delayed(AggregatedHttpResponse response, Duration delay, EventExecutor executor) {
```
trustin: Could use `ScheduledExecutorService` instead of `EventExecutor`?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mock/MockWebServerExtension.java
```diff
@@ -0,0 +1,182 @@
...
+        // Do nothing by default.
+    }
+
+    @Override
+    public final void beforeTestExecution(ExtensionContext context) {
+        mockResponses.clear();
+        recordedRequests.clear();
+    }
+
+    private class MockWebService implements Service<HttpRequest, HttpResponse> {
```
trustin: `implements HttpService`?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mock/RecordedRequest.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * A request that has been made to a {@link MockWebServerExtension}.
+ */
+public final class RecordedRequest {
+
+    private final AggregatedHttpRequest request;
+    private final ServiceRequestContext context;
+
+    public RecordedRequest(AggregatedHttpRequest request, ServiceRequestContext context) {
+        this.request = request;
```
ikhoon: nit: `requireNonNull`?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mock/RecordedRequest.java
```diff
@@ -0,0 +1,84 @@
...
+    public boolean equals(@Nullable Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (!(o instanceof RecordedRequest)) {
+            return false;
+        }
+
+        RecordedRequest that = (RecordedRequest) o;
```
minwoox: nit: final?

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mock/RecordedRequest.java
anuraaga: Thanks - added, though realized this constructor didn't need to be public anyways

testing/junit/src/main/java/com/linecorp/armeria/testing/junit/server/mock/RecordedRequest.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * A request that has been made to a {@link MockWebServerExtension}.
+ */
+public final class RecordedRequest {
+
+    private final AggregatedHttpRequest request;
+    private final ServiceRequestContext context;
+
+    public RecordedRequest(AggregatedHttpRequest request, ServiceRequestContext context) {
```
trustin: Sorry, but how about putting context first for consistency with other methods in our API?

anuraaga: Sure

trustin:
Any other comments, @ikhoon ?

PR #1937, opened 2019-07-30
Enable DeferredHttpResponse to have a defaultSubscriberExecutor

minwoox:
Motivation:
If a user calls `client.get(...).aggregate()`, it uses the `EventLoop` which is used for writing the elements
to the `HttpResponse` to subscribe. So the logic executed by one `EventLoop` seamlessly. However, if there's
a decorator which uses `DeferredHttpResponse` such as `ConcurrencyLimitingHttpClient` or `RetryingClient`,
it's executed not by one `EventLoop` but two `EventLoop`s which switching situation happens.

Modifications:
- Add `HttpResponse.from(stage, defaultSubscriberExecutor)`.
- Add a constructor which takes an `EventExecutor` to `DeferredHttpResponse`.
  - `DeferredHttpResponse` is public but it's deprecated a long time ago, so perhaps don't need to have a class which extends it.

Result:
- Aggregating `HttpResponse` becomes more efficient when `ConcurrencyLimitingHttpClient` or `RetryingClient` is used.
- Ready to move making connections to the beginning of client execution

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingHttpClient.java
@@ -25,6 +25,8 @@ import com.linecorp.armeria.client.ClientRequestContext;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 
+import io.netty.channel.EventLoop;
+
 /**
  * A {@link Client} decorator that limits the concurrent number of active HTTP requests.
  *
@@ -67,9 +69,10 @@ public final class ConcurrencyLimitingHttpClient extends ConcurrencyLimitingClie
 
     @Override
     protected Deferred<HttpResponse> defer(ClientRequestContext ctx, HttpRequest req) throws Exception {
+        final EventLoop eventLoop = ctx.eventLoop();
         return new Deferred<HttpResponse>() {
             private final CompletableFuture<HttpResponse> responseFuture = new CompletableFuture<>();
-            private final HttpResponse res = HttpResponse.from(responseFuture, ctx.eventLoop());
+            private final HttpResponse res = HttpResponse.from(responseFuture, eventLoop);
 
             @Override
             public HttpResponse response() {
--- a/core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingHttpClient.java
@@ -69,7 +69,7 @@ public final class ConcurrencyLimitingHttpClient extends ConcurrencyLimitingClie
     protected Deferred<HttpResponse> defer(ClientRequestContext ctx, HttpRequest req) throws Exception {
         return new Deferred<HttpResponse>() {
             private final CompletableFuture<HttpResponse> responseFuture = new CompletableFuture<>();
-            private final HttpResponse res = HttpResponse.from(responseFuture);
+            private final HttpResponse res = HttpResponse.from(responseFuture, ctx.eventLoop());
 
             @Override
             public HttpResponse response() {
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/linecorp/armeria/common/HttpResponseUtil.java
```diff
@@ -0,0 +1,43 @@
...
+import static java.util.Objects.requireNonNull;
+
+import java.util.concurrent.CompletionStage;
+
+import com.linecorp.armeria.common.util.Exceptions;
+
+final class HttpResponseUtil {
+
+    static void delegateWhenStageComplete(CompletionStage<? extends HttpResponse> stage,
+                                          DeferredHttpResponse res) {
```
ikhoon: Do we need to extract this into a Util class than `private static` in `HttpResponse`?

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -61,20 +62,30 @@ static HttpResponseWriter streaming() {
...
+     * {@link CompletionStage}. If the specified {@link CompletionStage} fails, the returned response will be
+     * closed with the same cause as well.
+     *
+     * @param stage the {@link CompletionStage} which will produce the actual {@link HttpResponse}
+     * @param defaultSubscriberExecutor the {@link EventExecutor} which will be used when a user subscribes
+     *                                  the returned {@link HttpResponse} using {@link #subscribe(Subscriber)}
+     *                                  or {@link #subscribe(Subscriber, SubscriptionOption...)}.
+     */
+    static HttpResponse from(CompletionStage<? extends HttpResponse> stage,
+                             EventExecutor defaultSubscriberExecutor) {
```
ikhoon: nit: `defaultSubscriberExecutor` -> `subscriberExecutor`? 
I think this is a specified executor than default. :-)

core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingHttpClient.java
```diff
@@ -69,7 +69,7 @@ private ConcurrencyLimitingHttpClient(Client<HttpRequest, HttpResponse> delegate
     protected Deferred<HttpResponse> defer(ClientRequestContext ctx, HttpRequest req) throws Exception {
         return new Deferred<HttpResponse>() {
             private final CompletableFuture<HttpResponse> responseFuture = new CompletableFuture<>();
-            private final HttpResponse res = HttpResponse.from(responseFuture);
+            private final HttpResponse res = HttpResponse.from(responseFuture, ctx.eventLoop());
```
trustin: How about extracting `ctx.eventLoop()` into a final local variable of `defer()`, so that the anonymous `Deferred` does not have the reference to the context but only to the event loop?

core/src/main/java/com/linecorp/armeria/common/HttpResponseUtil.java
minwoox: Couldn't have a private method in interface, yet. 😢

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
minwoox: Fixed. Thanks!

core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingHttpClient.java
minwoox: Fixed, thanks!

core/src/main/java/com/linecorp/armeria/common/HttpResponseUtil.java
ikhoon: Ah, sorry, I realized that is java 9 feature.
Then how about using a nested static class? 😆
https://dzone.com/articles/java-8-declare-private-and-protected-methods-in-in

minwoox: As we discussed, this will expose one more class which will bring us another breaking change when we force the minimum Java version to 9, so let's just leave it as it is. 😉

trustin:
Thanks, @minwoox !

minwoox:
Thanks for reviewing!

PR #1938, opened 2019-07-30
[WIP] Add HttpData#close, HttpData#toByteBuffer, and docs.

### LINKED ISSUE #1936: Make HttpData and/or AggregatedHttp* AutoCloseable
okhttp makes their `ResponseBody` `Closeable`, which is convenient for pooled objects.

https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/

```java
try (AggregatedHttpResponse response = client.get("/").aggregateWithPooledObjects().join()) {
  System.out.println(response.contentUtf8());
}
```

instead of

```java
AggregatedHttpResponse response = client.get("/").aggregateWithPooledObjects().join()
try {
  System.out.println(response.contentUtf8());
} finally {
  ReferenceCountUtil.safeRelease(response.content());
}
```

However, users should still be free to use non-pooled without closing as we want the default case to be easy to use.
```java
AggregatedHttpResponse response = client.get("/").aggregate().join();
System.out.println(response.contentUtf8());
```
is equivalent to this which implements `close()` as a no-op.
```java
try (AggregatedHttpResponse response = client.get("/").aggregate().join()) {
  System.out.println(response.contentUtf8());
}
```

If we phrase the Javadoc well, I think we can keep the simple API while also providing this simpler mechanism to advanced users.

anuraaga:
Also changes `ByteBufHttpData#toInputStream` to not create a new reference / release one. I think for usages of `HttpData`, it's clear enough that all the operations should happen in the scope of the try-with-resources so no need to worry about transfer of `InputStream` out of the block.

Fixes #1936

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpData.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpData.java
@@ -34,7 +34,6 @@ import com.linecorp.armeria.unsafe.ByteBufHttpData;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufUtil;
-import io.netty.util.ReferenceCountUtil;
 import it.unimi.dsi.fastutil.io.FastByteArrayInputStream;
 
 /**
@@ -430,7 +429,8 @@ public interface HttpData extends HttpObject, AutoCloseable {
      * A no-op in most cases. This method will only have an effect when using pooled objects, e.g., when using
      * methods like {@link HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}. If not using pooled
      * objects, you can ignore this method. If you are, please see the documentation at
-     * {@link com.linecorp.armeria.unsafe.ByteBufHttpData}.
+     * {@link com.linecorp.armeria.unsafe.ByteBufHttpData}. {@link #close()} can be called multiple times, only
+     * the first will have an effect.
      */
     @Override
     default void close() {
--- a/core/src/main/java/com/linecorp/armeria/common/HttpData.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpData.java
@@ -388,12 +388,6 @@ public interface HttpData extends HttpObject, AutoCloseable {
 
     /**
      * Returns a new {@link InputStream} that is sourced from this data.
-     *
-     * <p>Note, if this {@link HttpData} is pooled (e.g., it is the result of a call to
-     * {@link HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}), then this {@link InputStream} will
-     * increase the reference count of the underlying buffer. Make sure to call {@link InputStream#close()},
-     * usually using a try-with-resources invocation, to release this extra reference. And as usual, don't
-     * forget to call {@link ReferenceCountUtil#release(Object)} on this {@link HttpData} itself too.
      */
     default InputStream toInputStream() {
         return new FastByteArrayInputStream(array());
--- a/core/src/main/java/com/linecorp/armeria/common/HttpData.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpData.java
@@ -44,7 +44,7 @@ import it.unimi.dsi.fastutil.io.FastByteArrayInputStream;
  * <p>Implementations should generally extend {@link AbstractHttpData} to interact with other {@link HttpData}
  * implementations.
  */
-public interface HttpData extends HttpObject {
+public interface HttpData extends HttpObject, AutoCloseable {
 
     /**
      * Empty HTTP/2 data.
@@ -422,5 +422,24 @@ public interface HttpData extends HttpObject {
      */
     default Reader toReaderAscii() {
         return toReader(StandardCharsets.US_ASCII);
+    }
+
+    /**
+     * Returns a new {@link ByteBuffer} that wraps this data. Mutations to the {@link ByteBuffer} will be
+     * reflected in this {@link HttpData} and vice versa.
+     */
+    default ByteBuffer toByteBuffer() {
+        return ByteBuffer.wrap(array());
+    }
+
+    /**
+     * A no-op in most cases. This method will only have an effect when using pooled objects, e.g., when using
+     * methods like {@link HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}. If not using pooled
+     * objects, you can ignore this method. If you are, please see the documentation at
+     * {@link com.linecorp.armeria.unsafe.ByteBufHttpData}.
+     */
+    @Override
+    default void close() {
+        // No-op
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
@@ -369,7 +369,8 @@ public interface HttpRequest extends Request, StreamMessage<HttpObject> {
      * Aggregates this request. The returned {@link CompletableFuture} will be notified when the content and
      * the trailers of the request is received fully. {@link AggregatedHttpRequest#content()} will
      * return a pooled object, and the caller must ensure to release it. If you don't know what this means,
-     * use {@link #aggregate()}.
+     * use {@link #aggregate()}. If you are considering using this method, please read through the documentation
+     * at {@link com.linecorp.armeria.unsafe.ByteBufHttpData} first.
      */
     default CompletableFuture<AggregatedHttpRequest> aggregateWithPooledObjects(ByteBufAllocator alloc) {
         requireNonNull(alloc, "alloc");
@@ -383,7 +384,8 @@ public interface HttpRequest extends Request, StreamMessage<HttpObject> {
      * Aggregates this request. The returned {@link CompletableFuture} will be notified when the content and
      * the trailers of the request is received fully. {@link AggregatedHttpRequest#content()} will
      * return a pooled object, and the caller must ensure to release it. If you don't know what this means,
-     * use {@link #aggregate()}.
+     * use {@link #aggregate()}. If you are considering using this method, please read through the documentation
+     * at {@link com.linecorp.armeria.unsafe.ByteBufHttpData} first.
      */
     default CompletableFuture<AggregatedHttpRequest> aggregateWithPooledObjects(
             EventExecutor executor, ByteBufAllocator alloc) {
--- a/core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
+++ b/core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
@@ -57,7 +57,7 @@ import io.netty.buffer.Unpooled;
  * failure to release the reference will result in a memory leak and poor performance. You must make sure to do
  * this by calling {@link HttpData#close()}, usually in a try-with-resources structure to avoid side effects.
  *
- * <p>For example, <pre>{code
+ * <p>For example, <pre>{@code
  *
  * HttpResponse response = client.get("/");
  * response.aggregateWithPooledObjects(ctx.alloc(), ctx.executor())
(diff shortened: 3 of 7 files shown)


core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -388,12 +388,6 @@ default String toStringAscii() {
 
     /**
      * Returns a new {@link InputStream} that is sourced from this data.
-     *
```
adriancole: muchos grassyass

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -423,4 +417,23 @@ default Reader toReaderUtf8() {
     default Reader toReaderAscii() {
         return toReader(StandardCharsets.US_ASCII);
     }
+
+    /**
+     * Returns a new {@link ByteBuffer} that wraps this data. Mutations to the {@link ByteBuffer} will be
```
adriancole: seems powerful if actually works!

core/src/main/java/com/linecorp/armeria/common/HttpData.java
```diff
@@ -423,4 +417,23 @@ default Reader toReaderUtf8() {
...
+    }
+
+    /**
+     * A no-op in most cases. This method will only have an effect when using pooled objects, e.g., when using
+     * methods like {@link HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}. If not using pooled
+     * objects, you can ignore this method. If you are, please see the documentation at
+     * {@link com.linecorp.armeria.unsafe.ByteBufHttpData}.
+     */
+    @Override
+    default void close() {
```
adriancole: any terror in invoking twice?

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
```diff
@@ -369,7 +369,8 @@ default MediaType contentType() {
      * Aggregates this request. The returned {@link CompletableFuture} will be notified when the content and
      * the trailers of the request is received fully. {@link AggregatedHttpRequest#content()} will
      * return a pooled object, and the caller must ensure to release it. If you don't know what this means,
-     * use {@link #aggregate()}.
+     * use {@link #aggregate()}. If you are considering using this method, please read through the documentation
```
adriancole: good tip

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
...
+ *
+ * <p>The buffer backing a {@link ByteBufHttpData} is a pooled buffer - this means that it does not use normal
+ * Java garbage collection, and instead uses manual memory management using a reference count, similar to
+ * some constructs in languages like C++. Manual memory management is more fragile and not idiomatic in Java -
+ * you should only use this class in performance-sensitive situations and after being ready to deal with these
+ * very hard-to-debug issues.
+ *
+ * <p>You may interact with {@link ByteBufHttpData} when using methods that return pooled objects, such as
+ * {@link com.linecorp.armeria.common.HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}. If you don't
+ * use such methods, you will never see a {@link ByteBufHttpData} and don't need to read further.
```
adriancole: excellent break point

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
...
+ *
+ * <p>Any time you receive a {@link ByteBufHttpData} it will have a single reference that must be released -
+ * failure to release the reference will result in a memory leak and poor performance. You will want to make
```
adriancole: I would be a little stronger.

```
Any call sites of this must call {@link HttpData#close()}, usually in a try-with-resources structure to avoid side effects.
```

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
...
+ * sure to do this by calling {@link HttpData#close()}, usually in a try-with-resources structure.
+ *
+ * <p>For example, <pre>{code
+ *
+ * HttpResponse response = client.get("/");
+ * response.aggregateWithPooledObjects(ctx.alloc(), ctx.executor())
+ *     .thenApply(aggResp -> {
+ *        try (HttpData content = aggResp.content()) {
```
adriancole: add comment above try, as there is a lot of code to tell what line is important.

```
// HERE: try with resources on the content avoids reliance on downstream calls to do the same.
```

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
...
+ *                throw new IllegalArgumentException("Bad JSON: " + content.toStringUtf8());
+ *            }
+ *        }
+ *     });
+ *
+ * }</pre>
+ *
+ * <p>In this example, it is the initial {@code try (HttpData content = ...} that ensures the data is released.
+ * Calls to methods on {@link HttpData} will all work and can be called any number of times within this block.
+ * If called after the block, or a manual call to {@link HttpData#close}, these methods will either fail or
```
adriancole:
```suggestion
 * If called after the block, or a manual call to {@link HttpData#close}, these methods will fail or
```

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
...
+ * corrupt data.
+ *
+ * <p>When requesting pooled objects, it is not guaranteed that all {@link HttpData} are actually pooled, e.g.,
```
adriancole: rephrase like this is a request.

Ex. Requesting pooled objects is a request, not a mandate: it is not guaranteed...

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
...
+ * a decorator may not understand pooled objects at which points objects will be copied to the Java heap. Code
+ * should still be able to operate on either type of {@link HttpData} as long as it calls
+ * {@link HttpData#close()} and uses methods like {@link HttpData#toInputStream()},
+ * {@link HttpData#toByteBuffer()}, or {@link HttpData#toStringUtf8()} to access the content. The above example
+ * works well regardless of whether the {@link HttpData} is pooled or not.
+ *
+ * <p>In some cases, you may want to access the {@link ByteBuf} held by this {@link ByteBufHttpData}. This will
```
adriancole: use `<h3>` notes to break up this doc.

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
 /**
  * A {@link HttpData} that is backed by a {@link ByteBuf} for optimizing certain internal use cases. Not for
  * general use.
+ *
+ * <p>The buffer backing a {@link ByteBufHttpData} is a pooled buffer - this means that it does not use normal
```
adriancole: heading (I use h3) "what are pooled buffers"

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
...
+ * Java garbage collection, and instead uses manual memory management using a reference count, similar to
+ * some constructs in languages like C++. Manual memory management is more fragile and not idiomatic in Java -
+ * you should only use this class in performance-sensitive situations and after being ready to deal with these
+ * very hard-to-debug issues.
+ *
+ * <p>You may interact with {@link ByteBufHttpData} when using methods that return pooled objects, such as
+ * {@link com.linecorp.armeria.common.HttpResponse#aggregateWithPooledObjects(ByteBufAllocator)}. If you don't
+ * use such methods, you will never see a {@link ByteBufHttpData} and don't need to read further.
+ *
+ * <p>Any time you receive a {@link ByteBufHttpData} it will have a single reference that must be released -
```
adriancole: heading: impact of pooled buffers

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
adriancole: heading "pooled buffers is a request, not a mandate:

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -35,6 +37,69 @@
...
+ * final ByteBuf buf;
+ * if (data instanceof ByteBufHolder) {
+ *     buf = ((ByteBufHolder) data).content();
+ * } else {
+ *     buf = Unpooled.wrappedBuffer(data.array());
+ * }
+ *
+ * }</pre>
+ *
+ * <p>Using a {@link ByteBuf} directly is very advanced and can open up much more complicated management of
```
adriancole: heading "see also"

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
anuraaga: It seems fine as is, since otherwise we use the word `request` twice redundantly.

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -242,6 +259,9 @@ public ByteBuffer toByteBuffer() {
 
     @Override
     public void close() {
+        if (!closedUpdater.compareAndSet(this, 0, 1)) {
```
adriancole: good

minwoox:
I can't think of it either. 😢 The reason that you don't like it is that you think the `HttpData` is something that is not closeable?

ikhoon:
I also worry about top-level no-op `close()`. Many API which implement `HttpData` don't need `close()`. a `close()` method is only overrided at `ByteBufData` among `AbstractHttpData`, `ByteBufData, `ByteRangeHttpData` and `DefaultHttpData`.

How about making a delegate for making sure release instead of inheritance like:
```java
class CloseableHttpData implements HttpData, AutoCloseable {

    private HttpData httpData;

    static HttpData of(HttpData httpData) {
        return new CloseableHttpData(httpData);
    }

    CloseableHttpData(HttpData httpData) {
        this.httpData = httpData;
    }

    @Override
    public void close() {
        // Check HttpData Type
        if (httpData instanceof ByteBufData) {
            /// close stuff
        }
        ...
    }
    
    // HttpData overrides
    @Override
    public byte[] array() {
        return httpData.array();
    }
    ....

}
```

anuraaga:
I was thinking of the wrapper too, putting it into the unsafe package. I feel like if we do that we should do it for everything - move the current aggregateWithPooled methods and anything else that mentions pooled into unsafe. I think it might also imply moving some classes into the internal package though since they would probably need to become public.

Is this a good idea? It's pretty big refactoring

trustin:
I like the idea of providing an additional layer for enabling an advanced feature like this, e.g.

```java
try (PooledHttpRequest pooled = req.withPooling()) {
    pooled.aggregate()...;
}

interface PooledHttpRequest extends HttpRequest, AutoCloseable { ... }
```

What do you think?

adriancole:
@trustin I like the thinking.. I suppose it would be more telling if applied to existing call sites. Maybe something like that could go on a branch/PR and we can try applying it to real use using jitpack to serve the dependency

anuraaga:
Yeah I think it's a good idea - it'll take a lot of refactoring :) I'll put this PR on hold for now

trustin:
Superseded by #2448

Issue #1910, opened 2019-07-18
HealthCheckedEndpointGroup should probably wait for connectTimeout before throwing exception.


anuraaga:
Currently, `HttpHealthCheckedEndpointGroup` will throw an `EndpointGroupException` immediately when a request is attempted when there are no endpoints. It would be more intuitive and simpler to use if it was possible for the request to wait until `ClientFactory.connectTImeout()` before throwing the exception, allowing a health checked endpoint group to have similar connection behavior as just a single endpoint.

trustin:
That's a good idea.

trustin:
I guess we can implement this behavior like the following:

1. Define a dedicated exception that signifies there are no endpoints available.
2. Implement the delayed response somewhere in the beginning of client request execution path where the exception defined at (1) is caught.

trustin:
That being said, I guess it may not affect our current API, which means we may be able to move this to post-1.0. What do you think, @anuraaga?

anuraaga:
I agree this is just behavior change so can untarget 1.0

trustin:
We have a related issue where `RetryingClient` does not work at all when an `EndpointGroup` is empty, because the request fails even before `RetryingClient` tries anything.

To let the request be handled by a decorator even if there's a chosen endpoint, we're gonna have to change when `ClientRequestContext.endpoint()` returns `null` and to decide at which point we need to wait until an endpoint is available - Wait at the beginning or the end of invoking a decorator chain? What should we do when `RetryingClient` is involved?

trustin:
A related question from Andrey at Slack workspace:

> Hi guys, I use `RetryingHttpClient` with own implementation of `DynamicEndpointGroup` - the group periodically reads endpoints from an external source. The client retries on `UnprocessedRequestException` and this works nice but there are cases when the external source can return 0 endpoints for a short period of time (~ 10 to 30 seconds). In this case `UnprocessedRequestException` is thrown from `RoundRobinStrategy` but, obviously, there are no retries because it’s too early for retrying client to kick in. I really don’t want to have own retries on top of Armeria’s retries and think about adding a dummy artificial endpoint if there are 0 endpoints returned from the external source - so I can keep trying this “unreachable” endpoint until “good” endpoints arrive. Does this sound like a reasonable approach or maybe there are some better options?

trustin:
Although not ideal, a workaround at the moment could be make an `EndpointGroup` return `127.0.0.1:1` when there are no healthy endpoints.

[...]

PR #1939, opened 2019-07-30
Add CompositeEndpointGroup

### LINKED ISSUE #1897: Add CompositeEndpointGroup that aggregates multiple endpoint groups
I believe we're missing a simple way to register multiple endpoint groups as one - this would mostly be used when given a list of hosts that may resolve to multiple A records, each would be its own DnsAddressEndpointGroup but we'd like to round-robin over all the records. We currently have `OrElseEndpointGroup` that picks one or the other, but not one that iterates through all of them.

anuraaga:
Found while working on https://github.com/openzipkin/zipkin/blob/81a481b490399a797bc5d04c00ee7de400a4f015/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java#L527

Also makes `Endpoint` implement `EndpointGroup` as a group of itself so APIs can accept both groups and non-groups transparently.

Fixes #1897


core/src/main/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroup.java
```diff
@@ -0,0 +1,111 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.io.Closer;
+
+import com.linecorp.armeria.client.Endpoint;
+import com.linecorp.armeria.common.util.AbstractListenable;
+
+/**
+ * An {@link EndpointGroup} that merges the result of any number of other {@link EndpointGroup}s.
+ */
+public final class CompositeEndpointGroup extends AbstractListenable<List<Endpoint>> implements EndpointGroup {
```
trustin: How about making this package-private and adding `EndpointGroup.of(EndpointGroup...)` and `of(Iterable<EndpointGroup>)`?

core/src/main/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroup.java
```diff
@@ -0,0 +1,111 @@
...
+        for (EndpointGroup endpointGroup : endpointGroups) {
+            newEndpoints.addAll(endpointGroup.endpoints());
+        }
+
+        return merged = newEndpoints.build();
+    }
+
+    @Override
+    public void close() {
+        Closer closer = Closer.create();
```
trustin: nit: `final`?

core/src/main/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroup.java
```diff
@@ -0,0 +1,111 @@
...
+            return merged;
+        }
+
+        if (!dirtyUpdater.compareAndSet(this, 1, 0)) {
+            // Another thread might be updating merged at this time, but endpoint groups are allowed to take a
+            // little bit of time to reflect updates.
+            return merged;
+        }
+
+        ImmutableList.Builder<Endpoint> newEndpoints = ImmutableList.builder();
```
ikhoon: nit: `final`? +1

core/src/test/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroupTest.java
```diff
@@ -0,0 +1,53 @@
...
+class CompositeEndpointGroupTest {
+
+    private static final Endpoint FOO = Endpoint.of("foo");
+    private static final Endpoint BAR = Endpoint.of("bar");
+    private static final Endpoint CAT = Endpoint.of("cat");
+    private static final Endpoint DOG = Endpoint.of("dog");
+
+    @Test
+    void normal() {
+        DynamicEndpointGroup group1 = new DynamicEndpointGroup();
```
ikhoon: nit: final? +2

core/src/test/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroupTest.java
```diff
@@ -0,0 +1,53 @@
...
+        group1.setEndpoints(ImmutableList.of(FOO, BAR));
+        DynamicEndpointGroup group2 = new DynamicEndpointGroup();
```
ikhoon: nit: final? +3

core/src/test/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroupTest.java
```diff
@@ -0,0 +1,53 @@
...
+        group2.setEndpoints(ImmutableList.of(CAT, DOG));
+
+        CompositeEndpointGroup composite = new CompositeEndpointGroup(group1, group2);
```
ikhoon: nit: final? +4

core/src/main/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroup.java
anuraaga: Thanks - will make the change but putting this PR on hold until #1940

trustin: Merged #1940 :-)

anuraaga:
Just pushed WIP commits (now I need my deleted method back ;) )

One feeling I had while tweaking this is it might be nice if `Endpoint` was `EndpointGroup`, similar to `EventLoop` and `EventLoopGroup`.

trustin:
> One feeling I had while tweaking this is it might be nice if `Endpoint` was `EndpointGroup`, similar to `EventLoop` and `EventLoopGroup`. 

Indeed. I like the idea :-)

anuraaga:
Do you mind taking a skim of https://github.com/line/armeria/pull/1939/commits/e276167b8fa993ac6c3d5fc8eda1cf7605040bf5 ?

There are goods and bads of making `Endpoint` `EndpointGroup`, especially when writing documentation. How does it look?

anuraaga:
@trustin Just a reminder if you can give a quick skim of the general direction that'd be awesome

adriancole:
chasing up things, would be nice to have this merged for whatever next release

adriancole:
I don't know if this is the issue for it or not, but maybe we need to cache last exception so that if an endpoint group is empty we can know why then and there? here's an example https://github.com/openzipkin/zipkin/issues/2763#issuecomment-523006770

trustin:
@anuraaga Sorry I'm late. It looks OK-ish to me. The methods related with the initial endpoints look somewhat unnatural, but I think it's not that bad. What do you think, @minwoox ?

trustin:
@adriancole There's an issue for that: #1910

anuraaga:
Sorry for the huge delay, ready for another look

core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointGroup.java
```diff
@@ -41,6 +46,50 @@ static EndpointGroup empty() {
...
+     * {@code endpointGroups} can be instances of {@link Endpoint} as well, any {@link EndpointGroup}s and
+     * {@link Endpoint} will all be combined into a single {@link EndpointGroup} that contains the total set.
+     */
+    static EndpointGroup of(Iterable<EndpointGroup> endpointGroups) {
+        requireNonNull(endpointGroups, "endpointGroups");
+
+        final List<EndpointGroup> groups = new ArrayList<>();
+        final List<Endpoint> staticEndpoints = new ArrayList<>();
+        for (EndpointGroup endpointGroup : endpointGroups) {
+            if (endpointGroup instanceof Endpoint) {
```
trustin: Should we also spread the endpoints in the `StaticEndpointGroup`?

core/src/main/java/com/linecorp/armeria/client/endpoint/StaticEndpointGroup.java
```diff
@@ -26,7 +26,10 @@
 
 /**
  * A static immutable {@link EndpointGroup}.
+ *
+ * @deprecated use {@link EndpointGroup}
```
trustin: - use -> Use
- missing period
- `{@link EndpointGroup#of(EndpointGroup...)}` ?

core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointGroup.java
anuraaga: Sure also realized I was missing tests for these special cases

core/src/main/java/com/linecorp/armeria/client/endpoint/StaticEndpointGroup.java
trustin: I guess you forgot this :-)

anuraaga: Oops

core/src/test/java/com/linecorp/armeria/client/endpoint/EndpointGroupTest.java
```diff
@@ -18,20 +18,101 @@
...
 
         assertThat(emptyEndpointGroup.orElse(endpointGroup2).endpoints())
                 .isEqualTo(endpointGroup2.endpoints());
         assertThat(endpointGroup1.orElse(endpointGroup2).endpoints())
                 .isEqualTo(endpointGroup1.endpoints());
     }
+
+    @Test
+    void normal() {
+        DynamicEndpointGroup group1 = new DynamicEndpointGroup();
```
minwoox: nit: we could add `final` modifier for the variables used in this class

minwoox:
Thanks, @anuraaga!

PR #1940, opened 2019-07-30
Wait for delegate endpoints before beginning health check.

anuraaga:
Also moves logic for waiting for endpoints to `EndpointGroup`. Many of our groups just delegate, so aren't `DynamicEndpointGroup`, but users shouldn't have to know whether a group is dynamic or not to do the initial endpoint wait.

I noticed when writing tests that use a health checked client, they were extremely slow. Investigating, I realized it's almost impossible for a health checked endpoint of a dns endpoint to resolve endpoints without waiting one interval of retry, which defaults to relatively slow 3 seconds, since health checking starts within microseconds - even a fast DNS resolution will take milliseconds. It seems too tedious to require users to await on DNS endpoint groups, before then constructing a health checked one, and waiting on it. 

I'm not so confident that sticking a `join` into a constructor is a good idea though. Haven't come up with any better ideas though.

While writing this, I also realized that `OrElseEndpointGroup` (and the newly introduced `CompositeEndpointGroup` do not support waiting for initial endpoints. Going to need to fix that too in another PR.

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/DnsEndpointGroupBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/DnsEndpointGroupBenchmark.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.core.client.endpoint;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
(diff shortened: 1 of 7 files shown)


benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/DnsEndpointGroupBenchmark.java
```diff
@@ -0,0 +1,85 @@
...
+//        for (int i = 0; i < 100; i++) {
+//            benchmark.setUp();
+//            long currentTime = System.nanoTime();
+//            benchmark.resolveLocalhost();
+//            System.out.println(
+//                    "Took " + TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - currentTime) + " ms");
+//            benchmark.tearDown();
+//        }
+//        benchmark.stopServer();
+//    }
```
minwoox: This causes the check style failure.

adriancole:
agree with the rationale here. anything that has await and is a composite should await its members first to reduce tedium etc. thanks

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
```diff
@@ -91,6 +91,10 @@ protected HealthCheckedEndpointGroup(ClientFactory clientFactory,
     protected void init() {
         checkState(scheduledCheck == null, "init() must only be called once.");
 
+        if (delegate instanceof DynamicEndpointGroup) {
+            ((DynamicEndpointGroup) delegate).initialEndpointsFuture().join();
```
trustin: Should we pull this method up to `EndpointGroup`?

anuraaga: Yeah I was thinking of it - it would require quite a bit of refactoring so thinking of doing it in another PR. What do you think?

anuraaga: Went ahead and moved it in this one.

PR #1942, opened 2019-07-30
Don't add a second translated header when it already exists.

anuraaga:
During https://github.com/openzipkin/zipkin-aws/pull/135 we found that Armeria will add a second `host` header even if one is already set in the request headers before sending them. This can break behavior and I suppose must break some HTTP spec too.

I wasn't sure whether to respect or ignore the user's setting, but with this change, if a user themselves set an authority and host to different values, we'll leave the host as is. While this is a bad user, I guess we should trust they did it on purpose.

I had to write the unit test in Jetty since during the conversion from HTTP/1 headers to Armeria headers, the duplicate hosts got normalized. 

/cc @adriancole

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -40,7 +40,6 @@ dependencies {
     // Bouncy Castle
     compile 'org.bouncycastle:bcprov-jdk15on'
 
-
     // Jetty, for testing interoperability with other servers.
     testCompile 'org.eclipse.jetty:jetty-webapp'
 }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -39,6 +39,10 @@ dependencies {
 
     // Bouncy Castle
     compile 'org.bouncycastle:bcprov-jdk15on'
+
+
+    // Jetty, for testing interoperability with other servers.
+    testCompile 'org.eclipse.jetty:jetty-webapp'
 }
 
 sourceSets {
--- a/core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
@@ -868,7 +868,7 @@ public final class ArmeriaHttpUtil {
                 final AsciiString name = entry.getKey();
                 final String value = entry.getValue();
                 final AsciiString translatedName = translations.get(name);
-                if (translatedName != null) {
+                if (translatedName != null && !inputHeaders.contains(translatedName)) {
                     outputHeaders.add(translatedName, value);
                     continue;
                 }
--- a/core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
@@ -214,7 +214,9 @@ public final class ArmeriaHttpUtil {
     }
 
     /**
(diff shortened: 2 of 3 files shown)


core/build.gradle
```diff
@@ -39,6 +39,10 @@ dependencies {
 
     // Bouncy Castle
     compile 'org.bouncycastle:bcprov-jdk15on'
+
+
```
minwoox: nit: Could remove this extra line?

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -729,6 +742,47 @@ void testUpgradeRequestExecutesLogicOnlyOnce() throws Exception {
...
+        org.eclipse.jetty.server.Server jetty;
+
+        @BeforeAll
+        void startJetty() throws Exception {
+            jetty = new org.eclipse.jetty.server.Server(0);
+            jetty.setHandler(new AbstractHandler() {
+                @Override
+                public void handle(String target, Request baseRequest, HttpServletRequest request,
+                                   HttpServletResponse response) throws IOException, ServletException {
+                    if (Collections.list(request.getHeaders("host")).size() > 1) {
```
minwoox: nit: don't we need to check if the size of the host header is just one?
```
if (Collections.list(request.getHeaders("host")).size() == 1) {
    response.setStatus(HttpStatus.OK.code());
} else {
    response.sendError(HttpStatus.INTERNAL_SERVER_ERROR.code());
}
```

adriancole:
I don't know where to document this, but it is surprising that in a request callback there is no host. This would be surprising to anyone using clients that instantiate requests with a fully resolved url.

It may be worthwhile adding to Requests.headers() or authority() docs about the lazy nature of the addition of the host header.. that the user should not set it directly, but if they need to access the host data, they can via `ctx.endpoint().host()`

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -868,7 +868,7 @@ public static void toNettyHttp1(
                 final AsciiString name = entry.getKey();
                 final String value = entry.getValue();
                 final AsciiString translatedName = translations.get(name);
-                if (translatedName != null) {
+                if (translatedName != null && !inputHeaders.contains(translatedName)) {
```
adriancole: is there any chance a header that validly has multiple values could end up here? maybe a comment saying that no input header is valid with multiple values.

anuraaga: Added a comment on the translated headers that they should only contain headers that are present once in a request. Currently, the only translation is `:authority` -> `host` and maybe could just be hard-coded

core/build.gradle
trustin: @anuraaga ^^

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
trustin: @anuraaga ^^

core/build.gradle
anuraaga: Thanks for the reminder - too many PRs ><

Issue #1143, opened 2018-04-11
Difficult to request from an absolute URL with HttpClient


anuraaga:
Currently, `HttpClient`'s various `get`, `execute`, etc methods can only be used with a URL path. There are cases that a user has an absolute URL (such as from a `Location` header) and I think this can only be used by parsing the URL into authority + path and creating a client for the authority. It would also be nice to allow something like `HttpClient.DEFAULT.get("https://github.com/foo/bar/gaz")` as a drop-in replacement for simple usage of Apache or JDK.

trustin:
Agreed. A user doesn't really need to specify authority or scheme when creating an `HttpClient`. May require a fair bit of refactoring.

adriancole:
I would love something like this, but also settle for a http client function of endpoint

ikhoon:
I’d like to handle this issue. 😉

PR #1961, opened 2019-08-05
Provide a way to request from an absolute URL with HttpClient

ikhoon:
Motivation:
Currently, HttpClient executes can only be used by a URI path.
If HttpClient accepts absolute URI and executes with default options, it is easy for the first user to use it.

Modifications:
* Add static method `of()` in `HttpClient` to execute request with
  absolute URI
* Add package private `UNDEFINED_URI` in `HttpClientBuilder` to create
 `HttpClient#of()` without specifying URI

Result:
A user who wants to use HttpClient with default options can reduce boilerplate code.

Fixes: #1143

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
@@ -18,7 +18,7 @@ package com.linecorp.armeria.client;
 
 import static com.linecorp.armeria.client.HttpClientBuilder.isUndefinedUri;
 import static com.linecorp.armeria.internal.ArmeriaHttpUtil.concatPaths;
-import static com.linecorp.armeria.internal.ArmeriaHttpUtil.isAbsolutePath;
+import static com.linecorp.armeria.internal.ArmeriaHttpUtil.isAbsoluteUri;
 
 import java.net.URI;
 
@@ -54,7 +54,7 @@ final class DefaultHttpClient extends UserClient<HttpRequest, HttpResponse> impl
     private HttpResponse execute(@Nullable EventLoop eventLoop, HttpRequest req) {
         URI uri;
 
-        if (isAbsolutePath(req.path())) {
+        if (isAbsoluteUri(req.path())) {
             try {
                 uri = URI.create(req.path());
             } catch (Exception ex) {
@@ -65,28 +65,28 @@ final class DefaultHttpClient extends UserClient<HttpRequest, HttpResponse> impl
             uri = null;
         }
 
-        if (uri == null) {
-            if (isUndefinedUri(uri())) {
-                req.abort();
-                return HttpResponse.ofFailure(new IllegalArgumentException("no authority: " + req.path()));
-            } else {
-                final String originalPath = req.path();
-                final String newPath = concatPaths(uri().getRawPath(), originalPath);
-                final HttpRequest newReq;
-                // newPath and originalPath should be the same reference if uri().getRawPath() can be ignorable
-                if (newPath != originalPath) {
-                    newReq = HttpRequest.of(req, req.headers().toBuilder().path(newPath).build());
-                } else {
-                    newReq = req;
-                }
-                return execute(eventLoop, endpoint(), newReq);
-            }
-        } else {
+        if (uri != null) {
             final Endpoint endpoint = Endpoint.parse(uri.getAuthority());
             final RequestHeaders newHeaders = req.headers().toBuilder().path(uri.getRawPath()).build();
             final HttpRequest newReq = HttpRequest.of(req, newHeaders);
             return execute(eventLoop, endpoint, newReq);
         }
+
+        if (isUndefinedUri(uri())) {
+            req.abort();
+            return HttpResponse.ofFailure(new IllegalArgumentException("no authority: " + req.path()));
+        }
+
+        final String originalPath = req.path();
+        final String newPath = concatPaths(uri().getRawPath(), originalPath);
+        final HttpRequest newReq;
+        // newPath and originalPath should be the same reference if uri().getRawPath() can be ignorable
+        if (newPath != originalPath) {
+            newReq = HttpRequest.of(req, req.headers().toBuilder().path(newPath).build());
+        } else {
+            newReq = req;
+        }
+        return execute(eventLoop, endpoint(), newReq);
     }
 
     private HttpResponse execute(@Nullable EventLoop eventLoop, Endpoint endpoint, HttpRequest req) {
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
@@ -16,13 +16,17 @@
 
 package com.linecorp.armeria.client;
 
+import static com.linecorp.armeria.client.HttpClientBuilder.UNDEFINED_URI;
 import static com.linecorp.armeria.internal.ArmeriaHttpUtil.concatPaths;
+
+import java.net.URI;
 
 import javax.annotation.Nullable;
 
 import com.linecorp.armeria.common.AggregatedHttpRequest;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.RequestHeaders;
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.internal.PathAndQuery;
 
@@ -42,22 +46,38 @@ final class DefaultHttpClient extends UserClient<HttpRequest, HttpResponse> impl
     }
 
     private HttpResponse execute(@Nullable EventLoop eventLoop, HttpRequest req) {
-        final String originalPath = req.path();
-        final String newPath = concatPaths(uri().getRawPath(), originalPath);
-        final PathAndQuery pathAndQuery = PathAndQuery.parse(newPath);
+        final URI uri = URI.create(req.path());
+        final String authority = uri.getAuthority();
+
+        if (uri() == UNDEFINED_URI && authority == null) {
+            return HttpResponse.ofFailure(new IllegalArgumentException("no authority: " + req.path()));
+        } else if (authority != null) {
+            final Endpoint endpoint = Endpoint.parse(authority);
+            final RequestHeaders newHeaders = req.headers().toBuilder().path(uri.getRawPath()).build();
+            final HttpRequest newReq = HttpRequest.of(req, newHeaders);
+            return execute(eventLoop, endpoint, newReq);
+        } else {
+            final String originalPath = req.path();
+            final String newPath = concatPaths(uri().getRawPath(), originalPath);
+            final HttpRequest newReq;
+            // newPath and originalPath should be the same reference if uri().getRawPath() can be ignorable
+            if (newPath != originalPath) {
+                newReq = HttpRequest.of(req, req.headers().toBuilder().path(newPath).build());
+            } else {
+                newReq = req;
+            }
+            return execute(eventLoop, endpoint(), newReq);
+        }
+    }
+
+    private HttpResponse execute(@Nullable EventLoop eventLoop, Endpoint endpoint, HttpRequest req) {
+        final PathAndQuery pathAndQuery = PathAndQuery.parse(req.path());
         if (pathAndQuery == null) {
             req.abort();
-            return HttpResponse.ofFailure(new IllegalArgumentException("invalid path: " + newPath));
+            return HttpResponse.ofFailure(new IllegalArgumentException("invalid path: " + req.path()));
         }
-
-        final HttpRequest newReq;
-        if (newPath != originalPath) {
-            newReq = HttpRequest.of(req, req.headers().toBuilder().path(newPath).build());
-        } else {
-            newReq = req;
-        }
-
-        return execute(eventLoop, newReq.method(), pathAndQuery.path(), pathAndQuery.query(), null, newReq,
+        return execute(eventLoop, endpoint, req.method(),
+                       pathAndQuery.path(), pathAndQuery.query(), null, req,
                        (ctx, cause) -> HttpResponse.ofFailure(cause));
     }
 
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
@@ -16,7 +16,7 @@
 
 package com.linecorp.armeria.client;
 
-import static com.linecorp.armeria.client.HttpClientBuilder.UNDEFINED_URI;
+import static com.linecorp.armeria.client.HttpClientBuilder.isUndefinedUri;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
```diff
@@ -42,22 +45,38 @@ public HttpResponse execute(HttpRequest req) {
...
-        final HttpRequest newReq;
-        if (newPath != originalPath) {
-            newReq = HttpRequest.of(req, req.headers().toBuilder().path(newPath).build());
+        if (uri() == UNDEFINED_URI) {
+            final URI uri = URI.create(req.path());
+            final String authority = uri.getAuthority();
+            if (authority == null) {
+                return HttpResponse.ofFailure(new IllegalArgumentException("no authority: " + req.path()));
+            }
+            final Endpoint endpoint = Endpoint.parse(authority);
```
ikhoon: Do we need to cache `Endpoint`?

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -32,6 +34,13 @@
  */
 public interface HttpClient extends ClientBuilderParams {
 
+    /**
+     * Creates a new HTTP client using the {@link ClientFactory#DEFAULT} and the {@link ClientOptions#DEFAULT}.
+     */
+    static HttpClient of() {
```
ikhoon: `of()` or `ofDefault()`?

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
```diff
@@ -42,22 +45,38 @@ public HttpResponse execute(HttpRequest req) {
...
-        final PathAndQuery pathAndQuery = PathAndQuery.parse(newPath);
-        if (pathAndQuery == null) {
-            req.abort();
-            return HttpResponse.ofFailure(new IllegalArgumentException("invalid path: " + newPath));
-        }
-
-        final HttpRequest newReq;
-        if (newPath != originalPath) {
-            newReq = HttpRequest.of(req, req.headers().toBuilder().path(newPath).build());
+        if (uri() == UNDEFINED_URI) {
```
anuraaga: I was thinking instead of making some clients absolute and some not, we could examine the request instead, if it is an absolute url use it otherwise treat it as always. It has the big advantage of making it easier to handle Location header in redirects which I've found to be hard right now. Having to configure two clients in the same way, one to handle relative the other absolute is not so intuitive.

Would it be possible to do that?

ikhoon: @anuraaga Thanks for fast feedback.  I understood why it matters in retrying client. I will remove this condition and replace it with checking request header path. 
I assume if request header path is absolute URL then it would be higher priority than client level URI. 
Check it by adding test case for redirect. 😀

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
anuraaga: I think it's ok without caching, absolute URLs still probably aren't such a common request, and it's just one small object.

Technically, `before` is creating a whole new `HttpClient` to handle absolute URLs, so it's actually much less GC than that :P

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -35,6 +35,11 @@
  */
 public final class HttpClientBuilder extends AbstractClientOptionsBuilder<HttpClientBuilder> {
 
+    /**
+     * An undefined URI to create {@link HttpClient#of()} without specifying URI.
+     */
+    static final URI UNDEFINED_URI = URI.create("none+http://127.0.0.1");
```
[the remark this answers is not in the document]
ikhoon: Ok, that is a good idea. :-)

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
```diff
@@ -42,22 +46,38 @@ public HttpResponse execute(HttpRequest req) {
...
 
     private HttpResponse execute(@Nullable EventLoop eventLoop, HttpRequest req) {
-        final String originalPath = req.path();
-        final String newPath = concatPaths(uri().getRawPath(), originalPath);
-        final PathAndQuery pathAndQuery = PathAndQuery.parse(newPath);
-        if (pathAndQuery == null) {
-            req.abort();
-            return HttpResponse.ofFailure(new IllegalArgumentException("invalid path: " + newPath));
-        }
+        final URI uri = URI.create(req.path());
```
trustin: - Could catch an exception and assume `authority == null`.
- Could call `URI.create()` only when `req.path()` contains `://`.

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
ikhoon: I agreed with `Endpoint` is a small object. I kept the current logic. :-)

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
```diff
@@ -42,22 +46,38 @@ public HttpResponse execute(HttpRequest req) {
...
+        final String authority = uri.getAuthority();
 
-        final HttpRequest newReq;
-        if (newPath != originalPath) {
-            newReq = HttpRequest.of(req, req.headers().toBuilder().path(newPath).build());
+        if (isUndefinedUri(uri()) && authority == null) {
+            return HttpResponse.ofFailure(new IllegalArgumentException("no authority: " + req.path()));
```
minwoox: `req.abort()` before this?

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -35,6 +35,25 @@
  */
 public final class HttpClientBuilder extends AbstractClientOptionsBuilder<HttpClientBuilder> {
 
+    /**
+     * An undefined {@link URI} to create {@link HttpClient} without specifying {@link URI}.
+     */
+    private static final URI UNDEFINED_URI = URI.create("none+http://127.0.0.1");
```
trustin: How about using something more recognizable, like `none+http://undefined`?

core/src/test/java/com/linecorp/armeria/client/SimpleHttpClientTest.java
```diff
@@ -0,0 +1,110 @@
...
+    @Test
+    void custom_withRelativePath() {
+        final HttpClient client = HttpClient.of(server2.uri("/"));
+        final HttpRequest request = HttpRequest.of(HttpMethod.GET, "/simple-client");
+        final HttpResponse response = client.execute(request);
+        assertThat(response.aggregate().join().status()).isEqualTo(OK);
+    }
+
+    @Test
+    void redirect() {
```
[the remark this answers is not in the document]
trustin: YEEES :fireworks:

core/src/test/java/com/linecorp/armeria/client/SimpleHttpClientTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.ResponseHeaders;
+import com.linecorp.armeria.server.ServerBuilder;
+import com.linecorp.armeria.testing.junit.server.ServerExtension;
+
+class SimpleHttpClientTest {
```
trustin: Do we have a better name for this class? `HttpClientRequestPathTest`?

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -35,6 +35,25 @@
...
+
+    /**
+     * Returns an undefined {@link URI}.
+     */
+    static URI undefinedUri() {
+        return UNDEFINED_URI;
```
trustin: Probably OK to make `UNDEFINED_URI` package-local and accessing it directly, because it's an internal use only? I'm fine with as-is though.

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
ikhoon: Ah, thanks.

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -62,9 +81,12 @@ public HttpClientBuilder(String uri) {
      *                                  in {@link SessionProtocol}
      */
     public HttpClientBuilder(URI uri) {
```
minwoox: Because we have `HttpClient.of()` which does not take the `uri` as an argument, don't we need to add the same constructor `public HttpClientBuilder() {...}`?

core/src/test/java/com/linecorp/armeria/client/SimpleHttpClientTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @RegisterExtension
+    @Order(10)
+    static final ServerExtension server1 = new ServerExtension() {
+        @Override
+        protected void configure(ServerBuilder sb) throws Exception {
+            sb.http(0)
```
minwoox: I think we don't need this. 😄

core/src/test/java/com/linecorp/armeria/client/SimpleHttpClientTest.java
```diff
@@ -0,0 +1,110 @@
...
+              .service("/new-location", (ctx, req) -> HttpResponse.of(OK));
+        }
+    };
+
+    @RegisterExtension
+    @Order(20)
+    static final ServerExtension server2 = new ServerExtension() {
+        @Override
+        protected void configure(ServerBuilder sb) throws Exception {
+            sb.http(0)
```
minwoox: ditto

core/src/test/java/com/linecorp/armeria/client/SimpleHttpClientTest.java
ikhoon: I like the name! Thanks.

core/src/test/java/com/linecorp/armeria/client/SimpleHttpClientTest.java
ikhoon: I didn't know that, thanks!

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
ikhoon: Nice idea. :-)

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
ikhoon: fixed :-)

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
ikhoon: fixed :-)

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
```diff
@@ -46,27 +52,40 @@ public HttpResponse execute(HttpRequest req) {
...
-            if (newPath != originalPath) {
-                newReq = HttpRequest.of(req, req.headers().toBuilder().path(newPath).build());
+            uri = null;
+        }
+
+        if (uri == null) {
+            if (isUndefinedUri(uri())) {
+                req.abort();
+                return HttpResponse.ofFailure(new IllegalArgumentException("no authority: " + req.path()));
             } else {
```
trustin: nit: You could unwrap this `else` block because you always `return` in `if`.

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
```diff
@@ -46,27 +52,40 @@ public HttpResponse execute(HttpRequest req) {
...
+                    newReq = req;
+                }
+                return execute(eventLoop, endpoint(), newReq);
             }
-            return execute(eventLoop, endpoint(), newReq);
+        } else {
+            final Endpoint endpoint = Endpoint.parse(uri.getAuthority());
+            final RequestHeaders newHeaders = req.headers().toBuilder().path(uri.getRawPath()).build();
+            final HttpRequest newReq = HttpRequest.of(req, newHeaders);
+            return execute(eventLoop, endpoint, newReq);
```
trustin: How about inverting the `if` block and unwrapping the `else` block? i.e.

```java
if (uri != null) {
    ...
    return execute(eventLoop, endpoint, newReq);
}

if (isUndefinedUri(uri())) {
    ...
    return ...;
}

final String originalPath = req.path();
...
return execute(...);
```

This way, we get less indentation and cognitive load.

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
```diff
@@ -334,6 +334,25 @@ public static String decodePath(String path) {
         return new String(buf, 0, dstLen, StandardCharsets.UTF_8);
     }
 
+    /**
+     * Returns {@code true} if the specified {@code path} is an absolute {@code URI}.
+     */
+    public static boolean isAbsolutePath(@Nullable String path) {
```
trustin: `isAbsoluteUri()`?

core/src/test/java/com/linecorp/armeria/internal/ArmeriaHttpUtilTest.java
```diff
@@ -473,6 +475,22 @@ public void toArmeriaRequestHeaders() {
         assertThat(headers.authority()).isEqualTo("foo:36462");
     }
 
+    @Test
+    void isAbsolutePath() {
```
trustin: `isAbsoluteUri()`

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
ikhoon: Added [`isAbsolutePath()`](https://github.com/line/armeria/pull/1961/files#diff-6d9fb3249b1a725b4c80301a1f7d4373R340) method to check whether path is an absolute containing `://`.

core/src/main/java/com/linecorp/armeria/internal/ArmeriaHttpUtil.java
ikhoon: I was considering that name too. :-)
Does it need to change the parameter name or it is fine?

trustin: `pathOrUri` or `maybeUri`?

trustin:
@anuraaga Any other comments?

PR #1962, opened 2019-08-06
Fix not to raise ClosedPublisherException when the HttpResponse is ab…

minwoox:
…orted by client

Motivation:
`HttpResponse` can be aborted by the client. e.g, `RetryingClient` can just look at the header to
determine to retry or not and discard the rest of the response.
In that situation, because the response decoder writes to the aborted `HttpResponse`, it raises `ClosedPublisherException`.

Modification:
- Use `HttpResponse.tryWrite()` instead of `HttpResponse.write()` in `Http1ResponseDecoder`.

Result:
- `ClosedPublihserException` is no longer raised when writing to aborted `HttpResponse`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Http1ResponseDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Http1ResponseDecoder.java
@@ -161,8 +161,13 @@ final class Http1ResponseDecoder extends HttpResponseDecoder implements ChannelI
                             state = State.NEED_DATA_OR_TRAILERS;
                         }
 
-                        res.scheduleTimeout(channel().eventLoop());
-                        res.write(ArmeriaHttpUtil.toArmeria(nettyRes));
+                        // If this tryWrite() returns false, it means the response stream has been closed due to
+                        // disconnection or by the response consumer. We do not need to handle such cases here
+                        // because it will be notified to the response consumer anyway.
+                        if (!res.tryWrite(ArmeriaHttpUtil.toArmeria(nettyRes))) {
+                            // Schedule only when the response stream is still open.
+                            res.scheduleTimeout(ctx.channel().eventLoop());
+                        }
                     } else {
                         failWithUnexpectedMessageType(ctx, msg);
                     }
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/Http1ResponseDecoder.java
```diff
@@ -161,8 +161,13 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
                         }
 
-                        res.scheduleTimeout(channel().eventLoop());
-                        res.write(ArmeriaHttpUtil.toArmeria(nettyRes));
+                        // If this tryWrite() returns false, it means the response stream has been closed due to
+                        // disconnection or by the response consumer. We do not need to handle such cases here
+                        // because it will be notified to the response consumer anyway.
+                        if (!res.tryWrite(ArmeriaHttpUtil.toArmeria(nettyRes))) {
+                            // Schedule only when the response stream is still open.
+                            res.scheduleTimeout(ctx.channel().eventLoop());
```
trustin: Wouldn't this make timeout schedule a little bit late, because `write()` may invoke the subscriber directly?

minwoox: Yes, I can fix it. Thank!
I brought this logic from `Http2ResponseDecoder` and I thought it's fine unless the subscriber blocks the `EventLoop`.
But in `scheduleTimeout` method, it checks `isOpen()`, so we can call `scheduleTimeout` first. :-)

trustin: Thanks! Approved.

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -215,6 +215,13 @@ public boolean isOpen() {
             return delegate.isOpen();
         }
 
+        /**
+         * Writes the specified {@link HttpObject} to {@link DecodedHttpResponse}. This method is only called
+         * from {@link Http1ResponseDecoder} and {@link Http2ResponseDecoder}. If this returns {@code false},
+         * it means the response stream has been closed due to disconnection or by the response consumer.
+         * So the caller do not need to handle such cases because it will be notified to the response
+         * consumer anyway.
```
ikhoon: 👍

trustin:
Thanks, @minwoox !

minwoox:
Thanks for reviewing! :wink:

PR #1970, opened 2019-08-07
Remove the self-type parameters from abstract builders

trustin:
Motivation:

Self type parameters are useful when you write an abstract builder
with method chaining, but the Javadoc generated from it looks less
intuitive.

Modifications:

- Remove the usage of the self-type parameters from abstract builders.
- Miscellaneous:
  - Add `CircuitBreakerClientBuilder.mapping()`.
  - Deprecate `CircuitBreakerClientBuilder.circuitBreakerMapping()`.

Result:

- Looks prettier to users
- Increased maintanence burden for devs
- Deprecation of `CircuitBreakerClientBuilder.circuitBreakerMapping()` in favor of `mapping()`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClientBuilder.java
@@ -30,15 +30,13 @@ import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Response;
 
 /**
- * Builds a new {@link CircuitBreakerClient} or its decorator function.
+ * A skeletal builder implementation that builds a new {@link CircuitBreakerClient} or its decorator function.
  *
- * @param <T> the type of {@link CircuitBreakerClientBuilder}
- * @param <U> the type of the {@link Client} that this builder builds or decorates
+ * @param <T> the type of the {@link Client} that this builder builds or decorates
  * @param <I> the type of outgoing {@link Request} of the {@link Client}
  * @param <O> the type of incoming {@link Response} of the {@link Client}
  */
-public abstract class CircuitBreakerClientBuilder<
-        T extends CircuitBreakerClientBuilder<T, U, I, O>, U extends CircuitBreakerClient<I, O>,
+public abstract class CircuitBreakerClientBuilder<T extends CircuitBreakerClient<I, O>,
         I extends Request, O extends Response> {
 
     @Nullable
@@ -69,11 +67,6 @@ public abstract class CircuitBreakerClientBuilder<
         this.strategyWithContent = strategyWithContent;
     }
 
-    @SuppressWarnings("unchecked")
-    final T self() {
-        return (T) this;
-    }
-
     CircuitBreakerStrategy strategy() {
         checkState(strategy != null, "strategy is not set.");
         return strategy;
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerClientBuilder.java
```diff
@@ -88,27 +81,40 @@ CircuitBreakerStrategy strategy() {
...
+    }
+
+    /**
+     * Sets the {@link CircuitBreakerMapping}. If unspecified, {@link CircuitBreakerMapping#ofDefault()}
+     * will be used.
+     *
+     * @return {@code this} to support method chaining.
      */
-    public T circuitBreakerMapping(CircuitBreakerMapping mapping) {
+    public CircuitBreakerClientBuilder<T, I, O> mapping(CircuitBreakerMapping mapping) {
```
minwoox: 👍

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerRpcClientBuilder.java
```diff
@@ -36,12 +36,24 @@ public CircuitBreakerRpcClientBuilder(CircuitBreakerStrategyWithContent<RpcRespo
...
-    Function<Client<RpcRequest, RpcResponse>, CircuitBreakerRpcClient> newDecorator() {
+    public Function<Client<RpcRequest, RpcResponse>, CircuitBreakerRpcClient> newDecorator() {
         return this::build;
     }
+
+    // Methods that were overridden to change the return type.
+
+    @Override
+    public CircuitBreakerRpcClientBuilder circuitBreakerMapping(CircuitBreakerMapping mapping) {
+        return (CircuitBreakerRpcClientBuilder) super.circuitBreakerMapping(mapping);
```
minwoox: `super.circuitBreakerMapping(mapping);` -> `super.mapping(mapping);`?

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerHttpClientBuilder.java
```diff
@@ -50,14 +50,26 @@ public CircuitBreakerHttpClientBuilder(
...
     @Override
     public Function<Client<HttpRequest, HttpResponse>, CircuitBreakerHttpClient> newDecorator() {
         return this::build;
     }
+
+    // Methods that were overridden to change the return type.
+
+    @Override
+    public CircuitBreakerHttpClientBuilder circuitBreakerMapping(CircuitBreakerMapping mapping) {
+        return (CircuitBreakerHttpClientBuilder) super.circuitBreakerMapping(mapping);
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerRpcClientBuilder.java
trustin: Just replaced with `mapping(mapping)` which is simpler.

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/CircuitBreakerHttpClientBuilder.java
trustin: Just replaced with `mapping(mapping)` which is simpler.

trustin:
Thanks for reviewing.

PR #1971, opened 2019-08-07
Allows tracing of threads without a request context available using a…

### LINKED ISSUE #1964: Consider a blacklist approach for RequestContextCurrentTraceContext
An armeria app uses a tracing object that coexists with activity that is not request originated. Threads like these will result in the following message to be logged: 

```
                logger.warn("Attempted to propagate trace context, but no request context available. " +
                            "Did you forget to use RequestContext.contextAwareExecutor() or " +
                            "RequestContext.makeContextAware()?", new NoRequestContextException());
```

The advice is totally invalid in these cases as for example a scheduled thread or reporting loop has no relevance (afaik).

One way out is to allow configuration of a pattern of threads that should be ignored as opposed to spawning this message. This will allow options besides wild hacks or worse killing the log message.

Ex this is how sleuth handles one of the places that are relatively hard to shut-up :)
https://github.com/spring-cloud/spring-cloud-sleuth/blob/558900155adb1ae4a732cf450c216e2ff04a3f90/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rxjava/SleuthRxJavaSchedulersHook.java#L126

Another option is to introduce some sort of thread local which you can set to say "ignore". For example, I could set that in a thread factory in order to quiet the root thing I'm trying to do which is in some ways better than scary comments which may end up lying.

thoughts?

anuraaga:
… normal thread-local, and allows opting a thread out of the request context warning.

Currently threads that are run without a `RequestContext` cannot have traces since we return no-op scopes. It doesn't hurt to go ahead and use thread local scopes for these since they may be background threads. In general, we will still log a warning in this case, but also provide a knob to let threads opt-out of this warning if they are purely background threads.

Fixes #1964 

/cc @adriancole

### DIFF
--- a/brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
+++ b/brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
@@ -66,7 +66,7 @@ public final class RequestContextCurrentTraceContext extends CurrentTraceContext
     // Thread-local for storing TraceContext when invoking callbacks off the request thread.
     private static final ThreadLocal<TraceContext> THREAD_LOCAL_CONTEXT = new ThreadLocal<>();
 
-    private static final ThreadLocal<Boolean> THREAD_NOT_REQUEST_THREAD = ThreadLocal.withInitial(() -> false);
+    private static final ThreadLocal<Boolean> THREAD_NOT_REQUEST_THREAD = new ThreadLocal<>();
 
     private static final Scope INITIAL_REQUEST_SCOPE = new Scope() {
         @Override
@@ -250,7 +250,7 @@ public final class RequestContextCurrentTraceContext extends CurrentTraceContext
         @Override
         @Nullable
         public RequestContext get() {
-            if (THREAD_NOT_REQUEST_THREAD.get()) {
+            if (Boolean.TRUE.equals(THREAD_NOT_REQUEST_THREAD.get())) {
                 return null;
             }
             ClassLoaderHack.loadMe();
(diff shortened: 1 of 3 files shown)


brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
```diff
@@ -49,23 +49,24 @@
...
 public final class RequestContextCurrentTraceContext extends CurrentTraceContext {
 
+    /**
+     * Sets whether the current thread is not a request thread, meaning it is never executed in the scope of a
+     * server or client request and will never have a {@link RequestContext} available. This can be called from
+     * background threads, such as the thread that reports traced spans to storage, to prevent logging a
+     * warning when trying to start a trace without having a {@link RequestContext}.
+     */
+    public static void setCurrentThreadNotRequestThread(boolean value) {
+        THREAD_NOT_REQUEST_THREAD.set(value);
```
trustin: How about setting `TRUE` when `value` is `true` and removing when `false`?

anuraaga: Ah yeah good point.

adriancole:
any gotchas when a non request thread forks many client ones?

brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
```diff
@@ -142,8 +147,9 @@ private RequestContextCurrentTraceContext(Builder builder) {
     public TraceContext get() {
         final RequestContext ctx = getRequestContextOrWarnOnce();
```
adriancole: this answers my question. a new client context will prioritize vs the thread local one. This is correct imho

adriancole:
heh had a late thought. could we use FastThreadLocal internally?

trustin:
Oops! Pressed the merge button too early. :sweat_smile:

anuraaga:
The thread-locals are for non-netty threads (we can assume reasonably that netty threads will have the context since they process requests), so I think normal ThreadLocal is a bit better

adriancole:
ok I thought just from branding that eventhough it is a netty type it was
faster than stock (eg not dependent on any other behaviors). since we
always have this in classpath was thinking we could use it. but anyway
sounds fine!

PR #1974, opened 2019-08-07
Make `RetryingClient` respect the `Endpoint` selection order

### LINKED ISSUE #1973: RetryClient produces extra endpoint selection if the parent context has EndpointSelector
After version up to 0.89, I realized that, the request to upstream servers become imbalance because an `Endpoint` is already selected at `DefaultClientRequestContext#init` but `RetryingClient#newDerivedContext` also selects `Endpoint`.

Here is minimal reproducible code. You realized that, with `RetryingHttpClient`, it expects to send requests to `port:8080`, `port:8081` servers but it does not.

```java
    private static int port(Server server) {
        return server.activePort().get().localAddress().getPort();
    }

    @Test
    void endpoint() {
        List<Server> servers = new ArrayList<>();
        AtomicInteger port = new AtomicInteger(8080);
        for (int i = 0; i < 2; i++) {
            Server s = new ServerBuilder()
                    .http(port.getAndIncrement())
                    .service(Route.builder().catchAll().build(),
                             new AbstractHttpService() {
                                 @Override
                                 public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) {
                                     return HttpResponse.of(HttpStatus.OK, MediaType.ANY_TYPE,
                                                            Integer.toString(port(ctx.server())));
                                 }
                             })
                    .build();
            s.start().join();
            servers.add(s);
        }

        List<Endpoint> l = new ArrayList<>();
        for (Server server : servers) {
            l.add(Endpoint.of("127.0.0.1", port(server)).withWeight(1000));
        }

        StaticEndpointGroup g = new StaticEndpointGroup(l);

        assertThat(g.endpoints()).hasSize(2);
        EndpointGroupRegistry.register("test", g, EndpointSelectionStrategy.WEIGHTED_ROUND_ROBIN);
        HttpClient c = new HttpClientBuilder("http://group:test/")
                .decorator(new RetryingHttpClientBuilder(RetryStrategy.onServerErrorStatus()).newDecorator())
                .build();
        List<String> responses = ImmutableList.of(
                c.get("/monitor/l7check").aggregate().join().contentAscii(),
                c.get("/monitor/l7check").aggregate().join().contentAscii());
        assertThat(responses.get(0)).isEqualTo("8080");
        assertThat(responses.get(1)).isEqualTo("8081");
    }
```

trustin:
Motivation:

`RetryingClient.newDerivedContext()` always selects a new `Endpoint`,
but it should never for the first attempt, because the `Endpoint` for
the first attempt has been already selected by the parent context.

Modifications:

- Add `totalAttempt` parameter to `RetryingClient.newDerivedContext()`
  and do not select a new `Endpoint` when it's greater than 1.
- Add a test case

Result:

- Fixes #1973

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
@@ -101,10 +101,10 @@ public final class RetryingRpcClient extends RetryingClient<RpcRequest, RpcRespo
             return;
         }
 
-        final ClientRequestContext derivedCtx = newDerivedContext(ctx, req);
+        final int totalAttempts = getTotalAttempts(ctx);
+        final ClientRequestContext derivedCtx = newDerivedContext(ctx, req, totalAttempts);
         ctx.logBuilder().addChild(derivedCtx.log());
 
-        final int totalAttempts = getTotalAttempts(ctx);
         if (totalAttempts > 1) {
             derivedCtx.setAdditionalRequestHeader(ARMERIA_RETRY_COUNT, Integer.toString(totalAttempts - 1));
         }
--- a/core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientLoadBalancingTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientLoadBalancingTest.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.client.retry;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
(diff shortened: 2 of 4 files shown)


core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientLoadBalancingTest.java
```diff
@@ -0,0 +1,144 @@
...
+import com.linecorp.armeria.client.endpoint.EndpointGroupRegistry;
+import com.linecorp.armeria.client.endpoint.EndpointSelectionStrategy;
+import com.linecorp.armeria.client.endpoint.StaticEndpointGroup;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.logging.RequestLogAvailability;
+import com.linecorp.armeria.server.ServerBuilder;
+import com.linecorp.armeria.testing.junit.server.ServerExtension;
+
+public class RetryingClientLoadBalancingTest {
```
minwoox: Could remove public?

trustin: Oops. :-)

trustin:
Thanks @imasahiro for a nice test case.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingRpcClient.java
```diff
@@ -101,10 +101,10 @@ private void doExecute0(ClientRequestContext ctx, RpcRequest req,
             return;
         }
 
-        final ClientRequestContext derivedCtx = newDerivedContext(ctx, req);
+        final int totalAttempts = getTotalAttempts(ctx);
+        final ClientRequestContext derivedCtx = newDerivedContext(ctx, req, totalAttempts);
```
ikhoon: 👍

trustin:
Thanks for reviewing.

Issue #1949, opened 2019-08-05
Convert React class components to React hooks


trustin:
I find React hooks are much cleaner and less confusing than the old class-based React components. We convert our class-based components to hooks.

anuraaga:
I've had this on my mind and have migrated a few components recently so have it in my mind. Let me take care of this this week (unless someone hasn't already started).

PR #1975, opened 2019-08-07
Update docs-client dependencies, including migrating across breaking changes.

anuraaga:
In preparation for #1949 I've started by upgrading all dependencies, including some fairly major upgrades.

Major updates are material-ui - https://material-ui.com/guides/migration-v3/

I also made demo mode controllable with an environment variable, which is convenient since I can just set it as an IntelliJ run configuration without modifying code.

And finally fixed missing key warnings and fixed hot reload.


trustin:
I see some aesthetic issues though:

- Too much space between HTTP method name and RPC method name
- HTTP method name label has inconsistent widths.
  image
- Search bar font could be smaller?
  image

anuraaga:
Thanks - guess deprecations aren't only code...

I wasn't able to reproduce inconsistent width issue on Chrome Windows. Which browser did you see it? Also, maybe it won't reproduce anymore anyways since I completely recoded it using `Grid` since the behavior of `inset` seemed to change drastically.

I also restored some of the visuals of the goto-select, but am having trouble getting it to look good / same as before. The styling is pretty custom, and maybe a bad interaction with new default behavior of material-ui. @ikhoon do you think you could help take a look at what could be causing it?

image

trustin:
> Which browser did you see it?
Firefox - Just can't live without [Tree Style Tab](https://addons.mozilla.org/en-US/firefox/addon/tree-style-tab/).

ikhoon:
> do you think you could help take a look at what could be causing it?

@anuraaga Sure! I will investigate the issue soon. :->

anuraaga:
@trustin Thanks - started firefox for the first time in a while, don't see the issue with the latest code so hopefully it was fixed.

@ikhoon Thanks a lot!

trustin:
It looks much better now, but search bar still looks weird. I think @ikhoon will help us on it. :stuck_out_tongue_winking_eye: 

image

As you see, the text is not vertically aligned correctly and we could use a monospace font.

ikhoon:
@anuraaga I made another pull request based on your working branch. Sorry for the hard cording. :-(
https://github.com/anuraaga/armeria/pull/6

docs-client/package.json
```diff
@@ -9,21 +9,21 @@
     "start": "cross-env TS_NODE_PROJECT=\"tsconfig-webpack.json\" WEBPACK_DEV=true webpack-dev-server"
   },
   "dependencies": {
-    "@material-ui/core": "^3.3.2",
-    "@material-ui/icons": "^3.0.1",
+    "@material-ui/core": "^4.3.1",
```
ikhoon: nice!

docs-client/src/components/VariableList/index.tsx
```diff
@@ -170,11 +170,8 @@ class FieldInfos extends React.Component<FieldInfosProps, State> {
       <>
         {!this.state.isEmpty ? (
           this.props.variables.map((variable, index) => (
-            <>
-              <TableRow
-                key={FieldInfos.generateKey(variable.name)}
-                onClick={() => this.handleCollapse(index)}
-              >
+            <React.Fragment key={FieldInfos.generateKey(variable.name)}>
```
ikhoon: good :-)

anuraaga:
Thanks for the PR :) For the future, if you wanted to, feel free to push directly into the branch - I always tick the checkbox that maintainers can commit into my forks.

ikhoon:
messageImage_1565313680266

trustin:
Thanks, all. :-)

PR #1977, opened 2019-08-08
Implement long polling for `HealthCheckedEndpointGroup`

trustin:
Motivation:

`HealthCheckService` provides a way to get notified immediately when a
server becomes healthy or healthy. We should make use of it in
`HealthCheckedEndpointGroup`.

Modifications:

- Make `HttpHealthChecker` respect the `armeria-lphc` header.
- Miscellaneous:
  - Fix a bug where `RequestLog.responseCause()` is sometimes not
    recorded at all.

Result:

- `HealthCheckedEndpointGroup` knows really soon when a server becomes
  healthy or unhealthy, except when the server is completely
  unreachable, e.g. connection refused.
- `RequestLog.responseCause()` is recorded correctly.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
@@ -40,7 +40,6 @@ import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.RequestHeaders;
 import com.linecorp.armeria.common.RequestHeadersBuilder;
 import com.linecorp.armeria.common.SessionProtocol;
-import com.linecorp.armeria.common.logging.RequestLogAvailability;
 import com.linecorp.armeria.common.util.AsyncCloseable;
 
 import io.netty.util.AsciiString;
@@ -155,11 +154,6 @@ final class HttpHealthChecker implements AsyncCloseable {
                             TimeUnit.SECONDS.toMillis(maxLongPollingSeconds));
                     ctx.setResponseTimeoutMillis(newResponseTimeoutMillis);
                 }
-                ctx.log().addListener(log -> {
-                    if (log.responseHeaders().status().code() == 0) {
-                        System.err.println("?");
-                    }
-                }, RequestLogAvailability.COMPLETE);
             }
             return delegate().execute(ctx, req);
         }
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
@@ -18,26 +18,44 @@ package com.linecorp.armeria.client.endpoint.healthcheck;
 import java.net.StandardProtocolFamily;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
 
+import com.google.common.math.LongMath;
+
+import com.linecorp.armeria.client.Client;
 import com.linecorp.armeria.client.ClientOptionsBuilder;
+import com.linecorp.armeria.client.ClientRequestContext;
 import com.linecorp.armeria.client.Endpoint;
 import com.linecorp.armeria.client.HttpClient;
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -89,10 +132,36 @@ private synchronized void check() {
...
+                final long responseTimeoutMillis = ctx.responseTimeoutMillis();
+                if (responseTimeoutMillis > 0) {
+                    final long newResponseTimeoutMillis = LongMath.saturatedAdd(
+                            responseTimeoutMillis,
+                            TimeUnit.SECONDS.toMillis(maxLongPollingSeconds));
+                    ctx.setResponseTimeoutMillis(newResponseTimeoutMillis);
+                }
+                ctx.log().addListener(log -> {
+                    if (log.responseHeaders().status().code() == 0) {
+                        System.err.println("?");
```
minwoox: You missed this. 😄

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -70,15 +88,40 @@ void start() {
...
             }
+            ctx.updateHealth(isHealthy ? 1 : 0);
+            wasHealthy = isHealthy;
 
             try {
-                ctx.executor().schedule(this::check, ctx.nextDelayMillis(), TimeUnit.MILLISECONDS);
+                final ScheduledExecutorService executor = ctx.executor();
+                final long nextDelayMillis = ctx.nextDelayMillis();
+                if (res != null && maxLongPollingSeconds > 0) {
+                    executor.execute(this::check);
```
minwoox: I think we still need to schedule using `nextDelayMillis` to avoid aggressive health check.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
trustin: Oh, yeah :-D

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -89,10 +132,36 @@ private synchronized void check() {
 
     @Override
     public synchronized CompletableFuture<?> closeAsync() {
-        if (lastCheckFuture != null) {
-            return lastCheckFuture.handle((unused1, unused2) -> null);
+        if (lastResponse != null) {
+            lastResponse.abort();
+            return lastResponse.completionFuture().handle((unused1, unused2) -> null);
```
minwoox: I think we still have a chance that the `check` method is invoked.
If the task is scheduled by `executor.schedule(this::check, nextDelayMillis, TimeUnit.MILLISECONDS);`, and this method is called, the scheduled task will be executed.
How about introducing a flag for that?

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
trustin: That's why I only do this when `res != null`. What other conditions could we add to make this safer?

minwoox: Ah yeah, that makes sense. Thanks!
By the way, we can move the `final long nextDelayMillis = ctx.nextDelayMillis();` to the bottom `else` block, or inline it.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
trustin: Added the `closed` flag.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
trustin: Updated the logic so that it enables long polling only on 2xx and 5xx responses, so that we do not hammer a server on 4xx, etc.

minwoox: That's a good idea. 👍

Issue #1968, opened 2019-08-07
Add Fibonacci backoff implementation and update the default backoff spec


trustin:
See this tweet: https://twitter.com/sethvargo/status/1158838068805345280

We currently provide fixed, random and exponential backoff implementations and use exponential backoff as the default backoff:

- Backoff implementations: https://github.com/line/armeria/tree/master/core/src/main/java/com/linecorp/armeria/client/retry
- Default backoff spec: https://github.com/line/armeria/blob/fa3572a5698d5fcdca2dfe1295a23787bccffb7e/core/src/main/java/com/linecorp/armeria/common/Flags.java#L211-L212

We could add Fibonacci backoff implementation and update our default backoff specification.

anuraaga:
Read the papers and not finding a really relevant reason to change the default - the papers are framed in terms of mobile networks, where a fixed amount of bandwidth is allocated among many devices and a backoff that doesn't exponentially become very large allows better allocation and higher throughput. If I understand correctly, retrying in this use case is common and expected. I think our backoff is mostly to give a backend time to cool-down in times of problems, and retrying is not common or expected - so maybe exponential still makes sense as a default, though adding the fibonacci implementation seems nice.

trustin:
Thanks for the comment, @anuraaga. Yeah, at least we could add the Fibonacci implementation. Making it as a default is another story which needs more thought.

thinkgruen:
hello, I saw flag on this issues and would like to give it a shot this week. 😃

trustin:
Awesome! Feel free to let us know if you have any questions etc while working on it. :smile: By the way, are you using Armeria?

thinkgruen:
So far no real questions, PR is almost ready. I just encountered some silly arithmetic stuff along the way where MAX_LONG / 3 * 3 < MAX_LONG, but this is not a magnitude to lose sleep over, i guess. 😄

As for your question: not actively atm. I am somewhat familiar with Netty already as well as a LINE user, so I thought it would be interesting to have a look into Armeria.

PR #1979, opened 2019-08-08
implemented Fibonacci backoff

thinkgruen:
Motivation:
- Currently no Fibonacci backoff implementation is available. According to #1968 this strategy is considered as a default, though the decision is not yet final.

Modifications:
- Added Fibonacci backoff (iterative, starting with a multiplier of 1) alongside with tests. This backoff is also documented in and made available through a functional interface.

Result:
- Fibonacci backoff can be used as an alternative to random, fixed or exponential backoff.


core/src/main/java/com/linecorp/armeria/client/retry/FibonacciBackoff.java
```diff
@@ -0,0 +1,63 @@
...
+        int c;
+        for (int i = 0; i <= n - 2; i++) {
+            c = a;
+            a = b;
+            b += c;
+        }
+        return a;
+    }
+
+    private static long saturatedMultiply(long left, int right) {
```
anuraaga: Can use guava

https://static.javadoc.io/com.google.guava/guava/28.0-jre/com/google/common/math/LongMath.html#saturatedMultiply-long-long-

core/src/main/java/com/linecorp/armeria/client/retry/FibonacciBackoff.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    protected long doNextDelayMillis(int numAttemptsSoFar) {
+        final long nextDelay = saturatedMultiply(initialDelayMillis, fib(numAttemptsSoFar));
+        return Math.min(nextDelay, maxDelayMillis);
+    }
+
+    private static int fib(int n) {
+        int a = 1;
```
anuraaga: Instead of calculating a multiplier, can we do addition using nextDelay directly? It would make it easy to cap the calculation at maxDelay.

If I not mistaken, as the number of attempts goes up this calculation will take longer and longer, unbounded.

core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java
```diff
@@ -105,6 +116,7 @@ static Backoff random(long minDelayMillis, long maxDelayMillis, Supplier<Random>
      * <ul>
      *   <li>{@code exponential=200:10000:2.0,jitter=0.2} (default)</li>
      *   <li>{@code exponential=200:10000,jitter=0.2,maxAttempts=50} (multiplier omitted)</li>
+     *   <li>{@code fibonacci=200:10000,jitter=0.2,maxAttempts=50}</li>
```
minwoox: Could you add the parsing logic in `BackoffSpec` and add the test case for it in `BackoffSpecTest`, please?

core/src/test/java/com/linecorp/armeria/client/retry/FibonacciBackoffTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public void testOverflow() {
+        final Backoff backoff = new FibonacciBackoff(Long.MAX_VALUE / 3, Long.MAX_VALUE);
+        assertThat(backoff.nextDelayMillis(1)).isEqualTo(Long.MAX_VALUE / 3);
+        assertThat(backoff.nextDelayMillis(2)).isEqualTo(Long.MAX_VALUE / 3);
+        assertThat(backoff.nextDelayMillis(3)).isEqualTo(Long.MAX_VALUE / 3 * 2);
+        assertThat(backoff.nextDelayMillis(4)).isEqualTo(Long.MAX_VALUE / 3 * 3);
+        assertThat(backoff.nextDelayMillis(5)).isEqualTo(Long.MAX_VALUE);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
```
ikhoon: Thanks for your contribution! 😀
We are writing test code using [junit5](https://junit.org/junit5/docs/current/user-guide/) for the new test.
Could you migrate this code into junit5?
* Replace `org.junit.Test` -> `org.junit.jupiter.api.Test`.
* Remove `public` modifier.
* Use `assertThatThrownBy()` method to assert an exception.
```java
assertThatThrownBy(() -> new FibonacciBackoff(-5, 120))
    .isInstanceOf(IllegalArgumentException.class);
```

core/src/main/java/com/linecorp/armeria/client/retry/FibonacciBackoff.java
thinkgruen: Thanks for the review! Other solutions I thought of is by keeping the last two multipliers in fields. As this is in the client package, I assume that the backoff instance is not shared, but I would have to read more of the code to make sure.

Or another solution would be creating a look-table upon initialization, it only has to have a size of 90-something for long.

core/src/main/java/com/linecorp/armeria/client/retry/FibonacciBackoff.java
thinkgruen: Thanks for the review! This method was also suggested to me, but I think was flagged as unstable, would have to check later again why that was. I also thought that the underflow checks it does weren't required in this context, so I made it only for positive values.

core/src/test/java/com/linecorp/armeria/client/retry/FibonacciBackoffTest.java
thinkgruen: Completely missed this, I just used the default in my IDE without checking. Thanks!

core/src/main/java/com/linecorp/armeria/client/retry/FibonacciBackoff.java
anuraaga: A lookup table sounds nice, since in practice, it'll probably be far fewer, like 5 entries. We may as well do that for ExponentialBackoff too :D

core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java
thinkgruen: Thanks for the review, I will add it as well!

core/src/main/java/com/linecorp/armeria/client/retry/FibonacciBackoff.java
```diff
@@ -0,0 +1,91 @@
...
+
+    FibonacciBackoff(long initialDelayMillis, long maxDelayMillis) {
+        checkArgument(initialDelayMillis >= 0,
+                      "initialDelayMillis: %s (expected: >= 0)", initialDelayMillis);
+        checkArgument(initialDelayMillis <= maxDelayMillis, "maxDelayMillis: %s (expected: >= %s)",
+                      maxDelayMillis, initialDelayMillis);
+        this.initialDelayMillis = initialDelayMillis;
+        this.maxDelayMillis = maxDelayMillis;
+
+        final List<Long> precomputed = new ArrayList<>();
```
trustin: How about precomputing only when there will be many values, like @anuraaga did in his pull request? https://github.com/line/armeria/pull/1983/files#diff-2d744bfdcc8222dbb89cabef4de01117R45-R60

core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java
```diff
@@ -86,12 +94,15 @@ static Backoff random(long minDelayMillis, long maxDelayMillis, Supplier<Random>
 
     /**
      * Creates a new {@link Backoff} that computes backoff delay using one of
-     * {@link #exponential(long, long, double)}, {@link #fixed(long)} and {@link #random(long, long)} chaining
-     * with {@link #withJitter(double, double)} and {@link #withMaxAttempts(int)} from the
-     * {@code specification}. This is the format for the {@code specification}:
+     * {@link #exponential(long, long, double)}, {@link #fibonacci(long, long)}, {@link #fixed(long)}
+     * and {@link #random(long, long)} chainingwith {@link #withJitter(double, double)} and
```
minwoox: nit: chaningwith -> chaining with 😄

core/src/main/java/com/linecorp/armeria/client/retry/BackoffSpec.java
```diff
@@ -275,6 +299,8 @@ Backoff build() {
             backoff = Backoff.fixed(fixedDelayMillis);
         } else if (baseOption == BaseOption.random) {
             backoff = Backoff.random(randomMinDelayMillis, randomMaxDelayMillis);
+        } else if (baseOption == BaseOption.fibonacci) {
+            backoff = Backoff.fibonacci(randomMinDelayMillis, randomMaxDelayMillis);
```
minwoox: This should be
`backoff = Backoff.fibonacci(initialDelayMillis, maxDelayMillis);`

core/src/main/java/com/linecorp/armeria/client/retry/FibonacciBackoff.java
thinkgruen: I thought of this well. The problem I saw with the Fibonacci sequence is that you already calculate the whole sequence up until that point to find out if it becomes big. For 30, however, this can be estimated by using the same check as we do for like for the exponential one `Math.pow(1.6..., 29)`, but I think in practice I would set the bar way lower when it is worth performing the precalculation for Fibonacci.

core/src/main/java/com/linecorp/armeria/client/retry/BackoffSpec.java
thinkgruen: Well spotted, I completely missed it. 🙇🏻‍♂️

core/src/main/java/com/linecorp/armeria/client/retry/FibonacciBackoff.java
thinkgruen: Reading through the PR from @anuraaga again, I assume the purpose is the opposite. It ensures that the collection is not becoming too big. Having `maxDelay` = multiplier*initialDelay` would also perform the precalculation with a resulting Collection of two elements, I think.

In my Implementation, I kick off an iterative calculation of the 31st, 32nd... etc. delay then, capping the Collection size at 30 still. I directly make use of predeceding values (29th and 30th), while the exponential solution is not using precalculation at that point.

anuraaga: Agree there's no reason to make the precomputation conditional and I removed it from my PR. I was thinking to optimize the case where there are few possible values of delay, but realize that it doesn't hurt to optimize the case when there are lots of values too.

core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java
```diff
@@ -68,6 +68,14 @@ static Backoff exponential(long initialDelayMillis, long maxDelayMillis, double
         return new ExponentialBackoff(initialDelayMillis, maxDelayMillis, multiplier);
     }
 
+    /**
+     * Returns a {@link Backoff} for which the backoff delay increases in line with the Fibonacci sequence
+     * f(n) = f(n-1) + f(n-2).
```
minwoox: nit: how about adding `where f(1) = {@code initialDelayMillis}`

core/src/test/java/com/linecorp/armeria/client/retry/FibonacciBackoffTest.java
```diff
@@ -0,0 +1,64 @@
...
+ * under the License.
+ */
+package com.linecorp.armeria.client.retry;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.junit.jupiter.api.Test;
+
+public class FibonacciBackoffTest {
```
ikhoon: nit: Please remove this `public` too. :-)

core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java
thinkgruen: I like it, added :) Thanks!

core/src/test/java/com/linecorp/armeria/client/retry/FibonacciBackoffTest.java
thinkgruen: Ah true. Thanks! :)

trustin:
Thanks a lot, @thinkgruen! Looking forward to your next pull request. :-D

PR #1980, opened 2019-08-09
RequestContextCurrentTraceContext.setCurrentThreadNotRequestThread code

adriancole:
This shows how to use `setCurrentThreadNotRequestThread` until we have
a function of Thread instead.

### DIFF
--- a/brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
+++ b/brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
@@ -59,12 +59,14 @@ public final class RequestContextCurrentTraceContext extends CurrentTraceContext
      * a {@link java.util.concurrent.ThreadFactory} like the following:
      * <pre>{@code
      * > ThreadFactory factory = (runnable) -> new Thread(new Runnable() {
-     * >     @Override public void run() {
+     * >     @Override
+     * >     public void run() {
      * >         RequestContextCurrentTraceContext.setCurrentThreadNotRequestThread(true);
      * >         runnable.run();
      * >     }
      * >
-     * >     @Override public String toString() {
+     * >     @Override
+     * >     public String toString() {
      * >         return runnable.toString();
      * >     }
      * > });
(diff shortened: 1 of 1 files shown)


adriancole:
this is kindof worse when using other thread factories..

```java
      new NamedThreadFactory("zipkin-throttle-pool") {
        @Override public Thread newThread(Runnable runnable) {
          return super.newThread(new Runnable() {
            @Override public void run() {
              RequestContextCurrentTraceContext.setCurrentThreadNotRequestThread(true);
              runnable.run();
            }

            @Override public String toString() {
              return runnable.toString();
            }
          });
        }
      }
```

adriancole:
honestly I think it is too crufty to go this route.. we should do something different. wdyt @anuraaga?

adriancole:
PS this is contained for now, so I don't think it needs to necessarily block a release, just it doesn't neatly integrate with thread factories and that's where it is most likely to need to integrate. you can see some code here https://github.com/openzipkin/zipkin/pull/2743

brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
```diff
@@ -54,6 +54,20 @@
      * server or client request and will never have a {@link RequestContext} available. This can be called from
      * background threads, such as the thread that reports traced spans to storage, to prevent logging a
      * warning when trying to start a trace without having a {@link RequestContext}.
+     *
+     * Ex. To prevent warnings from an administrative thread controlled by a thread factory:
```
trustin: `Ex. To` -> `<p>For example, to`

brave/src/main/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContext.java
```diff
@@ -54,6 +54,20 @@
...
+     * <pre>{@code
+     * threadFactory = (runnable) -> new Thread(new Runnable() {
+     *   @Override public void run() {
```
trustin: You could use `>`:

```
 * <pre>{@code
 * > ThreadFactory threadFactory = ...
 * >     @Override
 * >     public void run() {
 * > ...
 * > });
 * }</pre>
```

trustin:
@adriancole Update the pull request a little bit to make the build pass. What do you think?

adriancole:
thanks @trustin works for me

PR #1982, opened 2019-08-09
Various improvements in health checking code

trustin:
Motivation:

- When a server stops, the clients who sent long-polling health check
  requests will currently get a GOAWAY frame or an immediate
  disconnection.
- The port number specified in `HealthCheckedEndpointGroupBuilder.port()`
  is not used when filling the `:authority` header.
- A user sometimes wants to send a `GET` request rather than `HEAD`.

Modifications:

- Modify `HealthCheckService` to send a response even when a server is
  stopping, so that the client is given a chance to receive them.
- Update `HttpHealthCheckedEndpointGroupLongPollingTest` to make sure
  the client gets the correct 'unhealthy' response with `armeria-lphc: 0`
  header when a server stops.
- Made sure the custom port numbers are respected when building URI and
  `:authority` header.
- Add `useGet(boolean)` option to the builder.

Result:

- No unexpected disconnection/GOAWAY when a server stops.
- No more incorrect `:authority` header.
- A user can choose to send `GET` health check requests instead of `HEAD`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/AbstractHealthCheckedEndpointGroupBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/AbstractHealthCheckedEndpointGroupBuilder.java
@@ -38,7 +38,6 @@ import com.linecorp.armeria.common.util.AsyncCloseable;
 public abstract class AbstractHealthCheckedEndpointGroupBuilder {
 
     private final EndpointGroup delegate;
-    private final Function<? super HealthCheckerContext, ? extends AsyncCloseable> checker;
 
     private SessionProtocol protocol = SessionProtocol.HTTP;
     private Backoff retryBackoff = DEFAULT_HEALTH_CHECK_RETRY_BACKOFF;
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/AbstractHealthCheckedEndpointGroupBuilder.java
```diff
@@ -155,6 +144,16 @@ public AbstractHealthCheckedEndpointGroupBuilder withClientOptions(
...
-                                              retryBackoff, configurator, checker);
+                                              retryBackoff, configurator, newCheckerFactory());
     }
+
+    /**
+     * Returns the {@link Function} that starts to send health check requests to the {@link Endpoint}
+     * specified in a given {@link HealthCheckerContext} when invoked. The {@link Function} must update
+     * the health of the {@link Endpoint} with a value between [0, 1] via
+     * {@link HealthCheckerContext#updateHealth(double)}. {@link HealthCheckedEndpointGroup} will call
+     * {@link AsyncCloseable#closeAsync()}on the {@link AsyncCloseable} returned by the {@link Function}
```
minwoox: nit: need a space before `on`

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -50,29 +50,55 @@
...
+                }
             }
+
+            if (port != protocol.defaultPort()) {
+                uriBuf.append(':').append(port);
+                authorityBuf.append(':').append(port);
+            }
+
+            builder = new HttpClientBuilder(uriBuf.toString());
+            this.authority = authorityBuf.toString();
```
anuraaga: Instead of saving to a field and manually adding it to headers, I think we can add it as a header to the client. Is that cleaner?

trustin: I think both are OK. I guess they don't make much difference except it will end up with making a single-header `HttpHeaders` internally, which might be tad bit more expensive.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/AbstractHealthCheckedEndpointGroupBuilder.java
trustin: Fixed :laughing:

PR #1990, opened 2019-08-12
Fix not to raise ConcurrentModificationException in HealthCheckedEndp…

### LINKED ISSUE #1989: `ConcurrentModificationException` in `HealthCheckedEndpointGroup.<init>`
```
Caused by: java.util.ConcurrentModificationException: null
	at java.base/java.util.HashMap$Values.forEach(HashMap.java:979)
	at com.linecorp.armeria.client.endpoint.healthcheck.HealthCheckedEndpointGroup.<init>(HealthCheckedEndpointGroup.java:126)
	at com.linecorp.armeria.client.endpoint.healthcheck.AbstractHealthCheckedEndpointGroupBuilder.build(AbstractHealthCheckedEndpointGroupBuilder.java:147)
```

minwoox:
…ointGroup.<init>

Fix #1989

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
@@ -42,6 +42,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.spotify.futures.CompletableFutures;
 
 import com.linecorp.armeria.client.ClientFactory;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
```diff
@@ -123,7 +123,11 @@ public static HealthCheckedEndpointGroupBuilder builder(EndpointGroup delegate,
         updateCandidates(delegate.initialEndpointsFuture().join());
 
         // Wait until the initial health of all endpoints are determined.
-        contexts.values().forEach(ctx -> ctx.initialCheckFuture.join());
```
anuraaga: Would be nice to add a test that previously failed - I think it's possible to do it deterministically by having a initial endpoints future that doesn't complete while adding endpoints.

minwoox: Ah, thanks! Let me make it in another PR.

PR #1992, opened 2019-08-12
Add a test for HealthCheckedEndpointGroup raising ConcurrentModificat…

minwoox:
…ionException previously

Related #1990

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupTest.java
@@ -22,7 +22,6 @@ import static org.awaitility.Awaitility.await;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -30,14 +29,19 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
 import com.google.common.collect.ImmutableList;
 
 import com.linecorp.armeria.client.Endpoint;
 import com.linecorp.armeria.client.endpoint.DynamicEndpointGroup;
 import com.linecorp.armeria.common.util.AsyncCloseable;
+import com.linecorp.armeria.testing.junit.common.EventLoopExtension;
 
 class HealthCheckedEndpointGroupTest {
+
+    @RegisterExtension
+    static final EventLoopExtension eventLoop = new EventLoopExtension();
 
     @Test
     void delegateUpdateCandidatesWhileCreatingHealthCheckedEndpointGroup() {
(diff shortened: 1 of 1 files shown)


core/src/test/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupTest.java
```diff
@@ -34,6 +38,43 @@
...
+    void delegateUpdateCandidatesWhileCreatingHealthCheckedEndpointGroup() {
+        final MockEndpointGroup delegate = new MockEndpointGroup();
+        final CompletableFuture<List<Endpoint>> future = delegate.initialEndpointsFuture();
+        future.complete(ImmutableList.of(Endpoint.of("127.0.0.1", 8080), Endpoint.of("127.0.0.1", 8081)));
+
+        final CountDownLatch latch = new CountDownLatch(1);
+
+        // Schedule the task which update the endpoint one second later to ensure that the change is happening
+        // while creating the HealthCheckedEndpointGroup.
+        Executors.newSingleThreadScheduledExecutor().schedule(
```
ikhoon: nit: use daemon thread?
`Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("health-check", true))`

minwoox: Fixed to use `EventLoopExtension`. 😉

anuraaga: Can also use a common event loop or forkjoinpool

core/src/test/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupTest.java
```diff
@@ -19,21 +19,66 @@
...
 class HealthCheckedEndpointGroupTest {
+
+    @RegisterExtension
+    static final EventLoopExtension eventLoop = new EventLoopExtension();
+
+    @Test
+    void delegateUpdateCandidatesWhileCreatingHealthCheckedEndpointGroup() {
+        final MockEndpointGroup delegate = new MockEndpointGroup();
+        final CompletableFuture<List<Endpoint>> future = delegate.initialEndpointsFuture();
+        future.complete(ImmutableList.of(Endpoint.of("127.0.0.1", 8080), Endpoint.of("127.0.0.1", 8081)));
```
anuraaga: Just to confirm with the old code, this test fails with the exception. If so then it's probably good. I was sort of expecting not to complete this future (so initialCheckFuture.join() blocks forever).

minwoox: Yeah, it failed before.
I have to complete the future because the thread should be at `forEach(ctx -> ctx.initialCheckFuture.join());`. So another thread updates the endpoints and triggers the `ConcurrentModificationException`. 😄

core/src/test/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupTest.java
minwoox: Changed to use `CommonPools`. (Can't use forkjoinpool because I have to schedule it. 😉 )

minwoox:
Thanks, for reviewing!

PR #1998, opened 2019-08-16
Handles gRPC status when deferred to trailers

adriancole:
It seems stackdriver trace api will wait until trailers to set the
status for certain errors. This seems legit looking at the gRPC spec.

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors

Example error:
```
2019-08-16 10:45:30.502  INFO 47885 --- [-worker-nio-2-2] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-08-16T02:45:30.498Z(1565923530498646), length=0B, duration=3786µs(3786891ns), headers=[:status=200, content-type=application/grpc, date=Fri, 16 Aug 2019 02:45:12 GMT, alt-svc=quic=":443"; ma=2592000; v="46,43,39"], trailers=[EOS, grpc-status=5, grpc-message=Requested entity was not found., google.rpc.resourceinfo-bin=EhNwcm9qZWN0cy96aXBraW4tZGV2, grpc-status-details-bin=CAUSH1JlcXVlc3RlZCBlbnRpdHkgd2FzIG5vdCBmb3VuZC4aRAordHlwZS5nb29nbGVhcGlzLmNvbS9nb29nbGUucnBjLlJlc291cmNlSW5mbxIVEhNwcm9qZWN0cy96aXBraW4tZGV2]}
```

### DIFF
--- a/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
+++ b/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
@@ -94,7 +94,7 @@ public class UnaryGrpcClient {
                                          "Non-successful HTTP response code: " + msg.status());
                              }
 
-                             // Status can either be in the headers or trailers depending on error
+                             // Status can either be in the headers or trailers depending on errorUnaryGrpcClientTest
                              String grpcStatus = msg.headers().get(GrpcHeaderNames.GRPC_STATUS);
                              if (grpcStatus != null) {
                                  checkGrpcStatus(grpcStatus, msg.headers());
--- a/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
+++ b/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
@@ -94,7 +94,7 @@ public class UnaryGrpcClient {
                                          "Non-successful HTTP response code: " + msg.status());
                              }
 
-                             // Status can either be in the headers or trailers depending on errorUnaryGrpcClientTest
+                             // Status can either be in the headers or trailers depending on error
                              String grpcStatus = msg.headers().get(GrpcHeaderNames.GRPC_STATUS);
                              if (grpcStatus != null) {
                                  checkGrpcStatus(grpcStatus, msg.headers());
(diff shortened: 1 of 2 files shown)


grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
```diff
@@ -91,21 +94,31 @@ public UnaryGrpcClient(HttpClient httpClient) {
...
-                                 String grpcMessage = msg.headers().get(GrpcHeaderNames.GRPC_MESSAGE);
-                                 if (grpcMessage != null) {
-                                     grpcMessage = StatusMessageEscaper.unescape(grpcMessage);
-                                 }
-                                 throw new ArmeriaStatusException(
-                                         Integer.parseInt(grpcStatus),
-                                         grpcMessage);
+                             // Status can either be in the headers or trailers depending on error
+                             String grpcStatus = msg.headers().get(GrpcHeaderNames.GRPC_STATUS);
+                             if (grpcStatus != null) {
```
anuraaga: Think a bit cleaner as

```java
if (grpcStatus == null) {
  grpcStatus = trailers.get(
```

adriancole: but we would still have to flip which we use to get the other header right? this also controls the condition of whether you get GRPC_MESSAGE from headers or trailers

adriancole: or are you saying having the same blocks flipped in order is the cleaner thing..

anuraaga: Ah nevermind that's right. Let's extend the test to also have a grpc-message so I don't accidentally break it ;)

adriancole: sgtm

adriancole:
sorry about the checkstyle commits. I was running `./gradlew :grpc:checkstyle` not `:grpc-protocol:checkstyleMain`

elefeint:
Thank you for figuring it out! I think this is why I've been struggling to make openzipkin/zipkin-gcp#131 work -- the service failure was not propagating.

trustin:
Thanks, @adriancole !

Issue #1969, opened 2019-08-07
totalDuration does not contain connectionAcquisition


minwoox:
Currently, `totalDuration` is:
```
request header sent <--------- total duration --------> receive response completely
```
If connecting to the endpoint takes long due to dns resolution, a user has to look at the [connectionAcquisition](http://line.github.io/armeria/advanced-structured-logging.html#client-connection-timing-properties) particularly besides the `totalDuration`.
Therefore, if we include `connectionAcquisition` in `totalDuration`, he/she might notice quickly when a problem happens while establishing a connection.

minwoox:
Superseded by #2272

PR #2002, opened 2019-08-17
Support connectionAcquisition for totalDuration in ClientRequest

JunoJunho:
Motivation:

 - totalDuration does not fully support request total duration

Modifications:

 - Accumulate totalDuration with connectionAcquisition metric
 - More refactor (?)

Result:

 - Handle #1969


trustin:
I think the approach made in this pull request is incorrect. A correct change would be to make `RequestLog.totalDuration*()` include the connection acquisition time, not just touching the exported metrics.

trustin:
Are you going to update this pull request as I suggested or ..?

JunoJunho:
Ah, sorry for late response, I am vacation on recent days.
I discussed this issue with @minwoox and @minwoox said that there is a possibility to change this metric acquisition might be changed on implementing other features.

Therefore, it is not a good timing to handle this now.

Let me close this pr.

PR #2004, opened 2019-08-20
Keep an endpoint's healthiness if long-poll health check returns NOT_…

imasahiro:
…MODIFIED status code

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
@@ -36,6 +36,7 @@ import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.RequestHeaders;
 import com.linecorp.armeria.common.RequestHeadersBuilder;
 import com.linecorp.armeria.common.util.AsyncCloseable;
@@ -108,6 +109,9 @@ final class HttpHealthChecker implements AsyncCloseable {
                         maxLongPollingSeconds = getMaxLongPollingSeconds(res);
                         break;
                     default:
+                        if (res.status() == HttpStatus.NOT_MODIFIED) {
+                            isHealthy = wasHealthy;
+                        }
                         // Do not use long polling on an unexpected status for safety.
                         maxLongPollingSeconds = 0;
                 }
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
@@ -110,10 +110,12 @@ final class HttpHealthChecker implements AsyncCloseable {
                         break;
                     default:
                         if (res.status() == HttpStatus.NOT_MODIFIED) {
+                            maxLongPollingSeconds = getMaxLongPollingSeconds(res);
                             isHealthy = wasHealthy;
+                        } else {
+                            // Do not use long polling on an unexpected status for safety.
+                            maxLongPollingSeconds = 0;
                         }
-                        // Do not use long polling on an unexpected status for safety.
-                        maxLongPollingSeconds = 0;
                 }
             } else {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -108,6 +109,9 @@ private synchronized void check() {
                         maxLongPollingSeconds = getMaxLongPollingSeconds(res);
                         break;
                     default:
+                        if (res.status() == HttpStatus.NOT_MODIFIED) {
+                            isHealthy = wasHealthy;
+                        }
                         // Do not use long polling on an unexpected status for safety.
                         maxLongPollingSeconds = 0;
```
mauhiz: we can't continue long polling on 304?

core/src/test/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupLongPollingTest.java
```diff
@@ -167,6 +171,44 @@ void periodicCheckWhenConnectionRefused() throws Exception {
...
+            assertThat(notModifiedResponseHeaders.getLong("armeria-lphc", 1)).isEqualTo(0);
+
+            // Check the next check respected backoff, because there's no point of
+            // sending a request immediately if you get a 'not modified'.
+            final Stopwatch stopwatch = Stopwatch.createStarted();
+            healthCheckRequestLogs.take();
+            assertThat(stopwatch.elapsed(TimeUnit.MILLISECONDS))
+                    .isGreaterThan(RETRY_INTERVAL.toMillis() * 4 / 5);
+        } finally {
+            this.healthCheckRequestLogs = null;
```
mauhiz: why?

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
imasahiro: Ah, we can use it. Fixed

core/src/test/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthCheckedEndpointGroupLongPollingTest.java
imasahiro: To make sure healthCheckRequestLogs is always null when the tests start?

mauhiz: I thought each test uses a new instance of the test class.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -108,8 +109,13 @@ private synchronized void check() {
                         maxLongPollingSeconds = getMaxLongPollingSeconds(res);
                         break;
                     default:
-                        // Do not use long polling on an unexpected status for safety.
-                        maxLongPollingSeconds = 0;
+                        if (res.status() == HttpStatus.NOT_MODIFIED) {
```
ikhoon: nit: Why don’t you move this into a case statement to keep consistency.

mauhiz:
Because he doesn't want to apply for whole 3XX range, I think

minwoox:
Perhaps @ikhoon wants it like:
```java
switch (res.status().codeClass()) {
    case SUCCESS:
        maxLongPollingSeconds = getMaxLongPollingSeconds(res);
        isHealthy = true;
        break;
	case REDIRECTION:
		if (res.status() == HttpStatus.NOT_MODIFIED) {
            maxLongPollingSeconds = getMaxLongPollingSeconds(res);
            isHealthy = wasHealthy;
        } else {
            // Do not use long polling on an unexpected status for safety.
            maxLongPollingSeconds = 0;
        }
    case SERVER_ERROR:
        maxLongPollingSeconds = getMaxLongPollingSeconds(res);
        break;
    default:
        // Do not use long polling on an unexpected status for safety.
        maxLongPollingSeconds = 0;
}
```
But I like the current way. :-)

imasahiro:
Yes, if we need to handle more case, I would like to add the code (https://github.com/line/armeria/pull/2004#issuecomment-523272751) but since `304` is the only special case we need to care at the moment, I would like to merge into `default` case.

trustin:
Thanks, @imasahiro !

PR #2005, opened 2019-08-21
Fix `byte[]` and `HttpData` conversion

### LINKED ISSUE #1999: ByteArrayResponseConverterFunction clamped to two MediaTypes
The following snippet should imho work:
```
    @Get("/favicon.ico")
    @Produces("image/gif")
    private ByteArray favicon() {
        return new byte[0]
    }
```

It does not work however, because if the `contentType` isn't null, and isn't one of APPLICATION_BINARY or OCTET_STREAM, the converter will not be used.

Discussion about problem:
https://app.slack.com/client/T1NH00FK9/C1NGPBUH2/thread/C1NGPBUH2-1565946569.136900

trustin:
Motivation:

The following annotated service method will fail because
`ByteArrayResponseConverterFunction` currently does not convert `byte[]`
or `HttpData` with the content type other than
`application/octet-stream` or `application/binary`.

Modifications:

- Convert `byte[]` or `HttpData` even if the content type is neither
  `application/octet-stream` nor `application/binary`.
- Also fix the case where `byte[]` or `HttpData` is not converted when a
  service method produces text content.

Result:

- Fixes #1999

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
+++ b/core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
@@ -80,6 +80,18 @@ import reactor.test.StepVerifier;
 public class AnnotatedHttpServiceResponseConverterTest {
 
     private static final ObjectMapper mapper = new ObjectMapper();
+    private static final String STRING = "₩";
+    private static final byte[] BYTEARRAY = STRING.getBytes(StandardCharsets.UTF_8);
+    private static final HttpData HTTPDATA = HttpData.wrap(BYTEARRAY);
+    private static final JsonNode JSONNODE;
+
+    static {
+        try {
+            JSONNODE = mapper.readTree("{\"a\":\"₩\"}");
+        } catch (IOException e) {
+            throw new Error(e);
+        }
+    }
 
     @ClassRule
     public static final ServerRule rule = new ServerRule() {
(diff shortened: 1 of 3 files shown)


trustin:
For @tobias-

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
```diff
@@ -563,15 +599,33 @@ public void produceTypeAnnotationBasedDefaultResponseConverter() throws Exceptio
...
+        System.err.println(res.contentUtf8());
+        assertThat(res.content().array()).isEqualTo(BYTEARRAY);
+
+        res = aggregated(client.get("/byteArrayTxt"));
+        assertThat(res.contentType()).isEqualTo(MediaType.PLAIN_TEXT_UTF_8);
+        assertThat(res.content().array()).isEqualTo(BYTEARRAY);
+
+        res = aggregated(client.get("/httpDataTxt"));
+        assertThat(res.contentType()).isEqualTo(MediaType.PLAIN_TEXT_UTF_8);
+        System.err.println(res.contentUtf8());
```
ikhoon: ditto

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
```diff
@@ -563,15 +599,33 @@ public void produceTypeAnnotationBasedDefaultResponseConverter() throws Exceptio
...
-        assertThat(res.content().array()).isEqualTo("¥".getBytes());
+        assertThat(res.content().array()).isEqualTo(BYTEARRAY);
+
+        res = aggregated(client.get("/byteArrayGif"));
+        assertThat(res.contentType()).isEqualTo(MediaType.GIF);
+        assertThat(res.content().array()).isEqualTo(BYTEARRAY);
+
+        res = aggregated(client.get("/httpDataPng"));
+        assertThat(res.contentType()).isEqualTo(MediaType.PNG);
+        System.err.println(res.contentUtf8());
```
ikhoon: nit: remove this? 😀

trustin: Oops :sweat_smile:

PR #2009, opened 2019-08-21
Update Tomcat from 9.0.22 to 9.0.24

trustin:


trustin:
Missed this one in #2007

Issue #1966, opened 2019-08-06
Add literal value field for enum in internal documentation


Waynefn:
In internal documentation for enums `internal/docs/#/enums`,
there are only `Name` and `Description` field for enum types.
It would be more convenient if the literal value (for example, int value) for this enum name is also displayed,
since in most cases, we use that value to debug request and response.

trustin:
That's a good idea. Are you interested in sending a pull request? :smile:

Ta2Rim:
I'll work on it

trustin:
@Ta2Rim Awesome!

Waynefn:
Cool. I found the improvement just recently when I using it!

PR #2014, opened 2019-08-22
Add literal value field for enum in internal documentation #1966

Ta2Rim:
**Motivation**
#1966

**Modification**
Add literal value field for enum in documentation.
 -> Add literal value as member variable to EnumValueInfo, modify following changes.

**Result**
User can see literal value of enum values in doc pages

There's a question I have working on this issue.
I wonder what is the standard to define 'enum' or 'struct',
Because I defined enums in thrift file but they're partially shown.


Ta2Rim:
I'll repost it because it's not a draft pull-request

trustin:
It was actually OK but that's fine anyway. Thanks!

PR #2015, opened 2019-08-22
Add literal value field for enum in internal documentation

Ta2Rim:
**Motivation**
#1966

**Modification**
Add literal value field for enum in documentation.
-> Add literal value as member variable to EnumValueInfo, modify following changes.

**Result**
User can see literal value of enum values in doc pages

There's a question I have working on this issue.
I wonder what is the standard to define 'enum' or 'struct',
Because I defined enums in thrift file but they're partially shown.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
+++ b/core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
@@ -17,11 +17,14 @@ package com.linecorp.armeria.server.docs;
 
 import static java.util.Objects.requireNonNull;
 
+import java.util.Objects;
+
 import javax.annotation.Nullable;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
 
 /**
@@ -32,6 +35,8 @@ public final class EnumValueInfo {
     private final String name;
     @Nullable
     private final String docString;
+    @Nullable
+    private final Integer intValue;
 
     /**
      * Creates a new instance.
@@ -39,17 +44,29 @@ public final class EnumValueInfo {
      * @param name the name of the enum value
      */
     public EnumValueInfo(String name) {
-        this(name, null);
+        this(name, null, null);
     }
 
     /**
      * Creates a new instance.
      *
      * @param name the name of the enum value
+     * @param intValue the integer value of the enum value
+     */
+    public EnumValueInfo(String name, @Nullable Integer intValue) {
+        this(name, intValue, null);
+    }
+
+    /**
+     * Creates a new instance.
+     *
+     * @param name the name of the enum value
+     * @param intValue the integer value of the enum value
      * @param docString the documentation string that describes the enum value
      */
-    public EnumValueInfo(String name, @Nullable String docString) {
+    public EnumValueInfo(String name, @Nullable Integer intValue, @Nullable String docString) {
         this.name = requireNonNull(name, "name");
+        this.intValue = intValue;
         this.docString = Strings.emptyToNull(docString);
     }
 
@@ -59,6 +76,16 @@ public final class EnumValueInfo {
     @JsonProperty
     public String name() {
         return name;
+    }
+
+    /**
+     * Returns the integer value of the enum value.
+     */
+    @JsonProperty
+    @JsonInclude(Include.NON_NULL)
+    @Nullable
+    public Integer intValue() {
+        return intValue;
     }
 
     /**
@@ -73,7 +100,7 @@ public final class EnumValueInfo {
 
     @Override
     public int hashCode() {
-        return name.hashCode();
+        return Objects.hash(name, intValue);
     }
 
     @Override
@@ -85,11 +112,17 @@ public final class EnumValueInfo {
             return false;
         }
 
-        return name.equals(((EnumValueInfo) o).name);
+        final EnumValueInfo that = (EnumValueInfo) o;
+        return name.equals(that.name) && Objects.equals(intValue, that.intValue);
     }
 
     @Override
     public String toString() {
-        return name();
+        return MoreObjects.toStringHelper(this)
+                          .omitNullValues()
+                          .add("name", name)
+                          .add("intValue", intValue)
+                          .add("docString", docString)
+                          .toString();
     }
 }
--- a/docs-client/src/containers/EnumPage/index.tsx
+++ b/docs-client/src/containers/EnumPage/index.tsx
@@ -45,6 +45,9 @@ export default class EnumPage extends React.PureComponent<Props> {
       return <>Not found.</>;
     }
 
+    const hasIntValue = data.values.some((value) => !!value.intValue);
+    const hasDocString = data.values.some((value) => !!value.docString);
+
     return (
       <>
         <Typography variant="h5">
@@ -62,7 +65,8 @@ export default class EnumPage extends React.PureComponent<Props> {
             <TableHead>
               <TableRow>
                 <TableCell>Name</TableCell>
-                <TableCell>Description</TableCell>
+                {hasIntValue && <TableCell>Int Value</TableCell>}
+                {hasDocString && <TableCell>Description</TableCell>}
               </TableRow>
             </TableHead>
             <TableBody>
@@ -72,7 +76,8 @@ export default class EnumPage extends React.PureComponent<Props> {
                     <TableCell>
                       <code>{value.name}</code>
                     </TableCell>
-                    <TableCell>{value.docString}</TableCell>
+                    {hasIntValue && <TableCell>{value.intValue}</TableCell>}
+                    {hasDocString && <TableCell>{value.docString}</TableCell>}
                   </TableRow>
                 ))
               ) : (
--- a/grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcDocServicePluginTest.java
+++ b/grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcDocServicePluginTest.java
@@ -211,9 +211,9 @@ public class GrpcDocServicePluginTest {
         final EnumInfo enumInfo = generator.newEnumInfo(CompressionType.getDescriptor());
         assertThat(enumInfo).isEqualTo(new EnumInfo(
                 "armeria.grpc.testing.CompressionType",
-                ImmutableList.of(new EnumValueInfo("NONE"),
-                                 new EnumValueInfo("GZIP"),
-                                 new EnumValueInfo("DEFLATE"))));
+                ImmutableList.of(new EnumValueInfo("NONE", 0),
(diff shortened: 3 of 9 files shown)


core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -32,24 +32,37 @@
     private final String name;
     @Nullable
     private final String docString;
+    private final String literalValue;
```
trustin: How about annotating this field with `@Nullable` and allowing `null` to represent the case where the literal value is not available?

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -32,24 +32,37 @@
...
+        this(name, null, null);
     }
 
     /**
      * Creates a new instance.
      *
      * @param name the name of the enum value
+     * @param literalValue literal value (for example, int value) for the enum value
+     */
+    public EnumValueInfo(String name, String literalValue) {
```
trustin: Ditto - please annotate with `@Nullable` because it's nullable.

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -32,24 +32,37 @@
...
+
+    /**
+     * Creates a new instance.
+     *
+     * @param name the name of the enum value
+     * @param literalValue literal value (for example, int value) for the enum value
      * @param docString the documentation string that describes the enum value
      */
-    public EnumValueInfo(String name, @Nullable String docString) {
+    public EnumValueInfo(String name, String literalValue, @Nullable String docString) {
```
trustin: Ditto - please annotate with `@Nullable` because it's nullable.

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -61,6 +74,14 @@ public String name() {
         return name;
     }
 
+    /**
+     * Returns the literal value (for example, int value) for the enum value.
+     */
+    @JsonProperty
+    public String literalValue() {
```
trustin: Missing `@Nullable` annotation

trustin:
Could you explain why you chose `String` to represent a literal value? Is there a service definition language who uses non-integer to represent an enum's literal values?

trustin:
Looks good enough to remove the WIP label

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
anuraaga: Let's add this field to equals, hashCode, and toString. This will cause some tests to need to be updated, which is good since it means the new code is tested :)

Ta2Rim:
> Could you explain why you chose `String` to represent a literal value? Is there a service definition language who uses non-integer to represent an enum's literal values?

I thought there's a possibility that string value can be considered as enum's literal values
But it doesn't need to be String
Since there's no service use non-integer to represent an enum's literal values definitions in Armeria 
I'll change it too.

trustin:
> I thought there's a possibility that string value can be considered as enum's literal values. But it doesn't need to be String. Since there's no service use non-integer to represent an enum's literal values definitions in Armeria, I'll change it too.

Thanks!

minwoox:
Could you check the test `testNewEnumInfo` in `ThriftDocServicePluginTest`, please?

docs-client/src/containers/EnumPage/index.tsx
```diff
@@ -72,6 +73,7 @@ export default class EnumPage extends React.PureComponent<Props> {
                     <TableCell>
                       <code>{value.name}</code>
                     </TableCell>
+                    <TableCell>{value.literalValue}</TableCell>
```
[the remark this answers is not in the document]
Ta2Rim: @trustin 
I google it and I found out jsx ignore booleans, null, undefined.
so It shows empty table cell
If there's good idea to alter empty cell, let me know. I'll work on it(maybe it would be another issue)
https://reactjs.org/docs/jsx-in-depth.html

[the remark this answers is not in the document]
minwoox: So we should show the value when it exists as we do in https://github.com/line/armeria/blob/master/docs-client/src/components/VariableList/index.tsx#L62

grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcDocServicePluginTest.java
```diff
@@ -211,9 +211,9 @@ public void newEnumInfo() throws Exception {
         final EnumInfo enumInfo = generator.newEnumInfo(CompressionType.getDescriptor());
         assertThat(enumInfo).isEqualTo(new EnumInfo(
                 "armeria.grpc.testing.CompressionType",
-                ImmutableList.of(new EnumValueInfo("NONE"),
-                                 new EnumValueInfo("GZIP"),
-                                 new EnumValueInfo("DEFLATE"))));
+                ImmutableList.of(new EnumValueInfo("NONE", 0),
+                                 new EnumValueInfo("GZIP", 1),
```
anuraaga: Can you change the proto file so these numbers aren't in sequence (e.g., change DEFLATE to `= 3`. Then this test verifies that we correctly display the proto value, not enum ordinal.

Ta2Rim: Yes that's a good idea

thrift/src/main/java/com/linecorp/armeria/server/thrift/ThriftDocServicePlugin.java
```diff
@@ -402,6 +405,17 @@ static TypeSignature toTypeSignature(FieldValueMetaData fieldValueMetaData) {
         return TypeSignature.ofUnresolved(firstNonNull(unresolvedName, "unknown"));
     }
 
+    @VisibleForTesting
+    static EnumInfo newEnumInfo(Class<? extends Enum<? extends TEnum>> enumType) {
+        requireNonNull(enumType, "exceptionClass");
```
ikhoon: I guess we could remove `requireNonNull` if this method is called only one place with non-null variable and `package-private`. 😀

docs-client/src/containers/EnumPage/index.tsx
```diff
@@ -45,6 +45,11 @@ export default class EnumPage extends React.PureComponent<Props> {
       return <>Not found.</>;
     }
 
+    const hasLiteralValue = data.values.some(
+      (variable) => !!variable.literalValue,
```
ikhoon: nit: remove unnecessary parentheses. and `variable` -> `value`?

docs-client/src/containers/EnumPage/index.tsx
```diff
@@ -45,6 +45,11 @@ export default class EnumPage extends React.PureComponent<Props> {
...
+    );
+    const hasDocString = data.values.some((variable) => !!variable.docString);
```
ikhoon: nit+ditto

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -32,24 +35,38 @@
...
     public EnumValueInfo(String name) {
-        this(name, null);
+        this(name, null, null);
+    }
+
+    /**
+     * Creates a new instance.
+     *
+     * @param name the name of the enum value
+     * @param literalValue literal value (for example, int value) for the enum value
```
minwoox: This is an `Integer` value so I think we don't need `(for example, int value)`.

minwoox: Question: The word `literal` is right for this? 
In [gRPC](https://developers.google.com/protocol-buffers/docs/proto3) and [Thrift](https://diwakergupta.github.io/thrift-missing-guide/#_enums), they call the value as a `constant` so I was wondering If we need to use that word too. /cc @anuraaga, @trustin

trustin: How about `intValue`?

minwoox: That's a good idea. `literal` is somewhat unclear to understand for me. 😉

Ta2Rim: I'm with you!

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -50,23 +50,23 @@ public EnumValueInfo(String name) {
     /**
      * Creates a new instance.
      *
-     * @param name the name of the enum value
-     * @param literalValue literal value (for example, int value) for the enum value
+     * @param name the integer value of the enum value
+     * @param intValue the current value of the enum value
```
trustin: name -> the name of the enum value
intValue -> the integer value of the enum value

docs-client/src/containers/EnumPage/index.tsx
```diff
@@ -62,7 +65,8 @@ export default class EnumPage extends React.PureComponent<Props> {
             <TableHead>
               <TableRow>
                 <TableCell>Name</TableCell>
-                <TableCell>Description</TableCell>
+                {hasIntValue && <TableCell>Literal Value</TableCell>}
```
ikhoon: `Int Value`?

Ta2Rim: oh I missed out thanks!

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -32,24 +35,38 @@
...
     public EnumValueInfo(String name) {
-        this(name, null);
+        this(name, null, null);
+    }
+
+    /**
+     * Creates a new instance.
+     *
+     * @param name the name of the enum value
+     * @param intValue the current value of the enum value
```
minwoox: nit: the current value -> the integer value

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -32,24 +35,38 @@
...
+     */
+    public EnumValueInfo(String name, @Nullable Integer intValue) {
+        this(name, intValue, null);
     }
 
     /**
      * Creates a new instance.
      *
      * @param name the name of the enum value
+     * @param intValue the current value of the enum value
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/server/docs/EnumValueInfo.java
```diff
@@ -61,6 +78,16 @@ public String name() {
         return name;
     }
 
+    /**
+     * Returns the current value of the enum value.
```
minwoox: ditto

trustin:
@anuraaga Any other comments?

PR #2018, opened 2019-08-23
Add `as()` to `Client`

### LINKED ISSUE #1883: Add Client.as which can return a decorated client in the form of one of its decorations.
Same as `Service.as` but for client.

jyblue:
Motivation:

A user often need to downcast decorated `Client` to certain type of `Client`.
If we have `Client.as()`, it would be possible.

Modifications:

* Add `as()` to `Client`
* Add `as()` to `DecoratingClient`

Result:
* Closes #1883

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Client.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Client.java
@@ -61,6 +61,6 @@ public interface Client<I extends Request, O extends Response> {
     default <T> Optional<T> as(Class<T> clientType) {
         requireNonNull(clientType, "clientType");
         return clientType.isInstance(this) ? Optional.of(clientType.cast(this))
-                                            : Optional.empty();
+                                              : Optional.empty();
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/Client.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Client.java
@@ -61,6 +61,6 @@ public interface Client<I extends Request, O extends Response> {
     default <T> Optional<T> as(Class<T> clientType) {
         requireNonNull(clientType, "clientType");
         return clientType.isInstance(this) ? Optional.of(clientType.cast(this))
-                                              : Optional.empty();
+                                           : Optional.empty();
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/Client.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Client.java
@@ -16,6 +16,10 @@
 
 package com.linecorp.armeria.client;
 
+import static java.util.Objects.requireNonNull;
+
+import java.util.Optional;
+
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.Request;
@@ -45,4 +49,18 @@ public interface Client<I extends Request, O extends Response> {
      * @return the {@link Response} to the specified {@link Request}
      */
     O execute(ClientRequestContext ctx, I req) throws Exception;
+
+    /**
+     * Undecorates this {@link Client} to find the {@link Client} which is an instance of the specified
+     * {@code clientType}.
+     *
+     * @param clientType the type of the desired {@link Client}
+     * @return the {@link Client} which is an instance of {@code clientType} if this {@link Client}
+     *         decorated such a {@link Client}. {@link Optional#empty()} otherwise.
+     */
+    default <T> Optional<T> as(Class<T> clientType) {
+        requireNonNull(clientType, "clientType");
+        return clientType.isInstance(this) ? Optional.of(clientType.cast(this))
+                                            : Optional.empty();
+    }
 }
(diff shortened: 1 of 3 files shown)


ikhoon:
@jyblue Thanks for contributing! btw, Could you add test code for your work? 😄

jyblue:
@trustin @ikhoon  Thank you for patience. I added a test case.
Please take a look at this.

To make this test better, 
I'm trying to fully understand how Client/Decorator works.

So my question is,
* **After client is created using `ClientBuilder` or `Clients`, 
how can user access their decorated delegate of `Client` directly?**

``` java
HttpClient client = new HttpClientBuilder(...)
        .decorator(MyDecoratorA.newDecorator())
        .decorator(MyDecoratorB.newDecorator())
        .build();

// error, HttpClient does not implement Client
MyDecoratorA decoratorA = client.as(MyDecoratorA.class);

// error, delegate() is not public
MyDecoratorB decoratorB = client.delegate().as(MyDecoratorB.class);
```

Would you please give me some guide or example of `Client.as()` use-case?

trustin:
@jyblue That's a great question. For `HttpClient`, we could add `as()` method there. For other client types such as gRPC and Thrift, it will be more complicated.

For Thrift and gRPC, we will have to add a separate API for this, e.g.

```java
MyThriftService.AsyncIface s = ...;
Clients.as(s, SomeClient.class);
```

I didn't think much about the implementation detail of this, but it could be:

1. Extract `Client.as()` into its own interface, such as `Unwrappable`:
   ```java
   public interface Unwrappable {
       <T> T as(Class<T> type);
   }

   public interface Client<...> extends Unwrappable { ... }
   public interface Service<...> extends Unwrappable { ... }
   ```
2. Add `<T, U> U Clients.as(T client, Class<U> type)`. See https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/Clients.java#L391 which converts an arbitrary client into a `ClientBuilderParams`. We could reuse this logic to convert a client into an `Unwrappable`.
3. Make `THttpClientInvocationHandler` and `HttpClient` implement `Unwrappable`.

This change could be done in a separate pull request though. Do you want to tackle into this in this pull request or leave it as it is for now?

core/src/main/java/com/linecorp/armeria/client/Client.java
```diff
@@ -45,4 +49,18 @@
...
+     * {@code clientType}.
+     *
+     * @param clientType the type of the desired {@link Client}
+     * @return the {@link Client} which is an instance of {@code clientType} if this {@link Client}
+     *         decorated such a {@link Client}. {@link Optional#empty()} otherwise.
+     */
+    default <T> Optional<T> as(Class<T> clientType) {
+        requireNonNull(clientType, "clientType");
+        return clientType.isInstance(this) ? Optional.of(clientType.cast(this))
+                                            : Optional.empty();
```
trustin: `?` and `:` should be aligned.

trustin:
Regardless of implementing `as()` for `HttpClient` and other clients that do not implement `Client`, how about extracting `Service.as()` and `Client.as()` into a new interface `Unwrappable`?

anuraaga:
For reference, I found that `ClientFactory` provides the interface to concert a user client into client params

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/ClientFactory.java#L204

I think this is because, while rare, some factories may not recognize all user client types. So I guess we would use it in this pattern too?

core/src/main/java/com/linecorp/armeria/client/Client.java
jyblue: Thanks, fixed it.

jyblue:
Thank you for review.

If you mind, 
I want to leave this PR now. Please open a new issue for `Unwrappable`.
I need time to digest :-)

core/src/test/java/com/linecorp/armeria/client/ClientOptionsBuilderTest.java
```diff
@@ -147,4 +147,45 @@ public void testShortcutMethods() {
...
+    @Test
+    public void testDecoratorDowncast() {
+        final FooClient inner = new FooClient();
+        final FooDecorator outer = new FooDecorator(inner);
+
+        assertThat(outer.as(inner.getClass())).isPresent();
+        assertThat(outer.as(outer.getClass())).isPresent();
+
+        if (outer.as(inner.getClass()).isPresent()) {
+            assertThat(outer.as(inner.getClass()).get() == inner);
```
ikhoon: nit: How about using `isSameAs()` for asserting? :-)
```java
assertThat(outer.as(inner.getClass()).get()).isSameAs(inner);
```

jyblue: @ikhoon 
When I use, `isSameAs()` without `isPresent()`
IDEA warns me
```
Warning:(156, 47) 'Optional.get()' without 'isPresent()' check
```
is it okay?

which one is preferred?
* Use `isSameAs` and leaving warnings
* Use `isSameAs` and suppress warnings using annotation

ikhoon: @jyblue Sorry for interrupting, I didn't know IntelliJ IDEA warns it. 
Now, I think this is good for me. 👍

jyblue: @ikhoon Thanks a lot !

Issue #1913, opened 2019-07-18
Disparity between HttpHealthCheckedEndpoint and HttpClient can be confusing.


anuraaga:
This might be hard to solve, but we found some difficulties using `HttpHealthCheckedEndpointGroup` because it is separate from `HttpClient`. We configured the `HttpClient` with https, and decorators, but we also had to do the duplicate configuration on the `HttpHealthCheckedEndpointGroup` too, which we forgot at first. As it's redundant, it would be nice if this only had to be set once.

trustin:
We could make `HttpHealthCheckedEndpointGroup` accept an `HttpClient` once we resolve #1143.

trustin:
Not sure this is THE solution but at least it will make it trivial to use the same settings: https://github.com/line/armeria/pull/2019

trustin:
Can't think of a good way to solve this. Should we close this now that it's relatively simple to use the same settings?

```java
HealthCheckedEndpointGroup.builder().clientOptions(webClient.options())...
```

anuraaga:
The code is much better. I think we can leave the API at that.

It's still a bit unfortunate that it's easy to forget though. Do you think it's possible for a client to handle 4xx errors and if it is using a health checked endpoint group which doesn't have the same options, show a warning that the user may have missed this?

trustin:
> Do you think it's possible for a client to handle 4xx errors and if it is using a health checked endpoint group which doesn't have the same options, show a warning that the user may have missed this?

Done in #2386

PR #2019, opened 2019-08-23
Add `HealthCheckedEndpointGroupBuilder.clientOptions()`

trustin:
Related: #1913
Motivation:

By adding a builder method that accepts `ClientOptions` directly instead
of the function that mutates `ClientOptions`, a user can easily reuse
his or her `ClientOptions` instance:

    HttpClient myClient = ...;
    // Before:
    builder.withClientOptions(b -> b.options(myClient.options()));
    // After:
    builder.clientOptions(myClient.options());

Modifications:

- Add `clientOptions(ClientOptions)` method to `HealthCheckedEndpointGroupBuilder`.

Result:

- User friendliness.


adriancole:
I very much like this  better

trustin:
> Though as mentioned in the issue, it seems hard to solve in a great way so maybe this is still the best we can do.

Yeah, we need to solve in a more fundamental way. This helps a little meanwhile.

trustin:
Thanks for reviewing.

PR #2021, opened 2019-08-23
Call channel.remoteAddress before the connection is closed

### LINKED ISSUE #2010: NPE in HttpSessionHandler.channelInactive with JRE11
The following issue is reproduced on latest armeria (0.90.x) with openjdk 11.0.2 when the remote server requires a downgrade from HTTP2 to H1C :

```
java.lang.NullPointerException: remoteAddress
	at io.netty.util.internal.ObjectUtil.checkNotNull(ObjectUtil.java:33)
	at io.netty.bootstrap.Bootstrap.connect(Bootstrap.java:141)
	at com.linecorp.armeria.client.HttpChannelPool.connect(HttpChannelPool.java:347)
	at com.linecorp.armeria.client.HttpSessionHandler.channelInactive(HttpSessionHandler.java:306)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:257)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:243)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:236)
	at io.netty.handler.codec.ByteToMessageDecoder.channelInputClosed(ByteToMessageDecoder.java:393)
	at io.netty.handler.codec.ByteToMessageDecoder.channelInactive(ByteToMessageDecoder.java:358)
	at io.netty.handler.codec.http2.Http2ConnectionHandler.channelInactive(Http2ConnectionHandler.java:415)
```

It appears that the Channel's remoteAddress is not set. (requestedRemoteAddress is set though)

This cannot be reproduced e.g. with oracle 10.0.2 where remoteAddress = requestedRemoteAddress

It could be an issue with netty (interop with recent openJDKs) or with the way Armeria uses Netty.

minwoox:
Motivation:
A connection can be closed while reading. If `channel.remoteAddress()` is called after
the connection is closed it will return `null`. So we should call `channel.remoteAddress()`
before a reading event happens so that we can cache the address and reuse it later.

Modifications:
- Call `channel.remoteAddress()` after the connection is established in `HttpSessionHandler` and where necessary
- Remove unused class: `LoggingUtil`

Result:
- You no longer see NPE when the remote server requires a downgrade
- Fix #2010 hopefully


core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -119,7 +119,11 @@ public final SessionProtocol sessionProtocol() {
     @SuppressWarnings("unchecked")
     public <A extends SocketAddress> A remoteAddress() {
         final Channel ch = channel();
-        return ch != null ? (A) ch.remoteAddress() : null;
+        if (ch == null) {
+            return null;
+        }
+        final SocketAddress remoteAddress = ch.remoteAddress();
+        return remoteAddress != null ? (A) remoteAddress : null;
```
trustin: It seems to me that this change doesn't really change anything. Could you explain why this is better?

core/src/main/java/com/linecorp/armeria/internal/logging/LoggingUtil.java
```diff
@@ -1,52 +0,0 @@
...
-
-import com.linecorp.armeria.common.HttpHeaderNames;
-import com.linecorp.armeria.common.HttpHeaders;
-
-import io.netty.channel.Channel;
-
-/**
- * A utility class for logging.
- */
-public final class LoggingUtil {
```
trustin: Thanks!

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -209,13 +209,16 @@ public HttpRequest request() {
...
     public <A extends SocketAddress> A remoteAddress() {
         final Channel ch = channel();
         assert ch != null;
+        final SocketAddress remoteAddress = ch.remoteAddress();
+        if (remoteAddress == null) {
+            return null;
+        }
         @SuppressWarnings("unchecked")
-        final A addr = (A) ch.remoteAddress();
+        final A addr = (A) remoteAddress;
```
trustin: Ditto. This doesn't seem to change anything to me.

core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
```diff
@@ -105,6 +107,9 @@
 
         this.channelPool = requireNonNull(channelPool, "channelPool");
         this.channel = requireNonNull(channel, "channel");
+        final SocketAddress remoteAddress = channel.remoteAddress();
+        assert remoteAddress != null;
```
anuraaga: I think we normally use assert when it's obvious in the same method / class. How about using requireNonNull like the others since it may still have a programming error?

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
minwoox: Yeah, this change brings nothing. Let me revert this, thanks!

core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
minwoox: I thought it was obvious that the `remoteAddress` is not `null` because this is called right after the connection is established by the `eventLoop` and there's no reading event between that. So the socket is not marked as closed which means that we can always bring the `remoteAddress`.
But I think `requireNonNull` is fine as well, so let me use that. Thanks! :)

minwoox:
Let me add a test case for this. (Not sure if I can make it thought. 😆)

core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
```diff
@@ -241,6 +241,20 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
...
+     * before {@code channel().remoteAddress()} is called which caches the address in it.
+     */
+    private SocketAddress remoteAddress(ChannelHandlerContext ctx) {
+        SocketAddress remoteAddress = ctx.channel().remoteAddress();
+        if (remoteAddress == null) {
+            // This is called after connect() method is called so we have the remoteAddress.
+            assert this.remoteAddress != null;
+            remoteAddress = this.remoteAddress;
+        }
+        return remoteAddress;
```
ikhoon: nit: How about simplifying this logic? 😉
```java
SocketAddress remoteAddress = ctx.channel().remoteAddress();
return firstNonNull(remoteAddress, this.remoteAddress);
```

ikhoon: I’m not expert in this logic. I just wonder we don’t have to update `this.remoteAddress` just in case  `remoteAdress != null` and `remoteAdress != this.remoteAddress`.

minwoox: > return firstNonNull(remoteAddress, this.remoteAddress);

That's a good suggestion. I will take it. Thanks!

> I just wonder we don’t have to update this.remoteAddress = remoteAddress

I think we don't have to. The `remoteAddress` is only used later to extract host in `UpgradeRequestHandler`. We only need the host and the port, so I think we don't have to do that.

minwoox:
I tried to close the socket as soon as it's connected from the server-side but I couldn't reproduce the failure. Let's see if this change fixes or not. 🙇

trustin:
Thanks, @minwoox and reviewers!

minwoox:
Thanks for reviewing!

Issue #1855, opened 2019-06-25
Provide a way to set an example request for AnnotatedServiceRegistrationBean


minwoox:
Currently, a user can set an example request for `DocService` when using `GrpcServiceRegistrationBean` and `ThriftServiceRegistrationBean`.
We need to provide the same way for `AnnotatedServiceRegistrationBean` so that a user is not surprised.

heowc:
I'll work on it

PR #2026, opened 2019-08-24
Provide a way to set an example request for AnnotatedServiceRegistrationBean

heowc:
Related: #1855 
Motivation:
when add for annotatedServices to `DocService` with `AnnotatedServiceRegistrationBean`,  this couldn't include`ExampleRequest`.

Modifications:
- Add for `AnnotatedExampleRequest` to `AnnotatedServiceRegistrationBean`
- Add for `GrpcExampleRequest` to `GrpcServiceRegistrationBean`
- Remove `ExampleRequest`
- Add for annotatedServices to `DocService` with `AnnotatedServiceRegistrationBean`
- Fix Test Case

Result:
- User friendliness.


trustin:
Thanks for a quick pull request. Let us take a look early next week. Have a happy weekend!

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ExampleRequest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public String getMethodName() {
+        return methodName;
+    }
+
+    public Object getExampleRequest() {
+        return exampleRequest;
+    }
+
+    public static ExampleRequest of(@NotNull String serviceType,
```
trustin: - Could you move this method to the top of the class so that public static methods come before member fields/methods?
- Could you add `toString()`?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ExampleRequest.java
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.spring;
+
+import javax.validation.constraints.NotNull;
+
+public final class ExampleRequest {
```
trustin: Please add Javadoc for public classes and methods.

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaAutoConfigurationTest.java
```diff
@@ -91,14 +98,20 @@
...
                     .setServiceName("annotatedService")
-                    .setService(new AnnotatedService())
+                    .setService(service)
                     .setPathPrefix("/annotated")
                     .setDecorators(LoggingService.newDecorator())
                     .setExceptionHandlers(ImmutableList.of(new IllegalArgumentExceptionHandler()))
                     .setRequestConverters(ImmutableList.of(new StringRequestConverterFunction()))
-                    .setResponseConverters(ImmutableList.of(new StringResponseConverter()));
+                    .setResponseConverters(ImmutableList.of(new StringResponseConverter()))
+                    .setExampleRequests(Collections.singleton(ExampleRequest.of(service.getClass().getName(),
```
ikhoon: nit: How about changing `Collections.singleton(...)` -> `ImmutableList.of(...)` to keep consistency?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ExampleRequest.java
heowc: Thank you for review. I fixed it in 0f88123.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ExampleRequest.java
heowc: Thank you for review. I fixed it in 30d27f9.

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaAutoConfigurationTest.java
heowc: Thank you for review. I fixed it in 5dd296c.

heowc:
> Thanks a lot for your work!
> I found out that our `XServiceRegistrationBean` has a problem which can screw up a user while he/she configures the example requests and headers.
> 
> * Cannot set example HTTP headers for gRPC Service.
> * Cannot set example HTTP headers in method level.
> * Cannot set example requests for Thrift in method level.
> 
> So I think we should handle this so that a user can configure example requests and headers in service or method levels.
> Do you mind handling it in this PR? or I think we can do it in another PR. :-)

I also think it's a good idea to handle the problem in another PR.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
```diff
@@ -157,4 +178,37 @@ public AnnotatedServiceRegistrationBean setResponseConverters(
...
+     */
+    @NotNull
+    public Collection<ExampleRequest> getExampleRequests() {
+        return exampleRequests;
+    }
+
+    /**
+     * Sets sample requests for {@link #getService()}.
+     */
+    public AnnotatedServiceRegistrationBean setExampleRequests(
```
minwoox: How about moving this method and `getExampleRequests` up to `AbstractServiceRegistrationBean` so that we can remove duplicate code and use it for gRPC, Thrift and annotated services?
(we should override the method in `HttpServiceRegistrationBean` and throw the `UnsupportedOperationException` though.)

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaAutoConfigurationTest.java
```diff
@@ -91,14 +98,20 @@
...
-                    .setService(new AnnotatedService())
+                    .setService(service)
                     .setPathPrefix("/annotated")
                     .setDecorators(LoggingService.newDecorator())
                     .setExceptionHandlers(ImmutableList.of(new IllegalArgumentExceptionHandler()))
                     .setRequestConverters(ImmutableList.of(new StringRequestConverterFunction()))
-                    .setResponseConverters(ImmutableList.of(new StringResponseConverter()));
+                    .setResponseConverters(ImmutableList.of(new StringResponseConverter()))
+                    .setExampleRequests(ImmutableList.of(ExampleRequest.of(service.getClass().getName(),
+                                                                                "post",
```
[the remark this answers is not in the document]
heowc: Thank you for review. I fixed it in 33c7d6c.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
heowc: Thank you for review. I fixed it in add07a2.

minwoox: @heowc I realized that each service needs different parameters for example requests:
- gRPC: service name, parameter name and the request object
  - `GrpcServiceRegistrationBean` can add multiple services using `GrpcServiceBuilder` so we need the service name to distinguish
- Annotated: parameter name and the request object
  - `AnnotatedServiceRegistrationBean` can add only one service so we don't need the service name.
- Thrift: request object
  - The service and method name can be inferred from the request object.

So we need different setters for each `ServiceRegistrationBean`, sorry about that. 😉 
Here's the thing we should do:
- Put the `ExampleRequest` back into `GrpcServiceRegistrationBean`
  - We are not going to use that class eventually but we need that now not to bring another breaking change.
  - Deprecate `ExampleRequest` class and related methods using `@Deprecated`.
    - This is because the class is only used in `GrpcServiceRegistrationBean`. It's not going to be used in other beans so it can confuse users.
  - Add new setter which will be: `setExampleRequests(String serviceType, String methodName, Object sampleRequest)`
- Add a setter to `AnnotatedServiceRegistrationBean`
  - We are not going to use `ExampleRequest` for here.
  - `setExampleRequests(String methodName, Object sampleRequest)`.
- Bring back the setters to `ThriftServiceRegistrationBean`.

If you have any questions, please let us know. 😄

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ExampleRequest.java
```diff
@@ -0,0 +1,75 @@
...
+    private final Object exampleRequest;
+
+    private ExampleRequest(String serviceType, String methodName, Object exampleRequest) {
+        this.serviceType = serviceType;
+        this.methodName = methodName;
+        this.exampleRequest = exampleRequest;
+    }
+
+    /**
+     * Return the service type of this {@link ExampleRequest}.
```
ikhoon: `Return -> Returns`. 
Conventionally, Javadoc uses 3rd person present regarding verbs.
https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AbstractServiceRegistrationBean.java
```diff
@@ -36,7 +38,7 @@
  * @param <T> the type of the service object to be registered
  * @param <U> the type of the implementation of this bean
  */
-public class AbstractServiceRegistrationBean<T, U> {
+public class AbstractServiceRegistrationBean<T, U, V> {
```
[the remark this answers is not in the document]
heowc: Thank you for review. But, the code has been removed due to force push.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ExampleRequest.java
heowc: Thank you for review. But, the code has been removed due to force push.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedExampleRequest.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.validation.constraints.NotNull;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * Used as a example request object in {@link AnnotatedServiceRegistrationBean}.
+ */
+public final class AnnotatedExampleRequest {
+
+    public static AnnotatedExampleRequest of(@NotNull String methodName,
```
trustin: Javadoc?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/GrpcExampleRequest.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.validation.constraints.NotNull;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * Used as a example request object in {@link GrpcServiceRegistrationBean}.
+ */
+public final class GrpcExampleRequest {
+
+    public static GrpcExampleRequest of(@NotNull String serviceType,
```
trustin: Javadoc?

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaAutoConfigurationTest.java
```diff
@@ -168,6 +179,13 @@ public AggregatedHttpResponse get() {
         public AggregatedHttpResponse getV2() {
             throw new IllegalArgumentException();
         }
+
+        @Post("/post")
+        @ProducesJson
+        public AggregatedHttpResponse post(@RequestObject JsonNode jsonNode) throws JsonProcessingException {
```
trustin: I guess you can just return the JsonNode as-is, without any conversion because Armeria will do that for you.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedExampleRequest.java
heowc: Thank you for review! 64d4581

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/GrpcExampleRequest.java
heowc: Thank you for review! 64d4581

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaAutoConfigurationTest.java
heowc: Thank you for review! a579b0a

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
```diff
@@ -157,4 +170,32 @@ public AnnotatedServiceRegistrationBean setResponseConverters(
...
+            @NotNull Collection<AnnotatedExampleRequest> exampleRequests) {
+        this.exampleRequests = exampleRequests;
+        return this;
+    }
+
+    /**
+     * Adds sample request for {@link #getService()}.
+     */
+    public AnnotatedServiceRegistrationBean addExampleRequest(String methodName, Object exampleRequest) {
+        requireNonNull(methodName, "methodName");
```
minwoox: Let's use `@NotNull` instead of `requireNonNull` for consistency.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
```diff
@@ -157,4 +170,32 @@ public AnnotatedServiceRegistrationBean setResponseConverters(
...
+    public Collection<AnnotatedExampleRequest> getExampleRequests() {
+        return exampleRequests;
+    }
+
+    /**
+     * Sets sample requests for {@link #getService()}.
+     */
+    public AnnotatedServiceRegistrationBean setExampleRequests(
+            @NotNull Collection<AnnotatedExampleRequest> exampleRequests) {
+        this.exampleRequests = exampleRequests;
```
minwoox: How about adding `exampleRequests` to the end of the existing field as we do in `HttpHeaders`?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
```diff
@@ -39,7 +43,8 @@
  * >             .setDecorators(LoggingService.newDecorator())
  * >             .setExceptionHandlers(new MyExceptionHandler())
  * >             .setRequestConverters(new MyRequestConverter())
- * >             .setResponseConverters(new MyResponseConverter());
+ * >             .setResponseConverters(new MyResponseConverter())
+ * >             .setExampleRequests(Lists.of(AnnotatedExampleRequest.of("post", "{\"foo\":\"bar\"}")));
```
minwoox: nit: How about using `myMethod` or `hello` instead of `post` for the method name?
Because the method is not an HTTP method but the method in a class, I think `post` can give confusion to a user.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/GrpcExampleRequest.java
```diff
@@ -0,0 +1,76 @@
...
+ * under the License.
+ */
+package com.linecorp.armeria.spring;
+
+import javax.validation.constraints.NotNull;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * Used as a example request object in {@link GrpcServiceRegistrationBean}.
```
minwoox: ditto

spring/boot-autoconfigure/src/test/java/com/linecorp/armeria/spring/ArmeriaAutoConfigurationTest.java
```diff
@@ -168,6 +177,11 @@ public AggregatedHttpResponse get() {
         public AggregatedHttpResponse getV2() {
             throw new IllegalArgumentException();
         }
+
+        @Post("/post")
+        public JsonNode post(@RequestObject JsonNode jsonNode) {
+            return objectMapper.valueToTree(jsonNode);
```
trustin: I guess you can just `return jsonNode` and remove `objectMapper`?

heowc: Fixed 8171329

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/AnnotatedServiceRegistrationBean.java
heowc: Sorry, I don't understand your comment. 😢 Could you explain more detail?

minwoox: Had a chat with @heowc. 😄

heowc: Thank you for your help at night, @minwoox!!

In summary, 
There is generally a consistency that `setX` is adds to the end of the existing field. However, they are not classes like `~~RegistrationBean`. So these classes will go to [another PR](https://github.com/line/armeria/pull/2026#pullrequestreview-281918050) that needs refactoring.

PR #2029, opened 2019-08-26
Add `Unwrappable` and `ClientFactory.unwrap()`

trustin:
Related: #2018
Motivation:

- `Client.as()` and `Service.as()` have the same logic which can be
  deduplicated.
- `Client.as()` is useless when a user tries to unwrap `HttpClient`,
  gRPC client or Thrift client, because they do not implement `Client`.

Modifications:

- Add `Unwrappable` and make `Client`, `Service` and `Backoff` extend it.
- Make `HttpClient` implement `Unwrappable`.
- Add `ClientFactory.unwrap(client, type)` so that a user can unwrap a
  gRPC or Thrift client.
- `ClientFactory.clientBuilderParam()` now delegates to `unwrap()`.
- Revise the Javadoc of `as()`.

Result:

- A non-`Client` client object can be unwrapped via `ClientFactory.unwrap()`.
- A user can implement his or her `Unwrappable`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Client.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Client.java
@@ -58,8 +58,8 @@ public interface Client<I extends Request, O extends Response> extends Unwrappab
      * HttpClient client = new HttpClientBuilder()
      *     .decorator(LoggingClient.newDecorator())
      *     .build();
-     * LoggingClient unwrapped1 = client.as(LoggingClient.class);
-     * LoggingClient unwrapped2 = ClientFactory.DEFAULT.unwrap(client, LoggingClient.class);
+     * LoggingClient unwrapped1 = client.as(LoggingClient.class).get();
+     * LoggingClient unwrapped2 = ClientFactory.DEFAULT.unwrap(client, LoggingClient.class).get();
      * assert unwrapped1 == unwrapped2;
      * }</pre>
      *
--- a/core/src/main/java/com/linecorp/armeria/client/Client.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Client.java
@@ -59,12 +59,15 @@ public interface Client<I extends Request, O extends Response> extends Unwrappab
      *     .decorator(LoggingClient.newDecorator())
      *     .build();
      * LoggingClient unwrapped1 = client.as(LoggingClient.class).get();
-     * LoggingClient unwrapped2 = ClientFactory.DEFAULT.unwrap(client, LoggingClient.class).get();
+     * LoggingClient unwrapped2 = Clients.unwrap(client, LoggingClient.class).get();
      * assert unwrapped1 == unwrapped2;
      * }</pre>
      *
      * @param type the type of the object to return
      * @return the object of the specified {@code type} if found. {@link Optional#empty()} if not found.
+     *
+     * @see Clients#unwrap(Object, Class)
+     * @see ClientFactory#unwrap(Object, Class)
      */
     @Override
     default <T> Optional<T> as(Class<T> type) {
--- a/core/src/main/java/com/linecorp/armeria/client/Client.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Client.java
@@ -58,9 +58,12 @@ public interface Client<I extends Request, O extends Response> extends Unwrappab
      * HttpClient client = new HttpClientBuilder()
      *     .decorator(LoggingClient.newDecorator())
      *     .build();
-     * LoggingClient unwrapped1 = client.as(LoggingClient.class).get();
+     *
+     * LoggingClient unwrapped = client.as(LoggingClient.class).get();
+     *
+     * // You can also use Clients.unwrap(), which is useful especially for
+     * // Thrift and gRPC where the client object does not implement the 'as()' method.
      * LoggingClient unwrapped2 = Clients.unwrap(client, LoggingClient.class).get();
-     * assert unwrapped1 == unwrapped2;
      * }</pre>
      *
      * @param type the type of the object to return
@@ -68,6 +71,7 @@ public interface Client<I extends Request, O extends Response> extends Unwrappab
      *
      * @see Clients#unwrap(Object, Class)
      * @see ClientFactory#unwrap(Object, Class)
+     * @see Unwrappable
      */
     @Override
     default <T> Optional<T> as(Class<T> type) {
--- a/core/src/main/java/com/linecorp/armeria/client/Client.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Client.java
@@ -26,6 +26,7 @@ import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Response;
 import com.linecorp.armeria.common.RpcRequest;
 import com.linecorp.armeria.common.RpcResponse;
+import com.linecorp.armeria.common.util.Unwrappable;
 
 /**
  * Sends a {@link Request} to a remote {@link Endpoint}.
@@ -41,7 +42,7 @@ import com.linecorp.armeria.common.RpcResponse;
  * @see UserClient
  */
 @FunctionalInterface
-public interface Client<I extends Request, O extends Response> {
+public interface Client<I extends Request, O extends Response> extends Unwrappable {
     /**
      * Sends a {@link Request} to a remote {@link Endpoint}, as specified in
      * {@link ClientRequestContext#endpoint()}.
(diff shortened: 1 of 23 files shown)


core/src/main/java/com/linecorp/armeria/common/util/Unwrappable.java
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * class Qux<T> extends AbstractWrapper<T> {
+     *     Qux(T delegate) {
+     *         super(delegate);
+     *     }
+     * }
+     *
+     * Qux qux = new Qux(new Bar(new Foo()));
+     * Foo foo = qux.as(Foo.class);
+     * Bar bar = qux.as(Bar.class);
```
minwoox: This should be optional?

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -201,7 +206,51 @@ static void disableShutdownHook() {
...
+     *
+     * @param client the client object
+     * @param type the type of the object to return
+     * @return the object of the specified {@code type} if found. {@link Optional#empty()} if not found.
+     */
+    default <T> Optional<T> unwrap(Object client, Class<T> type) {
+        requireNonNull(client, "client");
+        requireNonNull(type, "type");
+
+
```
jyblue: nit : please check extra line :-)

core/src/main/java/com/linecorp/armeria/common/util/Unwrappable.java
trustin: Appended `.get()`

core/src/main/java/com/linecorp/armeria/client/Client.java
```diff
@@ -51,16 +52,23 @@
...
     /**
-     * Undecorates this {@link Client} to find the {@link Client} which is an instance of the specified
-     * {@code clientType}.
+     * Unwraps this {@link Client} into the object of the specified {@code type}.
+     * Use this method instead of an explicit downcast. For example:
+     * <pre>{@code
+     * HttpClient client = new HttpClientBuilder()
+     *     .decorator(LoggingClient.newDecorator())
+     *     .build();
+     * LoggingClient unwrapped1 = client.as(LoggingClient.class);
```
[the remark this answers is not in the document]
trustin: Appended `.get()`

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
trustin: Oops. Thanks :-)

trustin:
@minwoox `Backoff` and `BackoffWrapper` now extends `Unwrappable` and `AbstractUnwrappable`.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -201,7 +206,50 @@ static void disableShutdownHook() {
...
+     * LoggingClient unwrapped1 = client.as(LoggingClient.class);
+     * LoggingClient unwrapped2 = ClientFactory.DEFAULT.unwrap(client, LoggingClient.class);
+     * assert unwrapped1 == unwrapped2;
+     * }</pre>
+     *
+     * @param client the client object
+     * @param type the type of the object to return
+     * @return the object of the specified {@code type} if found. {@link Optional#empty()} if not found.
+     */
+    default <T> Optional<T> unwrap(Object client, Class<T> type) {
```
minwoox: Don't we need a test for this?

trustin: `clientBuilderParams()` cover this method, so I didn't write one, but yeah I can add one.

trustin: Added tests, found and fixed some bugs :scream:

core/src/main/java/com/linecorp/armeria/client/Client.java
```diff
@@ -51,16 +52,26 @@
...
-     * Undecorates this {@link Client} to find the {@link Client} which is an instance of the specified
-     * {@code clientType}.
+     * Unwraps this {@link Client} into the object of the specified {@code type}.
+     * Use this method instead of an explicit downcast. For example:
+     * <pre>{@code
+     * HttpClient client = new HttpClientBuilder()
+     *     .decorator(LoggingClient.newDecorator())
+     *     .build();
+     * LoggingClient unwrapped1 = client.as(LoggingClient.class).get();
+     * LoggingClient unwrapped2 = Clients.unwrap(client, LoggingClient.class).get();
```
anuraaga: I find this example a little confusing since they look the same and I don't know which one to use. Would it make sense to instead mention explicitly - `for gRPC and Thrift clients, see Clients.unwrap`?

core/src/main/java/com/linecorp/armeria/common/util/Unwrappable.java
```diff
@@ -0,0 +1,56 @@
...
+ * under the License.
+ */
+package com.linecorp.armeria.common.util;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.Optional;
+
+/**
+ * Provides a way to unwrap an object in decorator pattern, as an alternative to down-casting.
```
anuraaga: I don't think it's an alternative, so maybe `similar to down-casting in an inheritance pattern`.

core/src/test/java/com/linecorp/armeria/client/HttpClientUnwrapTest.java
```diff
@@ -0,0 +1,59 @@
...
+        final ClientFactory factory = client.factory();
+
+        assertThat(factory.unwrap(client, HttpClient.class)).containsSame(client);
+
+        assertThat(factory.unwrap(client, RetryingHttpClient.class))
+                .containsInstanceOf(RetryingHttpClient.class);
+        assertThat(factory.unwrap(client, LoggingClient.class)).containsInstanceOf(LoggingClient.class);
+
+        assertThat(factory.unwrap(client, Unwrappable.class)).containsSame(client);
+        assertThat(factory.unwrap(client, Client.class)).containsInstanceOf(RetryingHttpClient.class);
```
anuraaga: `containsInstanceOf(Client.class)`? I think the method's contract is just to return something of the provided type and we don't need to be so explicit.

core/src/test/java/com/linecorp/armeria/client/HttpClientUnwrapTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+        assertThat(client.as(HttpClient.class)).containsSame(client);
+
+        assertThat(client.as(RetryingHttpClient.class)).containsInstanceOf(RetryingHttpClient.class);
+        assertThat(client.as(LoggingClient.class)).containsInstanceOf(LoggingClient.class);
+
+        assertThat(client.as(Unwrappable.class)).containsSame(client);
+        assertThat(client.as(Client.class)).containsInstanceOf(RetryingHttpClient.class);
+
+        assertThat(client.as(String.class)).isEmpty();
```
anuraaga: `String.class` being empty seems obvious and could be for other reasons. How about using a different decorator like `HttpDecodingClient.class` so it looks more like a real usage.

core/src/test/java/com/linecorp/armeria/client/HttpClientUnwrapTest.java
trustin: This assertion is intentional to make sure `LoggingClient` is not found first.

core/src/main/java/com/linecorp/armeria/client/Client.java
trustin: Good idea. Updated.

core/src/main/java/com/linecorp/armeria/common/util/Unwrappable.java
trustin: Updated.

core/src/test/java/com/linecorp/armeria/client/HttpClientUnwrapTest.java
trustin: Now using real client classes :-)

core/src/test/java/com/linecorp/armeria/client/HttpClientUnwrapTest.java
anuraaga: Since the user requested `Client` wouldn't `LoggingClient` be ok too? Since the semantics are similar to a downcast, I feel like it's an overly strict assertion. Maybe if multiple match the predicate we should throw an exception instead so user knows they need to specify more narrowly?

trustin: I'm not sure if we need to raise an exception when there are multiple matches because a user may be doing it intentionally. Also, it's going to be tricky when complex type hierarchy is involved.

Let me just add some comment to make this clear and other devs know what the test started to fail when we change the decorator internals.

trustin:
Thanks for reviewing. :smile:

Issue #1914, opened 2019-07-18
HTTP/1.1 requestLog doesn't record exception when getting ClosedSessionException


kojilin:
Testing with okhttp's MockWebServer.
`new MockResponse().setSocketPolicy(SocketPolicy.DISCONNECT_AT_START)`
`new MockResponse().setSocketPolicy(SocketPolicy.DISCONNECT_AFTER_REQUEST)`

Client side will get exception but there is no cause in RequestLog.

--
HTTP/2 works well with above socket policy.

kojilin:
I will investigate this issue, due to I can't reproduce with separate test.

trustin:
Thanks a lot, @kojilin!

kojilin:
https://gist.github.com/kojilin/ed5daa7369833abc3d93d57a301b06b7#file-httpclientrequestlogwithclosedsessiontest-java-L62

<img width="1420" alt="スクリーンショット 2019-07-26 20 45 26" src="https://user-images.githubusercontent.com/863983/61949597-54b50880-afe6-11e9-8ccd-04086b4ad421.png">

not sure why h1c pass, maybe due to brave's integration test is using existing HttpClient?

PR #2032, opened 2019-08-28
Remove unnecessary ignore of brave client integration test

kojilin:
Looks like some changes fix this problem(Not sure which one). So remove the ignore assumption and add back the test for HTTP/1.1 .

close https://github.com/line/armeria/issues/1914


trustin:
Thanks, @kojilin !

PR #2034, opened 2019-08-29
Fix resource leak in `DefaultHttpClient.execute()`

trustin:
Motivation:

The `HttpRequest` given to `DefaultHttpClient.execute()` must be aborted
or consumed completely at any circumstances. However, it currently does
not. For example, when a circuit is open and `CircuitBreakerClient`
raises an exception, the request will be left open forever.

Modifications:

- Abort the request when request failed unexpectedly.
- Miscellaneous:
  - Replace Guava `Tickers` with ours.

Result:

- `ByteBufHttpData` does not leak anymore.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
@@ -32,6 +32,7 @@ import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.RequestHeaders;
 import com.linecorp.armeria.common.SessionProtocol;
+import com.linecorp.armeria.common.logging.RequestLogAvailability;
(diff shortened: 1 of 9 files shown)


minwoox:
I realized that this is newly introduced because we used to abort the request using fallback when execute. This was in `DefaultHttpClient`:
```java
(ctx, cause) -> {
    if (ctx != null && !ctx.log().isAvailable(RequestLogAvailability.REQUEST_START)) {
        // An exception is raised even before sending a request, so abort the request to
        // release the elements.
        newReq.abort();
    }
    return HttpResponse.ofFailure(cause);
});
```
So if an exception is raised in a decorator, we abort the request and can be free from releasing the byte buffer.
I think there's still a chance that a user writes a decorator which raises an exception and forgets to release the buffer, so we still need that mechanism.

trustin:
@ikhoon @minwoox Updated the pull request to address @minwoox's comment.

Issue #2000, opened 2019-08-17
Response Timeouts both logged and returned to user


anuraaga:
Currently, it seems that here we log a response timeout despite returning it to the user

https://github.com/line/armeria/blob/f2b67df619b8bf5df9b92c4eb87dbc4bfea24f3f/core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java#L289

I think we're supposed to not be logging ResponseTimeoutException since we return it anyways and it ends up double logged.

anuraaga:
For reference this came up in https://github.com/openzipkin/zipkin/issues/2756

Issue #2001, opened 2019-08-17
Add UUID to request logging


anuraaga:
Currently logs look like

```
.c.l.LoggingClient                  : Request: {startTime=2019-08-17T12:39:19.994Z(1566045559994257), length=4970KiB(5089733B), duration=6198?s(6198300ns), scheme=none+h1c, headers=[]}
zipkin                      | 2019-08-17 12:39:20.117  WARN 1 --- [orker-epoll-2-4] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-08-17T12:39:20.117Z(1566045560117742), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
```

It would be useful, and probably simple, to assign a UUID to every request to add to the log to be able to associate requests and responses. Many users will be setting trace IDs that would perform similar utility, but not all of them and I think there's very little cost even if it ends up performing double-purpose.

trustin:
We need to update:

- `RequestContext.newDerivedContext()` to accept a new UUID
- `RequestContext.toString()` to return a nicely formatted string which includes the last few digits of the UUID.

PR #2035, opened 2019-08-30
Explain Response timeout

mauhiz:
We're observing a cryptic warn :

```
Unexpected exception:
```

and the cause has no trace:

```
com.linecorp.armeria.client.ResponseTimeoutException: null
```

This PR aims to give more info
- which host timed out
- after what delay

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
@@ -286,8 +286,15 @@ abstract class HttpResponseDecoder {
             if (cancelTimeout()) {
                 actionOnTimeoutCancelled.accept(cause);
             } else {
-                if (cause != null && !Exceptions.isExpected(cause)) {
-                    logger.warn("Unexpected exception:", cause);
+                if (cause != null && logger.isWarnEnabled() && !Exceptions.isExpected(cause)) {
+                    StringBuilder logMsg = new StringBuilder("Unexpected exception while closing");
+                    if (request != null) {
+                        logMsg.append(" request to ").append(request.authority());
+                    }
+                    if (cause instanceof ResponseTimeoutException) {
+                        logMsg.append(" after ").append(responseTimeoutMillis).append("ms");
+                    }
+                    logger.warn(logMsg.toString(), cause);
                 }
             }
 
--- a/core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
@@ -287,9 +287,12 @@ abstract class HttpResponseDecoder {
                 actionOnTimeoutCancelled.accept(cause);
             } else {
                 if (cause != null && logger.isWarnEnabled() && !Exceptions.isExpected(cause)) {
-                    StringBuilder logMsg = new StringBuilder("Unexpected exception while closing");
+                    final StringBuilder logMsg = new StringBuilder("Unexpected exception while closing");
                     if (request != null) {
-                        logMsg.append(" request to ").append(request.authority());
+                        final String authority = request.authority();
+                        if (authority != null) {
+                            logMsg.append(" a request to ").append(authority);
+                        }
                     }
                     if (cause instanceof ResponseTimeoutException) {
                         logMsg.append(" after ").append(responseTimeoutMillis).append("ms");


core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -286,8 +286,15 @@ private void close(@Nullable Throwable cause,
             if (cancelTimeout()) {
                 actionOnTimeoutCancelled.accept(cause);
             } else {
-                if (cause != null && !Exceptions.isExpected(cause)) {
-                    logger.warn("Unexpected exception:", cause);
+                if (cause != null && logger.isWarnEnabled() && !Exceptions.isExpected(cause)) {
+                    StringBuilder logMsg = new StringBuilder("Unexpected exception while closing");
+                    if (request != null) {
```
minwoox: I think we need to add:
`request.authority() != null`

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -286,8 +286,15 @@ private void close(@Nullable Throwable cause,
...
+                        logMsg.append(" request to ").append(request.authority());
```
minwoox: nit: ` a request to `?

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -286,8 +286,15 @@ private void close(@Nullable Throwable cause,
...
+                    }
+                    if (cause instanceof ResponseTimeoutException) {
+                        logMsg.append(" after ").append(responseTimeoutMillis).append("ms");
+                    }
```
minwoox: nit: `logMsg.append(':')`, here, will make it prettier, perhaps?

trustin:
I guess there was an issue related with double notification regarding `ResponseTimeoutException` filed by @anuraaga: #2000

Perhaps we should not log `ResponseTimeoutException` and just notify via `RequestLog.responseCause()`, i.e. Put the detailed exception message into `ResponseTimeoutException` by making `ResponseTimeoutException` a non-singleton?

/cc @anuraaga

anuraaga:
If I'm not mistaken the ResponseTimeoutException is already returned to the user's future so agree it's great if we don't also log it here, or lower to debug level.

Making the exception non-singleton (but still no stack trace which isn't needed) seems good to me.

anuraaga:
Also maybe solving this is enough even if it's singleton?

https://github.com/line/armeria/issues/2001

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -286,8 +286,15 @@ private void close(@Nullable Throwable cause,
             if (cancelTimeout()) {
                 actionOnTimeoutCancelled.accept(cause);
             } else {
-                if (cause != null && !Exceptions.isExpected(cause)) {
-                    logger.warn("Unexpected exception:", cause);
+                if (cause != null && logger.isWarnEnabled() && !Exceptions.isExpected(cause)) {
+                    StringBuilder logMsg = new StringBuilder("Unexpected exception while closing");
```
ikhoon: nit: `final`?

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
mauhiz: oh, is there a risk that `request.authority()` is null?

mauhiz:
Thanks for all the valuable feedback.

@trustin and @anuraaga please let me know whether this PR should be closed (if you have other plans)
If not, I'll address the comments

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
trustin: I don't think so. Authority is nullable only for user convenience when sending a client request (so that Armeria auto-fills it), i.e. it is non-null on the server side. To make IntelliJ IDEA inspection happy, you might want to add an assertion though.

trustin:
@mauhiz What do you think about:

- Modifying ResponseTimeoutException.get() so that the caller can specify an exception message and
- Moving your error message generation code to where ResponseTimeoutException.get() is called

..?

If you think this is too much for you, please feel free to let us know so we can work on it for you.

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
trustin: Never mind. This is client-side and thus authority can be null. Sorry! :sweat_smile:

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
mauhiz: 👍

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
mauhiz: I don't think so, there is going to be a newline right after anyways (yaml log? :D)

mauhiz:
@trustin while this PR declares it wants to "explain Response timeout", I don't think this is the only possible exception here; so any changes to `ResponseTimeoutException` only would fail to address the whole issue (which is a confusing message)

trustin:
@mauhiz I see what you mean. I think we can merge this in and resolve #2000 separately.

trustin:
Thanks, @mauhiz !

mauhiz:
Thank you for the reviews!

PR #2038, opened 2019-09-02
Uses RequestContext instead of RequestLog for trace parsing and…

adriancole:
This change renovates http tracing instrumentation, to use `RequestContext` instead of `RequestLog` for trace parsing and sampling. It also avoids redundant timestamps and provides hooks needed for [secondary sampling](https://github.com/openzipkin-contrib/zipkin-secondary-sampling/tree/master/brave). 

## Switch from `RequestLog` to `RequestContext`
Brave normally adds `span.remoteEndpoint`, and parses tags without custom logic. However, you can add armeria-specific tags by implementing `HttpClientParser` or `HttpServerParser`, and passing that to configuration (ex a `@Bean` method in Spring Boot):

```java
public <T> void response(HttpAdapter<?, T> adapter, T res, Throwable error, SpanCustomizer span) {
    super.response(adapter, res, error, span); // do portable parsing
    if (res instanceof RequestLog) { // do armeria-specific
        final SocketAddress raddr = ((RequestLog) res).context().remoteAddress();
        if (raddr != null) {
             customizer.tag("my-special-remote-tag", raddr.toString());
        }  
    }
}
```

This change moves from `RequestLog` to `RequestContext`, whic is a more direct path to custom data.

```diff
<     if (res instanceof RequestLog) { // do armeria-specific
<         final SocketAddress raddr = ((RequestLog) res).context().remoteAddress();
---
>     if (res instanceof RequestContext) { // do armeria-specific
>         final SocketAddress raddr = ((RequestContext) res).remoteAddress();
```

## Use `ServiceRequestContext` when extracting incoming trace state
Before, tracing (Brave) handlers used the `HttpHeaders` object when parsing incoming trace IDs. This is fine for simple use cases. However, there are advanced patterns, such as [secondary sampling](https://github.com/openzipkin-contrib/zipkin-secondary-sampling/tree/master/brave) which need to read the http path or url to trigger trace recording decisions in a fashion very similar to feature flags. To allow this use case, we now use `ServiceRequestContext` instead. End user code has no impact to this decision.

## Less copy/paste
Before, Brave's http instrumentation didn't provide hooks to supply external timestamps. To re-use Armeria's request log timestamps meant a lot of copy/pasting of logic. Brave was specifically changed to address this use case in 5.7, which both avoids copy/pasting and also the implicit redundant timestamp commands. Thanks @kojilin for the original work which led to the upstream change affording this.

See https://github.com/openzipkin/brave/pull/986


adriancole:
don't merge or consider this too much yet. I am thinking about a way to simplify everything and better to do in one change.

trustin:
OK. Let me know when it's ready!

adriancole:
https://github.com/openzipkin/brave/pull/986 once merged will refine this to use wrapper types instead of log object directly

adriancole:
Now dependent on https://github.com/openzipkin/brave/pull/989 as I forgot we were doing copy/pasta for adjusting the finish timestamp

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceIntegrationTest.java
```diff
@@ -106,8 +108,15 @@ public void supportsPortableCustomization() throws Exception {
             public <T> void response(HttpAdapter<?, T> adapter, T res, Throwable error,
                                      SpanCustomizer customizer) {
                 super.response(adapter, res, error, customizer);
+                // TODO: is there a way to get the URL visible earlier? Waiting until response is
```
adriancole: I probably missed the review comment, but it seems in the current implementation, the URL isn't available until response parsing. this will subvert the intent of most parsers and such. Since we are post-dating the span's timestamp anyway, I wonder if we can delay until the URL is available, but not so late as when the request is already sent. then, we can parse at that moment...

wdyt? @kojilin

adriancole: ex if this were organized differently, or if internals were shifted such that the URL is available earlier, the code simplifies to `adapter.url(request)`

trustin: Not sure if this is what you're looking for, but how about adding a listener to `RequestLog` so that you are called back when request headers are ready? There's no need to wait until the response.

```java
ctx.log().addListener(log -> {
    generateUrl(log);
}, RequestLogAvailability.REQUEST_HEADERS);
```

adriancole: thanks will try

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
```diff
@@ -54,32 +55,33 @@ private ArmeriaHttpClientParser() {
     @Override
     public <T> void response(HttpAdapter<?, T> rawAdapter, T res, Throwable error, SpanCustomizer customizer) {
         super.response(rawAdapter, res, error, customizer);
-        if (res instanceof RequestLog && rawAdapter instanceof ArmeriaHttpClientAdapter) {
-            final RequestLog requestLog = (RequestLog) res;
-            final ArmeriaHttpClientAdapter adapter = (ArmeriaHttpClientAdapter) rawAdapter;
-            customizer.tag(SpanTags.TAG_HTTP_HOST, adapter.authority(requestLog))
-                      .tag(SpanTags.TAG_HTTP_URL, adapter.url(requestLog))
-                      .tag(SpanTags.TAG_HTTP_PROTOCOL, adapter.protocol(requestLog));
+        if (!(res instanceof HttpClientResponse)) return;
```
adriancole: @anuraaga fyi I have an idea of how to hide this type completely, so that parsers get the "unwrapped" one. will raise a PR in a bit on brave...

adriancole: https://github.com/openzipkin/brave/pull/990

adriancole:
sorry I forgot about this. will pickup monday

adriancole:
Sitting this down until I get feedback on something. Basically ideal case is we can get a service decorator that works on the abstraction of a resolved URL. If not, it is crazy. Anything besides decorator is also ok as long as we can wrap execution in a trace scope (remember we permit current trace context to be served from something besides armeria)

anuraaga:
I think we may have run into a problem since we consolidated on `RequestLog`, which is a "weaker" type than others we have. Information about a request is currently split between `HttpRequest`, `RequestLog`, and `RequestContext`. I don't think this is a bad design, it can make some cases clearer, while other cases maybe less clear. But `RequestContext` conveniently provides accessors to the `HttpRequest` and `RequestLog` too so is the most powerful type - we should probably be wrapping that instead of `RequestLog`. You'll have the headers and can also listen on the log events as needed.

adriancole:
wrapping `RequestContext` is probably a better choice than, agreed. I think probably the best thing I can do here is to not solve the http url parsing thing, and leave that for a different issue, as it is a different concern vs the wrapping one and also much more subtle.

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceIntegrationTest.java
adriancole: https://github.com/line/armeria/issues/2089

adriancole:
PTAL I think this is ready to go

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -132,6 +120,7 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
             }
         }
 
+        // TODO: see if it makes sense to defer all parsing until here
```
adriancole: Remove TODO already tracked #2089
```suggestion
```

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
```diff
@@ -103,6 +88,8 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
                 return delegate().serve(ctx, req);
             }
         }
+
+        // TODO: see if it makes sense to defer all parsing until here
```
adriancole: Remove TODO already tracked #2089

```suggestion
```

adriancole:
PS the takeaway on this change is that you can now cast into RequestContext (Client or Service) when parsing, and casting the URL anomaly aside, path-based secondary sampling should work now. 

We can put a test here later to prove it https://github.com/openzipkin-contrib/zipkin-secondary-sampling

trustin:
@adriancole There are some Checkstyle violations that need fixes: https://travis-ci.com/line/armeria/builds/128273879#L1538

adriancole:
sorry. I'm still not in love with gradle. It let's you succeed like this!
I'll run the actual command..

± |single-arity → origin ?:5 ✗| → ./gradlew checkStyle
Configuration on demand is an incubating feature.

BUILD SUCCESSFUL in 17s
248 actionable tasks: 248 up-to-date

trustin:
I guess `checkstyle` has some bugs with incremental build. How about running `./gradlew clean checkstyle`?

adriancole:
Fixed what I could. Literally no idea why this one is still broke.. Odd to
be required to javadoc summaries on package private types.. we could
consider relaxing that vs requiring "this does what it says" statements and
resultant checkStyle busywork.

    [ant:checkstyle] [ERROR]
/Users/acole/oss/armeria/brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java:38:
Summary javadoc is missing. [SummaryJavadoc]

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -0,0 +1,179 @@
...
+        public int statusCode() {
+            if (!ctx.log().isAvailable(RequestLogAvailability.RESPONSE_HEADERS)) {
+                return 0;
+            }
+            return ctx.log().status().code();
+        }
+
+        @Override
+        public long finishTimestamp() {
+            return SpanContextUtil.wallTimeMicros(ctx.log(), ctx.log().responseEndTimeNanos());
```
minwoox: Do we need a check here?
```java
if (!ctx.log().isAvailable(RequestLogAvailability.RESPONSE_END)) {
    return 0L;
}
```

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -0,0 +1,185 @@
...
+
+import brave.http.HttpClientAdapter;
+
+/**
+ * Wraps a pair of {@link ClientRequestContext} and {@link RequestHeadersBuilder} in an {@link
+ * brave.http.HttpClientRequest}.
+ */
+final class ClientRequestContextAdapter {
+    /**
+     * @param ctx used to parse http properties
```
trustin: Here, you need to insert some summary, like `Returns a new {@link brave.http.HttpClientRequest} from the given {@link ClientRequestContext} and {@link RequestHeadersBuilder}.

brave/src/main/java/com/linecorp/armeria/server/brave/ServiceRequestContextAdapter.java
```diff
@@ -0,0 +1,210 @@
...
+                    return paths.get(paths.size() - 1);
+                case REGEX_WITH_PREFIX:
+                    return paths.get(1) + paths.get(0);
+            }
+            return null;
+        }
+
+        @Override
+        public long finishTimestamp() {
+            return SpanContextUtil.wallTimeMicros(ctx.log(), ctx.log().responseEndTimeNanos());
```
minwoox: ditto

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -144,38 +132,13 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
             if (log.isAvailable(RequestLogAvailability.RESPONSE_FIRST_BYTES_TRANSFERRED)) {
                 SpanTags.logWireReceive(span, log.responseFirstBytesTransferredTimeNanos(), log);
             }
-            SpanTags.updateRemoteEndpoint(span, log);
-            handleFinish(log, span);
+            SpanTags.updateRemoteEndpoint(span, ctx);
+            HttpClientResponse response = ClientRequestContextAdapter.asHttpClientResponse(ctx);
```
minwoox: nit: final

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
```diff
@@ -112,37 +98,12 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
             if (log.isAvailable(RequestLogAvailability.RESPONSE_FIRST_BYTES_TRANSFERRED)) {
                 SpanTags.logWireSend(span, log.responseFirstBytesTransferredTimeNanos(), log);
             }
-            handleFinish(log, span);
+            HttpServerResponse response = ServiceRequestContextAdapter.asHttpServerResponse(ctx);
```
minwoox: nit: final

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
adriancole: that's studdering. I'll just delete it. thanks

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
adriancole: note for future... this all is terribly crufty. it is sad that these operations cannot return absent values

minwoox: 😭

adriancole: thanks regardless. I'd hate for some accidental usage of this to crash a request. In hind sight, probably this method should have been protected.

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
adriancole: thanks for saving me.. the checkstyle being slightly under the type made me scratch head. I'm more used to things that align to the actual line of code.

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
adriancole: https://github.com/line/armeria/issues/2090 on my complaint :) thanks for being watchdog meanwhile. If this is solved, you and @kojilin etc can be less scared of api usage as it doesn't need to be so scary anyway

adriancole:
Thanks @trustin and @minwoox! please don't mistake my grumbling for lack of gratitude. Appreciate both your help and also the long term stewardship on this project.

minwoox:
@adriancole haha, no worries. 😄 We really appreciate your support, interest, enthusiasm, and grumbling(?).

trustin:
Need to rebase once https://github.com/line/armeria/pull/2091 is merged (and vice versa).

adriancole:
> Need to rebase once #2091 is merged (and vice versa).

happy to rebase

trustin:
Let me merge this first, because #2091 may take longer. Thanks, @adriancole !

trustin:
@adriancole Would you mind updating the PR subject and description to reflect the final outcome of this PR?

adriancole:
@trustin done. feel free to edit further as you see fit. thanks!

trustin:
@adriancole Thanks a lot for highly detailed description!

PR #2051, opened 2019-09-06
Update Awaitility to 4.0.1

minwoox:
As described in https://github.com/awaitility/awaitility/issues/140


trustin:
Thanks, @minwoox !

PR #2052, opened 2019-09-08
Support java9 api in RequestContextAwareCompletableFuture

### LINKED ISSUE #1991: Override new  Java 9 `CompletableFuture` API for  `RequestContextAwareCompletableFuture`
Armeria `RequestContextAwareCompletableFuture` extends `CompletableFuture` and override their methods based on Java 8.
https://github.com/line/armeria/blob/54e124d73178609b4f0af6d833ea562a1ce15bd9/core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java#L27
Java 9 comes with [some changes](https://www.baeldung.com/java-9-completablefuture) to the CompletableFuture class.
We need to override the following methods:
* [completeAsync​(Supplier<? extends T> supplier)
](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/CompletableFuture.html#completeAsync-java.util.function.Supplier-)
* [completeAsync​(Supplier<? extends T> supplier, Executor executor)](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/CompletableFuture.html#completeAsync-java.util.function.Supplier-java.util.concurrent.Executor-)

NB: To keep compatibility with Java 8
* Don't use `@Override` keyword
* Branch policy is required depending on Java version.

Additionally, We can detect un-override cases early by scanning `javadoc` and find a class which `implements` or `extends`, but is not `com.linecorp.armeria` package. 
Related work: https://github.com/line/armeria/blob/0419e5b24ef5a12cfa3e4937b08fcb0f24f501d3/site/build.gradle#L72-L77

cc: Thanks for inquiry on this issue. @jongyeol

young891221:
Motivation:
- Armeria `RequestContextAwareCompletableFuture` extends `CompletableFuture` and override their methods based on Java 8.

Modifications:
- Implement java9 completeAsync API

Result:
- completeAsync API support in java9 or later
- closed #1991


core/src/main/java/com/linecorp/armeria/common/util/Functions.java
```diff
@@ -194,5 +196,25 @@
...
+     * Converts the specified {@link Supplier} into a {@link Callable}
+     */
+    public static <T> Callable<T> getToCall(Supplier<T> supplier) {
+        return supplier::get;
+    }
+
+    /**
+     * Converts the specified {@link Callable} into a {@link Supplier}
+     */
+    public static <T> Supplier<T> callToGet(Callable<T> callable) {
```
trustin: How about 'fromCallable()'?

core/src/main/java/com/linecorp/armeria/common/util/Functions.java
```diff
@@ -194,5 +196,25 @@
...
+        return () -> {
+            try {
+                return callable.call();
+            } catch(Exception e) {
```
trustin: Could you use Throwables.throwIfUnchecked() first so that unchecked exceptions are thrown as is?

young891221: Yes, I can.
`Throwables.throwIfUnchecked ()` handles RuntimeException and Error but not all exceptions. But, this solution needs to manage post logic.
If `callable.call()` method throws an exception, I think that method should throw an exception in catch block. What do you think?

```java
} catch (Throwable e) {
      Throwables.throwIfUnchecked(e);
      throw new RuntimeException("fromCallable call exception", e);
}
```

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +253,96 @@
...
+                (Functions.fromCallable(ctx.makeContextAware(callable)), executor));
+    }
+
+    public CompletableFuture<T> orTimeout(long timeout, TimeUnit unit) {
+        if (unit == null)
+            throw new NullPointerException();
+        if (!this.isDone())
+            whenComplete(new Canceller(Delayer.delay(new Timeout(this),
+                    timeout, unit)));
+        return ctx.makeContextAware(this);
```
trustin: We could just return `this`, no?

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +253,96 @@
...
+    }
+
+    public CompletableFuture<T> completeOnTimeout(T value, long timeout,
+                                                  TimeUnit unit) {
+        if (unit == null)
+            throw new NullPointerException();
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +253,96 @@
...
+        if (!this.isDone())
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +253,96 @@
...
+            whenComplete(new Canceller(Delayer.delay(
+                    new DelayedCompleter<>(this, value),
+                    timeout, unit)));
+        return ctx.makeContextAware(this);
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +253,96 @@
...
+        static ScheduledFuture<?> delay(Runnable command, long delay,
+                                        TimeUnit unit) {
+            return delayer.schedule(command, delay, unit);
+        }
+
+        static final class DaemonThreadFactory implements ThreadFactory {
+            public Thread newThread(Runnable r) {
+                Thread t = new Thread(r);
+                t.setDaemon(true);
+                t.setName("CompletableFutureDelayScheduler");
```
trustin: How about prepending `armeria-`, not to be confused with the thread created by JDK?

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +253,96 @@
...
+        final CompletableFuture<U> f;
+        final U u;
+        DelayedCompleter(CompletableFuture<U> f, U u) { this.f = f; this.u = u; }
+        public void run() {
+            if (f != null)
+                f.complete(u);
+        }
+    }
+
+    static final class Canceller implements BiConsumer<Object, Throwable> {
```
trustin: Could implement `BiFunction` and return `null` to be used with `handle()`.

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +307,31 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+        final RequestContext context = createContext();
+        final CompletableFuture<String> originalFuture = new CompletableFuture<>();
+        final CompletableFuture<String> contextAwareFuture = context.makeContextAware(originalFuture);
+        final CompletableFuture<String> resultFuture = contextAwareFuture.completeAsync(() -> "success", executor);
+
+        originalFuture.complete("success");
+        assertDepth(1);
+        assertThat(resultFuture.get()).isEqualTo("success");
+    }
+
```
trustin: Please add the tests for timeout operations.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
young891221: Thank you for the good information 👍

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +253,96 @@
...
+        final Callable<? extends T> callable = Functions.toCallable(supplier);
+        return (CompletableFuture<T>) ctx.makeContextAware(supplyAsync
+                (Functions.fromCallable(ctx.makeContextAware(callable)), executor));
+    }
+
+    public CompletableFuture<T> orTimeout(long timeout, TimeUnit unit) {
+        if (unit == null)
+            throw new NullPointerException();
+        if (!this.isDone())
+            whenComplete(new Canceller(Delayer.delay(new Timeout(this),
```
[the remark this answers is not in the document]
young891221: Thank you for the good information 👍

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +255,111 @@
...
         return ctx.makeContextAware(super.exceptionally(ctx.makeContextAware(fn)));
     }
+
+    // support JDK9 functions
+
+    public <U> CompletableFuture<U> newIncompleteFuture() {
+        return new CompletableFuture<>();
+    }
+
+    public CompletableFuture<T> copy() {
```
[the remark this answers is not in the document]
trustin: Agreed. We do not need to override this method if we implemented `newIncompleteFuture()` correctly.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +255,111 @@
...
+    }
+
+    public CompletionStage<T> minimalCompletionStage() {
+        return this;
+    }
+
+    @SuppressWarnings("unchecked")
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier) {
+        final Callable<? extends T> callable = Functions.toCallable(supplier);
+        return (CompletableFuture<T>) ctx.makeContextAware(supplyAsync(
```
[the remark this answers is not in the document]
trustin: Adding `RequestContext.makeContextAware(Supplier)` introduces ambiguity against `makeContextAware(Callable)`, requiring the lambda expression to have the explicit type. +1 for having a private method.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +255,111 @@
...
+
+    @SuppressWarnings("unchecked")
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier,
+                                              Executor executor) {
+        final Callable<? extends T> callable = Functions.toCallable(supplier);
+        return (CompletableFuture<T>) ctx.makeContextAware(supplyAsync
+                (Functions.fromCallable(ctx.makeContextAware(callable)), executor));
+    }
+
+    public CompletableFuture<T> orTimeout(long timeout, TimeUnit unit) {
```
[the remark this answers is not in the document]
trustin: That's a good point. Yeah, `CompletableFuture.orTimeout()` and `completeOnTimeout()` relies on `whenComplete()` we overrode already, so there's no need to override `orTimeout()` and `completeOnTimeout()`. @young891221, sorry, but could you remove the two methods back?

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +255,111 @@
     public CompletableFuture<T> exceptionally(Function<Throwable, ? extends T> fn) {
         return ctx.makeContextAware(super.exceptionally(ctx.makeContextAware(fn)));
     }
+
+    // support JDK9 functions
+
+    public <U> CompletableFuture<U> newIncompleteFuture() {
```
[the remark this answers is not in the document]
young891221: I had the same idea during the holidays. Thank you for reply!

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +243,92 @@
...
+                return action.get();
+            }
+        };
+    }
+
+    private MinimalStage<T> uniAsMinimalStage() {
+        return new MinimalStage<>();
+    }
+
+    static final class MinimalStage<T> extends CompletableFuture<T> {
```
anuraaga: This isn't context aware since it doesn't extend `RequestContextAwareCompletableFuture` or return one. How about

```java
final class MinimalStage<T> extends RequestContextAwareCompletableFuture<T> {
  MinimalStage() {
    super(ctx);
    RequestContextAwareCompletableFuture.this.whenComplete((value, t) -> {
      if (t != null) {
        super.completeExceptionally(t);
      } else {
        super.complete(value);
   }
}
```

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +243,92 @@
...
+                (Supplier<? extends T> supplier) {
+            throw new UnsupportedOperationException(); }
+        public CompletableFuture<T> orTimeout
+                (long timeout, TimeUnit unit) {
+            throw new UnsupportedOperationException(); }
+        public CompletableFuture<T> completeOnTimeout
+                (T value, long timeout, TimeUnit unit) {
+            throw new UnsupportedOperationException(); }
+        @Override public CompletableFuture<T> toCompletableFuture() {
+            if (super.isDone()) {
```
anuraaga: I think this this can be context aware by doing

```java
CompletableFuture<T> future = newIncompleteFuture();
whenComplete((value, t) -> {
  if (t != null) {
    future.completeExceptionally(t);
 } else {
   future.complete(value);
  }
  return future;
}
```

I guess us propagating whenComplete into a new future is similar to the JDK's private methods using `uniPush`.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
anuraaga: Also given how complicated this API is, I think we can just `throw new UnsupportedOperationException` for now in `minimalCompletionStage()` - we could revisit after copying in the `CompletableFuture` tck so it's easier to verify the behavior works as the JDK expects, and in the meantime I think it's a very uncommon API.

trustin: I guess it's not allowed to fork the GPL'd TCK into our code base. I guess we can implement the `MinimalStage` by delegating all supported operations to `RequestContextAwareCompletableFuture` and wrapping the returned `CompletableFuture` with `MinimalStage`:

```java
final class MinimalStage<T> implements CompletionStage<T> {
    private final RequestContextAwareCompletableFuture<T> delegate;

    public CompletionStage<Void> thenRun(...) {
        return new MinimalStage<>(delegate.thenRun(...));
    }
    public boolean complete(...) {
        throw new UnsupportedOperationException();
    }
    public CompletableFuture toCompletableFuture() {
        return delegate.thenApply(Function.identity());
    }
}
```

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -251,7 +254,7 @@
     }
 
     public CompletionStage<T> minimalCompletionStage() {
-        return this.uniAsMinimalStage();
+        return ctx.makeContextAware(new RequestContextAwareMinimalStage<>(this));
```
trustin: I guess we do not need to make `RequestContextAwareMinimalStage` context-aware because `this` is already context-aware.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareMinimalStage.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    @Override
+    public CompletionStage<T> exceptionally(Function<Throwable, ? extends T> fn) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public CompletableFuture<T> toCompletableFuture() {
+        return delegate;
```
trustin: Calling `complete()` or `completeExceptionally()` on the returned future should not complete the minimal stage (`this`). How about:

```java
return delegate.thenApply(Function.identity());
```

.. so that `delegate` is not completed when a user does this:

```java
minimalStage.toCompletableFuture().complete(null);
```

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareMinimalStage.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private final RequestContextAwareCompletableFuture<T> delegate;
+
+    RequestContextAwareMinimalStage(RequestContextAwareCompletableFuture<T> delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public <U> CompletionStage<U> thenApply(Function<? super T, ? extends U> fn) {
+        throw new UnsupportedOperationException();
```
trustin: I think you have to implement all the methods that throw an `UnsupportedOperationException`:

```java
return new RequestContextAwareMinimalStage(
        (RequestContextAwareCompletableFuture<U>) delegate.thenApply(fn));
```

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +246,31 @@
     public CompletableFuture<T> exceptionally(Function<Throwable, ? extends T> fn) {
         return ctx.makeContextAware(super.exceptionally(ctx.makeContextAware(fn)));
     }
+
+    // support JDK9 functions
+
+    public <U> CompletableFuture<U> newIncompleteFuture() {
+        return new CompletableFuture<>();
```
trustin: I think you have to return `new RequestContextAwareCompletableFuture<>(ctx);` here, like @anuraaga mentioned in https://github.com/line/armeria/pull/2052/files#r323632680

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +243,92 @@
...
+    public <U> CompletableFuture<U> newIncompleteFuture() {
+        return new RequestContextAwareCompletableFuture<>(ctx);
+    }
+
+    public CompletionStage<T> minimalCompletionStage() {
+        return this.uniAsMinimalStage();
+    }
+
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier) {
+        return ctx.makeContextAware(supplyAsync(this.makeContextAware(supplier)));
```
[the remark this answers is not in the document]
trustin: As @anuraaga said, this method must return `this`, and `this` must be completed with the result of the specified `supplier`, i.e.

```java
supplyAsync(this.makeContextAware(supplier)).handle((res, cause) -> {
    if (cause != null) {
        completeExceptionally(cause);
    } else {
        complete(res);
    }
    return null;
});
return this;
```

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -243,4 +246,31 @@
...
+        return new RequestContextAwareMinimalStage<>(this);
+    }
+
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier) {
+        return ctx.makeContextAware(supplyAsync(this.makeContextAware(supplier)));
+    }
+
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier,
+                                              Executor executor) {
+        return ctx.makeContextAware(supplyAsync(this.makeContextAware(supplier), executor));
```
trustin: Ditto - See my comment above.

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +304,66 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
     }
 
+    @Test
+    public void minimalCompletionStageUsingCompleteAsync() throws Exception {
+        final RequestContext context = createContext(false);
+        final CompletableFuture<Integer> originalFuture = new CompletableFuture<>();
+        final RequestContextAwareCompletableFuture<Integer> contextAwareFuture = (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
+        final CompletionStage<Integer> completionStage = contextAwareFuture.minimalCompletionStage();
+
+        contextAwareFuture.completeAsync(() -> { throw new RuntimeException(); });
```
trustin: Perhaps `{` and `}` are unnecessary?

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +304,66 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
         }
     }
 
+    @Test
+    public void minimalCompletionStageUsingCompleteAsync() throws Exception {
+        final RequestContext context = createContext(false);
+        final CompletableFuture<Integer> originalFuture = new CompletableFuture<>();
+        final RequestContextAwareCompletableFuture<Integer> contextAwareFuture = (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
+        final CompletionStage<Integer> completionStage = contextAwareFuture.minimalCompletionStage();
```
trustin: `minimalStage`?

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +304,66 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+    @Test
+    public void minimalCompletionStageUsingWhenComplete() throws Exception {
+        final RequestContext context = createContext(false);
+        final CompletableFuture<Integer> originalFuture = new CompletableFuture<>();
+        final RequestContextAwareCompletableFuture<Integer> contextAwareFuture = (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
+        final CompletionStage<Integer> completionStage = contextAwareFuture.minimalCompletionStage();
+
+        AtomicInteger atomicInteger = new AtomicInteger(0);
+        AtomicReference<Throwable> atomicReference = new AtomicReference<>();
+        completionStage.whenComplete((v, e) -> {if (e != null) atomicReference.set(e); else atomicInteger.set(v);});
```
trustin: Could you format properly by inserting newlines and indentations?

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +304,66 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+    }
+
+    @Test
+    public void minimalCompletionStageUsingWhenComplete() throws Exception {
+        final RequestContext context = createContext(false);
+        final CompletableFuture<Integer> originalFuture = new CompletableFuture<>();
+        final RequestContextAwareCompletableFuture<Integer> contextAwareFuture = (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
+        final CompletionStage<Integer> completionStage = contextAwareFuture.minimalCompletionStage();
+
+        AtomicInteger atomicInteger = new AtomicInteger(0);
```
trustin: You don't need to specify `0`. Default constructor does the same job.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
young891221: `new RequestContextAwareCompletableFuture<>(ctx);` makes stack overflow on java9.
Because, several methods of CompletableFuture call `newIncompleteFuture()` on their own.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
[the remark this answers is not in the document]
anuraaga: Is there something wrong with my suggestion of `super.completeAsync`? ;)

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
anuraaga: Could you post the stack overflow?

Both `RequestContextAwareCompletableFuture` and `CompletableFuture` constructor have no method calls, just assign fields, so I wouldn't expect that.

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
young891221: `throw` needs `{` and `}`.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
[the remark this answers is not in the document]
trustin: `completeAsync` is unavailable in Java 8, no?

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
young891221: 스크린샷 2019-09-20 오후 6 22 31
스크린샷 2019-09-20 오후 6 23 07

young891221: It happens because methods call the` newIncompleteFuture()` method from inside Java9.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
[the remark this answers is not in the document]
anuraaga: `RequestContextAwareCompletableFuture` is private, so I don't think a user could access this method on Java 8 without reflection, which I guess is ok if it doesn't work as that would be a very weird use case.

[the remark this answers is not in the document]
anuraaga: Oh - `--release` :(

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
trustin: Then please insert some like breaks.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
anuraaga: Ah interesting - this is because for Java 8 we have to make the result of e.g., `super.handle` context aware, but don't need to on Java 9+ which added the `newIncompleteFuture` method exactly for our use case. Thanks for clarifying for me!

@trustin Does this warrant creating a separate class for Java 9+? This change has turned out to be very intense...

anuraaga: Or instead of a separate class, for now we could also do `instanceof RequestContextAwareFuture` in all the methods we make the result of a method context aware which should be compatible on both versions.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
young891221: I see

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
young891221: Got it.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
young891221: When handling with `instanceof RequestContextAwareFuture`, this is a recognition that is handled differently by java version. First of all, I have to check `this`, so I wonder if this is actually possible. At this point, I think there is no need to implement `newIncompleteFuture()`. What do you think?

anuraaga: We have somewhat similar code that runs different logic based on Java version - I guess it's inevitable when supporting new APIs

https://github.com/line/armeria/pull/2055

Here luckily the logic doesn't need to use reflection I think.

I'm not sure about `this`. Does this pattern not fix the stack overflow?

```java
final CompletableFuture<T> result = super.method(ctx.makeContextAware(param));
if (result instanceof Requestcontextawarecompletablefuture) 
  // Already context aware
  return result;
else
  return ctx.makeContextAware(result);

I think this is better than not implementing newIncompleteFuture since we don't know when a new JDK API will try to use it.

Comment (young891221) [reply core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java]: Very nice! Thanks for your support 👍

File: core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java [hunk 27]
```diff
@@ -34,213 +37,405 @@
 
     @Override
     public <U> CompletableFuture<U> thenApply(Function<? super T,? extends U> fn) {
-        return ctx.makeContextAware(super.thenApply(ctx.makeContextAware(fn)));
+        final CompletableFuture<U> result = super.thenApply(ctx.makeContextAware(fn));
+        if (result instanceof RequestContextAwareCompletableFuture) {
```
anuraaga: Let's go ahead and extract a private method `maybeMakeContextAware`. We can add a comment like ```If already an instance of Requestcontextawarecompletablefuture, it must either be `this` or the result of `newIncompleteFuture` so we don't have to do anything.```

young891221: Got it. I missed refactoring point.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,271 @@
...
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier) {
+        supplyAsync(this.makeContextAware(supplier)).handle((result, cause) -> {
+            if (cause != null) {
+                completeExceptionally(cause);
+            } else {
+                complete(result);
+            }
+            return null;
+        });
+        return maybeMakeContextAware(this);
```
trustin: `return this;` should be suffice.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,271 @@
...
+                                              Executor executor) {
+        supplyAsync(this.makeContextAware(supplier), executor).handle((result, cause) -> {
+            if (cause != null) {
+                completeExceptionally(cause);
+            } else {
+                complete(result);
+            }
+            return null;
+        });
+        return maybeMakeContextAware(this);
```
trustin: `return this;` should be suffice.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,271 @@
...
+    }
+
+    private <T> CompletableFuture<T> maybeMakeContextAware(CompletableFuture<T> result) {
```
trustin: `result` -> `future` ?

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,271 @@
...
 
     @Override
     public <U> CompletableFuture<U> applyToEitherAsync(CompletionStage<? extends T> other,
                                                        Function<? super T, U> fn) {
-        return ctx.makeContextAware(super.applyToEitherAsync(other, ctx.makeContextAware(fn)));
+        final CompletableFuture<U> result = super.applyToEitherAsync(other, ctx.makeContextAware(fn));
+        if (result instanceof RequestContextAwareCompletableFuture) {
+            return result;
+        }
+        return ctx.makeContextAware(result);
```
trustin: Could just use `maybeMakeContextAware()`?

```java
return super.applyToEitherAsync(...);
```

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareMinimalStage.java
```diff
@@ -0,0 +1,287 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Executor;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+@SuppressWarnings("unchecked")
+public class RequestContextAwareMinimalStage<T> implements CompletionStage<T> {
```
trustin: Could remove `public` and add `final`.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareMinimalStage.java
```diff
@@ -0,0 +1,287 @@
...
+    public CompletionStage<T> exceptionally(Function<Throwable, ? extends T> fn) {
+        return new RequestContextAwareMinimalStage(
+                (RequestContextAwareCompletableFuture<T>) delegate.exceptionally(fn));
+    }
+
+    @Override
+    public CompletableFuture<T> toCompletableFuture() {
+        return delegate.thenApply(Function.identity());
+    }
+}
```
trustin: How about implementing `toString()` using `MoreObjects.toStringHelper()`?

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+
+        contextAwareFuture.completeAsync(() -> {
+            throw new RuntimeException();
+        });
+        contextAwareFuture.complete(1);
+
+        assertThat(contextAwareFuture.join()).isEqualTo(1);
+        assertThat(contextAwareFuture.getNow(null)).isEqualTo(1);
+        assertThat(contextAwareFuture.get()).isEqualTo(1);
+        completionStage.toCompletableFuture().get();
```
trustin: Please add assertion for the result of this sentence.

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+    @Test
+    public void minimalCompletionStageUsingWhenComplete() throws Exception {
+        final RequestContext context = createContext(false);
+        final CompletableFuture<Integer> originalFuture = new CompletableFuture<>();
+        final RequestContextAwareCompletableFuture<Integer> contextAwareFuture =
+                (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
+        final CompletionStage<Integer> completionStage = contextAwareFuture.minimalCompletionStage();
+
+        AtomicInteger atomicInteger = new AtomicInteger();
+        AtomicReference<Throwable> atomicReference = new AtomicReference<>();
```
trustin: `causeCaptor`?

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+        });
+        contextAwareFuture.complete(1);
+
+        assertThat(contextAwareFuture.join()).isEqualTo(1);
+        assertThat(contextAwareFuture.getNow(null)).isEqualTo(1);
+        assertThat(contextAwareFuture.get()).isEqualTo(1);
+        completionStage.toCompletableFuture().get();
+    }
+
+    @SuppressWarnings("unchecked")
```
trustin: You do not need to add this annotation at the method level but limit the scope of suppression at variable level:

```java
final RequestContextAwareCompletableFuture<Integer> contextAwareFuture =
        (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
```

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,271 @@
...
+        return maybeMakeContextAware(super.handleAsync(ctx.makeContextAware(fn), executor));
     }
 
     @Override
     public CompletableFuture<T> exceptionally(Function<Throwable, ? extends T> fn) {
-        return ctx.makeContextAware(super.exceptionally(ctx.makeContextAware(fn)));
+        return maybeMakeContextAware(super.exceptionally(ctx.makeContextAware(fn)));
+    }
+
+    // support JDK9 functions
```
trustin: `// The methods not available in Java 8 but only in Java 9+`

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,267 @@
...
-        return ctx.makeContextAware(super.whenCompleteAsync(ctx.makeContextAware(action)));
+        return maybeMakeContextAware(super.whenCompleteAsync(ctx.makeContextAware(action)));
     }
 
     @Override
     public CompletableFuture<T> whenCompleteAsync(BiConsumer<? super T, ? super Throwable> action,
                                                   Executor executor) {
-        return ctx.makeContextAware(super.whenCompleteAsync(ctx.makeContextAware(action), executor));
+        return maybeMakeContextAware(super.whenCompleteAsync(ctx.makeContextAware(action),
+                                                                    executor));
```
trustin: nit: indentation?

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,266 @@
...
+                completeExceptionally(cause);
+            } else {
+                complete(result);
+            }
+            return null;
+        });
+        return this;
+    }
+
+    private <T> CompletableFuture<T> maybeMakeContextAware(CompletableFuture<T> future) {
```
minwoox: nit: type `T` is used as a result type for `CompletableFuture<T>`. We should use `<U>` instead.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,266 @@
...
+                complete(result);
+            }
+            return null;
+        });
+        return this;
+    }
+
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier,
+                                              Executor executor) {
+        supplyAsync(this.makeContextAware(supplier), executor).handle((result, cause) -> {
```
minwoox: We do not need `this` here.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
```diff
@@ -34,213 +37,266 @@
...
+    public <U> CompletableFuture<U> newIncompleteFuture() {
+        return new RequestContextAwareCompletableFuture<>(ctx);
+    }
+
+    public CompletionStage<T> minimalCompletionStage() {
+        return new RequestContextAwareMinimalStage<>(this);
+    }
+
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier) {
+        supplyAsync(this.makeContextAware(supplier)).handle((result, cause) -> {
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareMinimalStage.java
```diff
@@ -0,0 +1,296 @@
+/*
+ * Copyright 2017 LINE Corporation
```
minwoox: nit: 2019

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareMinimalStage.java
```diff
@@ -0,0 +1,296 @@
...
+
+    @Override
+    public CompletableFuture<T> toCompletableFuture() {
+        return delegate.thenApply(Function.identity());
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                .add("delegate", delegate)
```
minwoox: nit: indentation?
```java
                  .add("delegate", delegate)
                  .toString();
```

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+        assertThat(completionStage.toCompletableFuture().get()).isEqualTo(1);
+    }
+
+    @Test
+    public void minimalCompletionStageUsingWhenComplete() throws Exception {
+        final RequestContext context = createContext(false);
+        final CompletableFuture<Integer> originalFuture = new CompletableFuture<>();
+        @SuppressWarnings("unchecked")
+        final RequestContextAwareCompletableFuture<Integer> contextAwareFuture =
+                (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
```
minwoox: nit: `(RequestContextAwareCompletableFuture<Integer>)`

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+        final CompletionStage<Integer> completionStage = contextAwareFuture.minimalCompletionStage();
+
+        AtomicInteger atomicInteger = new AtomicInteger();
```
minwoox: nit: final

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+        AtomicReference<Throwable> causeCaptor = new AtomicReference<>();
```
minwoox: nit: final

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+        assertThat(causeCaptor.get()).isNull();
+    }
+
+    @Test
+    public void makeContextAwareCompletableFutureUsingCompleteAsync() throws Exception {
+        final RequestContext context = createContext(false);
+        final CompletableFuture<String> originalFuture = new CompletableFuture<>();
+        @SuppressWarnings("unchecked")
+        final RequestContextAwareCompletableFuture<String> contextAwareFuture =
+                (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
```
minwoox: nit: `(RequestContextAwareCompletableFuture<String>)`

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
         }
     }
 
+    @Test
+    public void minimalCompletionStageUsingCompleteAsync() throws Exception {
+        final RequestContext context = createContext(false);
+        final CompletableFuture<Integer> originalFuture = new CompletableFuture<>();
+        @SuppressWarnings("unchecked")
+        final RequestContextAwareCompletableFuture<Integer> contextAwareFuture =
+                (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
```
minwoox: ditto

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+    @Test
+    public void makeContextAwareCompletableFutureUsingCompleteAsyncWithExecutor() throws Exception {
+        final ExecutorService executor = Executors.newFixedThreadPool(2);
+        final RequestContext context = createContext(false);
+        final CompletableFuture<String> originalFuture = new CompletableFuture<>();
+        @SuppressWarnings("unchecked")
+        final RequestContextAwareCompletableFuture<String> contextAwareFuture =
+                (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
+        final CompletableFuture<String> resultFuture = contextAwareFuture.completeAsync(() -> "success",
+                executor);
```
minwoox: nit: indentation
```java
...completeAsync(() -> "success",
                 executor);
```

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+    }
+
+    @Test
+    public void makeContextAwareCompletableFutureUsingCompleteAsyncWithExecutor() throws Exception {
+        final ExecutorService executor = Executors.newFixedThreadPool(2);
+        final RequestContext context = createContext(false);
+        final CompletableFuture<String> originalFuture = new CompletableFuture<>();
+        @SuppressWarnings("unchecked")
+        final RequestContextAwareCompletableFuture<String> contextAwareFuture =
+                (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
```
minwoox: ditto

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -307,6 +308,83 @@ public void makeContextAwareCompletableFutureWithAsyncChaining() throws Exceptio
...
+        final CompletableFuture<Integer> originalFuture = new CompletableFuture<>();
+        @SuppressWarnings("unchecked")
+        final RequestContextAwareCompletableFuture<Integer> contextAwareFuture =
+                (RequestContextAwareCompletableFuture) context.makeContextAware(originalFuture);
+        final CompletionStage<Integer> completionStage = contextAwareFuture.minimalCompletionStage();
+
+        contextAwareFuture.completeAsync(() -> {
+            throw new RuntimeException();
+        });
+        contextAwareFuture.complete(1);
```
minwoox: What is this test for?
If we add `Thread.sleep(1000)` before this line the test result is totally different.
If we want to test `completeAsync` we need to remove this `contextAwareFuture.complete(1);`.
And we need to change all the `assertThat(...)` below.

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
young891221: Applied

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
young891221: Applied

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareCompletableFuture.java
young891221: Applied

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareMinimalStage.java
young891221: Applied

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
young891221: Applied

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
young891221: This part requires forced converting to be tested in Java8.

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
young891221: Thank you for the feedback.
It was a test on the `toCompletableFuture()` method call when converting to minimalStage. I deleted the unnecessary code and changed it to make sense.

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
minwoox: I mean you used `(RequestContextAwareCompletableFuture)` to downcast without the parameter type. If we add the type parameter here which is `<String>`, we no longer need `@SuppressWarnings("unchecked")` in line 358. 😄

core/src/main/java/com/linecorp/armeria/common/RequestContextAwareMinimalStage.java
```diff
@@ -0,0 +1,296 @@
...
+
+    @Override
+    public CompletableFuture<T> toCompletableFuture() {
+        return delegate.thenApply(Function.identity());
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                            .add("delegate", delegate)
```
minwoox: nit: `.` should be right under the `.` in the above line.
```java
MoreObjects.toStringHelper(this)
           .add("delegate", delegate)
```

Issue #1686, opened 2019-03-26
Use Java 9's stack-walking API to reduce the overhead of getting full stack trace in Thrift TEXT serialization format.


trustin:
In `StructContext`, we need to acquire the class name from the stack trace using `Thread.current().getStackTrace()`:

- https://github.com/line/armeria/blob/armeria-0.82.0/thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java#L104-L164

However, `getStackTrace()` is a high-overhead operation since it has to walk the call stack up to its root. We could leverage Java 9's stack-walking API so we can stop climbing the call stack when we found what we want.

See https://dzone.com/articles/whos-calling-me-an-overview-of-stack-walking-api-i for more information about Java 9 stack walking API.

Note that we still need to fall back to the old implementation if running on Java 8, so you might need to add some abstraction layer which allows switch between two different stack-walking implementations.

anuraaga:
This reminds me of the `SystemInfo` change we made recently - if we add more features targeting newer Java versions, it's probably important to tweak the build so we can do it with class loading instead of reflection which can be tedious or slower.

ByeongweonMoon:
I will handle this.

moonchanyong:
I will take this issue.

PR #2055, opened 2019-09-09
Use StackWalker for TTEXT serialization format when on Java 9+(#1686)

moonchanyong:
Motivation:
#1686  

Modification:
Use StackWalker for getting a full stack trace in Thrift TEXT serialization format from above java version 9.

Result:
choose stack trace api according to java version.

### DIFF
--- a/thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
+++ b/thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
@@ -45,8 +45,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 final class StackWalkingThriftMessageClassFinder extends AbstractThriftMessageClassFinder {
-    private final Logger logger =
+
+    private static final Logger logger =
             LoggerFactory.getLogger(StackWalkingThriftMessageClassFinder.class);
+
+    private static final String INVOKING_FAIL_MSG =
+            "Failed to invoke StackWalker.StackFrame.getDeclaringClass():";
 
     @Nullable
     private final Function<Stream<Object>, Class<?>> walkHandler;
@@ -63,7 +67,6 @@ final class StackWalkingThriftMessageClassFinder extends AbstractThriftMessageCl
                                     MethodType.methodType(Object.class, Function.class));
 
         final Class<?> stackFrameClass = classLoader.loadClass("java.lang.StackWalker$StackFrame");
-        final String invokingFailMsg = "Failed to invoke StackWalker.StackFrame.getDeclaringClass():";
         Function<Object, Class<?>> getClassByStackFrameTemp;
         Object instance;
 
@@ -91,7 +94,7 @@ final class StackWalkingThriftMessageClassFinder extends AbstractThriftMessageCl
                 try {
                     return getMatchedClass((Class<?>) getDeclaringClassMH.invoke(stackFrame));
                 } catch (Throwable t) {
-                    logger.warn(invokingFailMsg, t);
+                    logger.warn(INVOKING_FAIL_MSG, t);
                 }
                 return null;
             };
@@ -110,7 +113,7 @@ final class StackWalkingThriftMessageClassFinder extends AbstractThriftMessageCl
                 try {
                     return getMatchedClass(getClassByName(getClassNameMH.invoke(stackFrame).toString()));
                 } catch (Throwable t) {
-                    logger.warn(invokingFailMsg, t);
+                    logger.warn(INVOKING_FAIL_MSG, t);
                 }
                 return null;
             };
@@ -120,10 +123,10 @@ final class StackWalkingThriftMessageClassFinder extends AbstractThriftMessageCl
 
         final Function<Object, Class<?>> getClassByStackFrame = getClassByStackFrameTemp;
         walkHandler = stackFrameStream ->
-                        stackFrameStream.map(getClassByStackFrame)
-                                        .filter(Objects::nonNull)
-                                        .findFirst()
-                                        .orElse(null);
+                stackFrameStream.map(getClassByStackFrame)
+                                .filter(Objects::nonNull)
+                                .findFirst()
+                                .orElse(null);
     }
 
     @Nullable
--- a/thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
+++ b/thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+// =================================================================================================
+// Copyright 2011 Twitter, Inc.
+// -------------------------------------------------------------------------------------------------
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this work except in compliance with the License.
+// You may obtain a copy of the License in the LICENSE file, or at:
+//
+//  https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =================================================================================================
+package com.linecorp.armeria.common.thrift.text;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.invoke.MethodType;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import javax.annotation.Nullable;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class StackWalkingThriftMessageClassFinder extends AbstractThriftMessageClassFinder {
+    private final Logger logger =
+            LoggerFactory.getLogger(StackWalkingThriftMessageClassFinder.class);
+
+    @Nullable
+    private final Function<Stream<Object>, Class<?>> walkHandler;
+    private final MethodHandle walkMH;
+    private final Object stackWalker;
+
+    StackWalkingThriftMessageClassFinder() throws Throwable {
+        final ClassLoader classLoader = ClassLoader.getSystemClassLoader();
+        final Lookup lookup = MethodHandles.lookup();
+
+        final Class<?> stackWalkerClass = classLoader.loadClass("java.lang.StackWalker");
+        walkMH = lookup.findVirtual(stackWalkerClass,
+                                    "walk",
+                                    MethodType.methodType(Object.class, Function.class));
+
+        final Class<?> stackFrameClass = classLoader.loadClass("java.lang.StackWalker$StackFrame");
+        final String invokingFailMsg = "Failed to invoke StackWalker.StackFrame.getDeclaringClass():";
+        Function<Object, Class<?>> getClassByStackFrameTemp;
+        Object instance;
+
+        try {
+            // StackWalker instantiate with RETAIN_CLASS_REFERENCE option
+            final Class<?> Option = classLoader.loadClass("java.lang.StackWalker$Option");
+            final MethodHandle getInstanceMH =
+                    lookup.findStatic(stackWalkerClass,
+                                      "getInstance",
+                                      MethodType.methodType(stackWalkerClass, Option));
+            final Enum<?> RETAIN_CLASS_REFERENCE =
+                    Arrays.stream((Enum<?>[]) Option.getEnumConstants())
+                          .filter(op -> "RETAIN_CLASS_REFERENCE".equals(op.name()))
+                          .findFirst().orElseGet(null);
+
+            if (RETAIN_CLASS_REFERENCE == null) {
+                throw new IllegalStateException("Failed to get RETAIN_CLASS_REFERENCE option");
+            }
+            instance = getInstanceMH.invoke(RETAIN_CLASS_REFERENCE);
+            final MethodHandle getDeclaringClassMH = lookup.findVirtual(stackFrameClass,
+                                                                        "getDeclaringClass",
+                                                                        MethodType.methodType(Class.class));
+
+            getClassByStackFrameTemp = stackFrame -> {
+                try {
+                    return getMatchedClass((Class<?>) getDeclaringClassMH.invoke(stackFrame));
+                } catch (Throwable t) {
+                    logger.warn(invokingFailMsg, t);
+                }
+                return null;
+            };
+        } catch (Throwable throwable) {
+            // StackWalker instantiate without option
+            logger.warn("Falling back to StackWalker without option:", throwable);
+            final MethodHandle getInstanceMH =
+                    lookup.findStatic(stackWalkerClass,
+                                      "getInstance",
+                                      MethodType.methodType(stackWalkerClass));
+            final MethodHandle getClassNameMH =
+                    lookup.findVirtual(stackFrameClass, "getClassName", MethodType.methodType(String.class));
+
+            instance = getInstanceMH.invoke();
+            getClassByStackFrameTemp = stackFrame -> {
+                try {
+                    return getMatchedClass(getClassByName(getClassNameMH.invoke(stackFrame).toString()));
+                } catch (Throwable t) {
+                    logger.warn(invokingFailMsg, t);
+                }
+                return null;
+            };
+        }
+
+        stackWalker = instance;
+
+        final Function<Object, Class<?>> getClassByStackFrame = getClassByStackFrameTemp;
+        walkHandler = stackFrameStream ->
+                        stackFrameStream.map(getClassByStackFrame)
+                                        .filter(Objects::nonNull)
+                                        .findFirst()
+                                        .orElse(null);
+    }
+
+    @Nullable
+    @Override
+    public Class<?> get() {
+        try {
+            return (Class<?>) walkMH.invoke(stackWalker, walkHandler);
+        } catch (Throwable t) {
+            throw new IllegalStateException("Failed to invoke StackWalker.walk():", t);
+        }
+    }
+}
+
--- a/thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
+++ b/thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
@@ -30,15 +30,15 @@
 // =================================================================================================
 package com.linecorp.armeria.common.thrift.text;
 
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Modifier;
+import static com.linecorp.armeria.common.thrift.text.AbstractThriftMessageClassFinder.isTBase;
+
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.function.Supplier;
 
 import javax.annotation.Nullable;
 
-import org.apache.thrift.TApplicationException;
 import org.apache.thrift.TBase;
 import org.apache.thrift.TException;
 import org.apache.thrift.TFieldIdEnum;
@@ -56,7 +56,7 @@ import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.JsonNode;
 
-import com.linecorp.armeria.internal.thrift.TApplicationExceptions;
+import com.linecorp.armeria.common.util.SystemInfo;
 
 /**
  * A struct parsing context. Builds a map from field name to TField.
@@ -65,6 +65,21 @@ import com.linecorp.armeria.internal.thrift.TApplicationExceptions;
  */
 class StructContext extends PairContext {
     private static final Logger log = LoggerFactory.getLogger(StructContext.class);
+    private static final Supplier<Class<?>> thriftMessageClassFinder;
+
+    static {
+        Supplier<Class<?>> supplier = null;
+        if (SystemInfo.javaVersion() >= 9) {
+            try {
+                supplier = new StackWalkingThriftMessageClassFinder();
+            } catch (Throwable t) {
+                log.warn("Failed to initialize StackWalkingThriftMessageClassFinder. " +
+                         "Falling back to DefaultThriftMessageClassFinder:", t);
+            }
+        }
+
+        thriftMessageClassFinder = supplier != null ? supplier : new DefaultThriftMessageClassFinder();
+    }
 
     // When processing a given thrift struct, we need certain information
     // for every field in that struct. We store that here, in a map
@@ -132,63 +147,13 @@ class StructContext extends PairContext {
      * TProtocol.writeStructBegin(), rather than relying on the stack trace.
      */
     private static Class<?> getCurrentThriftMessageClass() {
-        final StackTraceElement[] frames =
-                Thread.currentThread().getStackTrace();
+        final Class<?> clazz = thriftMessageClassFinder.get();
 
-        for (StackTraceElement f : frames) {
-            final String className = f.getClassName();
-
-            try {
-                final Class<?> clazz = Class.forName(className);
-
-                // Note, we need to check
-                // if the class is abstract, because abstract class does not have metaDataMap
-                // if the class has no-arg constructor, because FieldMetaData.getStructMetaDataMap
-                //   calls clazz.newInstance
-                if (isTBase(clazz) && !isAbstract(clazz) && hasNoArgConstructor(clazz)) {
-                    return clazz;
-                }
-
-                if (isTApplicationException(clazz)) {
-                    return clazz;
-                }
-
-                if (isTApplicationExceptions(clazz)) {
-                    return TApplicationException.class;
-                }
-            } catch (ClassNotFoundException ex) {
-                log.warn("Can't find class: " + className, ex);
-            }
-        }
-        throw new RuntimeException("Must call (indirectly) from a TBase/TApplicationException object.");
-    }
-
-    private static boolean isTBase(Class<?> clazz) {
-        return TBase.class.isAssignableFrom(clazz);
-    }
-
-    private static boolean isTApplicationException(Class<?> clazz) {
-        return TApplicationException.class.isAssignableFrom(clazz);
-    }
-
-    private static boolean isTApplicationExceptions(Class<?> clazz) {
-        return clazz == TApplicationExceptions.class;
-    }
-
-    private static boolean isAbstract(Class<?> clazz) {
-        return Modifier.isAbstract(clazz.getModifiers());
-    }
-
-    private static boolean hasNoArgConstructor(Class<?> clazz) {
-        final Constructor<?>[] allConstructors = clazz.getConstructors();
-        for (Constructor<?> ctor : allConstructors) {
-            final Class<?>[] pType = ctor.getParameterTypes();
-            if (pType.length == 0) {
-                return true;
-            }
+        if (clazz == null) {
+            throw new RuntimeException("Must call (indirectly) from a TBase/TApplicationException object.");
         }
 
-        return false;
+        return clazz;
     }
 
     /**
--- a/thrift/src/test/java/com/linecorp/armeria/common/thrift/text/ThriftMessageClassFinderTest.java
+++ b/thrift/src/test/java/com/linecorp/armeria/common/thrift/text/ThriftMessageClassFinderTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.common.thrift.text;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.security.Permission;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import com.linecorp.armeria.common.thrift.text.RpcDebugService.doDebug_args;
+import com.linecorp.armeria.common.util.SystemInfo;
+
+class ThriftMessageClassFinderTest {
+    @ParameterizedTest(name = "testThriftMessageClassFinder {index}: finder={0}")
+    @MethodSource("testThriftMessageClassParameters")
+    void testThriftMessageClassFinder(Supplier<Class<?>> thriftMessageClassFinder) {
+        assertThat(thriftMessageClassFinder.get()).isNull();
+        assertThat(new MockArgs().proxy(thriftMessageClassFinder)).isNotNull();
+    }
+
+    private static Stream<Arguments> testThriftMessageClassParameters() throws Throwable {
+        Stream<Arguments> parameters = Stream.of(Arguments.of(new DefaultThriftMessageClassFinder()));
+
+        if (SystemInfo.javaVersion() >= 9) {
+            parameters = Stream.concat(
+                    parameters,
+                    Stream.of(
+                            Arguments.of(new StackWalkingThriftMessageClassFinder()),
+                            Arguments.of(getNoOptionStackWalkerInstance())));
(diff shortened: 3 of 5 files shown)


moonchanyong:
sorry I m late and poor English.

I wrote this code based below lines.

1. I don't care comment(naming, error msg) yet.
2. I wanna that logic to be a small function for sharing java 8 flow and 9. so I distribute in each function (```getClassName```, ```getClassByName```, ```matchClass```, ```streamProcessor```)
3. ```StackTraceElement``` & ```StackFrame(in java9)``` has same name method ```getClassName```. so I use reflection to Object parameter for made common function.


is it the right way that using stream and made small common functions even if using reflection in ```getClassName```?

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -132,35 +138,85 @@ protected TField getTFieldByName(String name) throws TException {
...
+                final Object stackWalkerInstance = stackWalker
+                        .getDeclaredMethod("getInstance")
+                        .invoke(null);
+                final Method walk = stackWalkerInstance.getClass().getDeclaredMethod("walk", Function.class);
+                result = (Optional<Class>) walk.invoke(
+                        stackWalkerInstance,
+                        (Function<Stream<Object>, Optional<Class>> ) s -> streamProcessor(s));
+            } catch (Throwable t) {
+                // @TODO: msg
+                log.debug("walwlawla", "StackWalker.getInstance.walk", t);
```
trustin: In this case, we could log at WARN level once and then fall back to Java 8 version.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -132,35 +138,85 @@ protected TField getTFieldByName(String name) throws TException {
      * TProtocol.writeStructBegin(), rather than relying on the stack trace.
      */
     private static Class<?> getCurrentThriftMessageClass() {
-        final StackTraceElement[] frames =
-                Thread.currentThread().getStackTrace();
-
-        for (StackTraceElement f : frames) {
-            final String className = f.getClassName();
-
+        if (SystemInfo.javaVersion() >= 9) {
```
trustin: Instead of doing this on every `getCurrentThriftMessageClass()`, we could detect the Java version only once at the class initializer (`static` block), e.g.

```java
final class StructContext ... {
    private static final Supplier<Class<?>> currentThriftMessageClass;
    static {
        Supplier<Class<?>> supplier = null;
        if (SystemInfo.javaVersion() >= 9) {
            try {
                supplier = new StackWalkingCurrentThriftMessageClassFinder();
            } catch (Throwable t) {
                log.warn(...);
            }
        }

        currentThriftMessageClass = supplier != null ? supplier : new DefaultCurrentThriftMessageClassFunder();
        }
    }
    ...
}
```

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -132,35 +138,85 @@ protected TField getTFieldByName(String name) throws TException {
...
-                // if the class has no-arg constructor, because FieldMetaData.getStructMetaDataMap
-                //   calls clazz.newInstance
-                if (isTBase(clazz) && !isAbstract(clazz) && hasNoArgConstructor(clazz)) {
-                    return clazz;
-                }
+                final Class<?> stackWalker = Class.forName("java.lang.StackWalker");
+                final Object stackWalkerInstance = stackWalker
+                        .getDeclaredMethod("getInstance")
+                        .invoke(null);
+                final Method walk = stackWalkerInstance.getClass().getDeclaredMethod("walk", Function.class);
```
trustin: Could we use `MethodHandle` which is way faster than Java reflection? https://www.baeldung.com/java-method-handles

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -132,35 +138,85 @@ protected TField getTFieldByName(String name) throws TException {
...
-                    return TApplicationException.class;
-                }
-            } catch (ClassNotFoundException ex) {
-                log.warn("Can't find class: " + className, ex);
+            if (result.isPresent()) {
+                return result.get();
             }
         }
-        throw new RuntimeException("Must call (indirectly) from a TBase/TApplicationException object.");
+        throw  new RuntimeException("Must call (indirectly) from a TBase/TApplicationException object.");
```
trustin: nit: Double space

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -132,35 +138,85 @@ protected TField getTFieldByName(String name) throws TException {
...
+                .map(StructContext::getClassByName)
+                .filter(Optional::isPresent)
+                .map(StructContext::matchClass)
+                .filter(Objects::nonNull)
+                .findFirst();
+    }
+
+    private static String getClassName(Object stackFrame) {
+        try {
+            final Method getClassName = stackFrame.getClass().getDeclaredMethod("getClassName");
```
trustin: This is inefficient. Use `StackTraceElement.getClassName()` directly for the default implementation and `MethodHandle` for the Java 9 implementation. Once you introduce some class hierarchy as I advised, you could provide two different `getClassName()` implementations.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -132,35 +138,85 @@ protected TField getTFieldByName(String name) throws TException {
...
+            return name.toString();
+        } catch (Throwable t) {
+            // @TODO:msg
+            log.debug("", "", t);
+        }
+        // @TODO: msg
+        throw new RuntimeException("");
+    }
+
+    private static Optional<Class> getClassByName(String className) {
```
trustin: Could we just return `Class` and annotated this method with `@Nullable`? `Optional` is nice but it's not certain whether JVM's escape analysis will be happy with it.

trustin:
@anuraaga wrote:

> @trustin What do you think about making a multi-release jar?

I like the idea. Let me create an issue so we can give it a try in a separate pull request.

trustin:
https://github.com/line/armeria/issues/2062

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,105 @@
...
+        Class<?> clazz = null;
+        try {
+            final Object sw = getInstanceMH.invoke();
+            clazz = (Class<?>) walkMH.invoke(sw, walkHandler);
+        } catch (Throwable t) {
+            log.warn("fail to invoke StackWalker.walk.", t);
+        }
+
+        if (clazz == null) {
+            log.warn("can't find CurrentThriftMessageClass using StackWalker. retry using java 8 version");
```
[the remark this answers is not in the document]
trustin: No, I meant doing that in the static initializer. :-)

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/DefaultCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public Class<?> get() {
+        final StackTraceElement[] frames =
+                Thread.currentThread().getStackTrace();
+
+        for (StackTraceElement f : frames) {
+            final String className = f.getClassName();
+            final Class<?> clazz = getClassByName(className);
+
+            if (clazz == null) {
```
trustin: Now we do not need this check anymore because getMatchedClass() handles this case.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,102 @@
...
+    private final Logger logger =
+            LoggerFactory.getLogger(StackWalkingCurrentThriftMessageClassFinder.class);
+    private final DefaultCurrentThriftMessageClassFinder defaultCurrentThriftClassFinder =
+            new DefaultCurrentThriftMessageClassFinder();
+
+    private final MethodHandle walkMH;
+    private final MethodHandle getInstanceMH;
+    private MethodHandle getClassNameMH;
+
+    StackWalkingCurrentThriftMessageClassFinder() throws Throwable {
```
trustin: 'throws Exception' should be enough.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,102 @@
...
+        final MethodHandles.Lookup lookup = MethodHandles.lookup();
+        final MethodType walkMT = MethodType.methodType(Object.class, Function.class);
+        final Class<?> stackWalker = Class.forName("java.lang.StackWalker");
```
trustin: Could you specify the class loader, like the system class loader.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,102 @@
...
+        walkMH = lookup.findVirtual(stackWalker, "walk", walkMT);
+
+        final MethodType getInstanceMT = MethodType.methodType(stackWalker);
+        getInstanceMH = lookup.findStatic(stackWalker, "getInstance", getInstanceMT);
+
+        final Class<?> stackFrame = Class.forName("java.lang.StackWalker$StackFrame");
```
trustin: Ditto - please specify the class loader for consistent behavior, because otherwise the context class loader is used.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,102 @@
...
+        final MethodType getClassNameMT = MethodType.methodType(String.class);
+        getClassNameMH = lookup.findVirtual(stackFrame, "getClassName", getClassNameMT);
+    }
+
+    @Nullable
+    private final Function<Stream<Object>, Class<?>> walkHandler =  s ->
+            s.map(stackFrame -> {
+                try {
+                    return Class.forName(getClassNameMH.invoke(stackFrame).toString());
```
trustin: Could use getClassByName() in the superclass?

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,102 @@
...
+                } catch (Throwable t) {
+                    logger.warn("fail to invoke StackWalker.StackFrame.getDeclaringClass.", t);
```
trustin: - fail -> Failed
- getDeclaringClass. -> getDeclaringClass():
  - Note colon and parens

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,102 @@
...
+            .orElse(null);
+
+    @Nullable
+    @Override
+    public Class<?> get() {
+        final Class<?> clazz;
+
+        try {
+            final Object sw = getInstanceMH.invoke();
+            clazz = (Class<?>) walkMH.invoke(sw, walkHandler);
```
trustin: No need for a local variable; just return.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,102 @@
...
+        } catch (Throwable t) {
+            logger.warn("fail to invoke StackWalker.walk. " +
+                        "retry using DefaultCurrentThriftMessageClassFinder", t);
+            return defaultCurrentThriftClassFinder.get();
```
trustin: I guess we do not need to fallback into the default implementation because it should never fail if this class is run from Java 9+. How about just rethrowing the exception? (wrap with an IllegalStateException if necessary.)

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -65,6 +65,21 @@
...
+    private static final Supplier<Class<?>> currentThriftMessageClass;
+
+    static {
+        Supplier<Class<?>> supplier = null;
+        if (SystemInfo.javaVersion() >= 9) {
+            try {
+                supplier = new StackWalkingCurrentThriftMessageClassFinder();
+            } catch (Throwable t) {
+                log.warn("can't use StackWalkingCurrentThriftMessageClassFinder. " +
+                         "use DefaultCurrentThriftMessageClassFinder.", t);
```
trustin: - can't use -> Failed to initialize
- use -> Falling back to 
- . -> :

thrift/src/test/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinderTest.java
```diff
@@ -0,0 +1,40 @@
...
+            currentThriftMessageClassFinder = new DefaultCurrentThriftMessageClassFinder();
+        }
+    }
+
+    @Test
+    void notFoundThriftMessageClass() {
+        final Class<?> clazz = currentThriftMessageClassFinder.get();
+        assertThat(clazz).isNull();
+    }
+}
```
trustin: Could you add the case for a successful lookup?

thrift/src/test/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinderTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.jupiter.api.Test;
+
+class StackWalkingCurrentThriftMessageClassFinderTest {
+    Supplier<Class<?>> currentThriftMessageClassFinder;
+
+    {
+        try {
+            currentThriftMessageClassFinder = new StackWalkingCurrentThriftMessageClassFinder();
+        } catch (Throwable throwable) {
+            currentThriftMessageClassFinder = new DefaultCurrentThriftMessageClassFinder();
```
trustin: You could use an assumption to skip the test on Java 8:

    assumeTrue(SystemInfo.javaVersion() >= 9);

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
anuraaga: @trustin Since we guard the construction of this class by the version check, do we actually need to use reflection? Since class loading is lazy, and we require Java 11 to build, I think it should be ok and makes the code simpler.

trustin: We can't just yet because we pass '-release 8'.

anuraaga: Ah ok :)

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,98 @@
...
+            s.map(stackFrame -> {
+                try {
+                    return getClassByName(getClassNameMH.invoke(stackFrame).toString());
+                } catch (Throwable t) {
+                    logger.warn("Failed to invoke StackWalker.StackFrame.getDeclaringClass():", t);
+                }
+
+                return null;
+            })
+            .map(AbstractCurrentThriftMessageClassFinder::getMatchedClass)
```
trustin: How about merging this map() with the map() above for efficiency?

    returb getMatchedClass(getClassByName(...));

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,98 @@
...
+            .findFirst()
+            .orElse(null);
+
+    @Nullable
+    @Override
+    public Class<?> get() {
+        final Class<?> clazz;
+
+        try {
+            clazz = (Class<?>) walkMH.invoke(getInstanceMH.invoke(), walkHandler);
```
trustin: Local variable unnecessary:

    return (Clazz<?>) walkMH...;

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/AbstractCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,110 @@
...
+
+abstract class AbstractCurrentThriftMessageClassFinder implements Supplier<Class<?>> {
+    private static final Logger logger = LoggerFactory.getLogger(AbstractCurrentThriftMessageClassFinder.class);
+
+    @Nullable
+    static Class<?> getClassByName(String className) {
+        try {
+            return ClassLoader.getSystemClassLoader().loadClass(className);
+        } catch (ClassNotFoundException ex) {
+            logger.warn("Can't find class: {}", className, ex);
```
trustin: class -> a class

thrift/src/test/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinderTest.java
```diff
@@ -0,0 +1,63 @@
...
+        assertThat(clazz).isNotNull();
+    }
+
+    @Test
+    void notFoundThriftMessageClass() {
+        final Class<?> clazz = currentThriftMessageClassFinder.get();
+        assertThat(clazz).isNull();
+    }
+
+    public static class MockArgs extends doDebug_args {
```
trustin: public -> private?

thrift/src/test/java/com/linecorp/armeria/common/thrift/text/DefaultCurrentThriftMessageClassFinderTest.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test
+    void foundThriftMessageClass() throws TException {
+        final MockArgs mock = new MockArgs();
+        final Class<?> clazz = mock.callThriftMessageClassFinder();
+        assertThat(clazz).isNotNull();
+    }
+
+    public static class MockArgs extends doDebug_args {
```
trustin: public -> private?

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,98 @@
...
+        } catch (Throwable t) {
+            throw new IllegalStateException("Failed to invoke StackWalker.walk:", t);
```
trustin: walk -> walk()

thrift/src/test/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinderTest.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.annotation.Nullable;
+
+import org.apache.thrift.TException;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.Test;
+
+import com.linecorp.armeria.common.thrift.text.RpcDebugService.doDebug_args;
+import com.linecorp.armeria.common.util.SystemInfo;
+
+class StackWalkingCurrentThriftMessageClassFinderTest {
```
trustin: I see code duplication between this test and DefaultCurrentThriftMessageClassFinderTest. How about merging them into one by using parameterized tests? (You could provide 2 finder instances for Java 9+ and only the default one for Java 8.)

thrift/src/test/java/com/linecorp/armeria/common/thrift/text/CurrentThriftMessageClassFinderTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import com.linecorp.armeria.common.thrift.text.RpcDebugService.doDebug_args;
+import com.linecorp.armeria.common.util.SystemInfo;
+
+class CurrentThriftMessageClassFinderTest {
+
+    @ParameterizedTest(name = "testThriftMessageClassFinder {index}: finder={0}, isLookupCase={1}")
+    @MethodSource("testThriftMessageClassParameters")
```
moonchanyong: is it fine using MethodSource, unlike other codes? 
other codes are using ArgumentsSource.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,93 @@
...
+            })
+            .filter(Objects::nonNull)
+            .findFirst()
+            .orElse(null);
+
+    @Nullable
+    @Override
+    public Class<?> get() {
+        try {
+            return (Class<?>) walkMH.invoke(getInstanceMH.invoke(), walkHandler);
```
trustin: Can we reuse the `StackWalker` instance retrieved from `StackWalker.getInstance()`? If so, we do not need to invoke `getInstanceMH.invoke()` every time.

thrift/src/test/java/com/linecorp/armeria/common/thrift/text/CurrentThriftMessageClassFinderTest.java
trustin: I think it's fine, but we don't need the second parameter - we can just write two tests, because they are testing two different scenarios.

trustin:
I had a chance to review the `StackWalker` API and found that we do not need `Class.forName()` if we get the `StackWalker` instance with the `RETAIN_CLASS_REFERENCE` option:

    StackWalker.getInstance(Option.RETAIN_CLASS_REFERENCE);

If `getInstance()` succeeds, we can get the class from `StackFrame.getDeclaringClass()`, which means a lot more performance boost.

If `getInstance()` fails with `SecurityException`, we can fall back to the default `StackWalker` retrieved from `StackWalker.getInstance()` like we do now.

What do you think about updating this pull request so that the `RETAIN_CLASS_REFERENCE` is utilized? Perhaps we could keep a `boolean` flag which tells if the `StackWalker` we retrieved has the option or not.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/AbstractCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        return null;
+    }
+
+    static boolean isTBase(Class<?> clazz) {
+        return TBase.class.isAssignableFrom(clazz);
+    }
+
+    static boolean isTApplicationExceptions(Class<?> clazz) {
```
ikhoon: Could be `private`?

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,93 @@
...
+        final MethodType getInstanceMT = MethodType.methodType(stackWalker);
+        getInstanceMH = lookup.findStatic(stackWalker, "getInstance", getInstanceMT);
+
+        final Class<?> stackFrame = classLoader.loadClass("java.lang.StackWalker$StackFrame");
+        final MethodType getClassNameMT = MethodType.methodType(String.class);
+        getClassNameMH = lookup.findVirtual(stackFrame, "getClassName", getClassNameMT);
+    }
+
+    @Nullable
+    private final Function<Stream<Object>, Class<?>> walkHandler = s ->
```
ikhoon: nit: rename `s` -> `stackFrameStream`?

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -65,6 +65,21 @@
  */
 class StructContext extends PairContext {
     private static final Logger log = LoggerFactory.getLogger(StructContext.class);
+    private static final Supplier<Class<?>> currentThriftMessageClass;
```
ikhoon: nit: How about rename `currentThriftMessageClass` -> `currentThriftMessageClassFinder`?

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/AbstractCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,110 @@
...
+import javax.annotation.Nullable;
+
+import org.apache.thrift.TApplicationException;
+import org.apache.thrift.TBase;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.linecorp.armeria.internal.thrift.TApplicationExceptions;
+
+abstract class AbstractCurrentThriftMessageClassFinder implements Supplier<Class<?>> {
```
minwoox: I think we can remove the word `current` from this class and all other subclasses.

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
```diff
@@ -132,63 +147,13 @@ protected TField getTFieldByName(String name) throws TException {
      * TProtocol.writeStructBegin(), rather than relying on the stack trace.
      */
```
minwoox: So this description goes to `DefaultCurrentThriftMessageClassFinder` right?

moonchanyong: `StackWalking` and `getStackTrace`  are getting the same way on the stacktrace ? [JEP-259](https://openjdk.java.net/jeps/259)

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingCurrentThriftMessageClassFinder.java
```diff
@@ -0,0 +1,93 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class StackWalkingCurrentThriftMessageClassFinder extends AbstractCurrentThriftMessageClassFinder {
+    private final Logger logger =
+            LoggerFactory.getLogger(StackWalkingCurrentThriftMessageClassFinder.class);
+
+    private final MethodHandle walkMH;
+    private final MethodHandle getInstanceMH;
+    private MethodHandle getClassNameMH;
```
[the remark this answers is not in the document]
moonchanyong: moved workHandler initializing code to constructor for attaching final

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
```diff
@@ -0,0 +1,137 @@
...
+                } catch (Throwable t) {
+                    logger.warn(invokingFailMsg, t);
+                }
+                return null;
+            };
+        }
+
+        stackWalker = instance;
+
+        final Function<Object, Class<?>> getClassByStackFrame = getClassByStackFrameTemp;
```
trustin: Can we make `getClassByStackFrameTemp` final? Then we would not need this local variable.

thrift/src/test/java/com/linecorp/armeria/common/thrift/text/ThriftMessageClassFinderTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private static Supplier<Class<?>> getNoOptionStackWalkerInstance() throws Throwable {
+        final SecurityManager SM = System.getSecurityManager();
+        System.setSecurityManager(new SecurityManager() {
+            @Override
+            public void checkPermission(Permission perm) {
+                // `getStackWalkerWithClassReference` is called by RETAIN_CLASS_REFERENCE option.
+                if (perm.getName().equals("getStackWalkerWithClassReference")) {
+                    throw new SecurityException("Fail SecurityManage.checkPermission() for Unit Testing");
```
trustin: - Fail -> Failing
- Unit Testing -> unit testing

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
```diff
@@ -0,0 +1,137 @@
...
+                                    "walk",
+                                    MethodType.methodType(Object.class, Function.class));
+
+        final Class<?> stackFrameClass = classLoader.loadClass("java.lang.StackWalker$StackFrame");
+        final String invokingFailMsg = "Failed to invoke StackWalker.StackFrame.getDeclaringClass():";
+        Function<Object, Class<?>> getClassByStackFrameTemp;
+        Object instance;
+
+        try {
+            // StackWlaker instantiate with RETAIN_CLASS_REFERENCE option
```
minwoox: nit: StackWlaker -> StackWalker

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
```diff
@@ -0,0 +1,137 @@
...
+                                      MethodType.methodType(stackWalkerClass, Option));
+            final Enum<?> RETAIN_CLASS_REFERENCE = Arrays.stream((Enum<?>[]) Option.getEnumConstants())
+                    .filter(op -> "RETAIN_CLASS_REFERENCE".equals(op.name())).findFirst().orElseGet(null);
+
+            if (RETAIN_CLASS_REFERENCE == null) {
+                throw new IllegalStateException("Failed to get RETAIN_CLASS_REFERENCE option");
+            }
+            instance = getInstanceMH.invoke(RETAIN_CLASS_REFERENCE);
+            final MethodHandle getDeclaringClassMH = lookup.findVirtual(stackFrameClass,
+                                                                "getDeclaringClass",
```
minwoox: nit: Could align this?:
```java
... lookup.findVirtual(stackFrameClass,
                       "getDeclaringClass",
                       MethodType.methodType(Class.class));
```

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        stackWalker = instance;
+
+        final Function<Object, Class<?>> getClassByStackFrame = getClassByStackFrameTemp;
+        walkHandler = stackFrameStream ->
+                        stackFrameStream.map(getClassByStackFrame)
+                        .filter(Objects::nonNull)
+                        .findFirst()
+                        .orElse(null);
```
minwoox: nit: indentation?
```java
walkHandler = stackFrameStream ->
        stackFrameStream.map(getClassByStackFrame)
                        .filter(Objects::nonNull)
                        .findFirst()
                        .orElse(null);
```

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
```diff
@@ -0,0 +1,137 @@
...
+
+        try {
+            // StackWlaker instantiate with RETAIN_CLASS_REFERENCE option
+            final Class<?> Option = classLoader.loadClass("java.lang.StackWalker$Option");
+            final MethodHandle getInstanceMH =
+                    lookup.findStatic(stackWalkerClass,
+                                      "getInstance",
+                                      MethodType.methodType(stackWalkerClass, Option));
+            final Enum<?> RETAIN_CLASS_REFERENCE = Arrays.stream((Enum<?>[]) Option.getEnumConstants())
+                    .filter(op -> "RETAIN_CLASS_REFERENCE".equals(op.name())).findFirst().orElseGet(null);
```
minwoox: nit: this needs indent as well. :-)

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StructContext.java
minwoox: Ah yes. Didn't read it carefully. Forget about the comment. 😉

thrift/src/main/java/com/linecorp/armeria/common/thrift/text/StackWalkingThriftMessageClassFinder.java
```diff
@@ -0,0 +1,139 @@
...
+        final ClassLoader classLoader = ClassLoader.getSystemClassLoader();
+        final Lookup lookup = MethodHandles.lookup();
+
+        final Class<?> stackWalkerClass = classLoader.loadClass("java.lang.StackWalker");
+        walkMH = lookup.findVirtual(stackWalkerClass,
+                                    "walk",
+                                    MethodType.methodType(Object.class, Function.class));
+
+        final Class<?> stackFrameClass = classLoader.loadClass("java.lang.StackWalker$StackFrame");
+        final String invokingFailMsg = "Failed to invoke StackWalker.StackFrame.getDeclaringClass():";
```
ikhoon: I think this is useless if an exception is not thrown. How about embedding into loggger.warn(...) with duplication?

```
logger.warn("Failed to invoke StackWalker.StackFrame.getDeclaringClass():", t);
```

trustin:
Thanks, @moonchanyong! :tada:

PR #2063, opened 2019-09-11
Support CORS options for actuator.

anuraaga:
It's common to enable CORS for actuator endpoints for use from an admin console served from a different domain. This registers the CORS decorator respecting the spring configuration for it.

Also adds `Iterable` versions of all vararg methods in the CORS builders to follow our general practice.


anuraaga:
Another way to think about it, it is a coincidence that `GrpcService` supports CORS, since it happens to use `doPost` and provide routes without any method restrictions. It seems weird that changing it to provide routes with `POST` restriction would cause CORS to stop working.

trustin:
@anuraaga Perhaps it’s good time to continue tinkering with this PR now that you can so `sb.routeDecorator().catchAll().build(CorsDecorator.newDecorator(...))`? 😉

anuraaga:
Indeed, it works!

spring/boot-actuator-autoconfigure/src/main/java/com/linecorp/armeria/spring/actuate/CorsEndpointProperties.java
```diff
@@ -0,0 +1,115 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.convert.DurationUnit;
+
+/**
+ * Fork of {@link org.springframework.boot.actuate.autoconfigure.endpoint.web.CorsEndpointProperties}. Upstream
+ * has a dependency on {@code CorsConfiguration} which we don't need.
```
trustin: If this is a fork, could we 1) retain their copyright header and 2) mention which version we forked from?

anuraaga: Oh yeah added copyright and permalink

core/src/main/java/com/linecorp/armeria/server/cors/AbstractCorsPolicyBuilder.java
```diff
@@ -68,15 +68,15 @@
         origins = Collections.emptySet();
     }
 
-    AbstractCorsPolicyBuilder(String... origins) {
+    AbstractCorsPolicyBuilder(List<String> origins) {
         requireNonNull(origins, "origins");
-        checkArgument(origins.length > 0, "origins is empty.");
-        for (int i = 0; i < origins.length; i++) {
-            if (origins[i] == null) {
+        checkArgument(origins.size() > 0, "origins is empty.");
```
minwoox: nit: `!origins.isEmpty()`

core/src/main/java/com/linecorp/armeria/server/cors/CorsServiceBuilder.java
```diff
@@ -207,6 +221,19 @@ public CorsServiceBuilder maxAge(long maxAge) {
...
+    /**
+     * Sets the CORS {@code "Access-Control-Max-Age"} response header and enables the
+     * caching of the preflight response for the specified time. During this time no preflight
+     * request will be made.
+     *
+     * @param maxAge the maximum time that the preflight response may be cached. Rounded to seconds.
+     * @return {@link CorsServiceBuilder} to support method chaining.
+     */
+    public CorsServiceBuilder maxAge(Duration maxAge) {
+        firstPolicyBuilder.maxAge(TimeUnit.MILLISECONDS.toSeconds(maxAge.toMillis()));
```
minwoox: nit: usually we delegate this to `maxAge(long maxAge) {...}`?

minwoox: nit: `requireNonNull`

anuraaga:
Thanks for cleaning up for me :)

trustin:
Thank you, too!

PR #2069, opened 2019-09-12
Add Java 9 overrides of platform-specific APIs.

### LINKED ISSUE #2062: Multi-release JAR
We currently rely on `MethodHandle`s or `java.lang.reflect` to access the classes or methods available in Java 9+ because our minimum requirement is Java 8. We could simplify this a lot by using the multi-release JAR:

https://github.com/melix/mrjar-gradle

anuraaga:
Works around proguard not supporting this by reassembling a multi-release JAR in the end. I guess at some point someone would copy the gradle changes into a `java-mrjar.gradle` in the scripts repo :) When doing so, I would add the `--release` argument globally without requiring any special tag as it's better to specify it than not to get animal-sniffer type checking automatically by javac.

Fixes #2062


core/build.gradle
```diff
@@ -1,4 +1,59 @@
...
+    manifest.attributes(
+            'Multi-Release': 'true'
+    )
+}
+
+// Proguard can't handle multi-release jars so we recreate it here.
+tasks.trimShadedJar.doLast {
+    // outjars is a file, so only one jar generated for sure
+    def trimmed = tasks.trimShadedJar.outJarFiles[0].toPath()
+    def copied = trimmed.resolveSibling(trimmed.fileName.toString() + '.bak')
```
trustin: This is not really copied. `tmp`?

core/build.gradle
```diff
@@ -1,4 +1,59 @@
...
+    java.nio.file.Files.move(trimmed, copied, java.nio.file.StandardCopyOption.REPLACE_EXISTING)
+    ant.zip(destfile: trimmed.toString()) {
```
trustin: Does `zip` produce a valid JAR? IIRC, a valid JAR's first entry must be `MANIFEST.MF` (or at least some implementations assume so).

core/build.gradle
```diff
@@ -1,4 +1,59 @@
+String compat(String src) {
```
minwoox: trim? I didn't know what that means when I first saw this. 😅

core/src/main/java9/com/linecorp/armeria/internal/Java9VersionSpecific.java
```diff
@@ -0,0 +1,34 @@
...
+package com.linecorp.armeria.internal;
+
+import java.time.Clock;
+import java.time.Instant;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Implementation of {@link JavaVersionSpecific} using Java 9 APIs.
+ */
+class Java9VersionSpecific extends Java8VersionSpecific {
```
ikhoon: I wonder extending `Java8VersionSpecific` is good for here.
How about extending `JavaVersionSpecific` and delegating `Java8VersionSpecific` if you need?

anuraaga: I think it's very common for newer versions to still use most of the APIs from previous ones, and override just one or two, so inheriting makes that simpler compared to always delegating everything in newer classes. For Java 9, it's not so useful since it's the first level of overriding, but I'd expect Java 10 to benefit from inheriting from Java 9. I figured Java 9 could inherit from 8 for consistency, but what do you think?

core/build.gradle
anuraaga: Thanks didn't know about this - and I found ant's jar command which cleaned this up even more thanks to the perfect `update` parameter :D

core/src/main/java9/com/linecorp/armeria/internal/Java9VersionSpecific.java
trustin: This makes me wonder if it's a good idea to merge Java8VersionSpecific into JavaVersionSpecific. WDYT?

ikhoon: Thanks for the kind explanation. The reason why I avoid inheritance is I think some [APIs could be removed](https://dzone.com/articles/apis-to-be-removed-from-java-11) in next versions of Java. But I agree with your opinion.

anuraaga: Good idea - merged 8 implementation into `JavaVersionSpecific` since it is indeed the "defualt implementation"

ikhoon: I have a question on dependency between multiple java versions.
```
main/java
    /java9
    /java10 
    /java1x
```
I'm not special on this. But I want to check this case. If some application work on Java 10, does it combine `java` + `java9` + `java10`?
I think it is possible to see only `java` + `java10` which do not include `java9`.

I will pull this PR and check it. :-)

trustin: Thanks @ikhoon. Let us know once the result is out.

core/src/main/java9/com/linecorp/armeria/internal/Java9VersionSpecific.java
```diff
@@ -0,0 +1,34 @@
...
+import java.time.Instant;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Implementation of {@link JavaVersionSpecific} using Java 9 APIs.
+ */
+class Java9VersionSpecific extends JavaVersionSpecific {
+
+    @Override
+    public final long currentTimeMicros() {
```
ikhoon: remove `final` to give a change override by the next version of Java?

ikhoon:
Base on [JEP 238: Multi-Release JAR Files](https://openjdk.java.net/jeps/238)
```
jar root
  - A.class
  - B.class
  - C.class
  - D.class
  - META-INF
     - versions
        - 9
           - A.class
           - B.class
        - 10
           - A.class
```

When running the above code with Java 10, we will get
```
A - Java 10
B - Java 9
C - root
D - root
```

Java 9 and Java 10 files dependent only root with [mrjar-gradle
  - `Added Java 10.` PR](https://github.com/melix/mrjar-gradle/pull/3/files) 

If Java 10 want to compile Java 9 `B.class`, we need another configuration. 

```gradle
java10Implementation files([sourceSets.main.output.classesDirs, sourceSets.java9.output.classesDirs]) { builtBy compileJava }
```

core/src/main/java9/com/linecorp/armeria/internal/Java9VersionSpecific.java
anuraaga: While I think the inheritance will help in general, I figured having one step to opt-in to overriding a method (remove final) e.g., in Java 10 is a good way to avoid accidental overriding which I guess you were worried about too :)

ikhoon: OK, That's a good idea! 👍

PR #2073, opened 2019-09-13
Adds EndpointGroup.isStaticIPs() to understand if endpoints are resolved

### LINKED ISSUE #2071: Signal to know if a composite endpoint group is dynamic
In zipkin, we have this code which now looks at a deprecated type:

```java
    if (endpointGroup instanceof StaticEndpointGroup && endpointGroup.endpoints().size() == 1) {
      // Just one non-domain URL, can connect directly without enabling load balancing.
      return endpointGroup.endpoints().get(0);
    }
```

Later, we look at `DynamicEndpointGroup`, but it isn't precisely the opposite of static. For example, `CompositeEndpointGroup` can contain dynamic endpoint groups.

There's no clear signal I can tell if an endpoint group is fully comprised of IP addresses or not, at least not with relying on a deprecated type.

adriancole:
When an `EndpointGroup` is not resolved, it cannot be guaranteed to
return a non-empty result. This adds `isStaticIPs()` to help determine
if it is necessary to await resolution or not.

Fixes #2071

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Endpoint.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Endpoint.java
@@ -622,6 +622,11 @@ public final class Endpoint implements Comparable<Endpoint>, EndpointGroup {
         }
     }
 
+    @Override
+    public boolean isStaticIPs() {
+        return hostType != null && hostType != HostType.HOSTNAME_ONLY;
+    }
+
     private void ensureGroup() {
         if (!isGroup()) {
             throw new IllegalStateException("not a group endpoint");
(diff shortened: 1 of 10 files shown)


trustin:
`isStaticIP()` doesn't sound right to me because it tries to do two things: 1) checking whether the endpoint group has static endpoints and 2) checking whether all the endpoints have IP addresses.

How about just adding `hasIpAddr()` to `EndpointGroup` and letting it return `true` when all endpoints have been resolved into IP addresses? It would not necessarily mean that the group is static, though, i.e. you might want to add `isStatic()` as well.

adriancole:
slight change though. `Endpoint.hasIpAddr()` currently throws. I think it would be better to have it stop doing that since EndpointGroup should be able to not throw and just return false, right?

```
     * @throws IllegalStateException if this endpoint is not a host but a group
```

adriancole:
not needed anymore

PR #2074, opened 2019-09-13
Make sure listeners are notified after first health check has resolved.

anuraaga:
Currently if all endpoints start out unhealthy, then the initial endpoints future for a health checked endpoint group never completes. The initial endpoints future should indicate when the endpoints have first been checked, not when they first become healthy - while an `EndpointGroup` with no `Endpoint`s is sort of useless, I think it's still a perfectly good `EndpointGroup` and deserves notifying listeners :)

/cc @adriancole

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
@@ -17,6 +17,8 @@ package com.linecorp.armeria.client.endpoint;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
```diff
@@ -106,7 +113,7 @@ protected final void setEndpoints(Iterable<Endpoint> endpoints) {
     }
 
     private void completeInitialEndpointsFuture(List<Endpoint> endpoints) {
-        if (!endpoints.isEmpty() && !initialEndpointsFuture.isDone()) {
```
adriancole: ah gotiti

core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
```diff
@@ -32,7 +34,12 @@
  * A dynamic {@link EndpointGroup}. The list of {@link Endpoint}s can be updated dynamically.
  */
 public class DynamicEndpointGroup extends AbstractListenable<List<Endpoint>> implements EndpointGroup {
-    private volatile List<Endpoint> endpoints = ImmutableList.of();
+
+    // An empty list of endpoints we also use as a marker that we have not initialized endpoints yet.
+    private static final List<Endpoint> UNINITIALIZED_ENDPOINTS = Collections.unmodifiableList(
```
minwoox: nit: `ImmutabliList.of()` for consistency?

trustin: I guess he intentionally used it so that `UNINITIALIZED_ENDPOINTS != ImmutableList.of()` because `ImmutableList.of()` will return a singleton, which means a `DynamicEndpointGroup` will enter the 'unintialized' state again when the endpoint list becomes empty later.

minwoox: Ah, I missed that. Thanks!

trustin:
Thanks, @ikhoon !

trustin:
Oops. Thanks, @anuraaga !

PR #2076, opened 2019-09-13
Support gRPC web trailers in client.

### LINKED ISSUE #2030: gRPC-web client doesn't apply framing
From slack, it seems that the gRPC web serialization format may not be working correctly on client side. Any of our formats should work both in server and client.

https://line-armeria.slack.com/archives/C1NGPBUH2/p1566800641169400?thread_ts=1566800641.169400&cid=C1NGPBUH2

anuraaga:
Largish diff since refactors deframed message to provide type to business logic to be able to handle special types (currently only one in existence is grpc-web trailers)

Fixes #2030

### DIFF
--- a/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaMessageDeframer.java
+++ b/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaMessageDeframer.java
@@ -84,8 +84,8 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
     private static final int HEADER_LENGTH = 5;
     private static final int COMPRESSED_FLAG_MASK = 1;
     private static final int RESERVED_MASK = 0x7E;
-    // Valid type always has 0s for middle bits.
-    private static final int UNINITIALIED_TYPE = 0xFF;
+    // Valid type is always positive.
+    private static final int UNINITIALIED_TYPE = -1;
 
     /**
      * A deframed message. For uncompressed messages, we have the entire buffer available and return it
--- a/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaMessageDeframer.java
+++ b/grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaMessageDeframer.java
@@ -141,7 +141,7 @@ public class ArmeriaMessageDeframer implements AutoCloseable {
 
             final DeframedMessage that = (DeframedMessage) o;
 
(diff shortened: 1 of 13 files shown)


grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/ArmeriaMessageDeframer.java
```diff
@@ -83,32 +83,37 @@
 
     private static final int HEADER_LENGTH = 5;
     private static final int COMPRESSED_FLAG_MASK = 1;
-    private static final int RESERVED_MASK = 0xFE;
+    private static final int RESERVED_MASK = 0x7E;
+    // Valid type always has 0s for middle bits.
+    private static final int UNINITIALIED_TYPE = 0xFF;
```
trustin: Maybe use `-1` which is even more impossible?

anuraaga: Isn't -1 FF? Is it clearer using the decimal?

trustin: You are using `ByteBuf.readUnsignedByte()` which never returns a negative value, so `0xFF` and `-1` are different. :wink:

anuraaga: Ah thanks that is better then!

grpc/src/main/java/com/linecorp/armeria/internal/grpc/GrpcStatus.java
```diff
@@ -206,6 +217,51 @@ public static ThrowableProto serializeThrowable(Throwable t) {
...
+        final String grpcMessage = headers.get(GrpcHeaderNames.GRPC_MESSAGE);
+        if (grpcMessage != null) {
+            status = status.withDescription(StatusMessageEscaper.unescape(grpcMessage));
+        }
+        final String grpcThrowable = headers.get(GrpcHeaderNames.ARMERIA_GRPC_THROWABLEPROTO_BIN);
+        if (grpcThrowable != null) {
+            status = addCause(status, grpcThrowable);
+        }
+
+        Metadata metadata = MetadataUtil.copyFromHeaders(headers);
```
minwoox: nit: final

trustin:
Let me merge if nits are squashed.

trustin:
Thanks, @anuraaga !

PR #2077, opened 2019-09-13
Refactors VariableList to simplify code and use hooks.

anuraaga:
When rewriting the docs client to use hooks, much of it was straight-forward, except for `VariableList`. The state management, with children having state managed by parents and propagating changes up, did not follow the model very well. It was also confusing for me to reason about. I think the biggest reason to have parents manage state is so when children are not rendered, the state stays - using CSS visibility instead of conditional rendering of children removes this problem and allows state to stay close to who owns it, the bean itself, and we don't need a global, indexed registry of bean state. If these were extremely huge amounts of DOM we may prefer conditional rendering, but these shouldn't see performance issues using CSS instead.

I also removed `hasLocation` prop since it can be derived from the variables list, which was simpler for me to reason about.

I'll send a follow-up PR with the rest of the migration, the rest is mechanical so wanted to separate out this more significant change.


docs-client/src/components/VariableList/index.tsx
```diff
@@ -83,168 +90,136 @@ export default function({
...
+  indent: number;
+  variable: Variable;
+  specification: Specification;
+  title: string;
+}
 
-  private static indentString(indent: number, s: string): string {
-    return `${'\xa0'.repeat(indent)}${s}`;
-  }
+const indentString = (indent: number, s: string): string => {
```
trustin: Question: Why are we not using `function` here?

docs-client/src/components/VariableList/index.tsx
```diff
@@ -83,168 +90,136 @@ export default function({
...
+        />
+      )}
+    </>
+  );
+};
+
+const FieldInfos: React.FunctionComponent<FieldInfosProps> = (props) => {
+  const styles = useStyles();
+
+  const isEmpty = props.variables.length === 0;
```
trustin: Question: Here, we are assuming the variable list won't change in runtime, right?

docs-client/src/components/VariableList/index.tsx
anuraaga: I've noticed most code using lambdas everywhere now. Since we don't support old browsers anymore there's no performance difference (not transpiled) and I guess it's nice to just not think about whether it's safe to use function or not and consistently use the safer lambda

docs-client/src/components/VariableList/index.tsx
anuraaga: Not really - if the specification is refetched and something changes, all the components will rerender to reflect the change when props change.

trustin: Got it. Thanks!

docs-client/src/components/VariableList/index.tsx
trustin: Thanks for the answer.

docs-client/src/components/VariableList/index.tsx
```diff
@@ -83,168 +90,136 @@ export default function({
...
-    const lowerCase = s.toLowerCase();
-    if ('unspecified' === lowerCase) {
-      return '-';
-    }
+  return lowerCase;
+};
 
-    return lowerCase;
-  }
+const formatLocation = (s: string): string => {
```
ikhoon: nit: After this function is migrated to lambda, I think this works for this (alias|wrap) case. 😉
``` ts
const formatLocation = formatRequirement
```

trustin:
@anuraaga Thanks! Looking forward to your hooks migration PR. :wink:

PR #2081, opened 2019-09-17
Support Spring WebFlux's changed body handling logic in Armeria's WebClient

### LINKED ISSUE #2080: Spring WebFlux client isn't working with body extraction methods
5.1.3.RELEASE or later, Spring WebFlux's body handling logic has changed when response has an error.

See below
https://github.com/spring-projects/spring-framework/blob/c187cb2fa13af2a6ff6e92d588ba70b458707460/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L443-L446

This changed logic handles body twice if status code of the response determines that there is an error. But `ArmeriaHttpClientResponseSubscriber` doesn't remember that the response body is canceled or completed. Therefore erroneous response using methods like `bodyToMono()` doesn't end.

socar-brad:
**Motivation:**

Spring WebFlux's body handling logic is changed. This changed logic handles body twice if response has an error. `ArmeriaHttpClientResponseSubscriber` doesn't remember that the response is completed or canceled. Then Armeria's WebClient using body handling methods doesn't work if response has an error.

**Modifications:**
 * Use EmitterProcessor remember if the response is completed or canceled.

**Result:**
 * Fix #2080

### DIFF
--- a/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
+++ b/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
@@ -18,13 +18,11 @@ package com.linecorp.armeria.spring.web.reactive;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.BiFunction;
 
-import javax.annotation.Nullable;
-
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
-import org.reactivestreams.Subscription;
 
 import com.linecorp.armeria.common.CommonPools;
+import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.HttpObject;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
@@ -34,6 +32,8 @@ import com.linecorp.armeria.common.stream.AbortedStreamException;
 import com.linecorp.armeria.common.stream.CancelledSubscriptionException;
 
 import io.netty.channel.EventLoop;
+import reactor.core.publisher.EmitterProcessor;
+import reactor.core.scheduler.Schedulers;
 
 /**
  * A {@link Subscriber} which reads the {@link ResponseHeaders} first from an {@link HttpResponse}.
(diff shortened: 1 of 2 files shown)


trustin:
Does EmitterProcessor remember the whole stream or only whether conpleted/errored? We do not want to keep the whole stream which may be very large.

socar-brad:
`EmitterProcessor`'s bufferSize  in `ArmeriaHttpClientResponseSubscriber` is 1 currently.
And `EmitterProcessor`'s buffer will be empty after processor's subscribers consume all items.
Also it remembers whether completed or errored after stream is completed or errored.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
```diff
@@ -42,78 +42,32 @@
...
+        final EmitterProcessor<HttpObject> httpObjectProcessor = EmitterProcessor.create(1);
+        httpObjectProcessor
+                .ofType(HttpData.class)
+                .subscribe(httpDataProcessor);
 
-    private boolean isCompleted;
-    @Nullable
-    private Throwable completedCause;
+        httpObjectProcessor
+                .ofType(ResponseHeaders.class)
```
minwoox: question: does this `httpObjectProcessor` relays the trailers which is `HttpHeaders` as well?

trustin:
Let me look into this issue in detail today because I'm curious if we can fix this with less changes rather than pulling in `EmitterProcessor` which seems somewhat heavyweight.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
socar-brad: `ArmeriaHttpClientResponseSubscriber`'s procedure is as follows.

1. `ArmeriaHttpClientResponseSubscriber` subscribes `HttpResponse`.
2. Receive an `HttpHeaders` from response of `1`
3. `Subscriber` completes `httpHeadersFuture`.
4. `ArmeriaClientHttpResponse` subscribes `ResponseBodyPublisher` to create a repsonse body flux.

If `ArmeriaHttpClientResponseSubscriber` doesn't use `EmitterProcessor` for HttpData, at the moment of third step all httpObjectProcessor's subscribers cancel their subscirptions. httpObjectProcessor also cancels it's subscription from `HttpResponse` and remembers itself completed. After that, `ArmeriaClientHttpResponse` subscribes `ResponseBodyPublisher` using `httpObjectProcessor`. It will not receive HttpData, and it will receive an only completed event.

If `ArmeriaHttpClientResponseSubscriber` uses `EmitterProcessor` for HttpData, at the moment of third step subscription of header is canceled. But subscription of HttpData still remains. And `httpObjectProcessor` also remains it's subscription. Therefore it can send all of httpData to `ArmeriaClientHttpResponse`.

trustin:
Created a fix that does not require `EmitterProcessor`: https://github.com/line/armeria/pull/2087

trustin:
@socar-brad By the way, are you a SOCAR/VCNC employee? Are you using Armeria in production?

socar-brad:
@trustin Yes. I'm working at SOCAR. We use Armeria in production to communicate with internal/external services. Thank you for make it open source.

trustin:
@socar-brad That's an awesome news. I hope you have a chance to join our Slack workspace and share your experience later. :cocktail:

trustin:
Let me close this in favor of #2087. Thank you again for reporting this problem and I hope you like the new fix.

PR #2086, opened 2019-09-18
Use negotiation cache to check if an endpoint does not support …

minwoox:
…n event loop scheduler

Motivation:
A user can set the default maximum number of event loops to handle HTTP/2 and HTTP/1 connections per endpoint. If he/she creates a client specifying `h1x` and `h2x`, we obviously know which HTTP version is used. However, if `http` or `https` is used, we do not know the specific version so it's hard to figure out how many event loops we need to use.

If we use the information from `SessionProtocolNegotiationCache` from the previous connection, we know that the endpoint supports the specific HTTP version, so we can use that.

Modifications:
- Use `SessionProtocolNegotiationCache` to find the endpoint supports HTTP/2
  - Use the information to assign the maximum number of event loops for the endpoint

Result:
- `DefaultEventLoopScheduler` respects `maxNumEventLoopsPerHttp(1|2)Endpoint` set from `ClientFactoryBuilder`.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/client/SessionProtocolNegotiationCacheTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/SessionProtocolNegotiationCacheTest.java
@@ -24,11 +24,19 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import io.netty.util.NetUtil;
 
 class SessionProtocolNegotiationCacheTest {
+
+    @BeforeEach
+    @AfterEach
+    void clearCache() {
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/linecorp/armeria/client/SessionProtocolNegotiationCacheTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+
+import org.junit.jupiter.api.Test;
+
+import io.netty.util.NetUtil;
+
+class SessionProtocolNegotiationCacheTest {
```
trustin: Could you clean the cache after the test is done, so that other tests are not affected?

minwoox: Oops, I missed that. Thanks!

minwoox:
Thanks for reviewing! 😉

PR #2087, opened 2019-09-18
Allow Spring WebFlux `DefaultWebClient` to subscribe to respons…

### LINKED ISSUE #2080: Spring WebFlux client isn't working with body extraction methods

trustin:
… many times

Inspired by @socar-brad's work: #2080 and #2081
Test case written by @socar-brad

Motivation:

Previously, Spring WebFlux `DefaultWebClient` subscribed only once to
the response body stream. However, since 5.1.3, it may subscribe more
than once:

- https://github.com/spring-projects/spring-framework/blob/c187cb2fa13af2a6ff6e92d588ba70b458707460/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L443-L446

Modifications:

- Improve `ArmetiaHttpClientResponseSubscriber` so that the caller can
  subscribe to its `ResponseBodyPublisher` multiple times, although only
  the first `Subscriber` will get the actual content.
  - However, this limitation is just enough because what
    `DefaultWebClient` tries to do is just draining the body stream.
- Clean-up
  - Use `SubscriptionOption` to simplify the completion logic.

Result:

- Fixes #2080

### DIFF
--- a/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
+++ b/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
@@ -15,8 +15,10 @@
  */
 package com.linecorp.armeria.spring.web.reactive;
 
+import static com.google.common.base.Preconditions.checkState;
+
 import java.util.concurrent.CompletableFuture;
-import java.util.function.BiFunction;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 import javax.annotation.Nullable;
 
@@ -32,6 +34,7 @@ import com.linecorp.armeria.common.HttpStatusClass;
 import com.linecorp.armeria.common.ResponseHeaders;
 import com.linecorp.armeria.common.stream.AbortedStreamException;
 import com.linecorp.armeria.common.stream.CancelledSubscriptionException;
+import com.linecorp.armeria.common.stream.SubscriptionOption;
 
 import io.netty.channel.EventLoop;
 
@@ -41,29 +44,27 @@ import io.netty.channel.EventLoop;
  * After that, it can act as a {@link Publisher} on behalf of the {@link HttpResponse},
  * by calling {@link #toResponseBodyPublisher()} which returns {@link ResponseBodyPublisher}.
  */
-final class ArmeriaHttpClientResponseSubscriber
-        implements Subscriber<HttpObject>, BiFunction<Void, Throwable, Void> {
+final class ArmeriaHttpClientResponseSubscriber implements Subscriber<HttpObject> {
 
-    private final CompletableFuture<ResponseHeaders> future = new CompletableFuture<>();
+    private final CompletableFuture<ResponseHeaders> headersFuture = new CompletableFuture<>();
     private final EventLoop eventLoop = CommonPools.workerGroup().next();
 
-    @Nullable
-    private ResponseBodyPublisher publisher;
+    private final ResponseBodyPublisher bodyPublisher = new ResponseBodyPublisher(this);
 
     @Nullable
-    private Subscription subscription;
+    private volatile Subscription subscription;
 
     private boolean isCompleted;
+
     @Nullable
     private Throwable completedCause;
 
     ArmeriaHttpClientResponseSubscriber(HttpResponse httpResponse) {
-        httpResponse.completionFuture().handle(this);
-        httpResponse.subscribe(this, eventLoop);
+        httpResponse.subscribe(this, eventLoop, SubscriptionOption.NOTIFY_CANCELLATION);
     }
 
-    CompletableFuture<ResponseHeaders> httpHeadersFuture() {
-        return future;
+    CompletableFuture<ResponseHeaders> headersFuture() {
+        return headersFuture;
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


trustin:
@socar-brad, PTAL.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
```diff
@@ -41,29 +44,27 @@
  * After that, it can act as a {@link Publisher} on behalf of the {@link HttpResponse},
  * by calling {@link #toResponseBodyPublisher()} which returns {@link ResponseBodyPublisher}.
  */
-final class ArmeriaHttpClientResponseSubscriber
-        implements Subscriber<HttpObject>, BiFunction<Void, Throwable, Void> {
```
minwoox: Nice to remove `BiFunction`. 😄

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
```diff
@@ -74,130 +75,187 @@ public void onSubscribe(Subscription s) {
...
+                } else {
+                    // The second subscriber.
+                    final Subscriber<HttpObject> newSubscriber = new CompositeSubscriber<>(oldSubscriber, s);
+                    if (subscriberUpdater.compareAndSet(this, oldSubscriber, newSubscriber)) {
+                        break;
+                    }
+                }
+            }
+
             s.onSubscribe(this);
```
minwoox: So the `subscription.request(n)` will be called by the second subscribers as well which might give more demand that the main subscriber requested. Is it okay?

trustin: Not okay. :sweat_smile: Fixed by using `Mono` of `HttpResponse.completionFuture()` for the late subscribers.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
```diff
@@ -74,130 +81,146 @@ public void onSubscribe(Subscription s) {
...
+                }
+            }
         }
 
         @Override
         public void request(long n) {
-            if (!isCompleted) {
-                upstreamSubscriber.ensureSubscribed().request(n);
-                return;
+            if (parent.completedCause == null) {
```
minwoox: Is there any chance that this method is called from another thread? If so, I think we can make this volatile.

trustin: Fixed by making it volatile

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
```diff
@@ -74,130 +81,146 @@ public void onSubscribe(Subscription s) {
...
-            publisher.relayOnComplete();
-        } else {
-            complete(null);
-        }
-    }
-
-    private void complete(@Nullable Throwable cause) {
-        isCompleted = true;
-        completedCause = cause;
+        complete(null);
```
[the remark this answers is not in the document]
trustin: Good catch!

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java
```diff
@@ -32,38 +34,42 @@
...
 import com.linecorp.armeria.common.stream.AbortedStreamException;
 import com.linecorp.armeria.common.stream.CancelledSubscriptionException;
+import com.linecorp.armeria.common.stream.SubscriptionOption;
 
 import io.netty.channel.EventLoop;
+import reactor.core.publisher.Mono;
 
 /**
  * A {@link Subscriber} which reads the {@link ResponseHeaders} first from an {@link HttpResponse}.
  * If the {@link ResponseHeaders} is consumed, it completes the {@code future} with the {@link ResponseHeaders}.
```
ikhoon: nit: `the {@code future}` -> `the {@code headersFuture}`?

trustin: Fixed by using `{@link #headersFuture}`.

PR #2091, opened 2019-09-19
Add `RequestContext.uri()` and `ClientRequestContext.scheme()`

### LINKED ISSUE #2089: Make it possible to decorate clients and servers at URL context
Currently, the parts needed to resolve an http URL come from different objects at different availabilities, notably not yet accessible when constructing a decorating service. This makes it complicated, and difficult to do things such as request sampling.

For example, to get a client URL requires the following logic and availabilities:
* RequestLogAvailability.SCHEME
* RequestLogAvailability.REQUEST_HEADERS

```java
    public static String generateUrl(RequestLog requestLog) {
        if (!requestLog.isAvailable(RequestLogAvailability.SCHEME) ||
            !requestLog.isAvailable(RequestLogAvailability.REQUEST_HEADERS)) {
            return null;
        }
        final Scheme scheme = requestLog.scheme();
        final String authority = requestLog.authority();
        final String path = requestLog.path();
        final String query = requestLog.query();
        final SessionProtocol sessionProtocol = scheme.sessionProtocol();
        final String uriScheme;
        if (SessionProtocol.httpValues().contains(sessionProtocol)) {
            uriScheme = "http://";
        } else if (SessionProtocol.httpsValues().contains(sessionProtocol)) {
            uriScheme = "https://";
        } else {
            uriScheme = sessionProtocol.uriText() + "://";
        }

        final StringBuilder uriBuilder = new StringBuilder(
                uriScheme.length() + authority.length() + path.length() +
                (query != null ? query.length() + 1 : 0));

        uriBuilder.append(uriScheme)
                  .append(authority)
                  .append(path);

        if (query != null) {
            uriBuilder.append('?').append(query);
        }

        return uriBuilder.toString();
    }
```

I think decoration at a resolved http request abstraction is a better fit for things like metrics and tracing, and also notably can avoid accidental retries inside a decorator. For example, in OkHttp, there is a separate concept of network interceptor from application, where the former has access to network details more fundamentally required.

trustin:
Motivation:

There is currently no simple way to retrieve the URI of the current
request.

Modifications:

- Add `RequestContext.uri()`
- Add `ClientRequestContext.scheme()` and `serializationFormat()`
- Replace `UserClient.sessionProtocol()` with `scheme()`.
- (Breaking) Push down `serializationFormat` from `NonWrappingRequestContext`
  to `DefaultClient/ServiceRequestContext`
  - You might be affected by this if you were extending it directly.
    Please consider migrating to `Client/ServiceRequestContextBuilder`.
- Remove `SpanTags.generateUrl()`

Result:

- Fixes #2089
- It's now much easier to get the URI of the current request.

### DIFF
--- a/brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
+++ b/brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
@@ -24,7 +24,6 @@ import com.linecorp.armeria.common.SerializationFormat;
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.common.logging.RequestLogAvailability;
-import com.linecorp.armeria.internal.brave.SpanTags;
 
 import brave.http.HttpClientAdapter;
 
@@ -59,7 +58,7 @@ final class ArmeriaHttpClientAdapter extends HttpClientAdapter<RequestLog, Reque
     @Override
     @Nullable
     public String url(RequestLog requestLog) {
-        return SpanTags.generateUrl(requestLog);
+        return requestLog.context().uri(true).toString();
     }
 
     @Override
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
@@ -19,6 +19,8 @@ package com.linecorp.armeria.client;
 import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.time.Duration;
 import java.util.Iterator;
 import java.util.Map.Entry;
@@ -39,6 +41,8 @@ import com.linecorp.armeria.common.HttpHeadersBuilder;
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.NonWrappingRequestContext;
 import com.linecorp.armeria.common.Request;
+import com.linecorp.armeria.common.Scheme;
+import com.linecorp.armeria.common.SerializationFormat;
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.logging.DefaultRequestLog;
 import com.linecorp.armeria.common.logging.RequestLog;
@@ -70,6 +74,7 @@ public class DefaultClientRequestContext extends NonWrappingRequestContext imple
     @Nullable
     private EventLoop eventLoop;
     private final ClientOptions options;
+    private final Scheme scheme;
     @Nullable
     private EndpointSelector endpointSelector;
     @Nullable
@@ -94,14 +99,49 @@ public class DefaultClientRequestContext extends NonWrappingRequestContext imple
      * the construction of this context.
      *
      * @param eventLoop the {@link EventLoop} associated with this context
-     * @param sessionProtocol the {@link SessionProtocol} of the invocation
+     * @param scheme the {@link Scheme} of the invocation
      * @param request the request associated with this context
      */
+    public DefaultClientRequestContext(
+            EventLoop eventLoop, MeterRegistry meterRegistry, Scheme scheme,
+            HttpMethod method, String path, @Nullable String query, @Nullable String fragment,
+            ClientOptions options, Request request) {
+        this(null, requireNonNull(eventLoop, "eventLoop"), meterRegistry, scheme,
+             method, path, query, fragment, options, request);
+    }
+
+    /**
+     * Creates a new instance. Note that {@link #init(Endpoint)} method must be invoked to finish
+     * the construction of this context.
+     *
+     * @param eventLoop the {@link EventLoop} associated with this context
+     * @param sessionProtocol the {@link SessionProtocol} of the invocation
+     * @param request the request associated with this context
+     *
+     * @deprecated Use {@link #DefaultClientRequestContext(EventLoop, MeterRegistry, Scheme, HttpMethod, String, String, String, ClientOptions, Request)}.
+     */
+    @Deprecated
     public DefaultClientRequestContext(
             EventLoop eventLoop, MeterRegistry meterRegistry, SessionProtocol sessionProtocol,
             HttpMethod method, String path, @Nullable String query, @Nullable String fragment,
             ClientOptions options, Request request) {
-        this(null, requireNonNull(eventLoop, "eventLoop"), meterRegistry, sessionProtocol,
+        this(eventLoop, meterRegistry, Scheme.of(SerializationFormat.NONE, sessionProtocol),
+             method, path, query, fragment, options, request);
+    }
+
+    /**
+     * Creates a new instance. Note that {@link #init(Endpoint)} method must be invoked to finish
+     * the construction of this context.
+     *
+     * @param factory the {@link ClientFactory} which is used to acquire an {@link EventLoop}
+     * @param scheme the {@link Scheme} of the invocation
+     * @param request the request associated with this context
+     */
+    public DefaultClientRequestContext(
+            ClientFactory factory, MeterRegistry meterRegistry, Scheme scheme,
+            HttpMethod method, String path, @Nullable String query, @Nullable String fragment,
+            ClientOptions options, Request request) {
+        this(requireNonNull(factory, "factory"), null, meterRegistry, scheme,
              method, path, query, fragment, options, request);
     }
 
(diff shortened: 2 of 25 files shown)


trustin:
@adriancole I guess you'll love this. Please review! :stuck_out_tongue_winking_eye:

trustin:
@jwills You might also wanna take a quick look. :wink:

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientAdapter.java
```diff
@@ -24,7 +24,6 @@
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.common.logging.RequestLogAvailability;
-import com.linecorp.armeria.internal.brave.SpanTags;
```
adriancole: this line I love

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -112,24 +152,28 @@ public DefaultClientRequestContext(
      * @param factory the {@link ClientFactory} which is used to acquire an {@link EventLoop}
      * @param sessionProtocol the {@link SessionProtocol} of the invocation
      * @param request the request associated with this context
+     *
+     * @deprecated Use {@link #DefaultClientRequestContext(ClientFactory, MeterRegistry, Scheme, HttpMethod, String, String, String, ClientOptions, Request)}.
```
adriancole: is it safe to assume internally, armeria makes no calls to this ctor anymore?

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -130,6 +128,33 @@ public final SessionProtocol sessionProtocol() {
...
+                cachedUri = uncachedUri(false);
+            }
+            return cachedUri;
+        }
+    }
+
+    /**
+     * Generates the {@link URI} of this request.
+     */
+    protected URI uncachedUri(boolean simplifyScheme) {
```
adriancole: maybe supplyUrl? I get that it is uncached. I scanned this PR and thought.. what?

then later I thought about supply. Basically this is a supply function used by uri, which is memoized.

protected is the right access for this.

trustin:
It's always available, regardless of the availability. Let me fix the test failures a bit later.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: Yes. I made sure nobody in our code calls this.

trustin: This ctor = the deprecated ctor, right?

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
trustin: Maybe just generateUri?

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -157,6 +161,62 @@
...
+        buf.append(firstNonNull(authority, "UNKNOWN"));
+        buf.append(path());
+
+        final String query = query();
+        if (query != null) {
+            buf.append('?').append(query);
+        }
+
+        try {
+            return new URI(buf.toString());
```
adriancole: not related to this change, and can be done externally, but seems an opportunity to use the multi-arity constructor vs forcing a parse?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
adriancole: yep

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
adriancole: WFM

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
trustin: If you look into the JDK source code, all `URI` constructors parse anyway. :scream:

adriancole: 😱 is right!

anuraaga:
It might also be a good time to add `httpRequest()` to `RequestContext` - that way decorators can always access the HTTP request simply without worrying about `RpcRequest`. I don't think `SpanTags.generateUrl` would have any timing issues if it used `HttpRequest` instead of `RequestLog`.

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -157,6 +161,62 @@
...
+     * @see #uri(boolean)
+     */
+    default URI uri() {
+        return uri(false);
+    }
+
+    /**
+     * Returns the {@link URI} of the current {@link Request}.
+     *
+     * @param simplifyScheme If {@code true}, the scheme of the returned {@link URI} will always be
```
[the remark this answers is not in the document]
trustin: I'm not sure I like it either, but this was introduced for the users who prefers using http/https over h1, h2, ... (Brave to name one :smile:) If we did not have this flag, a user will have to reconstruct the URI just to replace the scheme.

trustin:
@anuraaga wrote:

> I'm personally not a huge fan of the overall direction. It's very hard to reason about whether these values are user-provided, or connection-established, with similar but probably different fields in various places like `RequestContext`, `Request`, and `RequestLog`.
>
> One example of confusion is it's not clear to me why this new method uses authority instead of remote address. They're both `uri`s and this one seems to just happen to use authority but I'm not sure why.
>
> I think I would put methods like added here in `Request` somehow, not `RequestContext`, since that is closer to what the user requested, and try to have the latter reflect the actual state of a connection as much as possible (similar to the other `@Nullable` fields it has already).

I like the idea of keeping them in `Request`.

> It might also be worth considering holding off on this until after making the change to move connection establishment before creation of the context, as it could simplify things.

@minwoox actually tried moving it front but it didn't fit very well with some decorators such as `RetryingClient` and `CircuitBreakerClient` and we finally gave up on that direction (until we have a better idea for doing it).

> It might also be a good time to add `httpRequest()` to `RequestContext` - that way decorators can always access the HTTP request simply without worrying about `RpcRequest`. I don't think `SpanTags.generateUrl` would have any timing issues if it used `HttpRequest` instead of `RequestLog`.

That's a good point. Let me try to explore in this direction. Thanks for a suggestion!

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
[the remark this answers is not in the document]
anuraaga: I think that came up for brave when we noticed the scheme was h2c etc even though the user used http:// when defining the client (actually even more weird like none+h2c) since we were using established, not requested, scheme before. 

I feel as if the user requested h2:// it would be ok to use it should be used in instrumentation too since it's an advanced use case but explicit by the user. @adriancole does that make sense?

minwoox:
> @minwoox actually tried moving it front but it didn't fit very well with some decorators such as RetryingClient and CircuitBreakerClient and we finally gave up on that direction (until we have a better idea for doing it).

Yes, I have stopped to move it front. I should've thought more carefully before I bull ahead.
I could handle the `RetryingClient` by making the connection in two different places which are `UserClient` (that is before decorator) and `HttpClientDelegate` (that is after decorator).

But the real problem was `CircuitBreakerClient`. Because it's a decorator, if the circuit is open, we must not send a request and not try to establish a connection. So if the connection attempt is made before the request goes to the decorator, the `UserClient` must be aware of the state of the circuit. And we cannot store that information in `ClientFactory` which has the connection pools currently, because it will affect the other clients which are created from the `ClientFactory`. If we use another way, that would make the code really complicated.

So I gave up right now, but I'm still trying to find a way for that and I hope I could do it one day. :-)

anuraaga:
@minwoox Thanks I didn't know we found issues and stopped it but now I do :) Just food for thought is, as in https://github.com/line/armeria/pull/2063 we might find things that we've implemented as decorator so far but doesn't actually fit the model so well. We might feel like controlling whether connection happens or not is too low level for a decorator and should be implemented in a different way, or maybe it's still the best way.

minwoox:
@anuraaga Yes, agreed. The decorators should not deal with connections. And also agreed that some of the decorators may not fit the model so well. We might need another way for them or this might be the best way. Or just too immature to make a judgment? 😅

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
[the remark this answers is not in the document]
adriancole: it totally does not make sense, as no one would expect the tag http.uri to have custom schemes.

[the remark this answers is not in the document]
adriancole: someones when someone is asking for something obscure, it is good to do a poll and look for other telemetry systems. for example, how many expect or are able to parse successfully odd things like h2://

I think it would be a bug making machine to propagate this upstream iotw and it wouldn't take much querying to find majority would agree.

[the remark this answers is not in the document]
trustin: See #2092. I'm closing this.

trustin:
#2092 supersedes this pull request.

PR #2092, opened 2019-09-19
Add `RequestHeaders.uri()` and `HttpRequest.uri()`

trustin:
Motivation:

A user sometimes wants to get the request URI, without having to
construct it by him/herself.

Modifications:

- Make sure to auto-fill the `:scheme` and `:authority` headers in the
  beginning of the decorator chain.
- Add `RequestHeaders.uri()` and `HttpRequest.uri()` which generates the
  request URI from three headers - `:scheme`, `:authority` and `:path`.
- This will not work if a user removes `:scheme` or `:authority`
  headers, but I think it's acceptable to raise an exception so that the
  request fails somewhere in the decorator chain.

Result:

A user can get the request URI easily.


trustin:
Forgot to set the draft status to this pull request. Please keep in mind that this has failing tests.

trustin:
It needs more work, so it's too soon to approve. :wink:

core/src/test/java/com/linecorp/armeria/client/HttpClientSniTest.java
```diff
@@ -97,53 +95,55 @@ public static void destroy() throws Exception {
...
+        Assertions.assertEquals(HttpStatus.OK, response.status());
         return response.contentUtf8();
     }
 
     @Test
-    public void testCustomAuthority() throws Exception {
-        final HttpClient client = HttpClient.of(clientFactory, "https://127.0.0.1:" + httpsPort);
-        final AggregatedHttpResponse response =
-                client.execute(RequestHeaders.of(HttpMethod.GET, "/",
-                                                 HttpHeaderNames.AUTHORITY, "a.com:" + httpsPort))
```
trustin: `:authority` header will always be automatically set based on the specified or selected `Endpoint`, so you cannot set an `:authority` header manually anymore, although you can still send the `:authority` header you desire, by specifying an `Endpoint` explicitly.

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientAuthorityHeaderTest.java
```diff
@@ -81,18 +77,8 @@ public void authorityIsDifferentByBackendsWhenRetry() {
...
     }
 
-    @Test
-    public void authorityIsSameWhenSet() {
-        final HttpClient client = newHttpClientWithEndpointGroup();
-
-        final RequestHeaders headers = RequestHeaders.of(GET, "/",
-                                                         HttpHeaderNames.AUTHORITY, "www.armeria.com");
-        final AggregatedHttpResponse res = client.execute(headers).aggregate().join();
-        assertThat(res.contentUtf8()).contains("www.armeria.com");
```
trustin: Again, this doesn't work anymore because Armeria will auto-set the `:authority` header. I think it's not a problem because a user can specify the same host name for the `Endpoint`s in his or her `EndpointGroup`.

thrift/src/main/java/com/linecorp/armeria/server/thrift/THttpService.java
```diff
@@ -449,12 +449,20 @@ private SerializationFormat determineSerializationFormat(HttpRequest req) {
         if (contentType != null) {
             serializationFormat = findSerializationFormat(contentType);
             if (serializationFormat == null) {
-                return null;
+                // Browser clients often send a non-Thrift content type.
+                // Choose the default serialization format for some vague media types.
+                if (!("text".equals(contentType.type()) &&
+                      "plain".equals(contentType.subtype())) &&
+                    !("application".equals(contentType.type()) &&
+                      "octet-stream".equals(contentType.subtype()))) {
```
trustin: This pull request revealed a bug which made the `contentTypeNotThrift()` test does not test anything.

trustin:
I think it's ready now for reviews.

grpc/src/test/java/com/linecorp/armeria/server/grpc/GrpcServiceServerTest.java
```diff
@@ -204,12 +204,6 @@ public void errorWithMessage(SimpleRequest request, StreamObserver<SimpleRespons
             metadata.put(CUSTOM_VALUE_KEY, "custom value");
 
             final ServiceRequestContext ctx = ServiceRequestContext.current();
-            // Metadata takes priority, this trailer will not be written since it has the same name.
-            ctx.addAdditionalResponseTrailer(
```
trustin: `additionalResponse{Headers,Trailers}` now takes priority.

trustin:
Note that it is still not possible to remove the `RequestLogAvaiability` check in `SpanTags.generateUrl()`. I'm going to address this problem in a follow-up PR which will allow a user access `HttpRequest` from `ClientRequestContext` even for RPC requests.

trustin:
@imasahiro and @kojilin, could you also take a look into this change, which may be breaking in a very specific case where a user attempts to override `:authority` header?

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientAuthorityHeaderTest.java
```diff
@@ -81,18 +77,8 @@ public void authorityIsDifferentByBackendsWhenRetry() {
...
-    }
-
     private static HttpClient newHttpClientWithEndpointGroup() {
-        final EndpointGroup endpointGroup = new StaticEndpointGroup(
+        final EndpointGroup endpointGroup = EndpointGroup.of(
```
ikhoon: 👍

brave/src/main/java/com/linecorp/armeria/internal/brave/SpanTags.java
```diff
@@ -102,37 +101,11 @@ public static void addTags(Span span, RequestLog log) {
...
-        uriBuilder.append(uriScheme)
-                  .append(authority)
-                  .append(path);
-
-        if (query != null) {
-            uriBuilder.append('?').append(query);
-        }
 
-        return uriBuilder.toString();
+        return requestLog.requestHeaders().uri().toString();
```
ikhoon: Nice!👍

core/src/main/java/com/linecorp/armeria/common/RequestHeadersBuilder.java
```diff
@@ -39,20 +44,54 @@
...
+            case H2:
+            case H1:
+                scheme("https");
+                break;
+            case HTTP:
+            case H2C:
+            case H1C:
+                scheme("http");
+                break;
+            case PROXY:
```
adriancole: protocol isn't a scheme, and 'proxy' is obscure (FoxyProxy). We should map this to the semantic http scheme I think, right? https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

trustin: Aye, I had to throw IAE here. :-)

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -182,43 +185,70 @@ public boolean init(Endpoint endpoint) {
...
             ((StreamMessage<?>) req).abort();
         }
     }
 
+    private void autoFillSchemeAndAuthority() {
+        final Request req = request();
+        if (req instanceof HttpRequest) {
+            final HttpRequest httpReq = (HttpRequest) req;
+            final RequestHeaders oldHeaders = httpReq.headers();
+            final RequestHeadersBuilder newHeadersBuilder = oldHeaders.toBuilder();
```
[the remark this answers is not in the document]
trustin: Doh! Fixed :-)

core/src/main/java/com/linecorp/armeria/common/RequestHeadersBuilder.java
```diff
@@ -39,20 +44,54 @@
...
+            case H2:
+            case H1:
+                scheme("https");
+                break;
+            case HTTP:
+            case H2C:
+            case H1C:
+                scheme("http");
+                break;
+            default:
```
adriancole: for a rainy day: test of usage which triggers this (not unit test, like end-to-end).

trustin: Added tests! :rainbow:

trustin: Ah, it's impossible to trigger this in an integration test because ctx.sessioProtocol() never returns .

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -179,7 +179,7 @@ public void onNext(HttpObject o) {
                 final ResponseHeadersBuilder newHeaders = fillAdditionalHeaders(headers, additionalHeaders);
 
                 if (endOfStream && !additionalTrailers.isEmpty()) {
-                    newHeaders.setIfAbsent(additionalTrailers);
+                    newHeaders.set(additionalTrailers);
```
minwoox: What happens if a user sets `:status` using additional headers?
We should allow it? or need to check `status.isContentAlwaysEmpty()` after this?

core/src/main/java/com/linecorp/armeria/common/DefaultResponseHeaders.java
```diff
@@ -17,7 +17,7 @@
 
 import javax.annotation.Nullable;
 
-@SuppressWarnings("checkstyle:EqualsHashCode")
+@SuppressWarnings({ "checkstyle:EqualsHashCode", "EqualsAndHashcode" })
```
minwoox: Could we just leave `"EqualsAndHashcode"`?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -182,43 +184,68 @@ public boolean init(Endpoint endpoint) {
...
     }
 
+    private void autoFillSchemeAndAuthority() {
+        final Request req = request();
+        if (req instanceof HttpRequest) {
+            final HttpRequest httpReq = (HttpRequest) req;
+            final RequestHeaders headers = httpReq.headers();
+            final String authority = endpoint != null ? endpoint.authority() : "UNKNOWN";
+            if (headers.scheme() != null && authority.equals(headers.authority())) {
+                unsafeUpdateRequest(req);
```
minwoox: It seems like the `req` is not changed. Do we need this?

core/src/main/java/com/linecorp/armeria/common/DefaultResponseHeaders.java
trustin: Good to know that `checkstyle:` prefix is optional. :bow:

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -257,6 +284,17 @@ public ClientRequestContext newDerivedContext(Request request, Endpoint endpoint
         return new DefaultClientRequestContext(this, request, endpoint);
     }
 
+    @Override
+    protected void validateHeaders(RequestHeaders headers) {
+        // Do not validate/filter if the context is not fully initialized yet,
+        // because init() will trigger this method again via updateEndpoint().
+        if (eventLoop == null) {
```
minwoox: https://github.com/line/armeria/pull/2092/files#diff-d437ddb5e0b4572dba776949bacf9becR127
`eventLoop` can be passed as one of the constructor arguments. So I think another flag for that. 😉

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
trustin: That's a good point. I guess we need to prohibit setting some pseudo headers.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
```diff
@@ -643,7 +643,7 @@ private static void validateWeight(int weight) {
     }
 
     @Override
-    public boolean equals(Object obj) {
+    public boolean equals(@Nullable Object obj) {
```
minwoox: Question: do we decide to put `@Nullable` here?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -182,43 +184,68 @@ public boolean init(Endpoint endpoint) {
             }
 
             return true;
-        } catch (Exception e) {
+        } catch (Throwable t) {
```
minwoox: Question: Just want to know why you raised the level here. 😃

trustin: Just wanted to make sure `failEarly()` is called at any cases.

core/src/main/java/com/linecorp/armeria/client/Endpoint.java
trustin: Yeah, all `equals()` parameters are `@Nullable`, because `null` is accepted indeed.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: No, we don't. Fixed. :sweat_smile:

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
trustin: Now filters out some blacklisted headers. Thanks!

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: Doh! Fixed by adding `initialized` flag.

core/src/main/java/com/linecorp/armeria/common/DefaultResponseHeaders.java
trustin: It seems like Checkstyle only respects the suppressions that start with `checkstyle:`. Reverted.

minwoox: Sorry about that. 😅

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -57,6 +64,9 @@
     private static final AggregatedHttpResponse SERVICE_UNAVAILABLE_MESSAGE =
             AggregatedHttpResponse.of(HttpStatus.SERVICE_UNAVAILABLE);
 
+    private static final Set<AsciiString> HEADER_BLACKLIST = ImmutableSet.of(
```
minwoox: nit: how about `ADDITIONAL_HEADER_BLACKLIST`?

trustin: Renamed

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceResponseConverterTest.java
```diff
@@ -359,11 +359,11 @@ public Object expectOk() {
                 public void header() {}
 
                 @Get("/header-overwrite")
-                @AdditionalHeader(name = "header_name_1", value = "header_value_unchanged")
+                @AdditionalHeader(name = "header_name_1", value = "header_value_changed")
                 public HttpResponse headerOverwrite() {
                     return HttpResponse.of(ResponseHeaders.of(HttpStatus.OK,
                                                               HttpHeaderNames.of("header_name_1"),
-                                                              "header_value_changed"));
+                                                              "header_value_unchanged"));
```
trustin: Same here - `@AdditionalHeader` now takes priority. /cc @hyangtack

minwoox:
Thanks, @trustin!

trustin:
@imasahiro @kojilin @anuraaga It's been merged already but please feel free to review this PR so I can follow up.

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -737,20 +709,20 @@ void testUpgradeRequestExecutesLogicOnlyOnce() throws Exception {
...
-        assertEquals(HttpStatus.OK, response.status());
+        assertThat(response.status()).isEqualTo(HttpStatus.OK);
 
         clientFactory.close();
     }
 
     @Nested
     @TestInstance(Lifecycle.PER_CLASS)
-    class JettyInterop {
+    static class JettyInterop {
```
anuraaga: Ah didn't know about this :)

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -737,20 +709,20 @@ void testUpgradeRequestExecutesLogicOnlyOnce() throws Exception {
...
 
-        org.eclipse.jetty.server.Server jetty;
+        Server jetty;
```
anuraaga: I had figured someone down the line will get confused when trying to initialize an Armeria `Server` so didn't import.

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
trustin: IntelliJ IDEA inspection FTW :wink:

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
trustin: Ah, OK. Let me revert this.

anuraaga: Thanks - apparently not importing was also confusing so let's add a one-line comment too :)

imasahiro:
lgtm

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
trustin: #2097

adriancole:
great!

PR #2067, opened 2019-09-11
Make `EventLoopThreadFactory` more customizable

jyblue:
Motivation:

A user often need to perform additional work
before or after their `Runnable.run()` is run
which is created by `EventLoopThreadFactory`

Modifications:

Add `EventLoopThreadFactoryBuilder`
Deprecate public constructor of `EventLoopThreadFactory`

Result:

Closes #2047

PR #2095, opened 2019-09-20
Remove redundant `@[Dd]eprecated` inside of deprecated class

jyblue:
Inspired by :

https://github.com/line/armeria/pull/2067#discussion_r325448418

Motivation :

`@Deprecated` annotations and `@deprecated` javadoc sentences on methods
of deprecated classes can be removed.

Modifications :

Remove `@Deprecated` annotations on methods of deprecated classes.
Remove `@deprecated` javadoc sentences on methods of deprecated classes.

Result :

Code clean-up.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/StaticEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/StaticEndpointGroup.java
@@ -40,20 +40,14 @@ public final class StaticEndpointGroup implements EndpointGroup {
 
     /**
      * Creates a new instance.
-     *
-     * @deprecated Use {@link EndpointGroup#of(EndpointGroup...)}.
      */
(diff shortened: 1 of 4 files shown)


jyblue:
> `@Deprecated` annotations and `@deprecated` javadoc sentences on methods
> of deprecated classes can be removed.

I'm not sure that this rule is applicable for previous codes. PTAL :bowing_man:

trustin:
The `@deprecated` Javadoc tags in the methods provide some useful information which is not provided in the class-level Javadoc, so I guess it's better keeping them in this case?

jyblue:
had a chat with @trustin,
I agree that this kind of javadoc might be useful to users.
So, we better keeping them.

Thank you.

trustin:
Thanks for your understanding, @jyblue !

Issue #2031, opened 2019-08-28
AnnotatedHttpService: Allow `@PathPrefix` annotations in class scope


heowc:
I think it would be nice to be able to specify a prefix as an annotation without using pathPrefix.

For example.

```java
ServerBuilder sb = new ServerBuilder();
sb.annotatedService(new ForBarService());
//...
```
```java
@PathPrefix("/foo")
public class FooBarService {
    @Get
    @Path("/bar")
    public Object bar() { ... }

    @Get
    public Object fooGet() { ... }

    @Post
    public Object fooPost() { ... }
} 
```

I think there is an advantage to see the specified path at once.

trustin:
That's a good idea. We currently use `@Path` for other type of path mappings such as glob and regex, so we have two options:

- Restrict the format of `@Path` when it's used at class level.
- Define a different annotation such as `@PathPrefix`.

I prefer the second option, but what do you think? /cc @ikhoon @minwoox

ikhoon:
`@Path` at class level can give a user confused because of the glob. I also lean to `@PathPrefix` l. and we can use it method level too. 😀

minwoox:
+1 for `@PathPrefix` :)

sivaalli:
Hi, Is this something a first timer can work on? If yes, I'm more that happy to look at it

minwoox:
@sivaalli Thanks for your interest!
Yes, I think and I hope it's not that difficult. But I think we need to ask @heowc that he is going to deal with this or not before you take this. 😉

heowc:
@sivaalli Thanks for your opinion!
I will another PRs in progress and someday I will proceed this. 
but, If you need to this, I think it's not bad for you to deal with it. 👍

sivaalli:
Thanks you @minwoox and @heowc for your thoughts. I will work on this. BTW excellent work creating armeria and open sourcing it.

PR #2099, opened 2019-09-20
Adding @PathPrefix annotation

sivaalli:
Adding @PathPrefix annotation for a user to declare a prefix path that a service object handles.
Motivation :  https://github.com/line/armeria/issues/2031

This PR allows a user to declare a 

```
sb.annotatedService(new ForBarService());
```
and 
```
@PathPrefix("/foo")
public class FooBarService {
    @Get("/bar")
    public Object barGet() { ... }
}
```

But this PR does **not** allow something like this:

```
@PathPrefix("/foo")
public class FooBarService {
    @Get //without path
    public Object barGet() { ... }
}
```
Because a path has to be specified either in `@Path` or in a http method annotation(`@Get`) or an exception is thrown. But I think this is nice to have feature. I just want to get feedback on if this is the right direction I'm going and this is what expected. Because the issue did not directly talk about removing path to be specified in `@Path` or http method annotations. Please let me know if you want to see this.
Also, I will be doing documentation changes after this is fully done.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -258,7 +258,8 @@ public final class AnnotatedHttpServiceFactory {
      * Returns an {@link AnnotatedHttpService} instance defined to {@code method} of {@code object} using
      * {@link Path} annotation.
      */
-    private static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Method method,
+    @VisibleForTesting
+    static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Method method,
                                                       List<ExceptionHandlerFunction> baseExceptionHandlers,
                                                       List<RequestConverterFunction> baseRequestConverters,
                                                       List<ResponseConverterFunction> baseResponseConverters) {
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -764,9 +764,9 @@ public final class AnnotatedHttpServiceFactory {
     }
 
     /**
-     * Return the path prefix to use. If there is {@link PathPrefix} annotation on the class
+     * Returns the path prefix to use. If there is {@link PathPrefix} annotation on the class
      * then path prefix is computed by concatenating pathPrefix and value from annotation else
-     * return pathPrefix
+     * returns pathPrefix.
      */
     private static String computePathPrefix(Class<?> clazz, String pathPrefix) {
         ensureAbsolutePath(pathPrefix, "pathPrefix");
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -19,6 +19,8 @@ package com.linecorp.armeria.internal.annotation;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.Sets.toImmutableEnumSet;
+import static com.linecorp.armeria.internal.ArmeriaHttpUtil.concatPaths;
+import static com.linecorp.armeria.internal.RouteUtil.ensureAbsolutePath;
 import static com.linecorp.armeria.internal.annotation.AnnotatedValueResolver.toRequestObjectResolvers;
 import static com.linecorp.armeria.internal.annotation.AnnotationUtil.findAll;
 import static com.linecorp.armeria.internal.annotation.AnnotationUtil.findFirst;
@@ -273,10 +275,10 @@ public final class AnnotatedHttpServiceFactory {
         }
         final Class<?> clazz = object.getClass();
         final String pattern = findPattern(method, methodAnnotations);
-        final String selectedPathPrefix = selectPathPrefix(clazz, pathPrefix);
+        final String computedPathPrefix = computePathPrefix(clazz, pathPrefix);
 
         final Route route = Route.builder()
-                                 .pathWithPrefix(selectedPathPrefix, pattern)
+                                 .pathWithPrefix(computedPathPrefix, pattern)
                                  .methods(methods)
                                  .consumes(consumableMediaTypes(method, clazz))
                                  .produces(producibleMediaTypes(method, clazz))
@@ -763,11 +765,18 @@ public final class AnnotatedHttpServiceFactory {
 
     /**
      * Return the path prefix to use. If there is {@link PathPrefix} annotation on the class
-     * then the value from the annotation is used, else falls back to pathPrefix passed through
-     * method pram.
+     * then path prefix is computed by concatenating pathPrefix and value from annotation else
+     * return pathPrefix
      */
-    private static String selectPathPrefix(Class<?> clazz, String pathPrefix) {
-        return findFirst(clazz, PathPrefix.class).map(PathPrefix::value).orElse(pathPrefix);
+    private static String computePathPrefix(Class<?> clazz, String pathPrefix) {
+        ensureAbsolutePath(pathPrefix, "pathPrefix");
+        final Optional<PathPrefix> pathPrefixAnnotation = findFirst(clazz, PathPrefix.class);
+        if (!pathPrefixAnnotation.isPresent()) {
+            return pathPrefix;
+        }
+        final String pathPrefixFromAnnotation = pathPrefixAnnotation.get().value();
+        ensureAbsolutePath(pathPrefixFromAnnotation, "pathPrefixFromAnnotation");
+        return concatPaths(pathPrefix, pathPrefixFromAnnotation);
     }
 
     private AnnotatedHttpServiceFactory() {}
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -104,6 +104,7 @@ import com.linecorp.armeria.server.annotation.Options;
 import com.linecorp.armeria.server.annotation.Order;
 import com.linecorp.armeria.server.annotation.Patch;
 import com.linecorp.armeria.server.annotation.Path;
+import com.linecorp.armeria.server.annotation.PathPrefix;
 import com.linecorp.armeria.server.annotation.Post;
 import com.linecorp.armeria.server.annotation.ProduceType;
 import com.linecorp.armeria.server.annotation.Produces;
@@ -270,11 +271,12 @@ public final class AnnotatedHttpServiceFactory {
             throw new IllegalArgumentException(method.getDeclaringClass().getName() + '#' + method.getName() +
                                                " must have an HTTP method annotation.");
         }
-
         final Class<?> clazz = object.getClass();
         final String pattern = findPattern(method, methodAnnotations);
+        final String selectedPathPrefix = selectPathPrefix(clazz, pathPrefix);
+
         final Route route = Route.builder()
-                                 .pathWithPrefix(pathPrefix, pattern)
+                                 .pathWithPrefix(selectedPathPrefix, pattern)
                                  .methods(methods)
                                  .consumes(consumableMediaTypes(method, clazz))
                                  .produces(producibleMediaTypes(method, clazz))
@@ -757,6 +759,15 @@ public final class AnnotatedHttpServiceFactory {
             }
         }
         return null;
+    }
+
+    /**
+     * Return the path prefix to use. If there is {@link PathPrefix} annotation on the class
+     * then the value from the annotation is used, else falls back to pathPrefix passed through
+     * method pram.
+     */
+    private static String selectPathPrefix(Class<?> clazz, String pathPrefix) {
+        return findFirst(clazz, PathPrefix.class).map(PathPrefix::value).orElse(pathPrefix);
     }
 
     private AnnotatedHttpServiceFactory() {}
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -260,9 +260,9 @@ public final class AnnotatedHttpServiceFactory {
      */
     @VisibleForTesting
     static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Method method,
-                                                      List<ExceptionHandlerFunction> baseExceptionHandlers,
-                                                      List<RequestConverterFunction> baseRequestConverters,
-                                                      List<ResponseConverterFunction> baseResponseConverters) {
+                                              List<ExceptionHandlerFunction> baseExceptionHandlers,
+                                              List<RequestConverterFunction> baseRequestConverters,
+                                              List<ResponseConverterFunction> baseResponseConverters) {
 
         final Set<Annotation> methodAnnotations = httpMethodAnnotations(method);
         if (methodAnnotations.isEmpty()) {
--- a/core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
@@ -22,6 +22,7 @@ import static com.linecorp.armeria.internal.RouteUtil.EXACT;
 import static com.linecorp.armeria.internal.RouteUtil.GLOB;
 import static com.linecorp.armeria.internal.RouteUtil.PREFIX;
 import static com.linecorp.armeria.internal.RouteUtil.REGEX;
+import static com.linecorp.armeria.internal.RouteUtil.ensureAbsolutePath;
 import static com.linecorp.armeria.server.HttpHeaderUtil.ensureUniqueMediaTypes;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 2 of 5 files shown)


trustin:
@sivaalli, thank you so much for your first contribution! :clap: 

I agree that the following should not be allowed:

```java
@PathPrefix("/foo")
public class FooBarService {
    @Get //without path
    public Object barGet() { ... }
}
```

.. while the following should be allowed:

```java
@PathPrefix("/foo")
public class FooBarService {
    @Get("/bar") // mapped at "/foo/bar"
    public Object barGet() { ... }
}
```

Sounds good?

trustin:
Oops. Updated the above comment.

sivaalli:
Thank you @trustin  I'm really exited to contribute and looking forward to many more contributions as well. Also really appreciate your work on netty and armeria.

In my personal opinion, I can see a use case where a user wants to map a service to a prefix like `@PathPrefix("/users")` and have all the service methods handle different Http methods for ex:

```
    @PathPrefix("/users")
    public lass FooBarService {

        @Post
        public HttpResponse adduser(User user) { ... }

        @Get
        public HttpResponse getAllUsers() { ... }
    }
```
What do you think? Or am I missing something here.

ikhoon:
@sivaalli Thanks for contributing! `@PathPrefix` annotation looks good to me. By the way, are you using Armeria in production? Could you share your experience?😁

sivaalli:
> @sivaalli Thanks for contributing! `@PathPrefix` annotation looks good to me. By the way, are you using Armeria in production? Could you share your experience?😁

I work at Salesforce and a sub-cloud called Social, and we are currently on a monolith stack, but we are in talks of moving from JSON over REST to gRPC. I would definitely mention about Armeria when are are implementing it.

sivaalli:
Hi Again, If this is **not** expected behaviour
```
@PathPrefix("/foo")
public class FooBarService {
    @Get //without path
    public Object barGet() { ... }
}
```
Then I think this PR is ready for review. Please let me know If I can start making Documentation changes or any other changes with respect to this PR. :). Thanks @heowc for letting me contribute to this issue.

heowc:
Very nice! 👍

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -759,6 +761,15 @@ static String findDescription(AnnotatedElement annotatedElement) {
         return null;
     }
 
+    /**
+     * Return the path prefix to use. If there is {@link PathPrefix} annotation on the class
+     * then the value from the annotation is used, else falls back to pathPrefix passed through
+     * method pram.
+     */
+    private static String selectPathPrefix(Class<?> clazz, String pathPrefix) {
+        return findFirst(clazz, PathPrefix.class).map(PathPrefix::value).orElse(pathPrefix);
```
trustin: How about concatenating `pathPrefix` and `PathPrefix.value()`? For example:

```java
sb.annotatedService("/foo", new MyService());

@PathPrefix("/bar")
public class MyService { ... }
```

It seems natural to me that `MyService` is bound under `/foo/bar`. What do you think?

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.linecorp.armeria.server.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for mapping dynamic web requests onto specific class.
```
trustin: Could you rephrase this to reflect what this class does more correctly?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
```diff
@@ -122,6 +131,35 @@ public void ofUserDefinedRepeatableDecorator() throws NoSuchMethodException {
         assertThat(udd2.value()).isEqualTo(2);
     }
 
+    @org.junit.jupiter.api.Test
```
trustin: I wouldn't recommend mixing JUnit 4 and 5 annotations. If you are not going to migrate this test case to JUnit 5, I recommend sticking to JUnit 4.

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
```diff
@@ -122,6 +131,35 @@ public void ofUserDefinedRepeatableDecorator() throws NoSuchMethodException {
...
+                                           .flatMap(Collection::stream)
+                                           .collect(Collectors.toList());
+
+        assertThat(paths).allMatch(path -> path.startsWith(HOME_PATH_PREFIX + '/'));
+    }
+
+    @org.junit.jupiter.api.Test
+    public void testFindAnnotatedServiceElementsWithoutPathPrefixAnnotation() {
+        final Object serviceObject = new ServiceObject();
+        final String root = "/home";
```
trustin: Could be replaced with `HOME_PATH_PREFIX`?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
heowc: I agree. Otherwise, I think it can confuse the user. 👍

sivaalli: Sounds good. Will make those changes. Thanks for review:)

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
sivaalli: Sure. Thanks

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
sivaalli: Sure. Will use junit4 annotations.

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
sivaalli: Sure.

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -173,7 +173,6 @@ public RouteBuilder regex(Pattern regex) {
      * @see #path(String)
      */
     public RouteBuilder pathWithPrefix(String prefix, String pathPattern) {
-        prefix = ensureAbsolutePath(prefix, "prefix");
```
sivaalli: This validation is done in `computePathPrefix` method

trustin: Without this, a user will get an exception later than when he or she called this method. I think it gives a user better feed back when an exception is triggered as soon as a bad input is given.

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
```diff
@@ -0,0 +1,52 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.linecorp.armeria.server.ServerBuilder;
+
+/**
+ * Annotation that can be used on a class as a path prefix for all the
+ * methods that handle http request. For ex:
```
trustin: For example

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
```diff
@@ -0,0 +1,52 @@
...
+ * <pre>
+ * {@code @PathPrefix("/b")}
+ * public class MyService {
```
trustin: You can do this in our Javadoc:

```java
 * <pre>{@code
 * > @PathPrefix("/b")
 * > public class MyService {
 * >     @Get("/c")
 * >     public HttpResponse foo() { ... }
 * > }
 * }</pre>
```

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
```diff
@@ -0,0 +1,52 @@
...
+ *     {@code @Get("/c")}
+ *     public HttpResponse foo() {...}
+ * }
+ * </pre>
+ * And then can be registered to {@link ServerBuilder} like this
+ * <pre>{@code
+ * ServerBuilder sb = new ServerBuilder();
+ * sb.annotatedService("/a", new MyService());
+ * }</pre>
+ * In this case <b>foo</b> methods handles a request that matches path <b>a/b/c</b>
```
trustin:
```
 * }</pre>
 *
 * <p>In this case, the {@code foo()} method handles a request that matches the path 
 * {@code /a/b/c}.</p>
```

core/src/test/java/com/linecorp/armeria/internal/ArmeriaHttpUtilTest.java
```diff
@@ -71,6 +71,7 @@ void testConcatPaths() throws Exception {
 
         assertThat(concatPaths("/", "a")).isEqualTo("/a");
         assertThat(concatPaths("/", "/a")).isEqualTo("/a");
+        assertThat(concatPaths("/", "/")).isEqualTo("/");
```
trustin: Thanks for adding more test cases.

trustin:
Removed `[WIP]` :smile:

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
sivaalli: Added this back.

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
sivaalli: Corrected this :)

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
sivaalli: Used this block. Thank you.

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
sivaalli: Corrected this :)

core/src/test/java/com/linecorp/armeria/internal/ArmeriaHttpUtilTest.java
sivaalli: No Problem :)

sivaalli:
Build failure is unrelated to this PR.

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2016 LINE Corporation
```
minwoox: nit: 2019

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
```diff
@@ -0,0 +1,52 @@
...
+ * >     public HttpResponse foo() { ... }
+ * > }
+ * }</pre>
+ * And then can be registered to {@link ServerBuilder} like this
+ * <pre>{@code
+ * ServerBuilder sb = new ServerBuilder();
+ * sb.annotatedService("/a", new MyService());
+ * }</pre>
+ *
+ * <p>In this case {@code foo()} methods handles a request that matches path {@code /a/b/c}</p>
```
minwoox: nit: methods -> method

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
```diff
@@ -122,6 +132,35 @@ public void ofUserDefinedRepeatableDecorator() throws NoSuchMethodException {
...
     }
 
+    @Test
+    public void testFindAnnotatedServiceElementsWithPathPrefixAnnotation() {
+        final Object object = new PathPrefixServiceObject();
+        final List<AnnotatedHttpServiceElement> elements = find("/", object, new ArrayList<>());
+
+        final List<String> paths = elements.stream()
+                                           .map(AnnotatedHttpServiceElement::route)
+                                           .map(Route::paths)
```
minwoox: This returns the list which has two same paths for one `route`. How about just getting the first element here?
```java
final List<String> paths = elements.stream()
                                   .map(AnnotatedHttpServiceElement::route)
                                   .map(route -> route.paths().get(0))
                                   .collect(Collectors.toList());
assertThat(paths).containsExactly("/home/hello", "/home/");
...

File: core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java [hunk 13]
```diff
@@ -122,6 +132,35 @@ public void ofUserDefinedRepeatableDecorator() throws NoSuchMethodException {
...
+
+    @Test
+    public void testFindAnnotatedServiceElementsWithoutPathPrefixAnnotation() {
+        final Object serviceObject = new ServiceObject();
+        final List<AnnotatedHttpServiceElement> elements = find(HOME_PATH_PREFIX, serviceObject,
+                                                                new ArrayList<>());
+
+        final List<String> paths = elements.stream()
+                                           .map(AnnotatedHttpServiceElement::route)
+                                           .map(Route::paths)
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -759,6 +763,22 @@ static String findDescription(AnnotatedElement annotatedElement) {
         return null;
     }
 
+    /**
+     * Return the path prefix to use. If there is {@link PathPrefix} annotation on the class
```
minwoox: nit: Returns

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -759,6 +763,22 @@ static String findDescription(AnnotatedElement annotatedElement) {
...
+     * then path prefix is computed by concatenating pathPrefix and value from annotation else
+     * return pathPrefix
```
minwoox: nit: missing ending `.`

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
```diff
@@ -224,4 +259,31 @@ public String userDefinedRepeatableDecorator() {
             return "";
         }
     }
+
+    @PathPrefix(HOME_PATH_PREFIX)
+    static class PathPrefixServiceObject {
+
+        @Get("/hello")
```
ikhoon: Could you add test cases for verifying exception with `@Get`, `@Post`? 😀

core/src/main/java/com/linecorp/armeria/server/annotation/PathPrefix.java
trustin: Javadoc will show `>` but we have some JavaScript that strips it out and applies syntax highlighting, so please keep using `>`. :smile:

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
sivaalli: Yup I will do that. Thanks for suggestion.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -255,7 +258,8 @@ private static HttpStatus defaultResponseStatus(Optional<HttpStatus> defaultResp
      * Returns an {@link AnnotatedHttpService} instance defined to {@code method} of {@code object} using
      * {@link Path} annotation.
      */
-    private static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Method method,
+    @VisibleForTesting
+    static AnnotatedHttpServiceElement create(String pathPrefix, Object object, Method method,
                                                       List<ExceptionHandlerFunction> baseExceptionHandlers,
                                                       List<RequestConverterFunction> baseRequestConverters,
                                                       List<ResponseConverterFunction> baseResponseConverters) {
```
trustin: Indentation

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactoryTest.java
```diff
@@ -122,6 +141,46 @@ public void ofUserDefinedRepeatableDecorator() throws NoSuchMethodException {
...
+        final List<String> paths = elements.stream()
+                                           .map(AnnotatedHttpServiceElement::route)
+                                           .map(route -> route.paths().get(0))
+                                           .collect(Collectors.toList());
+
+        assertThat(paths).containsExactlyInAnyOrder(HOME_PATH_PREFIX + "/hello", HOME_PATH_PREFIX + '/');
+    }
+
+    @Test
+    public void testCreateAnnotatedServiceElementWithoutExplicitPathOnMethod() {
```
ikhoon: Nice!

sivaalli:
Thank you @trustin @minwoox @ikhoon for your support. I look forward to contribute more to Armeria.

PR #2101, opened 2019-09-20
Add documentation to `EventLoopThreadFactory`

jyblue:
Motivation :

Sometimes method-level documentation for deprecated method provides
useful information to users.

Modifications :

Add javadoc documentation to deprecated constructors of `EventLoopThreadFactory`

Result :

Better user information.
Related #2067

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
@@ -37,7 +37,12 @@ public final class EventLoopThreadFactory extends AbstractThreadFactory {
      * Creates a new factory that creates a non-daemon and normal-priority thread.
      *
      * @param threadNamePrefix the prefix of the names of the threads created by this factory.
+     *
+     * @deprecated Use {@link ThreadFactories#newEventLoopThreadFactory(String, boolean)} or
+     *             {@link ThreadFactories#builder(String)}. Make sure to set
+     *             {@link ThreadFactoryBuilder#eventLoop(boolean)} true to create event loop threads.
      */
+    @Deprecated
     public EventLoopThreadFactory(String threadNamePrefix) {
         this(threadNamePrefix, false);
     }
@@ -47,7 +52,12 @@ public final class EventLoopThreadFactory extends AbstractThreadFactory {
      *
      * @param threadNamePrefix the prefix of the names of the threads created by this factory.
      * @param daemon whether to create a daemon thread.
+     *
+     * @deprecated Use {@link ThreadFactories#newEventLoopThreadFactory(String, boolean)} or
+     *             {@link ThreadFactories#builder(String)}. Make sure to set
+     *             {@link ThreadFactoryBuilder#eventLoop(boolean)} true to create event loop threads.
      */
+    @Deprecated
     public EventLoopThreadFactory(String threadNamePrefix, boolean daemon) {
         this(threadNamePrefix, daemon, Thread.NORM_PRIORITY);
     }
@@ -57,7 +67,12 @@ public final class EventLoopThreadFactory extends AbstractThreadFactory {
      *
      * @param threadNamePrefix the prefix of the names of the threads created by this factory.
      * @param priority the priority of the threads created by this factory.
+     *
+     * @deprecated Use {@link ThreadFactories#newEventLoopThreadFactory(String, boolean)} or
+     *             {@link ThreadFactories#builder(String)}. Make sure to set
+     *             {@link ThreadFactoryBuilder#eventLoop(boolean)} true to create event loop threads.
      */
+    @Deprecated
     public EventLoopThreadFactory(String threadNamePrefix, int priority) {
         this(threadNamePrefix, false, priority);
     }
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
```diff
@@ -57,7 +67,12 @@ public EventLoopThreadFactory(String threadNamePrefix, boolean daemon) {
      *
      * @param threadNamePrefix the prefix of the names of the threads created by this factory.
      * @param priority the priority of the threads created by this factory.
+     *
+     * @deprecated Use {@link ThreadFactories#newEventLoopThreadFactory(String, boolean)} or
```
minwoox: Can't create a thread factocy whose `priority` is that we specified here if we use `ThreadFactories#newEventLoopThreadFactory(String, boolean)}`. Could you remove this phrase?

core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
```diff
@@ -68,7 +83,12 @@ public EventLoopThreadFactory(String threadNamePrefix, int priority) {
      * @param threadNamePrefix the prefix of the names of the threads created by this factory.
      * @param daemon whether to create a daemon thread.
      * @param priority the priority of the threads created by this factory.
+     *
+     * @deprecated Use {@link ThreadFactories#newEventLoopThreadFactory(String, boolean)} or
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
```diff
@@ -80,7 +100,12 @@ public EventLoopThreadFactory(String threadNamePrefix, boolean daemon, int prior
      * @param daemon whether to create a daemon thread.
      * @param priority the priority of the threads created by this factory.
      * @param threadGroup the {@link ThreadGroup}.
+     *
+     * @deprecated Use {@link ThreadFactories#newEventLoopThreadFactory(String, boolean)} or
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
```diff
@@ -37,7 +37,12 @@
      * Creates a new factory that creates a non-daemon and normal-priority thread.
      *
      * @param threadNamePrefix the prefix of the names of the threads created by this factory.
+     *
+     * @deprecated Use {@link ThreadFactories#newEventLoopThreadFactory(String, boolean)} or
```
minwoox: `(String, boolean)` -> `(String)`

jyblue: Question : We don't have a shortcut method `newEventLoopThreadFactory(String)`.
                 Then, should I remove javadoc phrase or add a new method?

minwoox: I misread the code, sorry about that. 😄 @ikhoon will leave a comment for this. 😉

jyblue: Thank you for review, and quick answer! :+1:

ikhoon: `newEventLoopThreadFactory` does not have an overload method for `(String)`.
I think this documentation can confuse a user who creates `EventLoopThreadFactory` with only  `threadNamePrefix`.

Could you update your documentation to `{@linkplain ThreadFactories#newEventLoopThreadFactory(String, boolean) ThreadFactories#newEventLoopThreadFactory(String, false)}` for guiding a user who uses default value.

core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
jyblue: Removed it.

core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
jyblue: Removed it.

core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
jyblue: Removed it.

core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
jyblue: Updated it. :smile:

core/src/main/java/com/linecorp/armeria/common/util/EventLoopThreadFactory.java
```diff
@@ -37,7 +37,13 @@
      * Creates a new factory that creates a non-daemon and normal-priority thread.
      *
      * @param threadNamePrefix the prefix of the names of the threads created by this factory.
+     *
+     * @deprecated Use {@linkplain ThreadFactories#newEventLoopThreadFactory(String, boolean)
+     *             ThreadFactories#newEventLoopThreadFactory(String, false)} or
```
jyblue: I'm not sure this line break is appropriate. PTAL :bowing_man:

trustin: I think you need to use `@link` because `@linkplain` will produce a non-monospace text.

trustin: Confirmed the the page is rendered correctly.

PR #2102, opened 2019-09-22
HTTP header/parameter-based path mapping

### LINKED ISSUE #1207: HTTP header/parameter-based path mapping
Spring framework provides HTTP header/parameter-based path mapping like below:

```
@GetMapping(value = "/hello", params = {"name"})
public String method(@RequestParam("name") String name) { ... }
```

I think that it's not a high priority issue, but we need to figure out how to support this kind of usage.

hyangtack:
Motivation:
Spring framework provides HTTP header/parameter-based path mapping like below:
```
@GetMapping(value = "/hello", params = {"name"})
public String method(@RequestParam("name") String name) { ... }
```
It would be useful for a user who wants to migrate his or her Spring based service to Armeria.

Modifications:
- Add `RoutingPredicate<T>` as an internal class in order to check whether the request headers 
  or parameters are matched by the condition that a user specified.
- Add `matchesParams` and `matchesHeader` methods to `RouteBuilder` and `AbstractBindingBuilder`.
- Add `@MatchesParam` and `@MatchesHeader` annotations for annotated service.
- Do not cache `Route` that has HTTP header/parameter-based path mapping.
- Add `requiresMatchingParamsPredicates` and `requiresMatchingHeadersPredicates` methods 
  to `RoutingContext` interface in order to control the route resolver.

Result:
- Closes #1207
- A user can specify HTTP header/parameter-based path mapping as follows:
```
Server.builder().route().post("/foo")
      .matchesHeaders("baz=qux")
      .matchesParams("quux=1")
...
```
- They also can specify their own value predicate:
```
Server.builder().route().post("/foo")
      .matchesHeaders("baz", "qux"::equals)
      .matchesParams("quux", value -> Integer.parseInt(value) < 10)
...
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
@@ -56,7 +56,7 @@ abstract class AbstractBindingBuilder {
     private Set<HttpMethod> methods = ImmutableSet.of();
     private Set<MediaType> consumeTypes = ImmutableSet.of();
     private Set<MediaType> produceTypes = ImmutableSet.of();
-    private Iterable<? extends String> paramPredicates = ImmutableList.of();
+    private Iterable<String> paramPredicates = ImmutableList.of();
     private Iterable<? extends CharSequence> headerPredicates = ImmutableList.of();
     private final Map<RouteBuilder, Set<HttpMethod>> routeBuilders = new LinkedHashMap<>();
     private final Set<RouteBuilder> pathBuilders = new LinkedHashSet<>();
@@ -336,7 +336,7 @@ abstract class AbstractBindingBuilder {
      *
      * @see ConditionalParam
      */
-    public AbstractBindingBuilder matchesParamPredicates(Iterable<? extends String> paramPredicates) {
+    public AbstractBindingBuilder matchesParamPredicates(Iterable<String> paramPredicates) {
         this.paramPredicates = ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates"));
         return this;
     }
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
@@ -42,8 +42,11 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 
+import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.server.annotation.ConditionalHeader;
+import com.linecorp.armeria.server.annotation.ConditionalParam;
 
 /**
  * An abstract builder class for binding something to a {@link Route} fluently.
@@ -53,6 +56,8 @@ abstract class AbstractBindingBuilder {
     private Set<HttpMethod> methods = ImmutableSet.of();
     private Set<MediaType> consumeTypes = ImmutableSet.of();
     private Set<MediaType> produceTypes = ImmutableSet.of();
+    private Iterable<? extends CharSequence> paramPredicates = ImmutableList.of();
+    private Iterable<? extends CharSequence> headerPredicates = ImmutableList.of();
     private final Map<RouteBuilder, Set<HttpMethod>> routeBuilders = new LinkedHashMap<>();
     private final Set<RouteBuilder> pathBuilders = new LinkedHashSet<>();
 
@@ -290,6 +295,93 @@ abstract class AbstractBindingBuilder {
     }
 
     /**
+     * Sets the {@link Route} to accept a request if it matches all the specified predicates for
+     * HTTP parameters. The predicate can be one of the following forms:
+     * <ul>
+     *     <li>{@code some-param=some-value} which means that the request must have a
+     *     {@code some-param=some-value} parameter</li>
+     *     <li>{@code some-param!=some-value} which means that the request must not have a
+     *     {@code some-param=some-value} parameter</li>
+     *     <li>{@code some-param} which means that the request must contain a {@code some-param} parameter</li>
+     *     <li>{@code !some-param} which means that the request must not contain a {@code some-param}
+     *     parameter</li>
+     * </ul>
+     *
+     * <p>Note that these predicates can be evaluated only with the query string of the request URI.
+     * Also note that each predicate will be evaluated with the decoded value of HTTP parameters,
+     * so do not use percent-encoded value in the predicate.
+     *
+     * @see ConditionalParam
+     */
+    public AbstractBindingBuilder matchesParamPredicates(CharSequence... paramPredicates) {
+        return matchesParamPredicates(ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates")));
+    }
+
+    /**
+     * Sets the {@link Route} to accept a request if it matches all the specified predicates for
+     * HTTP parameters. The predicate can be one of the following forms:
+     * <ul>
+     *     <li>{@code some-param=some-value} which means that the request must have a
+     *     {@code some-param=some-value} parameter</li>
+     *     <li>{@code some-param!=some-value} which means that the request must not have a
+     *     {@code some-param=some-value} parameter</li>
+     *     <li>{@code some-param} which means that the request must contain a {@code some-param} parameter</li>
+     *     <li>{@code !some-param} which means that the request must not contain a {@code some-param}
+     *     parameter</li>
+     * </ul>
+     *
+     * <p>Note that these predicates can be evaluated only with the query string of the request URI.
+     * Also note that each predicate will be evaluated with the decoded value of HTTP parameters,
+     * so do not use percent-encoded value in the predicate.
+     *
+     * @see ConditionalParam
+     */
+    public AbstractBindingBuilder matchesParamPredicates(Iterable<? extends CharSequence> paramPredicates) {
+        this.paramPredicates = ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates"));
+        return this;
+    }
+
+    /**
+     * Sets the {@link Route} to accept a request if it matches all the specified predicates for
+     * {@link HttpHeaders}. The predicate can be one of the following forms:
+     * <ul>
+     *     <li>{@code some-header=some-value} which means that the request must have a
+     *     {@code some-header: some-value} header</li>
+     *     <li>{@code some-header!=some-value} which means that the request must not have a
+     *     {@code some-header: some-value} header</li>
+     *     <li>{@code some-header} which means that the request must contain a {@code some-header} header</li>
+     *     <li>{@code !some-header} which means that the request must not contain a {@code some-header}
+     *     header</li>
+     * </ul>
+     *
+     * @see ConditionalHeader
+     */
+    public AbstractBindingBuilder matchesHeaderPredicates(CharSequence... headerPredicates) {
+        return matchesHeaderPredicates(
+                ImmutableList.copyOf(requireNonNull(headerPredicates, "headerPredicates")));
+    }
+
+    /**
+     * Sets the {@link Route} to accept a request if it matches all the specified predicates for
+     * {@link HttpHeaders}. The predicate can be one of the following forms:
+     * <ul>
+     *     <li>{@code some-header=some-value} which means that the request must have a
+     *     {@code some-header: some-value} header</li>
+     *     <li>{@code some-header!=some-value} which means that the request must not have a
+     *     {@code some-header: some-value} an header</li>
+     *     <li>{@code some-header} which means that the request must contain a {@code some-header} header</li>
+     *     <li>{@code !some-header} which means that the request must not contain a {@code some-header}
+     *     header</li>
+     * </ul>
+     *
+     * @see ConditionalHeader
+     */
+    public AbstractBindingBuilder matchesHeaderPredicates(Iterable<? extends CharSequence> headerPredicates) {
+        this.headerPredicates = ImmutableList.copyOf(requireNonNull(headerPredicates, "headerPredicates"));
+        return this;
+    }
+
+    /**
      * Returns a newly-created {@link Route}s based on the properties of this builder.
      */
     final List<Route> buildRouteList() {
@@ -311,6 +403,8 @@ abstract class AbstractBindingBuilder {
             builder.add(routeBuilder.methods(routeMethods)
                                     .consumes(consumeTypes)
                                     .produces(produceTypes)
+                                    .matchesParamPredicates(paramPredicates)
+                                    .matchesHeaderPredicates(headerPredicates)
                                     .build());
         });
 
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
@@ -56,7 +56,7 @@ abstract class AbstractBindingBuilder {
     private Set<HttpMethod> methods = ImmutableSet.of();
     private Set<MediaType> consumeTypes = ImmutableSet.of();
     private Set<MediaType> produceTypes = ImmutableSet.of();
-    private Iterable<? extends CharSequence> paramPredicates = ImmutableList.of();
+    private Iterable<? extends String> paramPredicates = ImmutableList.of();
     private Iterable<? extends CharSequence> headerPredicates = ImmutableList.of();
     private final Map<RouteBuilder, Set<HttpMethod>> routeBuilders = new LinkedHashMap<>();
     private final Set<RouteBuilder> pathBuilders = new LinkedHashSet<>();
@@ -313,7 +313,7 @@ abstract class AbstractBindingBuilder {
      *
      * @see ConditionalParam
      */
-    public AbstractBindingBuilder matchesParamPredicates(CharSequence... paramPredicates) {
+    public AbstractBindingBuilder matchesParamPredicates(String... paramPredicates) {
         return matchesParamPredicates(ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates")));
     }
 
@@ -336,7 +336,7 @@ abstract class AbstractBindingBuilder {
      *
      * @see ConditionalParam
      */
-    public AbstractBindingBuilder matchesParamPredicates(Iterable<? extends CharSequence> paramPredicates) {
+    public AbstractBindingBuilder matchesParamPredicates(Iterable<? extends String> paramPredicates) {
         this.paramPredicates = ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates"));
         return this;
     }
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
@@ -29,12 +29,14 @@ import static com.linecorp.armeria.common.HttpMethod.TRACE;
 import static com.linecorp.armeria.server.HttpHeaderUtil.ensureUniqueMediaTypes;
 import static java.util.Objects.requireNonNull;
 
+import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
@@ -46,8 +48,8 @@ import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.HttpParameters;
 import com.linecorp.armeria.common.MediaType;
-import com.linecorp.armeria.server.annotation.ConditionalHeader;
-import com.linecorp.armeria.server.annotation.ConditionalParam;
+import com.linecorp.armeria.server.annotation.MatchesHeader;
+import com.linecorp.armeria.server.annotation.MatchesParam;
 
 /**
  * An abstract builder class for binding something to a {@link Route} fluently.
@@ -57,8 +59,8 @@ abstract class AbstractBindingBuilder {
     private Set<HttpMethod> methods = ImmutableSet.of();
     private Set<MediaType> consumeTypes = ImmutableSet.of();
     private Set<MediaType> produceTypes = ImmutableSet.of();
-    private List<RoutingPredicate<HttpParameters>> paramPredicates = ImmutableList.of();
-    private List<RoutingPredicate<HttpHeaders>> headerPredicates = ImmutableList.of();
+    private List<RoutingPredicate<HttpParameters>> paramPredicates = new ArrayList<>();
+    private List<RoutingPredicate<HttpHeaders>> headerPredicates = new ArrayList<>();
     private final Map<RouteBuilder, Set<HttpMethod>> routeBuilders = new LinkedHashMap<>();
     private final Set<RouteBuilder> pathBuilders = new LinkedHashSet<>();
 
@@ -312,7 +314,7 @@ abstract class AbstractBindingBuilder {
      * Also note that each predicate will be evaluated with the decoded value of HTTP parameters,
      * so do not use percent-encoded value in the predicate.
      *
-     * @see ConditionalParam
+     * @see MatchesParam
      */
     public AbstractBindingBuilder matchesParams(String... paramPredicates) {
         return matchesParams(ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates")));
@@ -335,19 +337,22 @@ abstract class AbstractBindingBuilder {
      * Also note that each predicate will be evaluated with the decoded value of HTTP parameters,
      * so do not use percent-encoded value in the predicate.
      *
-     * @see ConditionalParam
+     * @see MatchesParam
      */
     public AbstractBindingBuilder matchesParams(Iterable<String> paramPredicates) {
-        this.paramPredicates = RoutingPredicate.copyOfParamPredicates(
-                requireNonNull(paramPredicates, "paramPredicates"));
+        this.paramPredicates.addAll(RoutingPredicate.copyOfParamPredicates(
+                requireNonNull(paramPredicates, "paramPredicates")));
         return this;
     }
 
     /**
-     * Sets the {@link Route} to accept a request if it the specified {@code predicate} returns {@code true}.
+     * Sets the {@link Route} to accept a request when the specified {@code valuePredicate} evaluates
+     * {@code true} with the value of the specified {@code paramName} parameter.
      */
-    public AbstractBindingBuilder matchesParams(RoutingPredicate<HttpParameters> predicate) {
-        paramPredicates = ImmutableList.of(requireNonNull(predicate, "predicate"));
+    public AbstractBindingBuilder matchesParams(String paramName, Predicate<String> valuePredicate) {
+        requireNonNull(paramName, "paramName");
+        requireNonNull(valuePredicate, "valuePredicate");
+        paramPredicates.add(RoutingPredicate.ofParams(paramName, valuePredicate));
         return this;
     }
 
@@ -364,7 +369,7 @@ abstract class AbstractBindingBuilder {
      *     header</li>
      * </ul>
      *
-     * @see ConditionalHeader
+     * @see MatchesHeader
      */
     public AbstractBindingBuilder matchesHeaders(String... headerPredicates) {
         return matchesHeaders(ImmutableList.copyOf(requireNonNull(headerPredicates, "headerPredicates")));
@@ -383,19 +388,22 @@ abstract class AbstractBindingBuilder {
      *     header</li>
      * </ul>
      *
-     * @see ConditionalHeader
+     * @see MatchesHeader
      */
     public AbstractBindingBuilder matchesHeaders(Iterable<String> headerPredicates) {
-        this.headerPredicates = RoutingPredicate.copyOfHeaderPredicates(
-                requireNonNull(headerPredicates, "headerPredicates"));
+        this.headerPredicates.addAll(RoutingPredicate.copyOfHeaderPredicates(
+                requireNonNull(headerPredicates, "headerPredicates")));
         return this;
     }
 
     /**
-     * Sets the {@link Route} to accept a request if it the specified {@code predicate} returns {@code true}.
+     * Sets the {@link Route} to accept a request when the specified {@code valuePredicate} evaluates
+     * {@code true} with the value of the specified {@code headerName} header.
      */
-    public AbstractBindingBuilder matchesHeaders(RoutingPredicate<HttpHeaders> predicate) {
-        headerPredicates = ImmutableList.of(requireNonNull(predicate, "predicate"));
+    public AbstractBindingBuilder matchesHeaders(CharSequence headerName, Predicate<String> valuePredicate) {
+        requireNonNull(headerName, "headerName");
+        requireNonNull(valuePredicate, "valuePredicate");
+        headerPredicates.add(RoutingPredicate.ofHeaders(headerName, valuePredicate));
         return this;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
@@ -59,8 +59,8 @@ abstract class AbstractBindingBuilder {
     private Set<HttpMethod> methods = ImmutableSet.of();
     private Set<MediaType> consumeTypes = ImmutableSet.of();
     private Set<MediaType> produceTypes = ImmutableSet.of();
-    private List<RoutingPredicate<QueryParams>> paramPredicates = new ArrayList<>();
-    private List<RoutingPredicate<HttpHeaders>> headerPredicates = new ArrayList<>();
+    private final List<RoutingPredicate<QueryParams>> paramPredicates = new ArrayList<>();
+    private final List<RoutingPredicate<HttpHeaders>> headerPredicates = new ArrayList<>();
     private final Map<RouteBuilder, Set<HttpMethod>> routeBuilders = new LinkedHashMap<>();
     private final Set<RouteBuilder> pathBuilders = new LinkedHashSet<>();
 
@@ -349,7 +349,7 @@ abstract class AbstractBindingBuilder {
      * Sets the {@link Route} to accept a request when the specified {@code valuePredicate} evaluates
      * {@code true} with the value of the specified {@code paramName} parameter.
      */
-    public AbstractBindingBuilder matchesParams(String paramName, Predicate<String> valuePredicate) {
+    public AbstractBindingBuilder matchesParams(String paramName, Predicate<? super String> valuePredicate) {
         requireNonNull(paramName, "paramName");
         requireNonNull(valuePredicate, "valuePredicate");
         paramPredicates.add(RoutingPredicate.ofParams(paramName, valuePredicate));
@@ -400,7 +400,8 @@ abstract class AbstractBindingBuilder {
      * Sets the {@link Route} to accept a request when the specified {@code valuePredicate} evaluates
      * {@code true} with the value of the specified {@code headerName} header.
      */
-    public AbstractBindingBuilder matchesHeaders(CharSequence headerName, Predicate<String> valuePredicate) {
+    public AbstractBindingBuilder matchesHeaders(CharSequence headerName,
+                                                 Predicate<? super String> valuePredicate) {
         requireNonNull(headerName, "headerName");
         requireNonNull(valuePredicate, "valuePredicate");
         headerPredicates.add(RoutingPredicate.ofHeaders(headerName, valuePredicate));
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
@@ -46,8 +46,8 @@ import com.google.common.collect.Sets;
 
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpMethod;
-import com.linecorp.armeria.common.HttpParameters;
 import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.common.QueryParams;
 import com.linecorp.armeria.server.annotation.MatchesHeader;
 import com.linecorp.armeria.server.annotation.MatchesParam;
 
@@ -59,7 +59,7 @@ abstract class AbstractBindingBuilder {
     private Set<HttpMethod> methods = ImmutableSet.of();
     private Set<MediaType> consumeTypes = ImmutableSet.of();
     private Set<MediaType> produceTypes = ImmutableSet.of();
-    private List<RoutingPredicate<HttpParameters>> paramPredicates = new ArrayList<>();
+    private List<RoutingPredicate<QueryParams>> paramPredicates = new ArrayList<>();
     private List<RoutingPredicate<HttpHeaders>> headerPredicates = new ArrayList<>();
     private final Map<RouteBuilder, Set<HttpMethod>> routeBuilders = new LinkedHashMap<>();
     private final Set<RouteBuilder> pathBuilders = new LinkedHashSet<>();
--- a/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
@@ -44,6 +44,7 @@ import com.google.common.collect.Sets;
 
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpParameters;
 import com.linecorp.armeria.common.MediaType;
 import com.linecorp.armeria.server.annotation.ConditionalHeader;
 import com.linecorp.armeria.server.annotation.ConditionalParam;
@@ -56,8 +57,8 @@ abstract class AbstractBindingBuilder {
     private Set<HttpMethod> methods = ImmutableSet.of();
     private Set<MediaType> consumeTypes = ImmutableSet.of();
     private Set<MediaType> produceTypes = ImmutableSet.of();
-    private Iterable<String> paramPredicates = ImmutableList.of();
-    private Iterable<? extends CharSequence> headerPredicates = ImmutableList.of();
+    private List<RoutingPredicate<HttpParameters>> paramPredicates = ImmutableList.of();
+    private List<RoutingPredicate<HttpHeaders>> headerPredicates = ImmutableList.of();
     private final Map<RouteBuilder, Set<HttpMethod>> routeBuilders = new LinkedHashMap<>();
     private final Set<RouteBuilder> pathBuilders = new LinkedHashSet<>();
 
@@ -313,8 +314,8 @@ abstract class AbstractBindingBuilder {
      *
      * @see ConditionalParam
      */
-    public AbstractBindingBuilder matchesParamPredicates(String... paramPredicates) {
-        return matchesParamPredicates(ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates")));
+    public AbstractBindingBuilder matchesParams(String... paramPredicates) {
+        return matchesParams(ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates")));
     }
 
     /**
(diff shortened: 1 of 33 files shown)


hyangtack:
To-Dos:
- More tests.
- Changes related to Documentation service. (@minwoox Please help me when I start to do this :-))

@trustin I'm curious that this change is reasonable. Please let me know what you think. :-)

minwoox:
> @minwoox Please help me when I start to do this :-))

Screen Shot 2019-09-23 at 10 41 19 AM

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -267,6 +279,102 @@ public RouteBuilder produces(Iterable<MediaType> produceTypes) {
...
+     *     {@code some-header: some-value} header</li>
+     *     <li>{@code some-header!=some-value} which the request must not have a
+     *     {@code some-header: some-value} header</li>
+     *     <li>{@code some-header} which the request must contain a {@code some-header} header</li>
+     *     <li>{@code !some-header} which the request must not contain a {@code some-header} header</li>
+     * </ul>
+     *
+     * @see Headers
+     */
+    public final RouteBuilder headers(String... headerPredicates) {
```
trustin: `CharSequence...`? We represent headers as `AsciiString`s, so all header names must be converted to header names via `HttpHeaderNames.of()`.

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -267,6 +279,102 @@ public RouteBuilder produces(Iterable<MediaType> produceTypes) {
...
+     *     {@code some-header: some-value} header</li>
+     *     <li>{@code some-header!=some-value} which the request must not have a
+     *     {@code some-header: some-value} an header</li>
+     *     <li>{@code some-header} which the request must contain a {@code some-header} header</li>
+     *     <li>{@code !some-header} which the request must not contain a {@code some-header} header</li>
+     * </ul>
+     *
+     * @see Headers
+     */
+    public RouteBuilder headers(Iterable<String> headerPredicates) {
```
trustin: `Iterable<? extends CharSequence>` with conversion to `AsciiString`s.

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -267,6 +279,102 @@ public RouteBuilder produces(Iterable<MediaType> produceTypes) {
...
+    }
+
+    /**
+     * TODO(hygntack) Is this useful?.
+     *
+     * @param id the identifier of the specified {@code headersPredicate} which is used for logging or
+     *           metric collection
+     * @param headersPredicate the predicates to be used when accepting a request by a {@link Route}
+     */
+    public RouteBuilder headers(String id, Predicate<HttpHeaders> headersPredicate) {
```
trustin: I think it will be useful, especially for ourselves; we could re-implement `consumes()` and `produces()` using this feature, and I guess a user may have a similar use case.

However, I'm not sure if it's a good idea to let a user specify `id` here. I think we should solve this problem at higher-level, like letting a user assign an `id` to a `Route`.

core/src/main/java/com/linecorp/armeria/server/RoutingContext.java
```diff
@@ -62,6 +64,11 @@
     @Nullable
     String query();
 
+    /**
+     * Returns the query retrieved from the request as a map.
+     */
+    Map<String, List<String>> queryParameters();
```
trustin: Could use/return `HttpParameters`?

core/src/main/java/com/linecorp/armeria/server/annotation/Params.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.linecorp.armeria.server.RouteBuilder;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(ParamsGroup.class)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+public @interface Params {
```
trustin: We should not use a plural form here because we only specify a single parameter. However, we are already using `@Param` for injecting parameters, so I guess in this case we should name it differently to avoid confusion. e.g. `@RequiredParam` ?

core/src/main/java/com/linecorp/armeria/server/annotation/Headers.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.linecorp.armeria.server.RouteBuilder;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(HeadersGroup.class)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+public @interface Headers {
```
trustin: We should not use a plural form here because we only specify a single header. However, we are already using `@Header` for injecting header values, so I guess in this case we should name it differently to avoid confusion. e.g. `@RequiredHeader` ?

trustin:
@hyangtack Gentle ping :wink:

core/src/main/java/com/linecorp/armeria/server/annotation/Params.java
hyangtack: Thanks! `@Required...` sounds like that the parameter or header must exist. But we have *!=* case as well. So I changed it `@ConditionalParam` and `@ConditionalHeader`. What do you think? :-)

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
hyangtack: Yeah, I agree with you. Would it be better to address it with follow-up PR?

hyangtack:
I haven't finished yet DocService and documentation, but it would be better to finish them after getting comments from you folks about this change. :-)

core/src/main/java/com/linecorp/armeria/server/annotation/ConditionalHeader.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.linecorp.armeria.server.RouteBuilder;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(ConditionalHeaders.class)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+public @interface ConditionalHeader {
```
trustin: Missing Javadoc

core/src/main/java/com/linecorp/armeria/server/annotation/ConditionalParam.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.linecorp.armeria.server.RouteBuilder;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(ConditionalParams.class)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+public @interface ConditionalParam {
```
trustin: Missing Javadoc

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,199 @@
...
+     * @see DefaultRoute where this predicate is evalued
+     */
+    @Override
+    public boolean test(T t) {
+        return delegate.test(t);
+    }
+
+    @VisibleForTesting
+    static final class ParsedPredicate {
+        final boolean isInverted;
```
trustin: `isNegated`?

minwoox:
> I haven't finished yet DocService and documentation, but it would be better to finish them after getting comments from you folks about this change. :-)

I think we can make it in the separate PR because `DocService` still works without the information of `ConditionalParam` and `ConditionalHeader`. And I think I can help you with that. 😄

core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
```diff
@@ -43,35 +47,58 @@
...
+        }
+        if (!paramPredicates.isEmpty()) {
+            complexity += 1 << 3;
+        }
+        if (!headerPredicates.isEmpty()) {
+            complexity += 1 << 4;
         }
         this.complexity = complexity;
+
+        if (paramPredicates.isEmpty() && headerPredicates.isEmpty()) {
```
minwoox: Couldn't we cache when `paramPredicates` is not empty()?
We already add `routingCtx.path()` to the `summary` which is used for `equals` in `DefaultRoutingContext` so if we add `routingCtx.query()` to the `summary`, isn't it cacheable?

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -318,6 +330,107 @@ public RouteBuilder produces(Iterable<MediaType> produceTypes) {
...
+     *     parameter</li>
+     * </ul>
+     *
+     * <p>Note that these predicates can be evaluated only with the query string of the request URI.
+     * Also note that each predicate will be evaluated with the decoded value of HTTP parameters,
+     * so do not use percent-encoded value in the predicate.
+     *
+     * @see ConditionalParam
+     */
+    public RouteBuilder matchesParamPredicates(CharSequence... paramPredicates) {
```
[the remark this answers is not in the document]
hyangtack: No. I didn't think about that deeply and just changed them all. 😢 Thanks!

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -318,6 +330,107 @@ public RouteBuilder produces(Iterable<MediaType> produceTypes) {
...
+    }
+
+    /**
+     * TODO(hyangtack) Is this useful?.
+     *
+     * @param id the identifier of the specified {@code headersPredicate} which is used for logging or
+     *           metric collection
+     * @param headerPredicate the predicates to be used when accepting a request by a {@link Route}
+     */
+    public RouteBuilder matchesHeaderPredicates(String id, Predicate<HttpHeaders> headerPredicate) {
```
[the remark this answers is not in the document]
hyangtack: Nope. I've left this for getting a comment. Let me remove it. :-)

core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
hyangtack: Yes, you're right. But I'm worried it makes cache hit ratio lower. Also, it'd affect the existing Armeria-based application which doesn't use this feature. 🤔

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
hyangtack: Thanks!

hyangtack:
I'm adding a new feature that supports `>=`, `>`, `<=` and `<` as a comparator. 
If the configured value can be parsed as `long` value, the values will be compared by numeric order. Otherwise, they will be compared by lexicographical order.
Does it make sense? 
I thought this feature because a user may want to handle a request separately with the client version header. But I'm worried whether this feature makes a user confused because the ordering is changed by the configured value.
What do you think?

core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
minwoox: Ah yes, let's leave it as it is then. 😄

minwoox:
> I'm adding a new feature that supports >=, >, <= and < as a comparator.

These operators shouldn't be applied only for numbers?

core/src/main/java/com/linecorp/armeria/server/Route.java
```diff
@@ -128,4 +128,9 @@ static RouteBuilder builder() {
      * Returns the {@link Set} of {@link MediaType}s that this {@link Route} produces.
      */
     Set<MediaType> produces();
+
+    /**
+     * Returns the cache key which can be used to check whether this {@link Route} is able to be cached.
```
trustin: is able to be -> can be

core/src/main/java/com/linecorp/armeria/server/RoutingContext.java
```diff
@@ -61,6 +63,11 @@
     @Nullable
     String query();
 
+    /**
+     * Returns the query retrieved from the request as an {@link HttpParameters}.
+     */
+    HttpParameters httpParameters();
```
trustin: How about just `params` or `queryParams`?

core/src/main/java/com/linecorp/armeria/server/RoutingContext.java
```diff
@@ -61,6 +63,11 @@
     @Nullable
     String query();
 
+    /**
+     * Returns the query retrieved from the request as an {@link HttpParameters}.
```
trustin: - `the query` -> `the query parameters`
- `from the request` -> `from the request path` (because we do not look into the content)
- Could remove `as an ...`?

core/src/main/java/com/linecorp/armeria/server/RoutingContext.java
```diff
@@ -74,6 +81,11 @@
      */
     List<MediaType> acceptTypes();
 
+    /**
+     * Returns {@link HttpHeaders} of the request.
+     */
+    HttpHeaders headers();
```
trustin: Could return `RequestHeaders`?

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,227 @@
...
+                    return value -> value.compareTo(configuredValue) > 0;
+                case ">=":
+                    return value -> value.compareTo(configuredValue) >= 0;
+                case "<":
+                    return value -> value.compareTo(configuredValue) < 0;
+                case "<=":
+                    return value -> value.compareTo(configuredValue) <= 0;
+            }
+        }
+        throw new IllegalStateException("Pattern has changed unexpectedly: " + COMPARE_PATTERN.pattern());
```
trustin: What does this mean? How about rephrasing the exception message so it's more user-friendly?

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    private static Predicate<String> longValueMatcher(Predicate<Long> comparator) {
```
trustin: `Predicate<Long>` -> `LongPredicate`?

core/src/main/java/com/linecorp/armeria/server/ServiceBindingBuilder.java
```diff
@@ -159,6 +159,26 @@ public ServiceBindingBuilder produces(Iterable<MediaType> produceTypes) {
         return (ServiceBindingBuilder) super.produces(produceTypes);
     }
 
+    @Override
+    public ServiceBindingBuilder matchesParamPredicates(String... paramPredicates) {
```
trustin: - How about just `matchesParams(String... expressions)` and `matchesHeaders(String... expressions)`?
- It seems like we use `CharSequence...` and `Iterable<? extends CharSequence>` in some places. Could we make it consistent in all places?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -132,6 +132,29 @@ you can use :api:`@StatusCode` annotation as follows.
...
+.. code-block:: java
+
+    public class MyAnnotatedService {
+
+        // Handles a request which contains 'client-type: android' header.
+        @Get("/users")
+        @ConditionalHeader("client-type=android")
+        public User getUsers1() { ... }
+
+        // Handles a request which contains 'client-type' header. Any values of the 'client-type' header are welcomed.
```
trustin: welcomed -> accepted?

site/src/sphinx/server-annotated-service.rst
```diff
@@ -132,6 +132,29 @@ you can use :api:`@StatusCode` annotation as follows.
...
+        @Get("/users")
+        @ConditionalHeader("client-type")
+        public User getUsers2() { ... }
+
+        // Handles a request which doesn't contain 'client-type: android' header.
```
trustin: `'client-type' header`?

trustin:
> If the configured value can be parsed as long value, the values will be compared by numeric order. Otherwise, they will be compared by lexicographical order. Does it make sense?

How about allowing a user to specify a custom `Predicate<String>`? A version number will not always be an integer but in a form of `x.y.z`:

```java
Server.builder()
      .route().get("/foo")
              .matchesParam("foo=bar")
              .matchesParam("clientVersion", value -> ...)
              .build()
``

File: core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java (hunk 17 above)
Comment (hyangtack) [reply core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java]: Removed. :-)

File: core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java (hunk 18 above)
Comment (hyangtack) [reply core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java]: Removed. :-)

File: core/src/main/java/com/linecorp/armeria/server/ServiceBindingBuilder.java (hunk 19 above)
Comment (hyangtack) [reply core/src/main/java/com/linecorp/armeria/server/ServiceBindingBuilder.java]: I couldn't remember why I changed `String` to `CharSequence` for headers. BTW, it was not necessary. Will change them to `String`. Thanks!

**Comment by hyangtack** [comment]
> I'm adding a new feature that supports >=, >, <= and < as a comparator.

>> These operators shouldn't be applied only for numbers?

> How about allowing a user to specify a custom Predicate<String>? A version number will not always be an integer but in a form of x.y.z:

@minwoox @trustin Thanks for the comments. 👍 
@trustin 's idea sounds good to me. So I removed the code related to >=, >, <=, and < operators. And will add a method for a custom value predicate.

File: core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java [hunk 22]
```diff
@@ -0,0 +1,236 @@
...
+        return StreamSupport.stream(predicates.spliterator(), false)
+                            .map(RoutingPredicate::ofParams).collect(toImmutableList());
+    }
+
+    static RoutingPredicate<HttpHeaders> ofHeaders(CharSequence headerName, Predicate<String> valuePredicate) {
+        return new RoutingPredicate<>(headerName, headers -> {
+            try {
+                return headers.getAll(headerName).stream().anyMatch(valuePredicate);
+            } catch (Throwable cause) {
+                if (Flags.verboseExceptionSampler().isSampled(cause.getClass())) {
```
hyangtack: I'm curious whether this usage is okay or not. The reason why I don't log error always is the abnormal request is not intended by our user. It might be from abusing. Let me know if I need to implement this by other way.

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Tests the specified {@code t} object.
+     *
+     * @see DefaultRoute where this predicate is evalued
+     */
+    public boolean test(T t) {
+        try {
+            return delegate.test(t);
+        } catch (Throwable cause) {
+            if (Flags.verboseExceptionSampler().isSampled(cause.getClass())) {
```
hyangtack: I'm curious whether this usage is okay or not. The reason why I don't want to write log message every time is because the abnormal request may not be intended by our user. It may be from abusing or hacking. Let me know if I need to implement this by other way.

core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
```diff
@@ -289,6 +295,110 @@ public AbstractBindingBuilder produces(Iterable<MediaType> produceTypes) {
...
+        this.headerPredicates = RoutingPredicate.copyOfHeaderPredicates(
+                requireNonNull(headerPredicates, "headerPredicates"));
+        return this;
+    }
+
+    /**
+     * Sets the {@link Route} to accept a request if it the specified {@code predicate} returns {@code true}.
+     */
+    public AbstractBindingBuilder matchesHeaders(RoutingPredicate<HttpHeaders> predicate) {
+        headerPredicates = ImmutableList.of(requireNonNull(predicate, "predicate"));
```
trustin: Perhaps we should add to the existing list, so that a user can specify multiple `RoutingPredicate`s fluently? e.g. `builder.matchesHeaders(pred1).matchesHeaders(pred2)`

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
```diff
@@ -142,7 +148,7 @@ private RouteCache() {}
             }
 
             final Routed<V> result = delegate.find(routingCtx);
-            if (result.isPresent()) {
+            if (result.isPresent() && !noCacheRoutes.contains(result.route())) {
```
trustin: I'm somewhat concerned about the performance implication of this check, because it will involve at least the comparison of 3 collections and 1 `PathMapping`. Would it still work if we use a set with identity comparison?

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
trustin: Sounds OK to me. Could you add some comment about it?

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,255 @@
...
+                logger.warn("Failed to evaluate the value of header or param '{}'. " +
+                            "This exception MUST be properly handled by a user.: " +
```
trustin: `You MUST catch and handle this exception properly:` ?

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,255 @@
...
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpParameters;
+
+/**
+ * A {@link Predicate} to evaluate whether a request can be accepted by a service method.
+ * Currently predicates for {@link HttpHeaders} and HTTP parameters are supported.
+ *
+ * @param <T> the type of the object to be tested
+ */
+public final class RoutingPredicate<T> {
```
trustin: How does a user create a new instance of this class? I don't see any public static factory methods.

core/src/test/java/com/linecorp/armeria/server/annotation/ConditionalHeaderTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+class ConditionalHeaderTest {
+
+    @RegisterExtension
+    public static final ServerExtension extension = new ServerExtension() {
+        @Override
+        protected void configure(ServerBuilder sb) throws Exception {
+            sb.annotatedService(new Object() {
+                @Get("/matches")
+                @ConditionalHeader("my-header=my-value")
```
trustin: How about `@MatchesHeader` and `@MatchesParam`? Which do you think is better?

core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
hyangtack: That's what I wanted at first. But the other API doesn't do that now. So I made it like this. Is it okay to make this method as you suggested? :-)

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
hyangtack: OMG...Sorry about that. I thought I made some method public. But I think it would be better to make this class as private like before. Because we can add a predicate when we call `matches*` method.
https://github.com/line/armeria/pull/2102#discussion_r356055297

core/src/test/java/com/linecorp/armeria/server/annotation/ConditionalHeaderTest.java
hyangtack: `@Matches*` looks better for consistency. Thanks. :-)

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
hyangtack: Agreed. The no cache routes may be matched by identity comparison. So I think it would work. Let me change the set to Set<Integer> which store Route's hash code. Does it make sense?

hyangtack:
Rebased and addressed @trustin 's comments. Please have a look. :-)

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
```diff
@@ -122,7 +122,7 @@ private RouteCache() {}
         private final Function<V, Route> routeResolver;
         private final Cache<RoutingContext, V> findCache;
         private final Cache<RoutingContext, List<V>> findAllCache;
-        private final Set<Route> noCacheRoutes;
+        private final Set<Integer> noCacheRoutes;
```
trustin: Even if `Route` did not override `hashCode()`, it is still possible that two different objects happen to have the same hash code, so I think you need to use something like `IdentityHashMap`.

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -83,7 +83,8 @@
         requireNonNull(headersPredicate, "headersPredicate");
         final Matcher m = CONTAIN_PATTERN.matcher(headersPredicate);
         if (m.matches()) {
-            final Predicate<HttpHeaders> predicate = headers -> headers.contains(m.group(2));
+            final CharSequence headerName = m.group(2);
```
trustin: How about converting to `AsciiString` using `HttpHeaderNames.of()`, so that `HttpHeaders.contains()` doesn't have to do the conversion internally every time?

core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
```diff
@@ -254,6 +311,14 @@ private static String generateMeterTag(String parentTag, Set<HttpMethod> methods
         addMediaTypes(name, "consumes", consumes);
         addMediaTypes(name, "produces", produces);
 
+        if (!paramPredicates.isEmpty()) {
+            name.add("params");
+            name.add(meterTagJoiner.join(paramPredicates.stream().map(RoutingPredicate::name).iterator()));
```
trustin: Maybe `distinct()`, just in case there are more than one predicate for the same param name?

core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
```diff
@@ -254,6 +311,14 @@ private static String generateMeterTag(String parentTag, Set<HttpMethod> methods
...
+        }
+        if (!headerPredicates.isEmpty()) {
+            name.add("headers");
+            name.add(meterTagJoiner.join(headerPredicates.stream().map(RoutingPredicate::name).iterator()));
```
trustin: Maybe `distinct()`, just in case there are more than one predicate for the same header name?

core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
```diff
@@ -209,8 +249,15 @@ public int complexity() {
         return produces;
     }
 
+    @Override
+    public String cacheKey() {
+        return cacheKey;
```
trustin: I'm not sure if it's a good idea to have a `String` key like this. What do you think about replacing `cacheKey` with two member fields and their getters:

- `Set<String> predicateParamNames()`
- `Set<AsciiString> predicateHeaderNames()`

Then, we could calculate `noCacheRoutes` by comparing `Route`s' properties. Thoughts?

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
hyangtack: Ah, I see. I forgot `IdentityHashMap`. :-( Thanks!

core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
hyangtack: Agreed. I removed cacheKey from Route. Now compare pathType, paths, methods, consumes and produces. Please have a look and let me know what you think about this change. :-)

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
```diff
@@ -122,7 +122,7 @@ private RouteCache() {}
         private final Function<V, Route> routeResolver;
         private final Cache<RoutingContext, V> findCache;
         private final Cache<RoutingContext, List<V>> findAllCache;
-        private final Set<Integer> noCacheRoutes;
+        private final IdentityHashMap<Route, Route> noCacheRoutes;
```
trustin: `Set<Route> noCacheRoutes = Collections.newSetFromMap(new IdentityHashMap<Route, Boolean>())`?

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
```diff
@@ -132,8 +132,10 @@ private RouteCache() {}
             this.routeResolver = requireNonNull(routeResolver, "routeResolver");
             this.findCache = requireNonNull(findCache, "findCache");
             this.findAllCache = requireNonNull(findAllCache, "findAllCache");
-            this.noCacheRoutes = requireNonNull(noCacheRoutes, "noCacheRoutes")
-                    .stream().map(Route::hashCode).collect(toImmutableSet());
+
+            final IdentityHashMap<Route, Route> map = new IdentityHashMap<>();
+            requireNonNull(noCacheRoutes, "noCacheRoutes").forEach(r -> map.put(r, r));
```
trustin: How about this:

```java
final Set<Route> newNoCachingRoutes = Collections.newSetFromMap(new IdentityHashMap<Route, Boolean>());
newNoCachingRoutes.addAll(requireNonNull(noCacheRoutes, "noCacheRoutes"));
this.noCachingRoutes = Collections.unmodifiableSet(newNoCachingRoutes);
```

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    static List<RoutingPredicate<HttpParameters>> copyOfParamPredicates(Iterable<String> predicates) {
+        return StreamSupport.stream(predicates.spliterator(), false)
+                            .map(RoutingPredicate::ofParams).collect(toImmutableList());
+    }
+
+    static RoutingPredicate<HttpHeaders> ofHeaders(CharSequence headerName,
+                                                   Predicate<String> valuePredicate) {
+        final AsciiString name = AsciiString.of(headerName);
```
trustin: `HttpHeaderNames.of(headerName)`

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,201 @@
...
+        return new RoutingPredicate<>(headerName, headers ->
+                headers.getAll(name).stream().anyMatch(valuePredicate));
+    }
+
+    @VisibleForTesting
+    static RoutingPredicate<HttpHeaders> ofHeaders(String headersPredicate) {
+        requireNonNull(headersPredicate, "headersPredicate");
+        return of(headersPredicate, AsciiString::of, name -> headers -> headers.contains(name),
```
trustin: Ditto - `HttpHeaderNames::of`

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
hyangtack: Thanks. Learned one more thing. :-)

core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
```diff
@@ -289,6 +297,116 @@ public AbstractBindingBuilder produces(Iterable<MediaType> produceTypes) {
...
+        this.paramPredicates.addAll(RoutingPredicate.copyOfParamPredicates(
+                requireNonNull(paramPredicates, "paramPredicates")));
+        return this;
+    }
+
+    /**
+     * Sets the {@link Route} to accept a request when the specified {@code valuePredicate} evaluates
+     * {@code true} with the value of the specified {@code paramName} parameter.
+     */
+    public AbstractBindingBuilder matchesParams(String paramName, Predicate<String> valuePredicate) {
```
trustin: `? super String`

core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
```diff
@@ -289,6 +297,116 @@ public AbstractBindingBuilder produces(Iterable<MediaType> produceTypes) {
...
+        this.headerPredicates.addAll(RoutingPredicate.copyOfHeaderPredicates(
+                requireNonNull(headerPredicates, "headerPredicates")));
+        return this;
+    }
+
+    /**
+     * Sets the {@link Route} to accept a request when the specified {@code valuePredicate} evaluates
+     * {@code true} with the value of the specified {@code headerName} header.
+     */
+    public AbstractBindingBuilder matchesHeaders(CharSequence headerName, Predicate<String> valuePredicate) {
```
trustin: `? super String`

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -320,6 +332,132 @@ public RouteBuilder produces(Iterable<MediaType> produceTypes) {
...
+        this.paramPredicates.addAll(RoutingPredicate.copyOfParamPredicates(
+                requireNonNull(paramPredicates, "paramPredicates")));
+        return this;
+    }
+
+    /**
+     * Sets the {@link Route} to accept a request when the specified {@code valuePredicate} evaluates
+     * {@code true} with the value of the specified {@code paramName} parameter.
+     */
+    public RouteBuilder matchesParams(String paramName, Predicate<String> valuePredicate) {
```
trustin: `? super String`

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -320,6 +332,132 @@ public RouteBuilder produces(Iterable<MediaType> produceTypes) {
...
+        this.headerPredicates.addAll(RoutingPredicate.copyOfHeaderPredicates(
+                requireNonNull(headerPredicates, "headerPredicates")));
+        return this;
+    }
+
+    /**
+     * Sets the {@link Route} to accept a request when the specified {@code valuePredicate} evaluates
+     * {@code true} with the value of the specified {@code headerName} header.
+     */
+    public RouteBuilder matchesHeaders(CharSequence headerName, Predicate<String> valuePredicate) {
```
trustin: `? super String`

core/src/main/java/com/linecorp/armeria/server/RoutingPredicate.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    CharSequence name() {
+        return name;
+    }
+
+    /**
+     * Tests the specified {@code t} object.
+     *
+     * @see DefaultRoute where this predicate is evalued
```
minwoox: evaluated?

site/src/sphinx/server-basics.rst
```diff
@@ -177,6 +177,7 @@ for a specific service, you can use fluent API:
       .post("/foo/bar")                          // Matched when the path is "/foo/bar" and the method is POST.
       .consumes(MediaType.JSON)                  // Matched when the "content-type" header is "application/json".
       .produces(MediaType.JSON)                  // Matched when the "accept" headers is "application/json".
+      .matchesHeaders("baz=qux")                 // Matched when the "baz" header is "qux".
```
minwoox: Let'a add `matchesParams` as well. 😆

site/src/sphinx/server-annotated-service.rst
```diff
@@ -132,6 +132,29 @@ you can use :api:`@StatusCode` annotation as follows.
         public void deleteUser(@Param("name") String name) { ... }
     }
 
+A service method can receive a request only if it contains a header or parameter the method requires.
```
minwoox: Not sure this is better but how about `You can define a service method which handles a request only if it ...`

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
ikhoon: I'm not sure I understand `noCacheRoutes` correctly. But don't we need to add 
`!noCacheRoutes.contains(result.route()))` predicate to `findAll(RoutingContext)` method?

site/src/sphinx/server-annotated-service.rst
hyangtack: Sounds better. Let me fix it. :-)

core/src/main/java/com/linecorp/armeria/server/AbstractBindingBuilder.java
```diff
@@ -53,6 +59,8 @@
     private Set<HttpMethod> methods = ImmutableSet.of();
     private Set<MediaType> consumeTypes = ImmutableSet.of();
     private Set<MediaType> produceTypes = ImmutableSet.of();
+    private List<RoutingPredicate<QueryParams>> paramPredicates = new ArrayList<>();
+    private List<RoutingPredicate<HttpHeaders>> headerPredicates = new ArrayList<>();
```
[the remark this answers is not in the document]
hyangtack: Oh, I missed them. Thanks!

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
```diff
@@ -119,14 +123,20 @@ private RouteCache() {}
...
                       Cache<RoutingContext, V> findCache,
-                      Cache<RoutingContext, List<V>> findAllCache) {
+                      Cache<RoutingContext, List<V>> findAllCache,
+                      Set<Route> noCacheRoutes) {
             this.delegate = requireNonNull(delegate, "delegate");
             this.routeResolver = requireNonNull(routeResolver, "routeResolver");
             this.findCache = requireNonNull(findCache, "findCache");
             this.findAllCache = requireNonNull(findAllCache, "findAllCache");
+
+            final Set<Route> newNoCacheRoutes = Collections.newSetFromMap(new IdentityHashMap<>());
```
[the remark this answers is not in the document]
hyangtack: Picked the first comment because the constructors of IdentityHashMap accept only size or Map. Thanks! :-)

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
hyangtack: @ikhoon I think you're right. I didn't think about that because there was no `findAll` method when I started this PR. :-( Let me figure out how I handle it. Thanks!

hyangtack:
@ikhoon Would you have a look on this commit? :-)
https://github.com/line/armeria/pull/2102/commits/c8c0b62f7f56270f370deb567c66972a7477f50a

`CachingRouter` caches the `Route`s which can be collected by matching cacheable parameters (path type, path, methods, consumes, produces). The non-cacheable parameters like the predicates for headers and/or query parameters will be evaluated later. 

/cc @trustin @minwoox

core/src/main/java/com/linecorp/armeria/server/RouteCache.java
hyangtack: @ikhoon Fixed. :-)

ikhoon:
> @ikhoon Would you have a look on this commit? :-)

Sure, Let me look at it!

ikhoon:
While reviewing your code in the cafe☕️, my laptop battery was gone. 😱 Let me check it soon.

ikhoon:
I'm not sure that I understand ambiguous routes and `CachingRoutingContext` logic. 😭
But I found the following test case fail though it is not related to the cache. 
```java
@Test
void headerParamMatchWithRoutingDecorator() {
    AtomicBoolean decorated = new AtomicBoolean(false);
    Server server = Server
            .builder()
            // want to decocrate only when header contains "header-match"
            .decorator(Route.builder()
                            .pathPrefix("/")
                            .matchesHeaders("header-match").build(),
                       (delegate, ctx, req) -> {
                           decorated.set(true);
                       })
            .annotatedService("/", new Object() {
                @MatchesHeader("header-match")
                @Get("/foo")
                public HttpResponse foo() {
                    return HttpResponse.of(HttpStatus.OK);
                }
            })
            .build();
    server.start().join();
    final WebClient client = WebClient.of("http://127.0.0.1:" + server.activeLocalPort());
    final AggregatedHttpResponse response = client.get("/foo").aggregate().join();
    assertThat(response.status()).isEqualByComparingTo(HttpStatus.NOT_FOUND);
    assertThat(decorated).isFalse(); // assert fail, decorated is true, should not be set true.
}
```

hyangtack:
@ikhoon Thanks. Let me have a look.

hyangtack:
@ikhoon I found `Route` for the decorator didn't have a method. It worked when I added `.methods(HttpMethod.GET)`. Is it a bug? Seems that all methods should be accepted if no method was specified.

ikhoon:
> I found Route for the decorator didn't have a method. It worked when I added .methods(HttpMethod.GET). Is it a bug? 

I guess it is not a bug. 😅 Currently, A decorator could be routed under a specific path and accept every methods. It is useful when authorizing some paths. If the empty method of route seems weird, I think we could add all methods when building a route without specifying methods. 

https://line.github.io/armeria/server-decorator.html#decorating-multiple-services-by-path-mapping

trustin:
I think the initial reason is it's easy and fast to use `.isEmpty()` to check is the route matches all methods. We could return all methods and add a method like `matchesAllMethods()`?

ikhoon:
> I think the initial reason is it's easy and fast to use .isEmpty() to check is the route matches all methods. We could return all methods and add a method like matchesAllMethods()?

Based on the @trustin 's review, we could find proper route with {header,param} matching by adding all known methods if method is empty.

```
diff --git a/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java b/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
index 5bd5a9611..0eab6da4f 100644
--- a/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
+++ b/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
@@ -96,10 +96,6 @@ final class DefaultRoute implements Route {
             return RoutingResult.empty();
         }
 
-        if (methods.isEmpty()) {
-            return builder.build();
-        }
-
         if (!methods.contains(routingCtx.method())) {
             // '415 Unsupported Media Type' and '406 Not Acceptable' is more specific than
             // '405 Method Not Allowed'. So 405 would be set if there is no status code set before.
diff --git a/core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java b/core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
index 28ce4955e..a9e5fccd5 100644
--- a/core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
@@ -467,6 +467,9 @@ public final class RouteBuilder {
             throw new IllegalStateException("Must set methods if consumes or produces is not empty." +
                                             " consumes: " + consumes + ", produces: " + produces);
         }
+        if (methods.isEmpty()) {
+            methods = HttpMethod.knownMethods();
+        }
         return new DefaultRoute(pathMapping, methods, consumes, produces, paramPredicates, headerPredicates);
     }
``` 

I think it is inefficient because maybe most of the routers defined by users don't have {header, param} predicates. Do we need special optimization here? 

For example:
```java
if (matchesAllMethods(methods) && !routingCtx.requiresMatchingParamsPredicates() && 
    !routingCtx.requiresMatchingHeadersPredicates()) {
    return builder.build();
```

hyangtack:
@ikhoon @trustin Sorry, I mis-understood the problem. I thought the decorator didn't work when I didn't specify the `method` of its `Route`. So I asked you whether it was a bug.  😅 

> I think it is inefficient because maybe most of the routers defined by users don't have {header, param} predicates. Do we need special optimization here?

I guess `consumes` and `produces` also have the same problem. So it seems better to execute the rest of the code. Anyway, let me make a change today.
Thanks, guys! 👍

trustin:
Still LGTM. Thanks @hyangtack :smile:

core/src/main/java/com/linecorp/armeria/server/DefaultRoutingContext.java
```diff
@@ -178,10 +202,13 @@ static boolean equals(RoutingContext self, @Nullable Object obj) {
 
         final RoutingContext other = (RoutingContext) obj;
         return self.virtualHost().equals(other.virtualHost()) &&
-               self.method().equals(other.method()) &&
+               self.method() == other.method() &&
                self.path().equals(other.path()) &&
-               Objects.equals(self.contentType(), other.contentType()) &&
-               self.acceptTypes().equals(other.acceptTypes());
```
ikhoon: Question: I wonder why `contentType` and `acceptTypes` are removed? 😀

ikhoon: Do we ignore them from the cache keys?

hyangtack: Because I added `headers()` instead. `contentType` and `acceptTypes`  are included in the headers so I think it would be okay to remove them. Please let me know if you have any concern about that. :-)

ikhoon: > Because I added headers() instead.

I didn't catch it. 😅

> contentType and acceptTypes are included in the headers

That sounds reasonable to me. 👍 LGTM

core/src/test/java/com/linecorp/armeria/server/RoutingContextTest.java
```diff
@@ -106,8 +106,9 @@ void testEquals() {
         ctx2 = new DefaultRoutingContext(virtualHost, "example.com",
                                          RequestHeaders.of(HttpMethod.GET, "/hello"), "/hello", "a=1", false);
 
-        assertThat(ctx1.hashCode()).isEqualTo(ctx2.hashCode());
-        assertThat(ctx1).isEqualTo(ctx2);
+        // Queries are different.
+        assertThat(ctx1.hashCode()).isNotEqualTo(ctx2.hashCode());
+        assertThat(ctx1).isNotEqualTo(ctx2);
```
ikhoon: Could you add an equal comparison for `ctx1.param = "a=1&b=1"` and `ctx2.param = "a=1&b=1"`? 😉

trustin:
Nice job, @hyangtack! :laughing:

minwoox:
🚀

ikhoon:
Awesome! 🌟

PR #2103, opened 2019-09-22
Attempts to remove special-casing in brave integration tests

adriancole:
There's one glitch left to analyze cc @trustin @kojilin 

See #2092

### DIFF
--- a/brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
+++ b/brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
@@ -19,6 +19,7 @@ package com.linecorp.armeria.client.brave;
 import javax.annotation.Nullable;
 
 import com.linecorp.armeria.client.ClientRequestContext;
+import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.RequestHeadersBuilder;
 import com.linecorp.armeria.common.RpcRequest;
 import com.linecorp.armeria.common.SerializationFormat;
@@ -26,7 +27,6 @@ import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.common.logging.RequestLogAvailability;
 import com.linecorp.armeria.internal.brave.SpanContextUtil;
-import com.linecorp.armeria.internal.brave.SpanTags;
 
 import brave.http.HttpClientAdapter;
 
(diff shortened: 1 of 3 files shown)


brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -74,7 +74,11 @@ public String path() {
         @Override
         @Nullable
         public String url() {
-            return SpanTags.generateUrl(ctx.log());
+            // This dies because lack of RequestLogAvailability.REQUEST_HEADERS
```
adriancole: This is the glitch. Note: the ServiceRequestContextAdapter works fine as-is

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -74,7 +74,11 @@ public String path() {
...
+            //return SpanTags.generateUrl(ctx.log());
+
+            // This doesn't die
+            return ((HttpRequest) ctx.request()).uri().toString();
```
adriancole: PS I don't mind doing this instead, just the casting seems accidental/awkward

kojilin: IIRC `ClientRequestContext#request` returns `RpcRequest` when it's RpcClient.

kojilin: IIRC `ctx.request` returns `RpcRequest` when it's a rpc client.

adriancole:
I see. anyway I am flexible on how. would be great to have this data
available without casting etc as it allows secondary sampling use case
client side. server side works already and is more important, but yeah this
is something we should try to formalize.

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
adriancole: ok so maybe we should add a test which ends up returning an RpcRequest to ensure we don't break it. Do you feel up to adding a commit for that?

kojilin: ok,  I will make one in another PR.

trustin: I'm thinking of splitting `request()` into `request(): HttpRequest` and `rpcRequest(): RpcRequest`, which is similar to @anuraaga's suggestion. Please stay tuned.

trustin:
Closing in favor of #2120

trustin:
Thanks @adriancole for the idea!

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
```diff
@@ -94,6 +79,16 @@ public void setupServer() {
         server.setProtocols(protocols);
     }
 
+    @Before
```
adriancole: @trustin I think this is the secret of the incas for your PR

PR #2104, opened 2019-09-23
Makes health configuration optional

adriancole:
A previous change fixed health status mapping, but in the process made
it required to have health autoconfiguration. This fixes the below
error:

```
Parameter 3 of method actuatorServerConfigurator in com.linecorp.armeria.spring.actuate.ArmeriaSpringActuatorAutoConfiguration required a bean of type 'org.springframework.boot.actuate.health.HealthStatusHttpMapper' that could not be found.
```


trustin:
Thanks @adriancole !

PR #2107, opened 2019-09-24
Make Dropwizard Metrics and Prometheus optional for Spring Boot integration

### LINKED ISSUE #2106: Accidental strict dep on codahale metrics
ArmeriaConfigurationUtil calls `meterRegistry instanceof DropwizardMeterRegistry` which results in `java.lang.ClassNotFoundException: com.codahale.metrics.json.MetricsModule` if you excluded codahale. I think we want to do Class.forName or similar to guard instead.

trustin:
Motivation:

A user may not want to pull in Dropwizard Metrics or Prometheus at all.

Modifications:

- Access anything related with Dropwizard or Prometheus only when the related classes are available at runtime.
- Make `io.dropwizard.metrics:metrics-json` and `io.micrometer:micrometer-registry-prometheus` optional.

Result:

- Fixes #2106
- A user can exclude `io.dropwizard.metrics:metrics-core` as well as
  `metrics-json`.
- A user can exclude `io.micrometer:micrometer-registry-prometheus`

### DIFF
--- a/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java
+++ b/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java
@@ -135,7 +135,7 @@ public final class ArmeriaConfigurationUtil {
         server.meterRegistry(meterRegistry);
 
         if (settings.isEnableMetrics() && !Strings.isNullOrEmpty(settings.getMetricsPath())) {
-            final boolean hasPrometheus = hasClasses(
+            final boolean hasPrometheus = hasAllClasses(
                     "io.micrometer.prometheus.PrometheusMeterRegistry",
                     "io.prometheus.client.CollectorRegistry");
 
@@ -147,7 +147,7 @@ public final class ArmeriaConfigurationUtil {
             }
 
             if (!addedPrometheusExposition) {
-                final boolean hasDropwizard = hasClasses(
+                final boolean hasDropwizard = hasAllClasses(
                         "io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry",
                         "com.codahale.metrics.MetricRegistry",
                         "com.codahale.metrics.json.MetricsModule");
@@ -169,7 +169,7 @@ public final class ArmeriaConfigurationUtil {
         }
     }
 
-    private static boolean hasClasses(String... classNames) {
+    private static boolean hasAllClasses(String... classNames) {
         for (String className : classNames) {
             try {
                 Class.forName(className, false, ArmeriaConfigurationUtil.class.getClassLoader());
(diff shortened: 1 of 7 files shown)


trustin:
@adriancole PTAL

adriancole:
testing now.. the code looks correct

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java
```diff
@@ -161,22 +159,42 @@ public static void configureServerWithArmeriaSettings(ServerBuilder server, Arme
...
-                                return HttpResponse.of(HttpStatus.OK, MediaType.JSON_UTF_8,
-                                                       objectMapper.writeValueAsBytes(dropwizardRegistry));
-                            }
-                        });
+            } else {
+                // Need to use reflection to access the classes related with Dropwizard Metrics
+                // because it's an optional dependency.
+                try {
+                    final ClassLoader classLoader = ArmeriaConfigurationUtil.class.getClassLoader();
+                    final Class<?> dropwizardMeterRegistryClass = Class.forName(
```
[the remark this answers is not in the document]
adriancole: IIRC you will need to have the code in a separate type (even static inner class should be fine)

ex https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/internal/Platform.java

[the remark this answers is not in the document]
adriancole: if the type was unstable reflection would still be best as otherwise you can pin library versions. however I think this type will be stable.

adriancole:
verified https://github.com/openzipkin/zipkin/pull/2818#issuecomment-534518339

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/PrometheusSupport.java
```diff
@@ -0,0 +1,68 @@
...
+
+final class PrometheusSupport {
+
+    static boolean addExposition(ArmeriaSettings settings, ServerBuilder server,
+                                 MeterRegistry meterRegistry) {
+
+        final String metricsPath = settings.getMetricsPath();
+        assert metricsPath != null;
+
+        for (;;) {
```
ikhoon: Question: Is there any chance this fall into a deadlock? Just in case:
```
fooCompositeMeterRegistry 
-(registries)-> [barCompositeMeterRegistry, ...]
-(registries)-> [fooCompositeMeterRegistry, ...]
```

trustin: I guess a user has a bigger problem in that case - adding a meter will recurse infinitely.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java
```diff
@@ -192,6 +169,17 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req)
         }
     }
 
+    private static boolean hasClasses(String... classNames) {
```
minwoox: nit: perhaps `hasAllClasses` is better to understand?

trustin: Sure. Renamed.

trustin:
Thanks for revieweing!

Issue #2105, opened 2019-09-23
Add sampling to the exceptions without stack traces.


trustin:
In production, it is currently recommended to run Armeria servers with the `verboseExceptions` flag disabled to reduce the overhead of `Throwable.fillInStackTrace()`. However, this makes it hard for users to determine the root cause of the exceptions such as `ResponseTimeoutException` or `AbortedStreamException`.

Instead of choosing between all or nothing, we could define some expression for defining a `Sampler`, e.t. `random=0.1` or `rateLimited=1`

trustin:
The sensible default would be like `rateLimited=10` which allows 10 `fillInStackTrace()` a second.

trustin:
Closed via #2111

PR #2108, opened 2019-09-24
Add `Sampler.of(String specification)`

trustin:
Motivation:

Like we can specify a string to instantiate a `Backoff`, we could also
specify a specification string to instantiate a `Sampler`, which will be
useful when sampling policy must be configurable.

Modifications:

- Add `Sampler.of()`
- Miscellaneous:
  - Minor documentation clean-up of `Backoff.of()`

Result:

- A user can create a `Sampler` from a specification string, which means
  sampling policy settings can be configurable from a configuration
  file, a system property or other external sources.
- Can start to work on #2105

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/Sampler.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/Sampler.java
@@ -29,8 +29,6 @@
  */
 package com.linecorp.armeria.common.util;
 
-import static java.util.Objects.requireNonNull;
-
 /**
  * Sampler is responsible for deciding if a particular trace should be "sampled", i.e. whether the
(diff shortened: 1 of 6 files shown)


trustin:
Thanks for reviewing.

core/src/main/java/com/linecorp/armeria/common/util/Sampler.java
```diff
@@ -82,6 +82,38 @@
         return cast;
     }
 
+    /**
+     * Returns a {@link Sampler} that is configured as specified in the given {@code specification} string.
```
adriancole: @narayaruna this is a change that might be interesting to you. Armeria is using specification strings to do sampling not just of tracing, but also whether or not to keep full stack traces.

PR #2110, opened 2019-09-25
Fix memory leak in `HealthCheckService`

trustin:
Motivation:

When a long-polling health check request times out, the data structure
that holds a pending response is not cleaned up.

Modifications:

Make sure to clean up the pending response.

Result:

The memory leak is gone.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
@@ -15,8 +15,6 @@
  */
 package com.linecorp.armeria.server.healthcheck;
 
-import java.util.ArrayDeque;
-import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ThreadLocalRandom;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java
```diff
@@ -285,16 +288,19 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
...
                     final PendingResponse pendingResponse = new PendingResponse(method, future, timeoutFuture);
-                    if (isHealthy) {
-                        pendingUnhealthyResponses.add(pendingResponse);
-                    } else {
-                        pendingHealthyResponses.add(pendingResponse);
-                    }
+                    pendingResponses.add(pendingResponse);
+                    timeoutFuture.addListener((FutureListener<Boolean>) f -> {
+                        synchronized (healthCheckerListener) {
+                            pendingResponses.remove(pendingResponse);
```
ikhoon: Question: Don't we have to synchronize the following block?
https://github.com/line/armeria/blob/69acc04e620ce52445ac23d1300508f6c95815c3/core/src/main/java/com/linecorp/armeria/server/healthcheck/HealthCheckService.java#L438-L444

ikhoon: This is not a problem. I misunderstood the logic. 😢

PR #2112, opened 2019-09-25
Move down the GitHub corner on mobile view

trustin:
Motivation:

On mobile view, the GitHub corner covers the hamburger button when the
menu is open.

Modification:

Move down the GitHub corner so that it never overlaps with the hamburger
button regardless whether the menu is open or not.

Result:

Better mobile web site experience


trustin:
After the change:
image

PR #2114, opened 2019-09-25
Prettier Java version-specific log message

trustin:


trustin:
Before:

```
Using version specific APIs from JavaVersionSpecific
or
Using version specific APIs from Java9VersionSpecific
```

After:

```
<nil>
or
Using the APIs optimized for: Java 9+
```

core/src/main/java/com/linecorp/armeria/internal/JavaVersionSpecific.java
```diff
@@ -29,12 +29,14 @@
...
 
-    private static Logger logger = LoggerFactory.getLogger(JavaVersionSpecific.class);
+    private static final Logger logger = LoggerFactory.getLogger(JavaVersionSpecific.class);
 
     private static final JavaVersionSpecific CURRENT = CurrentJavaVersionSpecific.get();
 
     static {
-        logger.info("Using version specific APIs from {}", CURRENT.getClass().getSimpleName());
+        if (CURRENT.getClass() != JavaVersionSpecific.class) {
+            logger.info("Using the APIs optimized for: {}", CURRENT.name());
```
ikhoon: 👍

Issue #1704, opened 2019-04-08
Umbrella issue for 1.0.0


trustin:
- [ ] More per-service settings with fluent route mapping. #1649
  - [x] This revamps the routing and path mapping API to allow a non-annotated service to accept certain methods or media types.
  - [ ] Related: More convenient decoration #199 (w.r.t specifying multiple decorators)
  - [x] Provide a way to configure content previewing to the specific service #1619 
  - [ ] HTTP header/parameter-based path mapping #1207 
- [x] Separate messages for AggregatedHttpMessage request / response #1702 
- [x] Add a StreamMessage subscription option that notifies cancellation to onError() #1700
- [x] Add `boolean shutdownOnStop` parameter to `ServerBuilder.blockingTaskExecutor()` #1685 
- [ ] Immutable HttpHeaders and Cookies #1567 
- [ ] Make sure all safely retriable exceptions are wrapped by `UnprocessedRequestException`
  - [x] Change to return `UnprocessedRequestException` in `RetryingClient` #1653 
  - [x] Wrap socket connect failures and more with `UnprocessedRequestException`
  - [x] Make sure `UnprocessedRequestException` always has a cause.
  - [ ] Add a section about UnprocessedRequestException in RetryingClient documentation #1395 
- [ ] Provide a way to build a generic `RetryStrategy`. #1634 
  - This may replace some static factory methods.
- [ ] Make it more configurable to extract the client IP address behind reverse-proxies #1631 
- [x] Decorate a service by path mapping #582 
- [ ] Create an abstraction for a LoggingContext containing request properties #1510 
- [x] Difficult to request from an absolute URL with HttpClient #1143 
- [ ] Provide an extension point in LoggingClient/Service for constructing a log message #1103 
  - Do a final review on the current way of message sanitization in `LoggingClient/Service`. Can we make it any better, especially in terms of performance and flexibility?
- [ ] EndpointGroup design issues #1084 
  - [ ] Disparity between HttpHealthCheckedEndpoint and HttpClient can be confusing. #1913
  - [ ] HealthCheckedEndpointGroup should probably wait for connectTimeout before throwing exception. #1910
- [ ] A user may get the wrong RequestContext when contexts are nested #1083 
- [ ] Allow resetting the request timeout with the same value #740
- [ ] Revisit the `blockingTaskExecutor`-related API
  - [ ] Make a `blockingTaskExecutor` a `SchedulerExecutorService`. #2269
[...]

PR #2120, opened 2019-09-27
Split `RequestContext.request()` into `request()` and `rpcReque…

### LINKED ISSUE #2089: Make it possible to decorate clients and servers at URL context

trustin:
Motivation:

It is currently impossible to get both `HttpRequest` and `RpcRequest`
from the current `RequestContext`, which makes it hard to retrieve the
information which should be available immediately otherwise.

Modifications:

- Make `RequestContext.request()` always return an `HttpRequest`.
- Make `RequestContext.updateRequest()` always update an `HttpRequest`.
  - It returns `void` now because it never fails, unless it's `null`.
- Add `RequestContext.rpcRequest()` and `updateRpcRequest()`.
- `RequestContext.newDerivedContext()` now requires both `HttpRequest`
  and `RpcRequest` when a user wants to override the request.
- `RequestLogBuilder.requestContent(Object, Object)` now calls
  `RequestContext.updateRpcRequest()` when the context has no
  `RpcRequest` set.
  - I'm not completely sure this is the best way, but we do not have to
    call `updateRpcRequest()` explicitly in many places.
- Cleaned up distributed tracing code by taking advantage of the above
  changes.
  - `armeria-zipkin` has been removed for further clean-up.

Result:

- A user can get the current `HttpRequest` and `RpcRequest` from
  `RequestContext`.
- Fixes #2089


trustin:
@adriancole PTAL :wink:

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
```diff
@@ -168,18 +168,17 @@ public void supportsPortableCustomization() throws Exception {
...
+                                       customizer.tag("http.url", ctx.request().uri().toString());
+                                   }
                                }
 
                                @Override
                                public <T> void response(HttpAdapter<?, T> adapter, T res,
                                                         Throwable error,
                                                         SpanCustomizer customizer) {
                                    super.response(adapter, res, error, customizer);
-                                   // TODO: make this possible at request scope https://github.com/line/armeria/issues/2089
```
trustin: Look ma, the TODO is gone! :laughing:

trustin:
This changeset opens the new possibility - the removal of `req` parameter in many places, e.g. `Service.serve()` and `Client.execute()`. I'm not fully sure if this is a good idea though because `ctx.request()` and `ctx.rpcRequest()` are mutable properties and thus it is not safe to access them after delegation.

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
```diff
@@ -153,7 +153,7 @@ public void redirect() throws Exception {
     @Override
     @Test
     public void supportsPortableCustomization() throws Exception {
```
adriancole: you should be able to delete this method override now!

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceIntegrationTest.java
```diff
@@ -101,17 +101,16 @@ public void supportsPortableCustomization() throws Exception {
             public <T> void request(HttpAdapter<T, ?> adapter, T req, SpanCustomizer customizer) {
                 customizer.tag("context.visible", String.valueOf(currentTraceContext.get() != null));
                 customizer.tag("request_customizer.is_span", String.valueOf(customizer instanceof brave.Span));
+                if (req instanceof ServiceRequestContext) {
```
adriancole: you should be able to delete supportsPortableCustomization() override now!

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -136,26 +136,51 @@ static ClientRequestContext of(RpcRequest request, URI uri) {
         return ClientRequestContextBuilder.of(request, uri).build();
     }
 
+    /**
+     * {@inheritDoc} For example, when you send an RPC request, this method will return {@code null} until
+     * the RPC request is translated into an HTTP request.
```
adriancole: informative!

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
trustin: After un-overriding, I had to set `RequestContextCurrentTraceContext` to `ITHttp.currentTraceContext` to make
`supportsPortableCustomization()` pass. Should I set it only for
`supportsPortableCustomization()` rather than for all tests?

adriancole: Please put a TODO(adrian) to look into it, that way this doesn't block your change. I can't look into it right now, but will and this is still massive improvement

adriancole:
I will pull this locally now also.

adriancole:
@trustin if you do this in the client and server tests, everything looks splendid

```diff
--- a/brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceIntegrationTest.java
+++ b/brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceIntegrationTest.java
@@ -26,6 +26,7 @@ import javax.annotation.Nullable;
 
 import org.junit.After;
 import org.junit.AssumptionViolatedException;
+import org.junit.Before;
 import org.junit.Test;
 
 import com.linecorp.armeria.common.HttpMethod;
@@ -38,10 +39,7 @@ import com.linecorp.armeria.server.Server;
 import com.linecorp.armeria.server.ServerBuilder;
 
 import brave.Tracing;
-import brave.Tracing.Builder;
-import brave.propagation.CurrentTraceContext;
 import brave.propagation.StrictScopeDecorator;
-import brave.sampler.Sampler;
 import brave.test.http.ITHttpServer;
 
 public class BraveServiceIntegrationTest extends ITHttpServer {
@@ -49,16 +47,14 @@ public class BraveServiceIntegrationTest extends ITHttpServer {
     @Nullable
     private Server server;
 
-    // Hide currentTraceContext in ITHttpServer
-    private final CurrentTraceContext currentTraceContext =
-            RequestContextCurrentTraceContext.builder()
-                                             .addScopeDecorator(StrictScopeDecorator.create())
-                                             .build();
-
+    @Before
     @Override
-    protected Builder tracingBuilder(Sampler sampler) {
-        return super.tracingBuilder(sampler)
-                    .currentTraceContext(currentTraceContext);
+    public void setup() throws Exception {
+        currentTraceContext =
+            RequestContextCurrentTraceContext.builder()
+                .addScopeDecorator(StrictScopeDecorator.create())
+                .build();
+        super.setup();
     }
 
     @Override
@@ -88,18 +84,6 @@ public class BraveServiceIntegrationTest extends ITHttpServer {
         server.start().join();
     }
 
-    @Override
-    public void supportsPortableCustomization() throws Exception {
-        // TODO(adriancole): Figure out why super.currentTraceContext has to be overridden.
-        final CurrentTraceContext oldSuperCurrentTraceContext = super.currentTraceContext;
-        super.currentTraceContext = currentTraceContext;
-        try {
-            super.supportsPortableCustomization();
-        } finally {
-            super.currentTraceContext = oldSuperCurrentTraceContext;
-        }
-    }
-
     @Override
     @Test
     public void notFound() {

```

kojilin:
Forget to add test for RPC, will you add one for `BraveIntegrationTest`? 
And I think we need to update RequestContext's request somewhere in UserClient or RPC client -> UserClient phase?

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogBuilder.java
```diff
@@ -183,6 +185,9 @@ void startRequest(Channel channel, SessionProtocol sessionProtocol, @Nullable SS
 
     /**
      * Sets the {@link RequestLog#requestContent()} and the {@link RequestLog#rawRequestContent()}.
+     * If the specified {@code requestContent} is a {@link RpcRequest} and
```
minwoox: nit: an {@link RpcRequest}

minwoox:
> And I think we need to update RequestContext's request somewhere in UserClient or RPC client -> UserClient phase? If not, above example's http.host will be null?

Yes I think we should set it in `THttpClientDelegate`.

minwoox:
Even though, we fill `HttpRequest` in `THttpClientDelegate`, if we use `RetryingRpcClient`, the original `ClientRequestContext` never has `HttpRequest`. I'm wondering if we need to consider this situation.

adriancole:
maybe rpcrequest fidelity is better a separate PR

anuraaga:
I think a recent related comment is this one

https://github.com/line/armeria/issues/1704#issuecomment-480726369

And I remember making a comment about possibly removing `RpcRequest` to simplify decorators. It might be good to give it another good think now - gRPC users use things like `ServerInterceptor` for RPC decoration which works well since it is fully aware of the RPC semantics. I think this is common for RPC to have specific aspects hard to abstract. Could it simplify many of the framework aspects if we move the current `RpcRequest` into the thrift layer out of the common framework?

trustin:
> Could it simplify many of the framework aspects if we move the current `RpcRequest` into the thrift layer out of the common framework?

Yeah, it will simplify things a lot (generics be gone!). However, I'm somewhat concerned about what will happen to some decorators which share common code between RPC-level and HTTP-level, such as `RetryingRpcClient` and `RetryingHttpClient`. We'll have to move various RPC-level decorators to `armeria-thrift`, which may result in a complex dependency web in the long term, as we introduce more modules that integrate with 3rd party libraries such as resilence4j.

Nevertheless, I think it's worth a shot - it may not be that complex after all if we refactor things carefully so that orthogonality between modules are preserved.

trustin:
Thanks for reviewing, folks!

PR #2123, opened 2019-09-28
Add a JMH benchmark for `ThriftMessageClassFinder`

moonchanyong:
Motivation:

add benchmarking result in line of PR: #2055 

Modification:

* add benchmarking result about between using Thread.getStackTrace, Stackwalker, and Stackwalker with RETAIN_CLASS_REFERENCE option

Result:

* Throughput of Stackwalker with RETAIN_CLASS_REFERENCE option is better than others 
* but is recoded high error

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/common/thrift/text/StackTracingBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/common/thrift/text/StackTracingBenchmark.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.common.thrift.text;
+
+import java.security.Permission;
+import java.util.function.Supplier;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.infra.Blackhole;
+
+import com.linecorp.armeria.thrift.services.HelloService.hello_args;
+
+/**
+ * Compare performance of {@link Thread#getStackTrace} vs StackWalker vs StackWalker with Option
+ * find getCurrentThriftMessageClass.
+ *
+ * <p>20190928 Macbook Pro 2018 2.2 GHz Intel Core i7
+ * <pre>
+ * # Run complete. Total time: 00:25:10
+ *
+ * Benchmark                                          Mode  Cnt       Score       Error  Units
+ * StackTracingBenchmark.defaultStackTrace           thrpt   25   65969.836 ±  4714.734  ops/s
+ * StackTracingBenchmark.usingStackWalker            thrpt   25  131919.125 ±  7730.472  ops/s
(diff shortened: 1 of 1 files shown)


trustin:
Thanks a lot! Could you paste the result?

trustin:
I think you must pass the value returned by .get().

trustin:
The result after adding `.get()` - 5 times faster!

```
Benchmark                                          Mode  Cnt       Score       Error  Units
StackTracingBenchmark.defaultStackTrace           thrpt    5  104310.565 ±  5994.780  ops/s
StackTracingBenchmark.usingStackWalker            thrpt    5  269399.892 ±  6504.989  ops/s
StackTracingBenchmark.usingStackWalkerWithOption  thrpt    5  538596.277 ± 96659.058  ops/s
```

moonchanyong:
I forgot to add get method after remove proxy method for calling on TBase implemented class.
I pushed again

benchmarks/src/jmh/java/com/linecorp/armeria/thrift/StackTracingBenchmark.java
```diff
@@ -0,0 +1,116 @@
...
+        });
+
+        stackWalkingThriftMessageClassFinder = (Supplier<Class<?>>) stackWalkerConstructor.newInstance();
+
+        System.setSecurityManager(SM);
+    }
+
+    @TearDown
+    public void stopServer() {
+    }
```
trustin: Perhaps we do not need this method?

benchmarks/src/jmh/java/com/linecorp/armeria/thrift/StackTracingBenchmark.java
```diff
@@ -0,0 +1,116 @@
...
+    private Supplier<Class<?>> defaultThriftMessageClassFinderfault;
+    private Supplier<Class<?>> stackWalkingThriftMessageClassFinder;
+    private Supplier<Class<?>> stackWalkingFinderWithRetainOption;
+
+    @Setup
+    public void makeFinder() throws Exception {
+        final Class<?> defaultThriftMessageClassFinderClass =
+                Class.forName("com.linecorp.armeria.common.thrift.text.DefaultThriftMessageClassFinder");
+        final Class<?> stackWalkingThriftMessageClassFinderClass =
+                Class.forName("com.linecorp.armeria.common.thrift.text.StackWalkingThriftMessageClassFinder");
```
trustin: Perhaps we do not use reflection if we put this benchmark class in the same package?

benchmarks/src/jmh/java/com/linecorp/armeria/thrift/StackTracingBenchmark.java
```diff
@@ -0,0 +1,116 @@
...
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.infra.Blackhole;
+
+import com.linecorp.armeria.thrift.services.HelloService.hello_args;
+
+/**
+ * Compare performance of Thread.getStackTrace vs StackWalker(reflected) vs StackWalker(reflected) with Option
```
ikhoon: nit: `{@link Thread.getStackTrace}` and others?

benchmarks/src/jmh/java/com/linecorp/armeria/thrift/StackTracingBenchmark.java
moonchanyong: I didn't think that. thanks for your advice.

benchmarks/src/jmh/java/com/linecorp/armeria/common/thrift/text/StackTracingBenchmark.java
```diff
@@ -0,0 +1,95 @@
...
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.infra.Blackhole;
+
+import com.linecorp.armeria.thrift.services.HelloService.hello_args;
+
+/**
+ * Compare performance of {@link Thread#getStackTrace} vs StackWalker vs StackWalker with Option
```
trustin:
```
Compare the performance between {@link Thread#getStackTrace()}, {@code StackWalker}
and {@code StackWalker} with the {@code RETAIN_CLASS_REFERENCE} option when finding
the current Thrift message class.
```

PR #2124, opened 2019-09-28
Add `SslContext` validation

### LINKED ISSUE #1844: Cryptic error message when TLS key store password is incorrect.
Slack thread: https://line-armeria.slack.com/archives/C1NGPBUH2/p1560738924028700

When a user starts a TLS server without correct key store password, the server has to refuse to start. Currently, it starts up without an error and fails to serve a request with a cryptic error message such as `NO_CERTIFICATE_SET` and `No available authentication scheme`.

/cc @adriancole

jyblue:
Motivation:

In some case, servers can start with miss-configured `SslContext` and
fail when users request ssl/tls connections.
If servers refuse to start in that case, users can fix their
configuration before severs start.

Modifications:

Modify `VirtualHostBuilder.validateSslContext()`

Result:

Fixes #1844


jyblue:
Hello, here is my draft.
And I'm not sure I'm on the right way to go. Please see my questions.
1. If `SSLEngine` is created without exceptions, is it guaranteed that `SslContext` is well-configured?
I read some articles including official docs but couldn't find answer.
2. If not guaranteed, **what configurations** do I have to validate manually?
I added some test cases and it seems that most of configurations are already validated by other classes before `validateSslContext()` method is hit.
3. I cannot reproduce [initial bug report](https://github.com/line/armeria/issues/1844#issuecomment-502589823) in my **test code**. Because when I try to load `KeyStore` without proper key, it throws a exception, which means the server fails to start. How do I create `KeyStore` with invalid key?

Thank you in advance :bowing_man:

jyblue:
One more question,
I added `JKS` file to resources directory for test purpose and source code contains its password.
Is it okay with any code security issue?

trustin:
@jyblue wrote:

> 1. If `SSLEngine` is created without exceptions, is it guaranteed that `SslContext` is well-configured?

I think so. If not, we can add more checks later.

> 3. I cannot reproduce [initial bug report](https://github.com/line/armeria/issues/1844#issuecomment-502589823) in my **test code**. Because when I try to load `KeyStore` without proper key, it throws a exception, which means the server fails to start. How do I create `KeyStore` with invalid key?

You have to try to reproduce the problem with 1) a PEM format private key and 2) OpenSSL enabled (which is enabled by default, but you'll have to check using `Flags.useOpenSsl()`).

> I added dummy `JKS` file to resources directory for test purpose and source code contains its password. Is it okay with any code security issue?

I don't think it has any security implication if you used a self-signed certificate and if it's not encrypted with your personal password. :smile:

anuraaga:
We have validation of created contexts here

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/SslContextUtil.java

I think the new checks here look the same as what we do so maybe they don't help with the issue. I guess we need to reproduce the error case we first to know for sure.

jyblue:
@anuraaga Thanks for good hint.

I think I found the reason.
In this case, `null` is given to `KeyStore.load()` as a password parameter and `KeyStore.load()` didn't perform password integrity check.
https://github.com/line/armeria/blob/45c1633e1cdea43fcfc729a03739dcdb8ad0b49c/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java#L500

> https://docs.oracle.com/javase/7/docs/api/java/security/KeyStore.html 
A password may be given to unlock the keystore , or to check the integrity of the keystore data. **If a password is not given for integrity checking, then integrity checking is not performed.**


If steps reproduced,
```
KeyStore.load(keyStore, null);  // not performed password check
...
// server starts
// wait for ssl/tls request
...
KeyStore.getEntry(password); // password check, exception occurs and server fails
```

I will write a test case to reproduce it.

trustin:
Thanks @jyblue! Looks like we are going into the right direction. Please keep us posted.

jyblue:
Hello, here is my research and code updated. PTAL :grinning: 

**Test conclusion**
If key store password is not given to `PKCS12` key store,  
a server is built, running without exceptions, but fails when ssl/tls request come.

**Internal cause**
When `netty.SslHandler` is trying to decode client's first packet, a exception occurs.
I think `SslHandler` doesn't have a valid access to key store (not for sure).
https://github.com/netty/netty/blob/d8b1a2d93f556a08270e6549bf7f91b3b09f24bb/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L1329

**Validation**
I added validation method which tries dummy ssl/tls request,   
and it produces same exception as original error.
```
failed to validate SSL/TLS configuration : javax.net.ssl.SSLHandshakeException: error:100000ae:SSL routines:OPENSSL_internal:NO_CERTIFICATE_SET
java.lang.RuntimeException: failed to validate SSL/TLS configuration : javax.net.ssl.SSLHandshakeException: error:100000ae:SSL routines:OPENSSL_internal:NO_CERTIFICATE_SET
	at com.linecorp.armeria.server.VirtualHostBuilder.build(VirtualHostBuilder.java:843)
	at com.linecorp.armeria.server.ServerBuilder.build(ServerBuilder.java:1432)
```

**Note**
`JKS` key store doesn't produce any exception when key store password is not given.

FYI
Bug reproduction and detail report
https://github.com/jyblue/ArmeriaIssueReproduction1844

trustin:
Please remove the WIP status if it's ready for reviews. :wink:

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+                                     })
+                                     .build();
+    }
+
+    @Test
+    void testPkcs12KeyStoreWithNullPassword() throws Exception {
+        // Dummy keystore generation
+        // keytool -genkeypair -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore keystore.p12 -validity 3650
+        // keypassword = keystorepassword = password
+        assertThrows(RuntimeException.class, () -> {
```
trustin: Could you also check the exception message?

```java
assertThatThrownBy(() -> ....)
        .isInstanceOf(RuntimeException.class)
        .hasMessageContaining(...);
```

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+            final TrustManagerFactory tmf
+                    = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            tmf.init(keyStore);
+
+            final Server server
+                    = new ServerBuilder().service("/", (ctx, res) -> HttpResponse.of(HttpStatus.OK))
+                                         .tls(kmf, sslContextBuilder -> {
+                                             sslContextBuilder.keyManager(kmf);
+                                             sslContextBuilder.trustManager(tmf);
+                                         })
```
trustin: I don't think you need to set the `TrustManagerFactory` unless you want to authenticate clients. You can just pass a no-op customizer, e.g. `sslContextBuilder -> {}`.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+            keyStore.load(getClass().getResource("keystore.p12").openStream(),
+                          keyPass != null ? keyPass.toCharArray() : null);
+
+            final KeyManagerFactory kmf
+                    = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+            kmf.init(keyStore, "password".toCharArray());
+
+            final TrustManagerFactory tmf
+                    = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            tmf.init(keyStore);
```
trustin: No need to create this.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+        final KeyStore keyStore = KeyStore.getInstance("JKS");
+        keyStore.load(getClass().getResource("keystore.jks").openStream(),
+                      keyPass != null ? keyPass.toCharArray() : null);
+
+        final KeyManagerFactory kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+        kmf.init(keyStore, "keypassword".toCharArray());
+
+        final TrustManagerFactory tmf
+                = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+        tmf.init(keyStore);
```
trustin: No need to create this.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+        final Server server
+                = new ServerBuilder().service("/", (ctx, res) -> HttpResponse.of(HttpStatus.OK))
+                                     .tls(kmf, sslContextBuilder -> {
+                                         sslContextBuilder.keyManager(kmf);
+                                         sslContextBuilder.trustManager(tmf);
+                                     })
```
trustin: Could just pass a no-op customizer: `sslContextBuilder -> {}`.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+    @Test
+    void testPkcs12KeyStoreWithNullPassword() throws Exception {
+        // Dummy keystore generation
+        // keytool -genkeypair -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore keystore.p12 -validity 3650
+        // keypassword = keystorepassword = password
+        assertThrows(RuntimeException.class, () -> {
+            final String keyPass = null;
+            final KeyStore keyStore = KeyStore.getInstance("PKCS12");
+            keyStore.load(getClass().getResource("keystore.p12").openStream(),
+                          keyPass != null ? keyPass.toCharArray() : null);
```
trustin: You can just pass `null` and remove `keyPass`.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.TrustManagerFactory;
+
+import org.junit.jupiter.api.Test;
+
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+
+class ServerTlsTest {
```
trustin: `ServerTlsValidationTest`?

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+        /*
+         * Dummy keystore generation
+         * keytool -genkeypair -keyalg RSA -keysize 2048 -storetype JKS -keystore keystore.jks -validity 3650
+         * key store password = password
+         * key password = keypassword
+         */
+        final String keyPass = null;
+        final KeyStore keyStore = KeyStore.getInstance("JKS");
+        keyStore.load(getClass().getResource("keystore.jks").openStream(),
+                      keyPass != null ? keyPass.toCharArray() : null);
```
trustin: Could just pass `null` and remove `keyPass`.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+
+class ServerTlsTest {
+
+    @Test
+    void testJksKeyStoreWithNullPassword() throws Exception {
```
trustin: How about adding `@Ignore` (with comments on why) and making this test case verify if an exception is raised just like `testPkcs12KeyStoreWithNullPassword` does?

trustin:
You also have to release the two `SSLEngine`s you created during validation:

```
try {
    ...
} finally {
    ReferenceCountUtil.release(serverEngine);
    ReferenceCountUtil.release(clientEngine);
}
```

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -841,6 +852,33 @@ VirtualHost build() {
         return decorator != null ? virtualHost.decorate(decorator) : virtualHost;
     }
 
+    void validateSslContext(SslContext sslContext) throws Exception {
+        final SSLEngine serverEngine = sslContext.newEngine(ByteBufAllocator.DEFAULT);
+        serverEngine.setUseClientMode(false);
+        serverEngine.setNeedClientAuth(true);
```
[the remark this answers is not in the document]
anuraaga: I think a client could also be configured with a wrong keystore config which could be a reason for it.

But in that case, can we move this into `SslContextUtil` so it applies to both server and client?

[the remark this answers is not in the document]
trustin: Yeah we could also validate the client side if a `KeyManagerFactory` is set on the client side and TLS client auth is enabled. Maybe in a separate PR?

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -828,6 +831,14 @@ VirtualHost build() {
             }
         }
 
+        if (sslContext != null) {
+            try {
+                validateSslContext(sslContext);
+            } catch (Exception e) {
+                throw new RuntimeException("failed to validate SSL/TLS configuration : " + e.toString(), e);
```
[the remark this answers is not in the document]
jyblue: Fixed it.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -841,6 +852,33 @@ VirtualHost build() {
         return decorator != null ? virtualHost.decorate(decorator) : virtualHost;
     }
 
+    void validateSslContext(SslContext sslContext) throws Exception {
```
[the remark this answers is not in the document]
jyblue: Fixed it.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -841,6 +852,33 @@ VirtualHost build() {
...
+        serverEngine.setNeedClientAuth(true);
+
+        final SelfSignedCertificate ssc = new SelfSignedCertificate("foo.com");
+        final SslContext sslContextClient
+                = buildSslContext(() -> SslContextBuilder.forClient()
+                                                         .keyManager(ssc.certificate(), ssc.privateKey()),
+                                           sslContextBuilder -> {
+                    sslContextBuilder.keyManager(ssc.certificate(), ssc.privateKey());
+                    sslContextBuilder.trustManager(ssc.certificate());
+        });
```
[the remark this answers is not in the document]
jyblue: I tested both 
* with `sslContextBuilder.trustManager(InsecureTrustManagerFactory.INSTANCE);`
* with no-op customizer
and test results are same. So I used no-op customizer.

I thought it is necessary to set both key manager and trust manager. It was my misunderstanding :sob:

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -841,6 +852,33 @@ VirtualHost build() {
...
+
+        clientEngine.beginHandshake();
+        serverEngine.beginHandshake();
+
+        ByteBuffer clientPacket = ByteBuffer.allocate(clientEngine.getSession().getApplicationBufferSize());
+        clientEngine.wrap(clientPacket, clientPacket);
+        clientPacket.flip();
+
+        ByteBuffer serverPacket = ByteBuffer.allocate(serverEngine.getSession().getApplicationBufferSize());
+        serverEngine.unwrap(clientPacket, serverPacket);
```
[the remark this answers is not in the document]
jyblue: Question :
Can I use `appBuf` for both client and server? (which is allocated with client's app buf size).

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+        final KeyManagerFactory kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+        kmf.init(keyStore, "keypassword".toCharArray());
+
+        final TrustManagerFactory tmf
+                = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+        tmf.init(keyStore);
+
+        final Server server
+                = new ServerBuilder().service("/", (ctx, res) -> HttpResponse.of(HttpStatus.OK))
```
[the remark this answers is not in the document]
jyblue: Fixed it. Thank you.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Fixed it. Thank you.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Fixed it. Thank you.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Removed.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Removed

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/ServerTlsTest.java
jyblue: Updated with comments. PTAL :bowing_man:

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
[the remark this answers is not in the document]
jyblue: Changed it to `false`.

jyblue:
Here is my update.
* Rename `validateSslContext` to `validateKeyStoreNullPassword`
   Because it is confused with `VirtualHost.validateSslContext`
* Make `validateKeyStoreNullPassword` to be called inside of `tls(KeyManagerFactory, TlsCoustomizer)`
  Making the validation more close to user-side, so users can easily find their miss configured key store password

jyblue:
It seems `@Ignore` annotation on `testJksKeyStoreWithNullPassword()` doesn't work.
I will take a look and fix it.

jyblue:
I fixed `@Ignore` to `@Disabled`, and ready for next review.
Thank you in advance.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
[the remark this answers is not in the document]
trustin: I think so. TLS protocol has the same maximum application data size for both client and server side IIUC. I never saw an implementation that returns different capacities for client and server side.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -241,10 +245,55 @@ public VirtualHostBuilder tls(KeyManagerFactory keyManagerFactory,
...
+
+        tls(sslContext);
         return this;
     }
 
+    /**
+     * Validate key store which is used for building {@code sslContext}.
+     * If key store is initialized with null password, an exception will be raised.
+     */
+    private void validateKeyStoreNullPassword(SslContext sslContext) throws Exception {
```
trustin: - How about merging this method into `validateSslContext()`? We also should remove the `validateSslContext()` call in the `VirtualHost`'s constructor to avoid double validation, because now we validate the `SslContext` early in `tls(SslContext)`.
- Is it possible to make this method throw an `SSLException` rather than an `Exception`? If so, we will not have to catch and wrap it because `tls()` can throw an `SslException` already.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -241,10 +245,55 @@ public VirtualHostBuilder tls(KeyManagerFactory keyManagerFactory,
...
         requireNonNull(tlsCustomizer, "tlsCustomizer");
 
-        tls(buildSslContext(() -> SslContextBuilder.forServer(keyManagerFactory), tlsCustomizer));
+        final SslContext sslContext = buildSslContext(() -> SslContextBuilder.forServer(keyManagerFactory),
+                                                      tlsCustomizer);
+        try {
+            validateKeyStoreNullPassword(sslContext);
+        } catch (Exception e) {
+            throw new RuntimeException("failed to validate SSL/TLS configuration: " + e.toString(), e);
+        }
```
trustin: Once we merge this into `validateSslContext()`, we do not need to validate here but only in `tls(SslContext)`.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -241,10 +245,55 @@ public VirtualHostBuilder tls(KeyManagerFactory keyManagerFactory,
...
+        try {
+            serverEngine = sslContext.newEngine(ByteBufAllocator.DEFAULT);
+            serverEngine.setUseClientMode(false);
+            serverEngine.setNeedClientAuth(false);
+
+            final SelfSignedCertificate ssc = new SelfSignedCertificate("foo.com");
+            final SslContext sslContextClient =
+                    buildSslContext(() -> SslContextBuilder.forClient()
+                                                           .keyManager(ssc.certificate(), ssc.privateKey()),
+                                    sslContextBuilder -> {});
```
trustin: I guess we do not need to set the `KeyManagerFactory` for the client side context, because we do not perform client authentication.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -241,10 +245,55 @@ public VirtualHostBuilder tls(KeyManagerFactory keyManagerFactory,
...
+            clientEngine = sslContextClient.newEngine(ByteBufAllocator.DEFAULT);
+            clientEngine.setUseClientMode(true);
+
+            clientEngine.beginHandshake();
+            serverEngine.beginHandshake();
+
+            ByteBuffer appBuf = ByteBuffer.allocate(clientEngine.getSession().getApplicationBufferSize());
+            ByteBuffer packetBuf = ByteBuffer.allocate(clientEngine.getSession().getPacketBufferSize());
```
trustin: Missing `final`

core/src/test/java/com/linecorp/armeria/server/ServerTlsValidationTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+
+class ServerTlsValidationTest {
+
+    // Note: When key store password is not given to key store, it is expected that an exception occurs when
```
trustin: How about using `TODO`?

```
// TODO: Re-enable this test once we figure out why it does not raise an exception.
//       When key store password is not ....
```

core/src/test/java/com/linecorp/armeria/server/ServerTlsValidationTest.java
```diff
@@ -0,0 +1,84 @@
...
+            final KeyManagerFactory kmf =
+                    KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+            kmf.init(keyStore, "password".toCharArray());
+
+            final Server server = Server.builder()
+                                        .service("/", (ctx, res) -> HttpResponse.of(HttpStatus.OK))
+                                        .tls(kmf, sslContextBuilder -> {})
+                                        .build();
+        }).isInstanceOf(RuntimeException.class)
+          .hasMessageContaining("failed to validate SSL/TLS configuration");
```
ikhoon: Could you reduce `assertThatThrownBy` coverage? To catch error in exact place. 😄

jyblue: Fixed it. Thank you.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
jyblue: * I merged `validateKeyStoreNullPassword()` into `validateSslContext()` 
and removed the `validateSslContext()` call in the `VirtualHost`'s constructor.
* I made modified `validateSslContext()`, `ServerBuilder.tls(SslContext)`, `VirtualHostBuilder.tls(SslContext)` throws `SSLExcpetion`

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
jyblue: Removed.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
jyblue: Removed.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/ServerTlsValidationTest.java
jyblue: Updated javadoc. PTAL :bowing_man:

jyblue:
Thank you for review, again. Here is my update.

I fixed items from previous review 
and removed 6 deprecated method (had a chat with @trustin for this)
Please see below removed method list.
* `ServerBuilder.sslContext(sslContext)`
* `ServerBuilder.sslContext(protocol, keyCertChainFile, keyFile)`
* `ServerBuilder.sslContext(protocol, keyCertChainFile, keyFile, keyPassword)`
* `VirtualHostBuilder.sslContext(sslContext)`
* `VirtualHostBuilder.sslContext(protocol, keyCertChainFile, keyFile)`
* `VirtualHostBuilder.sslContext(protocol, keyCertChainFile, keyFile, keyPassword)`

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
```diff
@@ -195,10 +201,41 @@ private static boolean needsNormalization(String hostnamePattern) {
     }
 
     @Nullable
-    static SslContext validateSslContext(@Nullable SslContext sslContext) {
+    static SslContext validateSslContext(@Nullable SslContext sslContext) throws SSLException {
```
minwoox: Because this only called from `VirtualHostBuilder`, how about moving this into the class and make this private?

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
```diff
@@ -195,10 +201,41 @@ private static boolean needsNormalization(String hostnamePattern) {
...
+            serverEngine = sslContext.newEngine(ByteBufAllocator.DEFAULT);
+            serverEngine.setUseClientMode(false);
+            serverEngine.setNeedClientAuth(false);
+
+            final SslContext sslContextClient =
+                    VirtualHostBuilder.buildSslContext(SslContextBuilder::forClient, sslContextBuilder -> {});
+            clientEngine = sslContextClient.newEngine(ByteBufAllocator.DEFAULT);
+            clientEngine.setUseClientMode(true);
+
+            clientEngine.beginHandshake();
```
minwoox: Could you add a comment that why we call this method?

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
```diff
@@ -195,10 +201,41 @@ private static boolean needsNormalization(String hostnamePattern) {
...
+
+            final ByteBuffer appBuf = ByteBuffer.allocate(clientEngine.getSession().getApplicationBufferSize());
+            final ByteBuffer packetBuf = ByteBuffer.allocate(clientEngine.getSession().getPacketBufferSize());
+
+            clientEngine.wrap(appBuf, packetBuf);
+            appBuf.clear();
+            packetBuf.flip();
+            serverEngine.unwrap(packetBuf, appBuf);
+        } catch (SSLException e) {
+            throw new SSLException("failed to validate SSL/TLS configuration: " + e, e);
```
ikhoon: I think we don't have to put `e` two times. How about replacing with `sslContext`?
```
throw new SSLException("failed to validate SSL/TLS configuration: " + sslContext, e);
```

trustin: I'm not sure. `SslContext.toString()` does not always yield a good hint. The intention of appending `e` here is to make it easy to find the cause of the validation failure easily. However, I don't think we need to repeat the exception type. How about:

    "failed to ...." + e.getMessage()

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
ikhoon: I guess we can remove `@Nullable` here?

jyblue: After moving, 
can we also make this method non-static private method?
Do we have any reason this method should be static?

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
jyblue: Added comment on the method. PTAL :bowing_man:

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
minwoox: I think there's not so much difference if we add static or not because it's optimized anyway. However, we prefer the class level method rather than the instance level. Please check out [here](https://www.geeksforgeeks.org/static-methods-vs-instance-methods-java/). 😉

jyblue: Thank you, it helps me a lot to understand :+1:

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
minwoox: I mean why we need to call `beginHandshake();`.
And I found out from the Javadoc of `beginHandshake()`:
```java
* This method is not needed for the initial handshake, as the
* {@code wrap()} and {@code unwrap()} methods will
* implicitly call this method if handshaking has not already begun.
```
So we might not need to call this? Not sure this is right. 😄

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
jyblue: I agree,
* Users need useful message to find the cause easily.
* We don't need to repeat the exception type.

I modified to 
``` java
throw new SSLException("failed to validate SSL/TLS configuration: " + e.getMessage());
```

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
jyblue: AFAIU, we can remove `beginHandshake()`

I searched netty's `SSLEngine` implemantation of `RefernceCountedOpenSslEngine`
and I found following,

Conclusion:
* We don't need to call `beginHandshake()` when it is first handshake procedure.

Reference source code:
* from : io.netty.handler.ssl.ReferenceCountedOpenSslEngine.java

Checks:
* Initial value of `SSLEngine.handshakeState` is `NOT_STARTED`
* `SSLEngine.beginHandshake()` set `SSLEngine.handshakeState` to `STARTED_EXPLICITLY`
``` java
case NOT_STARTED:
  handshakeState = HandshakeState.STARTED_EXPLICITLY;
  if (handshake() == NEED_TASK) {
      // Set needTask to true so getHandshakeStatus() will return the correct value.
     needTask = true;
  }
  calculateMaxWrapOverhead();
  break;
```
* If `SSLEngine` try to first handshake without calling `beginHandshake()`,
`wrap()` or `unwrap()` method set `SSLEngine.handshakeState` to `STARTED_EXPLICITLY`
``` java
// wrap()
// Flush any data that may be implicitly generated by OpenSSL (handshake, close, etc..).
  SSLEngineResult.HandshakeStatus status = NOT_HANDSHAKING;
  // Prepare OpenSSL to work in server mode and receive handshake
  if (handshakeState != HandshakeState.FINISHED) {
      if (handshakeState != HandshakeState.STARTED_EXPLICITLY) {
          // Update accepted so we know we triggered the handshake via wrap
          handshakeState = HandshakeState.STARTED_IMPLICITLY;
      }
```
``` java
// unwrap()
SSLEngineResult.HandshakeStatus status = NOT_HANDSHAKING;
  // Prepare OpenSSL to work in server mode and receive handshake
  if (handshakeState != HandshakeState.FINISHED) {
      if (handshakeState != HandshakeState.STARTED_EXPLICITLY) {
          // Update accepted so we know we triggered the handshake via wrap
          handshakeState = HandshakeState.STARTED_IMPLICITLY;
      }
```

jyblue: I also run `validateSslContext()` with/without `beginHandshake()`.
And results are same.

So, I think we can remove `beginHandshake()` in this case.
Let me submit fixed commit.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -93,6 +96,46 @@
...
+
+            final ByteBuffer appBuf = ByteBuffer.allocate(clientEngine.getSession().getApplicationBufferSize());
+            final ByteBuffer packetBuf = ByteBuffer.allocate(clientEngine.getSession().getPacketBufferSize());
+
+            clientEngine.wrap(appBuf, packetBuf);
+            appBuf.clear();
+            packetBuf.flip();
+            serverEngine.unwrap(packetBuf, appBuf);
+        } catch (SSLException e) {
+            throw new SSLException("failed to validate SSL/TLS configuration: " + e.getMessage());
```
trustin: But don't we need to set the cause of the `SSLException`? `, e`

jyblue: Thanks! I fixed it.
I just understood what
>  I don't think we need to repeat the exception type.  

means. :sob:

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -93,6 +96,46 @@
             ApplicationProtocolNames.HTTP_2,
             ApplicationProtocolNames.HTTP_1_1);
 
+    /*
+     * Validate {@code sslContext} is configured properly. If {@code sslContext} is configured as client
+     * context, or key store password is not given to key store when {@code sslContext} is created using key
+     * manager factory, the validation will fail and an exception will be raised.
```
ikhoon: nit: an exception -> an `{@link SSLException}`

jyblue: I fixed it. Thank you for review :+1:

trustin: It has to be `an {@link SSLException}`. You omitted `an` :wink:

jyblue: Oops, I should've double checked it.
Please let me fix it.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -99,7 +99,7 @@
     /**
      * Validate {@code sslContext} is configured properly. If {@code sslContext} is configured as client
      * context, or key store password is not given to key store when {@code sslContext} is created using key
-     * manager factory, the validation will fail and an {@link SSLException} will be raised.
+     * manager factory, the validation will fail and a {@link SSLException} will be raised.
```
trustin: an is correct here. 'an es-es-el...'

jyblue: Oh.... I see :sob: :sob: :sob: 
Fixed it. Thank you for your patience.

trustin: No worries :-D

trustin:
Nice work, @jyblue! I've updated the PR description and commit message for you. :wink:

Issue #2117, opened 2019-09-26
Version information in `DocService` index page is not sorted.


trustin:
Could be sorted for aesthetics' sake:

![image](https://user-images.githubusercontent.com/173918/65683043-44c5ad80-e097-11e9-8d59-17283fa0703f.png)

young891221:
I'll work on it!

rcsalome:
Hello. I have a quick fix for this. Will get a PR created. Thanks.

beargummy:
Should this issue be closed?

trustin:
Yes, indeed. Thanks for the heads up!

PR #2125, opened 2019-09-29
Sort `DocService` Modules Alphabetically by artifactId

rcsalome:
Motivation:

Fix defect #2117

Modifications:

Use 'localeCompare' to sort before rendering table.

Result:

Modules are now sorted alphabetically.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/Version.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/Version.java
@@ -41,11 +41,11 @@ import java.security.PrivilegedAction;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Enumeration;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.TreeMap;
 
 import javax.annotation.Nullable;
 
@@ -134,7 +134,7 @@ public final class Version {
             artifactIds.add(artifactId);
         }
 
-        final Map<String, Version> versions = new HashMap<>();
+        final Map<String, Version> versions = new TreeMap<>();
         for (String artifactId: artifactIds) {
             versions.put(
                     artifactId,
--- a/core/src/main/java/com/linecorp/armeria/common/util/Version.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/Version.java
@@ -61,7 +61,7 @@ import com.google.common.io.Closeables;
  * it may not be possible to retrieve the information completely, depending on your environment, such as
  * the specified {@link ClassLoader}, the current {@link SecurityManager}.
  */
-public final class Version {
+public final class Version implements Comparable<Version>{
 
     // Forked from Netty 4.1.34 at d0912f27091e4548466df81f545c017a25c9d256
 
@@ -227,5 +227,10 @@ public final class Version {
     @JsonIgnore
     public boolean isClean() {
         return "clean".equals(repositoryStatus);
+    }
+
+    @Override
+    public int compareTo(Version ver) {
+        return this.artifactId().compareTo(ver.artifactId());
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/common/util/Version.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/Version.java
@@ -61,7 +61,7 @@ import com.google.common.io.Closeables;
  * it may not be possible to retrieve the information completely, depending on your environment, such as
  * the specified {@link ClassLoader}, the current {@link SecurityManager}.
  */
-public final class Version implements Comparable<Version>{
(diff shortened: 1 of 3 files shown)


ikhoon:
@rcsalome Thanks for contributing! Could you sign for CLA? 😀

trustin:
Please make sure the e-mail address in your commit is registered and validated in your GitHub account.

minwoox:
Thanks for your contribution!
I was kind of imagining sorting the version in `DocService` which is at [here](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/docs/DocService.java#L168) and passing the values. So we don't have to sort it right before rendering. What do you think? @trustin, @ikhoon

trustin:
@minwoox Ah, that's also a good idea. Yeah, how about sorting at the server side once and for ever, @rcsalome?

ikhoon:
Sorting of `Version` is meaningful itself because the API can use at other places. 👍 But I think how to display is a little bit client-side logic. 🧐

rcsalome:
Hello. Thanks for the feedback!

I agree with @ikhoon, that since we want to modify the display, the client should handle that. 
However, I see your points @minwoox and @trustin. Sorting on the server allows other areas of the code to make use of the sorted results.

I will make a PR today or tomorrow with sorting on the server, and we can compare. Need to set up my IDE first.

Also, I will sign the CLA and update my email. Thanks again!

rcsalome:
I have just pushed a new approach. Feedback welcome.

However, if you do not want me to modify the `Version` class, that is fine. I will move the comparison logic to the `DocService` itself.

core/src/main/java/com/linecorp/armeria/common/util/Version.java
```diff
@@ -228,4 +228,9 @@ public String toString() {
     public boolean isClean() {
         return "clean".equals(repositoryStatus);
     }
+
+    @Override
+    public int compareTo(Version ver) {
+        return this.artifactId().compareTo(ver.artifactId());
```
minwoox: We do not need `this.`.

core/src/main/java/com/linecorp/armeria/common/util/Version.java
```diff
@@ -61,7 +61,7 @@
  * it may not be possible to retrieve the information completely, depending on your environment, such as
  * the specified {@link ClassLoader}, the current {@link SecurityManager}.
  */
-public final class Version {
+public final class Version implements Comparable<Version>{
```
rcsalome: I will add a space here after the `<Version>`. Will set up the formatter in the IDE.

core/src/main/java/com/linecorp/armeria/common/util/Version.java
minwoox: If two `Version`s have the same `artifactId`, they will be treated as the same version while sorting. Of course, this is not a problem in `DocService`, because there's no two `Version`s which have equal `artifactId`. However, if there's a situation in another place that we have to sort `Version`s which have the same `artifactId` but different `artifactVersion`, that will be a problem. So I think if we want to use `Comparable`, we should use all the fields. What do you think?

rcsalome: Yes. That is a good idea. Maybe we don't need all fields. I think we can do `artifactId`, then `artifactVersion`, then `longCommitHash`.

But it's up to you. If you want me to do all of the fields. That is ok. Just let me know what order you want them.

minwoox: As you said, we don't need all fields. I think just `artifactId ` and `artifactVersion` are enough. :-)

core/src/main/java/com/linecorp/armeria/server/docs/DocService.java
```diff
@@ -164,7 +164,7 @@ public void serverStarting(Server server) throws Exception {
                 spec = addDocStrings(spec, services);
                 spec = addExamples(spec);
 
-                final List<Version> versions = ImmutableList.copyOf(
+                final List<Version> versions = ImmutableList.sortedCopyOf(
```
trustin: How about just modifying `Version.identify()` to use a `SortedMap` rather than a `HashMap`?

rcsalome: This seems like the simplest solution!

rcsalome:
I created a new PR https://github.com/line/armeria/pull/2145 . I hope that is ok.

I am having trouble signing the CLA, since my first commit was using an incorrect gitconfig file.

trustin:
Shall I close this PR, or do you want to preserve the frontend-side changes?

rcsalome:
@trustin, You can close this PR. I will work on #2145. Thanks.

trustin:
Thanks, @rcsalome !

PR #2127, opened 2019-09-29
Separate RequestContextAwareFuture version overrides into mrjar…

anuraaga:
…e sets.

Now that we are using mrjar for version-specific implementations, we can use it for the recent change for completablefuture to support Java 9, and future ones to support Java 12.

Advantages

- Can use other Java 9 APIs when implementing a Java 9 API (e.g., `super.completeAsync`)
- Maybe some small performance by not needing e.g., `maybeMakeContextAware`, `supplyAsync` for completion of current future
- Possible readability improvement by being able to use `@Override` and otherwise IDE-awareness of the API level

Disadvantage
- If user shades armeria into non-mrjar and runs on newer Java, newer APIs wouldn't function correctly

I think shading is advanced enough that the disadvantage is ok but wondering what everyone thinks.


core/build.gradle
```diff
@@ -26,8 +51,10 @@ tasks.withType(JavaCompile) {
 }
 
 tasks.withType(Jar) {
-    into('META-INF/versions/9') {
-        from sourceSets.java9.output
+    mrJarVersions.each {version->
```
trustin: nit: `{ version ->`

core/src/main/java9/com/linecorp/armeria/internal/Java9RequestContextAwareCompletableFuture.java
```diff
@@ -25,271 +25,247 @@
 import java.util.function.Function;
 import java.util.function.Supplier;
 
+import com.linecorp.armeria.common.RequestContext;
 import com.linecorp.armeria.common.util.SafeCloseable;
 
-final class RequestContextAwareCompletableFuture<T> extends CompletableFuture<T> {
+final class Java9RequestContextAwareCompletableFuture<T> extends CompletableFuture<T> {
```
trustin: Could this class extend `RequestContextAwareCompletableFuture` to reduce duplication?

anuraaga: The main reason not to extend is to avoid the complexity of `maybeContextAware`. I think since Java 8 API won't be added to, the duplication is ok since there's no maintenance burden. What do you think?

trustin: SGTM

core/src/main/java9/com/linecorp/armeria/internal/Java9RequestContextAwareMinimalStage.java
```diff
@@ -27,259 +27,222 @@
...
 
     @Override
     public <U> CompletionStage<Void> thenAcceptBothAsync(CompletionStage<? extends U> other,
                                                          BiConsumer<? super T, ? super U> action,
                                                          Executor executor) {
-        return new RequestContextAwareMinimalStage(
-                (RequestContextAwareCompletableFuture<Void>) delegate.thenAcceptBothAsync(other, action,
-                                                                                          executor));
+        return new Java9RequestContextAwareMinimalStage(delegate.thenAcceptBothAsync(other, action,
+                                                                                               executor));
```
ikhoon: nit: indent?

core/src/main/java9/com/linecorp/armeria/internal/Java9RequestContextAwareMinimalStage.java
```diff
@@ -27,259 +27,222 @@
...
     }
 
     @Override
     public CompletionStage<Void> runAfterBothAsync(CompletionStage<?> other, Runnable action,
                                                    Executor executor) {
-        return new RequestContextAwareMinimalStage(
-                (RequestContextAwareCompletableFuture<Void>) delegate.runAfterBothAsync(other, action,
-                                                                                        executor));
+        return new Java9RequestContextAwareMinimalStage(delegate.runAfterBothAsync(other, action,
+                                                                                             executor));
```
ikhoon: ditto

core/src/main/java9/com/linecorp/armeria/internal/Java9RequestContextAwareMinimalStage.java
```diff
@@ -27,259 +27,222 @@
...
     }
 
     @Override
     public CompletionStage<Void> acceptEitherAsync(CompletionStage<? extends T> other,
                                                    Consumer<? super T> action, Executor executor) {
-        return new RequestContextAwareMinimalStage(
-                (RequestContextAwareCompletableFuture<Void>) delegate.acceptEitherAsync(other, action,
-                                                                                        executor));
+        return new Java9RequestContextAwareMinimalStage(delegate.acceptEitherAsync(other, action,
+                                                                                             executor));
```
ikhoon: ditto

core/src/main/java9/com/linecorp/armeria/internal/Java9RequestContextAwareMinimalStage.java
```diff
@@ -27,259 +27,222 @@
...
 
     @Override
     public CompletionStage<Void> runAfterEitherAsync(CompletionStage<?> other,
                                                      Runnable action,
                                                      Executor executor) {
-        return new RequestContextAwareMinimalStage(
-                (RequestContextAwareCompletableFuture<Void>) delegate.runAfterEitherAsync(other, action,
-                                                                                          executor));
+        return new Java9RequestContextAwareMinimalStage(delegate.runAfterEitherAsync(other, action,
+                                                                                               executor));
```
ikhoon: ditto

core/src/main/java9/com/linecorp/armeria/internal/Java9VersionSpecific.java
```diff
@@ -36,4 +39,9 @@ public final long currentTimeMicros() {
         return TimeUnit.SECONDS.toMicros(now.getEpochSecond()) + TimeUnit.NANOSECONDS.toMicros(
                 now.getNano());
     }
+
+    @Override
+    public <T> CompletableFuture<T> newRequestContextCompletableFuture(RequestContext ctx) {
```
ikhoon: How about renaming `newRequestContextCompletableFuture` -> `newRequestContextAwareCompletableFuture`?

ikhoon: Question: Don't we need `final` here?

core/src/main/java9/com/linecorp/armeria/internal/Java9RequestContextAwareMinimalStage.java
trustin: I pushed a fix commit for this.

core/src/main/java9/com/linecorp/armeria/internal/Java9VersionSpecific.java
trustin: +1 for renaming. Maybe just `newRequestContextAwareFuture()` because it's too long? We could rename the classes as well.

anuraaga: @ikhoon Thanks for remembering my pattern for me! :D

@trustin Renamed the method. We already have a class named `RequestContextAwareFuture` though so left them

Issue #1880, opened 2019-07-09
Kotlin examples


ikhoon:
We don't have any [Kotlin](https://kotlinlang.org/) examples. 
Kotlin has been growing fast and achieved the 14th on the [PYPL(PopularitY of Programming Language)](http://pypl.github.io/PYPL.html).
Kotlin examples could be a good point for Kotlin users to start Armeria.

heowc:
https://github.com/line/armeria/pull/2335#issuecomment-569164808

cricket007:
Link back #1078

PR #2128, opened 2019-09-30
Add a Kotlin sample. #1880

cnevinc:
This commit transfer `spring-boot-minimal` into Kotlin.


ikhoon:
@cnevinc  Thanks for contributing! I'm happy to review this PR. 😊
And could you sign for CLA?

examples/spring-boot-kotlin/build.gradle
```diff
@@ -0,0 +1,46 @@
+buildscript {
+    ext.kotlin_version = '1.3.50'
```
trustin: How about adding `kotlin-stdlib-jdk8` to `dependencies.yml` and refer to the version via `managedVersion`?

settings.gradle
```diff
@@ -55,6 +55,7 @@ includeWithFlags ':examples:grpc-service',           'java'
 includeWithFlags ':examples:proxy-server',           'java'
 includeWithFlags ':examples:saml-service-provider',  'java'
 includeWithFlags ':examples:server-sent-events',     'java'
+includeWithFlags ':examples:spring-boot-kotlin',     'java'
```
trustin: `spring-boot-minimal-kotlin`?

trustin:
Could you also use the different package name with the `spring-boot-minimal` example? It seems like JaCoCo is confused. How about just add `.kt` to the package names?

examples/spring-boot-kotlin/src/main/java/example/springframework/boot/minimal/Main.kt
```diff
@@ -0,0 +1,14 @@
+package example.springframework.boot.minimal
+
+import org.springframework.boot.SpringApplication
+import org.springframework.boot.autoconfigure.SpringBootApplication
+
+@SpringBootApplication
+open class Main {
```
heowc: Spring is guiding you to [a better example](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-kotlin.html#boot-features-kotlin-api-runapplication) than the change from java to kotlin function provided by idea.

Wouldn't this be more familiar to others?

heowc:
Hello, Welcome to contribute code to Armeria! :smile:

Although I don't know much about Kotlin, I saw the code and would like to suggest it.

Rather than just converting, I think that using the generator provided by [`start.spring.io`](https://start.spring.io) will be an easier and more user-friendly example.

cnevinc:
> Rather than just converting, I think that using the generator provided by [`start.spring.io`](https://start.spring.io) will be an easier and more user-friendly example.

Good idea. I'll come up with a better example in the other PR. Thank you!

trustin:
@cnevinc Any news on a follow-up? :wink:

heowc:
@trustin  @cnevinc May I challenge this?

trustin:
@heowc Of course! Thanks in advance! :pray:

Issue #2041, opened 2019-09-04
No exception logged when JSON request conversion fails in an annotated service


trustin:
For example:

```java
public class MyService {
    @Post("/post")
    void post(MyValue value) {
        ...
    }
}

public class MyValue {
    final String value;
    public MyValue(String value) {
        this.value = value;
    }
}   
```

When an HTTP client posts a JSON document:

```
curl -X POST \
  http://localhost:8080/post \
  -H 'Content-Type: application/json' \
  -H 'Host: localhost:8080' \
  -d '{ "value": "foo" }'
```

.. a `400 Bad Request` response is sent by the server, but no log message is written, making it hard to know what is wrong. The problem goes away if a user annotates `MyValue.<init>` with `@JsonCreator` and the parameter `value` with `@JsonProperty("value")`.

sivaalli:
Is it possible for me to take a look at this and possibly fix it by adding a log statement?

trustin:
@sivaalli Sure! Thanks a lot for looking into it. :bow:

sivaalli:
Hi @trustin , while making a change I got a doubt and wanted to get it clarified. If a `ExceptionHandlerFunction` is declared and the conversion fails(at runtime), then the function will be able to catch exception and handle it. If we log it in `JacksonRequestConverterFunction` and the `ExceptionHandlerFunction` that the user provided also logs it, then log appears twice. What do you think about this.

```
public class Armeria {

    public static void main(String[] args) throws ExecutionException, InterruptedException {
        Server build = new ServerBuilder()
                .port(8080, SessionProtocol.HTTP)
                .annotatedService("/", new MyService()).build();
        build.start().get();

    }

    private static class MyService {

        @ExceptionHandler(LoggingExceptionHandlerFunction.class)
        @Post("/a")
        public HttpResponse a(MyValue value){
            System.out.println(value.value);
            return HttpResponse.of(200);
        }
    }

    public static class MyValue{
        @JsonProperty("value")
        public String value;
    }

    public static class LoggingExceptionHandlerFunction implements ExceptionHandlerFunction{

        @Override
        public HttpResponse handleException(ServiceRequestContext ctx, HttpRequest req, Throwable cause) {
            System.out.println(cause);
            return HttpResponse.of(400);
        }
    }
}
```

trustin:
[...]

PR #2130, opened 2019-09-30
Adding log statement on JSON parse errors

sivaalli:
Motivation:

When a JSON document cannot be converted into Java object, an `IllegalArgumentException` is thrown. And this is converted to 400 status code. But it would be nice to have a log statement in `DefaultExceptionHandler` to log what is wring with the request

Also closes: https://github.com/line/armeria/issues/2041


sivaalli:
Sorry, Let me fix the commits

PR #2134, opened 2019-10-01
Fix a bug where long-lived request is not fully sent when retry…

minwoox:
Motivation:
If we create a `RetryingHttpClient` using `RetryStrategy`, in order to decide to retry or not, we used the information in `ResponseHeaders`.
For example, if the status is `200`, we don't do the retry but keep receiving the responses.
However, the previous logic had a flaw. When it gets the `200` status, it considers that the original `HttpRequest` is fully sent,
so it calls `HttpReqeustDuplicator.close()` to clean up the `ByteBuf`s that the duplicator is holding.
In `HttpReqeustDuplicator.close()`, the duplicator aborts all children `HttpRequest`s.

This is not a problem when the `HttpResponse` is received after `HttpRequest` is fully sent.
But it's a problem when it's not such as sending a long-lived streaming `HttpRequest`.

Modifications:
- `StreamMessageDuplicator.close()` does not abort children `StreamMessage`s anymore.
  - It makes the duplicator not duplicable anymore.
  - The children `StreamMessage`s are completed when the original publisher is closed.
- Add `StreamMessageDuplicator.abort()` to abort all children `StreamMessage`s.
- Clean up `AbstractStreamMessageDuplicator`

Result:
- You no longer see `AbortedStreamException` when sending long-lived requests using `RetryingHttpClient`.
- (Breaking) `AbstractStreamMessageDuplicator.close()` does not abort all children `StreamMessage`s.
  - You should use `AbstractStreamMessageDuplicator.abort()` to abort all children `StreamMessage`s.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
@@ -51,6 +51,7 @@ import com.linecorp.armeria.common.util.SafeCloseable;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.EventLoop;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.ImmediateEventExecutor;
@@ -77,9 +78,6 @@ import io.netty.util.concurrent.ImmediateEventExecutor;
 public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage<T>>
         implements SafeCloseable {
 
-    @SuppressWarnings("rawtypes")
-    private static final CompletableFuture[] EMPTY_FUTURES = new CompletableFuture[0];
-
     private final StreamMessageProcessor<T> processor;
 
     private final EventExecutor duplicatorExecutor;
@@ -101,8 +99,10 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
         if (executor != null) {
             duplicatorExecutor = executor;
         } else {
-            duplicatorExecutor = RequestContext.mapCurrent(
+            final EventLoop currentExecutor = RequestContext.mapCurrent(
                     RequestContext::eventLoop, () -> CommonPools.workerGroup().next());
+            assert currentExecutor != null;
+            duplicatorExecutor = currentExecutor;
         }
 
         processor = new StreamMessageProcessor<>(publisher, signalLengthGetter,
@@ -149,6 +149,10 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
         processor.close();
     }
 
+    public void abort() {
+        processor.abort();
+    }
+
     @VisibleForTesting
     static class StreamMessageProcessor<T> implements Subscriber<T> {
 
@@ -168,10 +172,6 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
             CLOSED
         }
 
-        @SuppressWarnings("rawtypes")
-        private static final AtomicLongFieldUpdater<StreamMessageProcessor> requestedDemandUpdater =
-                AtomicLongFieldUpdater.newUpdater(StreamMessageProcessor.class, "requestedDemand");
-
         private final StreamMessage<T> upstream;
         private final SignalQueue signals;
         private final SignalLengthGetter<Object> signalLengthGetter;
@@ -185,10 +185,11 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
         volatile int downstreamSignaledCounter;
         volatile int upstreamOffset;
 
-        @SuppressWarnings("unused")
-        private volatile long requestedDemand;
+        private long requestedDemand;
         @Nullable
-        private volatile Subscription upstreamSubscription;
+        private Subscription upstreamSubscription;
+
+        private boolean cancelUpstream;
 
         private volatile State state = State.DUPLICABLE;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -149,6 +149,10 @@ public void close() {
         processor.close();
     }
 
+    public void abort() {
```
trustin: Javadoc?

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -305,14 +319,17 @@ private void doSubscribe(DownstreamSubscription<T> subscription) {
...
-                }
+                subscription.invokeOnSubscribe();
             }
         }
 
+        private static void failLateProcessorSubscriber(DownstreamSubscription<?> subscription) {
+            final Subscriber<?> lateSubscriber = subscription.subscriber();
+            lateSubscriber.onSubscribe(NoopSubscription.INSTANCE);
+            lateSubscriber.onError(
+                    new IllegalStateException("duplicator is closed or last downstream is added."));
```
trustin: .. or no more downstream can be added?

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
minwoox: Oops~!

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
minwoox: Fixed. :-)

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -407,9 +467,7 @@ private void doCleanup() {
                 completionFutures.add(future);
             });
             downstreamSubscriptions.clear();
-            final CompletableFuture<Void> allDoneFuture = CompletableFuture.allOf(
-                    completionFutures.toArray(EMPTY_FUTURES));
-            allDoneFuture.handle((unused1, unused2) -> {
+            CompletableFutures.successfulAsList(completionFutures, cause -> null).handle((unused1, unused2) -> {
```
ikhoon: If we don't need results, I think `CompletableFutures.allAsList` is faster than `CompletableFutures.successfulAsList`.
Because of the call stack `successfulAsList` -> `joinList` -> `allAsList` :-) 

https://github.com/spotify/completable-futures/blob/06af851211d335de98fa501f0eb26086b6520888/src/main/java/com/spotify/futures/CompletableFutures.java#L101-L107

ikhoon: Sorry, I had a wrong review. I realized that `allAsList` doesn't wait for all responses if any of the responses has an exception. 
Thanks for the nice work!

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -138,17 +138,23 @@ protected EventExecutor duplicatorExecutor() {
     }
 
     /**
-     * Closes this factory and stream messages who are invoked by
-     * {@link AbstractStreamMessageDuplicator#duplicateStream()}.
-     * Also, clean up the data published from {@code publisher}. If {@link #duplicateStream(boolean)} with
-     * {@code true} is called already, invoking this method does not affect and cleaning up occurs when all
-     * of the {@link Subscription}s are completed or cancelled.
+     * Closes this duplicator and makes this not duplicable. You will get an {@link IllegalStateException} if
```
trustin:
```
Closes this duplicator and prevents it from further duplication. 
{@link #duplicateStream()} will raise an {@link IllegalStateException} after
this method is invoked. Note that the previously {@linkplain #duplicateStream() duplicated streams}
will not be closed but will continue publishing data until the original {@link Publisher}
is closed. All the data published from the {@link Publisher} are cleaned up when
all {@linkplain #duplicateStream() duplicated streams} are complete.
```

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -138,17 +138,23 @@ protected EventExecutor duplicatorExecutor() {
...
+     * All the data published from the {@code publisher} are cleaned up when all
+     * of the {@linkplain AbstractStreamMessageDuplicator#duplicateStream() duplicated streams} are completed.
      */
     @Override
     public void close() {
         processor.close();
     }
 
+    /**
+     * Closes this duplicator and aborts all stream messages who are invoked by
```
trustin: who are invoked by -> returned by

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -138,17 +138,23 @@ protected EventExecutor duplicatorExecutor() {
...
+     * {@link AbstractStreamMessageDuplicator#duplicateStream()}. This will also, clean up the data published
```
trustin: You don't really need to specify a class name here.

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
minwoox: Much better! Thanks!

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
minwoox: Fixed. 😉

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
minwoox: Fixed. 😉

minwoox:
Thanks for reviewing!

PR #2138, opened 2019-10-02
Do not notify `ResponseTimeoutException` twice

### LINKED ISSUE #2000: Response Timeouts both logged and returned to user

trustin:
Motivation:

- `HttpResponseDecoder` logs a `ResponseTimeoutException` while
  notifying it to a user, which is a double notification.

Modifications:

- Do not log a `ResponseTimeoutException` when the response has been
  timed out and the `HttpResponse` was closed with the same exception.

Result:

- Fixes #2000

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
@@ -175,16 +175,6 @@ abstract class HttpResponseDecoder {
                     this, responseTimeoutMillis, TimeUnit.MILLISECONDS);
         }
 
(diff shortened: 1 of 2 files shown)


trustin:
Thanks, reviewers!

PR #2142, opened 2019-10-02
Provide a way to configure a CacheSpec for HttpFileService

### LINKED ISSUE #916: Provide a way to configure a CacheSpec for HttpFileService
Currently, there is no way to configure a cache for `DocService` instantiated by `ArmeriaAutoConfiguration#armeriaServer`.

To make it more configurable, it is better to provide a new flag to `Flags` (or `ArmeriaSettings`).

heowc:
Motivation:

A user cannot set a way to configure a CacheSpec for HttpFileService.

Modifications:

- Provide JVM option(`-Dcom.linecorp.armeria.fileServiceCache=<spec>`)
- Add `fileServiceCache ` for `Flags`
- Fix `maxCacheEntries` for `HttpFileServiceBuilder`
- Add `entryCacheSpec` for `HttpFileServiceBuilder`
- Fix doc

Result:

- Fixes #916


core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -282,6 +282,10 @@
     private static final Optional<String> HEADER_VALUE_CACHE_SPEC =
             caffeineSpec("headerValueCache", DEFAULT_HEADER_VALUE_CACHE_SPEC);
 
+    private static final String DEFAULT_HTTP_FILE_SERVICE_CACHE_SPEC = "maximumSize=1024";
+    private static final Optional<String> HTTP_FILE_SERVICE_CACHE_SPEC =
+            caffeineSpec("httpFileServiceCache", DEFAULT_HTTP_FILE_SERVICE_CACHE_SPEC);
```
trustin: `httpFile` -> `file`, `HTTP_FILE` -> `FILE` (I think we're gonna rename `HttpFileService` to `FileService` eventually)

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -98,11 +102,38 @@ public HttpFileServiceBuilder clock(Clock clock) {
     }
 
     /**
-     * Sets the maximum allowed number of cached file entries. If not set, {@value #DEFAULT_MAX_CACHE_ENTRIES}
+     * Sets the maximum allowed number of cached file entries. If not set, {@code DEFAULT_ENTRY_CACHE_SPEC}
```
trustin: I think this is not user friendly. How about just `{@code 1024}` at the expense of duplication?

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -98,11 +102,38 @@ public HttpFileServiceBuilder clock(Clock clock) {
...
+        if (!enableMaxCacheEntries) {
+            throw new IllegalStateException("already configured for the http file service: " + entryCacheSpec);
+        }
+        entryCacheSpec = HttpFileServiceConfig.validateMaxCacheEntries(maxCacheEntries);
+        enableEntryCacheSpec = false;
+        return this;
+    }
+
+    /**
+     * Sets the cache spec for caching file entries. If not set, {@code DEFAULT_ENTRY_CACHE_SPEC}
```
trustin: Ditto - `{@code "maximumSize=1024"}` ?

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -78,10 +80,12 @@ public static HttpFileServiceBuilder forVfs(HttpVfs vfs) {
 
     private final HttpVfs vfs;
     private Clock clock = Clock.systemUTC();
-    private int maxCacheEntries = DEFAULT_MAX_CACHE_ENTRIES;
+    private Optional<String> entryCacheSpec = DEFAULT_ENTRY_CACHE_SPEC;
     private int maxCacheEntrySizeBytes = DEFAULT_MAX_CACHE_ENTRY_SIZE_BYTES;
     private boolean serveCompressedFiles;
     private boolean autoIndex;
+    private boolean enableMaxCacheEntries = true;
+    private boolean enableEntryCacheSpec = true;
```
trustin: How about `canSetEntryCacheSpec` and `canSetMaxCacheEntries` ?

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -98,11 +102,38 @@ public HttpFileServiceBuilder clock(Clock clock) {
...
+     * > HttpFileServiceBuilder.forFileSystem(...)
+     * >                       .entryCacheSpec("maximumSize=1024")
+     * >                       .build();
+     * }</pre>
      */
+    @Deprecated
     public HttpFileServiceBuilder maxCacheEntries(int maxCacheEntries) {
-        this.maxCacheEntries = HttpFileServiceConfig.validateMaxCacheEntries(maxCacheEntries);
+        if (!enableMaxCacheEntries) {
+            throw new IllegalStateException("already configured for the http file service: " + entryCacheSpec);
```
trustin: How about:
```java
checkState(canSetMaxCacheEntries,
           "Cannot call maxCacheEntries() if called entryCacheSpec() already.");
```

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -98,11 +102,38 @@ public HttpFileServiceBuilder clock(Clock clock) {
...
+
+    /**
+     * Sets the cache spec for caching file entries. If not set, {@code DEFAULT_ENTRY_CACHE_SPEC}
+     * is used by default.
+     */
+    public HttpFileServiceBuilder entryCacheSpec(String entryCacheSpec) {
+        requireNonNull(entryCacheSpec, "entryCacheSpec");
+        if (!enableEntryCacheSpec) {
+            throw new IllegalStateException(
+                    "already configured for the http file service: " + this.entryCacheSpec);
```
trustin: How about:
```java
checkState(canSetEntryCacheSpec,
           "Cannot call entryCacheSpec() if called maxCacheEntries() already.");
```

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -98,11 +103,34 @@ public HttpFileServiceBuilder clock(Clock clock) {
...
      * is used by default.
+     *
+     * @deprecated Use {@link #entryCacheSpec(String)}.
+     * <pre>{@code
+     * > HttpFileServiceBuilder.forFileSystem(...)
+     * >                       .entryCacheSpec("maximumSize=1024")
+     * >                       .build();
+     * }</pre>
      */
+    @Deprecated
```
trustin: We do not wish to deprecate this method because it is still a useful shortcut.

site/src/sphinx/server-http-file.rst
```diff
@@ -84,13 +84,15 @@ By default, :api:`HttpFileService` caches up to 1024 files whose length is less
...
+    fsb.entryCacheSpec("maximumSize=4096");
     // Cache files whose length is less than or equal to 1 MiB.
     fsb.maxCacheEntrySizeBytes(1048576);
 
     HttpFileService fs = fsb.build();
 
-The cache can also be disabled by specifying ``0`` for ``maxCacheEntries()``.
+The cache can also be disabled by specifying ``off`` for ``entryCacheSpec()``.
+Or, you can override the default value of ``maximumSize=1024`` using the JVM system property
+``-Dcom.linecorp.armeria.fileServiceCache=<spec>``.
```
trustin: Could we undeprecate `maxCacheEntries()`? i.e.

```rst
The cache can be disabled by specifying ``0`` for ``maxCacheEntries()``.
You can also specify a custom cache specification using ``entryCacheSpec()``,
as defined in `Caffeine documentation <https://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.6.0/com/github/benmanes/caffeine/cache/CaffeineSpec.html>`_.
Or, you can override the default cache specification of ``maximumSize=1024`` using
the JVM property ...
```

core/src/test/java/com/linecorp/armeria/server/file/HttpFileServiceTest.java
```diff
@@ -84,7 +84,7 @@ protected void configure(ServerBuilder sb) {
             sb.serviceUnder(
                     "/uncached/fs/",
                     HttpFileServiceBuilder.forFileSystem(tmpDir)
-                                          .maxCacheEntries(0)
+                                          .entryCacheSpec("off")
```
trustin: Once undeprecated, the `entryCacheSpec("off")` calls in this PR could be reverted for simplicity.

site/src/sphinx/server-http-file.rst
heowc: Since the current version is `2.8.0`, we will modify the contents to reflect this.

core/src/test/java/com/linecorp/armeria/server/file/HttpFileServiceTest.java
heowc: I will revert code

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -772,6 +776,18 @@ public static int defaultMaxTotalAttempts() {
         return HEADER_VALUE_CACHE_SPEC;
     }
 
+    /**
+     * Returns the value of the {@code fileServiceCache} parameter. It would be used to create a Caffeine
+     * {@link Cache} instance using {@link CaffeineSpec} for caching file entries.
+     *
+     * <p>The default value of this flag is {@value DEFAULT_FILE_SERVICE_CACHE_SPEC}. Specify the
+     * {@code -Dcom.linecorp.armeria.fileServiceCache=<spec>} JVM option to override the default value.
```
ikhoon: Could you add caffeine cache spec example? 
```java
For example, {@code -Dcom.linecorp.armeria.fileServiceCache=maximumSize=1024,expireAfterAccess=600s}.
```

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceConfig.java
```diff
@@ -36,25 +38,43 @@
...
+        }
+        try {
+            CaffeineSpec.parse(entryCacheSpec.get());
+        } catch (Exception e) {
+            throw new IllegalArgumentException("invalid cache spec: " + entryCacheSpec, e);
+        }
+        return entryCacheSpec;
+    }
+
+    @Deprecated
```
minwoox: This was package-private before, so we don't have to put `@Deprecated` here?

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceConfig.java
```diff
@@ -36,25 +38,43 @@
...
+    static Optional<String> validateMaxCacheEntries(int maxCacheEntries) {
```
minwoox: We can move this method into `HttpFileServiceBuilder` and make this private.

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceConfig.java
heowc: Oh, My mistake. I will fix it.

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -98,11 +103,26 @@ public HttpFileServiceBuilder clock(Clock clock) {
     }
 
     /**
-     * Sets the maximum allowed number of cached file entries. If not set, {@value #DEFAULT_MAX_CACHE_ENTRIES}
+     * Sets the maximum allowed number of cached file entries. If not set, {@code "maximumSize=1024"}
      * is used by default.
```
trustin: `If not set, up to {@code 1024} entries are cached by default.` ?

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceConfig.java
```diff
@@ -83,10 +103,10 @@ public Clock clock() {
     }
 
     /**
-     * Returns the maximum allowed number of cached file entries.
+     * Returns the cache spec for file entry cache.
```
trustin: `the cache spec of the file entry cache, as defined in {@link CaffeineSpec}.`

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -98,11 +103,26 @@ public HttpFileServiceBuilder clock(Clock clock) {
...
+                   "Cannot call maxCacheEntries() if called entryCacheSpec() already.");
+        entryCacheSpec = HttpFileServiceConfig.validateMaxCacheEntries(maxCacheEntries);
+        canSetEntryCacheSpec = false;
+        return this;
+    }
+
+    /**
+     * Sets the cache spec for caching file entries. If not set, {@code "maximumSize=1024"} is used by default.
+     */
+    public HttpFileServiceBuilder entryCacheSpec(String entryCacheSpec) {
```
ikhoon: I just saw that code patch coverage is lower than the baseline. Could you add a test case for here :-)

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
```diff
@@ -98,11 +103,38 @@ public HttpFileServiceBuilder clock(Clock clock) {
...
-     * Sets the maximum allowed number of cached file entries. If not set, {@value #DEFAULT_MAX_CACHE_ENTRIES}
-     * is used by default.
+     * Sets the maximum allowed number of cached file entries. If not set, up to {@code 1024} entries
+     * are cached by default.
      */
     public HttpFileServiceBuilder maxCacheEntries(int maxCacheEntries) {
-        this.maxCacheEntries = HttpFileServiceConfig.validateMaxCacheEntries(maxCacheEntries);
+        checkState(canSetMaxCacheEntries,
+                   "Cannot call maxCacheEntries() if called entryCacheSpec() already.");
+        entryCacheSpec = validateMaxCacheEntries(maxCacheEntries);
```
minwoox: I found out that this was a little bit weird because it validates `maxCacheEntries` but returns a cache spec. So how about just inlining this?
```java
public HttpFileServiceBuilder maxCacheEntries(int maxCacheEntries) {
    checkState(canSetMaxCacheEntries,
               "Cannot call maxCacheEntries() if called entryCacheSpec() already.");
    validateNonNegativeParameter(maxCacheEntries, "maxCacheEntries");
    if (maxCacheEntries == 0) {
        entryCacheSpec = Optional.empty();
    } else {
        // Do not have to call validateEntryCacheSpec()
        entryCacheSpec = Optional.of(String.format("maximumSize=%d", maxCacheEntries));
    }
    canSetEntryCacheSpec = false;
    return this;
}
```

heowc: Yes, good idea. I'll fix it right away.

core/src/main/java/com/linecorp/armeria/server/file/HttpFileServiceBuilder.java
heowc: I added test case. Please tell me if it is not enough.

heowc:
I found a case where it wasn't checked normally during the existing test. baseUri contains `/cached` rather than`/cached/`. 58808d4

minwoox:
Found build failure: 
```java
com.linecorp.armeria.server.file.HttpFileServiceTest > testClassPathGetFromModule(String)[1] FAILED
    java.lang.AssertionError: 
    Expecting:
     <["/cached/by-entry/classes/java/lang/Object.class",
        "/cached/classes/java/lang/Object.class"]>
    to contain:
     <["/cached/by-entry/classes/java/lang/Object.class"]>
    but could not find:
     <["/cached/by-entry/classes/java/lang/Object.class"]>
        at 
```

heowc:
Let's check a little bit more. :sweat:

trustin:
Thank you, @heowc!

PR #2145, opened 2019-10-03
Use TreeMap in Version Class to Sort the Versions

rcsalome:
Motivation:

Fix sorting defect #2117

Modifications:

Use a TreeMap instead of a HashMap when identifying versions.

Result:

Versions are now sorted by their artifactId, which is the TreeMap key.
DocService displays the modules sorted alphabetically.


rcsalome:
For discussions, see https://github.com/line/armeria/pull/2125 .

This PR was created, since the CLA agreement on the last PR is in an unapproved state. This is due to a misconfigured gitconfig file that was present during the initial commit.

PR #2151, opened 2019-10-03
Switch default build JDK to Java 12.

### LINKED ISSUE #2141: Raise build JDK requirement to Java 12
We now have travis infrastructure for installing multiple JDKs, including ones not officially suported by travis. This opens up the possibility of raising our minimum JDK version to build Armeria to Java 12. This makes it simpler to work on supporting newer APIs like in #2119 

Shall we raise the requirement?

anuraaga:
Fixes #2141

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -154,7 +154,7 @@ allprojects {
     }
 }
 
-// Require to use JDK 11 when releasing.
+// Require to use JDK 12 when releasing.
 tasks.release.doFirst {
     if (JavaVersion.current() != JavaVersion.VERSION_12) {
         throw new IllegalStateException("You must release using JDK 12.");
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ allprojects {
 task checkJavaVersion {
     doLast {
         if (JavaVersion.current().majorVersion.toInteger() < 12) {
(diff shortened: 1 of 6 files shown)


trustin:
Could you also update the `release` task in `/build.gradle` to require Java 12? Maybe should we make the entire build require Java 12 rather than just the release process?

anuraaga:
Thanks indeed. Changed the release task to require Java 12, and added a check for normal compile that verifies >=12 (don't think we have to require exactly 12 for normal building during development)

build.gradle
```diff
@@ -45,6 +45,16 @@ allprojects {
     }
 }
 
+task checkJavaVersion {
+    doLast {
+        if (JavaVersion.current().majorVersion.toInteger() < 12) {
+            throw new GradleException("The java version used ${JavaVersion.current()} is too old to " +
```
trustin: - java -> Java
- How about wrapping `${JavaVersion.current()}` with parens?

build.gradle
```diff
@@ -146,7 +156,7 @@ allprojects {
 
 // Require to use JDK 11 when releasing.
```
minwoox: nit: JDK 12 😄

PR #2152, opened 2019-10-03
Use singleton headers where possible in gRPC calls.

anuraaga:
It's actually relatively rare for headers to change across requests in gRPC since most users don't use message compression, and especially armeria users usually use our standard decorators instead of populating `Metadata`. We can optimize this case by always returning the same instance of `ResponseHeaders`.

Added a microbenchmark that only executes service business logic to remove thread-switch / connection overhead.

After
```
Benchmark                    Mode  Cnt      Score      Error  Units
GrpcServiceBenchmark.empty  thrpt    5  44601.768 ▒ 3864.274  ops/s
```

Before
```
Benchmark                    Mode  Cnt      Score      Error  Units
GrpcServiceBenchmark.empty  thrpt    5  33842.070 ▒ 8515.208  ops/s
```

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
+++ b/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
@@ -147,7 +147,7 @@ public final class GrpcService extends AbstractHttpService
         defaultHeaders = supportedSerializationFormats
                 .stream()
                 .map(format -> {
-                    ResponseHeadersBuilder builder =
+                    final ResponseHeadersBuilder builder =
                             ResponseHeaders
                                     .builder(HttpStatus.OK)
                                     .contentType(format.mediaType())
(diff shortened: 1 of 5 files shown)


grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
```diff
@@ -136,6 +143,21 @@ public static GrpcServiceBuilder builder() {
         this.maxInboundMessageSizeBytes = maxInboundMessageSizeBytes;
 
         advertisedEncodingsHeader = String.join(",", decompressorRegistry.getAdvertisedMessageEncodings());
+
+        defaultHeaders = supportedSerializationFormats
+                .stream()
+                .map(format -> {
+                    ResponseHeadersBuilder builder =
```
minwoox: nit: final

anuraaga:
@trustin I'm still hoping we find some large low-hanging fruit :) But in the meantime probably nice to do some microoptimizations now

anuraaga:
Oops, wrong branch...

PR #2153, opened 2019-10-03
Initialize path params map lazily to optimize routing that does not u…

anuraaga:
…se path params.

It is common for RPC routes to not have path params, but every routing result requires the allocation of a small `Map` used by the builder.

Also added a small optimization to presize the path params map for routes that do have them but haven't benchmarked it yet, it's probably tiny but still worth it.

After
```
Benchmark                     Mode  Cnt        Score         Error  Units
RoutersBenchmark.exactMatch  thrpt    5  7424958.423 ▒ 1921984.551  ops/s
```

Before
```
Benchmark                     Mode  Cnt        Score         Error  Units
RoutersBenchmark.exactMatch  thrpt    5  6953946.614 ▒ 1127234.099  ops/s
```

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/server/RoutersBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/server/RoutersBenchmark.java
@@ -34,7 +34,7 @@ import com.linecorp.armeria.server.logging.AccessLogWriter;
 public class RoutersBenchmark {
 
     private static final Service<HttpRequest, HttpResponse> SERVICE =
-            ((ctx, req) -> HttpResponse.of(HttpStatus.OK));
+            (ctx, req) -> HttpResponse.of(HttpStatus.OK);
 
     private static final List<ServiceConfig> SERVICES;
     private static final VirtualHost HOST;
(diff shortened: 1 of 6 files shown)


benchmarks/src/jmh/java/com/linecorp/armeria/server/RoutersBenchmark.java
```diff
@@ -0,0 +1,73 @@
...
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.RequestHeaders;
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+import com.linecorp.armeria.server.logging.AccessLogWriter;
+
+public class RoutersBenchmark {
+
+    private static final Service<HttpRequest, HttpResponse> SERVICE =
+            ((ctx, req) -> HttpResponse.of(HttpStatus.OK));
```
minwoox: nit: can remove the parenthesis.

benchmarks/src/jmh/java/com/linecorp/armeria/server/RoutersBenchmark.java
```diff
@@ -0,0 +1,73 @@
...
+                ContentPreviewerFactory.disabled(), ContentPreviewerFactory.disabled(),
+                AccessLogWriter.disabled(), false);
+        ROUTER = Routers.ofVirtualHost(HOST, SERVICES, RejectedRouteHandler.DISABLED);
+    }
+
+    @Benchmark
+    public Routed<ServiceConfig> exactMatch() {
+        final RoutingContext ctx = DefaultRoutingContext.of(HOST, "localhost", METHOD1_HEADERS.path(),
+                                                            null, METHOD1_HEADERS, false);
+        Routed<ServiceConfig> routed = ROUTER.find(ctx);
```
minwoox: nit: final

PR #2154, opened 2019-10-04
Enable FinalLocalVariable checkstyle check

anuraaga:
I always forget `final` (I think all except one file in this diff was written by me...). This will prevent forgetting `final` ever again.


settings/checkstyle/checkstyle-suppressions.xml
```diff
@@ -11,7 +11,9 @@
   <!-- Suppress all checks in generated sources. -->
   <suppress checks=".*" files="[\\/]gen-src[\\/]" />
   <!-- Suppress checks in large forks to make diffing against upstream easier. -->
-  <suppress checks=".*" files="[\\/]grpc[\\/]ServerCallImpl.java$" />
+  <suppress checks=".*" files="[\\/]common[\\/]MediaType.java$" />
+  <suppress checks=".*" files="|[\\/]internal[\\/]DefaultAttributeMap.java" />
+  <suppress checks=".*" files="[\\/]reactive[\\/]DefaultSslInfo.java" />
```
trustin: Could we only suppress `FinalLocalVariable` for these files?

anuraaga: For forks, isn't it better to not apply any of our style at all for easier diffs?

trustin:
Thanks, @anuraaga !

PR #2156, opened 2019-10-04
Issue1645

mumgmangmange:


trustin:
@mumgmangmange Gentle ping.

- You need to make sure `OpenSsl` is accessed only when a user starts a server with SSL enabled. Currently, `OpenSsl` is loaded even if you start a plaintext HTTP server only. We could fix this by making `Flags.useOpenSsl()` lazy.
- You need to make sure `Epoll` is accessed only when the current operating system is Linux. This is simpler than the `OpenSsl` case. You could just guard the access to `Epoll` with a simple `if` statement which checks the current operating system.

minwoox:
Let me close this issue because you are working on #2184.

PR #2157, opened 2019-10-05
Switch from Travis CI to AppVeyor BYOC

trustin:
Motivation:

AppVeyor recently launched BYOC self-hosted jobs, which allows us to
utilize our own machines for building projects. It is highly
cost-effective because we can overcome the concurrency/performance limit
enforced by AppVeyor or Travis CI.

Modifications:

- Switch from Travis CI to AppVeyor BYOC for building on Linux.
- Ignore `ArmeriaGrpcServerInteropTest.deadlineExceeded()` temporarily
  on Windows.

Result:

- Faster build


trustin:
Successfully set up 5 dockerized build agents with 8 vCPUs, 16 GB RAM and 100GB SDD. Currently configured to run up to 20 concurrent jobs, i.e. 20 x 5 = 100 concurrent jobs.

trustin:
Seems to work fine now, except that the cache is missing. I could not set it up because somehow it keeps saying there's no such access key ID. If we fail to set up the cache, maybe we should look for something else.

anuraaga:
BYOC with transparent caching seems hard as a concept since cross-cloud transfers would be slow anyways.

Can we manually add steps to upload/download to verda object store?

anuraaga:
For reference, maybe something like this

https://github.com/curioswitch/curiostack/blob/master/cloudbuild.yaml#L37

trustin:
> Can we manually add steps to upload/download to verda object store?

Verda provides S3-compatible object storage but it just doesn't work for some reason. Let me contact the Verda admins early next week just in case I misconfigured anything. If it doesn't work out well, we could run an S3-compatible server like https://github.com/minio/minio

trustin:
Cache is now working. I just used a local volume and `rsync` for speed and simplicity. Let me know what you think.

.appveyor.sh
```diff
@@ -0,0 +1,148 @@
...
+      -exec echo 'Purging:' {} ';' \
+      -delete
+
+  if [[ "$PURGE_CACHE" != '1' ]]; then
+    # Restore the home directory from the cache.
+    BRANCH_CACHE_DIR="/var/cache/appveyor/$APPVEYOR_ACCOUNT_NAME/$APPVEYOR_PROJECT_SLUG/branches/$APPVEYOR_REPO_BRANCH"
+    if [[ -z "$APPVEYOR_PULL_REQUEST_NUMBER" ]]; then
+      CACHE_DIR="$BRANCH_CACHE_DIR"
+    else
+      CACHE_DIR="/var/cache/appveyor/$APPVEYOR_ACCOUNT_NAME/$APPVEYOR_PROJECT_SLUG/pulls/$APPVEYOR_PULL_REQUEST_NUMBER"
```
anuraaga: We should use the same cache for all the builds. If it's like this I think every build is slow unless it is a follow up commit in a PR.

.appveyor.sh
```diff
@@ -0,0 +1,148 @@
...
+if [[ "$COVERAGE" -eq 1 ]]; then
+  # Send coverage reports to CodeCov.io.
+  # Note: In Linux, AppVeyor sets 'true' to 'CI' and 'APPVEYOR',
+  #       but CodeCov expects them to be 'True', so we set 'True' to them.
+  export CI=True
+  export APPVEYOR=True
+  bash <(curl -s https://codecov.io/bash) || true
+fi
+
+# Update the cache directory.
```
anuraaga: We should only update cache on master branch build

.appveyor.sh
trustin: If CACHE_DIR does not exist, it will use BRANCH_CACHE_DIR as a seed, so it should be fine once we merge this PR and the cache for master branch is built, no?

.appveyor.sh
```diff
@@ -0,0 +1,148 @@
...
+      CACHE_DIR="$BRANCH_CACHE_DIR"
+    else
+      CACHE_DIR="/var/cache/appveyor/$APPVEYOR_ACCOUNT_NAME/$APPVEYOR_PROJECT_SLUG/pulls/$APPVEYOR_PULL_REQUEST_NUMBER"
+    fi
+
+    # Fetch the home directory from the cache directory.
+    if [[ -d "$CACHE_DIR" ]]; then
+      touch "$CACHE_DIR"
+      echo_and_run rsync -a --stats "$CACHE_DIR/" "$HOME"
+    elif [[ -d "$BRANCH_CACHE_DIR" ]]; then
```
trustin: @anuraaga Here

anuraaga: Ah didn't notice the fallback. LGTM

trustin:
This is now ready for reviews.

trustin:
Thanks for reviewing!

PR #2139, opened 2019-10-02
Migrate debug page to hooks.

anuraaga:
It'd be nice if you can check out the debug page in this branch locally to make sure it works ok - I tried various operations and think it's fine but there are a lot of possible variations.

Also fixes long-time pet peeve by prettifying gRPC unframed JSON.

Fixes #1949

PR #2045, opened 2019-09-05
Scroll down to the debug form when a DocService URL contains …

SooJungDev:
Motivation:
- User has to manually scroll down the method page to see the debug form.

Modifications:
- Add move scroll function 

Result:
-  without scrolling down to it.
-  closed #1682

PR #2158, opened 2019-10-05
Fix a bug where a method page scrolls down to its debug form when the service list is collapsed

SooJungDev:
Motivation:
- When 1) the current `DocService` page URL has a request encoded in it and 2) a user collapses or expands the service list on the left pane, the current method page scrolls down to the debug form.
- The expected behavior: The method page must scroll down to the debug page only when the method page is open first time.

Modifications:
- Add DebugScroll Class and flag set/get method for limiting scroll

Result:
- The method page does not scroll to the debug form when a user collapses or expands the service list.

### DIFF
--- a/docs-client/src/containers/App/index.tsx
+++ b/docs-client/src/containers/App/index.tsx
@@ -38,6 +38,7 @@ import { Route, RouteComponentProps, withRouter } from 'react-router-dom';
 import EnumPage from '../EnumPage';
 import HomePage from '../HomePage';
 import MethodPage from '../MethodPage';
+import { DebugScroll } from '../MethodPage/DebugScroll';
 import StructPage from '../StructPage';
 
 import {
@@ -47,7 +48,6 @@ import {
 } from '../../lib/specification';
 
 import GotoSelect from '../../components/GotoSelect';
-
 import {
   extractSimpleArtifactVersion,
   Version,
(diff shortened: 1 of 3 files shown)


docs-client/src/containers/App/index.tsx
```diff
@@ -330,19 +330,24 @@ interface OpenServices {
   [name: string]: boolean;
 }
 
-const toggle = (current: boolean) => !current;
-
 const App: React.FunctionComponent<Props> = (props) => {
+  const toggle = (current: boolean) => {
+    DebugScroll.setIsScroll(false);
```
ikhoon: How about using `useState`? 😀 I wonder there are any problem to apply it.

anuraaga:
Can you describe more what the actual bug is? I don't quite understand from the PR description and code.

ikhoon:
I know there is a bug in https://github.com/line/armeria/pull/2045
@SooJungDev Could you paste some screenshots for before and after :-)

SooJungDev:
I have attached the image 
I hope it helps you understand.

- before
before

- after
after

trustin:
@anuraaga Updated the pull request description. Does it make sense now? If you're OK, I'd like to merge this before your React Hooks PR. Are you?

anuraaga:
Thanks, the images make it very clear! I think both myself and @ikhoon are feeling that a global variable isn't the best solution to the issue but I have to think more after some sleep about it.  

And no worries on ordering of PR merges, can update as needed.

anuraaga:
So the root cause is our equality comparison in `componentDidUpdate`

https://github.com/line/armeria/blob/master/docs-client/src/containers/MethodPage/DebugPage.tsx#L118

The boolean guards here force the problem away but a better fix would be switch to comparing the contents of the object there rather than equality comparison.

Incidentally, `useEffect(..., [match.params])` does do a normal comparison so the bug is actually already fixed in https://github.com/line/armeria/pull/2139 . Do you think it makes sense to just focus on that one?

PR #2159, opened 2019-10-06
Fix where access log is not written when the method of a reques…

minwoox:
…ot allowed

Motivation:
There's some cases that we don't call `RequestLogBuilder.endRequest()`.
- When the method of the request is not allowed
- When there are no services that match the path of the request
- When the path of the request is invalid
- and so on

Because the `RequestLog` is never complete, `AccessLogWriter` is not called so we should fix it.

Modification:
- Call `RequestLogBuilder.endRequest()` even when `service.serve()` is not invoked

Result:
- You can now see the access logs when the method of a request is not allowed and there are no services that match the path.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/server/HttpServerHandlerTest.java
+++ b/core/src/test/java/com/linecorp/armeria/server/HttpServerHandlerTest.java
@@ -68,7 +68,7 @@ class HttpServerHandlerTest {
         final AggregatedHttpResponse res = client.delete("/hello").aggregate().join();
         assertThat(res.status()).isSameAs(HttpStatus.METHOD_NOT_ALLOWED);
         await().untilAsserted(() -> assertThat(logHolder.get().path()).isEqualTo("/hello"));
-        assertThat(logHolder.get().requestCause()).isExactlyInstanceOf(HttpStatusException.class);
+        assertThat(logHolder.get().requestCause()).isNull();
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/linecorp/armeria/server/HttpServerHandlerTest.java
```diff
@@ -0,0 +1,100 @@
...
+    void handleNonExistentMapping() {
+        final HttpClient client = HttpClient.of(server.uri("/"));
+        final AggregatedHttpResponse res = client.get("/non_existent").aggregate().join();
+        assertThat(res.status()).isSameAs(HttpStatus.NOT_FOUND);
+        await().untilAsserted(() -> assertThat(logHolder.get().path()).isEqualTo("/non_existent"));
+        assertThat(logHolder.get().requestCause()).isNull();
+    }
+
+    @Test
+    void httpStatusException() {
```
minwoox: I found out that we leave the `httpStatusException` as the cause but not for `httpResponseException`.
I think we should treat them in the same way and I think we do not log the causes in both cases.

trustin: Agreed. Could you update accordingly?

minwoox: Fixed. PTAL 🙇

trustin:
@ikhoon Still LGTY?

minwoox:
Thanks for reviewing! 😉

Issue #2150, opened 2019-10-03
Consider to remove sorting at MeterIdPrefix


imasahiro:
Currently, we sort the micrometer tags at MeterIdPrefix to create Meter ID stably.
https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefix.java#L95-L101

Since I saw this stacktrace from our service’s flamegraph, I would like to consider whether it is avoidable or not.  Since it doesn’t have many samples, we can be ignored but because it is not optimal, it would be happy if we can avoid this sort.

<img width="1305" alt="Screen Shot 2019-10-03 at 13 11 03" src="https://user-images.githubusercontent.com/73532/66105045-522deb00-e5f5-11e9-81cb-c4680be72b6a.png">

trustin:
We can make use of the following characteristics:

- The number of tag names is very small (less than 10 usually)
- We always construct a new sorted tag list because `MeterIdPrefix` is immutable.
- When merging two tag lists:
  - the first tag list is always sorted.
  - the second tag list is also very small.

anuraaga:
I don't think this would be an issue if we sorted during metric export instead of when recording metrics.

Do you think it's possible to move sorting to a different layer that only occurs during export?

trustin:
That's a good point. We could:

- Sort only when creating a meter.
- Make `MeterIdPrefix.equals()` and `hashCode()` order-independent.

anuraaga:
I played with this a little and found

- Micrometer sorts before creating a meter anyways, so we don't need to worry about it. Removing sorting still allows prometheus integration test to pass even though it adds tags out-of-order

- We need to do order-independent hashcode/equals for the memoizing that happens in MicrometerUtil.register. However, I don't know if it's possible to do this faster than the current sort. I tried these three techniques, all slower than `HEAD`, and in this order of performance. Do you know of any other trick for unordered `Equals` that might be faster than the sort we do?
  - `this.tags.containsAll(that.tags) && that.tags.containsAll(this.tags)` (probably fast right now since by default we don't have that many tags)
  - Store field `this.unorderedTags = ImmutableSet.copyOf(tags)`, `this.unorderedTags.equals(that.unorderedTags)`
  - `ImmutableSet.copyOf(this.tags).equals(ImmutableSet.copyOf(that.tags));`

Also of the optimizations in 

https://github.com/line/armeria/issues/2150#issuecomment-538215947

[...]

PR #2163, opened 2019-10-07
Add default tags in order.

anuraaga:
This is the smallest solution for helping with the performance of tag addition. Our default logic can at least add tags in order to prevent reordering when sorting (TimSort handles already-sorted arrays well without special logic).

Also removes comparator when sorting since `Tag` is `Comparable` and adds tests for `Equals/HashCode` for use in playing with those functions.

For #2150

After
```
Benchmark                                        Mode  Cnt        Score         Error  Units
RequestMetricSupportBenchmark.registerSameTags  thrpt    5  6816146.723 ▒ 1548641.211  ops/s
```

Before
```
Benchmark                                        Mode  Cnt        Score        Error  Units
RequestMetricSupportBenchmark.registerSameTags  thrpt    5  5813367.536 ▒ 154548.384  ops/s
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefix.java
+++ b/core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefix.java
@@ -17,17 +17,16 @@
 package com.linecorp.armeria.common.metric;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static java.util.Comparator.comparing;
 import static java.util.Objects.requireNonNull;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 
 import javax.annotation.Nullable;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
 
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.Tag;
@@ -97,12 +96,12 @@ public final class MeterIdPrefix {
             return ImmutableList.of();
         }
 
-        tags.sort(comparing(Tag::getKey));
+        Collections.sort(tags);
         return ImmutableList.copyOf(tags);
     }
 
     private static ImmutableList<Tag> copyAndSort(Iterable<Tag> tags) {
-        return sort(Lists.newArrayList(tags));
+        return ImmutableList.sortedCopyOf(tags);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -71,21 +71,24 @@ static MeterIdPrefixFunction ofDefault(String name) {
         return new MeterIdPrefixFunction() {
             @Override
             public MeterIdPrefix activeRequestPrefix(MeterRegistry registry, RequestLog log) {
-                final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(4);
-                buildTags(tagListBuilder, log);
+                // hostNamePattern, method, route
+                final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(3);
+                // For optimal performance, add tags in order.
```
trustin: alphabetical order or lexical order?

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -71,21 +71,24 @@ static MeterIdPrefixFunction ofDefault(String name) {
...
             }
 
             @Override
             public MeterIdPrefix apply(MeterRegistry registry, RequestLog log) {
-                // method, hostNamePattern, route, status
+                // hostNamePattern, httpStatus, method, route
                 final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(4);
-                buildTags(tagListBuilder, log);
-                RequestMetricSupport.appendHttpStatusTag(tagListBuilder, log);
+                // For optimal performance, add tags in order.
```
trustin: alphabetical order or lexical order?

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefix.java
```diff
@@ -97,12 +96,12 @@ private static void zip(List<Tag> list, String... tags) {
             return ImmutableList.of();
         }
 
-        tags.sort(comparing(Tag::getKey));
+        Collections.sort(tags);
```
adriancole: don't you want `Collections.ragSort(tags)`?

PR #2168, opened 2019-10-07
Upload test reports to file.io

trustin:
.. so that we can browse the test result more conveniently.


trustin:
It kind of works, but:

- it takes extremely long to upload all test reports.
- there's no way to see the standard output/error for a failed test. (can see an exception and its stack trace though)

I think it's maybe better compressing the test reports into a tarball and uploading it somewhere.

trustin:
We now upload the gzipped tarball of test reports to file.io when a build fails. We are capped at 100 uploads a day, but I guess it should be enough for us. :laughing:

trustin:
Thanks!

PR #2172, opened 2019-10-08
Enable parallel build in CI jobs / Run only essential tasks on Windows

trustin:
- Our build machines can afford a parallel build.
  - Enable parallelism with 4 maximum workers. 
  - This makes the build finish in 10 minutes (previously 20+ minutes).
- Run only essential tasks on Windows to reduce the build time.
  - No Checkstyle
  - No shading
  - No docs-client
  - Only a few tasks:
    - `:core:test`
    - `:grpc:test`
    - `:thrift:test`
    - `:it:server:test`

### DIFF
--- a/.appveyor.build.sh
+++ b/.appveyor.build.sh
@@ -136,11 +136,11 @@ fi
 
 msg "Building .."
 # Run 'checkstyle' first so we don't waste our time building ill-formatted code.
-echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=2 checkstyle
+echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=3 checkstyle
 # Run 'trimShadedJar' alone because it pollutes the console if other tasks are running in parallel.
-echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=2 trimShadedJar
+echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=3 trimShadedJar
 # Run the remaining tasks.
-echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=2 build -xtrimShadedJar
+echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=3 build -xtrimShadedJar
 
 if [[ "$COVERAGE" -eq 1 ]]; then
   # Send coverage reports to CodeCov.io.
--- a/.appveyor.build.sh
+++ b/.appveyor.build.sh
@@ -135,12 +135,7 @@ if [[ "$COVERAGE" -eq 1 ]]; then
 fi
 
 msg "Building .."
-# Run 'checkstyle' first so we don't waste our time building ill-formatted code.
-echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=4 checkstyle
-# Run 'trimShadedJar' alone because it pollutes the console if other tasks are running in parallel.
-echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=4 trimShadedJar
-# Run the remaining tasks.
-echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=4 build -xtrimShadedJar
+echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=4 checkstyle build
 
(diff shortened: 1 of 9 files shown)


trustin:
Will update the Windows side once Linux side is done. Not sure it will work though.

.appveyor.yml
```diff
@@ -1,6 +1,6 @@
 version: '{branch}.{build}'
 environment:
-  GRADLE_CLI_OPTS: '--no-daemon --console=verbose --stacktrace'
```
minwoox: Is there any reason that you remove this?

trustin: Because we now launch Gradle twice?

minwoox: I thought each build has its own process but it's probably a bad idea because we have to pay for the cost of JVM startup for every build. Thanks!

trustin: Each build runs in its own Docker container, so Gradle daemons are shared only among the current build tasks.

minwoox: Ah, I mean each `./gradlew` commands. Sorry for choosing the wrong word. 😄

.appveyor.build.sh
```diff
@@ -135,7 +135,12 @@ if [[ "$COVERAGE" -eq 1 ]]; then
 fi
 
 msg "Building .."
-echo_and_run ./gradlew $GRADLE_CLI_OPTS checkstyle check build
+# Run 'checkstyle' first so we don't waste our time building ill-formatted code.
+echo_and_run ./gradlew $GRADLE_CLI_OPTS --parallel --max-workers=4 checkstyle
```
ikhoon: Awwwwwwwwwwwesome!

trustin:
Build time has been reduced from 23 minutes to 12 minutes with 4 workers. Let me try smaller number of workers to see the difference.

.appveyor.yml
trustin: I added it back since we now run only a single `./gradlew` command. ProGuard emits some noisy messages but it's much less than before.

core/build.gradle
```diff
@@ -182,5 +182,7 @@ if (tasks.findByName('trimShadedJar')) {
         // Do not optimize the dependencies that access some fields via sun.misc.Unsafe or reflection only.
         keep "class com.linecorp.armeria.internal.shaded.caffeine.** { *; }"
         keep "class com.linecorp.armeria.internal.shaded.jctools.** { *; }"
+
+        dontnote
```
trustin: This reduces the verbosity of `ProGuardTask` when running in parallel.

trustin:
PTAL again, @ikhoon and @minwoox. /cc @anuraaga

.appveyor.yml
```diff
@@ -24,7 +24,7 @@ init:
 
 build_script:
   - cmd: 'gradlew.bat --version'
-  - cmd: 'gradlew.bat %GRADLE_CLI_OPTS% checkstyle check build'
+  - cmd: 'gradlew.bat %GRADLE_CLI_OPTS% test testStreaming testNg -xyarnSetup -xyarn -xbuildWeb -xcheckstyle -xcheckstyleMain -xcheckstyleJava9 -xcheckstyleTest -xcheckstyleJava9Test -xcheckstyleJmh'
```
anuraaga: We already have a `checkstyleEnabled` local variable in `java.gradle` - how about making a one line change so it checks a property like `com.linecorp.checkstyleDisabled`?

trustin: Now using `noCheckstyle` and also added `noWeb`

thrift0.11/build.gradle
```diff
@@ -44,5 +44,7 @@ ext {
 }
 
 // Disable checkstyle because it's checked by ':thrift'.
-tasks.checkstyleMain.onlyIf { false }
-tasks.checkstyleTest.onlyIf { false }
+if (!project.hasProperty('noCheckstyle')) {
```
anuraaga: Think it's better to guard with `plugins.withId('checkstyle')`

trustin: Realized that `Checkstyle` is a core task class which is always available, so I just used `tasks.withType(Checkstyle)` because it is safer.

trustin:
Not sure if it's the best idea, but I updated `.appveyor.yml` so it runs only the following tasks on Windows:

- `:core:test`
- `:grpc:test`
- `:thrift:test`
- `:it:server:test`

Please let me know if you think I went too far or if we need to add more tasks.

PR #2174, opened 2019-10-09
Add UUID To RequestContext (#2001)

yanghun0070:
Motivation:
- It would be useful, and probably simple, to assign a UUID to every request to add to the log to be able to associate requests and responses. 

Modifications:
- Add  `UUID` to RequestContext.

Result:
-  assign a different UUID to every requests and responses context.


trustin:
@yanghun0070 Could you please rebase your work?

yanghun0070:
@trustin I already rebase my work.

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
```diff
@@ -163,7 +164,12 @@ protected void getAsync(HttpClient client, String pathIncludingQuery) throws Exc
...
         DummyRequestContext() {
             super(NoopMeterRegistry.get(), SessionProtocol.HTTP,
-                  HttpMethod.GET, "/", null, HttpRequest.streaming(HttpMethod.GET, "/"), null);
+                  HttpMethod.GET, "/", UUID.randomUUID(), null, HttpRequest.streaming(HttpMethod.GET, "/"), null);
+        }
+
+        @Override
+        public RequestContext newDerivedContext() {
+            throw new UnsupportedOperationException();
         }
```
trustin: You don't need to override this method.

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
```diff
@@ -433,6 +437,13 @@ protected final String path() {
         return path;
     }
 
+    /**
+     * Returns the {@link UUID} of the request.
+     */
+    protected final UUID uuid() {
+        return uuid;
+    }
```
trustin: - Could you add a public setter of the `uuid` as well?
- We could start with `null` `UUID` and then call `UUID.randomUUID()` only when a user did not call `builder.uuid(UUID)` (i.e. `this.uuid == null`).

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -54,6 +55,7 @@
     private volatile HttpRequest req;
     @Nullable
     private volatile RpcRequest rpcReq;
+    protected UUID uuid;
```
trustin: - `private final` unless there is a good reason to make it mutable and protected.
- I guess we need to move this field right before `method`, because it is an essential non-null per-request field.

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -67,18 +69,20 @@
...
      *
      * @param sessionProtocol the {@link SessionProtocol} of the invocation
+     * @param uuid the {@link UUID} associated with this context
      * @param req the {@link HttpRequest} associated with this context
      * @param rpcReq the {@link RpcRequest} associated with this context
      */
     protected NonWrappingRequestContext(
             MeterRegistry meterRegistry, SessionProtocol sessionProtocol,
-            HttpMethod method, String path, @Nullable String query,
+            HttpMethod method, String path, UUID uuid, @Nullable String query,
```
trustin: How about moving `uuid` right before `method`? Please make the same change for all `RequestContext` constructors you modified.

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -67,18 +69,20 @@
...
             @Nullable HttpRequest req, @Nullable RpcRequest rpcReq) {
 
         this.meterRegistry = requireNonNull(meterRegistry, "meterRegistry");
         this.sessionProtocol = requireNonNull(sessionProtocol, "sessionProtocol");
         this.method = requireNonNull(method, "method");
         this.path = requireNonNull(path, "path");
+        this.uuid = requireNonNull(uuid, "uuid");
```
trustin: Ditto - Please move this assignment right before `method` assignment. Same for all similar changes in this PR.

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -191,6 +192,11 @@
     @Nullable
     String query();
 
+    /**
+     * Returns the {@link UUID} that contains the information about the current {@link Request}.
+     */
+    UUID uuid();
```
trustin: Could you move this method right before `method()`?

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -178,6 +182,11 @@ public final String query() {
         return query;
     }
 
+    @Override
+    public final UUID uuid() {
+        return uuid;
+    }
```
trustin: Could you move this method right before `method()`?

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -159,7 +160,7 @@ private DefaultServiceRequestContext(
             long requestStartTimeMicros) {
 
         super(meterRegistry, sessionProtocol,
-              requireNonNull(routingContext, "routingContext").method(), routingContext.path(),
+              requireNonNull(routingContext, "routingContext").method(), routingContext.path(), UUID.randomUUID(),
```
trustin: Please make `Default{Client,Service}RequestContext` have symmetric and consistent API, i.e. Let a user specify an UUID as a parameter just like you did in `DefaultClientRequestContext`.

trustin (changes requested):
When creating a derived context, a user might want or *not* want to assign a new UUID to the derived context. How about adding `UUID uuid` as an additional parameter of `newDerivedContext(HttpRequest, RpcRequest)` and `newDerivedContext(HttpRequest, RpcRequest, Endpoint)`?

I'm somewhat unsure about what the behavior of `newDerivedContext()` is supposed to be. Should we:

- Always assign a new random UUID,
- Add `UUID` as a parameter there as well or
- Just remove `newDerivedContext()` to get rid of ambiguity..?

Let me know what you think. @minwoox @ikhoon @anuraaga

trustin:
And perhaps we could add `RequestLog.uuid()` getter as well, which is a simple shortcut to `RequestLog.context().uuid()`.

minwoox:
> Just remove newDerivedContext() to get rid of ambiguity..?

+1 for this.
IIRC, the `newDerivedContext` is requested by one of our customers who was not using `RetryingClient`. He implemented his own retrying client because at that time `Retryingclient` had a bug that uses only one `requestLog` no matter how many retry is made so he could not log the causes and responses successfully.
But right now, we fixed it and it works well. I don't think we need it anymore.

benchmarks/src/jmh/java/com/linecorp/armeria/server/RoutersBenchmark.java
```diff
@@ -62,7 +63,7 @@
 
     @Benchmark
     public Routed<ServiceConfig> exactMatch() {
-        final RoutingContext ctx = DefaultRoutingContext.of(HOST, "localhost", METHOD1_HEADERS.path(),
+        final RoutingContext ctx = DefaultRoutingContext.of(HOST, "localhost", UUID.randomUUID(), METHOD1_HEADERS.path(),
```
minwoox: Seems like this line is longer than 112 which we constraint using checkstyle.
Could you rebase master and do `./gradlwe checkstyle` so that you can notice the violation?

minwoox: How about making a field for `uuid` and using it here?

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
```diff
@@ -163,7 +164,7 @@ protected void getAsync(HttpClient client, String pathIncludingQuery) throws Exc
     private static class DummyRequestContext extends NonWrappingRequestContext {
         DummyRequestContext() {
             super(NoopMeterRegistry.get(), SessionProtocol.HTTP,
-                  HttpMethod.GET, "/", null, HttpRequest.streaming(HttpMethod.GET, "/"), null);
+                  UUID.randomUUID(), HttpMethod.GET, "/",  null, HttpRequest.streaming(HttpMethod.GET, "/"), null);
```
minwoox: Could you remove extra space after `"/",`?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -97,15 +98,16 @@
      *
      * @param eventLoop the {@link EventLoop} associated with this context
      * @param sessionProtocol the {@link SessionProtocol} of the invocation
+     * @param uuid the {@link UUID} that contains the information about the current {@link HttpRequest}.
```
minwoox: `uuid the {@link UUID} that represents the current {@link Request} and {@link Response}`

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -260,7 +263,6 @@ private DefaultClientRequestContext(DefaultClientRequestContext ctx,
         if (ctx.rpcRequest() != null) {
             requireNonNull(rpcReq, "rpcReq");
         }
-
```
minwoox: revert?

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
```diff
@@ -433,6 +436,24 @@ protected final String path() {
...
+        return this;
+    }
+
+    /**
+     * Returns the {@link UUID} of the request.
+     */
+    protected final UUID uuid() {
+        if(uuid == null) {
+            uuid = UUID.randomUUID();
+        }
```
minwoox: We don't need this because it's always not null?

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
```diff
@@ -167,6 +168,13 @@ default void ensureAvailability(Iterable<RequestLogAvailability> properties) {
      */
     RequestContext context();
 
+    /**
+     * Returns the method of the {@link Request}. This method is a shortcut to {@code context().uuid()}.
```
minwoox: the method -> the uuid

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -159,6 +160,7 @@ private DefaultServiceRequestContext(
             long requestStartTimeMicros) {
 
         super(meterRegistry, sessionProtocol,
+              requireNonNull(routingContext, "routingContext").uuid(),
               requireNonNull(routingContext, "routingContext").method(), routingContext.path(),
```
minwoox: `requireNonNull(routingContext, "routingContext").method()` -> `requireNonNull(routingContext).method()` because we don't want to check it twice?

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -321,9 +322,8 @@ private void handleRequest(ChannelHandlerContext ctx, DecodedHttpRequest req) th
 
         final String hostname = hostname(headers);
         final VirtualHost host = config.findVirtualHost(hostname);
-
         final RoutingContext routingCtx =
-                DefaultRoutingContext.of(host, hostname, pathAndQuery.path(), pathAndQuery.query(),
+                DefaultRoutingContext.of(host, hostname, UUID.randomUUID(), pathAndQuery.path(), pathAndQuery.query(),
```
minwoox: I think we don't need this `uuid` for `RoutingContext`. If we need that we have to use the one from the `requestLog`. WDYT? /cc @trustin, @ikhoon

minwoox (commented):
I think we need to rebase and run the check style task 😄

minwoox:
From #2001, we need to leave uuid in `toStringRequestOnly` and `toStringResponseOnly`. Could you add it please?

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
trustin: Agreed.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
trustin: the method -> the UUID

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: `... represents the unique identifier of the current {@link Request} and {@link Response} pair.`

benchmarks/src/jmh/java/com/linecorp/armeria/server/RoutersBenchmark.java
trustin: +1

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
```diff
@@ -98,7 +100,7 @@ protected AbstractRequestContextBuilder(boolean server, HttpRequest req) {
         this.req = requireNonNull(req, "req");
         rpcReq = null;
         sessionProtocol = SessionProtocol.H2C;
-
+        uuid = UUID.randomUUID();
```
trustin: Could you begin `uuid` with `null`, and call `UUID.randomUUID()` only when a user did not set the UUID of the context?

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
trustin: Correct. `uuid` must start with `null`.

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
```diff
@@ -433,6 +436,24 @@ protected final String path() {
...
+    public AbstractRequestContextBuilder uuid(UUID uuid) {
+        this.uuid = requireNonNull(uuid, "uuid");
+        return this;
+    }
+
+    /**
+     * Returns the {@link UUID} of the request.
+     */
+    protected final UUID uuid() {
+        if(uuid == null) {
```
trustin: Missing space between `if` and `(`

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -67,18 +69,20 @@
...
             MeterRegistry meterRegistry, SessionProtocol sessionProtocol,
-            HttpMethod method, String path, @Nullable String query,
+            UUID uuid, HttpMethod method, String path, @Nullable String query,
             @Nullable HttpRequest req, @Nullable RpcRequest rpcReq) {
 
         this.meterRegistry = requireNonNull(meterRegistry, "meterRegistry");
         this.sessionProtocol = requireNonNull(sessionProtocol, "sessionProtocol");
         this.method = requireNonNull(method, "method");
         this.path = requireNonNull(path, "path");
+        this.uuid = requireNonNull(uuid, "uuid");
```
trustin: This assignment should occur before the assignment of `method`, so that the field initialization takes place in order.

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -167,6 +168,11 @@
     @Nullable
     SSLSession sslSession();
 
+    /**
+     * Returns the {@link UUID} that contains the information about the current {@link Request}.
```
trustin: `Returns the {@link UUID} that represents the unique identifier of ....` (See @minwoox's comment)

trustin (changes requested):

benchmarks/src/jmh/java/com/linecorp/armeria/server/RoutersBenchmark.java
```diff
@@ -17,6 +17,7 @@
 package com.linecorp.armeria.server;
 
 import java.util.List;
+import java.util.UUID;
```
trustin: Please fix Checkstyle violations.

brave/src/test/java/com/linecorp/armeria/client/brave/BraveClientIntegrationTest.java
trustin: This is not fixed yet.

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
```diff
@@ -98,7 +100,9 @@ protected AbstractRequestContextBuilder(boolean server, HttpRequest req) {
         this.req = requireNonNull(req, "req");
         rpcReq = null;
         sessionProtocol = SessionProtocol.H2C;
-
+        if (uuid == null) {
+            uuid = UUID.randomUUID();
+        }
```
trustin: This does not make any sense. Remove this block entirely.

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
```diff
@@ -433,6 +438,24 @@ protected final String path() {
         return path;
     }
 
+    /**
+     * Sets the UUID of the request. If not set, it is auto-generated with UUID
+     */
+    public AbstractRequestContextBuilder uuid(UUID uuid) {
+        this.uuid = requireNonNull(uuid, "uuid");
+        return this;
+    }
```
trustin: Make sure to override this method in `{Client,Service}RequestContextBuilder` and change the return type, e.g.

```java
class ClientRequestContextBuilder ... {
    ...
    @Override
    public ClientRequestContextBuilder uuid(UUID uuid) {
        return (ClientRequestContextBuilder) super.uuid(uuid);
    }
    ...
}
```

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -158,8 +159,8 @@ private DefaultServiceRequestContext(
             InetAddress clientAddress, boolean requestStartTimeSet, long requestStartTimeNanos,
             long requestStartTimeMicros) {
 
-        super(meterRegistry, sessionProtocol,
-              requireNonNull(routingContext, "routingContext").method(), routingContext.path(),
+        super(meterRegistry, sessionProtocol, UUID.randomUUID(),
+              requireNonNull(routingContext).method(), routingContext.path(),
```
trustin: As I asked before already, Make sure `DefaultServiceRequestContext` constructors require `UUID` as a parameter, just like `DefaultClientRequestContext` does.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -184,7 +185,6 @@ private DefaultServiceRequestContext(
         // there is some delay between the actual channel read and this logging, but it's the best we can do for
         // now.
         log.requestFirstBytesTransferred();
-
```
trustin: Revert

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -321,7 +322,6 @@ private void handleRequest(ChannelHandlerContext ctx, DecodedHttpRequest req) th
 
         final String hostname = hostname(headers);
         final VirtualHost host = config.findVirtualHost(hostname);
-
```
trustin: Revert

core/src/test/java/com/linecorp/armeria/common/RequestContextTest.java
```diff
@@ -442,7 +443,7 @@ private NonWrappingRequestContext createContext(boolean addContextAwareHandler)
     private class DummyRequestContext extends NonWrappingRequestContext {
         DummyRequestContext() {
             super(NoopMeterRegistry.get(), SessionProtocol.HTTP,
-                  HttpMethod.GET, "/", null, HttpRequest.streaming(HttpMethod.GET, "/"), null);
+                  UUID.randomUUID(), HttpMethod.GET, "/",null, HttpRequest.streaming(HttpMethod.GET, "/"), null);
```
trustin: Missing space between `,` and `null`

core/src/test/java/com/linecorp/armeria/server/RouteTest.java
```diff
@@ -23,6 +23,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
+
```
trustin: Revert

trustin (changes requested):

trustin:
Seems like the PR history has gone haywire :sob: Could you try to fix it so this pull request contains only your commits? Please consider creating a new pull request if you think it'll be easier for you.

ikhoon:
Close and replace with #2203

Issue #2023, opened 2019-08-23
Per-service setting is not possible for annotated service


minwoox:
A user can configure per-service settings using fluent `Route` API.
However, this is only for the services which implement `HttpService`, not for the annotated service.
We should provide a way for annotated service as well.

sivaalli:
Hi @trustin @minwoox @ikhoon ,

I think what we could do is, add an annotation called `@Configuration` that has properties/fields that reflect all the settings that `Route` api supports. For example a user can declare `@Configuration(maxRequestLength = 1024, requestTimeoutMillis=5000)` like this on a service method. Instead of creating a new Annotation(in this example `@Configuration`), we can also introduce route specific properties on `@Path`. What do you think? 

If you guys have anything else in mind please feel free to share it so I can implement it.

trustin:
How about just adding `buildAnnotated(Object)` to `ServiceBindingBuilder`?

minwoox:
Hello, @sivaalli! Thanks for the suggestion. However, I think adding more annotations is the last thing I want to do. 😄 

Had a chat with @trustin and we thought adding a builder would be enough for this.
```java
ServerBuilder sb = Server.builder()
                         .annotatedService() // Not sure this is the best name.
                         .requestTimeoutMillis(5000)
                         .contentPreview(...)
                         ...
                         .build(object); // Return to the ServerBuilder
```

All the properties that a service might want to set separately are gathered in `AbstractServiceBindingBuilder` thanks to @ikhoon.
However, the builder for annotated service just cannot inherit from it because of `AbstractBindingBuilder` which `AbstractServiceBindingBuilder` is inherited from.
```java
- AbstractBindingBuilder
  - AbstractServiceBindingBuilder
    - ServiceBindingBuilder
```
So I think we have to tear off `AbstractServiceBindingBuilder` from `AbstractBindingBuilder` first. We can make an interface for the service configuration binding, so `ServiceBindingBuilder` implements the interface while extending `AbstractBindingBuilder`, and `AnnotatedServiceBuilder`(?) will implements the interface only.
What do you think?

sivaalli:
Hi @trustin and @minwoox, I think I may have misunderstood this issue. So just want to reiterate the idea. For example, if this is my service:

```
class MyService{

    @Get("/foo")
    public HttpResponse foo(){
        return HttpResponse.of(200);
    }

    @Post("/bar")
    public HttpResponse bar(){
        return HttpResponse.of(200);
    }
}
```
[...]

PR #2180, opened 2019-10-10
Adding AnnotatedServiceBuilder to add service level settings to annotated service

sivaalli:
Motivation:

It is good to provide an ability to user to configure per service level settings for annotated service.

Closes: https://github.com/line/armeria/issues/2023

Modifications:

1. Pulled all methods that `AbstractServiceBindingBuilder`(minus AbstractBindingBuilder)  has into an interface called `ServiceBuilder`
2. `AbstractServiceBindingBuilder` implements this(ServiceBuilder) interface and also extends `AbstractBindingBuilder`
3. Provided a default implementation of `ServiceBuilder` called `DefaultServiceBuilder`, this is shared between `AbstractServiceBindingBuilder` and `AbstractServiceBuilder` to avoid code duplication

Result:
AnnotatedService can now be configured to have service specific properties.


core/src/main/java/com/linecorp/armeria/server/DefaultServiceBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+
+    public Service<HttpRequest, HttpResponse> decorate(Service<HttpRequest, HttpResponse> service) {
+        if (decorator == null) {
+            return service;
+        }
+        return service.decorate(decorator);
+    }
+
+    public ServiceConfigBuilder serviceConfigBuilder(Route route, Service<HttpRequest, HttpResponse> service) {
+        final ServiceConfigBuilder serviceConfigBuilder = new ServiceConfigBuilder(route, service);
```
sivaalli: I could decorate a service with the decorate here, but `AnnotatedServiceBindingBuilder` needs exceptionHandlingDecorators as the last decorator https://github.com/line/armeria/pull/2180/files#diff-162326dc40b1d719dcad478d6ac513bcR185

minwoox:
Thanks a lot, @sivaalli, for your work!
Your PR is mostly looking good except for the names of the classes. As you know we have so many classes whose names are similar such as `ServiceConfig`, `ServiceConfigBuilder`, `ServiceBindingBuilder`, `AbstractBindingBuilder`, `AbstractServiceBuilder`, `AbstractServiceBindingBuilder` and so on. This is not the problem you introduced, but we used to have from the past. So I think we need to clean that first.
@trustin, @ikhoon and I will discuss this naming and let you know how we can improve this situation.
Thanks a lot again and sorry for the late review. 🙇

sivaalli:
> Thanks a lot, @sivaalli, for your work!
> Your PR is mostly looking good except for the names of the classes. As you know we have so many classes whose names are similar such as `ServiceConfig`, `ServiceConfigBuilder`, `ServiceBindingBuilder`, `AbstractBindingBuilder`, `AbstractServiceBuilder`, `AbstractServiceBindingBuilder` and so on. This is not the problem you introduced, but we used to have from the past. So I think we need to clean that first.
> @trustin, @ikhoon and I will discuss this naming and let you know how we can improve this situation.
> Thanks a lot again and sorry for the late review. 🙇

No problem at all. :)

core/src/main/java/com/linecorp/armeria/server/DefaultServiceBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+        if (this.decorator != null) {
+            this.decorator = this.decorator.andThen(castDecorator);
+        } else {
+            this.decorator = castDecorator;
+        }
+
+        return this;
+    }
+
+    public Service<HttpRequest, HttpResponse> decorate(Service<HttpRequest, HttpResponse> service) {
```
trustin: The non-overridden public methods in this class can be package-private because this class is package-private.

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,191 @@
...
+ *   .pathPrefix("/foo")
+ *   .verboseResponses(true)
+ *   .contentPreview(500)
+ *   .buildAnnotated(new Service())                         // Return to the ServerBuilder.
+ *   .build();
+ * }</pre>
+ *
+ * @see ServiceBindingBuilder
+ */
+public class AnnotatedServiceBindingBuilder extends AbstractServiceBuilder {
```
trustin: Could be `final`?

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,191 @@
...
+
+    private final ServerBuilder serverBuilder;
+    private final Builder<ExceptionHandlerFunction> exceptionHandlerFunctionBuilder = ImmutableList.builder();
+    private final Builder<RequestConverterFunction> requestConverterFunctionBuilder = ImmutableList.builder();
+    private final Builder<ResponseConverterFunction> responseConverterFunctionBuilder = ImmutableList.builder();
+    private String pathPrefix = "/";
+
+    public AnnotatedServiceBindingBuilder(ServerBuilder serverBuilder) {
```
trustin: Could be package-private?

sivaalli:
@trustin, thanks for your review. I have made requested changes. Please take a look :)

core/src/main/java/com/linecorp/armeria/server/DefaultServiceBuilder.java
trustin: I see. How about adding that as a Javadoc of this method?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1035,6 +1035,15 @@ public ServerBuilder annotatedService(String pathPrefix, Object service,
         return this;
     }
 
+    /**
+     * Returns a {@link AnnotatedServiceBindingBuilder} to build annotated service.
+     *
```
trustin: nit: This empty Javadoc line could be removed.

core/src/main/java/com/linecorp/armeria/server/ServiceBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Decorates a {@link Service} with the specified {@code decorator}.
+     *
+     * @param decorator the {@link Function} that decorates the {@link Service}
+     * @param <T> the type of the {@link Service} being decorated
+     * @param <R> the type of the {@link Service} {@code decorator} will produce
+     */
+    <T extends Service<HttpRequest, HttpResponse>, R extends Service<HttpRequest, HttpResponse>>
+    ServiceBuilder decorator(Function<T, R> decorator);
+
```
trustin: nit: This empty line could be removed

core/src/main/java/com/linecorp/armeria/server/ServiceBuilder.java
```diff
@@ -0,0 +1,123 @@
+package com.linecorp.armeria.server;
```
trustin: If you rebase your branch against `master` you'll start to see many checkstyle violations, like missing license headers.

core/src/main/java/com/linecorp/armeria/server/AbstractServiceBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+import com.linecorp.armeria.server.logging.AccessLogWriter;
+
+/**
+ * A builder that implements {@link ServerBuilder} by delegating all calls
+ * to {@link DefaultServiceBuilder}
+ */
+abstract class AbstractServiceBuilder implements ServiceBuilder {
```
trustin: Could you explain why we need this class? It seems to me that we can just extend `DefaultServiceBuilder` and remove this class, no?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1035,6 +1035,15 @@ public ServerBuilder annotatedService(String pathPrefix, Object service,
...
+     */
+    public AnnotatedServiceBindingBuilder annotatedService() {
+        return new AnnotatedServiceBindingBuilder(this);
+    }
+
+
```
trustin: nit: Double empty lines (could be a single line)

core/src/main/java/com/linecorp/armeria/server/DefaultServiceBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    Service<HttpRequest, HttpResponse> decorate(Service<HttpRequest, HttpResponse> service) {
+        if (decorator == null) {
+            return service;
+        }
+        return service.decorate(decorator);
+    }
+
+    ServiceConfigBuilder serviceConfigBuilder(Route route, Service<HttpRequest, HttpResponse> service) {
```
trustin: `toServiceConfigBuilder`?

core/src/main/java/com/linecorp/armeria/server/ServiceBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+import java.time.Duration;
+import java.util.function.Function;
+
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.logging.ContentPreviewer;
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+import com.linecorp.armeria.server.logging.AccessLogWriter;
+
+interface ServiceBuilder {
```
trustin: How about renaming `ServiceBuilder` to `ServiceConfigSetters` and `DefaultServiceBuilder` to `DefaultServiceConfigSetters`? WDYT, @minwoox @ikhoon ?

trustin:
> This is not the problem you introduced, but we used to have from the past. So I think we need to clean that first.

How about just renaming `*ServiceBuilder` in this pull request to move forward? I guess we can clean up by ourselves after this is merged.

core/src/test/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilderTest.java
```diff
@@ -0,0 +1,131 @@
+package com.linecorp.armeria.server;
```
heowc: missing license headers

core/src/test/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilderTest.java
```diff
@@ -0,0 +1,131 @@
...
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.logging.ContentPreviewer;
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+import com.linecorp.armeria.server.annotation.ExceptionHandlerFunction;
+import com.linecorp.armeria.server.annotation.Get;
+import com.linecorp.armeria.server.logging.AccessLogWriter;
+import com.linecorp.armeria.testing.junit.server.ServerExtension;
+
+public class AnnotatedServiceBindingBuilderTest {
```
heowc: It doesn't seem to need to be `public`

core/src/test/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilderTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public static final ExceptionHandlerFunction handlerFunction = (ctx, req, cause) -> HttpResponse.of(501);
```
heowc: Wouldn't `private` be better than `public`?

core/src/main/java/com/linecorp/armeria/server/ServiceBuilder.java
ikhoon: That's an excellent idea! I love that name. :-) This class only used for configuring service.

core/src/main/java/com/linecorp/armeria/server/AbstractServiceBuilder.java
sivaalli: There is no much difference except `AbstractServiceConfigSetters`(which was AbstractServiceBuilder) provides `build0()` that builds `ServiceConfigBuilder` and calls the abstract method `serviceConfigBuilder` for sub-classes to do something with `ServiceConfigBuilder`.

`AnnotatedServiceBindingBuilder`'s `buildAnnotated` will look like this:

```
final ServiceConfigBuilder serviceConfigBuilder = toServiceConfigBuilder(e.route(), s);
serverBuilder.serviceConfigBuilder(serviceConfigBuilder);
```

instead of 

`build0(e.route(), s);`

I can remove this class if this class is accomplishing little/no value.

sivaalli:
@trustin @minwoox @ikhoon @heowc Implemented all the requested changed. Let me know what you think.

core/src/main/java/com/linecorp/armeria/server/AbstractServiceBuilder.java
minwoox: > I can remove this class if this class is accomplishing little/no value.

Then, let's remove this class and use `DefaultServiceConfigSetters`. :-)

core/src/main/java/com/linecorp/armeria/server/AbstractServiceBindingBuilder.java
```diff
@@ -40,175 +34,81 @@
...
-    private AccessLogWriter accessLogWriter;
-    private boolean shutdownAccessLogWriterOnStop;
-
-    /**
-     * Sets the timeout of an HTTP request. If not set, {@link VirtualHost#requestTimeoutMillis()}
-     * is used.
-     *
-     * @param requestTimeout the timeout. {@code 0} disables the timeout.
-     */
+abstract class AbstractServiceBindingBuilder extends AbstractBindingBuilder implements
```
minwoox: It'll be a lot much prettier if we break the line before `implements`.

core/src/main/java/com/linecorp/armeria/server/AbstractServiceBindingBuilder.java
```diff
@@ -218,35 +118,10 @@ final void build0(Service<HttpRequest, HttpResponse> service) {
...
-            }
-            if (responseContentPreviewerFactory != null) {
-                serviceConfigBuilder.responseContentPreviewerFactory(responseContentPreviewerFactory);
-            }
-            if (accessLogWriter != null) {
-                serviceConfigBuilder.accessLogWriter(accessLogWriter, shutdownAccessLogWriterOnStop);
-            }
+            service = defaultServiceConfigSetters.decorate(service);
+            final ServiceConfigBuilder serviceConfigBuilder = defaultServiceConfigSetters
+                                                                    .toServiceConfigBuilder(route, service);
```
minwoox: nit: indentation

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Registers the {@link ServiceConfigBuilder} that is created from this Builder and returns
+     * {@link ServerBuilder}.
+     * @param service annotated service object to handle incoming requests matching path prefix, which
+     *                  can be configured through {@link AnnotatedServiceBindingBuilder#pathPrefix(String)}.
+     *                  If path prefix is not set then this service is registered to handle requests matching
+     *                  {@code /}
+     * @return {@link ServerBuilder} to continue building {@link Server}
+     */
+    public ServerBuilder buildAnnotated(Object service) {
```
minwoox: How about `build()` as we do in `ServiceBindingBuilder`?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1000,6 +1000,13 @@ public ServerBuilder annotatedService(String pathPrefix, Object service,
         return this;
     }
 
+    /**
+     * Returns a {@link AnnotatedServiceBindingBuilder} to build annotated service.
+     */
+    public AnnotatedServiceBindingBuilder annotatedService() {
```
minwoox: Could you add this method to `VirtualHostBuilder` as well?

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+ *   .pathPrefix("/foo")
+ *   .verboseResponses(true)
+ *   .contentPreview(500)
+ *   .buildAnnotated(new Service())                         // Return to the ServerBuilder.
+ *   .build();
+ * }</pre>
+ *
+ * @see ServiceBindingBuilder
+ */
+public final class AnnotatedServiceBindingBuilder extends AbstractServiceConfigSetters {
```
minwoox: Could you make another class for `VirtualHostBuilder`? `VirtualHostAnnotatedServiceBindingBuilder`.

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+    public AnnotatedServiceBindingBuilder accessLogWriter(AccessLogWriter accessLogWriter,
+                                                          boolean shutdownOnStop) {
+        return (AnnotatedServiceBindingBuilder) super.accessLogWriter(accessLogWriter, shutdownOnStop);
+    }
+
+    /**
+     * Registers the {@link ServiceConfigBuilder} that is created from this Builder and returns
+     * {@link ServerBuilder}.
+     * @param service annotated service object to handle incoming requests matching path prefix, which
+     *                  can be configured through {@link AnnotatedServiceBindingBuilder#pathPrefix(String)}.
```
minwoox: nit indentation?
```suggestion
* @param service annotated ...
*                can be ...
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+    public AnnotatedServiceBindingBuilder responseHandler(ResponseConverterFunction responseConverterFunction) {
+        requireNonNull(responseConverterFunction, "responseConverterFunction");
+        responseConverterFunctionBuilder.add(responseConverterFunction);
+        return this;
+    }
+
+    /**
+     * Adds the given {@link RequestConverterFunction} to this {@link AnnotatedServiceBindingBuilder}.
+     */
+    public AnnotatedServiceBindingBuilder requestHandler(RequestConverterFunction requestConverterFunction) {
```
minwoox: We use the word `converter` for the request and response. Could you rename it? `requestConverter`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1000,6 +1000,13 @@ public ServerBuilder annotatedService(String pathPrefix, Object service,
         return this;
     }
 
+    /**
+     * Returns a {@link AnnotatedServiceBindingBuilder} to build annotated service.
```
minwoox: nit: a -> an

core/src/main/java/com/linecorp/armeria/server/AbstractServiceBindingBuilder.java
ikhoon: I think we need a new local variable here unless decorating a service multiple times.
```suggestion
            final decoratedService = defaultServiceConfigSetters.decorate(service);
            final ServiceConfigBuilder serviceConfigBuilder = defaultServiceConfigSetters
                                                                    .toServiceConfigBuilder(route, decoratedService);
```

ikhoon: I think we need a new local variable here unless decorating a service multiple times.
```java
final decoratedService = defaultServiceConfigSetters.decorate(service);
final ServiceConfigBuilder serviceConfigBuilder = defaultServiceConfigSetters                                                                    
    .toServiceConfigBuilder(route, decoratedService);
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+     * Registers the {@link ServiceConfigBuilder} that is created from this Builder and returns
+     * {@link ServerBuilder}.
+     * @param service annotated service object to handle incoming requests matching path prefix, which
+     *                  can be configured through {@link AnnotatedServiceBindingBuilder#pathPrefix(String)}.
+     *                  If path prefix is not set then this service is registered to handle requests matching
+     *                  {@code /}
+     * @return {@link ServerBuilder} to continue building {@link Server}
+     */
+    public ServerBuilder buildAnnotated(Object service) {
+        final ImmutableList<ExceptionHandlerFunction> exceptionHandlerFunctions =
```
ikhoon: Could you change this `ImmutableList` to `List` and the following assignments?
```suggestion
        final List<ExceptionHandlerFunction> exceptionHandlerFunctions =
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+        final ImmutableList<RequestConverterFunction> requestConverterFunctions =
+                requestConverterFunctionBuilder.build();
+        final ImmutableList<ResponseConverterFunction> responseConverterFunctions =
+                responseConverterFunctionBuilder.build();
+
+        final Iterable<?> exceptionHandlersAndConverters = Iterables.concat(exceptionHandlerFunctions,
+                                                                            requestConverterFunctions,
+                                                                            responseConverterFunctions);
+        final List<AnnotatedHttpServiceElement> elements =
+                AnnotatedHttpServiceFactory.find(pathPrefix, service, exceptionHandlersAndConverters);
```
ikhoon: How about adding a new find method because the find method split the merged list(`exceptionHandlersAndConverters`) again. It seems inefficient.

```java
public static List<AnnotatedHttpServiceElement> 
        find(String pathPrefix, Object object,
	         List<RequestConverterFunction> exceptionHandlersAndConverters,
	         List<RequestConverterFunction> requestConverterFunctions,
	         List<ResponseConverterFunction> responseConverterFunctions) {
  ...
}
```

core/src/test/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilderTest.java
```diff
@@ -0,0 +1,147 @@
...
+    static void checkResult(org.apache.http.HttpResponse res,
+                            int statusCode) throws IOException {
+        final HttpStatus status = HttpStatus.valueOf(statusCode);
+        assertThat(res.getStatusLine().toString()).isEqualTo(
+                "HTTP/1.1 " + status);
+    }
+
+    static HttpRequestBase get(String uri) {
+        return new HttpGet(server.httpUri(uri));
+    }
```
ikhoon: Could you use Armeria HTTP client for testing this code? 😊

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
sivaalli: OK. I used this name because it would be clear to distinguish building annotated service and building server. I can make this change.

sivaalli: For ex, it would look like:

```
  sb.annotatedService()                                    // Returns an instance of this class
    .requestTimeout(5000)
    .maxRequestLength(8192)
    .exceptionHandler((ctx, request, cause) -> HttpResponse.of(400))
    .pathPrefix("/foo")
    .verboseResponses(true)
    .contentPreview(500)
    .build(new Service())                         // Return to the ServerBuilder.
    .build();
```

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
sivaalli: Sure will do.

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
minwoox: Yes, I have thought about it which `build(..)` is called twice consecutively.
Currently, we do this for `route()`:
```java
Server.builder()
      .route()
      .get("/")
      .build((ctx, req) -> HttpResponse.of(200))
      .build();
```
If we rename `buildAnnotated`, it will be:
```java
Server.builder()
      .annotatedService()
      .build(new MyObject())
      .build();
```
Which is a little weird. 😄 
But there's another chance that a user uses it with the variable:
```java
final AnnotatedServiceBindingBuilder annotatedBuilder = Server.builder().annotatedService();
annotatedBuilder.build(new MyObject());
```
I'm not sure which one is better honestly. 😄 But I still believe that we need to use same method name. WDYT?

sivaalli: Thanks for the explanation. This totally makes sense. I will make this change along with adding `VirtualHostAnnotatedServiceBindingBuilder` and will notify you again for review.

sivaalli:
This is not complete with all the requested changes. I have made a change and wanted to get feedback.

Please see [this](https://github.com/line/armeria/blob/287fecb06bed4136095a82ac63d3ba66d7aa1126/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceElement.java). When we have an instance of `AnnotatedHttpServiceElement` we build decorate the service and build a `ServiceConfigBuilder` out of it like this. We decorate the `AnnotatedHttpServiceElement.service` by 
1. Decorating with `AnnotatedHttpServiceElement.decorator` 
2. Decorating with custom decorators passed as method param 
3. And finally decorate with `AnnotatedHttpServiceElement.service.exceptionHandlingDecorator`.

Corresponding code is :
```
elements.forEach(e -> {
            Service<HttpRequest, HttpResponse> s = e.service();	
            // Apply decorators which are specified in the service class.	
            s = e.decorator().apply(s);	
            // Apply decorators which are passed via annotatedService() methods.	
            s = decorator.apply(s);	

            // If there is a decorator, we should add one more decorator which handles an exception	
            // raised from decorators.	
            if (s != e.service()) {	
                s = e.service().exceptionHandlingDecorator().apply(s);	
            }	
            service(e.route(), s);	
        });
```

Basically all the decorators are available within `AnnotatedHttpServiceElement` except the decorator passed in through method param. So created a new method in `AnnotatedHttpServiceElement` to do most of the decorating so all the complex logic described above is at once place instead of duplicating at multiple places. Please let me know if this is fine. Or else I can revert this particular logic.

I also think that it would be good to build a `ServiceConfigBuilder` from `AnnotatedHttpServiceElement`, then we basically add the service level settings in `VirtualHostAnnotatedServiceBindingBuilder` and `AnnotatedServiceBindingBuilder`. But `ServiceConfigBuilder` is package-private so it cannot be accessed in `AnnotatedHttpServiceElement` class. WDYT.

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+ *   .pathPrefix("/foo")
+ *   .verboseResponses(true)
+ *   .contentPreview(500)
+ *   .build(new Service())                                   // Return to the ServerBuilder.
+ *   .build();
+ * }</pre>
+ *
+ * @see ServiceBindingBuilder
+ */
+public final class AnnotatedServiceBindingBuilder implements ServiceConfigSetters {
```
sivaalli: Instead of extending `DefaultServiceConfigSetter` I thought of implementing `ServiceConfigSetters ` and delegating calls to `DefaultServiceConfigSetters `. Hope this is fine

sivaalli:
> I'm not sure why we need to do that. Could you elaborate it more, please?

There is no concrete reason on why I think it is better, but I am thinking that we already have  what we need to build a `ServiceConfigBuilder`, so why not do it by exposing a method called `toServiceConfigBuilder(decorator)`. 

By building `ServiceConfigBuilder` we get closer to our goal rather than calling `buildSafeDecoratedService` and again building `ServiceConfigBuilder`, because after calling `buildSafeDecoratedService` all the callers have to build `ServiceConfigBuilder`.

core/src/test/java/com/linecorp/armeria/server/VirtualHostAnnotatedServiceBindingBuilderTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertThat(serviceConfig.accessLogWriter()).isEqualTo(accessLogWriter);
+        assertThat(serviceConfig.shutdownAccessLogWriterOnStop()).isTrue();
+        assertThat(serviceConfig.requestContentPreviewerFactory()).isEqualTo(factory);
+        assertThat(serviceConfig.verboseResponses()).isTrue();
+    }
+
+    @Test
+    @Ignore
+    void testServiceDecoration_shouldCatchException() throws Exception {
+        try (CloseableHttpClient hc = HttpClients.createMinimal()) {
```
sivaalli: I tried using Armeria's http client and it seems that it is not respecting Host header values. IDK if this is intended behaviour, but I feel a bit weird on why that would be an intended behaviour though. So had to use Apache's http client. 

My Armeria's http client code is as follows:

```
final HttpClient httpClient = new HttpClientBuilder(server.httpUri("/"))
                .decorator((delegate, ctx, req) -> {
                    ctx.addAdditionalRequestHeader(HttpHeaderNames.METHOD, "GET");
                    ctx.addAdditionalRequestHeader(HttpHeaderNames.HOST, TEST_HOST);
                    return delegate.execute(ctx, req);
                })
                .build();
        return httpClient.get("/foo/bar").aggregate().join();
```
And I get 404. Let me know if I should create an issue for this.

sivaalli:
@minwoox this is ready for review.

My local build is passing. IDK whats wrong with CI.

core/src/test/java/com/linecorp/armeria/server/VirtualHostAnnotatedServiceBindingBuilderTest.java
anuraaga: To use the Armeria client with a custom host I think you need to initialize it with an `Endpoint`

`Endpoint.of(host, port).withIpAddr(127.0.0.1);`

trustin: I think you need to use .AUTHORITY rather than .HOST. (because it's HTTP/2)

trustin: .. and therr's no beed to set .GET.

sivaalli: Thank you @anuraaga @trustin for the advice 🙏 . I was able to use Armeria http client.

trustin:
@sivaalli It seems like a compilation error:

```
.../core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java:257: error: incompatible types: inference variable E has incompatible bounds
                      .collect(toImmutableList());
                              ^
    equality constraints: AnnotatedHttpServiceElement
    lower bounds: List<AnnotatedHttpServiceElement>
  where E is a type-variable:
    E extends Object declared in method <E>toImmutableList()
```

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -216,6 +216,25 @@
...
+     * from the specified {@code object}, {@link ExceptionHandlerFunction}'s,
+     * {@link RequestConverterFunction}'s and {@link ResponseConverterFunction}'s.
+     */
+    public static List<AnnotatedHttpServiceElement> find(String pathPrefix, Object object,
+                                                         List<ExceptionHandlerFunction>
+                                                                 exceptionHandlerFunctions,
+                                                         List<RequestConverterFunction>
+                                                                 requestConverterFunctions,
+                                                         List<ResponseConverterFunction>
+                                                                 responseConverterFunctions) {
```
trustin: nit: I think you could break the line right after `find(` to make this look better.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceConfigSetters.java
```diff
@@ -0,0 +1,183 @@
...
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+import com.linecorp.armeria.internal.ArmeriaHttpUtil;
+import com.linecorp.armeria.server.logging.AccessLogWriter;
+
+/**
+ * A default implementation of {@link ServiceConfigSetters} that stores service related settings
+ * and provides a method {@link DefaultServiceConfigSetters#toServiceConfigBuilder(Route, Service)} to build
+ * {@link ServiceConfigBuilder}.
+ */
+class DefaultServiceConfigSetters implements ServiceConfigSetters {
```
trustin: Could be `final`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -983,23 +983,60 @@ public ServerBuilder annotatedService(String pathPrefix, Object service,
...
+        registerHttpServiceElement(elements, decorator);
+        return this;
+    }
+
+    /**
+     * Binds the specified annotated service object under the specified path prefix.
+     *
+     * @param exceptionHandlerFunctions a list of {@link ExceptionHandlerFunction}'s
+     * @param requestConverterFunctions a list of {@link RequestConverterFunction}'s
+     * @param responseConverterFunctions a list of {@link ResponseConverterFunction}'s
```
trustin: nit: `'`s could be removed?

core/src/main/java/com/linecorp/armeria/server/VirtualHostAnnotatedServiceBindingBuilder.java
```diff
@@ -0,0 +1,218 @@
...
+public final class VirtualHostAnnotatedServiceBindingBuilder implements ServiceConfigSetters {
+
+    private final DefaultServiceConfigSetters defaultServiceConfigSetters = new DefaultServiceConfigSetters();
+    private final VirtualHostBuilder virtualHostBuilder;
+    private final Builder<ExceptionHandlerFunction> exceptionHandlerFunctionBuilder = ImmutableList.builder();
+    private final Builder<RequestConverterFunction> requestConverterFunctionBuilder = ImmutableList.builder();
+    private final Builder<ResponseConverterFunction> responseConverterFunctionBuilder = ImmutableList.builder();
+    private String pathPrefix = "/";
+
+    public VirtualHostAnnotatedServiceBindingBuilder(VirtualHostBuilder virtualHostBuilder) {
```
trustin: Needs to be package-private

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -550,23 +550,59 @@ public VirtualHostBuilder annotatedService(
...
+
+        final List<AnnotatedHttpServiceElement> elements =
+                AnnotatedHttpServiceFactory.find(pathPrefix, service, exceptionHandlerFunctions,
+                                                 requestConverterFunctions,
+                                                 responseConverterFunctions);
+        registerHttpServiceElement(elements, decorator);
+        return this;
+    }
+
+    public VirtualHostAnnotatedServiceBindingBuilder annotatedService() {
```
trustin: Missing Javadoc

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -550,23 +550,59 @@ public VirtualHostBuilder annotatedService(
...
+        registerHttpServiceElement(elements, decorator);
         return this;
     }
 
+    /**
+     * Binds the specified annotated service object under the specified path prefix.
+     *
+     * @param exceptionHandlerFunctions a list of {@link ExceptionHandlerFunction}'s
+     * @param requestConverterFunctions a list of {@link RequestConverterFunction}'s
+     * @param responseConverterFunctions a list of {@link ResponseConverterFunction}'s
```
trustin: nit: `'` could be removed?

sivaalli:
> Regarding the compilation error, what version of JDK are you using?

@trustin I'm using JDK 12. I cannot detect this error locally. Because I was able to compile and build a jar and ran some manual tests.

trustin:
I tried the same JDK version with AppVeyor builds and it works fine, too. Please ignore the compilation error meanwhile. Could you also try adding `clean` to the  build targets in `.appveyor.build.sh`?

sivaalli:
Sorry, I have again messed with the commits. I will create a new PR fixing things.

PR #2183, opened 2019-10-11
Fix a bug where disables all check style

minwoox:


minwoox:
Found a few more violations, let me fix them.

Issue #1645, opened 2019-03-08
Startup time optimization


trustin:
Initial startup time of a minimal Armeria server is about 1.x seconds in my setup, which is longer than a usual expectation. What's interesting is that the second `Server` instance in the same JVM takes less than 100ms to start, which may mean some one-time initialization logic is taking longer time than expected.

minwoox:
I think this is solved by #2184

trustin:
Thanks, @minwoox

PR #2184, opened 2019-10-12
Set useOpenSsl in Flags lazily to reduce server startup time

mumgmangmange:
Motivation:
Related #1645
`useOpenSsl` in `Flags` is used only when TLS is enabled. However, it takes a lot to instantiate `OpenSsl` class so it adds up a tremendous time to the server startup time. We should fix to instantiate it only when it needs.

Modification:
- Set `useOpenSsl` in `Flags` lazily
- Check whether the OS is Linux before calling `Epoll.isAvailable()`
- Add `OsType`

Result:
- Reduced server startup time to 80 percent (20% decreased)

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -118,6 +118,7 @@ public final class Flags {
     private static final boolean HAS_WSLENV = System.getenv("WSLENV") != null;
     private static final boolean USE_EPOLL = getBoolean("useEpoll", isEpollAvailable(),
                                                         value -> isEpollAvailable() || !value);
+    @Nullable
     private static Boolean useOpenSsl;
 
     private static final boolean DUMP_OPENSSL_INFO = getBoolean("dumpOpenSslInfo", false);
@@ -406,18 +407,16 @@ public final class Flags {
      * {@code -Dcom.linecorp.armeria.useOpenSsl=false} JVM option to disable it.
      */
     public static boolean useOpenSsl() {
-        if (useOpenSsl != null) {
-            return useOpenSsl;
+        final boolean useOpenSsl = getBoolean("useOpenSsl", true);
+        if (!useOpenSsl) {
+            // OpenSSL explicitly disabled
+            return Flags.useOpenSsl = false;
         }
-        useOpenSsl = getBoolean("useOpenSsl", OpenSsl.isAvailable(), value -> OpenSsl.isAvailable() || !value);
         if (!OpenSsl.isAvailable()) {
             final Throwable cause = Exceptions.peel(OpenSsl.unavailabilityCause());
             logger.info("OpenSSL not available: {}", cause.toString());
-        } else if (useOpenSsl) {
-            logger.info("Using OpenSSL: {}, 0x{}",
-                    OpenSsl.versionString(),
-                    Long.toHexString(OpenSsl.version() & 0xFFFFFFFFL));
-
+        } else {
+            logger.info("Using OpenSSL: {}, 0x{}", OpenSsl.versionString(), Long.toHexString(OpenSsl.version() & 0xFFFFFFFFL));
             if (dumpOpenSslInfo()) {
                 final SSLEngine engine = SslContextUtil.createSslContext(
                         SslContextBuilder::forClient,
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -29,6 +29,7 @@ import javax.annotation.Nullable;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
 
+import com.linecorp.armeria.common.util.SystemInfo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -117,9 +118,7 @@ public final class Flags {
     private static final boolean HAS_WSLENV = System.getenv("WSLENV") != null;
     private static final boolean USE_EPOLL = getBoolean("useEpoll", isEpollAvailable(),
                                                         value -> isEpollAvailable() || !value);
-
-    private static final boolean USE_OPENSSL = getBoolean("useOpenSsl", OpenSsl.isAvailable(),
-                                                          value -> OpenSsl.isAvailable() || !value);
+    private static Boolean useOpenSsl;
 
     private static final boolean DUMP_OPENSSL_INFO = getBoolean("dumpOpenSslInfo", false);
 
@@ -312,34 +311,15 @@ public final class Flags {
         } else if (USE_EPOLL) {
             logger.info("Using /dev/epoll");
         }
-
-        if (!OpenSsl.isAvailable()) {
-            final Throwable cause = Exceptions.peel(OpenSsl.unavailabilityCause());
-            logger.info("OpenSSL not available: {}", cause.toString());
-        } else if (USE_OPENSSL) {
-            logger.info("Using OpenSSL: {}, 0x{}",
-                        OpenSsl.versionString(),
-                        Long.toHexString(OpenSsl.version() & 0xFFFFFFFFL));
-
-            if (dumpOpenSslInfo()) {
-                final SSLEngine engine = SslContextUtil.createSslContext(
-                        SslContextBuilder::forClient,
-                        false,
-                        unused -> {}).newEngine(ByteBufAllocator.DEFAULT);
-                logger.info("All available SSL protocols: {}",
-                            ImmutableList.copyOf(engine.getSupportedProtocols()));
-                logger.info("Default enabled SSL protocols: {}", SslContextUtil.DEFAULT_PROTOCOLS);
-                ReferenceCountUtil.release(engine);
-                logger.info("All available SSL ciphers: {}", OpenSsl.availableJavaCipherSuites());
-                logger.info("Default enabled SSL ciphers: {}", SslContextUtil.DEFAULT_CIPHERS);
-            }
-        }
     }
 
     private static boolean isEpollAvailable() {
-        // Netty epoll transport does not work with WSL (Windows Sybsystem for Linux) yet.
-        // TODO(trustin): Re-enable on WSL if https://github.com/Microsoft/WSL/issues/1982 is resolved.
-        return Epoll.isAvailable() && !HAS_WSLENV;
+        if (SystemInfo.isLinux()) {
+            // Netty epoll transport does not work with WSL (Windows Sybsystem for Linux) yet.
+            // TODO(trustin): Re-enable on WSL if https://github.com/Microsoft/WSL/issues/1982 is resolved.
+            return Epoll.isAvailable() && !HAS_WSLENV;
+        }
+        return false;
     }
 
     /**
@@ -426,7 +406,32 @@ public final class Flags {
      * {@code -Dcom.linecorp.armeria.useOpenSsl=false} JVM option to disable it.
      */
     public static boolean useOpenSsl() {
-        return USE_OPENSSL;
+        if (useOpenSsl != null) {
+            return useOpenSsl;
+        }
+        useOpenSsl = getBoolean("useOpenSsl", OpenSsl.isAvailable(), value -> OpenSsl.isAvailable() || !value);
+        if (!OpenSsl.isAvailable()) {
+            final Throwable cause = Exceptions.peel(OpenSsl.unavailabilityCause());
+            logger.info("OpenSSL not available: {}", cause.toString());
+        } else if (useOpenSsl) {
+            logger.info("Using OpenSSL: {}, 0x{}",
+                    OpenSsl.versionString(),
+                    Long.toHexString(OpenSsl.version() & 0xFFFFFFFFL));
+
+            if (dumpOpenSslInfo()) {
+                final SSLEngine engine = SslContextUtil.createSslContext(
+                        SslContextBuilder::forClient,
+                        false,
+                        unused -> {}).newEngine(ByteBufAllocator.DEFAULT);
+                logger.info("AllUSE_OPENSSL = null; available SSL protocols: {}",
+                        ImmutableList.copyOf(engine.getSupportedProtocols()));
+                logger.info("Default enabled SSL protocols: {}", SslContextUtil.DEFAULT_PROTOCOLS);
+                ReferenceCountUtil.release(engine);
+                logger.info("All available SSL ciphers: {}", OpenSsl.availableJavaCipherSuites());
+                logger.info("Default enabled SSL ciphers: {}", SslContextUtil.DEFAULT_CIPHERS);
+            }
+        }
+        return useOpenSsl;
     }
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -407,8 +407,8 @@ public final class Flags {
      * {@code -Dcom.linecorp.armeria.useOpenSsl=false} JVM option to disable it.
      */
     public static boolean useOpenSsl() {
-        if (Flags.useOpenSsl != null) {
-            return Flags.useOpenSsl;
+        if (useOpenSsl != null) {
+            return useOpenSsl;
         }
         final boolean useOpenSsl = getBoolean("useOpenSsl", true);
         if (!useOpenSsl) {
@@ -420,14 +420,15 @@ public final class Flags {
             logger.info("OpenSSL not available: {}", cause.toString());
             return Flags.useOpenSsl = false;
         }
-        logger.info("Using OpenSSL: {}, 0x{}", OpenSsl.versionString(), Long.toHexString(OpenSsl.version() & 0xFFFFFFFFL));
+        logger.info("Using OpenSSL: {}, 0x{}", OpenSsl.versionString(),
+                    Long.toHexString(OpenSsl.version() & 0xFFFFFFFFL));
         if (dumpOpenSslInfo()) {
             final SSLEngine engine = SslContextUtil.createSslContext(
                     SslContextBuilder::forClient,
                     false,
                     unused -> {}).newEngine(ByteBufAllocator.DEFAULT);
             logger.info("AllUSE_OPENSSL = null; available SSL protocols: {}",
-                    ImmutableList.copyOf(engine.getSupportedProtocols()));
+                        ImmutableList.copyOf(engine.getSupportedProtocols()));
             logger.info("Default enabled SSL protocols: {}", SslContextUtil.DEFAULT_PROTOCOLS);
             ReferenceCountUtil.release(engine);
             logger.info("All available SSL ciphers: {}", OpenSsl.availableJavaCipherSuites());
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -29,7 +29,6 @@ import javax.annotation.Nullable;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
 
-import com.linecorp.armeria.common.util.SystemInfo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -46,6 +45,7 @@ import com.linecorp.armeria.client.retry.RetryingHttpClient;
 import com.linecorp.armeria.client.retry.RetryingRpcClient;
 import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.common.util.Sampler;
+import com.linecorp.armeria.common.util.SystemInfo;
 import com.linecorp.armeria.internal.SslContextUtil;
 import com.linecorp.armeria.server.RoutingContext;
 import com.linecorp.armeria.server.ServerBuilder;
@@ -427,7 +427,7 @@ public final class Flags {
                     SslContextBuilder::forClient,
                     false,
                     unused -> {}).newEngine(ByteBufAllocator.DEFAULT);
-            logger.info("AllUSE_OPENSSL = null; available SSL protocols: {}",
+            logger.info("All available SSL protocols: {}",
                         ImmutableList.copyOf(engine.getSupportedProtocols()));
             logger.info("Default enabled SSL protocols: {}", SslContextUtil.DEFAULT_PROTOCOLS);
             ReferenceCountUtil.release(engine);
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -407,6 +407,9 @@ public final class Flags {
      * {@code -Dcom.linecorp.armeria.useOpenSsl=false} JVM option to disable it.
      */
     public static boolean useOpenSsl() {
+        if (Flags.useOpenSsl != null) {
+            return Flags.useOpenSsl;
+        }
         final boolean useOpenSsl = getBoolean("useOpenSsl", true);
         if (!useOpenSsl) {
             // OpenSSL explicitly disabled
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -117,9 +117,7 @@
     private static final boolean HAS_WSLENV = System.getenv("WSLENV") != null;
     private static final boolean USE_EPOLL = getBoolean("useEpoll", isEpollAvailable(),
```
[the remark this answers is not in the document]
minwoox: Didn't notice that we added `if` condition in `isEpollAvailable`. So we can leave this as it is. Thanks!

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -117,9 +117,7 @@
...
                                                         value -> isEpollAvailable() || !value);
-
-    private static final boolean USE_OPENSSL = getBoolean("useOpenSsl", OpenSsl.isAvailable(),
-                                                          value -> OpenSsl.isAvailable() || !value);
+    private static Boolean USE_OPENSSL = null;
```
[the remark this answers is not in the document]
minwoox: Because this is not a static final constant anymore, we should use camel case.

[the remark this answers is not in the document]
minwoox: Could you leave a comment about why we initialize this field lazily?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,6 +403,32 @@ public static boolean useEpoll() {
      * {@code -Dcom.linecorp.armeria.useOpenSsl=false} JVM option to disable it.
      */
     public static boolean useOpenSsl() {
+
```
ikhoon: Remove this blank line?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -308,34 +306,17 @@
...
-            }
-        }
     }
 
     private static boolean isEpollAvailable() {
         // Netty epoll transport does not work with WSL (Windows Sybsystem for Linux) yet.
         // TODO(trustin): Re-enable on WSL if https://github.com/Microsoft/WSL/issues/1982 is resolved.
-        return Epoll.isAvailable() && !HAS_WSLENV;
+
+        if (Ascii.toLowerCase(System.getProperty("os.name")).startsWith("linux")) {
```
ikhoon: Could you use `SystemInfo.isLinux()`?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,6 +403,32 @@ public static boolean useEpoll() {
      * {@code -Dcom.linecorp.armeria.useOpenSsl=false} JVM option to disable it.
      */
     public static boolean useOpenSsl() {
+
+        if (USE_OPENSSL == null) {
```
ikhoon: We prefer ['early-return' style code](https://github.com/line/armeria/blob/master/CONTRIBUTING.md#prefer-early-return-style). Could you change the code as the the following code style?
```java
if(useOpenSsl != null) {
	return useOpenSsl;
}

useOpenSsl = getBoolean("useOpenSsl", OpenSsl.isAvailable(), value -> OpenSsl.isAvailable() || !value);
...
```

trustin:
There are 3 Checkstyle violations:

```
[ant:checkstyle] [ERROR] /appveyor/projects/armeria/core/src/main/java/com/linecorp/armeria/common/Flags.java:120:28: Variable 'USE_OPENSSL' explicitly initialized to 'null' (default value for its type). [ExplicitInitialization]
[ant:checkstyle] [ERROR] /appveyor/projects/armeria/core/src/main/java/com/linecorp/armeria/common/Flags.java:412: Line is longer than 112 characters (found 116). [LineLength]
[ant:checkstyle] [ERROR] /appveyor/projects/armeria/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java:84:8: Unused import - io.netty.handler.ssl.OpenSsl. [UnusedImports]
```

Just in case it's unclear, the first violation means you do not need to assign `null` to a variable, because a variable's default value is `null` already. The other two look obvious, right?

core/src/main/java/com/linecorp/armeria/common/Flags.java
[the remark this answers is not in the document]
mumgmangmange: Thank you :)

core/src/main/java/com/linecorp/armeria/common/Flags.java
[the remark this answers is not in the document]
mumgmangmange: Changed to Camel. Thank you.

Do not call OpenSsl.isAvailable() when not using SSL.
0.2 to 0.3 seconds shortened.

core/src/main/java/com/linecorp/armeria/common/Flags.java
mumgmangmange: Ok. I removed this line. 😃

core/src/main/java/com/linecorp/armeria/common/Flags.java
mumgmangmange: OK. I applied it. 😃 
And to call the isLinux() of SystemInfo.java I Changed Access Modifier from private to public.

core/src/main/java/com/linecorp/armeria/common/Flags.java
mumgmangmange: OK. I applied it. Thank you 😄

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -308,34 +306,17 @@
...
-                ReferenceCountUtil.release(engine);
-                logger.info("All available SSL ciphers: {}", OpenSsl.availableJavaCipherSuites());
-                logger.info("Default enabled SSL ciphers: {}", SslContextUtil.DEFAULT_CIPHERS);
-            }
-        }
     }
 
     private static boolean isEpollAvailable() {
         // Netty epoll transport does not work with WSL (Windows Sybsystem for Linux) yet.
         // TODO(trustin): Re-enable on WSL if https://github.com/Microsoft/WSL/issues/1982 is resolved.
```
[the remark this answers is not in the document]
mumgmangmange: I changed. Thank you. 😄

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -832,13 +837,13 @@ public static ExceptionVerbosity annotatedServiceExceptionVerbosity() {
             }
         });
         return "off".equals(spec) ? Optional.empty()
-                                  : Optional.of(spec);
+                : Optional.of(spec);
```
ikhoon: revert?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -832,13 +837,13 @@ public static ExceptionVerbosity annotatedServiceExceptionVerbosity() {
...
     }
 
     private static ExceptionVerbosity exceptionLoggingMode(String name, String defaultValue) {
         final String mode = getNormalized(name, defaultValue,
-                                          value -> Arrays.stream(ExceptionVerbosity.values())
-                                                         .anyMatch(v -> v.name().equalsIgnoreCase(value)));
+                value -> Arrays.stream(ExceptionVerbosity.values())
+                        .anyMatch(v -> v.name().equalsIgnoreCase(value)));
```
ikhoon: Revert?

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -154,7 +154,7 @@ public static long currentTimeMicros() {
         return JavaVersionSpecific.get().currentTimeMicros();
     }
 
-    private static boolean isLinux() {
+    public static boolean isLinux() {
```
ikhoon: Please add javadoc for `public` method.

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -154,7 +154,10 @@ public static long currentTimeMicros() {
         return JavaVersionSpecific.get().currentTimeMicros();
     }
 
-    private static boolean isLinux() {
+    /**
+     * Return whether Linux is used.
+     */
+    public static boolean isLinux() {
         return Ascii.toLowerCase(System.getProperty("os.name", "")).startsWith("linux");
```
anuraaga: Now that this is a public method, we should go ahead and optimize the method. Can you add `private static final boolean IS_LINUX = Ascii.toLowerCase(System.getProperty("os.name", "")).startsWith("linux");` at the beginning and return that here?

Also, anyone have thoughts on making this public means we should instead return an enum of `LINUX`, `WINDOWS`, `MAC_OS_X`, `OTHER`?

trustin: Yes, how about replacing `isLinux()` with the following method?

```java
private static final OsType OS_TYPE;

static {
    // ... Detect OS type here ...
    OS_TYPE = ...;
}

public static OsType osType() {
    return OS_TYPE;
}
```

`OsType` could look like:

```java
public enum OsType {
    LINUX,
    WINDOWS,
    // MAC_OS_X, FREEBSD, ... (Do we need them?)
    OTHERS
}
```

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
mumgmangmange: I wrote an annotation. 😃 
But is this correct ❔

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
mumgmangmange: OK 😄 I understand. 
I modified with reference to comment.

core/src/main/java/com/linecorp/armeria/common/Flags.java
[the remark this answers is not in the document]
adriancole: could this change be in a patch release with boot 2.2? If it is that much overhead, this will reduce zipkin-slim startup by 15% (I can verify). We are nearing a redo of our release, and it would make the announcement much better I think.

cc @trustin @anuraaga

adriancole:
fyi this inspired me to notice and fix some things in brave. thanks! https://github.com/openzipkin/brave/pull/1006

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,7 +402,32 @@ public static boolean useEpoll() {
      * {@code -Dcom.linecorp.armeria.useOpenSsl=false} JVM option to disable it.
      */
     public static boolean useOpenSsl() {
-        return USE_OPENSSL;
+        if (useOpenSsl != null) {
+            return useOpenSsl;
+        }
+        useOpenSsl = getBoolean("useOpenSsl", OpenSsl.isAvailable(), value -> OpenSsl.isAvailable() || !value);
```
anuraaga: I think we can avoid `isAvailable` if the user explicitly disabled OpenSSL.

```java
final boolean useOpenSsl = getBoolean("useOpenSsl", true);
if (!useOpenSsl) {
  // OpenSSL explicitly disabled
  return Flags.useOpenSsl = false;
}
if (!OpenSsl.isAvailable(){ 
  ...
} else {
  ...
}
```

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,12 @@
                 JETTY_ALPN_OPTIONAL_OR_AVAILABLE = false;
             }
         }
+
+        try {
+            osType = OsType.valueOf(Ascii.toUpperCase(System.getProperty("os.name", "").split(" ")[0]));
```
anuraaga: OS detection is a bit more complicated unfortunately.

Can you use the if statements for the `OsType` OS's from this famous OS detection logic?

https://github.com/trustin/os-maven-plugin/blob/master/src/main/java/kr/motd/maven/os/Detector.java#L150

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -154,8 +162,20 @@ public static long currentTimeMicros() {
...
+     */
+    public static OsType osType() {
+        return osType;
+    }
+
+    public enum OsType {
+        WINDOWS,
+        LINUX,
+        MAC,
+        FREEBSD,
```
anuraaga: I would lean towards leaving out `FREEBSD` and `SOLARIS` - it's unlikely Armeria servers would have logic specific to them anyways so they can go in `OTHERS`

mumgmangmange: OK. 😄 
I thought FREEBSD and SOLARIS as OTHERS and changed it.

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
mumgmangmange: OK. 😃 Changed to an if statement.

core/src/main/java/com/linecorp/armeria/common/Flags.java
mumgmangmange: Is this useOpenSsl variable unconditionally true?

anuraaga: In this case, `true` is the default value unless a user sets `com.linecorp.armeria.useOpenSsl=false`.

At least I think it is let me know if I missed something :)

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,21 @@
...
+            String osName = Ascii.toUpperCase(System.getProperty("os.name", ""));
+            if (osName.startsWith("WINDOWS")) {
+                osType = osType().WINDOWS;
+            }
+            if (osName.startsWith("LINUX")) {
+                osType = osType().LINUX;
+            }
+            if (osName.startsWith("MAC")) {
+                osType = osType().MAC;
+            }
```
minwoox: We should add `osType = OsType.OTHERS;` here

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,21 @@
                 JETTY_ALPN_OPTIONAL_OR_AVAILABLE = false;
             }
         }
+
+        try {
+            String osName = Ascii.toUpperCase(System.getProperty("os.name", ""));
+            if (osName.startsWith("WINDOWS")) {
+                osType = osType().WINDOWS;
```
minwoox: `osType = OsType.WINDOWS`

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,21 @@
...
+            }
+            if (osName.startsWith("LINUX")) {
+                osType = osType().LINUX;
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,21 @@
...
+            }
+            if (osName.startsWith("MAC")) {
+                osType = osType().MAC;
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,21 @@
...
+            }
+        } catch (Throwable t) {
+            osType = osType().OTHERS;
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,21 @@
                 JETTY_ALPN_OPTIONAL_OR_AVAILABLE = false;
             }
         }
+
+        try {
+            String osName = Ascii.toUpperCase(System.getProperty("os.name", ""));
```
minwoox: nit: final

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,7 +402,32 @@ public static boolean useEpoll() {
...
+        if (!OpenSsl.isAvailable()) {
+            final Throwable cause = Exceptions.peel(OpenSsl.unavailabilityCause());
+            logger.info("OpenSSL not available: {}", cause.toString());
+        } else if (useOpenSsl) {
+            logger.info("Using OpenSSL: {}, 0x{}",
+                    OpenSsl.versionString(),
```
minwoox: nit: indentation

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -117,9 +118,7 @@
     private static final boolean HAS_WSLENV = System.getenv("WSLENV") != null;
     private static final boolean USE_EPOLL = getBoolean("useEpoll", isEpollAvailable(),
                                                         value -> isEpollAvailable() || !value);
-
-    private static final boolean USE_OPENSSL = getBoolean("useOpenSsl", OpenSsl.isAvailable(),
-                                                          value -> OpenSsl.isAvailable() || !value);
+    private static Boolean useOpenSsl;
```
minwoox: nit: `@Nullable`

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -308,34 +307,15 @@
...
-                logger.info("Default enabled SSL ciphers: {}", SslContextUtil.DEFAULT_CIPHERS);
-            }
-        }
     }
 
     private static boolean isEpollAvailable() {
-        // Netty epoll transport does not work with WSL (Windows Sybsystem for Linux) yet.
-        // TODO(trustin): Re-enable on WSL if https://github.com/Microsoft/WSL/issues/1982 is resolved.
-        return Epoll.isAvailable() && !HAS_WSLENV;
+        if (SystemInfo.osType() == SystemInfo.OsType.LINUX) {
```
minwoox: `if (SystemInfo.isLinux())`?

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -154,8 +171,18 @@ public static long currentTimeMicros() {
         return JavaVersionSpecific.get().currentTimeMicros();
     }
 
-    private static boolean isLinux() {
-        return Ascii.toLowerCase(System.getProperty("os.name", "")).startsWith("linux");
```
minwoox: Let's bring back this method.
```java
public static boolean isLinux() {
    return osType == OsType.LINUX;
```

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -51,6 +51,8 @@
 
     private static boolean JETTY_ALPN_OPTIONAL_OR_AVAILABLE;
 
+    private static OsType osType;
```
ikhoon: `final`?

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,21 @@
                 JETTY_ALPN_OPTIONAL_OR_AVAILABLE = false;
             }
         }
+
+        try {
+            String osName = Ascii.toUpperCase(System.getProperty("os.name", ""));
+            if (osName.startsWith("WINDOWS")) {
+                osType = osType().WINDOWS;
+            }
+            if (osName.startsWith("LINUX")) {
```
ikhoon: I think `else if` is better here
```suggestion
            else if (osName.startsWith("LINUX")) {
```

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -109,6 +111,21 @@
...
+                osType = osType().LINUX;
+            }
+            if (osName.startsWith("MAC")) {
```
ikhoon: ditto

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -154,8 +171,18 @@ public static long currentTimeMicros() {
...
-    private static boolean isLinux() {
-        return Ascii.toLowerCase(System.getProperty("os.name", "")).startsWith("linux");
+    /**
+     * Returns the operating system type.
+     */
+    public static OsType osType() {
+        return osType;
+    }
+
+    public enum OsType {
```
ikhoon: Add Javadoc for `public`. :-)

core/src/main/java/com/linecorp/armeria/common/Flags.java
mumgmangmange: Yes. I understand 😃

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
mumgmangmange: Yes. I removed the try catch statement and used the if else clause :)

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
mumgmangmange: ahtt...I changed it

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
mumgmangmange: I added again :)

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -154,8 +167,28 @@ public static long currentTimeMicros() {
...
+     */
+    public enum OsType {
+        WINDOWS,
+        LINUX,
+        MAC,
+        OTHERS
+    }
+
+    /**
+     * Return whether Linux is used.
```
ikhoon: nit: Returns `{@code true}` if the operating system is Linux.

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,7 +403,30 @@ public static boolean useEpoll() {
      * {@code -Dcom.linecorp.armeria.useOpenSsl=false} JVM option to disable it.
      */
     public static boolean useOpenSsl() {
-        return USE_OPENSSL;
```
anuraaga: I think we still need the null check that returns the value after we set it once like we had before (can use a different name for the local variable if it makes it confusing).

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,7 +403,30 @@ public static boolean useEpoll() {
...
+        final boolean useOpenSsl = getBoolean("useOpenSsl", true);
+        if (!useOpenSsl) {
+            // OpenSSL explicitly disabled
+            return Flags.useOpenSsl = false;
+        }
+        if (!OpenSsl.isAvailable()) {
+            final Throwable cause = Exceptions.peel(OpenSsl.unavailabilityCause());
+            logger.info("OpenSSL not available: {}", cause.toString());
```
anuraaga: `return Flags.useOpenSsl = false`

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,7 +403,30 @@ public static boolean useEpoll() {
...
+        } else {
```
anuraaga: Can remove this else after adding the return above

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,7 +403,30 @@ public static boolean useEpoll() {
...
+                        unused -> {}).newEngine(ByteBufAllocator.DEFAULT);
+                logger.info("AllUSE_OPENSSL = null; available SSL protocols: {}",
+                        ImmutableList.copyOf(engine.getSupportedProtocols()));
+                logger.info("Default enabled SSL protocols: {}", SslContextUtil.DEFAULT_PROTOCOLS);
+                ReferenceCountUtil.release(engine);
+                logger.info("All available SSL ciphers: {}", OpenSsl.availableJavaCipherSuites());
+                logger.info("Default enabled SSL ciphers: {}", SslContextUtil.DEFAULT_CIPHERS);
+            }
+        }
+        return useOpenSsl;
```
anuraaga: `return Flags.useOpenSsl = true`

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -422,7 +403,34 @@ public static boolean useEpoll() {
...
+            return Flags.useOpenSsl = false;
+        }
+        logger.info("Using OpenSSL: {}, 0x{}", OpenSsl.versionString(),
+                    Long.toHexString(OpenSsl.version() & 0xFFFFFFFFL));
+        if (dumpOpenSslInfo()) {
+            final SSLEngine engine = SslContextUtil.createSslContext(
+                    SslContextBuilder::forClient,
+                    false,
+                    unused -> {}).newEngine(ByteBufAllocator.DEFAULT);
+            logger.info("AllUSE_OPENSSL = null; available SSL protocols: {}",
```
anuraaga:
```suggestion
            logger.info("All available SSL protocols: {}",
```

Looks like an accidental typo here

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -154,8 +167,28 @@ public static long currentTimeMicros() {
...
-        return Ascii.toLowerCase(System.getProperty("os.name", "")).startsWith("linux");
+    /**
+     * Returns the operating system type.
+     */
+    public static OsType osType() {
+        return osType;
+    }
+
+    /**
+     * Operating system type enumeration value
```
ikhoon: nit: Add '.' to the end of the line?

anuraaga:
```suggestion
     * Operating system.
```

(type enumeration value is obvious from it being `enum` so no need to repeat in javadoc)

core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java
```diff
@@ -154,8 +167,28 @@ public static long currentTimeMicros() {
         return JavaVersionSpecific.get().currentTimeMicros();
     }
 
-    private static boolean isLinux() {
-        return Ascii.toLowerCase(System.getProperty("os.name", "")).startsWith("linux");
+    /**
+     * Returns the operating system type.
```
anuraaga:
```suggestion
     * Returns the operating system for the currently running process.
```

minwoox:
@anuraaga, any other comments?

minwoox:
@mumgmangmange Could you sign the CLA so that we can merge this PR?

minwoox:
congratulations on your first PR, @mumgmangmange!
Thank you~! and Thanks, reviewers!

PR #2187, opened 2019-10-14
Some annotated service methods are run from blocking task executors

### LINKED ISSUE #2078: Some annotated service methods are run from blocking task executors.
Related: #2064 

For example, when the return type of a service method is `AggregatedHttpResponse` or response object conversion is involved, the service method is currently executed via the blocking task executor. By contrast, if a service method returns `HttpResponse`, it will be executed from a event loop thread. This behavioral difference is too subtle and confusing.

We need to:

- Make the service methods run from an event loop thread by default, regardless of their return types.
- Introduce an annotation that makes the service methods run from a blocking task executor.
- Provide some migration plan, like:
  - Introduce a flag that enables this change, whose default is false (disabled).
  - When the flag is disabled, issue a warning log message to tell people to migrate.
  - After a few releases, change the default from false to true.

heowc:
Motivation:
The annotated services whose return type is neither `HttpResponse` nor `CompeltableFuture`, are run using a `blockingTaskExecutor` by default.
We should fix this to use `EventLoop` by default and let a user choose to use `blockingTaskExecutor` if he/she wants.

Modifications:
- Add `@Blocking` which makes the annotated service run using `blockingTaskExecutor`
- Make all annotated services run from `EventLoop`

Result:
- Fix #2078
- (Breaking) All annotated services are now run from `EventLoop` by default


core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpService.java
```diff
@@ -232,19 +235,48 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
...
+                                            cause == null ? convertResponse(ctx, req, null, result,
+                                                                            HttpHeaders.of())
+                                                          : exceptionHandler.handleException(ctx, req, cause));
+                }
+
             default:
-                return f.thenApplyAsync(msg -> convertResponse(ctx, req, null, invoke(ctx, req, msg),
-                                                               HttpHeaders.of()),
-                                        ctx.blockingTaskExecutor());
+                if (useBlocking) {
```
minwoox: If a user uses `@Blocking`, he/she always sees this warning even though he/she uses it correctly. How about adding a flag for that in `Flags`?
```java
if (useBlocking) {
    return f.thenApplyAsync(
            msg -> convertResponse(ctx, req, null, invoke(ctx, req, msg),
                                   HttpHeaders.of()),
            ctx.blockingTaskExecutor());
} else {
    if (!USE_EVENT_LOOP_FOR_ANNOTATED_SERVICE) { // default is false
        logger.warn(...);
        return f.thenApplyAsync(
                msg -> convertResponse(ctx, req, null, invoke(ctx, req, msg),
                                       HttpHeaders.of()),
                ctx.blockingTaskExecutor());
    } else {
        return f.thenApply(msg -> convertResponse(ctx, req, null, invoke(ctx, req, msg),
                                                  HttpHeaders.of()));
    }
}
```

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpService.java
```diff
@@ -232,19 +235,48 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
...
+                    logger.warn("The current thread is using blockTaskExecutor. It'll migrate to use" +
+                                " event loop thread when soon. Therefore, use @Blocking to avoid" +
+                                " unexpected problems.");
```
minwoox: How about adding the name so that a user easily notice? `object.getClass().getSimpleName() + '#' + method.getName()`

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -379,10 +384,20 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
...
-                                                                               defaultTrailers.build()),
+                                                                               defaultTrailers.build(),
+                                                                               useBlocking),
                                                decorator(method, clazz, initialDecorator));
     }
 
+    /**
+     * Returns the default value if you do not use {@link Blocking} in annotated methods.
+     */
+    private static boolean defaultBlocking(Method method) {
```
minwoox: Then we don't need this.

heowc:
Thank you for reviewing draft!
It will help a lot in further development 👍

core/src/main/java/com/linecorp/armeria/server/annotation/Blocking.java
```diff
@@ -0,0 +1,35 @@
...
+ * Specifies whether blocking of thread that would be produced by the annotated service method.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Blocking {
+
+    /**
+     * Whether blocking of thread for the annotated service method.
+     */
+    boolean value() default true;
```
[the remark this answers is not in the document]
heowc: Fixed 1381f7d

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpService.java
heowc: Fixed 1efe13d

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpService.java
heowc: Fixed 9cdc252

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
heowc: Fixed 66028ca

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpService.java
minwoox: @heowc Thanks for the update.
I think it's better to just remove the flag. So let's just make a breaking change here. I'm sorry about that. 😄 
We have talked about the migration plan again, but we thought it's too much and we can just leave the information about that in the release note well.

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceTest.java
```diff
@@ -985,6 +1084,64 @@ public void testReturnVoid() throws Exception {
...
 
+    @Test
+    public void testOriginBlockingTaskType() throws Exception {
+        final HttpClient client = HttpClient.of(server.uri("/"));
+
+        String path = "/12/httpResponse";
+        RequestHeaders headers = RequestHeaders.of(HttpMethod.GET, path);
+        AggregatedHttpResponse res = client.execute(headers).aggregate().join();
+        assertThat(res.status()).isSameAs(HttpStatus.OK);
+        assertThat(blockingCount).hasValue(0);
```
ikhoon: I think we can reduce test logic using `@ParameterizedTest` with `@CsvSource` :-) 
```java
@CsvSource({ "/12/httpResponse, 0", "/12/aggregatedHttpResponse, 1", .. })
```

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceTest.java
```diff
@@ -985,6 +1084,64 @@ public void testReturnVoid() throws Exception {
...
+
+    @Test
+    public void testOnlyBlockingTaskType() throws Exception {
+        final HttpClient client = HttpClient.of(server.uri("/"));
+
+        String path = "/13/httpResponse";
+        RequestHeaders headers = RequestHeaders.of(HttpMethod.GET, path);
+        AggregatedHttpResponse res = client.execute(headers).aggregate().join();
+        assertThat(res.status()).isSameAs(HttpStatus.OK);
+        assertThat(blockingCount).hasValue(1);
```
ikhoon: Ditto

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceTest.java
```diff
@@ -86,12 +97,32 @@
 import com.linecorp.armeria.server.annotation.ResponseConverterFunction;
 import com.linecorp.armeria.server.logging.LoggingService;
 import com.linecorp.armeria.testing.internal.AnticipatedException;
-import com.linecorp.armeria.testing.junit4.server.ServerRule;
+import com.linecorp.armeria.testing.junit.server.ServerExtension;
 
 public class AnnotatedHttpServiceTest {
```
ikhoon: Remove `public`?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceTest.java
```diff
@@ -672,6 +708,69 @@ public String headerWithoutValue(RequestContext ctx,
...
+        public CompletionStage<String> completionStage(RequestContext ctx) {
+            validateContext(ctx);
+            return CompletableFuture.supplyAsync(() -> "Armeria");
+        }
+    }
+
+    @BeforeEach
+    void clear() {
+        blockingCount.set(0);
+    }
```
ikhoon: How about moving this logic up near the `blockingCount` initialization? or remove `final` and create `new AtomicInteger` before each test?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -297,6 +297,9 @@
             exceptionLoggingMode("annotatedServiceExceptionVerbosity",
                                  DEFAULT_ANNOTATED_SERVICE_EXCEPTION_VERBOSITY);
 
+    private static final boolean DEFAULT_USE_EVENT_LOOP_FOR_ANNOTATED_SERVICE =
+            getBoolean("defaultUseEventLoopForAnnotatedService", false);
```
trustin: How about not introducing any flag for this new behavior and letting our users just use `@Blocking`, so that we do not have to maintain this way too long flag?

minwoox:
> I'm going to look a bit further to solve the problem. Or should we use @Blocking for this test?

Let's get rid of the test. 😄

heowc:
> > I'm going to look a bit further to solve the problem. Or should we use @Blocking for this test?
> 
> Let's get rid of the test. 😄

Oh. Is that okay?
If there is similar test code again, this should also be removed.

minwoox:
> Oh. Is that okay?
If there is similar test code again, this should also be removed.

I think so. 😄 That's the reason why we needed the migration plan, but we have decided to just make a breaking change.

ikhoon:
Nice work! 👍

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceTest.java
```diff
@@ -793,7 +783,7 @@ public void testParam() throws Exception {
 
     @Test
     public void testAdvancedAnnotatedHttpService() throws Exception {
```
minwoox: Could you remove all `public` modifiers in this class as well?

heowc: I was remove `public`

minwoox:
@heowc Oops, could you resolve the conflict, please?

heowc:
@minwoox I have resolved the conflict!

minwoox:
Thanks, @heowc! 👍

PR #2188, opened 2019-10-15
Add `Server` and `Date` header by default

### LINKED ISSUE #2137: Add `Server` and `Date` header by default.
.. also with the options that remove them. (I wonder if `Date` header is mandatory though.)

We also would need to make sure that `ctx.additionalResponseHeaders()` can override the default headers.

jyblue:
Motivation:
* Issue #2137

Modifications:
* Modify `HttpResponseSubscriber.onNext()` to add `SERVER` and `DATE` header by default
* Add boolean `enableServerHeader` field to `ServerBuilder` and `ServerConfig`
* Add boolean `enableDateHeader` field to `ServerBuilder` and `ServerConfig`
* Add `ServerBuilder.disableSeverHeader()` to disable default `Server` header.
* Add `ServerBuilder.disableDateHeader()` to disable default `Date` header.
* Add test cases

Result:
* Fixes #2137
* HttpResponse will include `Server` header by default with following format
```
server: Armeria/0.94.1
```
* HttpResponse will include `Date` header by default with following format (RFC 1123)
```
date: Wed, 16 Oct 2019 08:26:14 GMT
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -71,16 +71,11 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
     private static final Set<AsciiString> ADDITIONAL_HEADER_BLACKLIST = ImmutableSet.of(
             HttpHeaderNames.SCHEME, HttpHeaderNames.STATUS, HttpHeaderNames.METHOD, HttpHeaderNames.PATH);
 
-    private static final String ARTIFACT_ID = "armeria";
-    private static final String SERVER_NAME = "Armeria";
     private static final String SERVER_HEADER;
 
     static {
-        SERVER_HEADER = String.format("%s/%s (%s)",
-                                      SERVER_NAME,
-                                      Version.identify(HttpResponseSubscriber.class.getClassLoader())
-                                             .get(ARTIFACT_ID).artifactVersion(),
-                                      System.getProperty("os.name"));
+        SERVER_HEADER = "Armeria/" + Version.identify(HttpResponseSubscriber.class.getClassLoader())
+                                            .get("armeria").artifactVersion();
     }
 
     enum State {
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -73,14 +73,14 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
 
     private static final String ARTIFACT_ID = "armeria";
     private static final String SERVER_NAME = "Armeria";
-    private static final String serverHeader;
+    private static final String SERVER_HEADER;
 
     static {
-        serverHeader = String.format("%s/%s (%s)",
-                                     SERVER_NAME,
-                                     Version.identify(HttpResponseSubscriber.class.getClassLoader())
-                                            .get(ARTIFACT_ID).artifactVersion(),
-                                     System.getProperty("os.name"));
+        SERVER_HEADER = String.format("%s/%s (%s)",
+                                      SERVER_NAME,
+                                      Version.identify(HttpResponseSubscriber.class.getClassLoader())
+                                             .get(ARTIFACT_ID).artifactVersion(),
+                                      System.getProperty("os.name"));
     }
 
     enum State {
@@ -222,7 +222,7 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                 }
 
                 if (useServerHeader && !newHeaders.contains(HttpHeaderNames.SERVER)) {
-                    newHeaders.add(HttpHeaderNames.SERVER, serverHeader);
+                    newHeaders.add(HttpHeaderNames.SERVER, SERVER_HEADER);
                 }
 
                 if (useDateHeader && !newHeaders.contains(HttpHeaderNames.DATE)) {
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -71,6 +71,18 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
     private static final Set<AsciiString> ADDITIONAL_HEADER_BLACKLIST = ImmutableSet.of(
             HttpHeaderNames.SCHEME, HttpHeaderNames.STATUS, HttpHeaderNames.METHOD, HttpHeaderNames.PATH);
 
+    private static final String ARTIFACT_ID = "armeria";
+    private static final String SERVER_NAME = "Armeria";
+    private static final String serverHeader;
+
+    static {
+        serverHeader = String.format("%s/%s (%s)",
+                                     SERVER_NAME,
+                                     Version.identify(HttpResponseSubscriber.class.getClassLoader())
+                                            .get(ARTIFACT_ID).artifactVersion(),
+                                     System.getProperty("os.name"));
+    }
+
     enum State {
         NEEDS_HEADERS,
         NEEDS_DATA_OR_TRAILERS,
@@ -208,15 +220,12 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                 }
 
                 if (!newHeaders.contains(HttpHeaderNames.SERVER) &&
-                    config.defaultServerNameResponseHeader()) {
-                    final String verInfo = String.format("Armeria/%s (%s)",
-                                                         Version.identify().get("armeria").artifactVersion(),
-                                                         System.getProperty("os.name"));
-                    newHeaders.add(HttpHeaderNames.SERVER, verInfo);
+                    config.includeServerHeader()) {
+                    newHeaders.add(HttpHeaderNames.SERVER, serverHeader);
                 }
 
                 if (!newHeaders.contains(HttpHeaderNames.DATE) &&
-                    config.defaultServerDateResponseHeader()) {
+                    config.includeDateHeader()) {
                     newHeaders.add(HttpHeaderNames.DATE,
                                    DateTimeFormatter.RFC_1123_DATE_TIME.format(
                                            ZonedDateTime.now(ZoneOffset.UTC)));
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -85,8 +85,8 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
     private final HttpObjectEncoder responseEncoder;
     private final DecodedHttpRequest req;
     private final DefaultServiceRequestContext reqCtx;
-    private final boolean useServerHeader;
-    private final boolean useDateHeader;
+    private final boolean enableServerHeader;
+    private final boolean enableDateHeader;
     private final long startTimeNanos;
 
     @Nullable
@@ -100,13 +100,13 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
 
     HttpResponseSubscriber(ChannelHandlerContext ctx, HttpObjectEncoder responseEncoder,
                            DefaultServiceRequestContext reqCtx, DecodedHttpRequest req,
-                           boolean useServerHeader, boolean useDateHeader) {
+                           boolean enableServerHeader, boolean enableDateHeader) {
         this.ctx = ctx;
         this.responseEncoder = responseEncoder;
         this.req = req;
         this.reqCtx = reqCtx;
-        this.useServerHeader = useServerHeader;
-        this.useDateHeader = useDateHeader;
+        this.enableServerHeader = enableServerHeader;
+        this.enableDateHeader = enableDateHeader;
         startTimeNanos = System.nanoTime();
     }
 
@@ -213,11 +213,11 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                     newHeaders.remove(HttpHeaderNames.CONTENT_LENGTH);
                 }
 
-                if (useServerHeader && !newHeaders.contains(HttpHeaderNames.SERVER)) {
+                if (enableServerHeader && !newHeaders.contains(HttpHeaderNames.SERVER)) {
                     newHeaders.add(HttpHeaderNames.SERVER, SERVER_HEADER);
                 }
 
-                if (useDateHeader && !newHeaders.contains(HttpHeaderNames.DATE)) {
+                if (enableDateHeader && !newHeaders.contains(HttpHeaderNames.DATE)) {
                     newHeaders.add(HttpHeaderNames.DATE,
                                    DateTimeFormatter.RFC_1123_DATE_TIME.format(
                                            ZonedDateTime.now(ZoneOffset.UTC)));
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -71,12 +71,9 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
     private static final Set<AsciiString> ADDITIONAL_HEADER_BLACKLIST = ImmutableSet.of(
             HttpHeaderNames.SCHEME, HttpHeaderNames.STATUS, HttpHeaderNames.METHOD, HttpHeaderNames.PATH);
 
-    private static final String SERVER_HEADER;
-
-    static {
-        SERVER_HEADER = "Armeria/" + Version.identify(HttpResponseSubscriber.class.getClassLoader())
-                                            .get("armeria").artifactVersion();
-    }
+    private static final String SERVER_HEADER =
+            "Armeria/" + Version.identify(HttpResponseSubscriber.class.getClassLoader()).get("armeria")
+                                .artifactVersion();
 
     enum State {
         NEEDS_HEADERS,
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -19,6 +19,9 @@ package com.linecorp.armeria.server;
 import static com.linecorp.armeria.internal.ArmeriaHttpUtil.isTrailerBlacklisted;
 
 import java.nio.channels.ClosedChannelException;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -46,6 +49,7 @@ import com.linecorp.armeria.common.ResponseHeadersBuilder;
 import com.linecorp.armeria.common.logging.RequestLogAvailability;
 import com.linecorp.armeria.common.logging.RequestLogBuilder;
 import com.linecorp.armeria.common.stream.AbortedStreamException;
+import com.linecorp.armeria.common.util.Version;
 import com.linecorp.armeria.internal.Http1ObjectEncoder;
 import com.linecorp.armeria.internal.HttpObjectEncoder;
 
@@ -77,6 +81,7 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
     private final HttpObjectEncoder responseEncoder;
     private final DecodedHttpRequest req;
     private final DefaultServiceRequestContext reqCtx;
+    private final ServerConfig config;
     private final long startTimeNanos;
 
     @Nullable
@@ -89,11 +94,13 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
     private boolean loggedResponseHeadersFirstBytesTransferred;
 
     HttpResponseSubscriber(ChannelHandlerContext ctx, HttpObjectEncoder responseEncoder,
-                           DefaultServiceRequestContext reqCtx, DecodedHttpRequest req) {
+                           DefaultServiceRequestContext reqCtx, DecodedHttpRequest req,
+                           ServerConfig config) {
         this.ctx = ctx;
         this.responseEncoder = responseEncoder;
         this.req = req;
         this.reqCtx = reqCtx;
+        this.config = config;
         startTimeNanos = System.nanoTime();
     }
 
@@ -198,6 +205,21 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                     // prevent the trailers from being sent so we go ahead and remove content-length to force
                     // chunked encoding.
                     newHeaders.remove(HttpHeaderNames.CONTENT_LENGTH);
+                }
+
+                if (!newHeaders.contains(HttpHeaderNames.SERVER) &&
+                    config.defaultServerNameResponseHeader()) {
+                    final String verInfo = String.format("Armeria/%s (%s)",
+                                                         Version.identify().get("armeria").artifactVersion(),
+                                                         System.getProperty("os.name"));
+                    newHeaders.add(HttpHeaderNames.SERVER, verInfo);
+                }
+
+                if (!newHeaders.contains(HttpHeaderNames.DATE) &&
+                    config.defaultServerDateResponseHeader()) {
+                    newHeaders.add(HttpHeaderNames.DATE,
+                                   DateTimeFormatter.RFC_1123_DATE_TIME.format(
+                                           ZonedDateTime.now(ZoneOffset.UTC)));
                 }
 
                 headers = newHeaders.build();
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -219,13 +219,13 @@ final class HttpResponseSubscriber implements Subscriber<HttpObject>, RequestTim
                     newHeaders.remove(HttpHeaderNames.CONTENT_LENGTH);
                 }
 
-                if (!newHeaders.contains(HttpHeaderNames.SERVER) &&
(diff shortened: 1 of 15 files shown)


core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -366,6 +369,16 @@ private void handleRequest(ChannelHandlerContext ctx, DecodedHttpRequest req) th
                 protocol, routingCtx, routingResult, req, getSSLSession(channel),
                 proxiedAddresses, clientAddress);
 
+        if (config.defaultServerNameResponseHeader()) {
+            reqCtx.addAdditionalResponseHeader(HttpHeaderNames.SERVER, "armeria");
```
minwoox: Let add version and os name using so that the header will be `Armeria/0.94.0 (Linux)`.
We can use `Version` and `System.getProperty("os.name")`.

minwoox: This value never changes so let's make a `private static final String` field for this.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1502,6 +1505,30 @@ public ServerBuilder contentPreview(int length) {
         return contentPreview(length, ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET);
     }
 
+    /**
+     * Sets whether the response header will include Server header.
+     */
+    public ServerBuilder setDefaultServerNameResponseHeader(boolean defaultServerNameResponseHeader) {
```
minwoox: We don't use `set` prefix throughout this class. How about just `includeServerHeader`? Not sure this is the best name. /cc @trustin and @ikhoon

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1502,6 +1505,30 @@ public ServerBuilder contentPreview(int length) {
...
+    }
+
+    boolean defaultServerNameResponseHeader() {
+        return defaultServerNameResponseHeader;
+    }
+
+    /**
+     * Sets whether the response header will include Date header.
+     */
+    public ServerBuilder setDefaultServerDateResponseHeader(boolean defaultServerDateResponseHeader) {
```
minwoox: `includeDateHeader`?

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
minwoox: We should do this only when a user does not specify this. Let's consider that a user specifies a header in the same manner:
```java
reqCtx.addAdditionalResponseHeader(HttpHeaderNames.SERVER, "myServer");
```
Then the response header will contain both values which he/she does not expect.
We should add it [here](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java#L202).
Please note how we add `HttpHeaderNames.CONTENT_LENGTH`. We should do it in the same way.

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -200,6 +207,21 @@ public void onNext(HttpObject o) {
                     newHeaders.remove(HttpHeaderNames.CONTENT_LENGTH);
                 }
 
+                if (!newHeaders.contains(HttpHeaderNames.SERVER) &&
+                    config.defaultServerNameResponseHeader()) {
+                    final String verInfo = String.format("Armeria/%s (%s)",
+                                                         Version.identify().get("armeria").artifactVersion(),
+                                                         System.getProperty("os.name"));
```
minwoox: Let's make this as `private static final String serverHeader = ....`

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -109,6 +109,9 @@
     @Nullable
     private String strVal;
 
+    private boolean defaultServerNameResponseHeader;
```
minwoox: `includeServerHeader`?

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -109,6 +109,9 @@
...
+    private boolean defaultServerDateResponseHeader;
```
minwoox: `includeDateHeader`?

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -200,6 +207,21 @@ public void onNext(HttpObject o) {
                     newHeaders.remove(HttpHeaderNames.CONTENT_LENGTH);
                 }
 
+                if (!newHeaders.contains(HttpHeaderNames.SERVER) &&
+                    config.defaultServerNameResponseHeader()) {
```
trustin: Could you check the `config` first, which is relatively cheaper than `headers.contains()`?

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -200,6 +207,21 @@ public void onNext(HttpObject o) {
...
+                    final String verInfo = String.format("Armeria/%s (%s)",
+                                                         Version.identify().get("armeria").artifactVersion(),
+                                                         System.getProperty("os.name"));
+                    newHeaders.add(HttpHeaderNames.SERVER, verInfo);
+                }
+
+                if (!newHeaders.contains(HttpHeaderNames.DATE) &&
+                    config.defaultServerDateResponseHeader()) {
```
trustin: Ditto: Could you check the `config` first, which is relatively cheaper than `headers.contains()`?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -203,6 +203,9 @@
     private Function<VirtualHost, Logger> accessLoggerMapper = host -> LoggerFactory.getLogger(
             defaultAccessLoggerName(host.hostnamePattern()));
 
+    private boolean includeServerHeader = true;
+    private boolean includeDateHeader = true;
```
trustin: Which is better, `include`, `use` or `add`? I think `use` sounds more natural:

```java
Server.builder()
      .useServerHeader(false)
      .useDateHeader(false)
```

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
```diff
@@ -0,0 +1,114 @@
...
+import com.linecorp.armeria.client.HttpClient;
+import com.linecorp.armeria.common.AggregatedHttpResponse;
+import com.linecorp.armeria.common.HttpHeaderNames;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.testing.junit.server.ServerExtension;
+
+class HttpServerDefaultHeadersTest {
+    @RegisterExtension
+    static final ServerExtension server1 = new ServerExtension() {
```
trustin: `serverWithDefaults`

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
```diff
@@ -0,0 +1,114 @@
...
+        @Override
+        protected void configure(ServerBuilder sb) throws Exception {
+            sb.service("/", (ctx, req) -> {
+                return HttpResponse.of(HttpStatus.OK);
+            });
+        }
+    };
+
+    @RegisterExtension
+    static final ServerExtension server2 = new ServerExtension() {
```
trustin: `serverWithoutServerHeader`

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
```diff
@@ -0,0 +1,114 @@
...
+        protected void configure(ServerBuilder sb) throws Exception {
+            sb.includeDateHeader(false);
+            sb.service("/", (ctx, req) -> {
+                return HttpResponse.of(HttpStatus.OK);
+            });
+        }
+    };
+
+    @RegisterExtension
+    static final ServerExtension server4 = new ServerExtension() {
```
trustin: `serverWithServerHeaderOverridden`

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
```diff
@@ -0,0 +1,114 @@
...
+    };
+
+    @Test
+    void testServerNameAndDateHeaderIncludedByDefault() {
+        final HttpClient client = HttpClient.of(server1.httpUri("/"));
+        final AggregatedHttpResponse res = client.get("/").aggregate().join();
+        assertThat(res.status()).isEqualTo(HttpStatus.OK);
+        assertThat(res.headers().names()).contains(HttpHeaderNames.SERVER,
+                                                   HttpHeaderNames.DATE);
+        assertThat(res.headers().get(HttpHeaderNames.SERVER)).isEqualTo("armeria");
```
trustin: Could you also make sure that the `date` header is a valid date?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
minwoox: Let's use `use` then. 😄

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
jyblue: question : 
now `Server` header like this
```
server: Armeria/0.94.1-SNAPSHOT (Linux)
```
Is it okay? I'm not sure `-SNAPSHOT` should be included.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
jyblue: Fixed to use `useServerHeader`.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
jyblue: Fixed to `useDateHeader`.

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
jyblue: Fixed it.

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
jyblue: Fixed to `useServerHeader`.

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
jyblue: Fixed to `useDateHeader`.

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
jyblue: Fixed it. Thank you for education!

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
```diff
@@ -0,0 +1,114 @@
...
+        protected void configure(ServerBuilder sb) throws Exception {
+            sb.includeServerHeader(false);
+            sb.service("/", (ctx, req) -> {
+                return HttpResponse.of(HttpStatus.OK);
+            });
+        }
+    };
+
+    @RegisterExtension
+    static final ServerExtension server3 = new ServerExtension() {
```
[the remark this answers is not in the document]
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
jyblue: Fixed it.

core/src/test/java/com/linecorp/armeria/server/HttpServerDefaultHeadersTest.java
jyblue: Added `date` validation.

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -200,6 +219,18 @@ public void onNext(HttpObject o) {
...
+                if (config.useServerHeader() &&
+                    !newHeaders.contains(HttpHeaderNames.SERVER)) {
+                    newHeaders.add(HttpHeaderNames.SERVER, serverHeader);
+                }
+
+                if (config.useDateHeader() &&
+                    !newHeaders.contains(HttpHeaderNames.DATE)) {
+                    newHeaders.add(HttpHeaderNames.DATE,
+                                   DateTimeFormatter.RFC_1123_DATE_TIME.format(
+                                           ZonedDateTime.now(ZoneOffset.UTC)));
```
jyblue: question : 
AFAIU, Server need to use `RFC_1123_DATE_TIME` format by `RFC 1123`
But, I'm not sure which time zone should we use.
In this case, I just followed @anuraaga's example from [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/4a1aad25baa71e52827d6ea4d1f78dd30dfb153f/frameworks/Java/armeria/src/main/java/hello/App.java#L27)

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
ikhoon: I suggest new method name(if we enable `Server` and `Date` header by default).  🧐

```java
Server.builder()
      .disableServerHeader()
      .disableDateHeader()
```

ikhoon:
I left a comment about turning off the default headers. https://github.com/line/armeria/pull/2188#discussion_r336277515

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
ikhoon: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date
from the above: 
HTTP dates are always expressed in GMT, never in local time.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
minwoox: I think this is not a good idea. People will suddenly see the header after when they upgrade and be a bit surprised.

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
ikhoon: I think it doesn't matter because we always release without `-SNAPSHOT`. You can only see it while developing `Armeria` 😎

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
minwoox: Didn't know that the spec said sending `date` header is [mandatory](https://tools.ietf.org/html/rfc7231#section-7.1.1.2). Had a chat with @ikhoon and this is probably not a bad idea.
However, I think `disableDateHeader` can be a bit confusing because some could take that as not sending `date` header at all. So we should add `default` which is `disableDefaultDateHeader`. But it's too long, so I'm not sure which one is the best. 😄

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -67,6 +71,18 @@
...
+    private static final String ARTIFACT_ID = "armeria";
+    private static final String SERVER_NAME = "Armeria";
+    private static final String serverHeader;
+
+    static {
+        serverHeader = String.format("%s/%s (%s)",
+                                     SERVER_NAME,
+                                     Version.identify(HttpResponseSubscriber.class.getClassLoader())
+                                            .get(ARTIFACT_ID).artifactVersion(),
+                                     System.getProperty("os.name"));
```
ikhoon: Had chat about `os.name` and we decided to remove `os.name` from `Server` header. Could you make `Server` header with `Armeria/{artifact_id}`? :-)

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
jyblue: Thank you for education :+1:

core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
jyblue: I see, thank you! :grinning:

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
jyblue: Do you mean
```
Server: Armeria/armeria
```

or

```
Server: Armeria/0.94.1
```
It seems `Server: Armeria/{ARTIFACT_VERSION}` is corrent?

minwoox: yes the version. 😅

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -67,6 +71,18 @@
     private static final Set<AsciiString> ADDITIONAL_HEADER_BLACKLIST = ImmutableSet.of(
             HttpHeaderNames.SCHEME, HttpHeaderNames.STATUS, HttpHeaderNames.METHOD, HttpHeaderNames.PATH);
 
+    private static final String ARTIFACT_ID = "armeria";
+    private static final String SERVER_NAME = "Armeria";
```
minwoox: Let's just inline these. 😄

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
ikhoon: Oops, artifact version!

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
jyblue: Fixed it :grinning:

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
jyblue: Thank you, fixed it :grinning:

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -67,6 +71,13 @@
     private static final Set<AsciiString> ADDITIONAL_HEADER_BLACKLIST = ImmutableSet.of(
             HttpHeaderNames.SCHEME, HttpHeaderNames.STATUS, HttpHeaderNames.METHOD, HttpHeaderNames.PATH);
 
+    private static final String SERVER_HEADER;
+
+    static {
```
minwoox: We do not need this static block. 😄

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1502,6 +1505,30 @@ public ServerBuilder contentPreview(int length) {
         return contentPreview(length, ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET);
     }
 
+    /**
+     * Sets whether the response header will include Server header.
+     */
+    public ServerBuilder useServerHeader(boolean useServerHeader) {
```
minwoox: So could you rename this to `disableServerHeader` and remove the boolean parameter?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1502,6 +1505,30 @@ public ServerBuilder contentPreview(int length) {
         return contentPreview(length, ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET);
     }
 
+    /**
+     * Sets whether the response header will include Server header.
```
trustin: `{@code "Server"}`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1502,6 +1505,30 @@ public ServerBuilder contentPreview(int length) {
...
+        this.useServerHeader = useServerHeader;
+        return this;
+    }
+
+    boolean useServerHeader() {
+        return useServerHeader;
+    }
+
+    /**
+     * Sets whether the response header will include Date header.
```
trustin: `{@code "Date"}`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1502,6 +1505,30 @@ public ServerBuilder contentPreview(int length) {
         return contentPreview(length, ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET);
     }
 
+    /**
+     * Sets whether the response header will include {@code "Server"} header.
+     */
+    public ServerBuilder disableServerHeader() {
+        this.useServerHeader = false;
```
anuraaga: Especially since the method is `disable`, I think it's more readable if the variable is called `enableServerHeader` instead of `use`

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1502,6 +1505,30 @@ public ServerBuilder contentPreview(int length) {
...
+        return this;
+    }
+
+    boolean useServerHeader() {
```
anuraaga: @trustin Is this our normal style for boolean getter? It really confuses me since it looks like it should actually mutate the field to `true`

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
jyblue: Removed the static block!

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -67,6 +71,18 @@
     private static final Set<AsciiString> ADDITIONAL_HEADER_BLACKLIST = ImmutableSet.of(
             HttpHeaderNames.SCHEME, HttpHeaderNames.STATUS, HttpHeaderNames.METHOD, HttpHeaderNames.PATH);
 
+    private static final String ARTIFACT_ID = "armeria";
+    private static final String SERVER_NAME = "Armeria";
+    private static final String serverHeader;
```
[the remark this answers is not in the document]
jyblue: Fixed it!

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
jyblue: Fixed it !

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
jyblue: Fixed it!

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
jyblue: I tried to use same style of this class.
Like this,
https://github.com/line/armeria/blob/59d98c355b22aad1a566e94aefafd3848ae20dee/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java#L1430
But, I agree that it's confusing.
I think `useServerHeader()` sounds like `setter` because it start with `use`.
What about `IsServerHeaderEnabled()`? or any good name?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: `isServerHeaderEnabled` sounds better to me.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: `isServerHeaderEnabled` sounds better to me.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
jyblue: Renamed variables to `enableServerHeader` and `enableDateHeader`.
Thank you!

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
jyblue: Renamed to `isServerHeaderEnabled` and `isDateHeaderEnabled`.

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -109,6 +109,9 @@
     @Nullable
     private String strVal;
 
+    private boolean enableServerHeader;
+    private boolean enableDateHeader;
```
ikhoon: Add `final`?

jyblue: Fixed it :grinning:

PR #2191, opened 2019-10-15
Allow entering DocService demo mode using a Gradle property

trustin:
Motivation:

When launching a `*DocServiceTest` from Gradle via IntelliJ, the
`DOC_SERVICE_DEMO` environment variable is not propagated to the test
cases.

Modifications:

- Update the build script so that the `DOC_SERVICE_DEMO` environment
  variable is set when a user passes the `-Pdemo` option to Gradle.

Result:

- More convenient testing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,6 +47,10 @@ allprojects {
         if (project.hasFlags('coverage')) {
             systemProperty 'com.linecorp.armeria.testing.coverage', 'true'
         }
+        // Pass special environment variable to enable demo mode for `*DocServiceTest`.
+        if (rootProject.hasProperty('demo')) {
+            environment['DOC_SERVICE_DEMO'] = 'true'
+        }
         // Use a different JRE for testing if necessary.
         if (javaTestHome) {
             executable "${javaTestHome}/bin/java"


anuraaga:
Do you mean when using Gradle runner to run the test? Hmm, it works fine for me (a relatively, but not that, think late last year, recent version of IntelliJ fixed the env variable propagation for Gradle runner).

This sort of setup works fine for me
image

trustin:
That didn't work for me with IntelliJ IDEA EAP 2019.3, but let me try again :thinking:

anuraaga:
FWIW I haven't tried with EAP, hope it's not broken there...

build.gradle
```diff
@@ -47,6 +47,10 @@ allprojects {
         if (project.hasFlags('coverage')) {
             systemProperty 'com.linecorp.armeria.testing.coverage', 'true'
         }
+        // Pass special environment variable to enable demo mode for `*DocServiceTest`.
+        if (rootProject.hasProperty('demo')) {
```
ikhoon: nit: Mostly LGTM, but I am worried `demo` has a too wide meaning. How about renaming to  `docservice.demo`?

trustin:
Tried again today and it seems to work fine with the env var. :sweat_smile:

PR #2196, opened 2019-10-18
Add optimized supplier of current HTTP timestamp

anuraaga:
As TechEmpower suggests, this caches HTTP timestamps for any given second. I can wire this into the `Date` header after that's in if this makes sense :)

```
Benchmark                                  Mode  Cnt         Score         Error  Units
HttpTimestampSupplierBenchmark.cached     thrpt    5  37665728.115 ▒  324874.523  ops/s
HttpTimestampSupplierBenchmark.notCached  thrpt    5   2099035.297 ▒ 1176888.041  ops/s
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -19,9 +19,6 @@ package com.linecorp.armeria.server;
 import static com.linecorp.armeria.internal.ArmeriaHttpUtil.isTrailerBlacklisted;
 
 import java.nio.channels.ClosedChannelException;
-import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
-import java.time.format.DateTimeFormatter;
 import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/server/cors/CorsConfig.java
```diff
@@ -208,13 +208,12 @@ public String toString() {
...
-
-        static final InstantValueSupplier INSTANCE = new InstantValueSupplier();
+    enum TimestampSupplier implements Supplier<String> {
+        INSTANCE;
 
         @Override
-        public Instant get() {
-            return Instant.now();
+        public String get() {
+            return HttpTimestampSupplier.currentTime();
```
trustin: Perhaps we could remove this class and just use `HttpTimestampSupplier::currentTime`?

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -218,9 +216,7 @@ public void onNext(HttpObject o) {
                 }
 
                 if (enableDateHeader && !newHeaders.contains(HttpHeaderNames.DATE)) {
-                    newHeaders.add(HttpHeaderNames.DATE,
-                                   DateTimeFormatter.RFC_1123_DATE_TIME.format(
-                                           ZonedDateTime.now(ZoneOffset.UTC)));
+                    newHeaders.add(HttpHeaderNames.DATE, HttpTimestampSupplier.currentTime());
```
trustin: Thanks! /cc @jyblue

jyblue: Thank you @anuraaga 
This PR gave me a good education 👍

PR #2197, opened 2019-10-18
Add some missing Javadoc comments

### LINKED ISSUE #2136: Use `MissingJavadoc*` checks for Checkstyle
We should add the following checks:

- `MissingJavadocPackage`
- `MissingJavadocType`

And remove the `minLineCount` property from:

- `MissingJavadocMethod`

Stupremee:
For the following classes (in the bechmark module) I don't know what kind of Javadoc I want to write for them. That's why I need some help. 
- com.linecorp.armeria.core.client.retry.WithDuplicator
- com.linecorp.armeria.core.client.retry.WithoutDuplicator
- com.linecorp.armeria.grpc.downstream.DownstreamSimpleBenchmark
- com.linecorp.armeria.grpc.upstream.UpstreamSimpleBenchmark
- com.linecorp.armeria.retrofit2.downstream.DownstreamSimpleBenchmark
- com.linecorp.armeria.retrofit2.upstream.UpstreamSimpleBenchmark

Fixes #2136

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/core/HttpServerBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/core/HttpServerBenchmark.java
@@ -38,7 +38,6 @@ import com.linecorp.armeria.shared.AsyncCounters;
 
 /**
  * Microbenchmarks of a {@link Server}.
- *
  */
 @State(Scope.Benchmark)
 public class HttpServerBenchmark {
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/core/HttpServerBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/core/HttpServerBenchmark.java
@@ -36,11 +36,15 @@ import com.linecorp.armeria.server.Server;
 import com.linecorp.armeria.server.ServerPort;
 import com.linecorp.armeria.shared.AsyncCounters;
 
+/**
+ * Microbenchmarks of a {@link Server}.
+ *
+ */
 @State(Scope.Benchmark)
 public class HttpServerBenchmark {
 
     // JMH bug prevents it from using enums that override toString() (it should use name() instead...).
-    public enum Protocol {
+    private enum Protocol {
         H2C(SessionProtocol.H2C),
         H1C(SessionProtocol.H1C);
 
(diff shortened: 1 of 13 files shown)


benchmarks/src/jmh/java/com/linecorp/armeria/core/HttpServerBenchmark.java
```diff
@@ -36,11 +36,15 @@
 import com.linecorp.armeria.server.ServerPort;
 import com.linecorp.armeria.shared.AsyncCounters;
 
+/**
+ * Microbenchmarks of a {@link Server}.
+ *
```
jyblue: This empty line can be removed :smile:

benchmarks/src/jmh/java/com/linecorp/armeria/core/client/endpoint/WeightedRoundRobinStrategyBenchmark.java
```diff
@@ -31,6 +31,10 @@
 import com.linecorp.armeria.client.endpoint.EndpointSelector;
 import com.linecorp.armeria.client.endpoint.StaticEndpointGroup;
 
+/**
+ * Microbenchmarks of different {@link EndpointSelector} configurations.
+ *
```
jyblue: ditto: This empty line can be removed

trustin:
Thanks a lot for adding missing Javadocs, @Stupremee! It seems like there are still many Checkstyle violations to fix at the moment. Do you intend to fix them as well in this pull request?

Stupremee:
Yes, I'll fix the errors. But I have no idea how to comment on the classes listed above.

trustin:
I don't think you have to write Javadoc for what you are not familiar with. How about writing Javadoc only for what you can describe? It will still be a great step forward. We could add the `MissingJavadoc*` checks later in other pull request when everything is filled in. What do you think?

Stupremee:
Yes I can do it. But I still need to add some comments to make the build pass. Should I just use an empty comment?

trustin:
> Should I just use an empty comment?

An empty comment will fail the build anyway. My suggestion is to remove `Missing*` while keeping the Javadoc comments added. We can add the missing Javadoc blocks in another PR.

trustin:
Ping @ikhoon

trustin:
Thanks again, @Stupremee !

PR #2199, opened 2019-10-21
Add a comment what QWxhZGRpbjpPcGVuU2VzYW1l means

minwoox:
Got a bug bounty report that we have a hard-coded sensitive token in our repository.
But it's a not a sensitive token, but the base64-encoding of `Aladdin:OpenSesame` cited from https://en.wikipedia.org/wiki/Basic_access_authentication.


minwoox:
I thought just it's too much. So let me close this. 😄

Issue #2008, opened 2019-08-21
`NullPointerException` in `ArmeriaGrpcServerInteropTest`


trustin:
Found this from CI:

```
com.linecorp.armeria.server.grpc.interop.ArmeriaGrpcServerInteropTest > deadlineExceeded FAILED
    java.lang.NullPointerException
```

No stack trace unfortunately.

trustin:
Reproduced locally with the following stack trace:

```
java.lang.NullPointerException
	at java.base/java.util.regex.Matcher.getTextLength(Matcher.java:1770)
	at java.base/java.util.regex.Matcher.reset(Matcher.java:416)
	at java.base/java.util.regex.Matcher.<init>(Matcher.java:253)
	at java.base/java.util.regex.Pattern.matcher(Pattern.java:1133)
	at java.base/java.util.regex.Pattern.matches(Pattern.java:1174)
	at io.grpc.testing.integration.AbstractInteropTest.deadlineExceeded(AbstractInteropTest.java:1108)
```

trustin:
It seems like the upstream client expects the server to respond with `CANCELLED` rather than `DEADLINE_EXCEEDED`: https://github.com/grpc/grpc-java/blob/39dc883b342f4ffab175195bb0a2e809f8bcd164/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L654-L667

`deadlineExceeded()` does not fail anymore if I change `GrpcService` to close the stream with `CANCELLED` instead of `DEADLINE_EXCEEDED` here: https://github.com/line/armeria/blob/armeria-0.91.0/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java#L182

However, after changing that, I see an intermittent failure at `deaelineExceededServerStreaming()`:

```
java.lang.AssertionError: 
Expected :DEADLINE_EXCEEDED
Actual   :CANCELLED
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at io.grpc.testing.integration.AbstractInteropTest.deadlineExceededServerStreaming(AbstractInteropTest.java:1148)
```

@anuraaga Any suggestions about the right way to fix this?

anuraaga:
[...]

PR #2201, opened 2019-10-21
Allow handling response timeout with specific cause on the client-side

ikhoon:
Motivation:
`ServiceRequestContext` can configure response timeout handler by adding `setResponseTimeoutHandler`.
But `ClientRequestContext` is less customizable than server-side.
We need to add a response timeout handler in the client-side to solve the problem in #2008.

Modifications:
* Add `setResponseTimeoutHandler(Runnable responseTimeoutHandler)` to `ClientRequestContext`.
* Add `abort(Throwable)` and  abort(Supplier<? extends Throwable>) to `StreamMessage` and its implementations.
* Change default value of `goAwayReceived` to -1 in `Http2GoAwayHandler`
* Remove singleton instance of `AbortingSubscriber` and add new constructor `AbortingSubscriber(Supplier<? extends Throwable>)`.
* Update existing test logic.
* Dependency
  * Add TestNG(7.0.0) to core test dependency for running Reactive Streams TCK on modern IntelliJ IDEA  
 
Result:
A user can configure a custom response timeout handler to intercept default logic and abort a request with a specific cause.


brave/src/test/java/com/linecorp/armeria/it/brave/BraveIntegrationTest.java
```diff
@@ -416,7 +416,8 @@ public void testClientTimesOut() throws Exception {
 
         // Client timed out, so no response data was ever sent from the server. There is no wire send in the
         // server and no wire receive in the client.
-        assertThat(serverSpan.annotations()).hasSize(1);
+        // FIXME(ikhoon) the server receives RST_STREAM frame, does it make sence?
+        assertThat(serverSpan.annotations()).hasSize(2);
```
ikhoon: I wonder why this has changed. 🧐

licenses/LICENSE.testng.al20.txt
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
anuraaga: Don't know if this is needed for a developement-time dependency (we don't package it into Armeria). Otherwise, our web licenses file is also missing a lot of development-time dependencies.

core/build.gradle
```diff
@@ -117,6 +117,9 @@ dependencies {
     // Project Reactor
     testCompile 'io.projectreactor:reactor-test'
 
+    // TestNG
+    testCompile 'org.testng:testng'
```
ikhoon: Workaround for:
```java
java.lang.NoSuchMethodError: org.testng.ITestNGMethod.getConstructorOrMethod()Lorg/testng/internal/ConstructorOrMethod;
```

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -253,6 +253,21 @@ ClientRequestContext newDerivedContext(@Nullable HttpRequest req, @Nullable RpcR
...
 
+    /**
+     * Returns {@link Response} timeout handler which is executed when
+     * the {@link Response} is not completely received within the allowed {@link #responseTimeoutMillis()}
+     * or the default {@link ClientOption#RESPONSE_TIMEOUT_MILLIS}.
+     */
+    @Nullable
+    Runnable responseTimeoutHandler();
+
+    /**
```
anuraaga: Think we should document something like this:

```java
/**
     * Sets a handler to run when the response times out. {@code responseTimeoutHandler} must close the request,
     * e.g., by calling {@link HttpRequestWriter#close(Throwable)}. If not set, the response will be closed with
     * {@link TimeoutException}.
     *
     * <p>For example,
     * <pre>{@code
     *   HttpRequestWriter req = HttpRequest.streaming();
     *   ctx.setResponseTimeoutHandler(() -> {
     *      req.close(new IllegalStateException("Server is in a bad state."));
     *   });
     *   ...
     * }</pre>
     */
```

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -334,6 +334,7 @@ private void failAndRespond(Throwable cause) {
         if (response.isOpen()) {
             response.close(cause);
             error = Http2Error.INTERNAL_ERROR;
+            // FIXME(ikhoon) fix here? wrap an exception with AbortedStreamException?
         } else if (cause instanceof WriteTimeoutException || cause instanceof AbortedStreamException) {
             error = Http2Error.CANCEL;
```
ikhoon: Now, an abort can have various causes with `abort(Throwable)` and `abort(Supply<? extends Throwable>)`.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
ikhoon: Thanks for the kind review! I didn't know you have been working on this issue... 😱 I should've shared working on that 😭

ikhoon:
@anuraaga I think we need your gRPC logic from https://github.com/line/armeria/commit/1c4b4bf378d6ca53a23a8c4841ad96a490f48c29#diff-5c8780680d870615d57375aa561006f6 
Do you want to update that logic after this PR? Or I can fork your code this PR.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
anuraaga: No problem, I didn't share either nor get back to the branch. Glad you send a PR :)

anuraaga:
I'm ok with either :) I don't think my code works yet which is why I never got to send a PR so maybe it's a little annoying to finish it up but happy if you'd like to give it a try.

licenses/LICENSE.testng.al20.txt
syleeeee: @anuraaga 
I think it depends on who we define as the "receiver" of Armeria. 
- If we define the receiver as the person who clones this repository, then we should notify them of the test library license. 
- Or if we define the receiver as the person who downloads Armeria release, then we don't need to.

syleeeee: So what do you think? 
Since we don't notify the licenses of front-end dev-dependencies, we don't need to add testing lib license either as @anuraaga said?
But @ikhoon said there are some licenses we are notifying, such as `mockito` and `junit`.

anuraaga: I'd recommend removing the other current test dependency licenses too. The main reason for license notification IMO is so users can easily confirm whether there code is license-compatible with Armeria. This doesn't apply to test code, only published code.

Tests don't have any compatibility issue for users, only for this repository itself which we can confirm when adding a new test dependency.

ikhoon:
@anuraaga I'm not special on `gRPC` internal so I need some time to learn it and fix the [`deadlineExceeded`](https://github.com/line/armeria/issues/2008) problem. Could you make a PR for #2008 issue? or guide me to work around. 😄

anuraaga:
Thanks for giving it a look. Yeah let's decouple that from this PR and keep this non-grpc and I'll take another look at that issue.

ikhoon:
Thanks for your gentle feedback. I will focus on the response timeout handler. 😀

licenses/LICENSE.testng.al20.txt
trustin: > I'd recommend removing the other current test dependency licenses too.

+1

brave/src/test/java/com/linecorp/armeria/it/brave/BraveIntegrationTest.java
trustin: I think it makes sense because the client would reset the stream when aborting the request. Could we add some comments about this and add more assertions for the properties of `serverSpan.annotations().get(1)` instead?

I also wonder what happens for HTTP/1, too. Could you check?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -64,7 +64,7 @@
 
     private static final AtomicReferenceFieldUpdater<DefaultClientRequestContext, HttpHeaders>
             additionalRequestHeadersUpdater = AtomicReferenceFieldUpdater.newUpdater(
-                    DefaultClientRequestContext.class, HttpHeaders.class, "additionalRequestHeaders");
+                DefaultClientRequestContext.class, HttpHeaders.class, "additionalRequestHeaders");
```
trustin: Revert?

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
trustin: I think it's fine to leave this as it is. If a user aborted with a different exception than `AbortedStreamException`, I'd say he or she wanted signal something else than an abort.

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -161,10 +161,21 @@ private void doExecute0(ClientRequestContext ctx, HttpRequestDuplicator rootReqD
...
+            originalReq.completionFuture().handle((result, cause) -> {
+                handleException(ctx, rootReqDuplicator, future, cause, initialAttempt);
+                return null;
+            });
+            return;
+        }
+        if (returnedRes.isComplete()) {
+            returnedRes.completionFuture().handle((result, cause) -> {
+                final Throwable abortCause = firstNonNull(cause, AbortedStreamException.get());
+                handleException(ctx, rootReqDuplicator, future, abortCause, initialAttempt);
```
trustin: Is it OK to call `handleException` twice?

core/src/main/java/com/linecorp/armeria/common/stream/AbortingSubscriber.java
```diff
@@ -16,20 +16,19 @@
...
 
-    private static final AbortingSubscriber<Object> INSTANCE = new AbortingSubscriber<>();
+    private final Supplier<? extends Throwable> causeSupplier;
 
-    @SuppressWarnings("unchecked")
-    static <T> AbortingSubscriber<T> get() {
-        return (AbortingSubscriber<T>) INSTANCE;
+    AbortingSubscriber(Supplier<? extends Throwable> causeSupplier) {
+        this.causeSupplier = causeSupplier;
     }
```
trustin: Could we continue using a singleton for `abort()`? We could create a new `AbostringSubscriber` only when a supplier is given.

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -645,19 +665,32 @@ private static void failLateSubscriber(EventExecutor executor,
...
+        }
+
+        @Override
+        public void abort(Throwable cause) {
+            requireNonNull(cause, "cause");
+            abort(() -> cause);
+        }
+
+        @Override
+        public void abort(Supplier<? extends Throwable> causeSupplier) {
```
trustin: For example, we could define `abort0(@Nullable Supplier causeSupplier)` and use the singleton `AbortingSubscriber` when `causeSupplier` is `null`?

core/src/main/java/com/linecorp/armeria/internal/Http2GoAwayHandler.java
```diff
@@ -44,7 +44,7 @@
     private static final long CODE_INTERNAL_ERROR = Http2Error.INTERNAL_ERROR.code();
 
     private boolean goAwaySent;
-    private long goAwayReceived; // -1 if not received, errorCode if received.
+    private long goAwayReceived = -1; // -1 if not received, errorCode if received.
```
trustin: Thanks, @ikhoon!

dependencies.yml
```diff
@@ -217,6 +217,12 @@ javax.inject:
 javax.validation:
   validation-api: { version: '2.0.1.Final' }
 
+org.testng:
+  testng:
+    version: '7.0.0'
+    javadocs:
+      - https://static.javadoc.io/org.testng/testng/7.0.0/
```
trustin: We do not refer to TestNG API in our public Javadoc, so we do not need this.

ikhoon: Thanks for teaching! I will remove this line.

brave/src/test/java/com/linecorp/armeria/it/brave/BraveIntegrationTest.java
ikhoon: Sure, I will check HTTP/1 and add more assertions!

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -717,6 +736,59 @@ void testUpgradeRequestExecutesLogicOnlyOnce() throws Exception {
...
+        final HttpRequestWriter writer = HttpRequest.streaming(HttpMethod.POST, "/slow");
+        final HttpResponse response = client.execute(writer);
+        await().untilAsserted(() -> {
+            assertThat(logHolder.get().isAvailable(RequestLogAvailability.COMPLETE)).isTrue();
+        });
+
+        if (useResponseTimeoutHandler) {
+            assertThat(invokeResponseTimeoutHandler).isTrue();
+            assertThatThrownBy(() -> response.aggregate().join()).isInstanceOf(CompletionException.class)
+                    .hasCauseInstanceOf(ClosedSessionException.class);
```
minwoox: I'm not sure if emitting `ClosedSessionException` is right or not. I think `ResponseTimeoutException` is better than `ClosedSessionException`.
And I think `reqCause` is slightly better than `ResponseTimeoutException`.
WDYT?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -253,6 +255,30 @@ ClientRequestContext newDerivedContext(@Nullable HttpRequest req, @Nullable RpcR
      */
     void setResponseTimeout(Duration responseTimeout);
 
+    /**
+     * Returns {@link Response} timeout handler which is executed when
+     * the {@link Response} is not completely received within the allowed {@link #responseTimeoutMillis()}
+     * or the default {@link ClientOption#RESPONSE_TIMEOUT_MILLIS}.
```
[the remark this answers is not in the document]
trustin: I think it's intentional - to show a user that there are two ways to do the same job.

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
ikhoon: That is a good idea. I will address it :-)

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -717,6 +736,59 @@ void testUpgradeRequestExecutesLogicOnlyOnce() throws Exception {
...
 
+    @ParameterizedTest
+    @CsvSource({
+            "H1C, true", "H1C, false",
+            "H2C, true", "H2C, false"
+    })
+    void testResponseTimeoutHandler(SessionProtocol protocol, boolean useResponseTimeoutHandler) {
+        final AtomicReference<RequestLog> logHolder = new AtomicReference<>();
+        final IllegalStateException reqCause = new IllegalStateException("abort request");
+        final IllegalStateException logCause = new IllegalStateException("log cause");
```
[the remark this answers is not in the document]
ikhoon: Oops! Thanks for the nice catch!

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
```diff
@@ -165,6 +176,20 @@ protected EventExecutor defaultSubscriberExecutor() {
         return completionFuture;
     }
 
+    @Nullable
+    @Override
+    public Throwable completionCause() {
+        return completionCauseSupplier != null ? completionCauseSupplier.get() : null;
+    }
+
+    final boolean completionCause(Supplier<? extends Throwable> causeSupplier) {
```
trustin: `setCompletionCause()`?

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
```diff
@@ -165,6 +176,20 @@ protected EventExecutor defaultSubscriberExecutor() {
         return completionFuture;
     }
 
+    @Nullable
+    @Override
+    public Throwable completionCause() {
+        return completionCauseSupplier != null ? completionCauseSupplier.get() : null;
```
trustin: Could memoize the completion cause so it’s instantiated once. Please make sure that this method is used when creating a `CloseEvent`, too.

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -519,6 +548,12 @@ public boolean isEmpty() {
             return completionFuture;
         }
 
+        @Nullable
+        @Override
+        public Throwable completionCause() {
+            return completionCauseSupplier != null ? completionCauseSupplier.get() : null;
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
ikhoon: Now, it returns `reqCause` instead of `ClosedSessionExcepiton`.

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -133,32 +136,68 @@ final void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEven
...
+                setCompletionCause(CancelledSubscriptionException.INSTANCE);
+                closeEvent = CANCELLED_CLOSE;
+            }
         } else {
-            closeEvent = sampler.isSampled(AbortedStreamException.class) ?
-                         new CloseEvent(new AbortedStreamException()) : ABORTED_CLOSE;
+            setCompletionCause(cause);
+            // cause is always not-null if cancel == false
+            if (cause instanceof AbortedStreamException) {
+                if (cause == AbortedStreamException.INSTANCE) {
```
trustin: Outer `instanceof` is unnecessary.

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -131,19 +132,37 @@ SubscriptionImpl subscribe(SubscriptionImpl subscription) {
...
+    }
+
+    @Override
+    public void abort(Supplier<? extends Throwable> causeSupplier) {
+        requireNonNull(causeSupplier, "causeSupplier");
+        abort0(causeSupplier);
+    }
+
+    private void abort0(Supplier<? extends Throwable> causeSupplier) {
+        final Throwable cause = requireNonNull(causeSupplier.get(),
```
trustin: What is the point of using `Supplier` if you’re going to call `get()` regardless of whether `abort()` will fail?

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -133,32 +136,68 @@ final void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEven
...
+    @Override
+    public final void abort(Supplier<? extends Throwable> causeSupplier) {
+        requireNonNull(causeSupplier, "causeSupplier");
+        abort0(causeSupplier);
+    }
+
+    private void abort0(@Nullable Supplier<? extends Throwable> causeSupplier) {
+        final Supplier<? extends Throwable> causeOrAbortStreamExceptionSupplier =
+                causeSupplier != null ? causeSupplier : AbortedStreamException::get;
+        final Throwable cause = requireNonNull(causeOrAbortStreamExceptionSupplier.get(),
```
trustin: What is the point of using `Supplier` if you’re going to call `get()` regardless of whether `abort()` will fail?

core/src/main/java/com/linecorp/armeria/common/stream/PublisherBasedStreamMessage.java
```diff
@@ -204,21 +212,41 @@ private static void failLateSubscriber(EventExecutor executor,
...
+
+    @Override
+    public void abort(Supplier<? extends Throwable> causeSupplier) {
+        requireNonNull(causeSupplier, "causeSupplier");
+        abort0(causeSupplier);
+    }
+
+    private void abort0(Supplier<? extends Throwable> causeSupplier) {
         final AbortableSubscriber subscriber = this.subscriber;
+        final Throwable cause = requireNonNull(causeSupplier.get(),
```
trustin: What is the point of using `Supplier` if you’re going to call `get()` regardless of whether `abort()` will fail?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -276,6 +277,30 @@ ClientRequestContext newDerivedContext(@Nullable HttpRequest req, @Nullable RpcR
...
+    /**
+     * Sets a handler to run when the response times out. {@code responseTimeoutHandler} must close the request,
+     * e.g., by calling {@link HttpRequestWriter#close(Throwable)}. If not set, the response will be closed with
+     * {@link ResponseTimeoutException}.
+     *
+     * <p>For example,
+     * <pre>{@code
+     *   HttpRequestWriter req = HttpRequest.streaming();
+     *   ctx.setResponseTimeoutHandler(() -> {
+     *      req.close(new IllegalStateException("Server is in a bad state."));
```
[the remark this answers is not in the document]
ikhoon: Actually, I addressed [@anuraaga 's review](https://github.com/line/armeria/pull/2201#discussion_r336851759). If we change this, we should the documentation too. 
```java
{@code responseTimeoutHandler} must close the request, e.g., by calling {@link HttpRequestWriter#close(Throwable)}
```
WDYT? @anuraaga 🤔

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
ikhoon: The reason why I call `get()` here is I thought there is no `abort()` fail in `FixedStreamMessage`.
If I missed something, please let me know 😊

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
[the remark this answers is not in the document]
anuraaga: To be honest I'm not completely sure what the difference is between `HttpRequestWriter.close` and `HttpResponse.abort` for this handler - it's a point that confused me when trying to work on it.

But I figured since our example for `requestTimeoutHandler` in the server uses `HttpResponseWriter.close`, the client is supposed to be the opposite and call `HttpRequestWriter.close`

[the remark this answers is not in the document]
trustin: The difference between `close` and `abort` is that `abort` works even if there is no subscriber. If there's no subscriber, `abort` prevents a new subscription and completes the completion future immediately.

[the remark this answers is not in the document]
anuraaga: So when should a user use `close(Throwable)` vs `abort(Throwable)`? I think we need to be able to express this distinction clearly in Javadoc (probably of those methods themselves) for what seems to accomplish the same thing. Currently, we use `close(Throwable)` for all error handling, like when the connection is interrupted, and I'm not sure why not for timeout. Unless of course it makes sense to remove one of them, then it's simpler :D

ikhoon:
Updated, PTAL 🙏

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
ikhoon: `causeSupplier.get()` is deferred until creating `new CloseEvent(cause)`.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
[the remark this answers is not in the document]
anuraaga: Don't think this was answered - can we make more clear when a user is supposed to use `abort` instead of `close`?

Currently, for server we call close here

https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java#L211
https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java#L330

But is it supposed to use abort?

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -131,19 +129,28 @@ SubscriptionImpl subscribe(SubscriptionImpl subscription) {
...
         }
-
-        final SubscriptionImpl newSubscription = new SubscriptionImpl(
-                this, AbortingSubscriber.get(), ImmediateEventExecutor.INSTANCE, false, false);
-        if (subscriptionUpdater.compareAndSet(this, null, newSubscription)) {
-            // We don't need to invoke onSubscribe() for AbortingSubscriber because it's just a placeholder.
-            invokedOnSubscribe = true;
-        }
-        cancelOrAbort(false);
+        cancelOrAbort(false, causeSupplier);
```
trustin: If you move this subscription logic inside `if (setState(OPEN, CLEANUP)`, it will never be executed if a user closed the stream before a subscription is made, no?

```java
DefaultStreamMessage m = new DefaultStreamMessage();
m.write(...);
m.close();
m.abort();
m.subscribe(...); // Will succeed, but must fail.
```

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -133,32 +133,56 @@ final void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEven
...
+            }
+
+            setCompletionCause(cause);
+            if (cause == AbortedStreamException.INSTANCE) {
+                closeEvent = ABORTED_CLOSE;
+            } else {
+                closeEvent = new CloseEvent(cause);
+            }
         }
         if (closeEventUpdater.compareAndSet(this, null, closeEvent)) {
```
trustin: Is it possible to use the return value of `setCompletionCause()` instead of performing an additional CAS?

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -192,12 +193,24 @@ void logResponseFirstBytesTransferred() {
...
         public void run() {
-            final ResponseTimeoutException cause = ResponseTimeoutException.get();
-            delegate.close(cause);
-            logBuilder.endResponse(cause);
+            final Runnable responseTimeoutHandler;
+            if (logBuilder != RequestLogBuilder.NOOP) {
+                responseTimeoutHandler =
+                        ((ClientRequestContext) ((RequestLog) logBuilder).context()).responseTimeoutHandler();
+            } else {
+               responseTimeoutHandler = null;
```
trustin: nit: indentation

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -192,12 +193,24 @@ void logResponseFirstBytesTransferred() {
 
         @Override
         public void run() {
-            final ResponseTimeoutException cause = ResponseTimeoutException.get();
-            delegate.close(cause);
-            logBuilder.endResponse(cause);
+            final Runnable responseTimeoutHandler;
+            if (logBuilder != RequestLogBuilder.NOOP) {
```
trustin: After thinking a little bit, I thought `!(logBuilder instanceof DefaultRequestLog)` might be a little bit better? :sweat_smile:

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
trustin: Please add a test case for this.

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -172,6 +175,15 @@ default boolean isComplete() {
      */
     CompletableFuture<Void> completionFuture();
 
+    /**
+     * Returns a {@link Throwable} that was captured immediately when abortion or cancellation occurred.
```
trustin: when the stream is aborted, cancelled or completed exceptionally.

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -172,6 +175,15 @@ default boolean isComplete() {
...
+     * Note that this method will return {@code null} when this streams is not complete yet
+     * or complete successfully. An error cause in {@link #completionFuture()} could be obtained only
+     * when this stream is exceptionally complete.
```
trustin: I'm not sure what this sentence ('An error cause ...') means. Could you elaborate?

core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
```diff
@@ -198,9 +203,11 @@ private void doExecute0(ClientRequestContext ctx, HttpRequestDuplicator rootReqD
             } else {
                 final Throwable responseCause =
                         log.isAvailable(RequestLogAvailability.RESPONSE_END) ? log.responseCause() : null;
+                final Runnable closingOriginalResTask =
```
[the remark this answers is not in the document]
ikhoon: I forked the name from argument name of `RetryingHttpClient.handleBackoff`
https://github.com/line/armeria/blob/a2da1905ff38a16b8f1958fbf7b1bdc989c7c050/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java#L237-L244

Let me fix it too :-)

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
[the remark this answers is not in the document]
trustin: I think `abort()` is generally better for handling timeout because it can clean up everything produced so far even if no subscription is made. However, it will not make a difference if you made a subscription before timeout occurs, so the benefit might be subtle.

ikhoon:
We have changed the direction of this PR a few times. We decided to close this because the old reviews of the previous direction are obstructing reviewers.

I will open another PR after solving the following issues from this work:
* Remove `StreamMessage.completionCause()`
* Offer only `StreamMessage.abort(Throwable)` and Remove `StreamMessage.abort(Supply<? extends Throwable>)`
* Call `abort()` if we don't know the cause of `abort` or `close`

PR #2205, opened 2019-10-23
Update Dependencies

minwoox:
- BOM
  - Use Micrometer BOM 1.3.0
- Brave 5.7.0 -> 5.8.0
- Dropwizard 4.1.0 -> 4.1.1
- gax-grpc 1.49.0 -> 1.49.1
- gRPC 1.24.0 -> 1.24.1
- Netty 4.1.42 -> 4.1.43
- org.bouncycastle 1.63 -> 1.64
- Proguard 6.1.1 -> 6.2.0
- Prometheus 0.7.0 -> 0.8.0
- RxJava2 2.2.12 -> 2.2.13
- Spring Boot 2.1.8 -> 2.1.9
- Tomcat
  - 9.0.26 -> 9.0.27
  - 8.5.43 -> 8.5.47
- ZooKeeper 3.5.5 -> 3.5.6
- Build
  - Gradle 5.6.2 -> 5.6.3

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -198,6 +198,7 @@ io.zipkin.brave:
     # ':site:javadoc' fails when we use a newer version of Javadoc.
     javadocs:
     - https://static.javadoc.io/io.zipkin.brave/brave/5.6.3/
+    - https://static.javadoc.io/io.zipkin.brave/brave-instrumentation-http/5.6.3/
 
 it.unimi.dsi:
   fastutil:
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -5,12 +5,12 @@
 boms:
   - com.fasterxml.jackson:jackson-bom:2.10.0
   - io.dropwizard.metrics:metrics-bom:4.1.1
-  - io.grpc:grpc-bom:1.24.0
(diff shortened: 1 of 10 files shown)


dependencies.yml
```diff
@@ -408,7 +406,7 @@ org.slf4j:
 
 org.springframework.boot:
   spring-boot-starter:
-    version: &SPRING_BOOT_VERSION '2.1.8.RELEASE'
+    version: &SPRING_BOOT_VERSION '2.1.9.RELEASE'
```
minwoox: We have decided to update to 2.1.9 because 2.2.0 does not support Tomcat 8.5. So this release(Armeria 0.95.0) will be the last release which supports Tomcat 8.5.

minwoox: /cc @adriancole

adriancole: cool..  thanks for update. also the things available for performance in boot 2.2 are actually spring related and may imply some refactoring as options were added to the `@Configuration` annotation. It may not actually require refactoring, but anyway.

minwoox:
Thanks for reviewing!

PR #2208, opened 2019-10-24
Implement ArmeriaServerHttpRequest.getRemoteAddress()

eunchan-kim:
**Motivation**:
`ArmeriaServerHttpRequest` implements `org.springframework.http.server.reactive.ServerHttpRequest`.
However `ArmeriaServerHttpRequest` does not implement `getRemoteAddress` and relies on default method which returns null.

As a result when using Armeria as an underlying server of Spring WebFlux, we get `Optional.empty` remoteAddress from serverRequest handled by handlerFunction.
For example, suppose there's a routerFunction like below.
```
        @Bean
        public RouterFunction<ServerResponse> route() {
            return RouterFunctions
                    .route(RequestPredicates.GET("/route")
                                            .and(RequestPredicates.accept(MediaType.TEXT_PLAIN)),
                           serverRequest -> {
                                // remoteAddress() returns Optional.empty()
                               Optional<InetSocketAddress> remoteAddress = serverRequest.remoteAddress();
                               ServerResponse.ok().build();
                           });
```

**Modification**:

- Implement `ArmeriaServerHttpRequest.getRemoteAddress`

**Result**:

- `ArmeriaServerHttpRequest` supports `getRemoteAddress` method

### DIFF
--- a/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpRequest.java
+++ b/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpRequest.java
@@ -17,6 +17,7 @@ package com.linecorp.armeria.spring.web.reactive;
 
 import static java.util.Objects.requireNonNull;
 
+import java.net.InetSocketAddress;
 import java.net.URI;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


minwoox:
Thanks a lot, @eunchan-kim! I didn't know that we didn't override `remoteAddress()`. 😅 
I found out that there's another method in `ServerHttpRequest` which is `mutate()`. I think we do not override that but could you talk a look at it as well to make it sure?

eunchan-kim:
Thank you for quick review! 
@minwoox I also think we don't need to override `mutate()`

minwoox:
Great job, @eunchan-kim! 👍

PR #2209, opened 2019-10-24
Remove `RequestContext.newDerivedContext()` method

minwoox:
Motivation:
`newDerivedContext()` method is requested by one of our customers who was not using `RetryingClient`.
He implemented his own retrying client because at that time `Retryingclient` had a bug that uses only one `requestLog` no matter how many retry is made.
Thereafter he could not log the causes and responses successfully.
But right now, we fixed it and it works well. So we do not need anymore.

Modification:
- Remove `RequestContext.newDerivedContext()`

Result:
- (Breaking) You cannot derive a context using `RequestContext.newDerivedContext()`.
  - You should use `(Client|Service)RequestContextBuilder` to make another request context.


minwoox:
> writing his/her own retrying client, etc.

Couldn't think of it. 😅 Let me make a change. 😄

minwoox:
Superseded by #2224

PR #2210, opened 2019-10-24
Fix NPE in Http1ClientCodec

minwoox:
Motivation:
If the remote peer sends multiple responses for one request, which is not allowed by the spec but may still be possible,
NPE can be raised in `Http1ClientCodec`.

Modification:
- Copy `Http1ClientCodec` from the upstream Netty 4.1.42 at 39cc7a673939dec96258ff27f5b1874671838af0

Result:
- No more NPE in `Http1ClientCodec`


ikhoon:
Related: https://github.com/netty/netty/pull/9465

PR #2213, opened 2019-10-24
Fix document

hirakida:
I reflected this change in the document.
https://github.com/line/armeria/pull/2060


minwoox:
Thanks, @hirakida!
Could you also change the same mistake at line 732, please?

hirakida:
> Thanks, @hirakida!
> Could you also change the same mistake at line 732, please?

I did it!

PR #2214, opened 2019-10-25
Deactivate HttpSession in separate handler from codecs.

anuraaga:
Currently, HTTP/1 codec is forked from upstream to allow deactivating sessions in it, and HTTP/2 handler has some API machinery to deactivate sessions there too. We can avoid forking and separating logic by having a single handler that takes care of deactivation.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/Http2ClientConnectionHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Http2ClientConnectionHandler.java
@@ -63,11 +63,6 @@ final class Http2ClientConnectionHandler extends AbstractHttp2ConnectionHandler 
     }
 
     @Override
-    protected void onCloseRequest(ChannelHandlerContext ctx) throws Exception {
-        HttpSession.get(ctx.channel()).deactivate();
-    }
-
-    @Override
     protected boolean needsImmediateDisconnection() {
         return clientFactory.isClosing() || responseDecoder.goAwayHandler().receivedErrorGoAway();
     }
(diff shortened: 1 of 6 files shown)


trustin:
The reason why we set `false` to `active` on `ctx.close()` is to make sure `canSendRequest()` start to return `false` immediately, so that:

- the connection is not chosen from the connection pool when the connection is currently scheduled for disconnection. (See `HttpChannelPool.isHealthy()`)
  - Instead, we could remove the connection completely from the pool on `ctx.close()`, but I guess it'll be more complicated than the current code?
- the request sent between `ctx.close()` and `channelInactive()` is failed immediately with `ClosedSessionException`. (See `HttpRequestSubscriber.writeFirstHeader()`)
  - Maybe we can remove this part because the writes issued after `ctx.close()` will be failed by Netty anyway.

anuraaga:
Thanks! Do you think we can set `active = channel.isWritable` in `channelWritabilityChanged`? So many callbacks not sure exactly their order :) But I feel it should be possible for `HttpSessionHandler` to independently handle this without the help of the codec handlers.

trustin:
> Thanks! Do you think we can set `active = channel.isWritable()` in `channelWritabilityChanged`?

I guess not, because `channelWritabilityChanged` merely means a client is reading slowly?

anuraaga:
Thanks, sorry that was silly. I updated PR to keep the manual deactivation, but moved it to a separate handler. It seems to simplify the code while also removing the need for forking `Http1ClientCodec` (the main objective)

trustin:
I'm somewhat concerned the potential throughput loss due to the increased number of handlers in a Netty pipeline, and thus tend to keep the number of handlers minimal, but maybe that's fine because we already have quite a few handlers. Would you mind checking some benchmarks?

anuraaga:
A powerful benchmark that would probably stress the behavior best (HttpServerBenchmark.empty) kills my laptop 💀 

I agree that compared to the other handlers, it doesn't seem like it'd be much affected, but I can inherit from `ReadSuppressingHandler` to combine them into one - would that make sense?

trustin:
> I can inherit from `ReadSuppressingHandler` to combine them into one - would that make sense?

SGTM

PR #2217, opened 2019-10-30
Add RefreshingAddressResolver

### LINKED ISSUE #2017: Do not block a request when DNS resolution takes long time
We found that an issue in our DNS servers can lead to the increased total duration of requests. Instead of waiting for the DNS resolution result, we could initiate the DNS resolution in background and reuse the previous DNS resolution result (if available). If the previous DNS resolution result is not available, we will still have to wait until the DNS resolution is finished, but it will not usually be a problem because it affects only the hosts that appeared first.

We might need to wait for #1886 because a merge conflict is expected.

minwoox:
Motivation:
When the DNS server is not responsive, it takes a long time to get DNS addresses and it leads to the increased total duration of requests.
Instead of waiting for the DNS resolution result, we could initiate the DNS resolution in the background and reuse the previous DNS resolution result if available.
Also, we can share the DNS cache so that all `EventLoop`s do not have to send the DNS query for the same host.

Modifications:
- Add `RefreshingAddressResolver` which updates DNS caches automatically.
  - The `Resolver` refreshes DNS caches automatically in the background.

Result:
- Close #2017
- You may no longer see the increased total duration of requests due to [dnsResolutionDurationNanos](http://line.github.io/armeria/advanced-structured-logging.html#client-connection-timing-properties).
- (Breaking) `ClientFactoryBuilder.domainNameResolverCustomizer()` now takes `Consumer<? super DnsResolverGroupBuilder>`.
  - It was `Consumer<? super DnsNameResolverBuilder>`. `DnsResolverGroupBuilder` has additional properties for the refreshing.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
@@ -1,0 +1,370 @@
+/*
+ *  Copyright 2019 LINE Corporation
+ *
+ *  LINE Corporation licenses this file to you under the Apache License,
+ *  version 2.0 (the "License"); you may not use this file except in compliance
+ *  with the License. You may obtain a copy of the License at:
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  License for the specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package com.linecorp.armeria.client;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.linecorp.armeria.internal.dns.DnsUtil.extractAddressBytes;
+import static java.util.Objects.requireNonNull;
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
+import javax.annotation.Nullable;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
+
+import com.linecorp.armeria.client.retry.Backoff;
+import com.linecorp.armeria.internal.dns.DefaultDnsNameResolver;
+import com.linecorp.armeria.internal.dns.DnsQuestionWithoutTrailingDot;
+
+import io.netty.channel.EventLoop;
+import io.netty.handler.codec.dns.DnsQuestion;
+import io.netty.handler.codec.dns.DnsRecord;
+import io.netty.handler.codec.dns.DnsRecordType;
+import io.netty.resolver.AbstractAddressResolver;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.Promise;
+
+class AutoUpdatingAddressResolver extends AbstractAddressResolver<InetSocketAddress> {
+
+    private static final Logger logger = LoggerFactory.getLogger(AutoUpdatingAddressResolver.class);
+
+    @SuppressWarnings("rawtypes")
+    private static final AtomicReferenceFieldUpdater<CacheEntry, ScheduledFuture> futureUpdater =
+            AtomicReferenceFieldUpdater.newUpdater(CacheEntry.class, ScheduledFuture.class,
+                                                   "cacheUpdatingScheduledFuture");
+    @VisibleForTesting
+    static final ScheduledFuture<?> closed = new ScheduledFuture<Object>() {
+
+        @Override
+        public boolean cancel(boolean mayInterruptIfRunning) {
+            return false;
+        }
+
+        @Override
+        public long getDelay(TimeUnit unit) {
+            return Long.MIN_VALUE;
+        }
+
+        @Override
+        public int compareTo(Delayed o) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean isCancelled() {
+            return true;
+        }
+
+        @Override
+        public boolean isDone() {
+            return true;
+        }
+
+        @Override
+        public Object get() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public Object get(long timeout, TimeUnit unit) {
+            throw new UnsupportedOperationException();
+        }
+    };
+
+    private final ConcurrentMap<String, CompletableFuture<CacheEntry>> cache;
+    private final DefaultDnsNameResolver resolver;
+    private final List<DnsRecordType> dnsRecordTypes;
+    private final int minTtl;
+    private final int maxTtl;
+    private final Backoff autoUpdateBackoff;
+    private final long autoUpdateTimeoutMillis;
+
+    private volatile boolean resolverClosed;
+
+    protected AutoUpdatingAddressResolver(EventLoop eventLoop,
+                                          ConcurrentMap<String, CompletableFuture<CacheEntry>> cache,
+                                          DefaultDnsNameResolver resolver, List<DnsRecordType> dnsRecordTypes,
+                                          int minTtl, int maxTtl, Backoff autoUpdateBackoff,
+                                          long autoUpdateTimeoutMillis) {
+        super(eventLoop);
+        this.cache = cache;
+        this.resolver = resolver;
+        this.dnsRecordTypes = dnsRecordTypes;
+        this.minTtl = minTtl;
+        this.maxTtl = maxTtl;
+        this.autoUpdateBackoff = autoUpdateBackoff;
+        this.autoUpdateTimeoutMillis = autoUpdateTimeoutMillis;
+    }
+
+    @Override
+    protected boolean doIsResolved(InetSocketAddress address) {
+        return !requireNonNull(address, "address").isUnresolved();
+    }
+
+    @Override
+    protected void doResolve(InetSocketAddress unresolvedAddress, Promise<InetSocketAddress> promise)
+            throws Exception {
+        requireNonNull(unresolvedAddress, "unresolvedAddress");
+        requireNonNull(promise, "promise");
+        if (resolverClosed) {
+            promise.tryFailure(new IllegalStateException("resolver is already closed."));
+            return;
+        }
+        final String hostname = unresolvedAddress.getHostName();
+        final int port = unresolvedAddress.getPort();
+        final CompletableFuture<CacheEntry> entryFuture = cache.get(hostname);
+        if (entryFuture != null) {
+            handle(entryFuture, promise, port);
+            return;
+        }
+
+        final CompletableFuture<CacheEntry> result = new CompletableFuture<>();
+        final CompletableFuture<CacheEntry> previous = cache.putIfAbsent(hostname, result);
+        if (previous != null) {
+            handle(previous, promise, port);
+            return;
+        }
+
+        final List<DnsQuestion> questions =
+                dnsRecordTypes.stream()
+                              .map(type -> DnsQuestionWithoutTrailingDot.of(hostname, type))
+                              .collect(toImmutableList());
+        sendQueries(questions, hostname, result);
+        result.handle((entry, cause) -> {
+            if (cause != null) {
+                cache.remove(hostname);
+                promise.tryFailure(cause);
+                return null;
+            }
+            promise.trySuccess(new InetSocketAddress(entry.address(), port));
+            return null;
+        });
+    }
+
+    private static void handle(CompletableFuture<CacheEntry> future, Promise<InetSocketAddress> promise,
+                               int port) {
+        future.handle((entry, cause) -> {
+            if (cause != null) {
+                promise.tryFailure(cause);
+                return null;
+            }
+            entry.incrementCacheHit();
+            promise.trySuccess(new InetSocketAddress(entry.address(), port));
+            return null;
+        });
+    }
+
+    private void sendQueries(List<DnsQuestion> questions, String hostname,
+                             CompletableFuture<CacheEntry> result) {
+        final Future<List<DnsRecord>> recordsFuture = resolver.sendQueries(questions, hostname);
+        recordsFuture.addListener(f -> {
+            if (!f.isSuccess()) {
+                result.completeExceptionally(f.cause());
+                return;
+            }
+
+            @SuppressWarnings("unchecked")
+            final List<DnsRecord> records = (List<DnsRecord>) f.getNow();
+            InetAddress inetAddress = null;
+            long ttlMillis = -1;
+            try {
+                for (DnsRecord r : records) {
+                    final byte[] addrBytes = extractAddressBytes(r, logger, hostname);
+                    if (addrBytes == null) {
+                        continue;
+                    }
+                    try {
+                        inetAddress = InetAddress.getByAddress(hostname, addrBytes);
+                        ttlMillis = TimeUnit.SECONDS.toMillis(
+                                Math.max(Math.min(r.timeToLive(), maxTtl), minTtl));
+                        break;
+                    } catch (UnknownHostException e) {
+                        // Should never reach here because we already validated it in extractAddressBytes.
+                        result.completeExceptionally(
+                                new IllegalArgumentException("Invalid address: " + hostname, e));
+                        return;
+                    }
+                }
+            } finally {
+                records.forEach(ReferenceCountUtil::safeRelease);
+            }
+
+            if (inetAddress == null) {
+                result.completeExceptionally(
+                        new UnknownHostException("failed to receive DNS records for " + hostname));
+                return;
+            }
+            final CacheEntry entry = new CacheEntry(inetAddress, ttlMillis, questions);
+            result.complete(entry);
+            entry.scheduleCacheUpdate(ttlMillis);
+        });
+    }
+
+    @Override
+    protected void doResolveAll(InetSocketAddress unresolvedAddress, Promise<List<InetSocketAddress>> promise)
+            throws Exception {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void close() {
+        resolverClosed = true;
+        resolver.close();
+    }
+
+    final class CacheEntry implements Runnable {
+
+        private final InetAddress address;
+        private final long ttlMillis;
+        private final List<DnsQuestion> questions;
+        private final long deadlineNanos;
+
+        private int numAttemptsSoFar = 1; // Updated only by the executor().
+
+        private volatile int cacheHits;
+
+        @Nullable
+        volatile ScheduledFuture<?> cacheUpdatingScheduledFuture;
+
+        CacheEntry(InetAddress address, long ttlMillis, List<DnsQuestion> questions) {
+            this.address = address;
+            this.ttlMillis = ttlMillis;
+            this.questions = questions;
+
+            if (autoUpdateTimeoutMillis <= 0 || autoUpdateTimeoutMillis == Long.MAX_VALUE) {
+                deadlineNanos = 0;
+            } else {
+                deadlineNanos = System.nanoTime() +
+                                TimeUnit.MILLISECONDS.toNanos(ttlMillis + autoUpdateTimeoutMillis);
+            }
+        }
+
+        void incrementCacheHit() {
+            cacheHits++;
+        }
+
+        InetAddress address() {
+            return address;
+        }
+
+        long ttlMillis() {
+            return ttlMillis;
+        }
+
+        void scheduleCacheUpdate(long nextDelayMillis) {
+            for (;;) {
+                if (resolverClosed) {
+                    return;
+                }
+                final ScheduledFuture<?> oldFuture = futureUpdater.get(this);
+                if (oldFuture == closed) {
+                    return;
+                }
+
+                final ScheduledFuture<?> newFuture =
+                        executor().schedule(this, nextDelayMillis, TimeUnit.MILLISECONDS);
+                if (futureUpdater.compareAndSet(this, oldFuture, newFuture)) {
+                    return;
+                } else {
+                    newFuture.cancel(true);
+                }
+            }
+        }
+
+        void clear() {
+            assert resolverClosed;
+            final ScheduledFuture<?> future = futureUpdater.getAndSet(this, closed);
+            if (future != null) {
+                future.cancel(false);
+            }
+        }
+
+        @Override
+        public void run() {
+            if (resolverClosed) {
+                return;
+            }
+
+            final String hostName = address.getHostName();
+            if (cacheHits <= 0) { // Should we increase 0 to N?
+                cache.remove(hostName);
+                return;
+            }
+
+            final CompletableFuture<CacheEntry> result = new CompletableFuture<>();
+            sendQueries(questions, hostName, result);
+            result.handle((entry, cause) -> {
+                if (resolverClosed) {
+                    return null;
+                }
+
+                if (cause != null) {
+                    if (deadlineNanos == 0) {
+                        final long nextDelayMillis = autoUpdateBackoff.nextDelayMillis(numAttemptsSoFar++);
+                        scheduleCacheUpdate(nextDelayMillis);
+                        return null;
+                    }
+                    final long timeoutMillis = TimeUnit.NANOSECONDS.toMillis(deadlineNanos - System.nanoTime());
+                    if (timeoutMillis <= 0) {
+                        cache.remove(hostName);
+                        return null;
+                    }
+                    final long nextDelayMillis = autoUpdateBackoff.nextDelayMillis(numAttemptsSoFar++);
+                    scheduleCacheUpdate(Math.min(timeoutMillis, nextDelayMillis));
+                    return null;
+                }
+
+                // Got the response successfully so reset the state.
+                cacheHits = 0;
+                numAttemptsSoFar = 1;
+
+                if (entry.address().equals(address) && entry.ttlMillis() == ttlMillis) {
+                    scheduleCacheUpdate(ttlMillis);
+                } else {
+                    // Replace the old entry with the new one.
+                    cache.put(hostName, result);
+                    entry.scheduleCacheUpdate(entry.ttlMillis());
+                }
+                return null;
+            });
+        }
+
+        @Override
+        public String toString() {
+            return MoreObjects.toStringHelper(this)
+                              .add("address", address)
+                              .add("ttlMillis", ttlMillis)
+                              .add("questions", questions)
+                              .add("cacheHits", cacheHits)
+                              .toString();
+        }
+    }
+}
--- a/core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
@@ -60,7 +60,14 @@ class AutoUpdatingAddressResolver extends AbstractAddressResolver<InetSocketAddr
     private static final AtomicReferenceFieldUpdater<CacheEntry, ScheduledFuture> futureUpdater =
             AtomicReferenceFieldUpdater.newUpdater(CacheEntry.class, ScheduledFuture.class,
                                                    "cacheUpdatingScheduledFuture");
+
+    /**
+     * A {@link ScheduledFuture} which is set in {@link CacheEntry} when the
+     * {@link AutoUpdatingAddressResolver} is closed by replacing and cancelling previously scheduled automatic
+     * DNS cache updating.
+     */
     @VisibleForTesting
+    @SuppressWarnings("ComparableImplementedButEqualsNotOverridden")
     static final ScheduledFuture<?> closed = new ScheduledFuture<Object>() {
 
         @Override
@@ -109,11 +116,11 @@ class AutoUpdatingAddressResolver extends AbstractAddressResolver<InetSocketAddr
 
     private volatile boolean resolverClosed;
 
-    protected AutoUpdatingAddressResolver(EventLoop eventLoop,
-                                          ConcurrentMap<String, CompletableFuture<CacheEntry>> cache,
-                                          DefaultDnsNameResolver resolver, List<DnsRecordType> dnsRecordTypes,
-                                          int minTtl, int maxTtl, Backoff autoUpdateBackoff,
-                                          long autoUpdateTimeoutMillis) {
+    AutoUpdatingAddressResolver(EventLoop eventLoop,
+                                ConcurrentMap<String, CompletableFuture<CacheEntry>> cache,
+                                DefaultDnsNameResolver resolver, List<DnsRecordType> dnsRecordTypes,
+                                int minTtl, int maxTtl, Backoff autoUpdateBackoff,
+                                long autoUpdateTimeoutMillis) {
         super(eventLoop);
         this.cache = cache;
         this.resolver = resolver;
--- a/core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
@@ -135,10 +135,10 @@ class AutoUpdatingAddressResolver extends AbstractAddressResolver<InetSocketAddr
         requireNonNull(unresolvedAddress, "unresolvedAddress");
         requireNonNull(promise, "promise");
         if (resolverClosed) {
-            promise.tryFailure(new IllegalStateException("resolver is already closed."));
+            promise.tryFailure(new IllegalStateException("resolver is closed already."));
             return;
         }
-        final String hostname = unresolvedAddress.getHostName();
+        final String hostname = unresolvedAddress.getHostString();
         final int port = unresolvedAddress.getPort();
         final CompletableFuture<CacheEntry> entryFuture = cache.get(hostname);
         if (entryFuture != null) {
@@ -247,7 +247,10 @@ class AutoUpdatingAddressResolver extends AbstractAddressResolver<InetSocketAddr
         private final List<DnsQuestion> questions;
         private final long deadlineNanos;
 
-        private int numAttemptsSoFar = 1; // Updated only by the executor().
+        /**
+         * No need to be volatile because updated only by the {@link #executor()}.
+         */
+        private int numAttemptsSoFar = 1;
 
         private volatile int cacheHits;
 
--- a/core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolverGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolverGroup.java
@@ -1,0 +1,174 @@
+/*
+ *  Copyright 2019 LINE Corporation
+ *
+ *  LINE Corporation licenses this file to you under the Apache License,
+ *  version 2.0 (the "License"); you may not use this file except in compliance
+ *  with the License. You may obtain a copy of the License at:
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  License for the specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package com.linecorp.armeria.client;
+
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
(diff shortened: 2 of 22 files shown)


minwoox:
The `Resolver` updates a DNS cache automatically in the background if the initial query was successful. (i.e. It does not try if the initial attempt failed.)
I'm not sure whether we still need to send queries in the background in that case. 
/cc @trustin, @ikhoon, @delegacy

trustin:
> I'm not sure whether we still need to send queries in the background if the initial attempt failed.

I think it's OK not to send background queries on initial failure, because it's probably rare.

thrift/src/test/java/com/linecorp/armeria/it/metric/DropwizardMetricsIntegrationTest.java
```diff
@@ -94,8 +94,6 @@ public void normal() throws Exception {
             assertThat(dropwizardRegistry.getMeters()
                                          .get(clientMetricNameWithStatusAndResult("requests", 200, "failure"))
                                          .getCount()).isEqualTo(3L);
-            System.err.println(dropwizardRegistry.getMeters());
-            System.err.println(serverMetricNameWithStatusAndResult("requests", 200, "failure"));
```
trustin: Thanks!

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolverGroupBuilder.java
```diff
@@ -0,0 +1,348 @@
...
+ * <p>{@link AddressResolver} keeps cache hits so that it does not automatically update the cache for
+ * the DNS addresses used only once.
+ */
+public final class AutoUpdatingAddressResolverGroupBuilder {
+
+    private Backoff autoUpdateBackoff = Backoff.ofDefault();
+
+    private long autoUpdateTimeoutMillis = TimeUnit.MINUTES.toMillis(10);
+
+    private int minTtl;
```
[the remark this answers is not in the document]
minwoox: Ah, thanks! At first, the default value is `0`. But I noticed that `DnsEndpointGroupBuilder` has `1` as the default value, so I changed the value to `1` (in my mind only I guess). 😄

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
```diff
@@ -0,0 +1,370 @@
...
+    @Override
+    protected void doResolve(InetSocketAddress unresolvedAddress, Promise<InetSocketAddress> promise)
+            throws Exception {
+        requireNonNull(unresolvedAddress, "unresolvedAddress");
+        requireNonNull(promise, "promise");
+        if (resolverClosed) {
+            promise.tryFailure(new IllegalStateException("resolver is already closed."));
+            return;
+        }
+        final String hostname = unresolvedAddress.getHostName();
```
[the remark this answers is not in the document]
minwoox: Oops, thanks!

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    final class CacheEntry implements Runnable {
+
+        private final InetAddress address;
+        private final long ttlMillis;
+        private final List<DnsQuestion> questions;
+        private final long deadlineNanos;
+
+        private int numAttemptsSoFar = 1; // Updated only by the executor().
```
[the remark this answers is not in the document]
minwoox: Fixed, thanks!

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolverGroup.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    @Override
+    protected AddressResolver<InetSocketAddress> newResolver(EventExecutor executor) throws Exception {
+        assert executor instanceof EventLoop;
+        final EventLoop eventLoop = (EventLoop) executor;
+        final DnsNameResolverBuilder builder = new DnsNameResolverBuilder(eventLoop);
+        resolverConfigurator.accept(builder);
+        final DefaultDnsNameResolver resolver = new DefaultDnsNameResolver(builder.build(), eventLoop);
+        return new AutoUpdatingAddressResolver(eventLoop, cache, resolver, dnsRecordTypes, minTtl, maxTtl,
```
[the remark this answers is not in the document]
minwoox: It's because I wanted to make `List<DnsRecordType>` once and share it with the resolvers. 😄

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolverGroupBuilder.java
```diff
@@ -0,0 +1,348 @@
...
+ * Builds an {@link AddressResolverGroup} which builds {@link AddressResolver}s that update DNS cache
+ * automatically. An usual DNS resolver such as {@link DnsNameResolver}, cache is expired after TTL.
+ * On the other hand, the automatic updating {@link AddressResolver} updates the DNS cache spontaneously
+ * even after TTL has elapsed. If automatic updating fails, the {@link AddressResolver} will retry with
+ * {@link #autoUpdateBackoff(Backoff)} until {@link #autoUpdateTimeoutMillis(long)} has elapsed.
+ *
+ * <p>{@link AddressResolver} keeps cache hits so that it does not automatically update the cache for
+ * the DNS addresses used only once.
+ */
+public final class AutoUpdatingAddressResolverGroupBuilder {
```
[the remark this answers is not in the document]
minwoox: That's a brilliant idea. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -261,37 +264,66 @@ public ClientFactoryBuilder sslContextCustomizer(Consumer<? super SslContextBuil
...
     /**
      * Adds the specified {@link Consumer} which customizes the given {@link DnsNameResolverBuilder}.
      * This method is useful when you want to change the behavior of the default domain name resolver, such as
      * changing the DNS server list.
      *
-     * @throws IllegalStateException if {@link #addressResolverGroupFactory(Function)} was called already.
+     * @throws IllegalStateException if {@link #autoUpdatingAddressResolverCustomizer(Consumer)} or
+     *                               {@link #addressResolverGroupFactory(Function)} was called already.
      */
     public ClientFactoryBuilder domainNameResolverCustomizer(
```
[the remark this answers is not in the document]
minwoox: Fixed. I didn't change the method name though. 😄

thrift/src/test/java/com/linecorp/armeria/it/metric/DropwizardMetricsIntegrationTest.java
minwoox: 😄

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
```diff
@@ -0,0 +1,373 @@
...
+class AutoUpdatingAddressResolver extends AbstractAddressResolver<InetSocketAddress> {
+
+    private static final Logger logger = LoggerFactory.getLogger(AutoUpdatingAddressResolver.class);
+
+    @SuppressWarnings("rawtypes")
+    private static final AtomicReferenceFieldUpdater<CacheEntry, ScheduledFuture> futureUpdater =
+            AtomicReferenceFieldUpdater.newUpdater(CacheEntry.class, ScheduledFuture.class,
+                                                   "cacheUpdatingScheduledFuture");
+    @VisibleForTesting
+    static final ScheduledFuture<?> closed = new ScheduledFuture<Object>() {
```
trustin: How about adding Javadoc for this field, to explain what this is used for?

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
```diff
@@ -0,0 +1,373 @@
...
+    private final DefaultDnsNameResolver resolver;
+    private final List<DnsRecordType> dnsRecordTypes;
+    private final int minTtl;
+    private final int maxTtl;
+    private final Backoff autoUpdateBackoff;
+    private final long autoUpdateTimeoutMillis;
+
+    private volatile boolean resolverClosed;
+
+    protected AutoUpdatingAddressResolver(EventLoop eventLoop,
```
trustin: Perhaps OK to remove `protected`?

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
```diff
@@ -0,0 +1,348 @@
...
+import io.netty.resolver.dns.DnsNameResolver;
+import io.netty.resolver.dns.DnsNameResolverBuilder;
+import io.netty.resolver.dns.DnsQueryLifecycleObserverFactory;
+import io.netty.resolver.dns.DnsServerAddressStreamProvider;
+import io.netty.resolver.dns.NoopAuthoritativeDnsServerCache;
+import io.netty.resolver.dns.NoopDnsCache;
+
+/**
+ * Builds an {@link AddressResolverGroup} which builds {@link AddressResolver}s that update DNS caches
+ * automatically. An usual DNS resolver such as {@link DnsNameResolver}, a cache is expired after TTL.
```
trustin: A usual - https://english.stackexchange.com/a/8863

core/src/main/java/com/linecorp/armeria/internal/dns/DefaultDnsNameResolver.java
```diff
@@ -0,0 +1,113 @@
...
+        final int numQuestions = questions.size();
+        if (numQuestions == 1) {
+            // Simple case of single query
+            final DnsQuestion question = questions.get(0);
+            logger.debug("[{}] Sending a DNS query", logPrefix);
+            return delegate.resolveAll(question);
+        }
+
+        // Multiple queries
+        logger.debug("{} Sending DNS queries", logPrefix);
```
trustin: - `[{}]` ?
- Could log questions

core/src/main/java/com/linecorp/armeria/internal/dns/DefaultDnsNameResolver.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public Future<List<DnsRecord>> sendQueries(List<DnsQuestion> questions, String logPrefix) {
+        requireNonNull(questions, "questions");
+        requireNonNull(logPrefix, "logPrefix");
+        final int numQuestions = questions.size();
+        if (numQuestions == 1) {
+            // Simple case of single query
+            final DnsQuestion question = questions.get(0);
+            logger.debug("[{}] Sending a DNS query", logPrefix);
```
trustin: - Could log a question?

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
minwoox: Added. 😄

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
minwoox: Oops, Thank!

core/src/main/java/com/linecorp/armeria/internal/dns/DefaultDnsNameResolver.java
minwoox: Added Thanks!

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
```diff
@@ -0,0 +1,380 @@
...
+        final String hostname = unresolvedAddress.getHostString();
+        final int port = unresolvedAddress.getPort();
+        final CompletableFuture<CacheEntry> entryFuture = cache.get(hostname);
+        if (entryFuture != null) {
+            handle(entryFuture, promise, port);
+            return;
+        }
+
+        final CompletableFuture<CacheEntry> result = new CompletableFuture<>();
+        final CompletableFuture<CacheEntry> previous = cache.putIfAbsent(hostname, result);
```
ikhoon: I think DNS cache hit rate is very high. How about checking `previous` with `get` first to avoid additional allocation of `result`. 
Or we can split cache into result cache and pending work to concurrency. Does it seem overkill? WDYT? 😀

minwoox: It's up there. 😄

anuraaga: I think caffeine provides a high performance, asynchronous cache that takes care of a lot of this logic (configuring it to refresh on interval is exactly the behavior we want of returning stale value while resolving new one). Have you thought of using it?

minwoox: Every cache has to have a different expiration timeout and it's very inconvenient to invalidate the cache. (Have to call `synchronous()`). And There were some other situations that make me hard to use it. So I decided not to use that. 😄

anuraaga: I see - yeah it seems like it's missing value based refresh even though it has value based expiry. Filed issue - https://github.com/ben-manes/caffeine/issues/360

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
```diff
@@ -0,0 +1,380 @@
...
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public Object get(long timeout, TimeUnit unit) {
+            throw new UnsupportedOperationException();
+        }
+    };
+
+    private final ConcurrentMap<String, CompletableFuture<CacheEntry>> cache;
```
anuraaga: Does this need to be concurrent? I think resolution happens on a single event loop.

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
```diff
@@ -0,0 +1,380 @@
...
+import io.netty.channel.EventLoop;
+import io.netty.handler.codec.dns.DnsQuestion;
+import io.netty.handler.codec.dns.DnsRecord;
+import io.netty.handler.codec.dns.DnsRecordType;
+import io.netty.resolver.AbstractAddressResolver;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.Promise;
+
+class AutoUpdatingAddressResolver extends AbstractAddressResolver<InetSocketAddress> {
```
anuraaga: RefreshingAddressResolver sounds more natural tome

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
minwoox: Yes. 😄 The cache is created in `AutoUpdatingAddressResolverGroup` and shared across from all `DnsResolver`s so that only one `EventLoop` sends the query for the hostname.

core/src/main/java/com/linecorp/armeria/client/AutoUpdatingAddressResolver.java
minwoox: Thanks! I will take it. 🙇

minwoox:
@anuraaga All fixed. PTAL 🙇

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,380 @@
...
+        });
+    }
+
+    @Override
+    protected void doResolveAll(InetSocketAddress unresolvedAddress, Promise<List<InetSocketAddress>> promise)
+            throws Exception {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
```
anuraaga: Do you know when newResolver and close are called? It would help understanding the lifecycle of the refresh scheduler

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolverGroup.java
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+
+        defaultDnsRecordTypes = dnsRecordTypes(resolvedAddressTypes);
+    }
+
+    /**
+     * Returns {@code true} if any {@link NetworkInterface} supports {@code IPv6}, {@code false} otherwise.
+     */
+    private static boolean anyInterfaceSupportsIpV6() {
```
anuraaga: I remember when this was added to Netty ;) Could you mark any code copied from Netty? We should work on making them public or otherwise avoiding keeping copied code forever and missing out on upstream bug fixes. I think a lot of our DNS code is copied so we can probably help improve the abstraction in Netty itself.

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
```diff
@@ -0,0 +1,344 @@
...
+    public DnsResolverGroupBuilder refreshBackoff(Backoff refreshBackoff) {
+        this.refreshBackoff = requireNonNull(refreshBackoff, "refreshBackoff");
+        return this;
+    }
+
+    /**
+     * Sets the timeout of refreshing DNS caches. After this timeout, the previously resolved
+     * DNS cache is invalidated.
+     */
+    public DnsResolverGroupBuilder refreshTimeout(Duration refreshTimeout) {
```
anuraaga: Is this needed on top of the backoff's max attempts?

minwoox: I think so. Because there's jitter which makes the user hard to calculate the whole duration, I thought having `refreshTimeout` is better so the cache stays until TTL + `refreshTimeout`. Do you think we need to replace it or add  `max attempts`?

anuraaga: `refreshTimeout` seems to be relatively unscientific anyways (default is a large 10 min) so not sure if it matters so much whether the duration is precise with respect to jitter. On the flip side, the interaction between max attempts and refreshTimeout is confusing since two different variables affect the same behavior.

I feel that if we do want to add a hard max deadline on retries, it's better if we add it as another parameter `Backoff` than here, so at least it's kept together with max attempts. I think the similar concern can apply to other uses of `Backoff`.

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: Let me add a comment. :)

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolverGroup.java
minwoox: Right. 😃 Let me see what I can do. Thanks!

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
minwoox: Ah, I forgot about that we have `Backoff.withMaxAttempts()` so we might not need to add it explicitly. 🤣 So we can just remove the `refreshTimeout` here if we don't need it.

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
```diff
@@ -0,0 +1,344 @@
...
+        return this;
+    }
+
+    /**
+     * Sets the cache for {@code CNAME} mappings.
+     */
+    public DnsResolverGroupBuilder cnameCache(DnsCnameCache cnameCache) {
+        this.cnameCache = requireNonNull(cnameCache, "cnameCache");
+        return this;
+    }
```
trustin: Question: Do we need to expose this property? Could you explain why we allow a user to specify this while disabling other caches?

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,363 @@
...
+import io.netty.channel.EventLoop;
+import io.netty.handler.codec.dns.DnsQuestion;
+import io.netty.handler.codec.dns.DnsRecord;
+import io.netty.handler.codec.dns.DnsRecordType;
+import io.netty.resolver.AbstractAddressResolver;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.Promise;
+
+class RefreshingAddressResolver extends AbstractAddressResolver<InetSocketAddress> {
```
trustin: nit: final?

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
minwoox: Removed. 😃

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
```diff
@@ -0,0 +1,317 @@
...
+import io.netty.resolver.dns.DnsNameResolverBuilder;
+import io.netty.resolver.dns.DnsQueryLifecycleObserverFactory;
+import io.netty.resolver.dns.DnsServerAddressStreamProvider;
+import io.netty.resolver.dns.NoopAuthoritativeDnsServerCache;
+import io.netty.resolver.dns.NoopDnsCache;
+import io.netty.resolver.dns.NoopDnsCnameCache;
+
+/**
+ * Builds an {@link AddressResolverGroup} which builds {@link AddressResolver}s that update DNS caches
+ * automatically. A usual DNS resolver such as {@link DnsNameResolver}, a cache is expired after TTL.
```
anuraaga: `Standard {@link DnsNameResolver} will only expire a cache entry after TTL, meaning DNS queries after TTL will always take time to resolve. A refreshing {@link AddressResolver} on the other hand updates the DNS cache automatically when TTL elapses, meaning DNS queries after TTL will retrieve a refreshed result right away.`

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
```diff
@@ -0,0 +1,317 @@
...
+ * On the other hand, the refreshing {@link AddressResolver} updates the DNS cache spontaneously
+ * even after TTL has elapsed. If refreshing fails, the {@link AddressResolver} will retry with
+ * {@link #refreshBackoff(Backoff)}.
+ *
+ * <p>{@link AddressResolver} keeps cache hits so that it does not refresh the caches for
```
anuraaga: I'm not sure I understand this sentence, is it `The refreshing {@link AddressResolver} will only start auto refresh for a given hostname on the second access before TTL to avoid auto-refreshing for queries that only happen once (e.g., requests during server startup).`

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
```diff
@@ -0,0 +1,317 @@
...
+    public DnsResolverGroupBuilder ttl(int minTtl, int maxTtl) {
+        checkArgument(minTtl > 0 && minTtl <= maxTtl,
+                      "minTtl: %s, maxTtl: %s (expected: 1 <= minTtl <= maxTtl)", minTtl, maxTtl);
+        this.minTtl = minTtl;
+        this.maxTtl = maxTtl;
+        return this;
+    }
+
+    /**
+     * Sets if this resolver should generate the detailed trace information in an exception message so that
```
anuraaga: `generate detailed trace information in exception messages`

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,371 @@
...
+                    return;
+                }
+                final ScheduledFuture<?> oldFuture = futureUpdater.get(this);
+                if (oldFuture == closed) {
+                    return;
+                }
+
+                final ScheduledFuture<?> newFuture =
+                        executor().schedule(this, nextDelayMillis, TimeUnit.MILLISECONDS);
+                if (futureUpdater.compareAndSet(this, oldFuture, newFuture)) {
```
anuraaga: I think `scheduleCacheUpdate` is also only possible from the `executor()` same as `attemptsSoFar` since it should be called from the same thread as `doResolve` - do we need the updater?

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,371 @@
...
+        private final InetAddress address;
+        private final long ttlMillis;
+        private final List<DnsQuestion> questions;
+
+        /**
+         * No need to be volatile because updated only by the {@link #executor()}.
+         */
+        private int numAttemptsSoFar = 1;
+
+        private volatile int cacheHits;
```
anuraaga: I think making this `boolean servedFromCache` or similar is clearer - usually using `++` on a `volatile int` isn't threadsafe. But here it's not really a bug since we only care if the value is 0 or not - so we can model that better with a boolean.

anuraaga: Though I do see we have a comment considering whether the threshold should be configurable - if we're expecting it, then we should make this `AtomicInteger`

minwoox: Yes, I was going to change it to `AtomicInteger` if we have to. Let me use `boolean servedFromCache`. Thanks! 😉

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
minwoox: This is the sentence I wanted to write. 😉 Thanks a lot. 🙇

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: `scheduleCacheUpdate` can be set by another thread when `clear()` is called.
So I think we need the updater but do not need the `for` loop. I was a little lazy to remove that. 😜 Let me fix it.

minwoox:
@anuraaga All fixed. PTAL 😉

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,369 @@
...
+                    executor().schedule(this, nextDelayMillis, TimeUnit.MILLISECONDS);
+            if (!futureUpdater.compareAndSet(this, oldFuture, newFuture)) {
+                // clear() is called and the future is set to closed future. So we just cancel the newFuture
+                // we just made.
+                newFuture.cancel(true);
+            }
+        }
+
+        void clear() {
+            assert resolverClosed;
```
anuraaga: I think clear is only called after closing the resolver and I guess this assert agrees. Since we don't reschedule or process a response if the resolver is closed, I still feel like we shouldn't need the updater for the future. Closing the client factory to close the resolvers is rare, so if we want to cancel the future I think we can just schedule cancellation on executor () instead of making the standard case of updating more complicated by the updater. I might not even worry about cancel since even if the future runs it will finish in microseconds, doing nothing since the resolver is closed.

Maybe the summary is that since we already have one volatile boolean (resolverClosed) which we check everywhere, we probably don't need another one.

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
```diff
@@ -0,0 +1,319 @@
...
+
+    private boolean traceEnabled = true;
+    @Nullable
+    private DnsCnameCache cnameCache;
+    @Nullable
+    private Long queryTimeoutMillis;
+    @Nullable
+    private ResolvedAddressTypes resolvedAddressTypes;
+    @Nullable
+    private Boolean recursionDesired;
```
ikhoon: > This flag is enabled by default.

According to the documentation of the setter of this field,`boolean recursionDesired = true`?

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: There's a chance that the future is not canceled after the resolver is closed, but yes, we are checking `resolverClosed` and it won't be a big deal because it would finish in microseconds.
Let me change that. Thanks!

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
minwoox: All the fields are set only the user sets them. It not set, they will be set by the `DnsNameResolverBuilder` which has the default value.
Let me remove about the part explaining the default value from the Javadoc and add a link to it just in case the default value is changed and I missed that. Thanks!

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,306 @@
...
+        /**
+         * No need to be volatile because updated only by the {@link #executor()}.
+         */
+        private int numAttemptsSoFar = 1;
+
+        private volatile boolean servedFromCache;
+
+        @VisibleForTesting
+        @Nullable
+        ScheduledFuture<?> refreshingScheduledFuture;
```
anuraaga: I think I'd name it `refreshFuture` - `scheduled` is already part of the type and this future only runs once until rescheduling instead of using `scheduleWithFixedDelay` for example

minwoox: Fixed. ;)

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
ikhoon: Thanks for the detail explanation! LGTM :-)

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,306 @@
...
+        }
+
+        final List<DnsQuestion> questions =
+                dnsRecordTypes.stream()
+                              .map(type -> DnsQuestionWithoutTrailingDot.of(hostname, type))
+                              .collect(toImmutableList());
+        sendQueries(questions, hostname, result);
+        result.handle((entry, cause) -> {
+            if (cause != null) {
+                cache.remove(hostname);
```
ikhoon: I wonder removing the cached result is a good strategy. IMHO, I want to get cached DNS if there is an error on the DNS result.

minwoox: Then, the client will never get resolved addresses.

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,306 @@
...
+            }
+
+            if (inetAddress == null) {
+                result.completeExceptionally(
+                        new UnknownHostException("failed to receive DNS records for " + hostname));
+                return;
+            }
+            final CacheEntry entry = new CacheEntry(inetAddress, ttlMillis, questions);
+            result.complete(entry);
+            entry.scheduleRefresh(ttlMillis);
```
ikhoon: nit: How about moving this logic up `records.forEach(ReferenceCountUtil::safeRelease)` to complete the future quickly?

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: Add `negativeTtl` to give a choice to users to cache or not when DNS queries are failed. The default value is `0` which means the resolve does not cache when the query fails, so it will be no problem. 😄

minwoox: @anuraaga and @trustin PTAL again. 😉

core/src/main/java/com/linecorp/armeria/client/DnsResolverGroupBuilder.java
```diff
@@ -0,0 +1,345 @@
...
+        this.minTtl = minTtl;
+        this.maxTtl = maxTtl;
+        return this;
+    }
+
+    /**
+     * Sets the TTL of the cache for the failed DNS queries in seconds. The default value is {@code 0} which
+     * means that the {@link AddressResolver} does not cache when DNS queries are failed.
+     */
+    public DnsResolverGroupBuilder negativeTtl(int negativeTtl) {
```
anuraaga: `failedTtl`? Not sure what negative means here (`checkArgument(negativeTtl >= 0)` looks a little funny ;))

Also what gets cached if the query failed?

minwoox: I brought this from [netty](https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java#L161)
but yes, it took a little time for me to understand what it means. We don't have to follow it. 😄 

> Also what gets cached if the query failed?

It gets the exception when the first query fails. Mostly `UnknownHostException`.

trustin: Better using the same property name with Netty?

ikhoon: I think `negative` is pretty standard for the failed cache.:-)
* https://tools.ietf.org/html/rfc2308
* https://en.wikipedia.org/wiki/Negative_cache

anuraaga: Ah didn't know it's a standard term :)

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: I didn't get what you mean. 😄

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,327 @@
...
+                servedFromCache = true;
+                numAttemptsSoFar = 1;
+
+                if (entry.address().equals(address) && entry.ttlMillis() == ttlMillis) {
+                    scheduleRefresh(ttlMillis);
+                } else {
+                    // Replace the old entry with the new one.
+                    cache.put(hostName, result);
+                    entry.scheduleRefresh(entry.ttlMillis());
+                }
```
ikhoon: I have a question on `entry.scheduleRefresh(ttlMillis)` logic. The refreshing `entry` is scheduled in the end of `sendQueries(...)`. 
And there are different refreshing logics here. IIUC, doesn't it cause double scheduling?

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
ikhoon: I thought we can release `recodes` lazy... but I guess it is meaningless. Sorry for the confusion. 😀

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: Ah yes, thanks for finding this. 😄

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,331 @@
...
+        final List<DnsQuestion> questions =
+                dnsRecordTypes.stream()
+                              .map(type -> DnsQuestionWithoutTrailingDot.of(hostname, type))
+                              .collect(toImmutableList());
+        sendQueries(questions, hostname, result);
+        result.handle((entry, unused) -> {
+            final Throwable cause = entry.cause();
+            if (cause != null) {
+                if (negativeTtl > 0) {
+                    executor().schedule(() -> cache.remove(hostname), negativeTtl, TimeUnit.SECONDS);
```
ikhoon: I think Negative TTL is a kind of cache. `RefreshingAddressResolver` role is keeping cache and automatic refresh it. So the following logic will work to meet the feature that refreshes the failed result with `nagativeTtl`. 
```java
entry.scheduleRefresh(TimeUnit.SECONDS.toMillis(negativeTtl);
```

If the next result is successful, it will eventually call  `entry.scheduleRefresh(entry.ttlMillis())` next time. WDYT?

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -0,0 +1,331 @@
...
+                    if (nextDelayMillis < 0) {
+                        cache.remove(hostName);
+                        return null;
+                    }
+                    scheduleRefresh(nextDelayMillis);
+                    return null;
+                }
+
+                // Got the response successfully so reset the state.
+                servedFromCache = true;
```
minwoox: This is false. Let me fix that.

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: https://github.com/line/armeria/pull/2217#issuecomment-547723266
Please see this.

Issue #1719, opened 2019-04-16
Using 'builder()' factory method instead of constructor


trustin:
Currently, we instantiate a builder by calling a constructor or a static method in a builder class:

```java
Server server = new ServerBuilder()
        .foo()
        .bar()
        .build();
```

I wonder if it was a good idea or we need to reconsider this decision in favor of having `builder()` method:

```java
Server server = Server.builder()
                      .foo()
                      .bar()
                      .build();
```

The pros of `builder()` are:

- We can easily align at `'.'`, which is better in terms of aesthetics.
- We do not need to import `ServerBuilder` if we chain all.

eonezhang:
I like the `Server.builder()` way

trustin:
Let me make the proposed change.

trustin:
`builder()` or `newBuilder()`? Protobuf uses `newBuilder()` and Guava uses `builder()`. Which one is better?

anuraaga:
+1.001 for `builder` vs +1 for `newBuilder` :)

trustin:
Just came back from Spain and it reads like 1001 vs 1 :rofl:

4whomtbts:
I'm working on this

trustin:
The remaining builder classes that need migration:

- `ClientBuilder`
- `DefaultClientBuilderParams` 
- `ServerListenerBuilder`
- `HttpAuthServiceBuilder`
- `SimpleCompositeServiceBuilder`
- `SimpleCompositeRpcServiceBuilder`
- `CorsPolicyBuilder`
- `CorsServiceBuilder`
- `DocServiceBuilder`
- `HttpFileBuilder`
- `HttpFileServiceBuilder`
- `LoggingServiceBuilder`
- `JettyServiceBuilder`
- `ArmeriaRetrofitBuilder`
- `KeyStoreCredentialResolverBuilder`
- `SamlServiceProviderBuilder`
- `TomcatServiceBuilder`
- `ZooKeeperUpdatingListenerBuilder`

4whomtbts:
It has been on progress since this monday. Apologize for for slow progress (__)

ikhoon:
@4whomtbts It's good to be back! I think you need to know some changes in #2335. 
1) Please make sure your current branch is up to date from the master branch. Lots of class which have `HTTP` keyword have been deprecated.
2) `HttpAuthServiceBuilder` and `HttpFileServiceBuilder` have been done, you may want to skip them.
[...]

PR #2221, opened 2019-10-30
Switch to static `builder()` methods

trustin:
Related: #1719
Motivation:

We decided to switch from `new *Builder()` to `*.builder()`.

Modifications:

- Switch to static `builder()` methods in:
  - `ClientFactory`
  - `ClientOptions`
  - `ClientDecoration`
  - `ClientConnectionTimings`
  - `HttpClient`
  - `LoggingClient`
  - `LoggingService`
  - `EndpointInfo`
  - `FieldInfo`
  - `ClientCacheControl`
  - `ServerCacheControl`
  - `ClientRequestContext`
  - `ServiceRequestContext`
  - `RequestContextCurrentTraceContext`
  - `RetrofitMeterIdPrefixFunction`
    - (Breaking) `RetrofitMeterIfPrefixFunctionBuilder` is now a top-level class without a public constructor.
- Miscellaneous:
  - Increase the test timeout for `ArmeriaSpringActuatorAutoConfigurationTest` for less flakiness
  - Maybe fixes the flakiness in `GracefulShutdownSupportTest`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
@@ -43,10 +43,10 @@ import io.netty.channel.EventLoopGroup;
  *
  * <h3>Life cycle of the default {@link ClientFactory}</h3>
  * <p>
- * {@link Clients} or {@link ClientBuilder} uses {@link #DEFAULT}, the default {@link ClientFactory},
- * unless you specified a {@link ClientFactory} explicitly. Calling {@link #close()} on the default
- * {@link ClientFactory} will neither terminate its I/O threads nor release other related resources unlike
- * other {@link ClientFactory} to protect itself from accidental premature termination.
+ * {@link Clients} or {@link ClientBuilder} uses the default {@link ClientFactory} returned by
+ * {@link #ofDefault()}, unless you specified a {@link ClientFactory} explicitly. Calling {@link #close()}
+ * on the default {@link ClientFactory} will neither terminate its I/O threads nor release other related
+ * resources unlike other {@link ClientFactory} to protect itself from accidental premature termination.
  * </p><p>
  * Instead, when the current {@link ClassLoader} is {@linkplain ClassLoader#getSystemClassLoader() the system
  * class loader}, a {@linkplain Runtime#addShutdownHook(Thread) shutdown hook} is registered so that they are
@@ -60,8 +60,25 @@ public interface ClientFactory extends AutoCloseable {
 
     /**
      * The default {@link ClientFactory} implementation.
+     *
+     * @deprecated Use {@link #ofDefault()}.
      */
-    ClientFactory DEFAULT = new ClientFactoryBuilder().build();
+    @Deprecated
+    ClientFactory DEFAULT = builder().build();
+
+    /**
+     * Returns the default {@link ClientFactory} implementation.
+     */
+    static ClientFactory ofDefault() {
+        return DEFAULT;
+    }
+
+    /**
+     * Returns a newly created {@link ClientFactoryBuilder}.
+     */
+    static ClientFactoryBuilder builder() {
+        return new ClientFactoryBuilder();
+    }
 
     /**
      * Closes the default {@link ClientFactory}.
@@ -69,12 +86,12 @@ public interface ClientFactory extends AutoCloseable {
     static void closeDefault() {
         LoggerFactory.getLogger(ClientFactory.class).debug(
                 "Closing the default {}", ClientFactory.class.getSimpleName());
-        ((DefaultClientFactory) DEFAULT).doClose();
+        ((DefaultClientFactory) ofDefault()).doClose();
     }
 
     /**
      * Disables the {@linkplain Runtime#addShutdownHook(Thread) shutdown hook} which closes
-     * {@linkplain #DEFAULT the default <code>ClientFactory</code>}. This method is useful when you need
+     * {@linkplain #ofDefault() the default <code>ClientFactory</code>}. This method is useful when you need
      * full control over the life cycle of the default {@link ClientFactory}.
      */
     static void disableShutdownHook() {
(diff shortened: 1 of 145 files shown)


core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -61,7 +61,14 @@
     /**
      * The default {@link ClientFactory} implementation.
      */
-    ClientFactory DEFAULT = new ClientFactoryBuilder().build();
+    ClientFactory DEFAULT = builder().build();
```
trustin: Should we switch to `ofDefault()`?

trustin:
While working on this, I found the following builder classes do not match well with the new pattern:

- `ClientBuilder`
  - Adding `Client.builder()` doesn't sound right. We can add `Clients.builder()` though. Let me know if you have a better idea.
- `ArmeriaRetrofitBuilder`
  - There's no `ArmeriaRetrofit`. Should we create an interface `ArmeriaRetrofit` that extends `Retrofit`? Thoughts?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+    private String serviceTagName;
+    @Nullable
+    private String defaultServiceName;
+
+    RetrofitMeterIdPrefixFunctionBuilder(String name) {
+        this.name = name;
+    }
+
+    /**
+     * Make tag of {@link RetrofitMeterIdPrefixFunction} contains Retrofit service interface name.
```
jyblue: Should it be `Makes a tag ...`? (I think it's a typo previously.)

trustin: Thanks! Rephrased the Javadoc to make it more comprehensive.

minwoox:
+1 for `Clients.builder()`.

> Should we create an interface ArmeriaRetrofit that extends Retrofit?
Retrofit is a final class so we cannot extend it. I'm not sure which one is the best. 😢

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
ikhoon: +1 I would lean towards using `ofDefaults()`

trustin: OK. Let me take care of it in another pull request.

trustin:
Gentle ping, guys :laughing:

trustin:
Took advantage of review lag to migrate more builders. :stuck_out_tongue_closed_eyes:

trustin:
Note to self: Use `(class [^ ]+Builder|new [^ ]+(?<!String)Builder\(|\{@link [^}]+Builder\})` to find the candidates.

core/src/main/java/com/linecorp/armeria/client/ClientOptions.java
```diff
@@ -71,31 +73,45 @@
...
      * Returns the {@link ClientOptions} with the specified {@link ClientOptionValue}s.
      */
     public static ClientOptions of(ClientOptionValue<?>... options) {
         requireNonNull(options, "options");
         if (options.length == 0) {
-            return DEFAULT;
+            return of();
         }
-        return new ClientOptions(DEFAULT, options);
+        return new ClientOptions(of(), options);
```
anuraaga: I think we can delegate to the other methods instead of repeating the empty checks `of(of(), options)`

core/src/main/java/com/linecorp/armeria/server/RouteBuilder.java
```diff
@@ -55,6 +55,8 @@
 
     private Set<MediaType> produces = ImmutableSet.of();
 
+    RouteBuilder() {}
```
minwoox: nit: deprecated?

core/src/main/java/com/linecorp/armeria/client/ClientOptions.java
trustin: That's a good idea.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -133,7 +133,29 @@ static ClientRequestContext of(RpcRequest request, String uri) {
...
+
+    /**
+     * Returns a new {@link ClientRequestContextBuilder} created from the specified {@link HttpRequest}.
+     */
+    static ClientRequestContextBuilder builder(HttpRequest request) {
+        return new ClientRequestContextBuilder(request);
+    }
+
+    /**
+     * Returns a new {@link ClientRequestContextBuilder} created from the specified {@link RpcRequest} and URI.
```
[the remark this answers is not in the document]
trustin: Will use `{@code uri}` because the parameter type is not `URI` in this case.

core/src/main/java/com/linecorp/armeria/client/HttpClient.java
```diff
@@ -197,7 +200,45 @@ static HttpClient of(ClientFactory factory, SessionProtocol protocol, Endpoint e
      */
     static HttpClient of(ClientFactory factory, SessionProtocol protocol, Endpoint endpoint,
                          ClientOptions options) {
-        return new HttpClientBuilder(protocol, endpoint).factory(factory).options(options).build();
+        return builder(protocol, endpoint).factory(factory).options(options).build();
+    }
+
+    /**
+     * Returns a new {@link HttpClientBuilder} created without a base URI.
```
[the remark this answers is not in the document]
trustin: OK!

trustin:
Sneaking in two miscellaneous changes, but I guess it's OK. Will merge tomorrow morning. Leave a message if you'd like them as a separate PR.

PR #2203, opened 2019-10-22
Add UUID From RequestContext

yanghun0070:
Fix #2001 
Add UUID From RequestContext

PR #2224, opened 2019-11-01
Replace `UUID` with `RequestId` and make its generation configu…

trustin:
Related: #2001 #2203
Motivation:

- A request ID doesn't need to be a UUID. It'd be nice if a user can
  customize request ID generation
- A new request ID is always generated when creating a derived context.
  A user might want to specify a certain request ID instead of
  generating a new one.
- `UUID.randomUUID()` relies on `SecureRandom`, but it can lead to a
  performance problem when system is in short of entropy.
- It'd be nice if a user can log the request ID in an access log.

Modifications:

- Add `RequestId`
- Add `ServerBuilder.requestIdGenerator()`
- Add `ClientOption.REQUEST_ID_GENERATOR` and its related methods
- (Breaking) Remove `RequestContext.newDerivedContext()`
- (Breaking) Add `RequestId` as the first parameter of
  `RequestContext.newDerivedContext(HttpRequest, RpcRequest)`
- Add a new access log component `%I`
- Add `ThreadLocalByteArray` to reduce the unnecessary allocation of ephemeral byte arrays
  - Use `ThreadLocalByteArray` where applicable

Result:

- A user can configure how a request ID is generated.
- A user can override the request ID of a derived context.
- A user can write the request ID into access logs
- (Breaking) `RequestContext.newDerivedContext()` now always requires
  `RequestId`, `HttpRequest` and `RpcRequest`.


trustin:
@hyangtack Do you think the format string `%I` sounds OK?

core/src/main/java/com/linecorp/armeria/internal/UuidUtil.java
```diff
@@ -0,0 +1,79 @@
...
+     * except that it uses a cryptographically insecure random for higher performance.
+     */
+    public static Supplier<UUID> randomGenerator() {
+        return RandomUuidGenerator.INSTANCE;
+    }
+
+    /**
+     * Returns a newly generated random {@link UUID}.
+     */
+    public static UUID random() {
```
anuraaga: This PR made me realize, we should probably use `String` as the type of the ID rather than `UUID` itself - no reason to force the user to specify a `UUID` if they'd prefer to specify some other type of `String`.

trustin: That's a good idea. Let me send a follow-up PR for that once this is merged.

anuraaga: A follow up PR is ok but wouldn't it affect most of this change? Even abbrev becomes much simpler I think :)

trustin: That's a good point. :sweat_smile: What do you think about defining an interface for a request ID?

```java
interface RequestId {
    String asText();
    long asLong(); // Could be used as a Brave Span ID?
    String abbreviate();
}
```

trustin: @adriancole We are going to add an `id` property to `RequestContext`, which will be a random 64-bit integer by default, which is identical to Brave's span ID. I think it will be great if a user can use the same value for `RequestContext.id` and Span ID for easier correlation. However, it seems like span ID generation is currently hard-coded in `Tracer.nextId()`. What do you think about allowing the customization of span ID generation?

anuraaga: We've actually been talking about allowing customization of more Tracer / Tracing defaults when creating a span - this would fit well in that model. The biggest downside would be a bad function (e.g. not actually random) breaking tracing - unlike breaking logs, breaking tracing would affect other services/teams.

So given the potential danger, I would lean against customizing span ID since the benefit doesn't seem all that huge (the log integration already makes sure span ids are logged, they happen to be different from the Armeria ID but I think it's fairly expected). I'm also wary of tying the lifecycle of this uid to the lifecycle of distributed tracing (off the top of my head I don't know if our uid currently has the same behavior for retrying client for example and am not sure if it should or not).

FWIW, my plan for this customizer was to always return empty string since I always apply brave, which is a solution I'm happy with. Might be a little, but not that much happier to have a shared ID.

core/src/main/java/com/linecorp/armeria/common/RequestId.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Returns a {@link RequestId} created from a 64-bit random integer.
+     */
+    static RequestId random() {
+        return new DefaultRequestId();
+    }
+
+    /**
+     * Returns the long textual representation of this ID.
```
anuraaga: I think it can just be `text` and `shortText` since the long one should basically be a full representation.

core/src/test/java/com/linecorp/armeria/common/DefaultRequestIdTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.junit.jupiter.api.Test;
+
+class DefaultRequestIdTest {
+
+    @Test
+    void basic() {
+        final RequestId id = RequestId.random();
+        assertThat(id).isInstanceOf(DefaultRequestId.class);
+        assertThat(id.longText()).hasSize(16);
```
anuraaga: Think it's worth having a test that passes in a value and checks the actual values of `longText`, `shortText`, especially with a positive number

trustin: Agreed. Added a test case for that.

core/src/main/java/com/linecorp/armeria/common/RequestId.java
trustin: Yeah, renamed.

core/src/main/java/com/linecorp/armeria/common/DefaultRequestId.java
```diff
@@ -0,0 +1,114 @@
...
+        if (longText != null) {
+            return longText;
+        }
+
+        return longText = newLongText();
+    }
+
+    private String newLongText() {
+        final String newLongText;
+        if (Integer.numberOfLeadingZeros((int) (value >>> 32)) < 4) {
```
anuraaga: Think it's worth adding a small comment on why we need the slow branch or what would go wrong if we didn't.

core/src/main/java/com/linecorp/armeria/common/DefaultRequestId.java
```diff
@@ -0,0 +1,114 @@
...
+            newShortText = Integer.toHexString(value);
+        } else {
+            newShortText = newTextSlow(value, 8);
+        }
+        return newShortText;
+    }
+
+    @SuppressWarnings("deprecation")
+    private static String newTextSlow(long value, int digits) {
+        final byte[] bytes = new byte[digits];
```
anuraaga: While optimizing long -> hex in zipkin, I found ThreadLocal<byte[]> of fixed size is very beneficial. I would have expected escape analysis to work here but it didn't on Java 11.

Can use a 16-byte buffer for both long and short, using the constructor that accepts offset/length

core/src/main/java/com/linecorp/armeria/common/DefaultRequestId.java
trustin: Added comments and optimized it a little bit more.

core/src/main/java/com/linecorp/armeria/common/DefaultRequestId.java
trustin: Added `ThreadLocalByteArray` and used it where applicable.

core/src/main/java/com/linecorp/armeria/internal/PathAndQuery.java
```diff
@@ -471,7 +470,7 @@ private static void appendHexNibble(StringBuilder buf, int nibble) {
         int length;
 
         Bytes(int initialCapacity) {
-            data = new byte[initialCapacity];
+            data = ThreadLocalByteArray.get(initialCapacity);
```
anuraaga: I'm assuming this usage is safe due to the way `Bytes` is used, but I feel this optimization is really scarry when the buffer is allowed to survive past a syntactic function block. But will leave it to you on whether to keep as is, refactor into one method somehow, or revert.

anuraaga: Nevermind :)

trustin: It was not safe because we're dealing with two `Bytes` instances, so I had to revert it. :sweat_smile:

grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/StatusMessageEscaper.java
```diff
@@ -102,10 +105,8 @@ private static String doEscape(byte[] valueBytes, int ri) {
             }
             escapedBytes[wi++] = b;
         }
-        final byte[] dest = new byte[wi];
-        System.arraycopy(escapedBytes, 0, dest, 0, wi);
 
-        return new String(dest, StandardCharsets.US_ASCII);
+        return new String(escapedBytes, 0, wi, StandardCharsets.US_ASCII);
```
trustin: @anuraaga Is it OK to use `new String(escapedBytes, 0, 0, wi)` instead of this for even better performance? It seems so, but just in case :-)

And please note that I removed an extra byte array allocation :stuck_out_tongue_winking_eye:

trustin:
Note to self: Make `ClientOptions.DEFAULT_OPTIONS` have all options and stop using `getOrElse`.

core/src/main/java/com/linecorp/armeria/common/RequestId.java
```diff
@@ -0,0 +1,51 @@
...
+     * Returns the full textual representation of this ID.
+     */
+    String text();
+
+    /**
+     * Returns the human-friendly short textual representation of this ID.
+     */
+    default String shortText() {
+        final String longText = text();
+        return longText.length() < 8 ? longText : longText.substring(0, 8);
```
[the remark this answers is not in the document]
trustin: Thanks!

core/src/main/java/com/linecorp/armeria/client/UserClient.java
```diff
@@ -142,7 +142,7 @@ protected final O execute(@Nullable EventLoop eventLoop, Endpoint endpoint,
         final DefaultClientRequestContext ctx;
         final HttpRequest httpReq;
         final RpcRequest rpcReq;
-        final UUID uuid = UUID.randomUUID();
+        final RequestId id = options().requestIdGenerator().get();
```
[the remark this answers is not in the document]
trustin: That's a good point. Added some logging and fallback.

PR #2227, opened 2019-11-05
Fix not to push context multiple times when notifying multiple…

minwoox:
…tLogListener

Motivation:
We push `(Client|Service)RequestContext` to thread local before calling `RequestLogListener.onRequestLog()`.
When there are multiple `RequestLogListener`s, we push the context every time before calling `onRequestLog`.
This could be a problem when a user sets an expensive hook using `ctx.onEnter()`. So we should fix to just push once.

Modification:
- Push context only once before notifying multiple listeners

Result:
- Better performance

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/logging/RequestLogListenerInvoker.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/RequestLogListenerInvoker.java
@@ -62,7 +62,7 @@ public final class RequestLogListenerInvoker {
                 }
                 try {
                     listener.onRequestLog(log);
-                } catch (Exception e) {
+                } catch (Throwable e) {
                     logger.warn("onRequestLog() failed with an exception:", e);
                 }
             }
--- a/core/src/main/java/com/linecorp/armeria/common/logging/RequestLogListenerInvoker.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/RequestLogListenerInvoker.java
@@ -16,6 +16,8 @@
 
 package com.linecorp.armeria.common.logging;
 
+import static java.util.Objects.requireNonNull;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/linecorp/armeria/common/logging/RequestLogListenerInvokerTest.java
```diff
@@ -0,0 +1,51 @@
...
+        final ServiceRequestContext ctx = ServiceRequestContextBuilder.of(HttpRequest.of(HttpMethod.GET, "/"))
+                                                                      .build();
+        ctx.onEnter(c -> counter.incrementAndGet());
+        final RequestLog log = ctx.log();
+
+        log.addListener(l -> { /* no-op */ }, RequestLogAvailability.REQUEST_END);
+        log.addListener(l -> { /* no-op */ }, RequestLogAvailability.REQUEST_END);
+        ctx.logBuilder().endRequest();
+
+        await().until(() -> counter.get() == 1);
```
anuraaga: Do we need awaitility / sleep? These callbacks should all run directly.

minwoox: No need. Have implemented without much thinking. 😅 Let me fix that, thanks!

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogListenerInvoker.java
```diff
@@ -34,12 +36,38 @@
...
+        requireNonNull(listeners, "listeners");
+        requireNonNull(log, "log");
+        if (listeners.length == 0) {
+            return;
+        }
+
+        try (SafeCloseable ignored = log.context().push()) {
+            for (RequestLogListener listener : listeners) {
+                if (listener == null) {
+                    continue;
```
trustin: Shouldn’t this be `break`?

core/src/test/java/com/linecorp/armeria/common/logging/RequestLogListenerInvokerTest.java
```diff
@@ -0,0 +1,46 @@
...
+        final ServiceRequestContext ctx = ServiceRequestContextBuilder.of(HttpRequest.of(HttpMethod.GET, "/"))
+                                                                      .build();
+        ctx.onEnter(c -> counter.incrementAndGet());
+        final RequestLog log = ctx.log();
+
+        log.addListener(l -> { /* no-op */ }, RequestLogAvailability.REQUEST_END);
+        log.addListener(l -> { /* no-op */ }, RequestLogAvailability.REQUEST_END);
+        ctx.logBuilder().endRequest();
+
+        assertThat(counter.get()).isOne();
```
trustin: Should we add a test case that makes sure the counter stays at 0 when there are no listeners to notify?

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogListenerInvoker.java
minwoox: I thought either way is fine because the listener can be `null`. But let me change that. 😄

core/src/test/java/com/linecorp/armeria/common/logging/RequestLogListenerInvokerTest.java
minwoox: Let me add one. Thanks!

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogListenerInvoker.java
trustin: If `listener` cannot be `null`, you could remove the `if` completely.

trustin: But it still seems to me that `listener` can be `null`, according to `removeSatisfiedListeners()`.

minwoox: Ah yes, didn't read code carefully. 😅

minwoox:
Thanks for reviewing!

PR #2229, opened 2019-11-05
Update SLF4J and Logback Javadoc URL

trustin:
- www.slf4j.org and logback.qos.ch started to respond with non-200
  status.


minwoox:
👍

PR #2230, opened 2019-11-05
Introduce ClientFactoryOptions

### LINKED ISSUE #1890: Expose ClientFactory params
When trying to write a unit test to verify some builder logic configures `ClientFactory` correctly, I found that I can't seem to read the settings of a `ClientFactory`. It should probably have a `.options()` method, similar to how a `Client` has `options()`.

delegacy:
Motivation:

- Users might want to build a new ClientFactory which inherits an
existing ClientFactory's settings, while overriding a certain option
only.

Result:

- Fixes #1890


delegacy:
This is in-progress yet, and I'd like to hear reviewers' thoughts about this approach. (`ClientFactoryOptions` similar to `ClientOptions`)

trustin:
Looks like you’re going into the right direction. Just one thing though:

Is there a reason you did not add the options for name resolution and event loop scheduling? If we add options for them, we could remove `ClientFactoryBuilderParams` and just add `ClientFactory.options()`. We might have to add some validation logic so a user is still disallowed to set mutually exclusive options.

delegacy:
> Is there a reason you did not add the options for name resolution and event loop scheduling?

I had the same question in mind, actually. :) These two factories are parameters of `ClientFactoryBuilder`, while others are parameters of `ClientFactory`, I think.

If I try event loop scheduling as `ClientFactoryOption`, what would be the type?
`ClientFactoryOption<EventLoopScheduler>`?

delegacy:
I think it is ready for review.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -123,6 +96,9 @@
      */
     @Deprecated
     public ClientFactoryBuilder() {
+        options.put(ClientFactoryOption.CHANNEL_OPTIONS,
+                    ClientFactoryOption.CHANNEL_OPTIONS.newValue(channelOptions));
```
trustin: Could we start from an empty `options` so we can use the default `ClientFactoryOptions` instance when a user did not set any options, rather than creating a new `ClientFactoryOptions` instance?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -429,22 +405,61 @@ public ClientFactoryBuilder useHttp1Pipelining(boolean useHttp1Pipelining) {
...
+     */
+    @SuppressWarnings("unchecked")
+    public <T> ClientFactoryBuilder option(ClientFactoryOptionValue<T> optionValue) {
+        requireNonNull(optionValue, "optionValue");
+        final ClientFactoryOption<?> opt = optionValue.option();
+        if (opt == ClientFactoryOption.EVENT_LOOP_SCHEDULER) {
+            eventLoopSchedulerFactory(eventLoopGroup -> (EventLoopScheduler) optionValue.value());
+        } else if (opt == ClientFactoryOption.ADDRESS_RESOLVER_GROUP) {
+            addressResolverGroupFactory(
+                    eventLoopGroup -> (AddressResolverGroup<InetSocketAddress>) optionValue.value());
```
trustin: Why don't we just replace `EVENT_LOOP_SCHEDULER` and `ADDRESS_RESOLVER_GROUP` with `EVENT_LOOP_SCHEDULER_FACTORY` and `ADDRESS_RESOLVER_GROUP_FACTORY`?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -73,48 +71,23 @@
...
-    // Do not accept 1) the options that may break Armeria and 2) the deprecated options.
-    @SuppressWarnings("deprecation")
-    private static final Set<ChannelOption<?>> PROHIBITED_SOCKET_OPTIONS = ImmutableSet.of(
-            ChannelOption.ALLOW_HALF_CLOSURE, ChannelOption.AUTO_READ,
-            ChannelOption.AUTO_CLOSE, ChannelOption.MAX_MESSAGES_PER_READ,
-            ChannelOption.WRITE_BUFFER_HIGH_WATER_MARK, ChannelOption.WRITE_BUFFER_LOW_WATER_MARK,
-            EpollChannelOption.EPOLL_MODE);
-
     // Netty-related properties:
     private EventLoopGroup workerGroup = CommonPools.workerGroup();
```
trustin: This could be removed because we have the `WORKER_GROUP` option now?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,280 @@
...
+            ClientFactoryOption.HTTP1_MAX_HEADER_SIZE.newValue(Flags.defaultHttp1MaxHeaderSize()),
+            ClientFactoryOption.HTTP1_MAX_CHUNK_SIZE.newValue(Flags.defaultHttp1MaxChunkSize()),
+            ClientFactoryOption.IDLE_TIMEOUT_MILLIS.newValue(Flags.defaultClientIdleTimeoutMillis()),
+            ClientFactoryOption.USE_HTTP2_PREFACE.newValue(Flags.defaultUseHttp2Preface()),
+            ClientFactoryOption.USE_HTTP1_PIPELINING.newValue(Flags.defaultUseHttp1Pipelining()),
+            ClientFactoryOption.CONNECTION_POOL_LISTENER.newValue(DEFAULT_CONNECTION_POOL_LISTENER),
+            ClientFactoryOption.METER_REGISTRY.newValue(Metrics.globalRegistry)
+    };
+
+    public static final ClientFactoryOptions DEFAULT = new ClientFactoryOptions(DEFAULT_OPTIONS);
```
trustin: Could we make this `private` and add a static factory method `of()` which returns `DEFAULT`, just like we do in `ClientOptions.of()`?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * TBW.
+     */
+    public static ClientFactoryOptions of(ClientFactoryOptionValue<?>... options) {
+        requireNonNull(options, "options");
+        if (options.length == 0) {
+            return DEFAULT;
+        }
+        return new ClientFactoryOptions(DEFAULT, options);
```
trustin: Could add `of(ClientFactoryOptions baseOptions, ...)` and simplify into `return of(of(), options)`. Please make this class have the same set of static factory methods with `ClientOptions`.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,280 @@
...
+            optionValue = castOptionValue;
+        }
+
+        return optionValue;
+    }
+
+    private static Map<ChannelOption<?>, Object> filterChannelOptions(
+            Map<ChannelOption<?>, Object> channelOptions) {
+
+        requireNonNull(channelOptions, "channelOptions");
```
trustin: You need to make an immutable copy first because `channelOptions` may be mutable.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,280 @@
...
+     */
+    public Map<ClientFactoryOption<Object>, ClientFactoryOptionValue<Object>> asMap() {
+        return asMap0();
+    }
+
+    /**
+     * TBW.
+     */
+    public EventLoopGroup workerGroup() {
+        return getOrElse0(ClientFactoryOption.WORKER_GROUP, CommonPools.workerGroup());
```
trustin: Please use `get(option).get()` instead of `getOrElse()`, so we keep the default values only in `DEFAULT_OPTIONS`, i.e. less duplication.

delegacy:
Ready for the next round.

trustin:
Resolved the conflict for you ;-)

delegacy:
> Looks good to me except Javadoc. Could you fill Javadoc?

Done. :)

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,366 @@
...
+                                 Iterable<ClientFactoryOptionValue<?>> options) {
+
+        super(ClientFactoryOptions::filterValue, clientFactoryOptions, options);
+    }
+
+    private ClientFactoryOptions(ClientFactoryOptions clientFactoryOptions,
+                                 ClientFactoryOptions options) {
+
+        super(clientFactoryOptions, options);
+    }
```
trustin: Please move the constructors before other member methods.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -128,6 +128,11 @@ static void disableShutdownHook() {
      */
     MeterRegistry meterRegistry();
 
+    /**
+     * Returns the {@link ClientFactoryOptions} that has been used to create this {@link ClientFactory}.
+     */
+    ClientFactoryOptions options();
```
trustin: Could you move this method after `setMeterRegistry()`, so that `meterRegistry()` and `setMeterRegistry()` are grouped together?

trustin:
Could you also move the implementations of `ClientFactory.options()` are located in the same order as in `ClientFactory`? For example, you could move `options()` so that it's placed after `setMeterRegistry()` and before `newClient()`, *not* after `close()`.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,367 @@
...
+        return channelOptions;
+    }
+
+    private ClientFactoryOptions(ClientFactoryOptionValue<?>... options) {
+        super(ClientFactoryOptions::filterValue, options);
+    }
+
+    private ClientFactoryOptions(ClientFactoryOptions clientFactoryOptions,
+                                 ClientFactoryOptionValue<?>... options) {
+
```
ikhoon: nit: remove this blank line?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,367 @@
...
+        super(ClientFactoryOptions::filterValue, clientFactoryOptions, options);
+    }
+
+    private ClientFactoryOptions(ClientFactoryOptions clientFactoryOptions,
+                                 Iterable<ClientFactoryOptionValue<?>> options) {
+
```
ikhoon: nit: remove this blank line?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,367 @@
...
+        super(ClientFactoryOptions::filterValue, clientFactoryOptions, options);
+    }
+
+    private ClientFactoryOptions(ClientFactoryOptions clientFactoryOptions,
+                                 ClientFactoryOptions options) {
+
```
ikhoon: nit: remove this blank line?

trustin:
@ikhoon I think the empty lines are inserted deliberately not to make the code look too compact?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,367 @@
...
+/**
+ * A set of {@link ClientFactoryOption}s and their respective values.
+ */
+@SuppressWarnings("OptionalGetWithoutIsPresent")
+public final class ClientFactoryOptions extends AbstractOptions {
+    private static final EventLoopGroup DEFAULT_WORKER_GROUP = CommonPools.workerGroup();
+
+    private static final Function<? super EventLoopGroup, ? extends EventLoopScheduler>
+            DEFAULT_EVENT_LOOP_SCHEDULER_FACTORY =
+            eventLoopGroup -> new DefaultEventLoopScheduler(eventLoopGroup, 0, 0, Collections.emptyList());
```
minwoox: I think this should be `eventLoopGroup, 1, 0, Collections.emptyList()` so that only one connection is made for HTTP/2

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOption.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * The {@link Consumer} which can arbitrarily configure the {@link SslContextBuilder} that will be
+     * applied to the SSL session.
+     */
+    public static final ClientFactoryOption<Consumer<? super SslContextBuilder>> SSL_CONTEXT_CUSTOMIZER =
+            valueOf("SSL_CONTEXT_CUSTOMIZER");
+
+    /**
+     * The factory that creates a {@link AddressResolverGroup} which resolves remote addresses into
```
minwoox: nit: a -> an

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -0,0 +1,367 @@
...
+    /**
+     * Returns the {@link Consumer} which can arbitrarily configure the {@link SslContextBuilder} that will be
+     * applied to the SSL session.
+     */
+    public Consumer<? super SslContextBuilder> sslContextCustomizer() {
+        return get0(ClientFactoryOption.SSL_CONTEXT_CUSTOMIZER).get();
+    }
+
+    /**
+     * Returns the factory that creates a {@link AddressResolverGroup} which resolves remote addresses into
```
minwoox: nit: a -> an

ikhoon:
@delegacy Could you give me permission to resolve the conflicting files?

ikhoon:
@delegacy Sorry, the permission was granted already. I was my mistake 😅
Updated: 
* Merge master: https://github.com/line/armeria/pull/2230/commits/a480d9e65cb091d3dd1210042271c0fa9882b5ef
* Address comments by @minwoox: https://github.com/line/armeria/pull/2230/commits/431f23f06c2a1326c8ce8b7e3c67d22fb327fc28

delegacy:
Thanks, reviewers.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
delegacy: I passed `0` to (implicitly) express that it will follow the default value. (The previous `ClientFactoryBuilder` also set `0` if `maxNumEventLoopsPerEndpoint` is not specified.) It will eventually use `1` (`DEFAULT_MAX_NUM_EVENT_LOOPS`) inside `DefaultEventLoopScheduler`.

I understand `1` for `maxNumEventLoopsPerEndpoint` is more semantically right. IMO, `DEFAULT_MAX_NUM_EVENT_LOOPS` should be used instead of `1`, then.

minwoox: Ah, yes. I completely forgot about it. 😱

minwoox: Let me send the fixing PR. Thanks!

PR #2231, opened 2019-11-05
Fix not to cache DNS failure when timed out

minwoox:
Motivation:
In Netty, `DnsNameResolver` only caches if the failure was not because of an IO error or timeout, but that was caused by query itself.
So we should follow the behavior.

Modifications:
- Cache only when the raised `UnknownHostException` does not have cause.

Result:
- Do not cache when IO error or timeout happens

To-do:
- Sending a PR to upstream so that we don't rely on this checking cause to figure out the cause is cacheable or not.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
+++ b/core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
@@ -115,7 +115,7 @@ final class RefreshingAddressResolver extends AbstractAddressResolver<InetSocket
         result.handle((entry, unused) -> {
             final Throwable cause = entry.cause();
             if (cause != null) {
-                if (entry.isCacheableCause() && negativeTtl > 0) {
+                if (entry.hasCacheableCause() && negativeTtl > 0) {
                     executor().schedule(() -> cache.remove(hostname), negativeTtl, TimeUnit.SECONDS);
                 } else {
                     cache.remove(hostname);
@@ -151,20 +151,20 @@ final class RefreshingAddressResolver extends AbstractAddressResolver<InetSocket
             if (!f.isSuccess()) {
                 final Throwable cause = f.cause();
 
-                // TODO(minwoox): In Netty, DnsNameResolver only cache if the failure was not because of an
+                // TODO(minwoox): In Netty, DnsNameResolver only caches if the failure was not because of an
                 //                IO error / timeout that was caused by the query itself.
-                //                To figure out that, we need to check the cause of the UnknownHostException.
+                //                To figure that out, we need to check the cause of the UnknownHostException.
                 //                If it's null, then we can cache the cause. However, this is very fragile
                 //                because Netty can change the behavior while we are not noticing that.
                 //                So sending a PR to upstream would be the best solution.
-                final boolean isCacheableCause;
+                final boolean hasCacheableCause;
                 if (cause instanceof UnknownHostException) {
                     final UnknownHostException unknownHostException = (UnknownHostException) cause;
-                    isCacheableCause = unknownHostException.getCause() == null;
+                    hasCacheableCause = unknownHostException.getCause() == null;
                 } else {
-                    isCacheableCause = false;
+                    hasCacheableCause = false;
                 }
-                result.complete(new CacheEntry(null, -1, questions, cause, isCacheableCause));
+                result.complete(new CacheEntry(null, -1, questions, cause, hasCacheableCause));
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -149,7 +149,25 @@ private void sendQueries(List<DnsQuestion> questions, String hostname,
...
         recordsFuture.addListener(f -> {
             if (!f.isSuccess()) {
-                result.complete(new CacheEntry(null, -1, questions, f.cause()));
+                final Throwable cause = f.cause();
+
+                // In Netty, only cache if the failure was not because of an IO error / timeout that was
+                // caused by the query itself. To figure out that, we need to check the cause of the
+                // UnknownHostException. If it's null, then we can cache the cause.
+                // However, this is very vulnerable because Netty can change the behavior while we are not
+                // noticing that. So sending a PR to upstream would be the best solution.
```
anuraaga: Would it be possible to at least file an issue now and link it here so not forgotten?

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -149,7 +149,25 @@ private void sendQueries(List<DnsQuestion> questions, String hostname,
         final Future<List<DnsRecord>> recordsFuture = resolver.sendQueries(questions, hostname);
         recordsFuture.addListener(f -> {
             if (!f.isSuccess()) {
-                result.complete(new CacheEntry(null, -1, questions, f.cause()));
+                final Throwable cause = f.cause();
+
+                // In Netty, only cache if the failure was not because of an IO error / timeout that was
+                // caused by the query itself. To figure out that, we need to check the cause of the
+                // UnknownHostException. If it's null, then we can cache the cause.
+                // However, this is very vulnerable because Netty can change the behavior while we are not
```
anuraaga: `fragile` instead of `vulnerable`

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: Yes, let me make one. 😃

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -149,7 +149,22 @@ private void sendQueries(List<DnsQuestion> questions, String hostname,
         final Future<List<DnsRecord>> recordsFuture = resolver.sendQueries(questions, hostname);
         recordsFuture.addListener(f -> {
             if (!f.isSuccess()) {
-                result.complete(new CacheEntry(null, -1, questions, f.cause()));
+                final Throwable cause = f.cause();
+
+                // TODO(minwoox): In Netty, DnsNameResolver only cache if the failure was not because of an
```
trustin: cache -> caches

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -149,7 +149,22 @@ private void sendQueries(List<DnsQuestion> questions, String hostname,
...
+                //                IO error / timeout that was caused by the query itself.
+                //                To figure out that, we need to check the cause of the UnknownHostException.
```
trustin: To figure that out

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
```diff
@@ -219,6 +234,7 @@ public void close() {
 
         @Nullable
         private final Throwable cause;
+        private final boolean isCacheableCause;
```
trustin: hasCacheableCause()?

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/TestDnsServer.java
```diff
@@ -63,6 +64,11 @@ public static DnsRecord newAddressRecord(String name, String ipAddr, long ttl) {
     private volatile Map<DnsQuestion, DnsResponse> responses;
 
     public TestDnsServer(Map<DnsQuestion, DnsResponse> responses) {
+        this(responses, new ChannelInboundHandlerAdapter());
```
trustin: How about passing `null` and not adding anything?

core/src/main/java/com/linecorp/armeria/client/RefreshingAddressResolver.java
minwoox: Fixed. 😄

core/src/test/java/com/linecorp/armeria/client/endpoint/dns/TestDnsServer.java
minwoox: Fixed. 😄

trustin:
Nice work, @minwoox !

minwoox:
Thanks for reviewing!

PR #2235, opened 2019-11-06
Take weight into account when comparing `Endpoint`s.

trustin:
Motivation:

`Endpoint.equals()` and `compareTo()` currently do not take
`Endpoint.weight()` into account. It was a deliberate decision, but it
is confusing. For example, when a user tries to update an endpoint's
weight, the following code will not work:

    private Endpoint oldEndpoint = ...;

    void update(Endpoint newEndpoint) {
        if (oldEndpoint.equals(newEndpoint)) {
            throw new IllegalArgumentException("can't set twice");
        }
        oldEndpoint = newEndpoint;
    }

    update(Endpoint.of("foo").withWeight(100));

    // This will fail.
    update(Endpoint.of("foo").withWeight(200));

Modifications:

- Update `Endpoint.equals()`, `compareTo()` and `hashCode()` to take
  `weight` into account.
- Make sure `weight` is not taken in account for certain cases:
  - `HealthCheckedEndpointGroup`
  - `HealthCheckedEndpointGroupMetrics`

Result:

- (Breaking) The behavior of `Endpoint.equals()`, `compareTo()` and
  `hashCode()` has been changed in a breaking yet good way.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
@@ -20,7 +20,6 @@ import static java.util.Objects.requireNonNull;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.function.Consumer;
 
 import com.linecorp.armeria.client.Endpoint;
@@ -28,7 +27,6 @@ import com.linecorp.armeria.common.metric.MeterIdPrefix;
 
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.binder.MeterBinder;
-import it.unimi.dsi.fastutil.Hash;
 import it.unimi.dsi.fastutil.objects.Object2ObjectOpenCustomHashMap;
 
 /**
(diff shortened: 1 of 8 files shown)


trustin:
@anuraaga Yeah, I'm not really sure which is less surprising behavior. Maybe just better documenting better and telling people weight is not part of comparison, and fix `DynamicEndpointGroup`?

anuraaga:
Yeah I think I would lean towards that option.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroupMetrics.java
```diff
@@ -71,29 +69,31 @@ public void bindTo(MeterRegistry registry) {
...
+                    healthMap.put(endpoint, healthy);
+                    registry.gauge(idPrefix.name(),
+                                   idPrefix.tags("authority", endpoint.authority(),
+                                                 "ip", firstNonNull(endpoint.ipAddr(), "")),
+                                   this, unused -> healthMap.get(endpoint) ? 1 : 0);
+                });
+            }
         }
     }
+
```
imasahiro: remove unused newline?

trustin:
Let me close this pull request and just fix `DynamicEndpointGroup`.

PR #2236, opened 2019-11-06
Add Dropwizard module

### LINKED ISSUE #2165: Implementing Armeria GRPC in Dropwizard?
Hi there!

We have a project built in Dropwizard, and we are switching from using REST with Jersey to use GRPC instead. Dropwizard uses Jetty, so if I could attach GRPC would be great.

In Dropwizards `initialize`-method, I wrote like this:

```java
@Override
public void initialize(final Bootstrap<QueryServConfiguration> bootstrap) {
      sb.service(
              GrpcService.builder()
              .addService(new HelloService())
              .build()
      );
      sb.serviceUnder("/", new JettyServiceBuilder().build());
}
```

It builds but I cannot make any requests towards it. Do I have to attach it with a handler, something like [this?](http://line.github.io/armeria/server-servlet.html#embedding-jetty)

cricket007:
Adding module for dropwizard integration

**Why**

1) I use Dropwizard at work, currently, I'm sure others do as well, and trust it's built-in Jetty http server, but might want other features. There already is a Dropwizard gRPC Bundle, but doesn't handle Thrift, AFAIK (not that that is a major feature for me, anyway). 
2) Wanted to see how difficult swapping out Dropwizard Jetty servers would be. 😄 
3) I am curious how much "faster" Dropwizard would be with Armeria. src: https://www.techempower.com/benchmarks/#section=data-r18&hw=ph&test=plaintext&l=zik0vz-f

**Progress**

- [x] Dropwizard Bundle
  - [x] Buildable Source
  - Tests
    - [x] YAML configurations
    - [x] Dropwizard lifecycle of Armeria server 
    - [x] Testing of services within an Dropwizard application instance (maybe via AccessLog capture) 
- [x] Runnable Example
- [x] Docs

Closes #2165


dropwizard/build.gradle
```diff
@@ -0,0 +1,20 @@
+plugins {
+    id "com.github.hierynomus.license" version "0.15.0"
+}
+
+//license {
+//  header project.file('../CODE-LICENSE.txt')
+//  strictCheck true
+//   exclude "**/*.properties"
+//}
```
cricket007: TODO: Remove

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,222 @@
...
+
+        if (this.connector != null) {
+            try {
+                setupAmeriaHttp(this.serverBuilder, this.connector);
+            } catch (SSLException | CertificateException e) {
+                LOGGER.error("Unable to define TLS Server", e);
+            }
+        }
+
+        // TODO: Add more items to server builder via Configuration
```
cricket007: Would appreciate a comment on what else should be added to the builder here based on values available at https://www.dropwizard.io/en/stable/manual/configuration.html#servers

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,222 @@
...
+                Duration.ofMillis(getShutdownGracePeriod().toMilliseconds()));
+    }
+
+    private void setupAmeriaHttp(final ServerBuilder sb, final ConnectorFactory connector)
+        throws SSLException, CertificateException {
+        if (connector instanceof ArmeriaHttpConnectorFactory) {
+            LOGGER.debug("Building Armeria HTTP Server");
+            final ArmeriaHttpConnectorFactory factory = (ArmeriaHttpConnectorFactory) connector;
+            sb.http(factory.getPort());
+            // TODO: More HTTP1 settings
```
cricket007: Same as above

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,222 @@
...
+            if (factory.isSelfSigned()) {
+                sb.tlsSelfSigned();
+            }
+            final File cert = new File(factory.getKeyCertChainFile());
+            final File keyStore = new File(factory.getKeyStorePath());
+            if (factory.isValidKeyStorePassword()) {
+                sb.tls(cert, keyStore, factory.getKeyStorePassword());
+            } else {
+                sb.tls(cert, keyStore);
+            }
```
cricket007: I guessed on how the `.tls` method is supposed to be invoked... Dropwizard seems to expose different & more options here

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,222 @@
...
+import io.dropwizard.jersey.setup.JerseyEnvironment;
+import io.dropwizard.jetty.ConnectorFactory;
+import io.dropwizard.jetty.ContextRoutingHandler;
+import io.dropwizard.server.SimpleServerFactory;
+import io.dropwizard.setup.Environment;
+import io.dropwizard.util.Size;
+import io.dropwizard.validation.MinSize;
+
+@JsonTypeName(ArmeriaServerFactory.TYPE)
+public class ArmeriaServerFactory extends SimpleServerFactory {
```
cricket007: Could just implement `ServerFactory` rather than bring in everything from `SimpleServerFactory`, and copy only necessary pieces.

trustin:
Whoa. This is awesome! Let me review some time tomorrow!

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,222 @@
...
+            if (factory.isValidKeyStorePassword()) {
+                sb.tls(cert, keyStore, factory.getKeyStorePassword());
+            } else {
+                sb.tls(cert, keyStore);
+            }
+            // TODO: More HTTPS settings
+        } else {
+            // TODO: HTTP isn't required for just gRPC servers. Is a null connector representative of that?
+            throw new ClassCastException("server.connector must be one of types " +
+                  ArmeriaHttpConnectorFactory.TYPE + " or " + ArmeriaHttpsConnectorFactory.TYPE);
```
cricket007: `connector` has a default value of `ArmeriaHttpConnectorFactory`... 

TODO: Try removing the default and using gRPC / Thrift only

examples/dropwizard-bundle/build.gradle
```diff
@@ -0,0 +1,11 @@
+plugins {
+    id 'application'
+}
+
+application {
+    mainClassName = 'example.dropwizard.DropwizardArmeriaApplication'
+}
```
cricket007: TODO: pass arguments for YAML config file

examples/dropwizard-bundle/src/main/java/example/dropwizard/DropwizardArmeriaApplication.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "dropwizard-armeria";
+    }
+
+    @Override
+    public void initialize(final Bootstrap<DropwizardArmeriaConfiguration> bootstrap) {
+        // TODO: application initialization
```
cricket007: TODO: Remove TODO comments from example

settings.gradle
```diff
@@ -50,6 +51,7 @@ includeWithFlags ':site'
 
 // Examples
 includeWithFlags ':examples:annotated-http-service', 'java'
+includeWithFlags ':examples:dropwizard',             'java'
```
cricket007: I would prefer the example module was `:examples:dropwizard`, but I had to rename the folder to `dropwizard-bundle` avoid a circular dependency on `project(':dropwizard')` for some reason?

I assume that the gradle module and folder name need to match, though

dropwizard/README.md
```diff
@@ -0,0 +1,108 @@
...
+This bundle introduces a new Dropwizard `ServerFactory` of type `armeria` that implements the same contract of the [`simple` ServerFactory](https://www.dropwizard.io/en/stable/manual/configuration.html#simple) as well as [HTTP(S) `ConnectorFactory`](https://www.dropwizard.io/en/stable/manual/configuration.html#http) implementations of `armeria-http` or `armeria-https`. 
+
+A minimal configuration file looks like so. 
+
+```yaml
+server:
+  type: armeria
+  connector:
+    type: armeria-http
+    port: 8080
```
cricket007: Currently, the `connector` property has a default, so this isn't the most minimal config

dropwizard/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+Support
+-------
+Please file bug reports and feature requests in [GitHub issues](https://github.com/line/armeria/issues).
+
+License
+-------
+This library is licensed under the Apache License, Version 2.0.
+
+See http://www.apache.org/licenses/LICENSE-2.0.html or the [LICENSE](../LICENSE.txt) file in this repository for the full license text.
```
cricket007: This section was ported from my repo. Feel free to comment to correct / remove.

dropwizard/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+
+
+dependencies {
+    compile project(':jetty')
```
cricket007: Dropwizard includes its own Jetty. Need to look into having exclusions.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,222 @@
...
+
+        return server;
+    }
+
+    private void setupArmeria(final Server server, final MetricRegistry metricRegistry) {
+        LOGGER.debug("Building Armeria Server");
+        final JettyService jettyService = JettyService.forServer(server);
+        this.serverBuilder = com.linecorp.armeria.server.Server.builder()
+                .meterRegistry(DropwizardMeterRegistries.newRegistry(metricRegistry))
+                .serviceUnder("/", jettyService);
```
cricket007: This alone appears to work for every Jersey resource as well as the AdminServlet added above here in `build()`... 

Not sure what happens when Jersey and Armeria have the same paths exposed

cricket007:
Master builds failing? 

`Task :examples:spring-boot-minimal:compileTestJava FAILED`

trustin:
> Master builds failing?

That is weird. Build passes at master. Let me check what is happening in this branch.

trustin:
My guess is that `dropwizard-bom` includes other dependencies than `dropwizard-`, interfering with Armeria's dependencies, such as Spring Boot 2. Could you remove `dropwizard-bom` and specify individual dependencies?

cricket007:
Am seeing this Exception when running my example module with common AccessLogWriter

```
WARN  [2019-11-07 03:52:13,183] com.linecorp.armeria.common.logging.RequestLogListenerInvoker: onRequestLog() failed with an exception:
! java.lang.ClassCastException: class com.linecorp.armeria.server.HttpServerHandler$EarlyRespondingRequestContext cannot be cast to class com.linecorp.armeria.server.ServiceRequestContext (com.linecorp.armeria.server.HttpServerHandler$EarlyRespondingRequestContext and com.linecorp.armeria.server.ServiceRequestContext are in unnamed module of loader 'app')
! at com.linecorp.armeria.server.logging.AccessLogger.write(AccessLogger.java:109)
! at com.linecorp.armeria.server.logging.AccessLogWriter.lambda$common$0(AccessLogWriter.java:38)
! at com.linecorp.armeria.common.logging.RequestLogListenerInvoker.invokeOnRequestLog(RequestLogListenerInvoker.java:38)
! at com.linecorp.armeria.common.logging.DefaultRequestLog.addListener(DefaultRequestLog.java:346)
! at com.linecorp.armeria.common.logging.DefaultRequestLog.addListener(DefaultRequestLog.java:322)
! at com.linecorp.armeria.server.HttpServerHandler.lambda$respond0$6(HttpServerHandler.java:608)
! at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:577)
! at io.netty.util.concurrent.DefaultPromise.notifyListeners0(DefaultPromise.java:570)
! at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:549)
! at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:490)
```

trustin:
> Am seeing this Exception when running my example module with common AccessLogWriter

Doh! Let me send a fix PR for that.

minwoox:
Oops, that was recently introduced by me. #2159 😅

cricket007:
Now getting NoSuchMethodError related to the Guava relocation. 

```
> Task :dropwizard:shadedTest
com.linecorp.armeria.server.dropwizard.ArmeriaServerFactoryTest > typesAreDiscoverable() FAILED
    java.lang.NoSuchMethodError: io.dropwizard.jackson.DiscoverableSubtypeResolver.getDiscoveredSubtypes()Lcom/linecorp/armeria/internal/shaded/guava/collect/ImmutableList;
        at com.linecorp.armeria.server.dropwizard.ArmeriaServerFactoryTest.typesAreDiscoverable(ArmeriaServerFactoryTest.java:45)

com.linecorp.armeria.server.dropwizard.ArmeriaConnectorFactoryTest > typesAreDiscoverable() FAILED
    java.lang.NoSuchMethodError: io.dropwizard.jackson.DiscoverableSubtypeResolver.getDiscoveredSubtypes()Lcom/linecorp/armeria/internal/shaded/guava/collect/ImmutableList;
        at com.linecorp.armeria.server.dropwizard.ArmeriaConnectorFactoryTest.typesAreDiscoverable(ArmeriaConnectorFactoryTest.java:49)

com.linecorp.armeria.server.dropwizard.ArmeriaConnectorFactoryTest > com.linecorp.armeria.server.dropwizard.ArmeriaConnectorFactoryTest$ArmeriaHttpsConnectorFactoryTest > withoutKeyStore_shouldNotValidate() FAILED
    java.lang.NoSuchMethodError: io.dropwizard.configuration.ConfigurationValidationException.getConstraintViolations()Lcom/linecorp/armeria/internal/shaded/guava/collect/ImmutableSet;
        at com.linecorp.armeria.server.dropwizard.ArmeriaConnectorFactoryTest$ArmeriaHttpsConnectorFactoryTest.withoutKeyStore_shouldNotValidate(ArmeriaConnectorFactoryTest.java:85)

com.linecorp.armeria.server.dropwizard.logging.AccessLogWriterFactoryTest > typesAreDiscoverable() FAILED
    java.lang.NoSuchMethodError: io.dropwizard.jackson.DiscoverableSubtypeResolver.getDiscoveredSubtypes()Lcom/linecorp/armeria/internal/shaded/guava/collect/ImmutableList;
        at com.linecorp.armeria.server.dropwizard.logging.AccessLogWriterFactoryTest.typesAreDiscoverable(AccessLogWriterFactoryTest.java:44)

11 tests completed, 4 failed

> Task :dropwizard:shadedTest FAILED
```

trustin:
> Now getting NoSuchMethodError related to the Guava relocation.

Ah, we need to exclude dropwizard packages for `ShadowTask` Gradle task. Let me send you a patch for this. Please stay in your IDE meanwhile. :bowing_man:

trustin:
@cricket007 A little bit complicated than imagined, but here's the fix:

```patch
diff --git dropwizard/build.gradle dropwizard/build.gradle
index 099afb9a2..12dfd7bd9 100644
--- dropwizard/build.gradle
+++ dropwizard/build.gradle
@@ -1,3 +1,12 @@
+buildscript {
+    repositories {
+        gradlePluginPortal()
+    }
+    dependencies {
+        classpath "com.github.jengelman.gradle.plugins:shadow:${managedVersions['com.github.jengelman.gradle.plugins:shadow']}"
+    }
+}
+
 dependencies {
     compile project(':jetty')
     // Dropwizard
@@ -9,3 +18,16 @@ dependencies {
 
     testCompile 'io.dropwizard:dropwizard-testing'
 }
+
+// Do not relocate Guava because it's part of Dropwizard's public API.
+[tasks.shadedJar, tasks.shadedTestJar].each { task ->
+    task.relocators.clear()
+    project.ext.relocations.each { Map<String, String> props ->
+        def from = props['from']
+        def to = props['to']
+        if (from in ['com.google.common', 'com.google.thirdparty.publicsuffix']) {
+            return
+        }
+        task.relocate from, to
+    }
+}
```

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
trustin: You could pass `Consumer<SslContextBuilder> tlsCustomizer` as the last parameter of `tls()`. `SslContextBuilder` exposes all TLS settings.

settings.gradle
trustin: It seems like it's a known issue in Gradle. You could name the project as `:examples:dropwizard-example` and specify the project directory explicitly. In `/settings.gradle`:

```java
includeWithFlags ':examples:dropwizard-example' 'java'
project(':examples:dropwizard-example').projectDir = file('examples/dropwizard')
```

See https://github.com/gradle/gradle/issues/847#issuecomment-450259535 for more information.

dropwizard/README.md
trustin: Question: Do Dropwizard users expect a Dropwizard server to have a default port of 8082? What is the most seamless behavior?

dropwizard/README.md
trustin: So far, all other modules did not have their own `README.md` but had a dedicated `.rst` page in the official site. Could we do the same for this module?

dropwizard/build.gradle
trustin: That's interesting. Does it ship its own Jetty fork, or does it depend on a specific Jetty version? Exclusion can be done like this:

```gradle
dependencies {
    compile(project(':jetty')) {
        exclude group: 'org.eclipse.jetty', module: 'jetty-server'
        exclude group: 'org.eclipse.jetty', module: '...'
    }
}
```

See https://stackoverflow.com/a/33931450/55808 for more information.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
trustin: The server added first will always have precedence. In an obvious case, it refuses to start and ask a user to fix the problem.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    /**
+     * Sets up the Armeria ServerBuilder with values from the Dropwizard Configuration.
+     * Ref <a href="https://line.github.io/armeria/advanced-production-checklist.html">Production Checklist</a>
+     *
+     * @param serverBuilder A non-production ready ServerBuilder
+     * @return A production-ready ServerBuilder
+     */
+    // visible for testing
```
trustin: You can use `@VisibleForTesting` annotation.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,311 @@
...
+import io.dropwizard.setup.Environment;
+import io.dropwizard.util.Size;
+import io.dropwizard.validation.MinSize;
+import io.micrometer.core.instrument.MeterRegistry;
+
+@JsonTypeName(ArmeriaServerFactory.TYPE)
+public class ArmeriaServerFactory extends SimpleServerFactory {
+
+    public static final String TYPE = "armeria";
+    private static final Logger LOGGER = LoggerFactory.getLogger(ArmeriaServerFactory.class);
```
trustin: nit: `logger` is used in our project.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,311 @@
...
+     */
+    // visible for testing
+    ServerBuilder decorateServerBuilderFromConfig(final ServerBuilder serverBuilder) {
+        Objects.requireNonNull(serverBuilder);
+        serverBuilder.maxNumConnections(getMaxQueuedRequests());
+        serverBuilder.blockingTaskExecutor(Executors.newFixedThreadPool(getMaxThreads()), true);
+        serverBuilder.maxRequestLength(maxRequestLength.toBytes());
+        serverBuilder.idleTimeoutMillis(getIdleThreadTimeout().toMilliseconds());
+        serverBuilder.gracefulShutdownTimeout(
+                Duration.ofSeconds(30L),
```
trustin: I guess you can use `getShotdownGracePeriod()` for both quiet period and timeout.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,311 @@
...
+     * Sets up the Armeria ServerBuilder with values from the Dropwizard Configuration.
+     * Ref <a href="https://line.github.io/armeria/advanced-production-checklist.html">Production Checklist</a>
+     *
+     * @param serverBuilder A non-production ready ServerBuilder
+     * @return A production-ready ServerBuilder
+     */
+    // visible for testing
+    ServerBuilder decorateServerBuilderFromConfig(final ServerBuilder serverBuilder) {
+        Objects.requireNonNull(serverBuilder);
+        serverBuilder.maxNumConnections(getMaxQueuedRequests());
```
trustin: There's no exact match for `getMaxQueuedRequests()` in Armeria at the moment.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,311 @@
...
+
+    @JsonGetter("disableServerHeader")
+    private boolean isServerHeaderDisabled() {
+        return disableServerHeader;
+    }
+
+    @JsonGetter("disableDateHeader")
+    public boolean isDateHeaderDisabled() {
+        return disableDateHeader;
+    }
```
trustin: How about renaming to `isServerHeaderEnabled()` and `isDateHeaderEnabled()` to avoid double negation which may be confusing? The default value could be `true` for both properties.

dropwizard/build.gradle
cricket007: It just has a transitive dependency on Jetty and other libraries - http://dropwizard.io/en/stable/getting-started.html

I know how to exclude, and I would rather exclude from Dropwizard if Armeria is going to take over that interaction. I meant I need to look if the versions are indeed different, or Gradle is smart enough to take care of that.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,311 @@
...
+
+    @JsonSetter("disableDateHeader")
+    public void disableDateHeaderDisabled(boolean disabled) {
+        this.disableDateHeader = disabled;
+    }
+
+    @JsonSetter("disableServerHeader")
+    public void setDisableServerHeader(boolean disabled) {
+        this.disableServerHeader = disabled;
+    }
```
trustin: `setDateHeaderEnabled()` and `setServerHeaderEnabled()`?

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,311 @@
...
+                Duration.ofMillis(getShutdownGracePeriod().toMilliseconds()));
+        if (isDateHeaderDisabled()) {
+            serverBuilder.disableDateHeader();
+        }
+        if (isServerHeaderDisabled()) {
+            serverBuilder.disableServerHeader();
+        }
+        serverBuilder.verboseResponses(hasVerboseResponses());
+        Optional.ofNullable(getDefaultHostname()).ifPresent(serverBuilder::defaultHostname);
+        // TODO: Add more items to server builder via Configuration
```
trustin: You might want to set the `MeterRegistry` of the server:

```java
serverBuilder.meterRegistry(DropwizardMeterRegistries.newRegistry(dropwizardMetricsRegistry));
```

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
trustin: The rule of thumb is to call the methods like `serviceUnder` as late as possible so that more specific routes are picked up first. In this case, you could add `jettyService` after all other Armeria services are bound.

dropwizard/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+A bundle for delegating application traffic through [Armeria](https://line.github.io/armeria/index.html) endpoints in Dropwizard applications.
+
+Dependency Info
+---------------
+```xml
+<dependency>
+    <groupId>com.linecorp.armeria</groupId>
+    <artifactId>armeria-dropwizard</artifactId>
+    <version>0.95.0</version>
```
cricket007: Should probably move all this to the RST so that the version can be templated

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ManagedArmeriaServer.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    public void stop() throws Exception {
+        if (server != null) {
+            LOGGER.info("Stopping Armeria Server");
+            server.stop().thenRunAsync(() -> LOGGER.info("Stopped Armeria Server"));
+        }
+    }
+
+    public interface BuilderCallback {
```
trustin: We prefer using a top-level class unless there is a very specific reason. Could we move this interface to top-level and name it like `ArmeriaServerConfigurator`, which is same with what our Spring Boot integration got?

dropwizard/README.md
cricket007: 8080 is the default. I don't have a compelling reason to keep it at 8082

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/connector/ArmeriaH2ConnectorFactory.java
```diff
@@ -0,0 +1,151 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+import com.linecorp.armeria.common.Flags;
+import com.linecorp.armeria.server.ServerBuilder;
+
+import io.dropwizard.jetty.ConnectorFactory;
+
+@JsonTypeName(ArmeriaH2ConnectorFactory.TYPE)
+public class ArmeriaH2ConnectorFactory extends ArmeriaHttpsConnectorFactory {
```
trustin: Is it necessary to have this `ConnectorFactory`? `ArmeriaHttpsConnectorFactory` already is capable of HTTP/2. Isn't it better merging this class into `ArmeriaHttpsConnectorFactory`?

examples/dropwizard-bundle/src/main/java/example/dropwizard/armeria/services/http/HelloService.java
```diff
@@ -0,0 +1,24 @@
...
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.server.annotation.Get;
+import com.linecorp.armeria.server.annotation.Produces;
+
+public class HelloService {
+
+    @Get("/hello")
+    @Produces("text/plain")
```
trustin: `@ProducesText`

examples/dropwizard-bundle/src/main/java/example/dropwizard/armeria/services/http/HelloService.java
```diff
@@ -0,0 +1,24 @@
...
+    public HttpResponse helloText() {
+        // Return a text document to the client.
+        return HttpResponse.of(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "Armeria");
+    }
+
+    @Get("/hello")
+    @Produces("application/json")
```
trustin: `@ProducesJson`

examples/dropwizard-bundle/src/main/java/example/dropwizard/armeria/services/http/HelloService.java
```diff
@@ -0,0 +1,24 @@
...
+import com.linecorp.armeria.server.annotation.Get;
+import com.linecorp.armeria.server.annotation.Produces;
+
+public class HelloService {
+
+    @Get("/hello")
+    @Produces("text/plain")
+    public HttpResponse helloText() {
+        // Return a text document to the client.
+        return HttpResponse.of(HttpStatus.OK, MediaType.PLAIN_TEXT_UTF_8, "Armeria");
```
trustin: `HttpResponse.of("Armeria")` or you can just return `"Armeria"` and change the return type to `String`.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/connector/proxy/ArmeriaHttpProxyConnectorFactory.java
```diff
@@ -0,0 +1,65 @@
...
+
+import io.dropwizard.jetty.ConnectorFactory;
+
+@JsonTypeName(ArmeriaHttpProxyConnectorFactory.TYPE)
+public class ArmeriaHttpProxyConnectorFactory extends ArmeriaProxyConnectorFactory {
+
+    public static final String TYPE = "armeria-http-proxy";
+    // TODO: This is not configurable
+    @JsonIgnore
+    private final ArmeriaServerDecorator innerFactory;
```
cricket007: Is a single SessionProtocol of PROXY not valid? I saw an error that HTTP or HTTPS needed to be included when I was doing `ServerBuilder.build()`

That being said, some refactoring would have to be done between the non-proxy connectors and these to handle duplicate ServerBuilder code 

One way I can think of this right now is to have a non-Proxy `ConnectorFactory` within the `ProxyConnectorFactory` 

Otherwise, PROXY support seems like a secondary feature to this whole PR

examples/dropwizard-bundle/src/main/java/example/dropwizard/armeria/services/http/HelloService.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    @Get("/hello")
+    @Produces("application/json")
+    public HttpResponse helloJson() {
+        // Return a JSON object to the client.
+        return HttpResponse.of(HttpStatus.OK, MediaType.JSON_UTF_8, "{ \"name\": \"Armeria\" }");
```
trustin: You might want to define a bean which has a property `name`, and let Jackson do the conversion for you.

site/src/sphinx/advanced-dropwizard-integration.rst
```diff
@@ -0,0 +1,121 @@
+.. _advanced-dropwizard-integration:
+
+Using Armeria with Dropwizard
+=================================
```
trustin: This is great! Thanks for spending your time on documentation. :heart:

dropwizard/README.md
trustin: Yes, please. That would be awesome.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/connector/proxy/ArmeriaProxyConnectorFactory.java
```diff
@@ -0,0 +1,82 @@
...
+                           @Nullable ThreadPool threadPool) {
+        return null; // un-used
+    }
+
+    @Override
+    public void decorate(final ServerBuilder sb) throws CertificateException, SSLException {
+        LOGGER.debug("Building Armeria Proxy Server");
+        sb.port(getPort(), getSessionProtocols())
+          .proxyProtocolMaxTlvSize((int) getMaxTlvSize().toBytes());
+        // are there other proxy settings?
```
cricket007: Nothing else stands out in the methods

dropwizard/src/test/java/com/linecorp/armeria/server/dropwizard/ArmeriaConnectorFactoryTest.java
```diff
@@ -0,0 +1,280 @@
...
+                    .isInstanceOf(Discoverable.class)
+                    .isInstanceOf(ArmeriaH2ConnectorFactory.class);
+        }
+
+        @Test
+        public void testFactoryDefaults() throws CertificateException, SSLException {
+            final String f1 = ResourceHelpers.resourceFilePath("f.crt");
+            final String f2 = ResourceHelpers.resourceFilePath("f.jks");
+            final ArmeriaH2ConnectorFactory factory =
+                    (ArmeriaH2ConnectorFactory) ArmeriaH2ConnectorFactory.build(f1, f2, null);
```
cricket007: Is there a way to generate legitimate JKS certificates for testing?

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
trustin: Alternatively, you could add the `DropwizardMeterRegistry` to `Metrics.globalRegistry`, so the metrics created via Micrometer API are also exported to Dropwizard Metrics:

```java
Metrics.globalRegistry.addRegistry(DropwizardMeterRegistries.newRegistry(...));
// No need to call `serverBuilder.meterRegistry()` because it defaults to `Metrics.globalRegistry`.
```

dropwizard/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+Usage
+-----
+Add an `ArmeriaBundle` to your [Application](https://javadoc.io/static/io.dropwizard/dropwizard-core/1.3.16/io/dropwizard/Application.html) class.
+
+```java
+@Override
+public void initialize(Bootstrap<MyConfiguration> bootstrap) {
+    // ...
+    final ArmeriaBundle bundle = new AbstractArmeriaBundle<DropwizardArmeriaConfiguration>() {
```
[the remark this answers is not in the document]
cricket007: Whoops. Renamed that

dropwizard/build.gradle
trustin: I see. I'd prefer using the Jetty version we provide, unless there's a conflict. Armeria usually depends on the latest version of Jetty, so I guess Gradle will choose what `armeria-jetty` depends on. To check the resolved dependency versions, you can use `./gradlew :dropwizard:dependencies`. (Let me know if you already know about this, so I don't provide you too much information. :wink:)

trustin:
I'm still looking into fixing the `ClassCastException` you mentioned. Please ignore it meanwhile.

dropwizard/README.md
```diff
@@ -0,0 +1,108 @@
...
+server:
+  type: armeria
+  accessLogWriter:
+    type: custom
+    format: "%d - sample - %s"
+```
+
+### Additional Connector properties
+
+The default `port` is 8082, to not interfere with Dropwizard's default application or admin servlets.
```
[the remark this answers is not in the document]
cricket007: I was having issues at first with Jetty's Connector class. Since I'm not having Dropwizard define that anymore, then setting the ports back to the default should be fine.

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
cricket007: Remove this then? Or is there a match for `maxNumConnections`?

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
cricket007: Already set it. 
https://github.com/line/armeria/pull/2236/files/2d5e7a08793cc21d655a114a45f887fc74b957ed#diff-6c4e58dd555e8837e0fc639603cd52ecR105-R107

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/connector/ArmeriaH2ConnectorFactory.java
cricket007: Sure. Can move the `.http2` methods over here. 

Or rename this to H2C

examples/dropwizard-bundle/src/main/java/example/dropwizard/armeria/services/http/HelloService.java
cricket007: 🤷‍♂ I was just copying the documentation here

site/src/sphinx/advanced-dropwizard-integration.rst
cricket007: Welcome! 📚

dropwizard/build.gradle
cricket007: Always looking to learn more Gradle after dealing with Maven during the day 🤣

examples/dropwizard-bundle/src/main/java/example/dropwizard/armeria/services/http/HelloService.java
trustin: Ah, no worries, then. :-)

trustin:
The `ClassCastException` has been fixed in `master`.

dropwizard/README.md
```diff
@@ -0,0 +1,40 @@
+### Additional `server` properties
+
+> ***Note:*** Not all Dropwizard configurations can be passed into the Armeria server.  Currently supported parameters are:   
+>   - `maxQueuedRequests`
+>   - `maxThreads`
+>   - `maxRequestLength`
+>   - `maxRequestLength`
```
trustin: Two `maxRequestLength`s

dropwizard/README.md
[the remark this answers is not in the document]
trustin: OK!

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaBundle.java
```diff
@@ -0,0 +1,40 @@
...
+package com.linecorp.armeria.server.dropwizard;
+
+import com.linecorp.armeria.server.ServerBuilder;
+
+import io.dropwizard.Configuration;
+import io.dropwizard.ConfiguredBundle;
+import io.dropwizard.setup.Bootstrap;
+import io.dropwizard.setup.Environment;
+
+public abstract class ArmeriaBundle<C extends Configuration>
```
trustin: Javadoc?

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
trustin: Couldn’t find one. Perhaps define one, like you did for server/dateHeader?

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
trustin: Ah, OK!

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/connector/proxy/ArmeriaHttpProxyConnectorFactory.java
trustin: > Is a single SessionProtocol of PROXY not valid? I saw an error that HTTP or HTTPS needed to be included when I was doing `ServerBuilder.build()`

Correct. PROXY can’t do anything without HTTP or HTTPS. I think it’s fine as long as 1) a user is allowed to specify PROXY and HTTP(S) connectors at the same port number and 2) the server fails to start up when only PROXY connector is specified.

If you are not familiar with Armeria’s multi-port configuration, we can skip PROXY support for now and consider it later.

```yaml

```

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/connector/proxy/ArmeriaProxyConnectorFactory.java
trustin: I think that is all at the protocol level.

dropwizard/src/test/java/com/linecorp/armeria/server/dropwizard/ArmeriaConnectorFactoryTest.java
trustin: So far we used `SelfSignedCertificateRule/Extension` but it doesn’t generate JKS but PEM.. I think it’s fine to use a pre-generated resource like you did.

dropwizard/src/test/resources/yaml/server/https-server-minimal.yaml
```diff
@@ -0,0 +1,5 @@
+type: armeria
+connector:
+  type: armeria-https
+  keyStorePath: /some/path/keystore.jks
+  keyStorePassword: changeme
```
trustin: nit: Please make sure all `.yaml` files’ last lines end with a newline.

dropwizard/build.gradle
cricket007: Looks like it pulls in the upgraded Jetty version

```
|    |    +--- io.dropwizard:dropwizard-lifecycle:1.3.16
|    |    |    +--- org.slf4j:slf4j-api:1.7.26 -> 1.7.28
|    |    |    +--- com.google.guava:guava:24.1.1-jre -> 28.1-jre (*)
|    |    |    +--- org.eclipse.jetty:jetty-server:9.4.18.v20190429 -> 9.4.22.v20191022 (*)
|    |    |    \--- io.dropwizard:dropwizard-util:1.3.16 (*)
```

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
cricket007: Btw, in `thrift0.9` it is `LOGGER`

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/ArmeriaServerFactory.java
trustin: Could you add a `TODO:` or `XXX:` about this?

site/src/sphinx/advanced-dropwizard-integration.rst
```diff
@@ -0,0 +1,260 @@
...
+
+    Not all Dropwizard configurations can be passed into the Armeria server.  Currently supported parameters are:
+
+    - ``maxThreads``
+    - ``maxRequestLength``
+    - ``idleThreadTimeout``
+    - ``shutdownGracePeriod``
+
+Where defined, the Armeria ServerFactory will prefer Armeria's default properties over Dropwizard's.
+The following additional properties are able to be added to configure the ``ServerBuilder`` before being
```
trustin: :api:`ServerBuilder`

dropwizard/src/main/java/com/linecorp/armeria/server/dropwizard/connector/proxy/ArmeriaHttpProxyConnectorFactory.java
cricket007: Leaving the Abstract proxy factory. Removed the others.

dropwizard/src/test/java/com/linecorp/armeria/server/dropwizard/ArmeriaConnectorFactoryTest.java
cricket007: I just touched empty files. Wasn't sure if there was a better way to make valid files.

examples/dropwizard/src/main/java/example/dropwizard/DropwizardArmeriaApplication.java
```diff
@@ -0,0 +1,51 @@
...
+        return "dropwizard-armeria";
+    }
+
+    @Override
+    public void initialize(final Bootstrap<DropwizardArmeriaConfiguration> bootstrap) {
+        final ArmeriaBundle bundle = new ArmeriaBundle() {
+            @Override
+            public void configure(final ServerBuilder builder) {
+                builder.service("/", (ctx, res) -> HttpResponse.of(MediaType.HTML_UTF_8, "<h2>It works!</h2>"));
+                builder.service("/armeria", (ctx, res) -> HttpResponse.of("Hello, Armeria!"));
```
cricket007: Not able to test this endpoint... 

Could make `ArmeriaConnector extend ServerConnector`, but that seemed to require a few more changes.

```class com.linecorp.armeria.server.jetty.ArmeriaConnector cannot be cast to class org.eclipse.jetty.server.ServerConnector (com.linecorp.armeria.server.jetty.ArmeriaConnector and org.eclipse.jetty.server.ServerConnector are in unnamed module of loader 'app')
java.lang.ClassCastException: class com.linecorp.armeria.server.jetty.ArmeriaConnector cannot be cast to class org.eclipse.jetty.server.ServerConnector (com.linecorp.armeria.server.jetty.ArmeriaConnector and org.eclipse.jetty.server.ServerConnector are in unnamed module of loader 'app')
	at io.dropwizard.testing.DropwizardTestSupport.getLocalPort(DropwizardTestSupport.java:249)
	at io.dropwizard.testing.junit5.DropwizardAppExtension.getLocalPort(DropwizardAppExtension.java:178)
	at example.dropwizard.DropwizardArmeriaApplicationTest.setUp(DropwizardArmeriaApplicationTest.java:49)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:675)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:125)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:132)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptLifecycleMethod(TimeoutExtension.java:111)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptBeforeAllMethod(TimeoutExtension.java:60)
	at org.junit.jupiter.engine.execution.ExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.lambda$invoke$0(ExecutableInvoker.java:105)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:104)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:62)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:43)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:35)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:104)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:98)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$invokeBeforeAllMethods$8(ClassBasedTestDescriptor.java:371)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.invokeBeforeAllMethods(ClassBasedTestDescriptor.java:369)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.before(ClassBasedTestDescriptor.java:193)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.before(ClassBasedTestDescriptor.java:77)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:132)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:125)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:135)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:122)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:80)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1540)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:125)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:135)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:122)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:80)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.submit(SameThreadHierarchicalTestExecutorService.java:32)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:57)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:51)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:229)
	at org.junit.platform.launcher.core.DefaultLauncher.lambda$execute$6(DefaultLauncher.java:197)
	at org.junit.platform.launcher.core.DefaultLauncher.withInterceptedStreams(DefaultLauncher.java:211)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:191)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:128)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor$CollectAllTestClassesExecutor.processAllTestClasses(JUnitPlatformTestClassProcessor.java:99)
	...
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:56)
	at java.base/java.lang.Thread.run(Thread.java:835)

example.dropwizard.DropwizardArmeriaApplicationTest > initializationError FAILED
    java.lang.ClassCastException: class com.linecorp.armeria.server.jetty.ArmeriaConnector cannot be cast to class org.eclipse.jetty.server.ServerConnector (com.linecorp.armeria.server.jetty.ArmeriaConnector and org.eclipse.jetty.server.ServerConnector are in unnamed module of loader 'app')
        at io.dropwizard.testing.DropwizardTestSupport.getLocalPort(DropwizardTestSupport.java:249)
        at io.dropwizard.testing.junit5.DropwizardAppExtension.getLocalPort(DropwizardAppExtension.java:178)
        at example.dropwizard.DropwizardArmeriaApplicationTest.setUp(DropwizardArmeriaApplicationTest.java:49)
1 test completed, 1 failed
> Task :examples:dropwizard-example:test FAILED```

Comment (trustin) [reply examples/dropwizard/src/main/java/example/dropwizard/DropwizardArmeriaApplication.java]: Doh! Let me check.

**Comment by trustin** [comment]
Didn't have a chance to look into the error yet. Will try this week. :bowing_man:

**Comment by cricket007** [comment]
Any updates on this one?

**Comment by trustin** [comment]
@cricket007 Let me take a look today. :pray:

**Comment by trustin** [comment]
@cricket007 I've just pushed a fix commit to your branch. PTAL

**Comment by cricket007** [comment]
Awesome, thanks! Looks like more tests need added for coverage?

**Comment by trustin** [comment]
> Awesome, thanks! Looks like more tests need added for coverage?

Yes, please! :bowing_man:

**Comment by cricket007** [comment]
Finally. Ready for review, again 👍

File: dropwizard/src/main/java/com/linecorp/armeria/dropwizard/ArmeriaServerFactory.java [hunk 39]
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    @Override
+    public ConnectorFactory getConnector() {
+        return connector;
+    }
+
+    @Override
+    public void setConnector(final ConnectorFactory factory) {
+        connector = Objects.requireNonNull(factory, "server.connector");
```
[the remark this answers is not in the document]
cricket007: This is saying that "server.connector" in the config was null, which happens to be an implementation of the connector factory

cricket007:
Thanks for the review 

By the way, Dropwizard 2.0 was just released. I'm not sure it makes sense to immediately jump to it, though. Maybe wait for any potential bugs to get worked out

trustin:
@cricket007 Is it possible to make all classes except `ArmeriaBundle` and `DropwizardArmeriaConfiguration` package-private? It seems to me that they are the only classes being used by a user in the example.

dropwizard/src/main/java/com/linecorp/armeria/dropwizard/connector/ArmeriaHttpConnectorFactory.java
```diff
@@ -0,0 +1,94 @@
...
+import com.linecorp.armeria.server.ServerBuilder;
+
+import io.dropwizard.jetty.ConnectorFactory;
+import io.dropwizard.jetty.HttpConnectorFactory;
+import io.dropwizard.util.Size;
+import io.dropwizard.validation.MaxSize;
+import io.dropwizard.validation.MinSize;
+
+@JsonTypeName(ArmeriaHttpConnectorFactory.TYPE)
+public class ArmeriaHttpConnectorFactory extends HttpConnectorFactory
```
minwoox: I think we don't have to extend `HttpConnectorFactory`. The user who sets the properties in `HttpConnectorFactory` wouldn't know that the server is not working as expected right away.
So how about introducing a new class `ArmeriaConnectorFactory` and share the logic from `ArmeriaHttpConnectorFactory` and `ArmeriaHttpsConnectorFactory `? In that case, we can remove `ArmeriaServerDecorator`  as well. 😄

cricket007: Could. But then, I'd just be copying [large chunks of code out of `HttpConnectorFactory`](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jetty/src/main/java/io/dropwizard/jetty/HttpConnectorFactory.java#L607-L700) and same for HttpsConnectorFactory

cricket007:
> Is it possible to make all classes package-private
> @trustin 

ArmeriaServerFactory imports these, so they cannot be package private

```
import com.linecorp.armeria.dropwizard.connector.ArmeriaHttpConnectorFactory;
import com.linecorp.armeria.dropwizard.connector.ArmeriaServerDecorator;
import com.linecorp.armeria.dropwizard.logging.AccessLogWriterFactory;
import com.linecorp.armeria.dropwizard.logging.CommonAccessLogWriterFactory;
```

dropwizard/src/main/java/com/linecorp/armeria/dropwizard/connector/ArmeriaHttpConnectorFactory.java
minwoox: I'm asking maybe I don't know much about how `Dropwizard` works 😄 but in the code that you gave to me, `HttpConnectorFactory` is making a new `ServerConnector` which has its own `acceptor`, `acceptChannel` and etc.
So are we accepting the connection using the `Connector`, not Armeria?

cricket007: I actually don't know that level of dropwizard myself.

If I did implement ConnectorFactory instead, how would I go about implementing the Connector class from armeria-jetty for the linked buildConnector method?

minwoox: I Don't know either. 😆 
I think we can use some classes in the jetty module and we can build our own connector factory so that Armeria server accepts all the connections. Just give me some time to look at the code in Jetty module and Dropwizard please. 😄

minwoox: > Could. But then, I'd just be copying large chunks of code out of HttpConnectorFactory  and same for HttpsConnectorFactory

So I found out that the code you linked which is  `HttpConnectorFactory.build()` is not called because we don't build the `connector` and add the `connector` to the `server` using `server.addConnector(connector)`.

But we have to support the configuration [properties](https://www.dropwizard.io/en/stable/manual/configuration.html) (even though we cannot support all of them), so I think we can just leave it as it is and add more properties that Armeria provides.

One thing I found is that if we add a fallback service, it doesn't work properly:
```java
public void configure(final ServerBuilder builder) {
    ....
    builder.serviceUnder("/", (ctx, res) -> HttpResponse.of("fall back!"));
}
```
It's because of [this](https://github.com/line/armeria/pull/2236/files#diff-2826258b9b33a09bf523f114e73b71faR280) so the requests which are matched by the router get into the `jettyService`. But I think we can handle it in the next PR because I think we might need to add and refactor our jetty module.

So, yes I think we are on the right way to go and I will do the rest of the review. Thanks and good job @cricket007!

jetty/src/main/java/com/linecorp/armeria/server/jetty/ArmeriaConnector.java
```diff
@@ -16,65 +16,42 @@
...
-    private final ArmeriaConnectionFactory connectionFactory;
-    private final Collection<ConnectionFactory> connectionFactories;
 
     private volatile boolean isShutdown;
 
-    ArmeriaConnector(Server server) {
-        this.server = server;
-        executor = server.getThreadPool();
+    ArmeriaConnector(Server server, com.linecorp.armeria.server.Server armeriaServer) {
+        super(server, 0, 0, new ArmeriaConnectionFactory());
```
ikhoon: I found that `-1` is default value for `acceptors`, and `selectors`.
If you are not intended, how about passing `-1` to super method arguments?
 
https://github.com/eclipse/jetty.project/blob/fcc18b05300f0e3a9e357da2b9509f2768fc7011/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java#L94-L104

dropwizard/src/main/java/com/linecorp/armeria/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,335 @@
...
+        final ContextRoutingHandler routingHandler = new ContextRoutingHandler(ImmutableMap.of(
+                getApplicationContextPath(), applicationHandler,
+                getAdminContextPath(), adminHandler));
+        final Handler gzipHandler = buildGzipHandler(routingHandler);
+        server.setHandler(addStatsHandler(addRequestLog(server, gzipHandler, environment.getName())));
+    }
+
+    private ServerBuilder getArmeriaServerBuilder(final Server server,
+                                                  @Nullable final ConnectorFactory connector,
+                                                  @Nullable final MetricRegistry metricRegistry) {
```
minwoox: These two fields are not nullable.
Could you also remove the `final` keyword from the parameter declaration of all methods, please? 😄

dropwizard/src/main/java/com/linecorp/armeria/dropwizard/ArmeriaServerFactory.java
```diff
@@ -0,0 +1,335 @@
...
+        logger.debug("Building Armeria Server");
+        final ServerBuilder serverBuilder = com.linecorp.armeria.server.Server.builder();
+        try {
+            decorateServerBuilder(
+                    serverBuilder, connector, accessLogWriter,
+                    metricRegistry != null ? DropwizardMeterRegistries.newRegistry(metricRegistry) : null);
```
minwoox: So this could be just `metricRegistry`

dropwizard/src/main/java/com/linecorp/armeria/dropwizard/logging/CommonAccessLogWriterFactory.java
```diff
@@ -0,0 +1,32 @@
...
+package com.linecorp.armeria.dropwizard.logging;
+
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+import com.linecorp.armeria.server.logging.AccessLogWriter;
+
+@JsonTypeName("common")
+public class CommonAccessLogWriterFactory implements AccessLogWriterFactory {
+
+    public CommonAccessLogWriterFactory() {
```
minwoox: Question: Do we need no-arg constructor for any reason?

minwoox:
I think we are almost there. Thank you!

> ArmeriaServerFactory imports these, so they cannot be package private

Then, could you add Javadoc to the public classes and methods, please?

jetty/src/main/java/com/linecorp/armeria/server/jetty/ArmeriaConnector.java
trustin: IIRC 0 and -1 have different meaning, but it may not matter anymore since we override most methods.

cricket007:
Looks like spring is failing 

```
> Task :examples:spring-boot-minimal:compileTestJava FAILED
> Task :examples:spring-boot-tomcat:compileTestJava FAILED
```

dropwizard/src/test/java/com/linecorp/armeria/dropwizard/ArmeriaServerFactoryTest.java
```diff
@@ -0,0 +1,226 @@
...
+        final ServerBuilder sb = com.linecorp.armeria.server.Server.builder()
+                                                                   .service("/", (ctx, req) -> HttpResponse
+                                                                           .of(HttpStatus.OK));
+
+        final ClassCastException ex =
+                assertThrows(ClassCastException.class,
+                             // decorate with an instance that does not implement ArmeriaServerDecorator
+                             () -> factory.decorateServerBuilder(sb,
+                                                                 (server, metrics, name, threadPool) -> null,
+                                                                 null, null));
```
minwoox: We should fix not to pass `null` for the `meterRegistry`. 😄

cricket007:
Thanks again, everyone!

trustin:
:tada::tada::tada:

Thanks, @cricket007!

PR #2240, opened 2019-11-08
Make `DynamicEndpointGroup` respect a weight change

trustin:
Motivation:

`DynamicEndpointGroup` currently does not accept an update request which
changes only the weight of an existing `Endpoint`.

Modifications:

- Fix `DynamicEndpointGroup` so that it accepts the case where only
  weights are changed.
- Improve `Endpoint.hashCode()` so it does not invoke `authority()`
  unnecessarily.

Result:

- `DynamicEndpointGroup` always notifies its listeners when meaningful
  changes are detected.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
@@ -97,7 +97,7 @@ public class DynamicEndpointGroup extends AbstractListenable<List<Endpoint>> imp
         final List<Endpoint> oldEndpoints = this.endpoints;
         final List<Endpoint> newEndpoints = ImmutableList.sortedCopyOf(endpoints);
 
-        if (oldEndpoints != UNINITIALIZED_ENDPOINTS && !hasChanges(oldEndpoints, newEndpoints)) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/DynamicEndpointGroup.java
```diff
@@ -112,6 +112,22 @@ protected final void setEndpoints(Iterable<Endpoint> endpoints) {
...
     }
 
+    private static boolean hasChanges(List<Endpoint> oldEndpoints, List<Endpoint> newEndpoints) {
+        if (oldEndpoints.size() != newEndpoints.size()) {
+            return true;
+        }
+
+        for (int i = 0; i < oldEndpoints.size(); i++) {
+            final Endpoint a = oldEndpoints.get(i);
+            final Endpoint b = newEndpoints.get(i);
```
ikhoon: Is there any chance that the `newEnpoints.size()` is smaller than `oldEndpoints.size()`?

trustin: Nope, because we made sure both lists have the same size.

trustin:
Thanks for reviewing. :bowing_man:

PR #2241, opened 2019-11-08
Make 'none+' serialization format as optional.

### LINKED ISSUE #2219: Accept URIs without `none+` and return URIs without `none+`
I see usability issues with `none+` prefix in URI scheme:

- A user must specify a URL that starts with `none+` to create an `HttpClient` with a `ClientBuilder`.
  - We could assume `none+` exists when a user did not prepend it.
  - It becomes a little bit more ambiguous, I think it will not be that confusing. It's like removing 0 from 0.1 (= .1).
- `HttpClient.uri()` returns a URI that starts with `none+`, which is somewhat confusing. For example, a user cannot convert the returned URI into a URL:
  ```java
  HttpClient client = ...;
  client.uri().toURL(); // Fail
  ```

I'd like to:

- Make the `none+` prefix optional when creating a client.
- Make `ClientBuildParams.uri()` return a URI without the `none+` prefix.

rmohta:
Motivation:

Improve API UX, by not forcing application to provide *none+* serialization format.

Modifications:

Update AbstractClientFactory.validateScheme(URI), to default to _none+_, if no serialization format is passed.
Update DefaultClientFactory to normalize URI, by removing _none+_ from URI scheme.

Result:

Fixes #2219

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
@@ -19,14 +19,12 @@ package com.linecorp.armeria.client;
 import static java.util.Objects.requireNonNull;
 
 import java.net.URI;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
 import javax.annotation.Nullable;
 
 import com.linecorp.armeria.common.Scheme;
-import com.linecorp.armeria.common.SerializationFormat;
 
 /**
  * A skeletal {@link ClientFactory} implementation.
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
@@ -98,20 +98,17 @@ public abstract class AbstractClientFactory implements ClientFactory {
     protected final Scheme validateScheme(URI uri) {
         requireNonNull(uri, "uri");
 
-        if (Objects.isNull(uri.getAuthority())) {
+        if (uri.getAuthority() == null) {
             throw new IllegalArgumentException("URI with missing authority: " + uri);
         }
 
         final String scheme = uri.getScheme();
-        if (Objects.isNull(scheme)) {
+        if (scheme == null) {
             throw new IllegalArgumentException("URI with missing scheme: " + uri);
         }
-        Optional<Scheme> parsedSchemeOpt = Scheme.tryParse(scheme);
+        final Optional<Scheme> parsedSchemeOpt = Scheme.tryParse(scheme);
         if (!parsedSchemeOpt.isPresent()) {
-            parsedSchemeOpt = Scheme.tryParse(SerializationFormat.NONE.uriText() + '+' + scheme);
-            if (!parsedSchemeOpt.isPresent()) {
-                throw new IllegalArgumentException("URI with undefined scheme: " + uri);
-            }
+            throw new IllegalArgumentException("URI with undefined scheme: " + uri);
         }
 
         final Scheme parsedScheme = parsedSchemeOpt.get();
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
@@ -19,12 +19,14 @@ package com.linecorp.armeria.client;
 import static java.util.Objects.requireNonNull;
 
 import java.net.URI;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
 import javax.annotation.Nullable;
 
 import com.linecorp.armeria.common.Scheme;
+import com.linecorp.armeria.common.SerializationFormat;
 
 /**
  * A skeletal {@link ClientFactory} implementation.
@@ -96,18 +98,20 @@ public abstract class AbstractClientFactory implements ClientFactory {
     protected final Scheme validateScheme(URI uri) {
         requireNonNull(uri, "uri");
 
-        final String scheme = uri.getScheme();
-        if (scheme == null) {
-            throw new IllegalArgumentException("URI with missing scheme: " + uri);
-        }
-
-        if (uri.getAuthority() == null) {
+        if (Objects.isNull(uri.getAuthority())) {
             throw new IllegalArgumentException("URI with missing authority: " + uri);
         }
 
-        final Optional<Scheme> parsedSchemeOpt = Scheme.tryParse(scheme);
+        final String scheme = uri.getScheme();
+        if (Objects.isNull(scheme)) {
+            throw new IllegalArgumentException("URI with missing scheme: " + uri);
+        }
+        Optional<Scheme> parsedSchemeOpt = Scheme.tryParse(scheme);
         if (!parsedSchemeOpt.isPresent()) {
-            throw new IllegalArgumentException("URI with unknown scheme: " + uri);
+            parsedSchemeOpt = Scheme.tryParse(SerializationFormat.NONE.uriText() + '+' + scheme);
+            if (!parsedSchemeOpt.isPresent()) {
+                throw new IllegalArgumentException("URI with undefined scheme: " + uri);
+            }
         }
 
         final Scheme parsedScheme = parsedSchemeOpt.get();
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
@@ -21,8 +21,6 @@ import java.net.URI;
 
 import com.google.common.base.MoreObjects;
 
-import com.linecorp.armeria.common.SerializationFormat;
-
 /**
  * Default {@link ClientBuilderParams} implementation.
  */
@@ -39,7 +37,7 @@ public class DefaultClientBuilderParams implements ClientBuilderParams {
     public DefaultClientBuilderParams(ClientFactory factory, URI uri, Class<?> type,
                                       ClientOptions options) {
         this.factory = requireNonNull(factory, "factory");
-        this.uri = cleanUri(requireNonNull(uri, "uri"));
+        this.uri = requireNonNull(uri, "uri");
         this.type = requireNonNull(type, "type");
         this.options = requireNonNull(options, "options");
     }
@@ -71,15 +69,5 @@ public class DefaultClientBuilderParams implements ClientBuilderParams {
                           .add("uri", uri)
                           .add("type", type)
                           .add("options", options).toString();
-    }
-
-    protected final URI cleanUri(final URI uri) {
-        if (!uri.getScheme().contains(SerializationFormat.NONE.uriText())) {
-            return uri;
-        }
-        final String newScheme = uri.getScheme()
-                                  .replace(SerializationFormat.NONE.uriText(), "")
-                                  .replace("+", "");
-        return URI.create(newScheme + uri.toString().substring(uri.getScheme().length()));
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
@@ -77,9 +77,9 @@ public class DefaultClientBuilderParams implements ClientBuilderParams {
         if (!uri.getScheme().contains(SerializationFormat.NONE.uriText())) {
             return uri;
         }
-        final String nScheme = uri.getScheme()
+        final String newScheme = uri.getScheme()
                                   .replace(SerializationFormat.NONE.uriText(), "")
                                   .replace("+", "");
-        return URI.create(nScheme + uri.toString().substring(uri.getScheme().length()));
+        return URI.create(newScheme + uri.toString().substring(uri.getScheme().length()));
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
@@ -74,11 +74,12 @@ public class DefaultClientBuilderParams implements ClientBuilderParams {
     }
 
     protected final URI cleanUri(final URI uri) {
-        if (uri.getScheme().contains(SerializationFormat.NONE.uriText())) {
-            final String nScheme = uri.getScheme().replace(SerializationFormat.NONE.uriText(), "")
-                                      .replace("+", "");
-            return URI.create(nScheme + uri.toString().substring(uri.getScheme().length()));
+        if (!uri.getScheme().contains(SerializationFormat.NONE.uriText())) {
+            return uri;
         }
-        return uri;
+        final String nScheme = uri.getScheme()
+                                  .replace(SerializationFormat.NONE.uriText(), "")
+                                  .replace("+", "");
+        return URI.create(nScheme + uri.toString().substring(uri.getScheme().length()));
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
@@ -21,6 +21,8 @@ import java.net.URI;
 
 import com.google.common.base.MoreObjects;
 
+import com.linecorp.armeria.common.SerializationFormat;
+
 /**
  * Default {@link ClientBuilderParams} implementation.
  */
@@ -37,7 +39,7 @@ public class DefaultClientBuilderParams implements ClientBuilderParams {
     public DefaultClientBuilderParams(ClientFactory factory, URI uri, Class<?> type,
                                       ClientOptions options) {
         this.factory = requireNonNull(factory, "factory");
-        this.uri = requireNonNull(uri, "uri");
+        this.uri = cleanUri(requireNonNull(uri, "uri"));
         this.type = requireNonNull(type, "type");
         this.options = requireNonNull(options, "options");
     }
@@ -69,5 +71,14 @@ public class DefaultClientBuilderParams implements ClientBuilderParams {
                           .add("uri", uri)
                           .add("type", type)
                           .add("options", options).toString();
+    }
+
+    protected final URI cleanUri(final URI uri) {
+        if (uri.getScheme().contains(SerializationFormat.NONE.uriText())) {
+            final String nScheme = uri.getScheme().replace(SerializationFormat.NONE.uriText(), "")
+                                      .replace("+", "");
+            return URI.create(nScheme + uri.toString().substring(uri.getScheme().length()));
+        }
+        return uri;
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
@@ -16,6 +16,8 @@
 
 package com.linecorp.armeria.client;
 
+import static com.linecorp.armeria.client.HttpClientBuilder.isUndefinedUri;
+
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
@@ -175,7 +177,7 @@ final class DefaultClientFactory extends AbstractClientFactory {
     }
 
     private URI normalizeUri(URI uri, Scheme scheme) {
-        if (HttpClientBuilder.isUndefinedUri(uri)) {
+        if (isUndefinedUri(uri)) {
             // See https://github.com/line/armeria/pull/2241 discussion for more details.
             return uri;
         }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
@@ -175,7 +175,7 @@ final class DefaultClientFactory extends AbstractClientFactory {
     }
 
     private URI normalizeUri(URI uri, Scheme scheme) {
-        if (uri == HttpClientBuilder.UNDEFINED_URI) {
+        if (HttpClientBuilder.isUndefinedUri(uri)) {
             // See https://github.com/line/armeria/pull/2241 discussion for more details.
             return uri;
         }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
@@ -174,7 +174,7 @@ final class DefaultClientFactory extends AbstractClientFactory {
         clientFactoriesToClose.forEach(ClientFactory::close);
     }
 
-    private URI normalizeUri(final URI uri, final Scheme scheme) {
+    private URI normalizeUri(URI uri, Scheme scheme) {
         if (scheme.serializationFormat() != SerializationFormat.NONE) {
             return uri;
         }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
@@ -188,9 +188,7 @@ final class DefaultClientFactory extends AbstractClientFactory {
             return uri;
         }
 
-        final String newScheme = uri.getScheme()
-                                    .replace(SerializationFormat.NONE.uriText(), "")
-                                    .replace("+", "");
-        return URI.create(newScheme + uri.toString().substring(uri.getScheme().length()));
+        return URI.create(scheme.sessionProtocol() + "://" +
+                          uri.toString().substring(uri.getScheme().length()));
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
@@ -178,9 +178,12 @@ final class DefaultClientFactory extends AbstractClientFactory {
 
     private URI normalizeUri(URI uri, Scheme scheme) {
         if (isUndefinedUri(uri)) {
-            // See https://github.com/line/armeria/pull/2241 discussion for more details.
+            // We use a special singleton marker URI for clients that do not explicitly define a
+            // host or scheme at construction time.
+            // As this isn't created by users, we don't need to normalize it.
             return uri;
         }
+
         if (scheme.serializationFormat() != SerializationFormat.NONE) {
             return uri;
         }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
@@ -35,6 +35,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Streams;
 
 import com.linecorp.armeria.common.Scheme;
+import com.linecorp.armeria.common.SerializationFormat;
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.util.ReleasableHolder;
 
@@ -128,6 +129,7 @@ final class DefaultClientFactory extends AbstractClientFactory {
     @Override
     public <T> T newClient(URI uri, Class<T> clientType, ClientOptions options) {
         final Scheme scheme = validateScheme(uri);
+        uri = normalizeUri(uri, scheme);
         return clientFactories.get(scheme).newClient(uri, clientType, options);
     }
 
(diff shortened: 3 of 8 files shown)


core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -96,18 +98,20 @@
     protected final Scheme validateScheme(URI uri) {
         requireNonNull(uri, "uri");
 
-        final String scheme = uri.getScheme();
-        if (scheme == null) {
-            throw new IllegalArgumentException("URI with missing scheme: " + uri);
-        }
-
-        if (uri.getAuthority() == null) {
+        if (Objects.isNull(uri.getAuthority())) {
```
trustin: `Objects.isNull()` could be replaced with ` == null` for simplicity.

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -96,18 +98,20 @@
...
             throw new IllegalArgumentException("URI with missing authority: " + uri);
         }
 
-        final Optional<Scheme> parsedSchemeOpt = Scheme.tryParse(scheme);
+        final String scheme = uri.getScheme();
+        if (Objects.isNull(scheme)) {
```
trustin: Ditto - could be `== null`

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -96,18 +98,20 @@
...
+            throw new IllegalArgumentException("URI with missing scheme: " + uri);
+        }
+        Optional<Scheme> parsedSchemeOpt = Scheme.tryParse(scheme);
```
trustin: Instead of implementing the fallback logic here, how about modifying `Scheme.tryParse()` and `Scheme.parse()` so they accept a scheme without `none+`?

core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
```diff
@@ -70,4 +72,14 @@ public String toString() {
                           .add("type", type)
                           .add("options", options).toString();
     }
+
+    protected final URI cleanUri(final URI uri) {
+        if (!uri.getScheme().contains(SerializationFormat.NONE.uriText())) {
```
trustin: How about this:

```java
final String nonePrefix = SerializationFormat.NONE.uriText() + '+';
if (!uri.getScheme().startsWith(nonePrefix)) { ... }
```

core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
```diff
@@ -70,4 +72,14 @@ public String toString() {
...
+            return uri;
+        }
+        final String nScheme = uri.getScheme()
+                                  .replace(SerializationFormat.NONE.uriText(), "")
+                                  .replace("+", "");
```
trustin: - `nScheme` -> `newScheme`
- `uri.getScheme().substring(nonePrefix.length)`?

core/src/main/java/com/linecorp/armeria/common/Scheme.java
```diff
@@ -75,10 +76,9 @@
      *         there is no such {@link Scheme} available
      */
     public static Optional<Scheme> tryParse(@Nullable String scheme) {
-        if (scheme == null) {
+        if (Objects.isNull(scheme)) {
```
trustin: Revert?

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
rmohta: Done. Habit of working on my code, where we prefer isNull() for readability :-)

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
rmohta: Done

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
rmohta: Done. I have also removed some duplicate logic from Scheme.parse(). Now Scheme.parse() invoked tryParse(), and returns the value.

core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
rmohta: I thought serialization protocol could be *none+http* or *http+none*, based on Scheme.SCHEMES map. Hence I had not checked for starts with, and replaced specific sequences.

core/src/main/java/com/linecorp/armeria/common/Scheme.java
rmohta: done

core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
trustin: Ah, OK. Then how about reconstructing the URI at `DefaultClientFactory.newClient(URI, Class, ClientOptions)` if `uri.getScheme()` is not equal to the validated `scheme.uriText()`, so that the user-specified URI always has normalized scheme?

core/src/main/java/com/linecorp/armeria/common/Scheme.java
```diff
@@ -78,8 +78,13 @@
...
         }
-
-        return Optional.ofNullable(SCHEMES.get(Ascii.toLowerCase(scheme)));
+        final Optional<Scheme> parsedScheme = Optional.ofNullable(SCHEMES.get(Ascii.toLowerCase(scheme)));
+        if (parsedScheme.isPresent()) {
+            return parsedScheme;
+        }
+        return Optional.ofNullable(SCHEMES.get(Ascii.toLowerCase(SerializationFormat.NONE.uriText() +
+                                                                 '+' +
+                                                                 scheme)));
```
trustin: The two lines could be merged?

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
trustin: Thanks! Looking nicer.

core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
```diff
@@ -70,4 +72,14 @@ public String toString() {
                           .add("type", type)
                           .add("options", options).toString();
     }
+
+    protected final URI cleanUri(final URI uri) {
```
ikhoon: Could be private?
```suggestion
    private URI normalizeUri(URI uri) {
```

core/src/main/java/com/linecorp/armeria/common/Scheme.java
ikhoon: `SerializationFormat.NONE.uriText()` is defined in lowercase. How about reusing lowercased scheme? 😀
```java
final String lowerCaseScheme = Ascii.toLowerCase(scheme);
final Optional<Scheme> parsedScheme = Optional.ofNullable(SCHEMES.get(lowerCaseScheme));
if (parsedScheme.isPresent()) {
    return parsedScheme;
}
return Optional.ofNullable(SCHEMES.get(SerializationFormat.NONE.uriText() +
                                                                 '+' +
                                                                 lowerCaseScheme)));
```

core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
rmohta: That's a good idea. I have moved it to DefaultClientFactory. And I am using the derived Scheme to check for SerializationFormat.None (instead of uriText comparison).

core/src/main/java/com/linecorp/armeria/common/Scheme.java
rmohta: Done

rmohta: Done.

core/src/main/java/com/linecorp/armeria/client/DefaultClientBuilderParams.java
rmohta: Done.

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -38,13 +38,13 @@
...
      */
-    private static final URI UNDEFINED_URI = URI.create("none+http://undefined");
+    private static final URI UNDEFINED_URI = URI.create("http://undefined");
 
     /**
      * Returns {@code true} if the specified {@code uri} is an undefined {@link URI}.
      */
     static boolean isUndefinedUri(URI uri) {
-        return UNDEFINED_URI == uri;
+        return UNDEFINED_URI.equals(uri);
```
ikhoon: Revert this. The reference comparison is intended. 😅

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
```diff
@@ -38,13 +38,13 @@
     /**
      * An undefined {@link URI} to create {@link HttpClient} without specifying {@link URI}.
      */
-    private static final URI UNDEFINED_URI = URI.create("none+http://undefined");
+    private static final URI UNDEFINED_URI = URI.create("http://undefined");
```
ikhoon: Revert?

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
```diff
@@ -171,4 +173,15 @@ public void close() {
     void doClose() {
         clientFactoriesToClose.forEach(ClientFactory::close);
     }
+
+    private URI normalizeUri(final URI uri, final Scheme scheme) {
```
ikhoon: Could you remove final from the arguments? 😀

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
rmohta: See above comment regarding isUndefinedUri change

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
rmohta: If I revert this, then isUndefinedUri(uri) will always return false, as == is comparing if they are exactly the same instance.
Due to this change (normalizeURI method), we will create __a new instance of URI__ by removing none+ from the URI. 
Also, I stumbled on this - because 2 test cases were failing. One of them was HttpClientRequestPathTest.default_withRelativePath. And one more from same test class (sorry don't remember the method name)

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
rmohta: Done. Sorry didn't know this convention :-)  
Would you be open, if I open another PR later to add this as a checkstyle check (to prevent this in future)?

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
ikhoon: Thanks for the detail explanations. 🙏 I thought you encountered a problem and fixed this.

How about skipping URI normalization on `UNDEFINDED`? Because it is used only internally. 😀

core/src/main/java/com/linecorp/armeria/common/Scheme.java
```diff
@@ -90,12 +94,11 @@
...
-        final Scheme res = SCHEMES.get(Ascii.toLowerCase(requireNonNull(scheme, "scheme")));
-        if (res == null) {
-            throw new IllegalArgumentException("scheme: " + scheme);
+        final Optional<Scheme> parsedScheme = tryParse(requireNonNull(scheme, "scheme"));
+        if (parsedScheme.isPresent()) {
+            return parsedScheme.get();
         }
-
-        return res;
+        throw new IllegalArgumentException("scheme: " + scheme);
```
ikhoon: nit: How about using `Optional.orElseThrow(...)`? 😉

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
rmohta: @ikhoon I think that might be an additional complexity - additional branch - for an edge case (I mean, how often will some one create a client and not specify the hostname/authority).

If you are certain, we would have to do - I am more than happy to make the change.

core/src/main/java/com/linecorp/armeria/common/Scheme.java
rmohta: Done

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
ikhoon: > I think that might be an additional complexity - additional branch - for an edge case 

I understand what you are worried about. But `UNDEFINED_URI` is created to build `HttpClient` without hostname. #1143 #1961

I'm still sure to revert this. :-)

rmohta: Cool! Thank you for the links.
I have made required changes, to handle special case around__UNDEFINED_URI__

core/src/main/java/com/linecorp/armeria/client/HttpClientBuilder.java
rmohta: I have made changes to handle UNDEFINED_URI, but haven't reverted the value of constant. It's because, per this PR - we don't want uri() to return _none+_. 
I think, it would not be idiomatic, to have _none+_ returned for this special case. We are doing an == check for this constant though.

rmohta:
> nit: Remove `public` here and the following methods?

Done. Removed in ClientBuilderTest too.

ikhoon:
> Would you be open, if I open another PR later to add this as a checkstyle check (to prevent this in future)?

We had chat and decided to need more discussion on this issue. I filed an issue for that. #2247 
Thank you for nice suggestion and if you have a good idea, leave your opinion on it :-)

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
```diff
@@ -171,4 +173,19 @@ public void close() {
     void doClose() {
         clientFactoriesToClose.forEach(ClientFactory::close);
     }
+
+    private URI normalizeUri(URI uri, Scheme scheme) {
+        if (uri == HttpClientBuilder.UNDEFINED_URI) {
```
ikhoon: Could you revert https://github.com/line/armeria/pull/2241/files#diff-8db0b28d88b37c5051e9775d4df020bfL41 and use `isUndefinedUri(URI uri)`? We want to hide `UNDEFINED_URI` instance from outside. 😉
```suggestion
        if (isUndefinedUri(uri)) {
```

ikhoon:
> I have made changes to handle UNDEFINED_URI, but haven't reverted the value of constant. It's because, per this PR - we don't want uri() to return none+.
I think, it would not be idiomatic, to have none+ returned for this special case. We are doing an == check for this constant though.
WDYT?

That is a nice idea! Removing `none+` from `UNDEFINED_URI` is good for me too. Thanks for the detail :-)

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
```diff
@@ -171,4 +173,19 @@ public void close() {
...
+            // See https://github.com/line/armeria/pull/2241 discussion for more details.
```
ikhoon: 👍

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
rmohta: Done.

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
anuraaga: The PR is pretty long and I don't think it's practical to reference it to find this detail. We can just inline the explanation

```
// We use a special singleton marker URI for clients that do not explicitly define a host or scheme at construction time.
// As this isn't created by users, we don't need to normalize it.
```

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
```diff
@@ -171,4 +175,19 @@ public void close() {
     void doClose() {
         clientFactoriesToClose.forEach(ClientFactory::close);
     }
+
+    private URI normalizeUri(URI uri, Scheme scheme) {
+        if (isUndefinedUri(uri)) {
+            // See https://github.com/line/armeria/pull/2241 discussion for more details.
+            return uri;
+        }
+        if (scheme.serializationFormat() != SerializationFormat.NONE) {
```
anuraaga: Would add a line break since there is a logical break in what we're doing.

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
```diff
@@ -171,4 +175,19 @@ public void close() {
...
+            return uri;
+        }
+
+        final String newScheme = uri.getScheme()
```
anuraaga: I believe it's more clear to use `scheme.sessionProtocol()` instead of string replacement - we are always replacing something like `none+http` or `https+none` with the protocol so may as well write the code like that.

Might also consider using the non-string constructor - I think it's more clear since there's no thinking about "is `substring` length actually correct, will parse throw errors", etc, it's clearly doing exactly what we want to do, replacing the scheme.

```java
new URI(scheme.sessionProtocol().uriText(),
             u.getUserInfo(), u.getHost(), u.getPort(),
             u.getPath(), u.getQuery(),
             u.getFragment())
```

core/src/main/java/com/linecorp/armeria/common/Scheme.java
```diff
@@ -78,8 +78,12 @@
         if (scheme == null) {
             return Optional.empty();
         }
-
-        return Optional.ofNullable(SCHEMES.get(Ascii.toLowerCase(scheme)));
+        final String lowercaseScheme = Ascii.toLowerCase(scheme);
+        final Optional<Scheme> parsedScheme = Optional.ofNullable(SCHEMES.get(lowercaseScheme));
```
anuraaga: We could make sure to only create `Optional` once

```java
Scheme parsedScheme = SCHEMES.get(lowercaseScheme);
if (parsedScheme != null) {
  return Optional.of(parsedScheme);
}

return Optional.ofNullable(SCHEMES.get(SerializationFormat.NONE.uriText() + '+' + lowercaseScheme));
```

core/src/test/java/com/linecorp/armeria/client/ClientBuilderTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Test
+    void nonePlusSchemeProvided() {
+        final HttpClient client = new ClientBuilder("none+https://google.com/").build(HttpClient.class);
+        assertThat(client.uri().toString()).isEqualTo("https://google.com/");
+    }
+
+    @Test
+    void nonePlusSchemeUriToUrl() {
+        final HttpClient client = new ClientBuilder("none+https://google.com/").build(HttpClient.class);
+        assertThatCode(() -> client.uri().toURL()).doesNotThrowAnyException();
```
anuraaga: How about using `.isEqualTo(URL)` to compare the actual value? Since we're asserting anyways doesn't hurt to have a stricter assertion of the behavior.

anuraaga: Alternatively, I guess comparing to a known `URI` in the other test would implicitly check this and we wouldn't need it.

core/src/test/java/com/linecorp/armeria/common/SchemeTest.java
```diff
@@ -0,0 +1,60 @@
...
+class SchemeTest {
+
+    @Test
+    public void tryParse_null() {
+        assertThat(Scheme.tryParse(null)).isEmpty();
+    }
+
+    @Test
+    void tryParse_add_none() {
+        final Optional<Scheme> got = Scheme.tryParse("http");
```
anuraaga: May as well call `.get()` here once instead of twice below, or alternatively can use `assertThat(Optional).hasValueSatisfying(value -> assertThat(value.serializationFormat())....`

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
trustin: I'd prefer using a single-string constructor because multi-string constructors concatenates the components and parses it anyway.

core/src/test/java/com/linecorp/armeria/client/ClientBuilderTest.java
rmohta: I have made the change.

This UT was included, as prior to this PR uri().toURL() will fail with MalformedURLException. It's to avoid any regression check, once this PR is merged.

anuraaga: Yeah - I think it's too lenient of a test though since it's easy to pass with a broken implementation that just returns a const `URI.create("github.com")` or something.

We're more concerned that the actual `URI` returned by Armeria is correct, while exception semantics of `toURL()` are implementation details of JDK, not Armeria. The other tests seem to take care of this, though I think it's even clearer without using `toString()` in them.

core/src/test/java/com/linecorp/armeria/common/SchemeTest.java
rmohta: Done

core/src/main/java/com/linecorp/armeria/common/Scheme.java
rmohta: Done

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
rmohta: Done

core/src/test/java/com/linecorp/armeria/client/ClientBuilderTest.java
anuraaga: Ah sorry I may have been confused by your comment - with the change it's probably fine :)

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
rmohta: Done

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
rmohta: @trustin @anuraaga : Sorry, I am a bit confused. any change is needed for this? Without going into URI construction complexities, this one tackles the scheme only. 
WDYT?  

```
final String newScheme = uri.getScheme()
                                    .replace(SerializationFormat.NONE.uriText(), "")
                                    .replace("+", "");
        return URI.create(newScheme + uri.toString().substring(uri.getScheme().length()));
```

core/src/test/java/com/linecorp/armeria/common/SchemeTest.java
```diff
@@ -36,15 +37,25 @@ public void tryParse_null() {
     @Test
     void tryParse_add_none() {
         final Optional<Scheme> got = Scheme.tryParse("http");
-        assertThat(got.get().serializationFormat()).isEqualTo(SerializationFormat.NONE);
-        assertThat(got.get().sessionProtocol()).isEqualTo(SessionProtocol.HTTP);
+        final Condition<Scheme> condition = new Condition<>(scheme ->
```
anuraaga: I think it's more idiomatic to call `assertThat` in the lambda instead of returning a boolean when using `satisfying` similar to here - https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractOptionalAssert.html#hasValueSatisfying-java.util.function.Consumer-

Otherwise, all assertions get bundled into a single error message - `Condition` I think is more for framework authors than test writers.

```java
assertThat(Scheme.tryParse("http")).hasValueSatisfying(scheme -> {
  assertThat(scheme.sessionProtocol()).isEqualTo(HTTP);
  assertThat(scheme.serializationFormat()).isEqualTo(NONE);
});
```

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
anuraaga: I personally still like the multi-string constructor since it's indeed implemented similar to `URI.create(String)` but makes very clear which part we're replacing.

Otherwise, if going with the single string, I'd still recommend being explicit `scheme.sessionProtocol() + "://" + uri.substring(`. There doesn't seem to be a good reason to rely on the coincidence that session protocols never have the word none inside them, i.e., the current code wouldn't work for a scheme like `none+httpnone://google.com`. It's a contrived example, but I think shows that the current logic is more indirect than it needs to be.

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
```diff
@@ -171,4 +175,22 @@ public void close() {
...
+            return uri;
+        }
+
+        if (scheme.serializationFormat() != SerializationFormat.NONE) {
+            return uri;
+        }
+
+        final String newScheme = uri.getScheme()
+                                    .replace(SerializationFormat.NONE.uriText(), "")
+                                    .replace("+", "");
```
minwoox: Can we just replace it once because `nonehttp://...` was not valid uri?

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
rmohta: Agree. I have updated it to use
```
        return URI.create(scheme.sessionProtocol() + "://" +
                          uri.toString().substring(uri.getScheme().length()));
```

core/src/test/java/com/linecorp/armeria/common/SchemeTest.java
rmohta: I was using _hasValueSatisfying(Condition<? super VALUE> condition)_.
But I see what the Javadoc says. Will update to use the Consumer function.

> Should be used as a way of deeper asserting on the containing object, as further requirement(s) for the value.

rmohta: Done

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
rmohta: I couldn't replace it once, as scheme could be http+none or none+http. Hence I had to do twice.

I have updated the code, where there is no longer replace. We are using parsed Scheme.sessionProtocol() as prefix for new URI construction.

core/src/main/java/com/linecorp/armeria/client/AbstractClientFactory.java
```diff
@@ -96,18 +96,17 @@
...
-        if (scheme == null) {
-            throw new IllegalArgumentException("URI with missing scheme: " + uri);
-        }
-
         if (uri.getAuthority() == null) {
             throw new IllegalArgumentException("URI with missing authority: " + uri);
         }
 
+        final String scheme = uri.getScheme();
+        if (scheme == null) {
```
minwoox: Is there any reason that you changed the order of checking `null`?

rmohta: To group processing. We only check and work on value of getAuthority() once. Rest of the code is related to uri scheme. So I updated the order, to have all scheme related processing grouped together. More for readability.

minwoox: Yeah, that makes sense. ;)

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
```diff
@@ -171,4 +175,20 @@ public void close() {
...
+            // host or scheme at construction time.
+            // As this isn't created by users, we don't need to normalize it.
+            return uri;
+        }
+
+        if (scheme.serializationFormat() != SerializationFormat.NONE) {
+            return uri;
+        }
+
+        return URI.create(scheme.sessionProtocol() +
```
trustin: `scheme.sessionProtocol().uriText()`?

rmohta: toString() also returns the uriText().

rmohta: Made changes, to explicitly call uriText().

trustin: Yeah, but it may not be safe to rely on the behavior of `toString()` in the long term. Thanks for making a change!

trustin:
:tada: Thanks a lot, @rmohta!

PR #2248, opened 2019-11-11
Allow handling response timeout with specific cause on the clie…

ikhoon:
Motivation:
`ServiceRequestContext` can configure response timeout handler by adding `setResponseTimeoutHandler`.
But `ClientRequestContext` is less customizable than server-side.
We need to add a response timeout handler in the client-side to solve the problem in #2008.

Modifications:
* Add `setResponseTimeoutHandler(Runnable responseTimeoutHandler)` to `ClientRequestContext`.
* Add `abort(Throwable)` to StreamMessage and its implementations.
* Add new factory method `AbortingSubscriber.get(Throwable)`.
* Add `Runnable requestTimeoutHandler()` to `ServiceRequestContext`.
* Update existing test logic.
* Dependency
  * Add TestNG(7.0.0) to core test dependency for running Reactive Streams TCK on modern IntelliJ
* Deprecate `NoopRequestLogBuilder`.

Result:
A user can configure a custom response timeout handler to intercept default logic and abort request with a specific cause.


ikhoon:
This PR is reopened, The origin PR was #2201 PTAL 🙇‍♂️

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
```diff
@@ -133,32 +132,41 @@ final void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEven
...
-                         new CloseEvent(new CancelledSubscriptionException()) : CANCELLED_CLOSE;
+        if (cause == CancelledSubscriptionException.INSTANCE) {
+            closeEvent = CANCELLED_CLOSE;
+        } else if (cause == AbortedStreamException.INSTANCE) {
+            closeEvent = ABORTED_CLOSE;
         } else {
-            closeEvent = sampler.isSampled(AbortedStreamException.class) ?
-                         new CloseEvent(new AbortedStreamException()) : ABORTED_CLOSE;
+            closeEvent = new CloseEvent(cause);
         }
```
trustin: I see similar logic in a few places. Can we move this common logic to a static method in `AbstreamStreamMessage`?

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -100,7 +104,7 @@
     private static final AtomicReference<ByteBuf> releasedByteBuf = new AtomicReference<>();
 
     // Used to communicate with test when the response can't be used.
-    private static final AtomicReference<Boolean> completed = new AtomicReference<>();
+    private static final AtomicBoolean completed = new AtomicBoolean();
```
trustin: Could be just a `volatile boolean`?

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -341,6 +345,12 @@ public void onComplete() {
                 }
             });
 
+            sb.service("/client-aborted", (ctx, req) -> {
+                // Don't need to return a real response since the client will timeout.
+                completed.compareAndSet(false, true);
```
trustin: No point of using CAS - could just assign.

core/src/test/java/com/linecorp/armeria/client/HttpClientResponseTimeoutHandlerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    void testResponseTimeoutHandler(SessionProtocol protocol, boolean useResponseTimeoutHandler) {
+        final AtomicReference<RequestLog> logHolder = new AtomicReference<>();
+        final IllegalStateException reqCause = new IllegalStateException("abort request");
+        final AtomicBoolean invokeResponseTimeoutHandler = new AtomicBoolean(false);
+        final HttpClient client = HttpClient.builder(server.uri(protocol, "/"))
+                .responseTimeout(Duration.ofSeconds(2))
+                .decorator((delegate, ctx, req) -> {
+                    if (useResponseTimeoutHandler) {
+                        ctx.setResponseTimeoutHandler(() -> {
+                            ctx.request().abort(reqCause);
```
trustin: Could we provide two tests? One that aborts a request and the other that aborts a response.

core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
```diff
@@ -457,26 +459,40 @@ public void shouldGetExceptionWhenFactoryIsClosed() {
...
+            assertThat(responseAbortServiceCallCounter.get()).isOne();
+            assertThat(log.requestCause()).isNull();
+            if (abortCause == null) {
+                assertThat(log.responseCause()).isExactlyInstanceOf(AbortedStreamException.class);
+            } else {
+                assertThat(log.responseCause()).isExactlyInstanceOf(abortCause.getClass())
+                                               .hasMessageContaining("abort stream with a specified cause");
+            }
+
+            // Sleep 3 seconds more to check if there was another retry.
```
trustin: 3 more seconds

core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
```diff
@@ -504,26 +520,39 @@ public void onComplete() {}
...
+            TimeUnit.SECONDS.sleep(1);
+            // No request is made.
+            assertThat(responseAbortServiceCallCounter.get()).isZero();
+            final RequestLog log = context.get().log();
+            if (abortCause == null) {
+                assertThat(log.requestCause()).isExactlyInstanceOf(AbortedStreamException.class);
+                assertThat(log.responseCause()).isExactlyInstanceOf(AbortedStreamException.class);
+            } else {
+                assertThat(log.requestCause()).isExactlyInstanceOf(abortCause.getClass());
+                assertThat(log.responseCause()).isExactlyInstanceOf(abortCause.getClass());
```
trustin: `isSameAs()`?

core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientTest.java
```diff
@@ -457,26 +459,40 @@ public void shouldGetExceptionWhenFactoryIsClosed() {
...
+            }
+
+            final RequestLog log = context.get().log();
+            await().untilAsserted(() -> assertThat(log.isAvailable(RequestLogAvailability.COMPLETE)).isTrue());
+            assertThat(responseAbortServiceCallCounter.get()).isOne();
+            assertThat(log.requestCause()).isNull();
+            if (abortCause == null) {
+                assertThat(log.responseCause()).isExactlyInstanceOf(AbortedStreamException.class);
+            } else {
+                assertThat(log.responseCause()).isExactlyInstanceOf(abortCause.getClass())
```
trustin: `isSameAs()`?

core/src/test/java/com/linecorp/armeria/common/stream/PublisherBasedStreamMessageTest.java
```diff
@@ -156,28 +173,54 @@ void invokeOnSubscribe() {
...
+
+        @Override
+        public Stream<? extends Arguments> provideArguments(final ExtensionContext extensionContext)
+                throws Exception {
+            return Arrays.asList(null,
+                                 new IllegalStateException("abort stream with a specified cause"))
+                         .stream()
+                         .map(Arguments::of);
         }
     }
```
trustin: Could de-duplicate by moving top-level?

core/src/test/java/com/linecorp/armeria/common/stream/StreamMessageDuplicatorTest.java
```diff
@@ -66,6 +68,9 @@
     @Rule
     public TestRule globalTimeout = new DisableOnDebug(new Timeout(10, TimeUnit.SECONDS));
 
+    private static final List<Throwable> ABORT_CAUSES =
+            Arrays.asList(null, new IllegalStateException("abort stream with a specified cause"));
```
trustin: Could be provided by the top-level class mentioned above.

core/src/main/java/com/linecorp/armeria/common/stream/FixedStreamMessage.java
ikhoon: Sure, I will address it!

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
```diff
@@ -74,8 +74,9 @@ private HttpResponse execute(@Nullable EventLoop eventLoop, HttpRequest req) {
         }
 
         if (isUndefinedUri(uri())) {
-            req.abort();
-            return HttpResponse.ofFailure(new IllegalArgumentException("no authority: " + req.path()));
+            final IllegalArgumentException cause = new IllegalArgumentException("no authority: " + req.path());
+            req.abort(cause);
```
anuraaga: I think this might best help me understand the different use cases of `abort` and `close`, which I'm still not comfortable about (I'm still not sure we need two methods).

This PR was about allowing customizing of response timeout, but it looks like we're changing existing behavior not related to timeout. So I wonder, what is the result of this change for the user? Did something change / get better by passing `cause` to `abort` everywhere? Usability-wise, I think it's worse to have this sort of duplicated logic, so I wonder what the effect is.

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -724,6 +734,39 @@ void testUpgradeRequestExecutesLogicOnlyOnce() throws Exception {
...
+        final HttpRequestWriter request = HttpRequest.streaming(HttpMethod.GET, "/client-aborted");
+
+        final HttpResponse response = client.execute(request);
+        final IllegalStateException badState = new IllegalStateException("bad state");
+        if (isAbort) {
+            request.abort(badState);
+        } else {
+            request.close(badState);
+        }
+        // request cause is obtained immediately
```
anuraaga: What does this comment mean? I guess `join` could take time if it wanted to so we're not verifying anything immediate.

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -724,6 +734,39 @@ void testUpgradeRequestExecutesLogicOnlyOnce() throws Exception {
...
+                .isInstanceOf(CompletionException.class)
+                .hasCause(badState);
+    }
+
+    @Test
+    void responseAbortWithException() throws InterruptedException {
+        final HttpClient client = HttpClient.of(server.httpUri("/"));
+        final HttpRequest request = HttpRequest.streaming(HttpMethod.GET, "/client-aborted");
+        final HttpResponse response = client.execute(request);
+        await().untilAsserted(() -> assertThat(completed).isTrue());
```
anuraaga: I guess this can be `await().until(() => completed)`. Though actually it might be a reason to stick with `AtomicReference` since there's also `await().untilTrue(AtomicReference)`

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
ikhoon: Oops, I should've deleted this comment. There was a code that holds an exception when a stream has been aborted. Thanks for reviewing!

ikhoon:
> Sorry for continuing to be confused - feel free to tell me to go away if it's getting boring ;)

I really appreciate your detail and thoughtful idea. Your review is always welcome! 😊

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
ikhoon: > This PR was about allowing customizing of response timeout, but it looks like we're changing existing behavior not related to timeout.

I agree with your opinion. This PR is not just about adding a response timeout handler. To customize the response timeout handler, we need a new abort method that consumes an abort cause. While adding the method, we wanted to migrate and fix the existing abort logic. Sorry for the confusion. I think the adding `abort(cause)` issue should be handled first. 

> Usability-wise, I think it's worse to have this sort of duplicated logic, so I wonder what the effect is.

This seems to have duplicated logic, but I believe we could deliver more detail abort cause to `log.responseCause()`, `log.requestCause()`, and `completionFuture()`.

core/src/test/java/com/linecorp/armeria/client/HttpClientResponseTimeoutHandlerTest.java
trustin: Ah, missed that one. Never mind.

core/src/test/java/com/linecorp/armeria/common/stream/AbortCauseArgumentProvider.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+
+class AbortCauseArgumentProvider implements ArgumentsProvider {
+
+    static final List<Throwable> ABORT_CAUSES =
+            Arrays.asList(null, new IllegalStateException("abort stream with a specified cause"));
```
trustin: Could use `ImmutableList.of()` because `Arrays.asList()` returns a mutable list.

core/src/main/java/com/linecorp/armeria/client/DefaultHttpClient.java
trustin: It actually fixes a bug where the specified request is never consumed when a bad request is given. This may yield a memory leak, so he fixed it by aborting the request. Besides this, I don't see any behavioral changes, right?

anuraaga: We were already calling `req.abort()` without a cause so I wouldn't expect any difference in terms of memory leak. From what I understand, now `log.requestCause` will have the actual cause, which is the same as `log.responseCause` when before it was just `AbortedStreamException` for the request. While it's sort of double-logging, I guess it's still more useful than generic `AbortedStreamException` so it makes sense. I sort of wish returning a failed response automatically closed the request but that's not so related to this PR.

trustin: Ahh. Sorry, I was seeing a wrong diff. :sweat_smile:

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -619,7 +629,7 @@ public void onComplete() {
         req.write(HttpData.ofUtf8("not finishing this stream, sorry."));
         await().untilAsserted(() -> assertThat(obj).hasValue(ResponseHeaders.of(HttpStatus.OK)));
         factory.close();
-        await().untilAsserted(() -> assertThat(completed).hasValue(true));
+        await().untilAsserted(() -> assertThat(completed));
```
anuraaga: Can update here too with `until`

core/src/test/java/com/linecorp/armeria/common/stream/AbortCauseArgumentProvider.java
ikhoon: At first, I wanted to use it. However `ImmutableList.of()` check not-null for every element. Let me wrap this with `Collections.unmodifiableList()`.

core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
```diff
@@ -63,6 +63,7 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
             // - `ClientRequestContext.endpoint()` returns `null` only when the context initialization failed.
             // - `ClientUtil.initContextAndExecuteWithFallback()` will use the fallback response rather than
             //   what we return here.
+            req.abort(CONTEXT_INITIALIZATION_FAILED);
```
minwoox: 👍

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -192,12 +194,24 @@ void logResponseFirstBytesTransferred() {
 
         @Override
         public void run() {
-            final ResponseTimeoutException cause = ResponseTimeoutException.get();
-            delegate.close(cause);
-            logBuilder.endResponse(cause);
+            final Runnable responseTimeoutHandler;
+            if (logBuilder instanceof DefaultRequestLog) {
+                responseTimeoutHandler =
+                        ((ClientRequestContext) ((RequestLog) logBuilder).context()).responseTimeoutHandler();
```
minwoox: How about just passing `ctx` instead of `logBuilder` when creating `HttpResponseWrapper` so we don't need to do this type casting?

core/src/main/java/com/linecorp/armeria/common/stream/DeferredStreamMessage.java
```diff
@@ -100,8 +98,8 @@ protected void delegate(StreamMessage<T> delegate) {
             throw new IllegalStateException("delegate set already");
         }
 
-        if (abortPending != 0) {
-            delegate.abort();
+        if (abortCause != null) {
```
minwoox: We should do:
```java
final Throwable abortCause = this.abortCause;
if (abortCause != null) {
    delegate.abort(abortCause);
}
```

core/src/main/java/com/linecorp/armeria/internal/Http2GoAwayHandler.java
```diff
@@ -44,7 +44,7 @@
     private static final long CODE_INTERNAL_ERROR = Http2Error.INTERNAL_ERROR.code();
 
     private boolean goAwaySent;
-    private long goAwayReceived; // -1 if not received, errorCode if received.
+    private long goAwayReceived = -1; // -1 if not received, errorCode if received.
```
minwoox: 👍

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -283,6 +283,14 @@ default MediaType negotiatedProduceType() {
      */
     void setRequestTimeout(Duration requestTimeout);
 
+    /**
+     * Returns {@link Request} timeout handler which is executed when
+     * receiving the current {@link Request} and sending the corresponding {@link Response}
+     * is not completely within the allowed {@link #requestTimeoutMillis()}.
```
minwoox: completely -> completely received

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
ikhoon: Originally, I thought of passing `Supplier<Runnable> responseTimeoutHandlerSupplier` or `ctx` too. But I don't have a strong convince whether it is good for not. 

The disadvantage of this is that we should cast `logBuilder`. However `RequestContext` and `ClientRequestContext` is mutable, so I wanted to stay `ctx` in `HttpSessionHandler` and hide it in `HttpResponseDecoder`. Let me know if you think it would be better to add `ctx` as a parameter.

minwoox: > However RequestContext and ClientRequestContext is mutable

Is there any difference if we pass the `ctx` instread of using `logBuilder` to retrieve the `ctx` in the point of view of mutability?

ikhoon: There is no difference in mutability. IMHO, the main difference is the scope of `ctx`. We can directly access `ctx` from `logBuilder`. In this case, `ctx` is only visible in this `if` block. :-) As you know, if we pass `ctx` with the constructor, it could be accessed everywhere in this class. 

Either is fine. But this idea was suggested by @trustin. I wanted to hear his opinion 👂

trustin: +1 for just passing `ctx` if possible.

ikhoon: Fixed it https://github.com/line/armeria/pull/2248/commits/f944b32506c1d743a12cee3a01f52c7abe98a793

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -141,6 +141,8 @@ final boolean needsToDisconnectWhenFinished() {
         @Nullable
         private final HttpRequest request;
         private final DecodedHttpResponse delegate;
+        @Nullable
+        private final ClientRequestContext ctx;
         private final RequestLogBuilder logBuilder;
```
trustin: Could you then remove this field and guard with `if (ctx != null)`?

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogBuilder.java
```diff
@@ -40,11 +40,6 @@
  */
 public interface RequestLogBuilder {
 
-    /**
-     * A dummy {@link RequestLogBuilder} that discards everything it collected.
-     */
-    RequestLogBuilder NOOP = new NoopRequestLogBuilder();
```
minwoox: Wonder if we need to mention this in the release note. Probably there's no one who has used this. 🤣

trustin: Maybe `@Deprecate` first? :sob:

ikhoon: Oops, Sorry. I will revert it and `@Deprecated` it.

ikhoon: Fix it https://github.com/line/armeria/pull/2248/commits/79fd34ef6172333f41855ad49a7783d75723a2d1
Wrote documentation for the deprecation. But I am wondering if it is a good explanation. 😅

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogBuilder.java
```diff
@@ -42,7 +42,13 @@
 
     /**
      * A dummy {@link RequestLogBuilder} that discards everything it collected.
+     *
+     * @deprecated This field will be removed without a replacement. The {@link NoopRequestLogBuilder}
+     *             means that it does not perform any operations on {@link RequestLogBuilder}.
+     *             The {@link NoopRequestLogBuilder} behavior can be achieved by performing operations
+     *             on {@link RequestContext#logBuilder()} if {@link RequestContext} is not null.
```
trustin: `NoopRequestLogBuilder` is not part of public API, so it's not a good idea to add a link to it in public API Javadoc. I also think the proposed explanation here is only valid in the context of this pull request, i.e. Users will have no idea why you're saying this. It has to be explained in our users' standpoint. How about:

```
This field will be removed without a replacement. Use {@link DefaultRequestLog} or a mock.
```

ikhoon: Thanks for reviewing! I will address it. 🙏

trustin:
:tada: :tada: :tada: Thanks for your patience, @ikhoon

minwoox:
🎉 🎉 🎉

PR #2239, opened 2019-11-07
Use HttpService and RpcService instead of Service<I, O>

minwoox:
Rule of thumbs:
- Every service should be either `HttpService` or `RpcService`.
- A service that is for HTTP does not have `HTTP` in its name.
  - e.g. `LoggingService`, `MetricCollectingService`, `SimpleCompositeService` and so on
  - Some core APIs should have a type of hierarchy. In that case, it does contain HTTP in its name.
    - `Service`, `HttpService`, `RpcService`
    - `SimpleDecoratingService`, `SimpleDecoratingHttpService`, `SimpleDecoratingRpcService`
    - `ServiceWithRoutes`, `HttpServiceWithRoutes`, `RpcServiceWithRoutes`
    - `TransientService`, `TransientHttpService`, `TransientRpcService`

Modifications:
- Every service is either `HttpService` or `RpcService`.
- Use `SimpleDecoratingHttpService` or `SimpleDecoratingRpcService` if possible.
- Follwe PECS for decorators: `Function<? super HttpService, ? extends HttpService>`
  - No more explicit type conversion.
- `Service.decorate(Function<T, R> decorator)` is pulled down
  - `HttpService.decorate(Function<? super HttpService, R> decorator)`
  - `RpcService.decorate(Function<? super RpcService, R> decorator)`
- `Service.decorate(DecoratingServiceFunction<I, O> function)` is pulled down  
  - `HttpService.decorate(DecoratingHttpServiceFunction function)`
  - `RpcService.decorate(DecoratingRpcServiceFunction function)`
  - `DecoratingServiceFunction` is split into `DecoratingHttpServiceFunction` and `DecoratingRpcServiceFunction`
- Provode only HTTP decorators which does not use `Request` and `Response` in its `serve()` method
  - `LoggingService`, `MetricCollectingService`, `StructuredLoggingService` and `KafkaStructuredLoggingService`
- Split into HTTP and RPC
  - `(Http|Rpc)ServiceWithRoutes`
  - `Transient(Http|Rpc)Service`
- AbstractCompositeService
  - Type parameters are changed: `<Request, Response>` -> `<Service, Request, Response>`
  - `SimpleCompositeService<Request, Response>` -> `SimpleCompositeService` and `SimpleCompositeRpcService`
  - `CompositeServiceEntry<HttpRequest, HttpResponse>` -> `CompositeServiceEntry`
- `THttpService` now extends `DecoratingService`
  - `RpcService.serviceAdded()` and `RpcService.as())` are now invoked.

PR #2251, opened 2019-11-12
Fix incorrect use of `extends` keyword in `Function`

trustin:
`super` should be used for the type of the object being consumed, but I
mistakenly used `extends`.


trustin:
Let me close this because it's been fixed already by @minwoox in #2239

PR #2253, opened 2019-11-12
Allow `HttpStatusException` to have a cause

trustin:
Motivation:

Users may want to record why they are throwing `HttpStatusException` by
specifying its cause.

Modifications:

- Add the static factory methods that require `Throwable cause` to
  `HttpStatusException`
- Fixed a bug where `HttpStatusException.of(int)` always returns a
  non-verbose exception.

Result:

- Users can add a cause to an `HttpStatusException`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
@@ -78,21 +78,26 @@ public final class HttpStatusException extends RuntimeException {
     public static HttpStatusException of(HttpStatus status, Throwable cause) {
         requireNonNull(status, "status");
         requireNonNull(cause, "cause");
-        return new HttpStatusException(status, cause);
+        return of0(status, cause);
+    }
+
+    private static HttpStatusException of0(HttpStatus status, @Nullable Throwable cause) {
+        final boolean sampled = Flags.verboseExceptionSampler().isSampled(HttpStatusException.class);
+        if (sampled || cause != null) {
+            return new HttpStatusException(status, sampled, cause);
+        }
+
+        final int statusCode = status.code();
+        if (statusCode >= 0 && statusCode < 1000) {
+            return EXCEPTIONS[statusCode];
+        } else {
+            return new HttpStatusException(HttpStatus.valueOf(statusCode), false, null);
+        }
     }
 
     private static final long serialVersionUID = 3341744805097308847L;
 
     private final HttpStatus httpStatus;
-
-    /**
-     * Creates a new instance.
-     */
-    private HttpStatusException(HttpStatus httpStatus, @Nullable Throwable cause) {
-        this(httpStatus,
-             Flags.verboseExceptionSampler().isSampled(HttpStatusException.class),
-             cause);
-    }
 
     /**
      * Creates a new instance.
--- a/core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
@@ -17,6 +17,8 @@ package com.linecorp.armeria.server;
 
 import static java.util.Objects.requireNonNull;
 
+import javax.annotation.Nullable;
+
 import com.linecorp.armeria.common.Flags;
 import com.linecorp.armeria.common.HttpStatus;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -32,51 +34,69 @@
...
     private final HttpStatus httpStatus;
 
     /**
-     * Creates a new instance with the specified {@link HttpStatus}.
+     * Creates a new instance.
      */
-    private HttpStatusException(HttpStatus httpStatus) {
-        super(requireNonNull(httpStatus, "httpStatus").toString());
-        this.httpStatus = httpStatus;
+    private HttpStatusException(HttpStatus httpStatus, @Nullable Throwable cause) {
```
minwoox: nit: we can remove `@Nullable`.

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -32,51 +34,69 @@
...
+        if (!sampled) {
+            final int statusCode = status.code();
+            if (statusCode >= 0 && statusCode < 1000) {
+                return EXCEPTIONS[statusCode];
             } else {
-                return new HttpStatusException(status, false);
+                return new HttpStatusException(HttpStatus.valueOf(statusCode), false, null);
             }
-        } else {
-            return EXCEPTIONS[statusCode];
```
anuraaga: Did we fix a bug that normal status codes would never be `verbose`?

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -32,51 +34,69 @@
...
     /**
-     * Creates a new instance with the specified {@link HttpStatus}.
+     * Creates a new instance.
      */
-    private HttpStatusException(HttpStatus httpStatus) {
-        super(requireNonNull(httpStatus, "httpStatus").toString());
-        this.httpStatus = httpStatus;
+    private HttpStatusException(HttpStatus httpStatus, @Nullable Throwable cause) {
+        this(httpStatus,
+             Flags.verboseExceptionSampler().isSampled(HttpStatusException.class),
```
anuraaga: Having some sampler happen in factories, other in constructor is a bit mysterious. How about moving it to the factory and removing this constructor?

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
trustin: Yes.

trustin: Updated PR description

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -32,51 +34,69 @@
...
-            final HttpStatus status = HttpStatus.valueOf(statusCode);
-            if (Flags.verboseExceptionSampler().isSampled(HttpStatusException.class)) {
-                return new HttpStatusException(status);
+        return of(HttpStatus.valueOf(statusCode));
+    }
+
+    /**
+     * Returns a new {@link HttpStatusException} instance with the specified HTTP status code and {@code cause}.
+     */
+    public static HttpStatusException of(int statusCode, Throwable cause) {
```
anuraaga:
```suggestion
    public static HttpStatusException of(int statusCode, Throwable cause) {
        requireNonNull(cause, "cause");
```

I think normally we check null at all public APIs even if a delegate would check it.

core/src/main/java/com/linecorp/armeria/server/HttpStatusException.java
```diff
@@ -32,51 +34,69 @@
...
+        return of(HttpStatus.valueOf(statusCode), cause);
+    }
+
+    /**
+     * Returns a new {@link HttpStatusException} instance with the specified {@link HttpStatus}.
+     */
+    public static HttpStatusException of(HttpStatus status) {
```
anuraaga:
```suggestion
    public static HttpStatusException of(HttpStatus status) {
        requireNonNull(status, "status");
```

trustin:
Thanks for reviewing!

@anuraaga Thanks for providing a change suggestion. That saved many keystrokes.

PR #2255, opened 2019-11-13
Make route decorators handle the requests for non-existent mapp…

### LINKED ISSUE #1625: Have armeria always add a fallback Service for not found requests?
I found in https://github.com/openzipkin/zipkin/pull/2348 that there is somewhat confusing behavior regarding `ServerBuilder.decorator` - as it only decorates mapped services, requests that aren't mapped and returned as `NOT_FOUND` are not handled by the decorator. If the decorator is for metrics collection, and has logic for not found requests, it doesn't work.

Does it make sense to have armeria always add a service at the end for not_found requests to make sure behavior is consistent with mapped requests?

trustin:
Related: #1625
Motivation:

Currently, the requests that are not matched by any `Route`s are handled
differently than other matched requests. Most importantly, such requests
re not evaluated by route decorators. For example, a request to `/bar`,
which is a non-existent path, is currently never handled by
`CorsService` in the following code:

    Server.builder()
          .service("/foo", ...)
          .routeDecorator().pathPrefix(/")
                           .build(CorsService.newDecorator(...))

Modifications:

- Add `FallbackService`
  - Move the CORS handling code in `CompositeRouter` to
    `VirtualHost.findServiceConfig()`.
- Make `VirtualHost.findServiceConfig()` fall back to the
  `FallbackService` when necessary.
- Rename `RoutingContext.delayedThrowable` to `deferredStatusException`
  - Change the type of `deferredStatusException` from `Optional<Throwable>`
    to `HttpStatusException`
- Add `Route.ofCatchAll()` and use it wherever possible.
- Add a test case to `HttpCorsServiceTest` to show the fallback service
  works as expected.

Result:

- Fixes #1625
- More requests are handled via the ordinary request processing pipeline.
  - 404, 405, 415


thrift/src/test/java/com/linecorp/armeria/server/thrift/AbstractThriftOverHttpTest.java
```diff
@@ -447,4 +448,16 @@ protected static String newUri(String scheme, String path) {
...
+        for (;;) {
+            final RequestLog log = requestLogs.take();
+            if (log.method() == HttpMethod.HEAD) {
+                // Skip the upgrade request.
+                continue;
+            }
+
+            return log;
+        }
+    }
```
trustin: Now even an HTTP/2 upgrade request has a request log :tada:

trustin:
@anuraaga The implementation is different than what's been discussed in #1625 because I needed to handle non-404 responses produced by a `Router`.

core/src/main/java/com/linecorp/armeria/server/FallbackService.java
```diff
@@ -0,0 +1,69 @@
...
+        final String pathWithSlash = path + '/';
+        if (!ctx.virtualHost().findServiceConfig(routingCtx.overridePath(pathWithSlash)).isPresent()) {
+            // '/path/' does not exist. Send 404.
+            throw cause;
+        }
+
+        // '/path/' exists. Send a redirect response.
+        final String location;
+        final String originalPath = req.path();
+        if (path.length() == originalPath.length()) {
```
anuraaga: Does `req.path()` get mutated? I think `path == originalPath == req.path()`

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
```diff
@@ -348,15 +354,46 @@ public boolean shutdownAccessLogWriterOnStop() {
...
+     * @return the {@link ServiceConfig} wrapped by a {@link Routed} if there's a match.
+     *         {@link Routed#empty()} if there's no match.
+     */
+    public Routed<ServiceConfig> findServiceConfig(RoutingContext routingCtx, boolean useFallbackService) {
+        final Routed<ServiceConfig> routed = router.find(requireNonNull(routingCtx, "routingCtx"));
+        if (routed.isPresent() || !useFallbackService) {
+            return routed;
+        }
+
+        if (routingCtx.isCorsPreflight()) {
```
anuraaga: Is there a reason to have this logic here and not in the fallback service?

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
```diff
@@ -348,15 +354,46 @@ public boolean shutdownAccessLogWriterOnStop() {
...
+            // '403 Forbidden' is better for a CORS preflight request than '404 Not Found'.
+            routingCtx.deferStatusException(HttpStatusException.of(HttpStatus.FORBIDDEN));
+        } else if (routingCtx.deferredStatusException() == null) {
+            routingCtx.deferStatusException(HttpStatusException.of(HttpStatus.NOT_FOUND));
+        }
+
+        return Routed.of(fallbackServiceConfig.route(),
```
anuraaga: Is there a reason not to add the fallback service to the router?

trustin: I wasn't sure if it's a good idea to cache the resulting `Route`. Thoughts?

anuraaga: Ah interesting point - I remember our path parsing cache only stores into the cache when the service succeeded, perhaps the routing cache should do something similar. But either way, for this PR that makes sense to prevent exploding cache - let's add a comment :)

core/src/main/java/com/linecorp/armeria/server/FallbackService.java
trustin: Brainfart 1

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
trustin: Brainfart 2

core/src/main/java/com/linecorp/armeria/server/FallbackService.java
trustin: The intention of this comparison was to check if the request path had a query string or not. I had to get `path` from `routingCtx.path()` which returns a path with a query string removed.

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
trustin: Moved

core/src/main/java/com/linecorp/armeria/server/VirtualHost.java
trustin: Added a comment :-)

core/src/main/java/com/linecorp/armeria/server/FallbackService.java
```diff
@@ -0,0 +1,89 @@
...
+        if (!ctx.virtualHost()
+                .findServiceConfig(routingCtx.overridePath(pathWithSlashAndWithoutQuery))
+                .isPresent()) {
+            // No need to send a redirect response because '/path/' (or '/path/?query') does not exist.
+            throw cause;
+        }
+
+        // '/path/' (or '/path/?query') exists. Send a redirect response.
+        final String location;
+        final String pathWithQuery = req.path();
```
anuraaga: Possibly more readable using `req.query()` since that's what we're concerned with

```java
if (req.query() == null) {
  location = req.path();
} else {
  location = req.path() + '?' + req.query();
}
```

I guess it's a tiny bit faster than current code (unless changed to `StringBuilder.append(CharSequence, to, end)`) too without intermediary `substring`

core/src/main/java/com/linecorp/armeria/server/FallbackService.java
```diff
@@ -0,0 +1,89 @@
...
+        if (pathWithoutQuery.length() == pathWithQuery.length()) {
```
minwoox: Couldn't `routeCtx.query() == null` clearer?

core/src/main/java/com/linecorp/armeria/server/FallbackService.java
trustin: Does this look better? https://github.com/line/armeria/pull/2255/commits/8b315fab0c5a7c88eb68f0281139ed5538a50001

core/src/main/java/com/linecorp/armeria/server/FallbackService.java
minwoox: I was too late. 😅

trustin: Fixed (already) :laughing:

ikhoon:
Check style 😅
```
[ant:checkstyle] [ERROR] /appveyor/projects/armeria/core/src/main/java/com/linecorp/armeria/server/FallbackService.java:18:15: Unused import - java.util.Objects.requireNonNull. [UnusedImports]
[ant:checkstyle] [ERROR] /appveyor/projects/armeria/core/src/main/java/com/linecorp/armeria/server/VirtualHost.java:35:8: Unused import - com.linecorp.armeria.common.HttpStatus. [UnusedImports]
```

core/src/main/java/com/linecorp/armeria/server/Route.java
```diff
@@ -31,6 +31,13 @@
  */
 public interface Route {
 
+    /**
+     * Returns the catch-all {@link Route} which matches every request.
+     */
+    static Route catchAll() {
+        return RouteBuilder.CATCH_ALL_ROUTE;
```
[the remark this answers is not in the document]
trustin: That's a good idea. Renamed.

trustin:
Thanks for reviewing!

PR #2259, opened 2019-11-13
Always use `ServiceRequestContext` on the server side

trustin:
Motivation:

`HttpServerHandler` creates an `EarlyRespondingRequestContext` which
does not implement `ServiceRequestContext` when handling an invalid
request. The default `AccessLogWriter` implementations expects all
`RequestContext`s on the server side are `ServiceRequestContext`. As a
result, an invalid request triggers a `ClassCastException` in
`AccessLogWriter`.

Such an assumption that a service-side context is always
`ServiceRequestContext` is reasonable, so we should always provide a
`ServiceRequestContext` rather than just a `RequestContext`.

Modifications:

- Remove `EarlyRespondingRequestContext`.
- Always use `DefaultServiceRequestContext`, even for invalid requests
  and the `OPTIONS *` request.
- Enable access logs for the test cases that tests invalid paths.
- Simplify the response generation logic in `HttpServerHandler`.
- Simplify how `HttpServerHandler` retrieves the current `SSLSession`.

Result:

- `AccessLogWriter` always sees a `ServiceRequestContext`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
@@ -22,7 +22,6 @@ import static com.linecorp.armeria.common.SessionProtocol.H2;
 import static com.linecorp.armeria.common.SessionProtocol.H2C;
 import static com.linecorp.armeria.common.stream.SubscriptionOption.WITH_POOLED_OBJECTS;
 import static com.linecorp.armeria.internal.ArmeriaHttpUtil.isCorsPreflightRequest;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java
```diff
@@ -469,27 +466,13 @@ private static String hostname(RequestHeaders headers) {
         return authority.substring(0, hostnameColonIdx);
     }
 
-    // TODO(minwoox) Refactor response() methods so that they are easily read
```
minwoox: 👍

trustin:
Thanks for reviewing.

Issue #2243, opened 2019-11-09
Domain name resolution failure in Windows


trustin:
Not sure if it affects all Windows machines, but got a report from at least one user. Full DNS trace:

```
2019-11-09 21:34:02.482 DEBUG 16800 --- [-worker-nio-2-2] o.s.w.s.adapter.HttpWebHandlerAdapter    : [24b42788] HTTP GET "/random"
2019-11-09 21:34:02.506 DEBUG 16800 --- [-worker-nio-2-2] s.w.r.r.m.a.RequestMappingHandlerMapping : [24b42788] Mapped to com.linecorp.devday.handson.demo.controller.DemoController#getRandom()
2019-11-09 21:34:02.546 DEBUG 16800 --- [-worker-nio-2-2] o.s.w.r.r.m.a.ResponseBodyResultHandler  : Using 'application/json' given [*/*] and supported [application/json, application/*+json, text/event-stream]
2019-11-09 21:34:02.546 DEBUG 16800 --- [-worker-nio-2-2] o.s.w.r.r.m.a.ResponseBodyResultHandler  : [24b42788] 0..1 [com.linecorp.devday.handson.demo.model.DemoModel]
2019-11-09 21:34:02.614 DEBUG 16800 --- [-worker-nio-2-2] o.s.w.r.f.client.ExchangeFunctions       : [7f5065ca] HTTP GET http://echo.jsontest.com/randomNumber/3/uuid/8a5870e4-834b-4359-aaf7-671289f72102
2019-11-09 21:34:02.683 DEBUG 16800 --- [-worker-nio-2-1] i.n.util.ResourceLeakDetectorFactory     : Loaded default ResourceLeakDetector: io.netty.util.ResourceLeakDetector@74984073
2019-11-09 21:34:02.698 DEBUG 16800 --- [-worker-nio-2-1] io.netty.resolver.dns.DnsQueryContext    : [id: 0x294edb68] WRITE: UDP, [30527: /10.114.0.1:53], DefaultDnsQuestion(echo.jsontest.com. IN A)
2019-11-09 21:34:02.703 DEBUG 16800 --- [-worker-nio-2-1] io.netty.resolver.dns.DnsQueryContext    : [id: 0x294edb68] WRITE: UDP, [57985: /10.114.0.1:53], DefaultDnsQuestion(echo.jsontest.com. IN AAAA)
[...]

PR #2261, opened 2019-11-14
Allow configuring DefaultAddressResolverGroup by flag

ikhoon:
Motivation:
Domain name resolution failure was reported on Windows OS. #2243
Before solving the original problem, this workaround enables [`DefaultAddressResolverGroup`](https://netty.io/4.1/api/io/netty/resolver/class-use/DefaultAddressResolverGroup.html) by the flag. This flag is disabled by default.
The `DefaultAddressResolverGroup` uses JDK's built-in domain name lookup mechanism. See [here](https://netty.io/4.1/api/io/netty/resolver/DefaultNameResolver.html) for more information.

Modification:
* Add `useJdkDnsResolver` flag to enable `DefaultAddressResolverGroup`.
* Use `DefaultAddressResolverGroup.INSTANCE` if it is possible.

Result:
A user can now use  `DefaultAddressResolverGroup` by enabling `useJdkDnsResolver` flag.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
@@ -455,10 +455,16 @@ public final class ClientFactoryBuilder {
                                                                maxNumEventLoopsFunctions);
         }
 
+        // FIXME(ikhoon) This workaround enables DefaultAddressResolverGroup by default for Windows OS.
+        // The DefaultAddressResolverGroup which calls blocking operation uses JDK's built-in domain name lookup
+        // mechanism. However DNS result is easy to cache, so it doesn't have a big impact on performance.
+        // The code that uses DefaultAddressResolverGroup could be removed after resolving the following issue.
+        // https://github.com/line/armeria/issues/2243
         if (Flags.useDefaultDnsResolver() &&
             addressResolverGroupFactory == null && dnsResolverGroupCustomizers == null) {
             addressResolverGroupFactory(unused -> DefaultAddressResolverGroup.INSTANCE);
         }
+
         final AddressResolverGroup<InetSocketAddress> addressResolverGroup;
         if (addressResolverGroupFactory != null) {
             @SuppressWarnings("unchecked")
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
@@ -455,7 +455,7 @@ public final class ClientFactoryBuilder {
                                                                maxNumEventLoopsFunctions);
         }
 
-        // FIXME(ikhoon) This workaround enables DefaultAddressResolverGroup by default for Windows OS.
+        // FIXME(ikhoon) This workaround enables DefaultAddressResolverGroup by `useDefaultDnsResolver` flag.
         // The DefaultAddressResolverGroup which calls blocking operation uses JDK's built-in domain name lookup
         // mechanism. However DNS result is easy to cache, so it doesn't have a big impact on performance.
         // The code that uses DefaultAddressResolverGroup could be removed after resolving the following issue.
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
@@ -455,12 +455,9 @@ public final class ClientFactoryBuilder {
                                                                maxNumEventLoopsFunctions);
         }
 
-        // FIXME(ikhoon) This workaround enables DefaultAddressResolverGroup by `useDefaultDnsResolver` flag.
-        // The DefaultAddressResolverGroup which calls blocking operation uses JDK's built-in domain name lookup
-        // mechanism. However DNS result is easy to cache, so it doesn't have a big impact on performance.
-        // The code that uses DefaultAddressResolverGroup could be removed after resolving the following issue.
-        // https://github.com/line/armeria/issues/2243
-        if (Flags.useDefaultDnsResolver() &&
+        // FIXME(ikhoon): Remove DefaultAddressResolverGroup registration after fixing Window domain name
+        //                resolution failure. https://github.com/line/armeria/issues/2243
+        if (Flags.useJdkDnsResolver() &&
             addressResolverGroupFactory == null && dnsResolverGroupCustomizers == null) {
             addressResolverGroupFactory(unused -> DefaultAddressResolverGroup.INSTANCE);
         }
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
@@ -53,6 +53,7 @@ import io.netty.channel.epoll.EpollChannelOption;
 import io.netty.handler.codec.http2.Http2CodecUtil;
 import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.resolver.AddressResolverGroup;
+import io.netty.resolver.DefaultAddressResolverGroup;
 import io.netty.resolver.dns.DnsNameResolverBuilder;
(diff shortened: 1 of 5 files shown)


anuraaga:
Looking at the linked doc this seems to do blocking I/O - so I think we'd have to do more refactoring to use blocking executor. Not sure if it's worth it without getting more reports. I guess with Windows it's easy to enable the firewall on your app which could block udp packets coming back

ikhoon:
> Looking at the linked doc this seems to do blocking I/O.

When I was implementing this logic, I also read the documentation and got advice from @trustin. 
He said we don't need to change the executor(EventLoop). It was late evening. So I was also wondering about the reason. I think this is the time to hear his opinion. 😎

@trustin Could you tell us why we don't need to change the given executor? 🙏

ikhoon:
> I think we'd have to do more refactoring to use blocking executor.

Had talked with @trustin. `DefaultNameResolver` does not run a domain name resolution on the given `EventExecutor`. 
https://github.com/netty/netty/blob/b91dd0678d290a70817ddc44bb5ced2baed275fd/resolver/src/main/java/io/netty/resolver/DefaultNameResolver.java#L40-L47
To make `DefaultNameResolver` use blocking executor, I think we might need to override upstream behavior. 

I'm still wondering if it is worth implementing? Because DNS is cacheable, we will fix the original issue on the next step.

I will leave some comments about this on code :-) WDYT?

anuraaga:
The performance might be ok - we'd need to verify that it's not possible to deadlock. I'm not sure what would happen if a client and DNS request use the same event loop, but I have seen deadlock when blocking the event loop during a client request so I'm very scared of doing it.

I might rethink whether this is really needed (I'm currently assuming we've only had one report, the attached one, which might not indicate such a big hammer but if there are more you know about then it might be), but even if adding I'd recommend not making it default. Tests working on appveyor  (and my laptop and [GitHub's Azure VMs](https://github.com/anuraaga/armeria/commit/af8e4ade09c31469ec9fd300b25b6ca9c56847a5/checks?check_suite_id=306486137) :) ) I think give us reasonable confidence in our current defaults

ikhoon:
> I have seen deadlock when blocking the event loop during a client request so I'm very scared of doing it.

Just curious, Could you explain to me how blocking causes deadlock? 😀

anuraaga:
I've seen it here - https://github.com/openzipkin/zipkin/pull/2703/files

It happens when a client request is made in a blocking way from a server request - if the client and server request are being handled by the same event loop, the client will wait for I/O forever while the server has blocked the event loop waiting for the client request to finish. I'm not sure if the same is an issue in this case, but it makes me very nervous :)

ikhoon:
>  Tests working on appveyor (and my laptop and GitHub's Azure VMs :) 

Could you take a DNS test for `http://echo.jsontest.com` and `http://json.org`? If you can test it, I want to get your result. 😀

>  but even if adding I'd recommend not making it default.

We have discussed Windows domain resolution issue and decide to divide this into 3 steps.

#### 1st step
* Offer DefaultAddressResolverGroup which resolves DNS using Jdk resolver.
* Turn off default option for every OS. 
* Use Blocking operation.

If this is fine, we could merge this PR after turning off the flag. If you don't agree with the blocking operation even though it turns off the default option, I will fork Netty DefaultAddressResolverGroup and fetch it to run domain resolution with a given executor. 🛠

#### 2nd step 
* After testing, if Netty async DNS resolver has a problem with basic Windows configuration, turning on default flag only for Window.

#### 3th step
* Investing the original problem and fetch Netty or Armeria code.

WDYT? @anuraaga /cc @minwoox @trustin

anuraaga:
I think that looks fine since we don't change the default without more investigation. Only thing I'd add is for 2nd step, if changing the default also be sure that the blocking won't cause deadlock. I guess it won't - the pattern is probably doing one more I/O after the block which wouldn't happen since DNS is the lowest level possible - but want to be confidence since it's very confusing to debug

ikhoon:
> Only thing I'd add is for 2nd step, if changing the default also be sure that the blocking won't cause deadlock.

That sounds good to me. When we need to change the default, I will change default DNS resolver to run with blocking executor. 👍

ikhoon:
Updated: 
* This flag is disabled by default.
* Update PR title, description, and code.

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -297,6 +298,8 @@
             exceptionLoggingMode("annotatedServiceExceptionVerbosity",
                                  DEFAULT_ANNOTATED_SERVICE_EXCEPTION_VERBOSITY);
 
+    private static final boolean USE_DEFAULT_DNS_RESOLVER = getBoolean("useDefaultDnsResolver", false);
```
trustin: `useJdkDnsResolver` ?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -845,6 +848,19 @@ public static ExceptionVerbosity annotatedServiceExceptionVerbosity() {
         return ANNOTATED_SERVICE_EXCEPTION_VERBOSITY;
     }
 
+    /**
+     * Enables {@link DefaultAddressResolverGroup} that resolves using JDK's built-in domain name
```
trustin: resolves -> resolves domain names

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -845,6 +848,19 @@ public static ExceptionVerbosity annotatedServiceExceptionVerbosity() {
...
+     * lookup mechanism.
+     * Note that this resolver performs a blocking name lookup from the caller thread.
```
trustin: - this resolver -> JDK's built-in resolver
- thread. -> thread, and thus this flag should be enabled only when the default asynchronous resolver does not work as expected.

core/src/test/java/com/linecorp/armeria/client/ClientFactoryBuilderTest.java
```diff
@@ -21,6 +21,10 @@
 import static org.mockito.Mockito.mock;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
+import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
```
trustin: Thanks for education :-)

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -454,6 +455,16 @@ public ClientFactory build() {
                                                                maxNumEventLoopsFunctions);
         }
 
+        // FIXME(ikhoon) This workaround enables DefaultAddressResolverGroup by `useDefaultDnsResolver` flag.
+        // The DefaultAddressResolverGroup which calls blocking operation uses JDK's built-in domain name lookup
+        // mechanism. However DNS result is easy to cache, so it doesn't have a big impact on performance.
+        // The code that uses DefaultAddressResolverGroup could be removed after resolving the following issue.
+        // https://github.com/line/armeria/issues/2243
```
trustin: - The first sentence of a FIXME/TODO should explicitly state 'what has to be done'.
- Please indent properly if this whole paragraphs is part of the FIXME:
  ```
  // FIXME(ikhoon): blah
  //                blaah
  //                blaaah
  ```
- Missing colon after FIXME

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -845,6 +848,20 @@ public static ExceptionVerbosity annotatedServiceExceptionVerbosity() {
         return ANNOTATED_SERVICE_EXCEPTION_VERBOSITY;
     }
 
+    /**
+     * Enables {@link DefaultAddressResolverGroup} that resolves domain name using JDK's built-in domain name
+     * lookup mechanism.
+     * Note that JDK's built-in resolver performs a blocking name lookup from the caller thread, and thus
+     * this flag should be enabled only when the default asynchronous resolver does not work as expected.
```
anuraaga: Maybe can add something like `, for example by always throwing a {@code FooException}.`

PR #2263, opened 2019-11-14
Make sure a CORS preflight request doesn't go to a wrong service

trustin:
Motivation:

Currently, with the following code:

    Server.builder()
          .route().get("/foo").build(someService);

a CORS preflight request (OPTIONS method) will go to `someService`,
which is not an expected behavior.

It is mainly because of the hack we added some time ago for adding
support for `@CorsDecorator`:

- `DefaultRoute` - https://github.com/line/armeria/blob/armeria-0.95.0/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java#L88-L99
- `AnnotatedHttpServiceFactory` - https://github.com/line/armeria/blob/armeria-0.95.0/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java#L368-L378

Modifications:

- Add `RoutingResultType` to represent three routing state:
  - `MATCHED`
  - `NOT_MATCHED`
  - `CORS_PREFLIGHT` (not matched, but need to handle a CORS preflight request)
- Modify the routing logic so that:
  - If the routing result type is `CORS_PREFLIGHT` and the service is
    decorated with a `CorsService`, the request goes to the service.
  - If the routing result type is `CORS_PREFLIGHT` and the service is
    NOT decorated with a `CorsService`, the request goes to the fallback
    service.
- Add a test case for what's explained in the 'motivation' section.

Result:

- Slightly cleaner hack
- A CORS preflight request does not go to a wrong service anymore.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/RoutingResultBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/RoutingResultBuilder.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.server;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.linecorp.armeria.server.RoutingResult.LOWEST_SCORE;
 import static java.util.Objects.requireNonNull;
 
@@ -30,6 +31,8 @@ import com.linecorp.armeria.internal.ArmeriaHttpUtil;
  * Builds a new {@link RoutingResult}.
  */
 public final class RoutingResultBuilder {
+
+    private RoutingResultType type = RoutingResultType.MATCHED;
 
     @Nullable
     private String path;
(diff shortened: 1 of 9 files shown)


trustin:
@hyangtack @anuraaga You might be interested in this change.

core/src/main/java/com/linecorp/armeria/server/RoutingResultBuilder.java
```diff
@@ -51,6 +54,20 @@
...
+    /**
+     * Sets the result type.
+     *
+     * @param type {@link RoutingResultType#MATCHED} or {@link RoutingResultType#CORS_PREFLIGHT}.
+     */
+    public RoutingResultBuilder type(RoutingResultType type) {
+        requireNonNull(type, "type");
+        checkArgument(type != RoutingResultType.NOT_MATCHED,
+                      "type: %s (expected: %s or %s)",
+                      RoutingResultType.MATCHED, RoutingResultType.CORS_PREFLIGHT);
```
minwoox: nit: `type, RoutingResultType.MATCHED, RoutingResultType.CORS_PREFLIGHT`

trustin:
Added the 'breaking-change' label because it adds some behavioral changes in the routing API. I think it will not affect most users, though.

PR #2269, opened 2019-11-16
Change blockingTaskExecutor from Executor to ScheduledExecutorS…

minwoox:
Motivation:
It will be good if a user can schedule a task using `blockingTaskExecutor`.

Modification:
- Change `blockingTaskExecutor` from `Executor` to `ScheduledExecutorService`.

Result:
- You can now schedule a task using `blockingTaskExecutor`.
- (Breaking) `ServerBuilder.blockingTaskExecutor(Executor executor, ...)` is now takes `ScheduledExecutorService`.

### DIFF
--- a/site/src/sphinx/advanced-production-checklist.rst
+++ b/site/src/sphinx/advanced-production-checklist.rst
@@ -29,7 +29,7 @@ You may want to consider the following options before putting your Armeria appli
       import com.linecorp.armeria.server.ServerBuilder;
 
       ServerBuilder sb = Server.builder();
-      sb.blockingTaskExecutor(my600ThreadsExecutor);
+      sb.blockingTaskExecutor(myScheduledExecutorService);
 
 - Specify the default limits of an HTTP request or response.
 
--- a/site/src/sphinx/advanced-production-checklist.rst
+++ b/site/src/sphinx/advanced-production-checklist.rst
@@ -21,15 +21,15 @@ You may want to consider the following options before putting your Armeria appli
 
 - Specify an alternative ``blockingTaskExecutor`` based on expected workload if your server has
   a service that uses it, such as :api:`TomcatService`, :api:`JettyService` and :api:`THttpService` with
-  synchronous service implementation. The default is a simple ``ThreadPoolExecutor`` with 200 threads and an
-  *unbounded* queue, provided by :api:`CommonPools`.
+  synchronous service implementation. The default is a simple ``ScheduledThreadPoolExecutor`` with maximum
+  200 threads, provided by :api:`CommonPools`.
 
   .. code-block:: java
 
       import com.linecorp.armeria.server.ServerBuilder;
 
       ServerBuilder sb = Server.builder();
-      sb.blockingTaskExecutor(myBoundedExecutor);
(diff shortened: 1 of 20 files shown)


minwoox:
@renaudb You might want to see this because you are using the customized `ThreadPoolExecutor`.

trustin:
@renaudb I think you can still achieve the same effect by taking account of the number of pending tasks using an `AtomicInteger` and rejecting the request before calling `ExecutorService.execute()`. However, if you think the inability to specify a `SynchronousQueue` as a task queue is a problem, as a follow-up PR, I guess we can provide a `ScheduledExecutorService` implementation which is a hybrid of `ThreadPoolExecutor` and `ScheduledThreadPoolExecutor`:

```java
public class HybridExecutorService implements ScheduledExecutorService {

    ThreadPoolExecutor executor = ...;
    ScheduledThreadPoolExecutor scheduledExecutor = ...;

    public void execute(Runnable task) { 
        executor.execute(task);
    }

    public ScheduledFuture<?> schedule(Runnable task, ...) {
        scheduledExecutor.schedule(() -> execute(task), ...);
        ... some future delegation ...
    }
}
```

Thoughts?

ikhoon:
nit: Don't we need to update the `blockingTaskExecutor` documentation in `Production checklist`?
https://github.com/line/armeria/blob/f03b400cff136a710dbdbeb75e0e4f741387d296/site/src/sphinx/advanced-production-checklist.rst#L22-L25

renaudb:
No, this is fine, we might just assign a specific context aware ExecutorService just to Solr and build our own API on top to make it look async.

minwoox:
> nit: Don't we need to update the blockingTaskExecutor documentation in Production checklist?

Thanks a lot for finding this. 😄

minwoox:
@ikhoon Fix. PTAL 🙇

site/src/sphinx/advanced-production-checklist.rst
```diff
@@ -21,15 +21,15 @@ You may want to consider the following options before putting your Armeria appli
...
+  synchronous service implementation. The default is a simple ``ScheduledThreadPoolExecutor`` with maximum
+  200 threads, provided by :api:`CommonPools`.
 
   .. code-block:: java
 
       import com.linecorp.armeria.server.ServerBuilder;
 
       ServerBuilder sb = Server.builder();
-      sb.blockingTaskExecutor(myBoundedExecutor);
+      sb.blockingTaskExecutor(my600ThreadsExecutor);
```
ikhoon: nit: I'm not sure this suggestion is good. 😅 `myScheduledExecutorService`?

ikhoon:
@minwoox Thanks for listening to my opinion 👍 Cool Beans!

site/src/sphinx/advanced-production-checklist.rst
minwoox: The queue of default scheduled pool executor is always unbounded. So I used your first suggestion. 😉

trustin:
Nice job, @minwoox !

minwoox:
Thanks for reviewing! :wink:

Issue #2272, opened 2019-11-19
Request time does not take into account connection creation


anuraaga:
Reported in https://line-armeria.slack.com/archives/C1NGPBUH2/p1574122806097100

I believe this is a regression after moving `Endpoint` selection to the begging of the request. We call `startRequest` only when about to send the request, but not anymore. This is affecting spans, but I think it also affects our `totalDuration` metric equally.

anuraaga:
To some extent, where we start and stop the request timing is up to us - but especially since we have the wire annotations now, it seems fine to me to have the total span duration match the invocation of the stub method (I guess basically `invoke() -> completionFuture()`). A good enough approximation I'm seeing is to just set the start time when a `ClientRequestContext` is constructed.

trustin:
+1 for setting the start time when the context is constructed

anuraaga:
It might be great timing to reconsider the `RequestLog` API - I find the `RequestLogAvailability` to be so fine-grained that it's fairly hard to understand, and it makes working with logs (e.g., adding a new logged metric) in Armeria very tedious. I wonder how many people need any availability other than `Complete` - I think log usually ends up being processed / stored into something after a request has finished in most (all?) use cases. With the brave sampler integration, I think we found that the availability didn't really matter and was making the code more complicated when it just needed to either have the information if already available or skip it. Any thoughts? Should we open another issue?

anuraaga:
So - any thoughts on the change to the logging API regards to availability setting? :) @trustin @minwoox @ikhoon

trustin:
Sorry, missed your comment. How about assuming `COMPLETE` when a user specified no availabilities? That would be less confusing I guess.

anuraaga:
@trustin I find it more difficult to actually log things, since it involves figuring out or making a new availability (which are technically not infinite with the current implementation using a bitset). I think your suggestion is referring to consuming logs, but I'm more concerned about adding logs. 

For this issue, right now, we have `REQUEST_START` which contains the start time, channel, and other bits, so if we want to change the timing of when we log `requestStartTimeMillis`, it's very tedious due to the coupling. It was similar when adding the `firstBytesTransferred` metrics. I'm not sure the value add of these availabilities is enough to warrant it being so difficult to add logging.

trustin:
[...]

PR #2273, opened 2019-11-19
Add connection timing annotations to client spans.

### LINKED ISSUE #2271: Add information from ClientConnectionTimings in BraveClient
Armeria already provides detailed connection timing information such as DNS resolution time, but BraveClient currently does not take advantage of it yet.

anuraaga:
Annotations are displayed in the zipkin UI as a table to be able to see time spent in individual steps of a span. It should be useful to users to be able to see client connection timings in spans.

Interestingly, until #2272 it's likely these annotations will happen before the span start time, which is not how spans are normally recorded. It shouldn't hurt though to have them since they're just numbers in a table. The other issue is actually a lot more work, so in the meantime at least this provides useful numbers.

Misc: I found that there would always be a 30-60s pause between `yarnSetup` and `yarn` on my computer and found it's because `gradle-node-plugin` adds `node_modules` as a task output, causing Gradle to hash every single file in there which takes forever. Without defining it as an output, it means if a developer mucks with `node_modules` with manual edits, `yarn` won't run again, which is arguably a good thing anyways.

Fixes #2271

### DIFF
--- a/brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
+++ b/brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
@@ -18,6 +18,7 @@ package com.linecorp.armeria.client.brave;
 
 import static com.linecorp.armeria.internal.brave.TraceContextUtil.ensureScopeUsesRequestContext;
 
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
 import javax.annotation.Nullable;
@@ -25,6 +26,7 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.linecorp.armeria.client.ClientConnectionTimings;
 import com.linecorp.armeria.client.ClientRequestContext;
 import com.linecorp.armeria.client.HttpClient;
 import com.linecorp.armeria.client.SimpleDecoratingHttpClient;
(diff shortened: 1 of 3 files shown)


brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -141,4 +163,9 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
             return delegate().execute(ctx, req);
         }
     }
+
+    private void logTiming(Span span, String prefix, long startTimeMicros, long durationNanos) {
+        span.annotate(startTimeMicros, prefix + ".start");
+        span.annotate(startTimeMicros + TimeUnit.NANOSECONDS.toMicros(durationNanos), prefix + ".end");
```
trustin: Will JVM optimize away the string concatenation here?

anuraaga: Guess not - switched to hardcoding

trustin:
> Misc: I found that there would always be a 30-60s pause between yarnSetup and yarn on my computer and found it's because gradle-node-plugin adds node_modules as a task output, causing Gradle to hash every single file in there which takes forever. Without defining it as an output, it means if a developer mucks with node_modules with manual edits, yarn won't run again, which is arguably a good thing anyways.

Are you going to send a PR for this?

anuraaga:
That got merged with my other PR (all my recent PRs had the patch ;) )

PR #2275, opened 2019-11-20
Return a CompletableFuture from API methods that checks whether…

anuraaga:
…ing methods are called from an event loop.

We have heard many stories of bad experiences because of blocking the event loop in an armeria server request. While we wouldn't be able to check all cases of this (e.g., calling JDBC from an event loop), it's fairly simple to check this for futures returned by Armeria methods and I think this covers a lot of cases.  This changes to return a special implementation of `CompletableFuture` that logs from blocking methods with a stack trace.

Let me know any general feedback and then I'll apply to other methods (though I guess `aggregate()` is the most common way to get a future from Armeria).

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -873,12 +873,12 @@ public final class Flags {
      * Returns whether {@link CompletableFuture}s returned by Armeria methods log a warning if
      * {@link CompletableFuture#join()} or {@link CompletableFuture#get()} are called from an event loop thread.
      * Blocking an event loop thread in this manner reduces performance significantly, possibly causing
-     * deadlocks, so it should be avoided at all costs (e.g. using thenApply type methods to execute
+     * deadlocks, so it should be avoided at all costs (e.g. using {@code thenApply()} type methods to execute
      * asynchronously or running the logic using {@link ServiceRequestContext#blockingTaskExecutor()}.
      *
-     * <p>This flag is disabled by default.
+     * <p>This flag is enabled by default.
      * Specify the {@code -Dcom.linecorp.armeria.reportBlockedEventLoop=false} JVM option
-     * to enable it.
+     * to disable it.
      */
     public static boolean reportBlockedEventLoop() {
         return REPORT_BLOCKED_EVENT_LOOP;
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -21,6 +21,7 @@ import java.nio.channels.ClosedChannelException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.IntPredicate;
 import java.util.function.LongPredicate;
 import java.util.function.Predicate;
@@ -51,6 +52,7 @@ import com.linecorp.armeria.server.RoutingContext;
 import com.linecorp.armeria.server.ServerBuilder;
 import com.linecorp.armeria.server.Service;
 import com.linecorp.armeria.server.ServiceConfig;
+import com.linecorp.armeria.server.ServiceRequestContext;
 import com.linecorp.armeria.server.annotation.ExceptionHandler;
 import com.linecorp.armeria.server.annotation.ExceptionVerbosity;
 
@@ -300,6 +302,9 @@ public final class Flags {
                                  DEFAULT_ANNOTATED_SERVICE_EXCEPTION_VERBOSITY);
 
     private static final boolean USE_JDK_DNS_RESOLVER = getBoolean("useJdkDnsResolver", false);
+
+    private static final boolean REPORT_BLOCKED_EVENT_LOOP =
+            getBoolean("reportBlockedEventLoop", true);
 
     static {
         if (!isEpollAvailable()) {
@@ -862,6 +867,21 @@ public final class Flags {
      */
     public static boolean useJdkDnsResolver() {
         return USE_JDK_DNS_RESOLVER;
+    }
+
+    /**
+     * Returns whether {@link CompletableFuture}s returned by Armeria methods log a warning if
+     * {@link CompletableFuture#join()} or {@link CompletableFuture#get()} are called from an event loop thread.
+     * Blocking an event loop thread in this manner reduces performance significantly, possibly causing
+     * deadlocks, so it should be avoided at all costs (e.g. using thenApply type methods to execute
+     * asynchronously or running the logic using {@link ServiceRequestContext#blockingTaskExecutor()}.
+     *
+     * <p>This flag is disabled by default.
+     * Specify the {@code -Dcom.linecorp.armeria.reportBlockedEventLoop=false} JVM option
+     * to enable it.
+     */
+    public static boolean reportBlockedEventLoop() {
+        return REPORT_BLOCKED_EVENT_LOOP;
     }
 
     private static Optional<String> caffeineSpec(String name, String defaultValue) {
--- a/core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingCompletableFuture.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingCompletableFuture.java
@@ -60,7 +60,7 @@ public final class EventLoopCheckingCompletableFuture<T> extends CompletableFutu
         return super.join();
     }
 
-    private void maybeLogIfOnEventLoop() {
+    private static void maybeLogIfOnEventLoop() {
         if (!Flags.reportBlockedEventLoop()) {
             return;
         }
--- a/core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingCompletableFuture.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingCompletableFuture.java
@@ -66,10 +66,10 @@ public final class EventLoopCheckingCompletableFuture<T> extends CompletableFutu
         }
         final Thread thread = Thread.currentThread();
         if (thread instanceof NonBlocking && REPORTED_THREADS.add(thread)) {
-            logger.warn("Calling a blocking method on CompletableFuture from an event loop thread. You " +
-                        "should never do this as this will usually result in significantly reduced " +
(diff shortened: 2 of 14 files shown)


trustin:
Maybe a user can just use BlockHound? https://github.com/reactor/BlockHound
Netty integration is coming soon: https://github.com/netty/netty/pull/9687

anuraaga:
BlockHound seems cool and could be a best practice when running an Armeria app to catch other things like JDBC. But I think registering an agent is tedious enough that it could still be worth having something out of the box. The cost / performance of manually instrumenting these three common methods seems pretty good to me. What do you think?

anuraaga:
@trustin Any more thoughts on this? I notice there could be JDK version issues in the future with BlockHound - https://github.com/reactor/BlockHound/issues/33. It's a very cool tool, but agents are so complex I think this is could be a small but effective improvement.

trustin:
Sorry. Let me check soon. Didn't have a chance to check the code yet.

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -301,6 +303,9 @@
 
     private static final boolean USE_JDK_DNS_RESOLVER = getBoolean("useJdkDnsResolver", false);
 
+    private static final boolean CHECK_FUTURE_JOIN_ON_EVENT_LOOP =
+            getBoolean("checkFutureJoinOnEventLoop", true);
```
trustin: How about something a little bit more generic like `reportBlockedEventLoop`?

core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingCompletableFuture.java
```diff
@@ -0,0 +1,68 @@
...
+        if (Flags.checkFutureJoinOnEventLoop() && Thread.currentThread() instanceof EventLoopThread) {
+            logger.warn("Calling a blocking method on CompletableFuture from an event loop thread. You " +
+                        "should never do this as this will usually result in significantly reduced " +
+                        "performance of the server, generally crippling its ability to handle high load, or " +
+                        "even result in deadlock which cannot be recovered from. Use " +
+                        "ServiceRequestContext.blockingExecutor to run this logic instead or switch to using " +
+                        "asynchronous methods like thenApply. If you really believe it is fine to block the " +
+                        "event loop like this, you can disable this log message by specifying the " +
+                        "-Dcom.linecorp.armeria.checkFutureJoinOnEventLoop=false system property",
+                        new IllegalStateException("Blocking event loop, don't do this."));
```
trustin: - Perhaps we should log only once for each thread? Could keep a weak set of `Thread`s?
- What about referring to the URL of some documentation which explains this well?

core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopThread.java
```diff
@@ -22,8 +22,8 @@
  * A {@link FastThreadLocalThread} that implements {@link NonBlocking} tag interface to make Project Reactor
  * happy.
  */
-final class EventLoopThread extends FastThreadLocalThread implements NonBlocking {
-    EventLoopThread(ThreadGroup threadGroup, Runnable r, String name) {
+public final class EventLoopThread extends FastThreadLocalThread implements NonBlocking {
```
trustin: How about using `instanceof NonBlocking`? It looks hacky though :sweat_smile:

core/src/test/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingCompletableFutureTest.java
```diff
@@ -0,0 +1,114 @@
...
+                .addAppender(appender);
+    }
+
+    @AfterEach
+    void cleanupLogger() {
+        ((ch.qos.logback.classic.Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME))
+                .detachAppender(appender);
+    }
+
+    @Test void joinOnEventLoop() {
```
trustin: nit: Line break between `@Test` and `void`?

core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingCompletableFuture.java
anuraaga: Think we might have to write that documentation 😄

anuraaga:
I have switched to `EventLoopCheckingCompletableFuture` where we return a future to users. Though it might be simpler to just use it everywhere, let me know.

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -864,6 +869,21 @@ public static boolean useJdkDnsResolver() {
         return USE_JDK_DNS_RESOLVER;
     }
 
+    /**
+     * Returns whether {@link CompletableFuture}s returned by Armeria methods log a warning if
+     * {@link CompletableFuture#join()} or {@link CompletableFuture#get()} are called from an event loop thread.
+     * Blocking an event loop thread in this manner reduces performance significantly, possibly causing
+     * deadlocks, so it should be avoided at all costs (e.g. using thenApply type methods to execute
```
trustin: `thenApply` -> `{@code thenApply()}` ?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -864,6 +869,21 @@ public static boolean useJdkDnsResolver() {
...
+     * asynchronously or running the logic using {@link ServiceRequestContext#blockingTaskExecutor()}.
+     *
+     * <p>This flag is disabled by default.
+     * Specify the {@code -Dcom.linecorp.armeria.reportBlockedEventLoop=false} JVM option
+     * to enable it.
```
trustin: Not enabling by default?

anuraaga: Oops

anuraaga:
Hmm test doesn't seem to work well on Windows. Happen to have any ideas why that could be?

trustin:
Maybe `REPORTED_THREADS.add()` returned `false`? How about using a different event loop for each test method?

anuraaga:
Doh duh... thanks for finding it!

trustin:
`maybeLogIfOnEventLoop` could be `static` by the way :-)

core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingCompletableFuture.java
```diff
@@ -0,0 +1,80 @@
...
+        maybeLogIfOnEventLoop();
+        return super.join();
+    }
+
+    private static void maybeLogIfOnEventLoop() {
+        if (!Flags.reportBlockedEventLoop()) {
+            return;
+        }
+        final Thread thread = Thread.currentThread();
+        if (thread instanceof NonBlocking && REPORTED_THREADS.add(thread)) {
```
ikhoon: IIUC this [`NonBlocking` interface](https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/scheduler/NonBlocking.java) is created in the project reactor.

https://github.com/reactor/reactor-core/blob/628258f644bdddd095e72d1fc81c5d6ac0e3fa12/reactor-core/src/main/java/reactor/core/scheduler/ReactorThreadFactory.java#L92-L96

It seems an edge case but a [non-eventloop thread](https://github.com/reactor/reactor-core/blob/628258f644bdddd095e72d1fc81c5d6ac0e3fa12/reactor-core/src/main/java/reactor/core/scheduler/ReactorThreadFactory.java#L92) might pass this condition and log this message. I am wondering if it is acceptable. Unless I suggest changing `EventLoopCheckingCompletableFuture` to `NonBlockCheckingCompletableFuture` and related message. 😀

minwoox: Armeria `EventLoopThread` implements `Nonblocking`. 😄

anuraaga: I originally checked `instanceof EventLoopThread` but changed to `NonBlocking` I think to avoid moving a package-local `EventLoopThread` to public in `internal` package. @trustin Any thoughts? I lean towards it being small enough not to worry about

ikhoon: I know that :-) Just worried other non-event loop thread could implement `NonBlocking`. 
But it seems too many worries. 😜

minwoox: I think we can warn that as well because a thread which implements `NonBlocking` is doing blocking operation. 😄

ikhoon: Originally I wasn't worried about the behavior. My concern is the log message.
Sorry for the confusion. 😅
```diff
- Calling a blocking method on CompletableFuture from an event loop thread.
+ Calling a blocking method on CompletableFuture from a non-blocking thread.
```

ikhoon: > I lean towards it being small enough not to worry about

But I agree with this opinion!

minwoox: Ah then it makes sense. 😄

trustin: So do we need to update the WARN log message?

ikhoon: I'm not completely sure it is worth to update the message.
If we do, then leave the original sentence and put extra information just in case. 😀

```diff
- Calling a blocking method on CompletableFuture from an event loop thread.
+ Calling a blocking method on CompletableFuture from an event loop thread or non-blocking thread.
```

trustin: @anuraaga Could you update the log message?

anuraaga: Sure

PR #2277, opened 2019-11-21
Fix outdated documentation in `client-custom-http-headers.rst`

JunoJunho:
Hello Armeria fellows,

I found that some custom header example is old-fashion (not working), so I changed the example.

While I am working this job, I found that `ClientOptionsBuilder` example is now deprecated, so it would be better to change another example to utilize `ClientOptionsBuilder`, but all are deprecated. If you have any idea for that, I will update for it also.

### DIFF
--- a/site/src/sphinx/client-custom-http-headers.rst
+++ b/site/src/sphinx/client-custom-http-headers.rst
@@ -78,8 +78,8 @@ If you want more freedom on how you manipulate the request headers, use a decora
 
     // Add a decorator that inserts the custom header.
     cb.decorator((delegate, ctx, req) -> { // See DecoratingHttpClientFunction and DecoratingRpcClientFunction.
-        req.headers().set(AUTHORIZATION, credential);
-        return delegate.execute(ctx, req);
+        RequestHeaders req0 = req.headers().set(AUTHORIZATION, credential).toBuilder().build();
+        return delegate.execute(ctx, HttpRequest.of(req0, req));
     });
 
     HelloService.Iface client = cb.build(HelloService.Iface.class);
--- a/site/src/sphinx/client-custom-http-headers.rst
+++ b/site/src/sphinx/client-custom-http-headers.rst
@@ -78,8 +78,9 @@ If you want more freedom on how you manipulate the request headers, use a decora
 
     // Add a decorator that inserts the custom header.
     cb.decorator((delegate, ctx, req) -> { // See DecoratingHttpClientFunction and DecoratingRpcClientFunction.
-        RequestHeaders authorized = req.headers().toBuilder().set(AUTHORIZATION, credential).build();
-        return delegate.execute(ctx, HttpRequest.of(req, authorized));
+        HttpRequest newReq = req.withHeaders(req.headers().toBuilder().set(AUTHORIZATION, credential));
+        ctx.updateRequest(newReq);
+        return delegate.execute(ctx, newReq);
     });
 
     HelloService.Iface client = cb.build(HelloService.Iface.class);
--- a/site/src/sphinx/client-custom-http-headers.rst
+++ b/site/src/sphinx/client-custom-http-headers.rst
@@ -78,8 +78,8 @@ If you want more freedom on how you manipulate the request headers, use a decora
 
     // Add a decorator that inserts the custom header.
     cb.decorator((delegate, ctx, req) -> { // See DecoratingHttpClientFunction and DecoratingRpcClientFunction.
-        RequestHeaders req0 = req.headers().set(AUTHORIZATION, credential).toBuilder().build();
-        return delegate.execute(ctx, HttpRequest.of(req0, req));
+        RequestHeaders authorized = req.headers().toBuilder().set(AUTHORIZATION, credential).build();
+        return delegate.execute(ctx, HttpRequest.of(req, authorized));
     });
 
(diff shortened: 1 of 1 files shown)


site/src/sphinx/client-custom-http-headers.rst
```diff
@@ -78,8 +78,8 @@ If you want more freedom on how you manipulate the request headers, use a decora
 
     // Add a decorator that inserts the custom header.
     cb.decorator((delegate, ctx, req) -> { // See DecoratingHttpClientFunction and DecoratingRpcClientFunction.
-        req.headers().set(AUTHORIZATION, credential);
-        return delegate.execute(ctx, req);
+        RequestHeaders req0 = req.headers().set(AUTHORIZATION, credential).toBuilder().build();
```
anuraaga: How about naming it `authorized` instead of `req0`?

site/src/sphinx/client-custom-http-headers.rst
```diff
@@ -78,8 +78,8 @@ If you want more freedom on how you manipulate the request headers, use a decora
...
+        return delegate.execute(ctx, HttpRequest.of(req0, req));
```
anuraaga: Ooh - this should be `HttpRequest.of(req, authorized)`. But it's confusing indeed that we have many methods that take `newHeaders` as the first parameter but this one takes it second and they both compile. I have filed #2278

site/src/sphinx/client-custom-http-headers.rst
anuraaga: Also I think this is `req.headers().toBuilder().set(...).build();`

anuraaga:
Thanks a lot for helping keep the docs updated! For `ClientOptionsBuilder`, are you referring to the deprecated constructors? We have switched to using `ClientOptions.builder()` but the class and its methods are still undeprecated. If you see any usage of `new ClientOptionsBuilder()` and would like to fix it, that'd be great :)

ikhoon:
> Junho.Kim seems not to be a GitHub user.

@JunoJunho Could you amend commit author name for CLA? 😀

JunoJunho:
@ikhoon I did it (_ _)

site/src/sphinx/client-custom-http-headers.rst
ikhoon: [`RequestHeader`](https://line.github.io/armeria/apidocs/com/linecorp/armeria/common/RequestHeaders.html) is immutable. It doesn't have `set(...)` methods. We need to call `toBuilder()` to modify and set another attribute to it.  😀

minwoox: And we need to update after https://github.com/line/armeria/pull/2283 is merged.

JunoJunho: > RequestHeader is immutable. It doesn't have set(...) methods. We need to call toBuilder() to modify and set another attribute to it. 😀

Oh,, I see. Thanks for explanation!

JunoJunho: > And we need to update after #2283 is merged.

How do I deal with it?

It seems currently in `master` branch?

JunoJunho: I checked #2283, I notice that I need to use `HttpRequest.withHeaders` instead of `HttpRequest.of` ?

minwoox: Yes!, and don't forget to call `ctx.updateRequest`

site/src/sphinx/client-custom-http-headers.rst
```diff
@@ -78,8 +78,8 @@ If you want more freedom on how you manipulate the request headers, use a decora
 
     // Add a decorator that inserts the custom header.
     cb.decorator((delegate, ctx, req) -> { // See DecoratingHttpClientFunction and DecoratingRpcClientFunction.
-        req.headers().set(AUTHORIZATION, credential);
-        return delegate.execute(ctx, req);
+        RequestHeaders authorized = req.headers().toBuilder().set(AUTHORIZATION, credential).build();
+        return delegate.execute(ctx, HttpRequest.of(req, authorized));
```
trustin: How about:

```java
HttpRequest newReq = req.withHeaders(req.headers().toBuilder()
                                        .set(AUTHORIZATION, credential));
ctx.updateRequest(newReq);
return delegate.execute(ctx, newReq);
```

trustin:
Re: `ClientOptionsBuilder` - Just replacing `new ClientOptionsBuilder()` with `ClientOptions.builder()` would be enough.

site/src/sphinx/client-custom-http-headers.rst
JunoJunho: As `Michael Jordan` wishes.

JunoJunho:
Is there any way to rebuild this?

My local machine built well..

trustin:
You can simply amend the last commit and force push. I sometimes use this shell script:

```bash
#!/bin/bash -e
HEAD_COMMIT_ID=$(git rev-parse HEAD)
if [[ -n "$HEAD_COMMIT_ID" ]]; then
  git commit --amend --reuse-message="$HEAD_COMMIT_ID"
  git push --force
fi
```

trustin:
Thanks, @JunoJunho! :bowing_man:

Issue #760, opened 2017-09-22
ClientRequestContext is not propagated to client callbacks


trustin:
`THttpClientInvocationHandler` and `THttpClientDelegate` do not seem to propagate `ClientRequestContext` properly to `AsyncMethodCallback` and the callbacks registered to `RpcResponse`.

Similarly, `HttpClientDelegate` doesn't seem to do so either.

`UserClient.execute(...)` pushes the client context, but that is not enough.

trustin:
`Subscriber.onSubscribe()` as well.

minwoox:
> Subscriber.onSubscribe() as well.

Let me work on this.
As @anuraaga [suggested](https://github.com/line/armeria/pull/2311#issuecomment-564434584) I might add `SubscriptionOption` later to give the user a choice not to push the context if we need.

One thing I'm not sure is that we were arguing in #2311 that we need to push the context when invoking all of `onX` method.
My initial implementation was pushing context when calling `onSubscribe`, `onComplete` and `onError`. But yes it's weird not to push `onNext`.
But I'm wondering if it's okay pushing the context only when invoking `onSubscribe`. 🤔

minwoox:
Thanks to @trustin, we can use context captor to get `ClientRequestContext` before subscribing which means that a user who wants to use the context in `onSubscribe` can use that.

So I think we can close this issue. 😄

PR #2282, opened 2019-11-26
Provide a way not to push context into thread-local when calling a lo…

minwoox:
…g listener

Motivation:
Currently, this raises an `IllegalStateException` in certain circumstances:
```java
final WebClient client =
        WebClient.builder(server.uri("/"))
                 .decorator(RetryingHttpClient...)
                 .build();

final CompletableFuture<AggregatedHttpResponse> future = client.get("/").aggregate();
serviceReqeustCtx.makeContextAware(future).handle(...);
```
It's because the `eventLoop` which executes the callback has `ClientReqeustContext` in its thread-local while calling `serviceRequestCtx.pushIfAbsent()` which raises an `IllegalStateException`.

More specifically, in `RetryingHttpClient`, we are using `RequestLog.addListener` to make a decision to retry or not. The `listener` is called with the `ClientRequestContext` in its thread's thread-local stack, so it raises an exception when pushing `ServiceRequestContext`.

To simply solve this, I chose to add `ReqeustLog.addListner` with the parameter `pushContext` which a user can determine to push the context or not.
This is not the best solution but I think we still need that API because a user might want not to push context when they have some heavy logic in `onEnter()`.

After fixing the bug, I will work on #760 and other `requestContext`-related issues to solve the problem in a general way.

Modifications:
- Add `RequestLog.addListener(...)` which a user can determine not to push context into the thread-local.

Result:
- You can now call a `requestLog` listener without having ctx in the thread-local.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
+++ b/core/src/main/java/com/linecorp/armeria/client/retry/RetryingHttpClient.java
@@ -219,7 +219,7 @@ public final class RetryingHttpClient extends RetryingClient<HttpRequest, HttpRe
                                .handle(handleBackoff(ctx, derivedCtx, rootReqDuplicator, originalReq,
                                                      returnedRes, future, response, originalResClosingTask));
             }
-        }, RequestLogAvailability.RESPONSE_HEADERS);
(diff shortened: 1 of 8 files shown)


minwoox:
Let me close this and solve it in another way. 😉

PR #2283, opened 2019-11-26
Clean up the factory methods in `Http{Request,Response}`

### LINKED ISSUE #2278: API inconsistency for creating request with new headers
Currently, we have many factories in `HttpRequest` that take headers as the first parameter

```java
HttpRequest.of(RequestHeaders, HttpData...)
HttpRequest.of(RequestHeaders headers, Publisher<? extends HttpObject> publisher)
```

However updating a request, it's with a second parameter

```java
HttpRequest of(HttpRequest request, RequestHeaders newHeaders)
```

This seems inconsistent to me (error prone's default checks wouldn't allow this either I believe there might be an effective java entry about inconsistent overrides). However, much worse than that is that both methods compile for an `HttpRequest`

```java
HttpRequest.of(newHeaders, oldRequest);  // Broken
HttpRequest.of(oldRequest, newHeaders);
```

The first results in an object that returns `newHeaders` from the `headers()` method but publishes old headers. If I'm not mistaken, this wouldn't actually work.

I propose

- Remove HttpRequest of(RequestHeaders headers, Publisher<? extends HttpObject> publisher) since it seems like a low value-add API which is very hard to reason about since the publisher may publish headers. If keeping it, it could be changed to `Publisher<HttpData>` which satisfies the current single usage in the Armeria codebase, but that leaves an issue of trailing headers. I think using `Publisher` is advanced enough that expecting users to filter it could be enough

- Remove HttpRequest of(HttpRequest request, RequestHeaders newHeaders) and add `HttpRequest.withHeaders()`. I'm guessing the reason the argument order is inconsistent is the behavior is a little different, overriding instead of creating a new request. We could sidestep the confusion by moving it into an instance method.

trustin:
Motivation:

- The following two methods look confusingly similar to each other and
  `HttpRequest` implements `Publisher`:
  - `HttpRequest.of(RequestHeaders, Publisher)`
  - `HttpRequest.of(HttpRequest, RequestHeaders)`
- Some static factory methods could be just member methods:
  - `HttpRequest.of(HttpRequest, RequestHeaders)`
  - `HttpRequest.of(AggregatedHttpRequest)`
  - `HttpResponse.of(AggregatedHttpResponse)`

Modifications:

- Deprecated `HttpRequest.of(HttpRequest, RequestHeaders)` in favor of
  `HttpRequest.of(RequestHeaders, HttpRequest)`
- Improved `HttpRequest.of(RequestHeaders, Publisher)` and
  `HttpResponse.of(Publisher)` so that they handle `HttpRequest` and
  `HttpResponse` specially.
- Added `HttpRequest.withHeaders(RequestHeaders)`
- Deprecated `HttpRequest.of(AggregatedHttpRequest)` and
  `HttpResponse.of(AggregatedHttpResponse)` in favor of
  `AggregatedHttpRequest.toHttpRequest()` and
  `AggregatedHttpResponse.toHttpResponse()`.
  - These methods were removed from `AggregatedHttpMessage` some time
    ago due to ambiguity, but now we are reviving them because we have
    separate types for requests and responses.

Result:

- Fixes #2278

### DIFF
--- a/brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
+++ b/brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
@@ -107,7 +107,7 @@ public final class BraveClient extends SimpleDecoratingHttpClient {
         final RequestHeadersBuilder newHeaders = req.headers().toBuilder();
         final HttpClientRequest request = ClientRequestContextAdapter.asHttpClientRequest(ctx, newHeaders);
         final Span span = handler.handleSend(request);
-        req = req.withHeaders(newHeaders.build());
+        req = req.withHeaders(newHeaders);
         ctx.updateRequest(req);
 
         // Ensure the trace context propagates to children
--- a/brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
+++ b/brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
@@ -107,7 +107,7 @@ public final class BraveClient extends SimpleDecoratingHttpClient {
         final RequestHeadersBuilder newHeaders = req.headers().toBuilder();
         final HttpClientRequest request = ClientRequestContextAdapter.asHttpClientRequest(ctx, newHeaders);
         final Span span = handler.handleSend(request);
-        req = HttpRequest.of(req, newHeaders.build());
+        req = req.withHeaders(newHeaders.build());
         ctx.updateRequest(req);
 
         // Ensure the trace context propagates to children
--- a/core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
+++ b/core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
@@ -26,6 +26,8 @@ import java.util.Locale;
 
 import com.google.common.collect.ImmutableList;
 
+import com.linecorp.armeria.common.FixedHttpResponse.RegularFixedHttpResponse;
+
 /**
  * A complete HTTP response whose content is readily available as a single {@link HttpData}.
  */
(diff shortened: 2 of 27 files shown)


brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -107,7 +107,7 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
         final RequestHeadersBuilder newHeaders = req.headers().toBuilder();
         final HttpClientRequest request = ClientRequestContextAdapter.asHttpClientRequest(ctx, newHeaders);
         final Span span = handler.handleSend(request);
-        req = HttpRequest.of(req, newHeaders.build());
+        req = req.withHeaders(newHeaders.build());
```
anuraaga: We can do it in separate PR but noticed that it'd be reasonable to add `withHeaders(RequestHeadersBuilder h) { withHeaders(h.build()); }` - inspired by protobuf which generally accepts both built and builders in its methods I think

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -227,4 +229,38 @@ static AggregatedHttpResponse of(Iterable<ResponseHeaders> informationals, Respo
      * Returns the {@linkplain HttpHeaderNames#STATUS STATUS} of this response.
      */
     HttpStatus status();
+
+    /**
+     * Converts this request into a new complete {@link HttpResponse}.
```
minwoox: nit: request -> response

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -227,4 +229,38 @@ static AggregatedHttpResponse of(Iterable<ResponseHeaders> informationals, Respo
...
+     *
+     * @return the new {@link HttpResponse} converted from this request.
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
```diff
@@ -279,24 +292,48 @@ static HttpRequest of(RequestHeaders headers, Publisher<? extends HttpObject> pu
      * >     }
      * > }
      * }</pre>
+     *
+     * @see #withHeaders(RequestHeaders)
      */
-    static HttpRequest of(HttpRequest request, RequestHeaders newHeaders) {
+    static HttpRequest of(RequestHeaders newHeaders, HttpRequest request) {
```
minwoox: Can we just remove this method and make users to use `withHeaders` directly?

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
trustin: That's a good idea. Added.

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
trustin: Oops :sweat_smile:

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
trustin: Hmm. Maybe that's OK.

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
```diff
@@ -350,6 +353,74 @@ default MediaType contentType() {
...
+            // Just check the reference only to avoid heavy comparison.
+            return this;
+        }
+
+        return new HeaderOverridingHttpRequest(this, newHeaders);
+    }
+
+    /**
+     * Returns a new {@link HttpRequest} derived from this {@link HttpRequest} by replacing its
+     * {@link RequestHeaders} with what's built from the specified {@code newHeaders}. Note that the content
```
minwoox: nit: newHeadersBuilder

trustin:
Thanks, reviewers!

PR #2284, opened 2019-11-29
Respect gRPC stub deadline.

### LINKED ISSUE #2008: `NullPointerException` in `ArmeriaGrpcServerInteropTest`

anuraaga:
This adds support for the `deadline` option of a gRPC client stub. Now, if this is provided, for example by one of the gRPC integration tests, it is handled correctly at the client side - previously it only sort of worked because it would often be handled at the server side.

I also added an option, not available in usptream, to ignore a client-specified timeout in the server. I've always felt it unsafe that gRPC server uses a client header to determine timeout, and this flag can be used for situations without trusted clients (and unit tests :) ).

Fixes #2008

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
+++ b/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
@@ -212,7 +212,7 @@ public final class GrpcService extends AbstractHttpService implements HttpServic
         final ArmeriaServerCall<?, ?> call = startCall(
                 methodName, method, ctx, req.headers(), res, serializationFormat);
         if (call != null) {
-            ctx.setRequestTimeoutHandler(() -> call.close(Status.DEADLINE_EXCEEDED, new Metadata()));
+            ctx.setRequestTimeoutHandler(() -> call.close(Status.CANCELLED, new Metadata()));
             req.subscribe(call.messageReader(), ctx.eventLoop(), WITH_POOLED_OBJECTS);
             req.completionFuture().handleAsync(call.messageReader(), ctx.eventLoop());
         }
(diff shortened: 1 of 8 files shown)


anuraaga:
Forgot to update the server to match upstream logic. Added a commit to do so, based on https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L666

grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
```diff
@@ -188,15 +191,17 @@ protected HttpResponse doPost(ServiceRequestContext ctx, HttpRequest req) throws
...
+        if (useClientTimeoutHeader) {
+            final String timeoutHeader = req.headers().get(GrpcHeaderNames.GRPC_TIMEOUT);
+            if (timeoutHeader != null) {
+                try {
+                    final long timeout = TimeoutHeaderUtil.fromHeaderValue(timeoutHeader);
+                    ctx.setRequestTimeout(Duration.ofNanos(timeout));
+                } catch (IllegalArgumentException e) {
+                    return HttpResponse.of(
+                            ArmeriaServerCall.statusToTrailers(
+                                    ctx, GrpcStatus.fromThrowable(e), new Metadata(), false));
```
minwoox: nit: This calls `static HttpResponse of(HttpObject... objs) {...}` instead of `static HttpResponse of(ResponseHeaders headers) {...}`. How about type-casting?
```java
return HttpResponse.of(
        (ResponseHeaders) ArmeriaServerCall.statusToTrailers(
                ctx, GrpcStatus.fromThrowable(e), new Metadata(), false));
```

grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
```diff
@@ -188,15 +191,17 @@ protected HttpResponse doPost(ServiceRequestContext ctx, HttpRequest req) throws
                             false));
```
minwoox: This one as well.

grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcService.java
anuraaga: Sure added the cast

PR #2286, opened 2019-12-01
Add immutable HTTP cookie API

trustin:
Motivation:

We currently rely on Netty's HTTP cookie API, but it has the following
drawbacks:

- It is mutable.
- It will not have `SameSite` attribute until Netty 5.
  - https://github.com/netty/netty/issues/8161
- It will not allow cookies with the same name until Netty 5.
  - https://github.com/netty/netty/issues/7210

Modifications:

- Add `Cookie`, which is an immutable representation of HTTP Cookie.
- Add `CookieBuilder` which builds a `Cookie`.
- Fork `{Client,Server}Cookie{Encoder,Decoder}` from Netty so it is
  possible to encode and decode our `Cookie`.
  - Modify `ServerCookieDecoder` to allow cookies with the same name.

Result:

- Immutable HTTP cookie API
- (Deprecation) `Cookies.copyOf()` has been deprecated in favor of `Cookies.of()`.
- (Breaking) `Cookies` now use Armeria's own `Cookie`.
- (Breaking?) `Cookies` may contain the cookies with the same name.
  - RFC does not prohibit duplicate names at all, so this is a correct behavior.
- (Breaking?) Annotated services now accept only the first `"Cookie"` header when decoding cookies in an `HttpRequest`. Sending multiple `"Cookie"` headers was a violation of RFC 6265 anyway.


core/src/main/java/com/linecorp/armeria/common/ClientCookieDecoder.java
```diff
@@ -0,0 +1,289 @@
...
+ * A <a href="http://tools.ietf.org/html/rfc6265">RFC6265</a> compliant cookie decoder for client side.
+ *
+ * <p>It will store the way the raw value was wrapped in {@link Cookie#isValueQuoted()} so it can be sent back
+ * to the origin server as is.</p>
+ *
+ * @see ClientCookieEncoder
+ */
+public final class ClientCookieDecoder extends CookieDecoder {
+
+    // Forked from netty-4.1.43
```
anuraaga: Let's add permalink too

core/src/main/java/com/linecorp/armeria/common/ClientCookieEncoder.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /**
+     * Encodes the specified cookie into a {@code "Cookie"} header value.
+     *
+     * @param name the cookie name
+     * @param value the cookie value
+     * @return a Rfc6265 style Cookie header value
+     */
+    public String encode(String name, String value) {
```
anuraaga: From what I can tell, `ClientCookieEncoder` and `ServerCookieEncoder` expose the same API. Does it make sense to instead only expose `CookieEncoder` interface with `clientStrict()`, `clientLax()`, etc factories and the public methods?

core/src/main/java/com/linecorp/armeria/common/ServerCookieEncoder.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Encodes the specified {@link Cookie}s into the {@code "Set-Cookie"} header values.
+     *
+     * @param cookies the {@link Cookie}s to encode.
+     * @return the corresponding {@code "Set-Cookie"} header values.
+     */
+    public List<String> encode(Iterable<? extends Cookie> cookies) {
+        final Iterator<? extends Cookie> cookiesIt = requireNonNull(cookies, "cookies").iterator();
```
anuraaga: It's probably worth optimizing the `Collection` case as it should be much more common than an unsized `Iterable`, allowing presizing the returned list. We could have two paths, though maybe it's overkill and better to keep the code simple and either use `ImmutableList.copyOf` or `Streams.stream(cookies).map(this::encode).collect(toImmutableList())`, both should presize the returned list.

core/src/main/java/com/linecorp/armeria/common/ClientCookieEncoder.java
trustin: They are expected to generate the header string(s) which adheres to different syntax, so I'm not sure if it's a good idea to expose a common interface between them.

What do you think about just adding a bunch of static methods to `Cookie` and then hide all encoders and decoders?

```java
    static String toCookieHeader(boolean strict, Cookie... cookies);
    static String toCookieHeader(boolean strict, Iterable<? extends Cookie> cookies);

    // Note that multiple `Cookie` headers are prohibited by RFC 6265
    // and thus we do not accept multiple headers.
    static Cookies fromCookieHeader(boolean strict, String cookieHeader);

    static List<String> toSetCookieHeaders(boolean strict, Cookie... cookies);
    static List<String> toSetCookieHeaders(boolean strict, Iterable<? extends Cookie> cookies);

    static Cookies fromSetCookieHeaders(boolean strict, String... setCookieHeaders);
    static Cookies fromSetCookieHeaders(boolean strict, Iterable<String> setCookieHeaders);

    // member methods
    String toCookieHeader(boolean strict);
    String toSetCookieHeader(boolean strict);
```

trustin: Went ahead and:

- Added the static methods and member methods mentioned above.
  - Also added the versions without `boolean strict` which default to strict mode.
- Hid `{Client,Server}Cookie{Encoder,Decoder}` from the public API.

core/src/main/java/com/linecorp/armeria/common/ServerCookieEncoder.java
trustin: Done!

core/src/main/java/com/linecorp/armeria/common/CookieUtil.java
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        return Cookie.builder(name, unwrappedValue.toString()).valueQuoted(valueQuoted);
+    }
+
+    // The methods newly added in the fork.
+
+    static Cookies fromSetCookieHeaders(ImmutableSet.Builder<Cookie> builder,
+                                        boolean strict, Iterator<String> it) {
+        do {
```
anuraaga: I think this do / while is to save a single call to `hasNext`, whereas usually we're not sure if the `it` is empty or not and use a normal while. I'm not sure this optimization is enough to not use a normal iteration style though. If we really wanted to keep it, I'd probably still add `assert it.hasNext()` at the beginning to make clear why we're using do/while here.

core/src/main/java/com/linecorp/armeria/common/Cookies.java
```diff
@@ -13,32 +13,57 @@
...
+        return DefaultCookies.EMPTY;
+    }
+
     /**
      * Creates an instance with a copy of the specified set of {@link Cookie}s.
      */
     static Cookies of(Cookie... cookies) {
-        return new DefaultCookies(ImmutableSet.copyOf(requireNonNull(cookies, "cookies")));
+        requireNonNull(cookies, "cookies");
+        if (cookies.length == 0) {
```
anuraaga: This is fine too but we could do `return of(ImmutableSet.copyOf(cookies)` for less code, and no change in allocations since `copyOf` of empty array returns empty set, and `copyOf` of `ImmutableSet` returns the same instance.

trustin: Sure. No problem.

core/src/main/java/com/linecorp/armeria/common/CookieUtil.java
trustin: Good idea. Added an assertion.

core/src/main/java/com/linecorp/armeria/common/Cookies.java
```diff
@@ -13,32 +13,52 @@
...
 
+    /**
+     * Returns an empty {@link Set} if {@link Cookie}s.
+     */
+    static Cookies empty() {
+        return DefaultCookies.EMPTY;
+    }
+
     /**
      * Creates an instance with a copy of the specified set of {@link Cookie}s.
```
anuraaga: I think either this could be `array of {@link Cookie}s` for consistency with the other javadoc (this parameter isn't a set), or even better is probably to use `specified set of {@link Cookie}s` in the other javadoc same as here since semantically they're the same method.

trustin: Changed to 'specified set of Cookies'

trustin:
(Breaking?) Annotated services now accept only the first "Cookie" header when decoding cookies in an HttpRequest. Sending multiple "Cookie" headers was a violation of RFC 6265 anyway.

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+    public CookieBuilder valueQuoted(boolean valueQuoted) {
+        this.valueQuoted = valueQuoted;
+        return this;
+    }
+
+    /**
+     * Sets the domain of the {@link Cookie}.
+     */
+    public CookieBuilder domain(String domain) {
+        this.domain = validateAttributeValue("domain", requireNonNull(domain, "domain"));
```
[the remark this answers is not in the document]
trustin: Actually, `validateAttributeValue()` was calling `requireNonNull()` already. Removed all `requireNonNull()`s.

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+     * Sets the security status of the {@link Cookie}. If unspecified, {@code false} will be used.
+     */
+    public CookieBuilder secure(boolean secure) {
+        this.secure = secure;
+        return this;
+    }
+
+    /**
+     * Sets whether the {@link Cookie} is HTTP only. If {@code true}, the {@link Cookie} cannot be accessed
+     * by a client side script. However, this works only if the browser supports it. For for information,
```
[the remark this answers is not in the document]
trustin: :sweat_smile:

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+        for (int i = 0; i < cs.length(); i++) {
+            final char c = cs.charAt(i);
+            if (!bits.get(c)) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private static String validateAttributeValue(String name, @Nullable String value) {
```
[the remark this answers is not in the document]
trustin: It's now nullable.

core/src/main/java/com/linecorp/armeria/common/Cookies.java
```diff
@@ -13,32 +13,52 @@
...
-import io.netty.handler.codec.http.cookie.Cookie;
-
 /**
- * An interface which holds decoded {@link Cookie} instances for an HTTP request.
+ * An immutable {@link Set} of {@link Cookie}s.
  */
 public interface Cookies extends Set<Cookie> {
 
+    /**
+     * Returns an empty {@link Set} if {@link Cookie}s.
```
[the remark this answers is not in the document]
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+    private static BitSet validCookieAttributeValueOctets() {
+        final BitSet bits = new BitSet();
+        for (int i = 32; i < 127; i++) {
+            bits.set(i);
+        }
+        bits.set(';', false);
+        return bits;
+    }
+
+    private static int firstInvalidOctet(CharSequence cs, BitSet bits) {
```
minwoox: I saw the same method in `CookieUtil`. Can we reuse it?

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+        return -1;
+    }
+
+    private static String validateAttributeValue(String value, String valueName) {
+        value = requireNonNull(value, valueName).trim();
+        checkArgument(!value.isEmpty(), "%s is empty.", valueName);
+        final int i = firstInvalidOctet(value, VALID_COOKIE_ATTRIBUTE_VALUE_OCTETS);
+        checkArgument(i == -1,
+                      "%s contains a prohibited character: %s", valueName, value);
```
minwoox: How about showing the invalid character as we do in `CookieUtil`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaServerHttpResponse.java
```diff
@@ -283,20 +282,23 @@ public String toString() {
...
-            cookie.setPath(resCookie.getPath());
+            builder.path(resCookie.getPath());
         }
-        cookie.setSecure(resCookie.isSecure());
-        cookie.setHttpOnly(resCookie.isHttpOnly());
-        return cookie;
+        builder.secure(resCookie.isSecure());
+        builder.httpOnly(resCookie.isHttpOnly());
+        if (resCookie.getSameSite() != null) {
+            builder.sameSite(resCookie.getSameSite());
```
ikhoon: 👍

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedValueResolver.java
```diff
@@ -566,15 +564,11 @@ private static AnnotatedValueResolver ofInjectableTypes0(AnnotatedElement annota
...
+                        final String value = ctx.request().headers().get(HttpHeaderNames.COOKIE);
+                        if (value == null) {
+                            return Cookies.empty();
                         }
-                        final ImmutableSet.Builder<Cookie> cookies = ImmutableSet.builder();
-                        values.stream()
-                              .map(ServerCookieDecoder.STRICT::decode)
-                              .forEach(cookies::addAll);
-                        return Cookies.copyOf(cookies.build());
+                        return Cookie.fromCookieHeader(value);
```
ikhoon: 👍

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+        return this;
+    }
+
+    /**
+     * Sets the <a href="https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-4.1.2.7"
+     * >{@code SameSite}</a> attribute of the {@link Cookie}. The value is supposed to be one of {@code "Lax"},
+     * {@code "Strict"} or {@code "None"}. Note that this attribute is server-side only.
+     */
+    public CookieBuilder sameSite(String sameSite) {
+        this.sameSite = validateAttributeValue(sameSite, "sameSite");
```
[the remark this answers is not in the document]
trustin: It's intentionally left as is because web is moving fast, adding more values.

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+    int expiresStart;
+    int expiresEnd;
+
+    CookieBuilder(String name, String value) {
+        this.name = requireNonNull(name, "name");
+        this.value = requireNonNull(value, "value");
+    }
+
+    CookieBuilder(Cookie cookie) {
+        name = cookie.name();
```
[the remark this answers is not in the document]
trustin: It's always called with `this`, so a user will never see a NPE.

core/src/main/java/com/linecorp/armeria/common/Cookie.java
```diff
@@ -0,0 +1,590 @@
...
+     * @param strict whether to validate that cookie names and values are in the valid scope
+     *               defined in RFC 6265 and to sort the {@link Cookie}s into order of decreasing path length,
+     *               as specified in RFC 6265. If {@code false}, the {@link Cookie}s are encoded in the order
+     *               in which they are given.
+     * @param cookies the {@link Cookie}s to encode.
+     * @return the encoded {@code "Cookie"} header value.
+     */
+    static String toCookieHeader(boolean strict, Cookie... cookies) {
+        requireNonNull(cookies, "cookies");
+        checkArgument(cookies.length != 0, "cookies is empty.");
```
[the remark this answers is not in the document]
trustin: cookies here is an array parameter, so it's `cookies` that's empty, not that individual cookies are empty.

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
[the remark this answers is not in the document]
ikhoon: web is moving fast, adding more values.
Ahh, got it. I totally agree that :-)

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/CookieBuilder.java
trustin: Fixed

ikhoon:
reference: https://github.com/line/armeria/issues/1567

PR #2288, opened 2019-12-02
Make sure boss `EventLoopGroup`s are terminated on startup fail…

trustin:
Related: https://stackoverflow.com/q/58891320/55808
Motivation:

When a server fails to start up, its boss `EventLoopGroup` is not
terminated, leaving a dangling non-daemon thread which prevents a JVM
process from terminating itself.

Modifications:

- Make sure the boss `EventLoopGroup`s are always terminated by keeping
  the boss groups created by `Server`.

Result:

- No more dangling non-daemon threads on server startup failure

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Server.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Server.java
@@ -113,7 +113,6 @@ public final class Server implements AutoCloseable {
     @VisibleForTesting
     ServerBootstrap serverBootstrap;
 
-
     Server(ServerConfig config, @Nullable DomainNameMapping<SslContext> sslContexts) {
         this.config = requireNonNull(config, "config");
         this.sslContexts = sslContexts;
--- a/core/src/main/java/com/linecorp/armeria/server/Server.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Server.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.server;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static java.util.Objects.requireNonNull;
 
 import java.net.InetAddress;
@@ -30,7 +31,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -43,6 +43,7 @@ import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 
+import org.jctools.maps.NonBlockingHashSet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -633,14 +637,13 @@ private void logStopFailure(Throwable cause) {
 
         @Override
         public void operationComplete(ChannelFuture f) {
-            final Channel ch = f.channel();
+            final ServerChannel ch = (ServerChannel) f.channel();
             assert ch.eventLoop().inEventLoop();
 
-            if (f.isSuccess()) {
```
anuraaga: Is the logic change this one? The others appear to be no-op to me. We used to close using a set of channels and calling `ch.eventLoop().parent ()` but now store that as keys of a map. If I'm not mistaken in it being the same as before, I might consider the previous version with the `Set` to be easier to reason about.

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -502,16 +501,19 @@ private void doStop(CompletableFuture<Void> future,
...
                             finishDoStop(future);
                             return;
                         }
+
                         // Shut down all boss groups and wait until they are terminated.
-                        final AtomicInteger remainingBossGroups = new AtomicInteger(serverChannels.size());
-                        serverChannels.forEach(ch -> {
-                            final EventLoopGroup bossGroup = ch.eventLoop().parent();
+                        final AtomicInteger remainingBossGroups = new AtomicInteger(bossGroups.size());
+                        bossGroups.forEach(bossGroup -> {
```
anuraaga: I think this `bossGroup` is shadowing another one so maybe a different name is better.

trustin: Hmm? What does it shadow?

ikhoon: I couldn't find shadow too. Maybe GitHub diff fold lots of lines, and the 405 line makes you confused 😉

anuraaga: Oops yeah that was it

core/src/main/java/com/linecorp/armeria/server/Server.java
trustin: Good point. Made `serverChannels` a `Set<ServerChannel>`.

core/src/test/java/com/linecorp/armeria/server/ServerTest.java
```diff
@@ -194,8 +200,43 @@ public void testDelayedResponseApiInvocationExpectedTimeout() throws Exception {
...
-        assertThat(server.server().serverBootstrap().config()
-                         .options().get(ChannelOption.SO_BACKLOG)).isEqualTo(1024);
+        final ServerBootstrap bootstrap = server.server().serverBootstrap;
+        assertThat(bootstrap).isNotNull();
+        assertThat(bootstrap.config().options().get(ChannelOption.SO_BACKLOG)).isEqualTo(1024);
+    }
+
+    @Test
+    public void unsuccessfulStartupTerminatesBossGroup() {
+        final Predicate<ThreadInfo>  predicate = info -> {
```
anuraaga:
```suggestion
        final Predicate<ThreadInfo> predicate = info -> {
```

trustin: Thanks :laughing:

trustin:
Thanks for reviewing!

trustin:
Special thanks to @magnuschatt for reporting this issue via StackOverflow - https://stackoverflow.com/questions/58891320/how-to-make-armeria-exit-on-an-address-already-in-use-error

PR #2289, opened 2019-12-02
Deprecate `Sampler.rateLimited()` in favor of `rateLimiting()`

### LINKED ISSUE #2122: Naming of `Sampler.rateLimited(int)`
I'm curious which is most natural:

- `Sampler.rateLimited(int samplesPerSecond)` (Current form)
- `Sampler.rateLimit(int samplesPerSecond)`
- `Sampler.rateLimiting(int samplesPerSecond)`

We also use `rate-limited` in specification strings, but I wonder which is most natural:

- `rate-limited=10` (Current form)
- `rate-limit=10`
- `rate-limiting=10`

I think the specification string should follow the factory method name for consistency, though.

trustin:
Motivation:

Modifications:

- Add `Sampler.rateLimiting()`.
- Deprecate `Sampler.rateLimited()`.
- Make `Sampler.of(String)` accept `rate-limiting=n` and `rate-limit=n`
  as well as `rate-limited=n`.

Result:

- Fixes #2122
- (Deprecation) `Sampler.rateLimites()` has been deprecated in favor of
  `rateLimiting()`.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/Flags.java
+++ b/core/src/main/java/com/linecorp/armeria/common/Flags.java
@@ -77,7 +77,7 @@ public final class Flags {
 
     private static final int NUM_CPU_CORES = Runtime.getRuntime().availableProcessors();
 
-    private static final String DEFAULT_VERBOSE_EXCEPTION_SAMPLER_SPEC = "rate-limited=10";
+    private static final String DEFAULT_VERBOSE_EXCEPTION_SAMPLER_SPEC = "rate-limit=10";
     private static final String VERBOSE_EXCEPTION_SAMPLER_SPEC;
     private static final Sampler<Class<? extends Throwable>> VERBOSE_EXCEPTION_SAMPLER;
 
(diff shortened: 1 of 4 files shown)


ikhoon:
Shouldn't we update the following documentation and message with `rate-limit`?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -77,7 +77,7 @@
 
     private static final int NUM_CPU_CORES = Runtime.getRuntime().availableProcessors();
 
-    private static final String DEFAULT_VERBOSE_EXCEPTION_SAMPLER_SPEC = "rate-limited=10";
+    private static final String DEFAULT_VERBOSE_EXCEPTION_SAMPLER_SPEC = "rate-limit=10";
```
minwoox: rate-limiting?

trustin: I think `rate-limit` is better than `rate-limiting` when used as a part of spec string since limit is a noun anyway.

trustin:
Thanks for reviewing.

PR #2290, opened 2019-12-02
Add GitHub actions workflow

anuraaga:
This is a strawman's proposal to add a GitHub actions-based build for Armeria. I have been playing with it, mostly to get used to the syntax. I wasn't so interested in merging it in without finding a significant advantage to the current build - then we had the breakage on MacOS X :) My CI runs were broken on Mac at the time but since I was experimenting with the scripts still, I assumed it was a problem with the CI, not the code. It made me think that it'd be good to have continuous build running on Mac - GitHub Actions makes it easy to run on Linux, Windows, and Mac.

Pros:
- Adds MacOS X build
- Fairly simple YAML syntax. In particular, support for caching removes a lot of boilerplate we have related to it
- Builds viewable directly in GitHub UI. In particular, it's nice that artifacts are available from the UI instead of finding the file.io link.
- No tricks to use cache, all OS's support it out of the box

Cons:
- Slower machines, 2CPU vs I think we currently use 4CPU ones. The number of parallel builds is generous, so with many of them we could claim lots of resources, but in general I guess builds are slower. This build with cache hits took ~25M. Most time is spent on the coverage build. https://github.com/anuraaga/armeria/commit/6a9a9e76d79a1aca3a32397bedf46c00b765766d/checks?check_suite_id=337250391
- Vastly slower Windows machine. Everyone seems to be finding the Windows machine to be much slower than Appveyor's. This doesn't hurt us so much in terms of build time since the reduced build commands means it still finishes earlier than others. But many more test flakes become apparent due to timing issues at the seconds-level. Admittedly, this could be considered an issue with the tests and not the CI.
- No great support for codecov right now (https://github.com/codecov/codecov-action/issues/29). That being said, I'm not sure there's a huge danger to embedding `CODECOV_TOKEN` as a non-secret.

Options:
- Only use GitHub Actions for MacOS X build. We avoid the slow machines issue but pay the price of maintaining two different build configs. We used to do it for Windows so maybe not a big deal.
- Add the full GitHub Actions workflow without removing AppVeyor. Since we don't pay for resources, it doesn't really hurt to keep it in action (pun intended) and allows monitoring it's build performance. But duplicate builds can confuse users
- Remove Appveyor and only use GitHub Actions. We get all the cons above, and would need to expose a codecov token publically (or not have coverage on pull requests which is probably worse)
- Don't use GitHub Actions. Actually since I started on this, it seems that Appveyor supports Mac OS X too now on request, so maybe don't need GitHub Actions. https://www.appveyor.com/blog/2019/11/20/build-macos-projects-with-appveyor/

Note :the GitHub Actions build will mostly fail until #2284

### DIFF
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -1,0 +1,115 @@
+name: Continuous Build
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    env:
+      CI: true
+    strategy:
+      matrix:
+        os: 
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+        jre: [13, 11, 8]
+        include:
+          - os: macos-latest
+            gradle_build_args: -PnoCheckstyle -PnoWeb :core:test :grpc:test :thrift:test :it:server:test
+          - os: windows-latest
+            gradle_build_args: -PnoCheckstyle -PnoWeb :core:test :grpc:test :thrift:test :it:server:test
+          - os: ubuntu-latest
+            gradle_build_args: -PnoCheckstyle -PnoWeb build
+          - os: ubuntu-latest
+            jre: 11
+            gradle_build_args: -Pcoverage checkstyle build
+        exclude: 
+          # We only test multiple JVM types on Linux to conserve resources. Unfortunately we cannot add jobs through include and
+          # must use many excludes instead.
+          - os: macos-latest
+            jre: 8
+          - os: macos-latest
+            jre: 13
+          - os: windows-latest
+            jre: 8
+          - os: windows-latest
+            jre: 13
+    steps:
+    - uses: actions/checkout@v1
+
+    # We can't reference env.HOME in jobs.*.env so we set variables that use it here.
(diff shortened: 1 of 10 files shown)


anuraaga:
Actually just noticed, it seems since LINE is a paid organization, actions don't work even though the repo is open source... I suspect it should be allowed though so maybe need to ask support if going through with the idea.

image

syleeeee:
> it seems since LINE is a paid organization, actions don't work even though the repo is open source

Yes, LINE is a paid org but stays in the old plan. 
GitHub Actions can be used in the user-based-plan(which is the new plan). We are in the middle of the discussion to move forward to use the new plan. But cannot confirm when. Sorry. 😭 

I'll let you know when we change the plan.

anuraaga:
@syleeeee Thanks for clarifying, I didn't know they had such a plan migration. Wish they didn't show the tab in that case ;)

Guess not currently supported is the biggest con in that list then 😭

anuraaga:
As it'll take some time to even be able to try this will go ahead and close - would be good to get feedback on whether this seems like an ok idea in general though.

trustin:
We're definitely interested and would be awesome if we can automate most of the release process using GitHub actions. :+1:

anuraaga:
Thanks for the comment - in that case look forwarding to bringing this back soon :)

anuraaga:
Just to record so not forgotten, I forgot one great pro of anything non-Appveyor is much easier to retry flaky builds.

PR #2292, opened 2019-12-02
Automatic redirection with a trailing slash

### LINKED ISSUE #2116: `prefix` path mapping is overly strict.
For example, if a user bound a service at `prefix:/docs`, the service
will get requests only when the request path starts with `/docs/`, which
means `/docs` will not be matched.

This is not a problem when there is no other mappings that match `/docs`
because Armeria will automatically redirect you to `/docs/`.

However, the following code will not work:

```java
// `/docs` will not be redirected to `/docs/` but handled by
// `HttpFileService`.
Server server =
    Server.builder()
          .service("prefix:/docs", new DocService())
          .service("prefix:/", HttpFileService.of(...))
          .build();
```

... because the `HttpFileService` will handle the request to `/docs`.

To address this issue, we need to introduce some special casing for the
path without a trailing slash.

However, the redirection should not occur when there's an explicit binding
for the path without a trailing slash:

```java
// `/docs` will not be redirected to `/docs/` but handled by
// `someOtherService`.
Server server =
    Server.builder()
          .service("prefix:/docs", new DocService())
          .service("/docs", someOtherService)
          .build();
```

trustin:
Motivation:

For example, if a user bound a service at `prefix:/docs`, the service
will get requests only when the request path starts with `/docs/`, which
means `/docs` will not be matched.

This is not a problem when there is no other mappings that match `/docs`
because Armeria will automatically redirect you to `/docs/`.

However, the following code will not work:

    // `/docs` will not be redirected to `/docs/` but handled by
    // `HttpFileService`.
    Server server =
        Server.builder()
              .service("prefix:/docs", new DocService())
              .service("prefix:/", HttpFileService.of(...))
              .build();

... because the `HttpFileService` will handle the request to `/docs`.

To address this issue, we need to introduce some special casing for the
path without a trailing slash.

However, the redirection should not occur when there's an explicit binding
for the path without a trailing slash:

    // `/docs` will not be redirected to `/docs/` but handled by
    // `someOtherService`.
    Server server =
        Server.builder()
              .service("prefix:/docs", new DocService())
              .service("/docs", someOtherService)
              .build();

Modifications:

- Modified `Routers.router()` to add the redirect mappings that maps to
  the fallback service for `EXACT`, `PREFIX` and `PARAMETERIZED` path
  types.
  - `FallbackService.handleNotFound()` will send a redirect response.
- Modified `RoutingTrie` to have precedence (high or low) for values.
  - Redirect mappings will have low precedence while other will have
    high precedence, so that a user-specified mapping has precedence.
- Miscellaneous:
  - `Router.findAll()` now returns `List` instead of `Stream`.
  - Tiny optimization in `CompositeRouter.findAll()`

Result:

- Fixes #2116
- Known issues:
  - `GLOB`, `REGEX` and `REGEX_WITH_PREFIX` path types are unsupported,
    although this is expected due to complexity and ambiguity.
  - (Breaking) This may break some servers which relied on the routing
    behavior.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Routers.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Routers.java
@@ -252,10 +252,10 @@ public final class Routers {
         if (logger.isDebugEnabled()) {
             logger.debug("Router created for {} service(s): {}",
                          values.size(), router.getClass().getSimpleName());
-            values.forEach(c -> {
-                final Route route = routeResolver.apply(c);
+            for (V v : values) {
+                final Route route = routeResolver.apply(v);
                 logger.debug("meterTag: {}, complexity: {}", route.meterTag(), route.complexity());
-            });
+            }
         }
         values.clear();
         return router;
--- a/core/src/main/java/com/linecorp/armeria/server/Routers.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Routers.java
@@ -226,37 +226,23 @@ public final class Routers {
         final Router<V> router;
         if (isTrie) {
             final RoutingTrieBuilder<V> builder = new RoutingTrieBuilder<>();
-            // Set a comparator to sort services by the number of conditions to be checked in a descending
-            // order.
+            // Set a comparator to sort services by the number of conditions to check in a descending order.
             builder.comparator(valueComparator);
-            values.forEach(v -> {
+            for (V v : values) {
                 final Route route = routeResolver.apply(v);
-                final String path = route.paths().get(0);
-                final int pathLen = path.length();
-                if (fallbackValue != null && pathLen > 1) {
-                    // Add some extra routes to handle the case where a client sends a request
-                    // without a trailing slash. `PathMapping.apply()` will produce a `RoutingResult`
-                    // whose type is `NEEDS_REDIRECT`.
-                    switch (route.pathType()) {
-                        case EXACT:         // `/foo`   or `/foo/`
-                        case PARAMETERIZED: // `/foo/:` or `/foo/:/`
-                            if (path.charAt(pathLen - 1) == '/') {
-                                // Add an extra route without a trailing slash for a redirect.
-                                builder.add(path.substring(0, pathLen - 1),
-                                            fallbackValue, /* hasHighPrecedence */ false);
-                            }
-                            break;
-                        case PREFIX: // `/foo/*`
-                            // Add an extra route of an exact match.
-                            builder.add(path.substring(0, pathLen - 1),
-                                        fallbackValue, /* hasHighPrecedence */ false);
-                            break;
-                        default:
-                            throw new Error("Unexpected path type: " + route.pathType());
+                builder.add(route.paths().get(1), v);
+
+                if (fallbackValue != null) {
+                    // Add an extra route without a trailing slash for a redirect.
+                    // Note that `path.length()` must be greater than 1 because path is `/` when 1.
+                    final String path = route.paths().get(0);
+                    final int pathLen = path.length();
+                    if (pathLen > 1 && path.charAt(pathLen - 1) == '/') {
+                        builder.add(path.substring(0, pathLen - 1),
+                                    fallbackValue, /* hasHighPrecedence */ false);
                     }
                 }
-                builder.add(route.paths().get(1), v);
-            });
+            }
             router = new TrieRouter<>(builder.build(), routeResolver);
         } else {
             values.sort(valueComparator);
--- a/core/src/main/java/com/linecorp/armeria/server/Routers.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Routers.java
@@ -32,7 +32,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
-import java.util.stream.Stream;
 
 import javax.annotation.Nullable;
 
@@ -42,7 +41,6 @@ import org.slf4j.LoggerFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 
-import com.linecorp.armeria.server.RoutingTrie.Builder;
 import com.linecorp.armeria.server.composition.CompositeServiceEntry;
 
 /**
@@ -71,7 +69,8 @@ public final class Routers {
             }
         };
 
-        return wrapVirtualHostRouter(defaultRouter(configs, ServiceConfig::route, rejectionConsumer));
+        return wrapVirtualHostRouter(defaultRouter(configs, virtualHost.fallbackServiceConfig(),
+                                                   ServiceConfig::route, rejectionConsumer));
     }
 
     /**
@@ -82,7 +81,7 @@ public final class Routers {
         requireNonNull(entries, "entries");
 
         final Router<CompositeServiceEntry<T>> delegate = wrapCompositeServiceRouter(defaultRouter(
-                entries, CompositeServiceEntry::route,
+                entries, /* fallbackValue */ null, CompositeServiceEntry::route,
                 (mapping, existingMapping) -> {
                     final String a = mapping.toString();
                     final String b = existingMapping.toString();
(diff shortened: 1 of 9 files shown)


trustin:
@anuraaga and @hyangtack might be interested in reviewing.

core/src/main/java/com/linecorp/armeria/server/Routers.java
```diff
@@ -227,7 +230,34 @@
             // Set a comparator to sort services by the number of conditions to be checked in a descending
             // order.
             builder.comparator(valueComparator);
-            values.forEach(v -> builder.add(routeResolver.apply(v).paths().get(1), v));
+            values.forEach(v -> {
+                final Route route = routeResolver.apply(v);
+                final String triePath = route.paths().get(1);
```
minwoox: How about using the first path so we can simplify the switch logic?`
```java
route.paths().get(0); // This will be "/foo/" when the type is PREFIX.
```

core/src/main/java/com/linecorp/armeria/server/Routers.java
```diff
@@ -295,9 +325,9 @@
         return result;
     }
 
-    private static <V> Stream<Routed<V>> findAll(RoutingContext routingCtx, List<V> values,
+    private static <V> List<Routed<V>> findAll(RoutingContext routingCtx, List<V> values,
                                                  Function<V, Route> routeResolver) {
```
hyangtack: nit: alignment

core/src/main/java/com/linecorp/armeria/server/RoutingTrie.java
```diff
@@ -261,8 +259,19 @@ private void dump(PrintWriter p, Node<V> node, int depth) {
...
         Builder<V> add(String path, V value) {
+            return add(path, value, true);
+        }
+
+        /**
+         * Adds a path and a value to be built as {@link RoutingTrie}.
+         *
+         * @param path  the path to serve
+         * @param value the value belonging to the path
+         * @param hasHighPrecedence whether the path being added has high precedence or not
```
anuraaga: Would be good to document what the precedence does. I'm not totally sure but is it `Any values that match a path with high precedence will be returned in front of matches with low precedence. Within the same precedence, values are returned in the order the paths were defined.`

core/src/main/java/com/linecorp/armeria/server/RoutingTrie.java
```diff
@@ -437,7 +459,12 @@ private void addRoute(Node<V> node, String path, V value) {
 
         // These values are sorted every time a new value is added.
         @Nullable
-        private List<V> values;
+        private List<V> valuesWithHighPrecedence;
+        @Nullable
+        private List<V> valuesWithLowPrecedence;
+        // Lazily populated from `valuesWithHighPrecedence` and `valuesWithLowPrecedence`.
```
anuraaga: If I'm not mistaken, the mutations all happen from a `Builder` of the router. Is it possible to split a `NodeBuilder` and a built `Node` which only has an immutable `allValues`? I think it would be simpler code than having to maintain an `allValues` based on two other lists.

Alternatively, since a router is mostly a singleton, I'm not sure we need to lazy initialize these lists, and it would make it much easier to read without the lazy calls.

core/src/main/java/com/linecorp/armeria/server/RoutingTrie.java
```diff
@@ -515,16 +552,29 @@ public String toString() {
...
+
             values.add(value);
 
             // Sort the values using the given comparator.
             if (comparator != null && values.size() > 1) {
                 values.sort(comparator);
             }
+
+            // Invalidate the copy of concat(nonRedirectValues, redirectValues).
+            allValues = null;
```
anuraaga: I think this is thread-safe only because we assume reads of `allValues` will only ever happen on a built router. Doing this explicitly by separating `NodeBuilder` and `Node` would be clearer IMO and maybe somewhat better performance.

core/src/main/java/com/linecorp/armeria/server/RoutingTrie.java
```diff
@@ -400,6 +410,18 @@ private void addRoute(Node<V> node, String path, V value) {
             final Node<V> parent = child.parent();
             return parent == null ? child : parent.addChild(child);
         }
+
+        static final class Entry<V> {
```
ikhoon: nit: Add `private`?
```suggestion
        private static final class Entry<V> {
```

trustin: Done

core/src/main/java/com/linecorp/armeria/server/RoutingTrie.java
trustin: Agreed. `RoutingTrie` is now immutable.

core/src/main/java/com/linecorp/armeria/server/RoutingTrie.java
trustin: Updated as advised.

core/src/main/java/com/linecorp/armeria/server/Routers.java
trustin: Fixed :-)

core/src/main/java/com/linecorp/armeria/server/Routers.java
trustin: Simplified

trustin:
Please review again, because this pull request has been refactored significantly to make `RoutingTrie` immutable.

trustin:
Let me rebase and squash. Something is not right with the build.

core/src/main/java/com/linecorp/armeria/server/Routers.java
```diff
@@ -216,18 +218,45 @@
...
+                        case PREFIX: // `/foo/*`
+                            // Add an extra route of an exact match.
+                            builder.add(path.substring(0, pathLen - 1),
+                                        fallbackValue, /* hasHighPrecedence */ false);
+                            break;
+                        default:
+                            throw new Error("Unexpected path type: " + route.pathType());
+                    }
+                }
+                builder.add(route.paths().get(1), v);
```
minwoox: Not sure this is a good idea, but how about just checking the last character if it's `/` or not because the path type is one of `EXACT`, `PARAMETERIZED` and `PREFIX`?
```suggestion
final Route route = routeResolver.apply(v);
builder.add(route.paths().get(1), v);
final String path = route.paths().get(0);
final int pathLen = path.length();
if (fallbackValue != null && pathLen > 1) {
    // Add some extra routes to handle the case where a client sends a request
    // without a trailing slash. `PathMapping.apply()` will produce a `RoutingResult`
    // whose type is `NEEDS_REDIRECT`.
    if (path.charAt(pathLen - 1) == '/') {
        // Add an extra route without a trailing slash for a redirect.
        builder.add(path.substring(0, pathLen - 1),
                    fallbackValue, /* hasHighPrecedence */ false);
    }
}
```

trustin: I think it simplifies a lot. Thanks for suggestion.

core/src/main/java/com/linecorp/armeria/server/RoutingTrieBuilder.java
```diff
@@ -0,0 +1,422 @@
...
+            current = next;
+            path = path.substring(same);
+        }
+    }
+
+    /**
+     * Converts the given character to the key of the children map.
+     * This is only used while building a {@link RoutingTrie}.
+     */
+    static char convertKey(char key) {
```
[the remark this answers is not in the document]
trustin: It's accessed by `NodeBuilder`, so Java compiler will generate a synthetic bridge method for it if we make it private. https://javax0.wordpress.com/2014/02/26/syntethic-and-bridge-methods/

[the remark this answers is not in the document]
ikhoon: I learned a lot! Thanks for sharing the link.

trustin:
Thanks for reviewing.

anuraaga:
Sorry for late followup - LGTM

PR #2293, opened 2019-12-03
Implement RoutingContextWrapper equals / hashCode / toString

anuraaga:
While looking through routing code, I randomly found that this class doesn't have equals / hashCode. It means it will break any routing cache (not only is it itself never cached but it will pollute the cache with garbage and cause it to be ineffective). Luckily, it seems to only be used in rarer cases of falling back to trailing slash and prefix + regex paths, which are rare, and composite service router, which doesn't use the server's main router so has a limited scope for cache pollution.

In another PR, I'm going to remove `summary()`, it seems like an inefficient way to save just a bit of code and isn't our normal pattern for hashCode / equals in other classes. But want to take it a step at a time if it's ok.

Also fixed JMH compilation that probably stopped working with a version update.

After
```
Benchmark                             Mode  Cnt        Score        Error  Units
RoutersBenchmark.exactMatch_wrapped  thrpt    5  6175815.558 ▒ 276154.231  ops/s
```

Before
```
Benchmark                             Mode  Cnt        Score        Error  Units
RoutersBenchmark.exactMatch_wrapped  thrpt    5  2531313.388 ▒ 118154.560  ops/s
```

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/common/stream/StreamMessageBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/common/stream/StreamMessageBenchmark.java
@@ -54,9 +54,9 @@ public class StreamMessageBenchmark {
     }
 
     @State(Scope.Thread)
-    private static class StreamObjects {
+    public static class StreamObjects {
 
-        private enum StreamType {
+        public enum StreamType {
             DEFAULT_STREAM_MESSAGE,
             FIXED_STREAM_MESSAGE,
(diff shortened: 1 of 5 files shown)


trustin:
@anuraaga wrote:

> In another PR, I'm going to remove `summary()`, it seems like an inefficient way to save just a bit of code and isn't our normal pattern for hashCode / equals in other classes. But want to take it a step at a time if it's ok.

Yes, please. Thanks in advance!

Issue #1631, opened 2019-03-06
Log the client IP address behind reverse-proxies


tobias-:
This issue is really two parts.

The first part is discovering the ip address of the client. The current Armeria solution is based on searching through different sources for the first acceptable address that passes the filters. 

Problems:
1. I don't know the proxy's address given how GKE assigns the address I can't just filter it out. It's not a private address either.
2.  It may be susceptible to the following trick: `curl https://host/ -H 'x-forwarded-for: 8.8.8.8'` (untested). (basically forging the origin of the server)

Most logs seems to log the contents of the `x-forwarded-for` header verbatim to get around that. (which may be the correct solution tbh).

One solution that does the trick:
```kotlin
    // Takes the nth proxy/server from the end to find the real client ip
    fun RequestLog.remoteHost(proxyIndex: Int): String {
        val header = requestHeaders()[X_FORWARDED_FOR_HEADER] ?: ""
        val remoteSocketIp = (context().remoteAddress<SocketAddress>() as? InetSocketAddress)?.address?.toString() ?: "local?"
        val hosts = (header.splitToSequence(", ") + remoteSocketIp).toList()
        return if (hosts.size > proxyIndex) {
            hosts[hosts.size - 1 - proxyIndex]
        } else {
            // fallback to remote host, just to have some value
            hosts.last()
        }
    }
```
---

The second problem is how to get that ip into the access log. Most of the stuff in the log interfaces are package private so it's not just a matter of dropping in a function that provides that field into the log formats. I would've liked to just use the `AccessLogFormats.COMBINED`, make a copy of the list where I replaced the first entry, and then just create a method same as the `AccessLogWriter.combined()`, possibly via a 
```java
    static AccessLogWriter custom(List<AccessLogComponent> accessLogComponents) {
        ImmutableList<AccessLogComponent> logComponents = ImmutableList.toImmutableList(accessLogComponents);
        return requestLog -> AccessLogger.write(logComponents, requestLog);
    }
```

tobias-:
This is the discussion that led up to the issue:
https://line-armeria.slack.com/conversation/C1NGPBUH2/p1551785804013500

trustin:
@ikhoon will work on this.

PR #2294, opened 2019-12-03
Provide a way to specify client IP from proxied addresses

ikhoon:
Motivation:
It is not easy to log a user's desired IP among multiple proxied IP addresses. See #1631

Modification:
* Add `clientAddressMapper()` to `ServerBuilder` and `ServerRequestContext`
* Revise `ProxiedAddresses` to hold multiple destination addresses

Result:
* A user can now select a client IP address from proxied addresses.

Fixes: #1631

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
@@ -78,8 +78,8 @@ public class DefaultServiceRequestContext extends NonWrappingRequestContext impl
     @Nullable
     private final SSLSession sslSession;
 
-    @Nullable
     private final ProxiedAddresses proxiedAddresses;
+
     private final InetAddress clientAddress;
 
     private final DefaultRequestLog log;
@@ -117,14 +117,13 @@ public class DefaultServiceRequestContext extends NonWrappingRequestContext impl
      * @param routingResult the result of finding a matched {@link Route}
      * @param request the request associated with this context
      * @param sslSession the {@link SSLSession} for this invocation if it is over TLS
-     * @param proxiedAddresses source and destination addresses retrieved from PROXY protocol header
+     * @param proxiedAddresses source and destination addresses delivered through proxy servers
      * @param clientAddress the address of a client who initiated the request
      */
     public DefaultServiceRequestContext(
             ServiceConfig cfg, Channel ch, MeterRegistry meterRegistry, SessionProtocol sessionProtocol,
             RequestId id, RoutingContext routingContext, RoutingResult routingResult, HttpRequest request,
-            @Nullable SSLSession sslSession, @Nullable ProxiedAddresses proxiedAddresses,
-            InetAddress clientAddress) {
+            @Nullable SSLSession sslSession, ProxiedAddresses proxiedAddresses, InetAddress clientAddress) {
         this(cfg, ch, meterRegistry, sessionProtocol, id, routingContext, routingResult, request,
              sslSession, proxiedAddresses, clientAddress, false, 0, 0);
     }
@@ -142,7 +141,7 @@ public class DefaultServiceRequestContext extends NonWrappingRequestContext impl
      * @param routingResult the result of finding a matched {@link Route}
      * @param request the request associated with this context
      * @param sslSession the {@link SSLSession} for this invocation if it is over TLS
-     * @param proxiedAddresses source and destination addresses retrieved from PROXY protocol header
+     * @param proxiedAddresses source and destination addresses delivered through proxy servers
      * @param clientAddress the address of a client who initiated the request
      * @param requestStartTimeNanos {@link System#nanoTime()} value when the request started.
      * @param requestStartTimeMicros the number of microseconds since the epoch,
@@ -151,17 +150,18 @@ public class DefaultServiceRequestContext extends NonWrappingRequestContext impl
     public DefaultServiceRequestContext(
             ServiceConfig cfg, Channel ch, MeterRegistry meterRegistry, SessionProtocol sessionProtocol,
             RequestId id, RoutingContext routingContext, RoutingResult routingResult, HttpRequest request,
-            @Nullable SSLSession sslSession, @Nullable ProxiedAddresses proxiedAddresses,
-            InetAddress clientAddress, long requestStartTimeNanos, long requestStartTimeMicros) {
+            @Nullable SSLSession sslSession, ProxiedAddresses proxiedAddresses, InetAddress clientAddress,
+            long requestStartTimeNanos, long requestStartTimeMicros) {
         this(cfg, ch, meterRegistry, sessionProtocol, id, routingContext, routingResult, request,
-             sslSession, proxiedAddresses, clientAddress, true, requestStartTimeNanos, requestStartTimeMicros);
+             sslSession, proxiedAddresses, clientAddress, true, requestStartTimeNanos,
+             requestStartTimeMicros);
     }
 
     private DefaultServiceRequestContext(
             ServiceConfig cfg, Channel ch, MeterRegistry meterRegistry, SessionProtocol sessionProtocol,
             RequestId id, RoutingContext routingContext, RoutingResult routingResult, HttpRequest req,
-            @Nullable SSLSession sslSession, @Nullable ProxiedAddresses proxiedAddresses,
-            InetAddress clientAddress, boolean requestStartTimeSet, long requestStartTimeNanos,
+            @Nullable SSLSession sslSession, ProxiedAddresses proxiedAddresses, InetAddress clientAddress,
+            boolean requestStartTimeSet, long requestStartTimeNanos,
             long requestStartTimeMicros) {
 
         super(meterRegistry, sessionProtocol, id,
@@ -174,7 +174,7 @@ public class DefaultServiceRequestContext extends NonWrappingRequestContext impl
         this.routingContext = routingContext;
         this.routingResult = routingResult;
         this.sslSession = sslSession;
-        this.proxiedAddresses = proxiedAddresses;
+        this.proxiedAddresses = requireNonNull(proxiedAddresses, "proxiedAddresses");
         this.clientAddress = requireNonNull(clientAddress, "clientAddress");
 
         log = new DefaultRequestLog(this, cfg.requestContentPreviewerFactory(),
@@ -245,7 +245,7 @@ public class DefaultServiceRequestContext extends NonWrappingRequestContext impl
 
         final DefaultServiceRequestContext ctx = new DefaultServiceRequestContext(
                 cfg, ch, meterRegistry(), sessionProtocol(), id, routingContext,
-                routingResult, req, sslSession(), proxiedAddresses(), clientAddress);
+                routingResult, req, sslSession(), proxiedAddresses(), clientAddress());
 
         if (rpcReq != null) {
             ctx.updateRpcRequest(rpcReq);
@@ -531,7 +531,6 @@ public class DefaultServiceRequestContext extends NonWrappingRequestContext impl
         return removeAdditionalResponseHeader(additionalResponseTrailersUpdater, name);
     }
 
-    @Nullable
     @Override
     public ProxiedAddresses proxiedAddresses() {
         return proxiedAddresses;
--- a/core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
@@ -36,6 +36,8 @@ import com.google.common.base.CharMatcher;
 import com.google.common.base.Splitter;
 import com.google.common.base.Splitter.MapSplitter;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import com.google.common.net.HostAndPort;
 
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpHeaders;
@@ -59,7 +61,8 @@ final class HttpHeaderUtil {
     static final Function<String, String> FORWARDED_CONVERTER = value -> TOKEN_SPLITTER.split(value).get("for");
 
     /**
-     * Returns an {@link InetAddress} of a client who initiated a request.
+     * Returns {@link ProxiedAddresses} which were delivered through a proxy server.
+     * Returns the specified {@code remoteAddress} if no valid proxy address is found.
      *
      * @param headers the HTTP headers which were received from the client
      * @param clientAddressSources a list of {@link ClientAddressSource}s which are used to determine
@@ -70,44 +73,49 @@ final class HttpHeaderUtil {
      * @see <a href="https://tools.ietf.org/html/rfc7239">Forwarded HTTP Extension</a>
      * @see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For">X-Forwarded-For</a>
      */
-    static InetAddress determineClientAddress(HttpHeaders headers,
-                                              List<ClientAddressSource> clientAddressSources,
-                                              @Nullable ProxiedAddresses proxiedAddresses,
-                                              InetAddress remoteAddress,
-                                              Predicate<InetAddress> filter) {
+    static ProxiedAddresses determineProxiedAddresses(HttpHeaders headers,
+                                                      List<ClientAddressSource> clientAddressSources,
+                                                      @Nullable ProxiedAddresses proxiedAddresses,
+                                                      InetSocketAddress remoteAddress,
+                                                      Predicate<InetAddress> filter) {
         for (final ClientAddressSource source : clientAddressSources) {
-            final InetAddress addr;
             if (source.isProxyProtocol()) {
                 if (proxiedAddresses != null) {
-                    addr = ((InetSocketAddress) proxiedAddresses.sourceAddress()).getAddress();
-                    if (filter.test(addr)) {
-                        return addr;
+                    if (filter.test(proxiedAddresses.sourceAddress().getAddress()) &&
+                        filter.test(proxiedAddresses.destinationAddresses().get(0).getAddress())) {
+                        return proxiedAddresses;
                     }
                 }
             } else {
+                final ImmutableList.Builder<InetSocketAddress> builder = ImmutableList.builder();
                 final AsciiString name = source.header();
                 if (name.equals(HttpHeaderNames.FORWARDED)) {
-                    addr = getFirstValidAddress(
-                            headers.get(HttpHeaderNames.FORWARDED), FORWARDED_CONVERTER, filter);
+                    headers.getAll(HttpHeaderNames.FORWARDED).forEach(forwarded -> {
+                        getAllValidAddress(forwarded, FORWARDED_CONVERTER, filter, builder);
+                    });
                 } else {
-                    addr = getFirstValidAddress(headers.get(name), Function.identity(), filter);
+                    headers.getAll(name).forEach(header -> {
+                        getAllValidAddress(header, Function.identity(), filter, builder);
+                    });
                 }
-                if (addr != null) {
-                    return addr;
+
+                final ImmutableList<InetSocketAddress> addresses = builder.build();
+                if (!addresses.isEmpty()) {
+                    return ProxiedAddresses.of(addresses.get(0), addresses.subList(1, addresses.size()));
                 }
             }
         }
         // We do not apply the filter to the remote address.
-        return remoteAddress;
+        return ProxiedAddresses.of(remoteAddress);
     }
 
     @VisibleForTesting
-    @Nullable
-    static InetAddress getFirstValidAddress(@Nullable String headerValue,
-                                            Function<String, String> valueConverter,
-                                            Predicate<InetAddress> filter) {
+    static void getAllValidAddress(@Nullable String headerValue,
+                                   Function<String, String> valueConverter,
+                                   Predicate<InetAddress> filter,
+                                   ImmutableList.Builder<InetSocketAddress> accumulator) {
         if (Strings.isNullOrEmpty(headerValue)) {
-            return null;
+            return;
         }
         for (String value : CSV_SPLITTER.split(headerValue)) {
             final String v = valueConverter.apply(value);
@@ -115,19 +123,18 @@ final class HttpHeaderUtil {
                 continue;
             }
             try {
-                final InetAddress addr = createInetAddress(v);
-                if (addr != null && filter.test(addr)) {
-                    return addr;
+                final InetSocketAddress addr = createInetSocketAddress(v);
+                if (addr != null && filter.test(addr.getAddress())) {
+                    accumulator.add(addr);
                 }
             } catch (UnknownHostException e) {
-                logger.debug("Failed to resolve hostname: {}", v, e);
+                logger.debug("Failed to create an InetSocketAddress: {}", v, e);
             }
         }
-        return null;
     }
 
     @Nullable
-    private static InetAddress createInetAddress(String address) throws UnknownHostException {
+    private static InetSocketAddress createInetSocketAddress(String address) throws UnknownHostException {
         final char firstChar = address.charAt(0);
         if (firstChar == '_' ||
             (firstChar == 'u' && "unknown".equals(address))) {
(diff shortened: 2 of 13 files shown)


core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -277,7 +277,8 @@ public Object getMessage0(RequestLog log) {
                     } else {
                         // %a means the client address who initiated a request.
                         final ServiceRequestContext ctx = (ServiceRequestContext) log.context();
-                        return ctx.clientAddress().getHostAddress();
+                        return ctx.clientAddressMapper().apply(ctx.proxiedAddresses())
+                                  .getHostString();
```
ikhoon: Question: I just log `host` only and want to hear other folks' opinions.
* Should we need to log `host:port`? 🤔
* If we need, fix `REMOTE_HOST` and `LOCAL_IP_ADDRESS` too?

trustin: Address only because most components here stem from a format string specification of some old HTTPd.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -174,8 +180,9 @@ private DefaultServiceRequestContext(
         this.routingContext = routingContext;
         this.routingResult = routingResult;
         this.sslSession = sslSession;
-        this.proxiedAddresses = proxiedAddresses;
-        this.clientAddress = requireNonNull(clientAddress, "clientAddress");
+        this.proxiedAddresses = requireNonNull(proxiedAddresses, "proxiedAddresses");
+        this.clientAddressMapper = requireNonNull(clientAddressMapper, "clientAddressMapper");
```
trustin: How about removing `clientAddressMapper`? I think a user can still get it from `ServerConfig`, so it's better keeping request-scope members only for reduced memory footprint.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -531,12 +538,16 @@ public boolean removeAdditionalResponseTrailer(CharSequence name) {
...
-    @Nullable
     @Override
     public ProxiedAddresses proxiedAddresses() {
         return proxiedAddresses;
     }
 
+    @Override
+    public Function<ProxiedAddresses, InetSocketAddress> clientAddressMapper() {
+        return clientAddressMapper;
+    }
```
trustin: I guess this will not be used very often, so it's probably better removing it.

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -159,7 +159,10 @@ static ServiceRequestContextBuilder builder(HttpRequest request) {
 
     /**
      * Returns the address of the client who initiated this request.
+     *
+     * @deprecated Use {@link #proxiedAddresses()} with {@link #clientAddressMapper()}
      */
+    @Deprecated
     default InetAddress clientAddress() {
```
trustin: I think this method should not be deprecated because it is fairly often that a user wants to know about the actual client address when writing a webapp.

Could you make sure this method returns the correct client address calculated with `clientAddressMapper` and `proxiedAddresses`?

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -419,8 +422,9 @@ default MediaType negotiatedProduceType() {
     boolean removeAdditionalResponseTrailer(CharSequence name);
 
     /**
-     * Returns the proxied addresses if the current {@link Request} is received through a proxy.
+     * Returns the proxied addresses of the current {@link Request}.
      */
-    @Nullable
     ProxiedAddresses proxiedAddresses();
+
+    Function<ProxiedAddresses, InetSocketAddress> clientAddressMapper();
```
trustin: Remove this

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -56,7 +56,7 @@
  * An access log component to generate a log message.
  */
 @FunctionalInterface
-interface AccessLogComponent {
+public interface AccessLogComponent {
```
trustin: The classes you made public are meant to be used internally. Please explain why you had to make them public.

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogFormats.java
```diff
@@ -53,14 +53,14 @@
 /**
  * Pre-defined access log formats and the utility methods for {@link AccessLogComponent}.
  */
-final class AccessLogFormats {
+public final class AccessLogFormats {
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogType.java
```diff
@@ -47,7 +47,7 @@
  *
  * </table>
  */
-enum AccessLogType {
+public enum AccessLogType {
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
```diff
@@ -277,8 +277,7 @@ public Object getMessage0(RequestLog log) {
                     } else {
                         // %a means the client address who initiated a request.
                         final ServiceRequestContext ctx = (ServiceRequestContext) log.context();
-                        return ctx.clientAddressMapper().apply(ctx.proxiedAddresses())
-                                  .getHostString();
+                        return ctx.clientAddress().getHostAddress();
```
trustin: I didn't mean you have to revert. I meant port should not be included. Could you revert the revert?

ikhoon: Oops 😅. I have a question about `clientAddressMapper()`. `AccessLogComponent` is difficult to access `ServerConfig` in the current API because `CommonComponent` is made statically.
https://github.com/line/armeria/blob/15decdfbc881ee0e3b12ca13f396010a54e39ed2/core/src/main/java/com/linecorp/armeria/server/logging/AccessLogFormats.java#L82-L83
 
So I was thinking about two options. 
1) Add `clientAddressMapper()` to `ServiceRequestContext`
2) Add `clientAddressMapper()` to `DefaultRequestLog`

Do I choose the second option? Or if you have another idea, let me know. 😊

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
ikhoon: The things requested in #1631 was 1) a way to select client IP among proxied addresses and 2) modify existing `COMBINED` components to configure access log format.

>  I would've liked to just use the AccessLogFormats.COMBINED, make a copy of the list where I replaced the first entry, and then just create a method same as the AccessLogWriter.combined(), possibly via a
> ```java
>    static AccessLogWriter custom(List<AccessLogComponent> accessLogComponents) {
>        ImmutableList<AccessLogComponent> logComponents = ImmutableList.toImmutableList(accessLogComponents);
>        return requestLog -> AccessLogger.write(logComponents, requestLog);
>    }
> ```

Do you think we won't open the API?

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
trustin: You can just use `ServiceRequestContext.clientAddress()` because you applied the mapper there already.

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
trustin: No. The reason why @tobias- has to do that was because we could not get the correct client address. Because we fixed `%a` in this pull request, there's nothing to do for access logs.

ikhoon: Thanks for the reply! Let me rever this :-)

ikhoon: fixed

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
ikhoon: I revert this. Now `ctx.clientAddress()` is result of `ctx.clientAddressMapper().apply(ctx.proxiedAddresses())`.

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogComponent.java
ikhoon: > Could you revert the revert?

Had chat about `revert the revert` and we talked that the current code(reverted version) is fine. 😀

ikhoon:
Updated PR description

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
```diff
@@ -141,32 +147,54 @@ private static InetAddress createInetAddress(String address) throws UnknownHostE
...
-                // Remove the port part. e.g. [2001:db8:cafe::17]:4711 => [2001:db8:cafe::17]
-                final String withoutPort = addr.substring(0, delim + 1);
-                if (NetUtil.isValidIpV6Address(withoutPort)) {
-                    return InetAddress.getByAddress(NetUtil.createByteArrayFromIpAddressString(withoutPort));
-                }
-            }
+    @Nullable
+    private static InetSocketAddress createInetSocketAddress(String address, char delimiter)
+            throws UnknownHostException {
+        final int delim = address.indexOf(delimiter);
```
trustin: I think it is not prohibited to use an IPv4 address with brackets, or there might be some weird implementation that does so. What do you think about using `HostAndPort.fromString()` to extract host and port?

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
```diff
@@ -141,32 +147,54 @@ private static InetAddress createInetAddress(String address) throws UnknownHostE
...
+                } else {
+                    port = Integer.parseInt(address.substring(delim + 1));
                 }
             }
         }
-        // Skip if it is an invalid address.
-        logger.debug("Failed to parse an address: {}", address);
-        return null;
+        return new InetSocketAddress(
+                InetAddress.getByAddress(NetUtil.createByteArrayFromIpAddressString(ip)), port);
```
trustin: You need to handle the case where `createByteArrayFromIpAddressString()` returns `null`.

core/src/main/java/com/linecorp/armeria/server/ProxiedAddresses.java
```diff
@@ -18,51 +18,86 @@
...
+        return new ProxiedAddresses(sourceAddress, ImmutableList.of());
     }
 
-    private final SocketAddress sourceAddress;
-    private final SocketAddress destinationAddress;
+    /**
+     * Creates a new instance with the specified {@code sourceAddress} and {@code destinationAddress}.
+     */
+    public static ProxiedAddresses of(InetSocketAddress sourceAddress, InetSocketAddress destinationAddress) {
+        return new ProxiedAddresses(sourceAddress, ImmutableList.of(destinationAddress));
```
trustin: requirenonnull

core/src/main/java/com/linecorp/armeria/server/ProxiedAddresses.java
```diff
@@ -18,51 +18,86 @@
...
+    }
+
+    /**
+     * Creates a new instance with the specified {@code sourceAddress} and {@code destinationAddresses}.
+     */
+    public static ProxiedAddresses of(InetSocketAddress sourceAddress,
+                                      Iterable<? extends InetSocketAddress> destinationAddresses) {
+        return new ProxiedAddresses(sourceAddress, ImmutableList.copyOf(destinationAddresses));
```
trustin: requirenonnull

core/src/main/java/com/linecorp/armeria/server/ProxiedAddresses.java
```diff
@@ -18,51 +18,86 @@
...
+
+        final ProxiedAddresses that = (ProxiedAddresses) o;
+        return sourceAddress.equals(that.sourceAddress) &&
+               destinationAddresses.equals(that.destinationAddresses);
+    }
+
+    @Override
+    public int hashCode() {
+        return sourceAddress.hashCode() * 31 + destinationAddresses.hashCode();
+    }
```
trustin: `toString()` should come before `hashCode` and `equals`, like other classes.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1278,6 +1279,15 @@ public ServerBuilder clientAddressFilter(Predicate<InetAddress> clientAddressFil
         return this;
     }
 
+    /**
+     * Sets a {@link Function} to use when determining the client address from {@link ProxiedAddresses}.
```
trustin: Describe the default behavior.

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContextBuilder.java
```diff
@@ -169,6 +158,8 @@ public ServiceRequestContext build() {
         final RoutingResult routingResult =
                 this.routingResult != null ? this.routingResult
                                            : RoutingResult.builder().path(path()).query(query()).build();
+        final InetAddress clientAddress = server.config().clientAddressMapper().apply(proxiedAddresses)
+                                                .getAddress();
```
trustin: Seems like `clientAddressMapper` is not configurable when using `ServiceRequestContext`. Is this intentional?

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogFormats.java
```diff
@@ -69,10 +69,17 @@
                              ofPredefinedCommon(RESPONSE_STATUS_CODE), BLANK,
                              ofPredefinedCommon(RESPONSE_LENGTH));
 
+    /**
+     * Returns a common log format.
+     */
+    public static List<AccessLogComponent> ofCommon() {
+        return COMMON;
```
trustin: revert

core/src/main/java/com/linecorp/armeria/server/logging/AccessLogFormats.java
```diff
@@ -84,6 +91,13 @@
                              ofQuotedRequestHeader(HttpHeaderNames.USER_AGENT), BLANK,
                              ofQuotedRequestHeader(HttpHeaderNames.COOKIE));
 
+    /**
+     * Returns a combined log format.
+     */
+    public static List<AccessLogComponent> ofCombined() {
+        return COMBINED;
```
trustin: revert

core/src/test/java/com/linecorp/armeria/server/HttpHeaderUtilTest.java
```diff
@@ -22,205 +22,273 @@
...
+                                            new InetSocketAddress("192.0.2.61", 4711),
+                                            new InetSocketAddress("192.0.2.62", 4711)));
 
         // IPv6
         assertThat(xForwardedFor("[2001:db8:cafe::17],[2001:db8:cafe::18],[2001:db8:cafe::19]"))
-                .isEqualTo(InetAddress.getByName("2001:db8:cafe::17"));
+                .isEqualTo(ImmutableList.of(
+                        new InetSocketAddress("[2001:db8:cafe::17]", 0),
+                        new InetSocketAddress("[2001:db8:cafe::18]", 0),
+                        new InetSocketAddress("[2001:db8:cafe::19]", 0)));
```
trustin: global comment - Brackets are unnecessary when calling `new InetSocketAddress()`

core/src/test/java/com/linecorp/armeria/server/HttpHeaderUtilTest.java
```diff
@@ -22,205 +22,273 @@
...
+                .isEqualTo(ImmutableList.of(
+                        new InetSocketAddress("[2001:db8:cafe::17]", 4711),
+                        new InetSocketAddress("[2001:db8:cafe::18]", 4711)));
 
         // The following cases are not a part of X-Forwarded-For specifications, but the first element is
         // definitely not valid.
         // A obfuscated identifier
         assertThat(xForwardedFor("_superhost,[2001:db8:cafe::17]:4711,[2001:db8:cafe::18]:4711"))
-                .isEqualTo(InetAddress.getByName("2001:db8:cafe::17"));
+                .isEqualTo(ImmutableList.of(
```
trustin: global comment: `.containsExactly(new InetSocketAddress(...), new InetSocketAddress(...)...);`

core/src/test/java/com/linecorp/armeria/server/logging/AccessLogFormatsTest.java
```diff
@@ -263,18 +264,20 @@ void formatMessage() {
 
     @Test
     void logClientAddress() throws Exception {
-        final InetSocketAddress remote = new InetSocketAddress(InetAddress.getByName("10.1.0.1"), 5000);
+        final InetSocketAddress remote = new InetSocketAddress(InetAddress.getByName("10.1.0.1"), 0);
+        final ProxiedAddresses proxied = ProxiedAddresses.of(
+                new InetSocketAddress(InetAddress.getByName("10.1.0.2"), 5000));
```
trustin: `InetAddress.getByName()` seems redundant?

core/src/main/java/com/linecorp/armeria/server/ProxiedAddresses.java
```diff
@@ -18,51 +18,86 @@
...
-    public static ProxiedAddresses of(SocketAddress sourceAddress, SocketAddress destinationAddress) {
-        return new ProxiedAddresses(sourceAddress, destinationAddress);
+    public static ProxiedAddresses of(InetSocketAddress sourceAddress) {
+        return new ProxiedAddresses(sourceAddress, ImmutableList.of());
     }
 
-    private final SocketAddress sourceAddress;
-    private final SocketAddress destinationAddress;
+    /**
+     * Creates a new instance with the specified {@code sourceAddress} and {@code destinationAddress}.
```
trustin: Mention that port `0` means that the port number is unknown for all factory/getter Javadocs.

core/src/main/java/com/linecorp/armeria/server/ProxiedAddresses.java
minwoox: You mean `toString()` should come after ...? 😄

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
```diff
@@ -70,64 +72,68 @@
...
             try {
-                final InetAddress addr = createInetAddress(v);
-                if (addr != null && filter.test(addr)) {
-                    return addr;
+                final InetSocketAddress addr = createInetSocketAddress(v);
+                if (addr != null && filter.test(addr.getAddress())) {
+                    accumulator.add(addr);
                 }
             } catch (UnknownHostException e) {
                 logger.debug("Failed to resolve hostname: {}", v, e);
```
minwoox: not related this PR but I think we need to change this something like
`Failed to create an IP address:`

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
```diff
@@ -141,32 +147,54 @@ private static InetAddress createInetAddress(String address) throws UnknownHostE
...
-            return InetAddress.getByAddress(NetUtil.createByteArrayFromIpAddressString(addr));
+        final InetSocketAddress inetSocketAddress;
+        try {
+            if (addr.charAt(0) == '[') {
+                inetSocketAddress = createInetSocketAddress(addr, ']');
+            } else {
+                inetSocketAddress = createInetSocketAddress(addr, ':');
+            }
+            // Log debug message if it is an invalid address.
+            if (inetSocketAddress == null) {
```
minwoox: This never happens, right?

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
```diff
@@ -141,32 +147,54 @@ private static InetAddress createInetAddress(String address) throws UnknownHostE
...
-        if (addr.charAt(0) == '[') {
-            final int delim = addr.indexOf(']');
-            if (delim > 0) {
-                // Remove the port part. e.g. [2001:db8:cafe::17]:4711 => [2001:db8:cafe::17]
-                final String withoutPort = addr.substring(0, delim + 1);
-                if (NetUtil.isValidIpV6Address(withoutPort)) {
-                    return InetAddress.getByAddress(NetUtil.createByteArrayFromIpAddressString(withoutPort));
-                }
-            }
+    @Nullable
```
minwoox: not nullable?

core/src/main/java/com/linecorp/armeria/server/ProxiedAddresses.java
trustin: YEEES :rofl

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
ikhoon: That is good advice. Let me apply it. 👍

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
ikhoon: Yes. While refactoring I miss it. 😅

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
ikhoon: That is a nice point. 🙏

core/src/main/java/com/linecorp/armeria/server/ProxiedAddresses.java
ikhoon: I will keep in mind. 😎

core/src/test/java/com/linecorp/armeria/server/HttpHeaderUtilTest.java
ikhoon: Thanks for education.

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
ikhoon: > I think it is not prohibited to use an IPv4 address with brackets

I have removed `NetUtil.isValidIpV{4,6}Address()` because I thought the `NetUtil.createByteArrayFromIpAddressString()` validate address string already. 

https://github.com/netty/netty/blob/dcbfe17eeb588a8ba2acb95800776526d01cc5a2/common/src/main/java/io/netty/util/NetUtil.java#L368

But it was my mistake. After your review, I totally understand why the original code needs to check twice. 😊

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
```diff
@@ -146,57 +147,21 @@ private static InetSocketAddress createInetSocketAddress(String address) throws
...
-        final InetSocketAddress inetSocketAddress;
         try {
-            if (addr.charAt(0) == '[') {
-                inetSocketAddress = createInetSocketAddress(addr, ']');
-            } else {
-                inetSocketAddress = createInetSocketAddress(addr, ':');
+            final HostAndPort hostAndPort = HostAndPort.fromString(addr);
+            final byte[] addressBytes = NetUtil.createByteArrayFromIpAddressString(hostAndPort.getHost());
+            if (addressBytes == null) {
+                logger.debug("Fail to parse an address: {}", address);
```
trustin: Fail -> Failed

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
```diff
@@ -146,57 +147,21 @@ private static InetSocketAddress createInetSocketAddress(String address) throws
...
-            if (inetSocketAddress == null) {
-                logger.debug("Failed to parse an address: {}", address);
-            }
-            return inetSocketAddress;
-        } catch (NumberFormatException e) {
-            logger.debug("Fail to parse a port from an address: {}", address, e);
+            return new InetSocketAddress(InetAddress.getByAddress(addressBytes),
+                                         hostAndPort.getPortOrDefault(0));
+        } catch (IllegalArgumentException e) {
+            logger.debug(e.getMessage(), e);
```
trustin: I wouldn't rely on `e.getMessage()` because it may be `null` although `createByteArrayFromIpAddressString()` always throws with a message.

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
```diff
@@ -128,7 +128,7 @@ static void getAllValidAddress(@Nullable String headerValue,
                     accumulator.add(addr);
                 }
             } catch (UnknownHostException e) {
-                logger.debug("Failed to resolve hostname: {}", v, e);
+                logger.debug("Failed to create an IP address: {}", v, e);
```
trustin: Failed to create an InetSocketAddress? Obviously you did not create an IP address.

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
trustin: `logger.debug("Failed to parse an address: {}, address, e);`

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
ikhoon: Yes, it's my mistake. Thanks for pointing this out!

core/src/main/java/com/linecorp/armeria/server/HttpHeaderUtil.java
ikhoon: I will address this comment. https://github.com/line/armeria/pull/2294#discussion_r354091569 😀

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -124,6 +127,7 @@
             List<ClientAddressSource> clientAddressSources,
             Predicate<InetAddress> clientAddressTrustedProxyFilter,
             Predicate<InetAddress> clientAddressFilter,
+            Function<ProxiedAddresses, InetSocketAddress> clientAddressMapper,
```
trustin: `Function<? super ProxiedAddresses, ? extends InetSocketAddress>`?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1278,6 +1279,16 @@ public ServerBuilder clientAddressFilter(Predicate<InetAddress> clientAddressFil
...
+     * If not set, the {@link ProxiedAddresses#sourceAddress()}} is used as a client address.
+     */
+    public ServerBuilder clientAddressMapper(
+            Function<ProxiedAddresses, InetSocketAddress> clientAddressMapper) {
```
trustin: `Function<? super ProxiedAddresses, ? extends InetSocketAddress>`?

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
ikhoon: Oops! Thanks for pointing out. 😭

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContextBuilder.java
ikhoon: I thought `clientAddressMapper` could be configurable with `serverConfigurator(Consumer<? super ServerBuilder> serverConfigurator)`.

trustin:
@tobias- Any comments?

ikhoon:
Let me rebase and update the commit message.

PR #2295, opened 2019-12-04
Optimize RoutingContext by removing summary generation and othe…

anuraaga:
…l tweaks.

- Don't generate a summary array for use in hashing and do the computation directly to reduce allocations
- Make accept headers extraction non-lazy, it was being eagerly evaluated during summary extraction in the old code anyways. It isn't really possible to use routing cache with lazy attributes anyways
- Don't use `forEach` due to extra `Stream` allocation and lambda allocation during extraction (current `forEach` holds a reference to the builder)
- Don't allocate list builder when `getAll` headers and the header isn't present

After
```
Benchmark                             Mode  Cnt         Score        Error  Units
RoutersBenchmark.exactMatch          thrpt    5  12921841.803 ▒ 173790.211  ops/s
RoutersBenchmark.exactMatch_wrapped  thrpt    5  12047216.990 ▒ 351620.117  ops/s
```

Before (wrapped is slower due to double allocation of summary)
```
Benchmark                             Mode  Cnt        Score         Error  Units
RoutersBenchmark.exactMatch          thrpt    5  7817340.954 ▒  926608.528  ops/s
RoutersBenchmark.exactMatch_wrapped  thrpt    5  6175815.558 ▒ 276154.231  ops/s
```

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpHeadersBase.java
@@ -264,10 +264,15 @@ class HttpHeadersBase implements HttpHeaderGetters {
     }
 
     private ImmutableList<String> getAllReversed(CharSequence name) {
-        final ImmutableList.Builder<String> builder = ImmutableList.builder();
         final int h = AsciiString.hashCode(name);
         final int i = index(h);
         HeaderEntry e = entries[i];
+
+        if (e == null) {
+            return ImmutableList.of();
+        }
+
+        final ImmutableList.Builder<String> builder = ImmutableList.builder();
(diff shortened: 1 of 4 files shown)


anuraaga:
For reference, I tend to look hard at routing code when trying to squeeze out performance since it's a lot of useful functionality, but not needed in RPC-only frameworks like gRPC upstream. So to be able to compete in performance, we need to make sure it's as fast as possible - this means having the router cache and making sure construction of the routing context is as lightweight as possible. Hopefully we find more ways to reduce this extra processing even further :)

anuraaga:
@trustin your other PR inspired me to use do/while here :D

PR #2296, opened 2019-12-05
Fix noisy `WARN` logs on client disconnection

trustin:
Related: #2248
Motivation:

When a client disconnects from a server, the `HttpResponse` being
prepared by a service was simply aborted with an `AbortedStreamException`.

However, since 0.96.0, thanks to #2248, an `HttpResponse` is aborted
with more specific exception, such as:

- `ClosedSessionException` on a disconnection
- `Http2Exception.StreamException` on an HTTP/2 RST_STREAM frame

However, `HttpResponseSubscriber` was not updated to handle the above
exceptions, which means they will leave noisy `WARN` level logs.

Modifications:

- Handle the exceptions which may cancel the current stream properly.
  - Add `Exceptions.isStreamCancelling()`
- Fix a bug where `HttpResponseSubscriber` records
  `responseFirstBytesTransferredTimeNanos` incorrectly.
- Make sure to cancel the health check response timeout task when a
  client disconnected, so that the timeout task do not stay in the event
  loop task queue longer than necessary.

Result:

- Less noisy logging
- `ws` (wire send) timestamp for a server response is not recorded when it's not really sent.
- Tad bit more efficient long-polling health check

### DIFF
--- a/brave/src/test/java/com/linecorp/armeria/it/brave/BraveIntegrationTest.java
+++ b/brave/src/test/java/com/linecorp/armeria/it/brave/BraveIntegrationTest.java
@@ -451,35 +451,35 @@ class BraveIntegrationTest {
     }
 
     @Test
-    void testClientTimesOut() throws Exception {
-        assertThatThrownBy(() -> timeoutClientClientTimesOut.hello("name"))
-                .isInstanceOf(ResponseTimeoutException.class);
-        final Span[] spans = spanReporter.take(2);
-
-        final Span serverSpan = findSpan(spans, "service/timeout");
-        final Span clientSpan = findSpan(spans, "client/timeout");
-
-        // HTTP2 client timed out, so RST_STREAM frame was sent to the server.
-        // There is a wire send in the server and no wire receive in the client.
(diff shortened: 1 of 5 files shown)


brave/src/test/java/com/linecorp/armeria/it/brave/BraveIntegrationTest.java
```diff
@@ -451,35 +451,35 @@ void testServerTimesOut() throws Exception {
...
         final Span clientSpan = findSpan(spans, "client/timeout");
 
-        // HTTP1 client timed out, so no response data was ever sent from the server.
+        // Collect all annotations except for connection attempts.
+        final List<Annotation> serverAnnotations = serverSpan.annotations();
+        final List<Annotation> clientAnnotations = clientSpan.annotations().stream()
+                                                             .filter(a -> !a.value().contains("connect"))
+                                                             .collect(toImmutableList());
+
+        // Client timed out, so no response data was ever sent from the server.
```
ikhoon: Nice!

PR #2297, opened 2019-12-05
Fix flaky tests

trustin:
- Fix a race condition in `HttpHealthCheckedEndpointGroupLongPollingTest`
- Fix the failure to delete a temporary directory in `HttpFileServiceTest`
- Miscellaneous:
  - Fix too large test timeouts

### DIFF
--- a/brave/src/test/java/com/linecorp/armeria/it/brave/BraveIntegrationTest.java
+++ b/brave/src/test/java/com/linecorp/armeria/it/brave/BraveIntegrationTest.java
@@ -86,7 +86,7 @@ import zipkin2.Annotation;
 import zipkin2.Span;
 import zipkin2.reporter.Reporter;
 
-@Timeout(10000)
+@Timeout(10)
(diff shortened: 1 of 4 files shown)


trustin:
Still failing although the chance has gone lower:

```
com.linecorp.armeria.server.file.HttpFileServiceTest > executionError FAILED
    java.io.IOException: Failed to delete temp directory C:\Users\appveyor\AppData\Local\Temp\1\junit5085164456569485523. The following paths could not be deleted (see suppressed exceptions for details): 
```

anuraaga:
I've seen this on Windows too in another project - just gave up and don't delete the directory and rely on using `./gradlew clean` to clean up junk occasionally. It seems to be a hard problem to reliably know when a directory is ready to delete on Windows.

https://github.com/curioswitch/curiostack/blob/master/tools/gradle-plugins/gradle-test-helpers/src/main/java/org/curioswitch/gradle/testing/GradleTempDirectories.java#L38

PR #2300, opened 2019-12-06
Slim down docs client

### LINKED ISSUE #2299: Investigate the JAR size increase between 0.94.0 and 0.95.0
- 0.94.0 - 6780408 bytes
- 0.95.0 - 8322580 bytes

:thinking:

anuraaga:
The main reason for using favicons-webpack-plugin is to avoid having the same source file in the repo twice, but we don't need to generate a full suite of favicons for a debug page.

We also can avoid bundling in fonts to reduce size further. Worst case is if in a constrainted network fonts don't load, the page will still work fine (and maybe even with correct fonts if the OS contains roboto)

Incidentally, I found out that favicon won't work in most cases since it doesn't support relocation (e.g., mounting at `/docs`) until https://github.com/jantimon/favicons-webpack-plugin/pull/173

Fixes #2299

### DIFF
--- a/docs-client/webpack.config.ts
+++ b/docs-client/webpack.config.ts
@@ -95,7 +95,13 @@ const config: Configuration = {
     new HtmlWebpackPlugin({
       template: './src/index.html',
     }),
-    new FaviconsWebpackPlugin('./src/images/logo.png'),
+    new FaviconsWebpackPlugin({
+      logo: './src/images/logo.png',
+      // We don't need the many different icon versions of webapp mode and use light mode
+      // to keep JAR size down.
+      mode: 'light',
+      devMode: 'light',
+    }),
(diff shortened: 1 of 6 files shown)


docs-client/webpack.config.ts
```diff
@@ -95,7 +95,13 @@ const config: Configuration = {
     new HtmlWebpackPlugin({
       template: './src/index.html',
     }),
-    new FaviconsWebpackPlugin('./src/images/logo.png'),
+    new FaviconsWebpackPlugin({
+      logo: './src/images/logo.png',
+      // We don't need the many different icon versions of webapp mode and use light mode
```
ikhoon: 👍

trustin:
Thanks, @anuraaga !

PR #2301, opened 2019-12-06
Reduce bit operations for header validation

anuraaga:
While reading through https://github.com/line/armeria/commit/b597f7a865a527a84ee3d6937075cfbb4470ed20 to understand it, I had to look up what the `~` operator does 😅  I think it's very uncommonly used and better to avoid if it's not necessary to improve readability. I found that because we don't need to presize the `BitSet` (it's a static final) and `BitSet.get` efficiently checks characters outside the range of the `BitSet`, it doesn't seem necessary and we can make the code more readable. What do you think?

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/HttpHeaderNames.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpHeaderNames.java
@@ -69,6 +69,7 @@ public final class HttpHeaderNames {
 
     private static final BitSet PROHIBITED_NAME_CHARS;
     private static final String[] PROHIBITED_NAME_CHAR_NAMES;
+    private static final byte LAST_PROHIBITED_NAME_CHAR;
 
     static {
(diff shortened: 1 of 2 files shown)


trustin:
How about keeping the `& ...MASK` check? Most characters will hit that branch and we're gonna same some cycles. We could define the masks without `~`, like plain `0xFFF0`.

anuraaga:
I think the difference in operations is just one, the initial `< 0` check that `BitSet.get` does (otherwise just a shift and compare, similar to mask and compare). But if it's worth it, how about

```java
private static final byte LAST_PROHIBITED_NAME_CHAR = (byte) PROHIBITED_NAME_CHARS.size() - 1;
if (value > LAST_PROHIBITED_NAME_CHAR)
```

It's a tiny bit less precise than the mask for prohibited names (two additional characters get through `>` and `?`) but keeps the logic more self-contained (no need to have confidence the mask matches the bitset, it's all self-computed).

PR #2303, opened 2019-12-07
Add flag to control header validation.

anuraaga:
For https://github.com/line/armeria/security/advisories/GHSA-35fr-h7jr-hh86 we have added back header validation to prevent injection attacks. I think there are many many servers that don't have exposure to this issue, most of them because they only use server-server RPC which is always HTTP/2, and others because they don't have any chance of setting headers with untrusted content. In the interest of our recent goal of reducing overhead for RPC-only scenarios (for reference, I don't see such validation in grpc-java, where it shouldn't be needed as they're HTTP/2-only), I propose adding a flag to allow users to disable the validation.

```
Benchmark                                  Mode  Cnt         Score        Error  Units
HttpHeadersBenchmark.create_noValidation  thrpt    5  24338315.495 ▒ 347537.645  ops/s
HttpHeadersBenchmark.create_validation    thrpt    5   5541946.523 ▒  29086.369  ops/s
```

### DIFF
--- a/benchmarks/src/jmh/java/com/linecorp/armeria/common/HttpHeadersBenchmark.java
+++ b/benchmarks/src/jmh/java/com/linecorp/armeria/common/HttpHeadersBenchmark.java
@@ -19,11 +19,18 @@ package com.linecorp.armeria.common;
 import javax.annotation.Nullable;
 
 import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Fork;
 
 /**
  * Microbenchmarks of {@link DefaultHttpHeaders} construction.
  */
 public class HttpHeadersBenchmark {
+
+    private static final String AUTHORIZATION_TOKEN =
+            // Sample JWT from https://jwt.io. Most auth tokens will be significantly longer.
+            "bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9." +
+            "eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ." +
+            "SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
 
     @Nullable
     @Benchmark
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -864,6 +867,27 @@ public static boolean useJdkDnsResolver() {
         return USE_JDK_DNS_RESOLVER;
     }
 
+    /**
+     * Disables validation of HTTP headers for dangerous characters like newlines - such characters can be used
+     * for injecting arbitrary content into HTTP responses.
+     *
+     * <p><strong>DISCLAIMER:</strong>Do not enable this if you don't know what you are doing. It is recommended
```
minwoox: nit: Do not disable?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -301,6 +301,9 @@
 
     private static final boolean USE_JDK_DNS_RESOLVER = getBoolean("useJdkDnsResolver", false);
 
+    private static final boolean UNSAFE_DISABLE_HEADER_VALIDATION =
+            getBoolean("unsafeDisableHeaderValidation", false);
```
trustin: How about `headerValidationEnabled` with `true` default?

anuraaga: Sure - went with `validateHeaders`, which is somewhat like `useEpoll` and `dumpSslInfo`

benchmarks/src/jmh/java/com/linecorp/armeria/common/HttpHeadersBenchmark.java
```diff
@@ -40,4 +47,15 @@ public MediaType parseUnknown() {
...
         return headers.contentType();
     }
+
+    @Benchmark
+    public HttpHeaders create_validation() {
+        return HttpHeaders.of(HttpHeaderNames.AUTHORIZATION, AUTHORIZATION_TOKEN);
+    }
+
+    @Benchmark
+    @Fork(jvmArgsAppend = "-Dcom.linecorp.armeria.unsafeDisableHeaderValidation=true")
```
ikhoon:
```suggestion
    @Fork(jvmArgsAppend = "-Dcom.linecorp.armeria.validateHeaders=false")
```

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -864,6 +866,27 @@ public static boolean useJdkDnsResolver() {
         return USE_JDK_DNS_RESOLVER;
     }
 
+    /**
+     * Enables validation of HTTP headers for dangerous characters like newlines - such characters can be used
+     * for injecting arbitrary content into HTTP responses.
+     *
+     * <p><strong>DISCLAIMER:</strong>Do not disable this unless you know what you are doing. It is recommended
```
trustin: nit: Missing space before `Do not`

trustin:
Thanks, @anuraaga !

PR #2309, opened 2019-12-10
Fix a bug where query params disappear when `WebClient` executes

ildartuykin:
Bugfix- disappearing query params, during webclient request execution with undefined endpoint, right CLA commit email


trustin:
Thanks again, @ildartuykin!

Issue #1739, opened 2019-04-25
Provide a way to get `ClientRequestContext` without using a decorator.


trustin:
Currently, a user has to insert a decorator to access the `ClientRqeustContext` or its `RequestLog`. For example:

```java
HttpClient client = new HttpClientBuilder(...)
    .decorator((delegate, ctx, req) -> {
        // Access ctx and ctx.log()
    })
    .build();
```

It would be pretty convenient if a user can do this:

```java
HttpClient client = HttpClient.of(...);
HttpResponse res = client.get("/");
ClientRequestContext ctx = ClientRequestContext.get(res);
```

geminiKim:
Hi :)
I was wondering about this issue while looking into Armeria.
I think it is simple to solve this task if the HttpResponse has a RequestContext.

like:
```
// ClientRequestContext.java

    static ClientRequestContext get(HttpResponse res) {
        if(res instanceof DecodedHttpResponse) { 
            return ((DecodedHttpResponse) res).getRequestContext();
        } else {
            return null; // But there is not perfect.
        }
    }


// DecodedHttpResponse.java

    private final ClientRequestContext requestContext;

    public DecodedHttpResponse(EventLoop eventLoop, ClientRequestContext requestContext) {
        this.eventLoop = eventLoop;
        this.requestContext = requestContext;
    }

    public ClientRequestContext getRequestContext() {
        return requestContext;
    }
```

I wonder about your opinion on the right way.

minwoox:
Hello @geminiKim, thanks for the good suggestion.!
But, the returned `HttpResponse` couldn't be `DecodedHttpResponse`.
It can be `DeferredHttpResponse`, `FilteredHttpResponse`, `DuplicatedHttpResponse` or `HttpDecodedResponse` depending on the decorators that a user specified.
I think we have to consider all of these situations because otherwise, the user would get the `ctx` only in certain circumstance. :-)
By the way, are you using Armeria? 😄

geminiKim:
@minwoox  I understand and agreed, Thank you for good feedback, It is really help for me :D
I’m not yet using Armeria (in production), just started to studying to learn good code and the contribution :)

I think I'll ask you a lot of questions. Thank you in opinion. :D

minwoox:
Got it. I hope you will be able to use Armeria in production one day. :-)
For this issue, I think we can wrap the `HttpResponse` with the class which has the original `HttpResponse` and `ClientRequestContext` in `DefaultHttpClient`.
```
[...]

Issue #2232, opened 2019-11-05
When do we need to call ctx.push()?


minwoox:
Found out that we are pushing context while notifying log listeners:
```java
try (SafeCloseable ignored = log.context().push()) {
    listener.onRequestLog(log);
}
```
The listener can get the `ctx` from the method parameter `log` using `log.context()`.
Should we push the `ctx` in such a case in which the thread is able to get the `ctx` from the parameter?

minwoox:
One thing I noticed recently was that some of our clients are using `ctx.onEnter()` and there is some heavy initialization logic called multiple times. However, after #2227 is merged, it shouldn't be a problem though.

trustin:
#2375 will remove `onEnter()` and `onExit()` by the way.

trustin:
I agree that we need to avoid pushing a context unnecessarily, and thus we should not push a context when a callback is given a context as a parameter. Thoughts? /cc @ikhoon @imasahiro @kojilin @anuraaga

anuraaga:
I agree with the general idea of not pushing when `ctx` is referencable by the user. There is one caveat in that sometimes features require the context to actually be in a thread local to work - if we didn't push context in the brave decorators tracing wouldn't work properly. So sometimes it's not only about finding a `ctx` variable in scope or not. This will be a very low minority of the cases though.

minwoox:
Oops, I forgot to leave a comment for this.
I think it's okay to push the context when we call `onRequestLog()`.
(Sorry I've changed my mind. 🙏)

The reason I filed this issue is that 1) we had a bug which was raising an exception when calling `ctx.makeContextAware` with a `ClientRequestContext` in the thread-local 2) I saw some really heavy initialization logic in `onEnter()`.
But now, we fixed the bug and we will remove the `onEnter` and move to using sort of `Storage` which users wouldn't not put some heavy initial logic when they just call `doAttach()` (I guess 😅).

Also, as @anuraaga explained, a user could add a decorator which sends a request to another backend when `RequestLogAvailalibility.COMPLETE` is fulfilled. It will need the ctx in the thread-local.

So I think we need to push the ctx when calling `onRequestLog()`.
Sorry for making a noise again. 🙏

trustin:
[...]

PR #2311, opened 2019-12-11
Propagate ClientRequestContext to client callbacks

### LINKED ISSUE #760: ClientRequestContext is not propagated to client callbacks

### LINKED ISSUE #1083: A user may get the wrong RequestContext when contexts are nested
When a user writes a service that invokes a request using an Armeria client, the `ServiceRequestContext` will be *hidden* by the client's `ClientRequestContext`. This can cause various unexpected hard-to-spot behaviors, especially when a user is using frameworks like RxJava, such as:

- When a user accesses the properties of `RequestContext`, a wrong value may be returned.
- When a user assumes the context is `ServiceRequestContext`, a `ClassCastException` will be raised.
- When a user access an attribute, the attribute may be missing. A user has to copy all attributes using `onChild()` to work around this.

minwoox:
Motivation:
`ClientRequestContext` is not propagate to `AsyncMethodCallback` and the callbacks registered to `RpcResponse`.

Modifications:
- Push `ClientRequestContext` into thread-local in `DeferredHttpResponse`, `DecodedHttpResponse` and `HttpResponseDuplicator` when invoking:
  - `Subscriber.onSubscribe()`
  - `Subscriber.onError()`
  - `Subscrbier.onComplete()`
  - Please note that the context is not pushed when calling `Subscriber.onNext()`
- `DeferredStreamMessage` calls `Subscriber.onSubscribe()` to downstream always after it subscribes to upstream

Result:
- You can now get `ClientRequestContext` in `AsyncMethodCallback` and `Subscribers` which subscribes to `HttpResponse` using `RequestContext.currentOrNull()`.
- (Breaking) `DeferredHttpResponse` is removed from public APIs. Use `HttpResponse.from()`.
- Close #760

To-do:
- Should fix #1083 before this is merged

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
@@ -35,6 +35,7 @@ import com.spotify.futures.CompletableFutures;
 
 import com.linecorp.armeria.common.CommonPools;
 import com.linecorp.armeria.common.RequestContext;
+import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.armeria.internal.PooledObjects;
 
 import io.netty.util.ReferenceCountUtil;
@@ -166,11 +167,6 @@ abstract class AbstractStreamMessage<T> implements StreamMessage<T> {
     }
 
     /**
-     * Returns the current demand.
-     */
-    abstract long demand();
-
-    /**
      * Callback invoked by {@link Subscription#request(long)} to add {@code n} to demand.
      */
     abstract void request(long n);
@@ -187,6 +183,15 @@ abstract class AbstractStreamMessage<T> implements StreamMessage<T> {
     abstract void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEvent event);
 
     /**
+     * Pushes the context to the thread-local stack if this {@link StreamMessage} has it.
+     * This method is invoked before {@link Subscriber#onSubscribe(Subscription)},
+     * {@link Subscriber#onComplete()} and {@link Subscriber#onError(Throwable)} are called.
+     */
+    protected SafeCloseable pushContextIfExist() {
+        return () -> { /* no-op */};
+    }
+
+    /**
      * Invoked after an element is removed from the {@link StreamMessage} and before
      * {@link Subscriber#onNext(Object)} is invoked.
      *
@@ -194,18 +199,23 @@ abstract class AbstractStreamMessage<T> implements StreamMessage<T> {
      */
     protected void onRemoval(T obj) {}
 
-    static void failLateSubscriber(SubscriptionImpl subscription, Subscriber<?> lateSubscriber) {
+    private void failLateSubscriber(SubscriptionImpl subscription, Subscriber<?> lateSubscriber) {
         final Subscriber<?> oldSubscriber = subscription.subscriber();
         final Throwable cause = abortedOrLate(oldSubscriber);
 
         if (subscription.needsDirectInvocation()) {
-            lateSubscriber.onSubscribe(NoopSubscription.INSTANCE);
-            lateSubscriber.onError(cause);
+            failLateSubscriber0(lateSubscriber, cause);
         } else {
             subscription.executor().execute(() -> {
-                lateSubscriber.onSubscribe(NoopSubscription.INSTANCE);
-                lateSubscriber.onError(cause);
+                failLateSubscriber0(lateSubscriber, cause);
             });
+        }
+    }
+
+    private void failLateSubscriber0(Subscriber<?> lateSubscriber, Throwable cause) {
+        try (SafeCloseable ignored = pushContextIfExist()) {
+            lateSubscriber.onSubscribe(NoopSubscription.INSTANCE);
+            lateSubscriber.onError(cause);
         }
     }
 
@@ -347,7 +357,6 @@ abstract class AbstractStreamMessage<T> implements StreamMessage<T> {
         public String toString() {
             return MoreObjects.toStringHelper(Subscription.class)
                               .add("publisher", publisher)
-                              .add("demand", publisher.demand())
                               .add("executor", executor).toString();
         }
     }
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
@@ -107,7 +107,7 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
             duplicatorExecutor = currentExecutor;
         }
 
-        processor = new StreamMessageProcessor<>(upstream, signalLengthGetter,
+        processor = new StreamMessageProcessor<>(this, upstream, signalLengthGetter,
                                                  duplicatorExecutor, maxSignalLength);
     }
 
@@ -171,6 +171,26 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
         processor.abort(cause);
     }
 
+    /**
+     * Invoked when {@link Subscriber#onSubscribe(Subscription)} is called by the
+     * {@linkplain StreamMessage upstream}. The subclass who wants to push the context before calling callbacks
+     * on {@link Subscriber} should override this method and store the context.
+     *
+     * @see #pushContextIfExist()
+     */
+    protected void onSubscribeCalled() {}
+
+    /**
+     * Pushes the context to the thread-local stack if this {@link AbstractStreamMessageDuplicator} has it.
+     * This method is invoked before {@link Subscriber#onSubscribe(Subscription)},
+     * {@link Subscriber#onComplete()} and {@link Subscriber#onError(Throwable)} are called.
+     *
+     * @see #onSubscribeCalled()
+     */
+    protected SafeCloseable pushContextIfExist() {
+        return () -> { /* no-op */};
+    }
+
     @VisibleForTesting
     static class StreamMessageProcessor<T> implements Subscriber<T> {
 
@@ -190,6 +210,7 @@ public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage
             CLOSED
         }
 
+        private final AbstractStreamMessageDuplicator<?, ?> duplicator;
         private final StreamMessage<T> upstream;
         private final SignalQueue signals;
         private final SignalLengthGetter<Object> signalLengthGetter;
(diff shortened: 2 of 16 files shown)


anuraaga:
We already do this for gRPC here

https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java#L350

And I think doing something similar in thrift client makes lots of sense.

I'm less sure about the subscriber - subscriber, and even http response, are concepts that don't related to server or client. It seems pretty hacky to have them best-effort push a client context. Also why not push for `onNext`?

Is it not enough that when subscribing we can specify `ctx.contextAwareEventLoop` when subscribing?

minwoox:
> Is it not enough that we can specify ctx.contextAwareEventLoop when subscribing?

As described in https://github.com/line/centraldogma/pull/450/, it's hard to get `ClientRequestContext` if we don't insert a decorator to store the context. So we cannot call `ctx.contextAwareEventLoop`.

I thought the user will be expecting the subscriber has the context in the thread-local. But if you think this is too much then I can just do pushing context for `AsyncMethodCallback`.
But one thing I noticed is that If we push the context and if we implement `RequestContext.parent()`, then the user does not have to use `makeContextAware` most of the time.
```java
@Override
HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
    final CompletableFuture<AggregatedHttpResponse> future = webClient.execute(req).aggregate();
    future.handle((res, cause) -> { // Do not have call ctx.makeContextAware(future)
       webClient2.execute(newReq).aggregate();
       ...
    });
    ...
}
```


For `onNext`, it can be called multiple times unlike other methods, I have chosen not to push the context. But I think if we need it, I can do it.

anuraaga:
I've thought about automatic context propagation in the context of tracing, but run into the wall that even if Armeria handles it well, anytime using something else like a DB will require context awareness so have figured it's either all-in or all-out in some sense. But I guess outside of tracing, implicit transfer could still be good enough for many cases. For reference, `grpc-context` has no automatic transfer at all (there isn't even an automatic context, user's create the first one if they need it, for example in a tracing interceptor).

It reminds me of your question a while ago of whether we should even push context when a method signature provides access to it. `RequestContext.current()` I think has it's primary purpose as giving access to the context from non-Armeria RPC APIs where it's otherwise impossible - I don't think it's as important to push it elsewhere. For the `Subscriber`, I think semantically it is too strange to only push it for most, but not all the callbacks - but I agree the performance impact of mounting on `onNext` could be huge, and as an example, in gRPC we would have `onNext` for every binary payload in a stream but only really need to mount when notifying of a parsed message. So if all can't be context-aware, maybe that's a good enough argument for none of them to?

core/src/test/java/com/linecorp/armeria/client/ClientRequestContextPushedOnSubscriberTest.java
```diff
@@ -0,0 +1,275 @@
...
+            public void onError(Throwable t) {}
+
+            @Override
+            public void onComplete() {
+                requestContexts.add(RequestContext.current());
+                latch.countDown();
+            }
+        }, nextEventLoop());
+        latch.await();
+        assertThat(requestContexts.size()).isEqualTo(2);
```
trustin: This assertion is unnecessary.

core/src/test/java/com/linecorp/armeria/client/ClientRequestContextPushedOnSubscriberTest.java
```diff
@@ -0,0 +1,275 @@
...
+                requestContexts.add(RequestContext.current());
+                latch.countDown();
+            }
+
+            @Override
+            public void onComplete() {}
+        }, nextEventLoop());
+
+        latch.await();
+        assertThat(requestContexts.size()).isEqualTo(2);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/ClientRequestContextPushedOnSubscriberTest.java
```diff
@@ -0,0 +1,275 @@
...
+                latch.countDown();
+            }
+
+            @Override
+            public void onComplete() {}
+        }, nextEventLoop());
+        res.abort();
+
+        latch.await();
+        assertThat(requestContexts.size()).isEqualTo(2);
```
trustin: Ditto

core/src/test/java/com/linecorp/armeria/client/ClientRequestContextPushedOnSubscriberTest.java
```diff
@@ -0,0 +1,275 @@
...
+                requestContexts.add(RequestContext.current());
+                latch.countDown();
+            }
+
+            @Override
+            public void onComplete() {}
+        }, nextEventLoop());
+
+        latch.await();
+        assertThat(requestContexts.size()).isEqualTo(2);
```
trustin: Ditto

thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientDelegate.java
```diff
@@ -108,7 +109,9 @@ public RpcResponse execute(ClientRequestContext ctx, RpcRequest call) throws Exc
             }
         } catch (Throwable cause) {
             reply.completeExceptionally(cause);
-            return reply;
+            try (SafeCloseable ignored = ctx.push()) {
+                return reply;
```
trustin: Why do we need to push the context here?

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
```diff
@@ -186,6 +182,15 @@ protected EventExecutor defaultSubscriberExecutor() {
      */
     abstract void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEvent event);
 
+    /**
+     * Pushes the context to the thread-local stack if this {@link StreamMessage} has it.
+     * This method is invoked before {@link Subscriber#onSubscribe(Subscription)},
+     * {@link Subscriber#onComplete()} and {@link Subscriber#onError(Throwable)} are called.
+     */
+    protected SafeCloseable pushContextIfExist() {
+        return () -> { /* no-op */};
```
trustin: nit: Missing space between `*/` and `}`

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessageDuplicator.java
```diff
@@ -171,6 +171,26 @@ public void abort(Throwable cause) {
...
+
+    /**
+     * Pushes the context to the thread-local stack if this {@link AbstractStreamMessageDuplicator} has it.
+     * This method is invoked before {@link Subscriber#onSubscribe(Subscription)},
+     * {@link Subscriber#onComplete()} and {@link Subscriber#onError(Throwable)} are called.
+     *
+     * @see #onSubscribeCalled()
+     */
+    protected SafeCloseable pushContextIfExist() {
+        return () -> { /* no-op */};
```
trustin: Ditto

trustin:
> We already do this for gRPC here https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaClientCall.java#L350

Oh, shouldn't we only do that in `onReady` and `onClose()` (and maybe in `onMetadata()`)? It may be premature optimization though.

@minwoox Would you mind measuring the performance impact of pushing context on `onNext()`?

anuraaga:
`onMessage` is what invokes the gRPC stub code and while it is possible to deal with `onReady` and store context somewhere, most code doesn't even handle `onReady` - the standard callbacks are `onNext`, `onError`, and `onCompleted` for a gRPC client stub and if we didn't mount here then `onNext` would not have any access to context.

minwoox:
I am on a training session so I might be late to reply. Sorry about that.
> Would you mind measuring the performance impact of pushing context on onNext()?

Yes, I can try. But before that, I wanted to address that the logic in the hook which is attached using `ReqeustContxte.onEnter()` is the most affected factor.
And I think I want to remove the `onEnter()` because it has two drawbacks:
- `onEnter` isn't invoked when the context is first pushed into the thread-local
  - A user attaches the hook to the context which is already pushed.
- Most of the users do not know when `onEnter` is called.
  - They do not expect the `onEnter` is called when `RequestLogListener` is invoked.
Because of these reasons, I really get rid of the API. Or at least, I want it to change to be called when `makeContextAware` series.
Let's discuss more about it on #2232 

> I've thought about automatic context propagation in the context of tracing, but run into the wall that even if Armeria handles it well, anytime using something else like a DB will require context awareness so have figured it's either all-in or all-out in some sense. But I guess outside of tracing, implicit transfer could still be good enough for many cases. For reference, grpc-context has no automatic transfer at all (there isn't even an automatic context, user's create the first one if they need it, for example in a tracing interceptor).

Most of the questions about `makeContextAware` that I got from the users are the cases that I shared in https://github.com/line/armeria/pull/2311#issuecomment-564392082. So I think if we support the automatic propagating that will really reduce the burden of the entry-level users.
But yes, they do have to know how it works to get using Armeria better.
And I also don't disagree that the explicit could be better.
So let's put it apart and focus on whether we need the context in the subscriber methods or not.
Do we agree that we need to push the context if it is not degrading the performance a lot?

anuraaga:
I'd be fine with pushing the context in `Subscriber` if it applies to all the callbacks consistently without significant performance impact. If we find an impact, I don't think I'd go for one that only applies to some callbacks as it's a confusing restriction. Also, not necessarily in this PR, but if we do push for client, can we also push for server for consistency?

trustin:
I'm OK with the removal of `onEnter` events, but I'm curious if there is any concrete use case we can't solve without using `onEnter()`. Thoughts, @anuraaga?

minwoox:
Let me do the benchmark first. Thanks! 😃

minwoox:
I realized that pushing the context before `Subscriber.onComplete()` is not enough because the callback cannot get the context if it's attached after the future is complete:
```java
CompletableFuture future = webClient.execute(req).aggregate();
future.hande((msg, cause) -> {
    // This thread will have the ClientRequestContext in the thread-local
    // most of the time.
    ...
);
....
future.hande((msg, cause) -> {
    // This thread will not have the ClientRequestContext in the thread-local
    // most of the time because the future is already completed. 
    ...
);
```

I think returning `RequestContextAwareCompletableFuture` is the only way to solve this issue. If we do that, we might not need to push the context before `Subscriber.onX`. In order to do that, we need to fix #1739 first so that we can make the `RequestContextAwareCompletableFuture` when `StreamMessage.subscribe()` is called.

I want to work on #1083 first it must be addressed before we pushing the context into the thread which executes the callbacks, and then I will do #1739.

So let me close this issue for now and reopen this if we still need to push the context before `Subscriber.onX`.
Thanks!

anuraaga:
Thanks for investigating! By the way, what do you think about pushing the context in-band for thrift, similar to what we do in gRPC. I guess around here

https://github.com/line/armeria/blob/master/thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientDelegate.java#L281

minwoox:
Yes, I have thought about it, but I thought I need to make the `DefaultRpcResponse` extends `RequestContextAwareCompletableFuture` as well so that the callback attached after the `RpcResponse` is complete will have the context. 🤔

anuraaga:
Ah got it that probably makes more sense

thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientDelegate.java
minwoox: I don't know. 😅

minwoox:
> I'm OK with the removal of onEnter events, but I'm curious if there is any concrete use case we can't solve without using onEnter()

@delegacy might have an idea on this because I've seen him using it.

PR #2310, opened 2019-12-10
Add grpc-service-reactor example

cbornet:
This adds an example for gRPC using the [reactive-grpc](https://github.com/salesforce/reactive-grpc) library.

See https://github.com/line/armeria-examples/pull/2

I, @cbornet, hereby release the Git commits in [this pull request](https://github.com/line/armeria-examples/pull/2310) ('The Contribution') under [Creative Commons Zero 1.0](https://creativecommons.org/publicdomain/zero/1.0/), so that anyone can modify and/or redistribute The Contribution without any restriction.

PR #2312, opened 2019-12-11
Move `examples/grpc-service` to `examples/grpc`

trustin:
.. and renamed the project from `grpc-service` to `grpc-example`.


trustin:
I'm not 100% sure, but my guess is that `.gitkeep` files are to signify that a user can put some files there. I'm not sure if it's a good convention or not though. I'm OK with removing them.

settings.gradle
```diff
@@ -50,7 +50,8 @@ includeWithFlags ':site'
 
 // Examples
 includeWithFlags ':examples:annotated-http-service', 'java'
-includeWithFlags ':examples:grpc-service',           'java'
+includeWithFlags ':examples:grpc-example',           'java'
```
ikhoon: nit: How about `grpc-minimal`? And we could adapt #2310 as `grpc-reactive`?

I'm not sure this suggestion is better than `grpc-example`. 🙃 You can feel free to ignore this.

anuraaga:
I'd lean towards removing them or adding a text string inside with the motivation. But off the top of my head, I'm not sure if Java ignores dot files in resources - if not I guess it's not great to add dummy files to the result of the project.

trustin:
I don't think Java or Gradle ignores dot files. Do they?

settings.gradle
trustin: I don't know either. I guess it doesn't matter much anyway though. Let me leave it as it is :wink:

anuraaga:
Yeah I guess they do - so to keep output clean, maybe better to not have them. But not a huge deal either way

PR #2314, opened 2019-12-11
Make `RequestContextExporter` a part of public API

### LINKED ISSUE #1510: Create an abstraction for a LoggingContext containing request properties
Currently, armeria natively exports many useful attributes to logback logs

https://github.com/line/armeria/blob/master/logback/src/main/java/com/linecorp/armeria/common/logback/BuiltInProperty.java

But since these are hard-coded into the logback implementation itself, it can't be reused for log4j users. Could we add to a `RequestContext` a layer that accumulates key/value pair that logging layers can read without special processing?

Here is a similar idea, but misses out on armeria's built-in logging, which would be very tedious to duplicate.

https://github.com/curioswitch/curiostack/blob/master/common/server/framework/src/main/java/org/curioswitch/common/server/framework/logging/RequestLoggingContext.java#L39

ikhoon:
Motivation:
Armeria exports many useful properties to logback `MDC` via `RequestContextExportingAppender`.
`RequestContextExportingAppender` is implemented with private classes, so it can't be reused for log4j.
This PR move code related to `RequestContextExport` for reusing it in other logging frameworks.

Modification:
* Move `BuiltinProperty`, `BuiltInProperties`, `RequestContextExporter`
  and `RequestContextExporterBuilder` to `armeria.common.logging`.
* Add getters to `RequestContextExporter`
* Make `RequestContextExporter`, `RequestContextExporterBuilder` public
* Migrate the code related with `RequestContext` from `RequestContextExportingAppender`
  to `RequestContextExporter`
* Deprecation
  * The getters from `RequestContextExporterBuilder` and `RequestContextExportingAppender` have been deprecated without replacement.

Result:
You can export desired built-in properties by using `RequestContextExporterBuilder` and `RequestContextExporter`.

Fixes #1510


ikhoon:
1) I want to listen to other's opinions on whether this direction is right.
2) I'm not sure `RequestContextAwareExporter` is good name. Let me know it is weird or have an idea. 😀

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextAwareExporter.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+
+    /**
+     * Returns a {@link Map} whose key is set through {@code <export />} and the value is exported
+     * from {@link RequestContext} and {@link RequestLog}.
+     * Note that: this method returns {@code null} if current {@link RequestContext} is {@code null}.
+     */
+    @Nullable
+    public Map<String, String> exports() {
```
trustin: `export()` ? (verb)

trustin:
I'm not sure about why `RequestContextAwareExporter` is needed after all. Can't we just:

- merge `RequestContextAwareExporter` and `RequestContextExporter` into `RequestContextExporter`
- make `RequestContextExporterBuilder` and `RequestContextExporter` public
- let uses use `RequestContextExporterBuilder` and `RequestContextExporter`?

What's the advantage of having `RequestContextAwareExporter` and `RequestContextExporter`?

ikhoon:
@trustin Thanks for the kind explanation. 

> merge `RequestContextAwareExporter` and `RequestContextExporter` into `RequestContextExporter`

Actually, I thought of merging `RequestContextAwareExporter` and `RequestContextExporter`. But I was not sure it, so that I keep the original code and add another class. Your review is SGTM. I will address it!

> make `RequestContextExporterBuilder` and `RequestContextExporter` public

I just focused on moving `RequestContextExporter` related code. After reading your review, I totally understand what I missed. 😱

ikhoon:
Now ready for review. PTAL.

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
```diff
@@ -114,7 +123,31 @@
         }
     }
 
-    void export(Map<String, String> out, RequestContext ctx, RequestLog log) {
+    /**
+     * Returns a {@link Map} whose key is an export key set through {@code add*()} in
+     * {@link RequestContextExporterBuilder} and value is extracted from {@link RequestContext}
+     * Note that: this method returns {@code null} if current {@link RequestContext} is {@code null}.
```
trustin: How about returning an empty `Map` instead?

logback/src/main/java/com/linecorp/armeria/common/logback/RequestContextExportingAppender.java
```diff
@@ -194,83 +195,44 @@ public void setExport(String mdcKey) {
...
+            exporter = builder.build();
+        }
+        final Map<String, String> contextMap = exporter.export();
+        if (contextMap != null) {
             final Map<String, String> originalMdcMap = eventObject.getMDCPropertyMap();
             final Map<String, String> mdcMap;
 
             // Create a copy of 'state' to avoid the race between:
             // - the delegate appenders who iterate over the MDC map and
             // - this class who update 'state'.
```
trustin: You need to move this comment block to the exporter.

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
```diff
@@ -114,7 +123,31 @@
...
+     */
+    @Nullable
+    public Map<String, String> export() {
+        final RequestContext ctx = RequestContext.currentOrNull();
```
trustin: How about adding `public Map<String, String> export(RequestContext ctx)` as well?

core/src/main/java/com/linecorp/armeria/common/RequestContextExporterBuilder.java
```diff
@@ -0,0 +1,251 @@
...
+                                                      Function<?, String> stringifier) {
+        requireNonNull(alias, "alias");
+        requireNonNull(attrKey, "attrKey");
+        requireNonNull(stringifier, "stringifier");
+        attrs.add(new ExportEntry<>(attrKey, PREFIX_ATTRS + alias, stringifier));
+        return this;
+    }
+
+    /**
+     * Returns {@code true} if the specified {@link AttributeKey} is in export list.
```
minwoox: nit: is in the

core/src/main/java/com/linecorp/armeria/common/RequestContextExporterBuilder.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /**
+     * Returns all {@link AttributeKey}s in the export list.
+     *
+     * @return the {@link Map} whose key is an alias and value is an {@link AttributeKey}
+     */
+    public Map<String, AttributeKey<?>> getAttributes() {
+        return Collections.unmodifiableMap(attrs.stream().collect(
+                Collectors.toMap(e -> e.exportKey.substring(PREFIX_ATTRS.length()), e -> e.key)));
```
minwoox: Could be?
```suggestion
return attrs.stream().collect(
        toImmutableMap(e -> e.exportKey.substring(PREFIX_ATTRS.length()), e -> e.key));
```

ikhoon:
> Which package would fit better for the new public classes?

First, I moved it to `com.linecorp.armeria.common.logging` but changed it to `com.linecorp.armeria.common`.
Because I thought `RequestContextExporter` implementation focus on exporting desired RequestContext's properties to `Map` dynamically which is not strongly related to logging. 

I'm less sure but `RequestContextExporter` can provide more flexible way to access `RequestContext`.

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
ikhoon: That is a good idea. 👍

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
```diff
@@ -114,7 +123,31 @@
...
+        if (ctx != null) {
+            final State state = state(ctx);
+            final RequestLog log = ctx.log();
+            final Set<RequestLogAvailability> availabilities = log.availabilities();
+
+            // Note: This equality check is extremely fast.
+            //       See RequestLogAvailabilitySet for more information.
+            if (!availabilities.equals(state.availabilities)) {
+                state.availabilities = availabilities;
+                export(state, ctx, log);
```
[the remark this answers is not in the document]
ikhoon: I think after moving this class under `common` and being public class, a user could call this method any time. If we add `RequestLogAvailability.COMPLETE` then we need description in documentation or throwing an exception?

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
ikhoon: SGTM, let me address it.

core/src/main/java/com/linecorp/armeria/common/RequestContextExporterBuilder.java
ikhoon: That is a good suggestion!

logback/src/main/java/com/linecorp/armeria/common/logback/RequestContextExportingAppender.java
ikhoon: Oops, thanks!

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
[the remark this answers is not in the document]
trustin: > Don't we need to call only when `RequestLogAvailability` is `complete` because `exportAttributes()` and others can be called multiple times?

I think this is intentional, because context properties must be exported partially even if the request is not complete. Otherwise, log messages of the requests being processed will not have any MDC properties set.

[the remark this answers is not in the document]
minwoox: Thanks for the explanation. 😄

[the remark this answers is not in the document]
ikhoon: I've learned too. 🙇‍♂️

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
```diff
@@ -114,7 +125,40 @@
...
+     * Note that: this method returns an empty {@link Map} if current {@link RequestContext} is {@code null}.
+     */
+    public Map<String, String> export() {
+        final RequestContext ctx = RequestContext.currentOrNull();
+        return ctx != null ? export(ctx) : ImmutableMap.of();
+    }
+
+    /**
+     * Returns a {@link Map} whose key is an export key set through {@code add*()} in
+     * {@link RequestContextExporterBuilder} and value is extracted from the specified {@link RequestContext}
```
trustin: Missing period

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
```diff
@@ -114,7 +125,40 @@
         }
     }
 
-    void export(Map<String, String> out, RequestContext ctx, RequestLog log) {
+    /**
+     * Returns a {@link Map} whose key is an export key set through {@code add*()} in
+     * {@link RequestContextExporterBuilder} and value is extracted from {@link RequestContext}
+     * Note that: this method returns an empty {@link Map} if current {@link RequestContext} is {@code null}.
```
trustin: How about removing `:` between `that` and `this`?

core/src/main/java/com/linecorp/armeria/common/RequestContextExporter.java
```diff
@@ -433,7 +477,7 @@ private static void exportHttpHeaders(Map<String, String> out, HttpHeaders heade
                                           ExportEntry<AsciiString>[] requiredHeaderNames) {
         for (ExportEntry<AsciiString> e : requiredHeaderNames) {
             final String value = headers.get(e.key);
-            final String mdcKey = e.mdcKey;
+            final String mdcKey = e.exportKey;
```
trustin: `mdcKey` -> `exportKey`

core/src/main/java/com/linecorp/armeria/common/RequestContextExporterBuilder.java
```diff
@@ -0,0 +1,250 @@
...
+
+    private static final String PREFIX_ATTRS = "attrs.";
+    private static final String PREFIX_HTTP_REQ_HEADERS = "req.http_headers.";
+    private static final String PREFIX_HTTP_RES_HEADERS = "res.http_headers.";
+
+    private final Set<BuiltInProperty> builtIns = EnumSet.noneOf(BuiltInProperty.class);
+    private final Set<ExportEntry<AttributeKey<?>>> attrs = new HashSet<>();
+    private final Set<ExportEntry<AsciiString>> httpReqHeaders = new HashSet<>();
+    private final Set<ExportEntry<AsciiString>> httpResHeaders = new HashSet<>();
+
```
trustin: Missing package-private default constructor. This class must be instantiated only via `RequestContextExporter.builder()`.

core/src/main/java/com/linecorp/armeria/common/RequestContextExporterBuilder.java
```diff
@@ -0,0 +1,250 @@
...
+        requireNonNull(attrKey, "attrKey");
+        requireNonNull(stringifier, "stringifier");
+        attrs.add(new ExportEntry<>(attrKey, PREFIX_ATTRS + alias, stringifier));
+        return this;
+    }
+
+    /**
+     * Returns {@code true} if the specified {@link AttributeKey} is in the export list.
+     */
+    public boolean containsAttribute(AttributeKey<?> key) {
```
trustin: How about hiding all getters from the public API by making them package-private or removing them? Then, we could add the getters to `RequestContextExporter`, without `get` method prefix.

core/src/main/java/com/linecorp/armeria/common/RequestContextExporterBuilder.java
ikhoon: Oops! Thanks for detecting it.🙏

core/src/main/java/com/linecorp/armeria/common/RequestContextExporterBuilder.java
ikhoon: Nice suggestion. I thought that removing getters from the `RequestContextExporterBuilder` too. But if remove them, the getters from `RequestContextExportingAppender` could be breaking changes because the getters could be called while building `RequestContextExporter`.
https://github.com/line/armeria/blob/29befcd911fae48e01761e37638a6473c2e76e30/logback/src/main/java/com/linecorp/armeria/common/logback/RequestContextExportingAppender.java#L117-L123

I guess the public getters from `RequestContextExportingAppender` could be only used for advanced cases so the breaking changes are not a big impact for most of the users.

ikhoon: Had discussed with @trustin, we decide to deprecate the getters from `RequestContextExporterBuilder` and `RequestContextExportingAppender`. https://github.com/line/armeria/pull/2314/commits/49ab82b7d6e060e4c553e39a35d8b46d98d80274

ikhoon:
PR description is updated.

trustin:
Please also fix Checkstyle violations.

ikhoon:
> How about moving to common.logging?

Sure, I will move it to `common.logging`

> Please also fix Checkstyle violations.

Oops, I missed it. 😱

trustin:
Thanks a lot, @ikhoon !

PR #2316, opened 2019-12-11
Add more ways to add global annotated service extensions

### LINKED ISSUE #1752: Provide a way to customize the default `ObjectMapper` used by Armeria
Armeria currently uses a vanilla `ObjectMapper` instance with the default settings when converting a request or a response, but it seems like some users want to customize it globally. There is currently no way to change the settings of the `ObjectMapper` used by `JacksonRequest/ResponseConverterFunction`. Perhaps we could provide a global access method for the `ObjectMapper` which is used when serializing or deserializing user-defined objects.

heowc:
Motivation:

Add more ways to add global annotated service extensions

Modifications:

- Add a way set AnnotatedHttpServiceExtensions from `{VirtualHost,Server}Builder`
- Change creation time about `{VirtualHost}AnnotatedServiceBindingBuilder` 

Result:

Closes #1752

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -257,29 +257,16 @@ public final class AnnotatedHttpServiceFactory {
                                                 .build());
                 }).collect(toImmutableList());
 
-        final List<ExceptionHandlerFunction> ehFromAnnotated =
-                getAnnotatedInstances(method, clazz, ExceptionHandler.class, ExceptionHandlerFunction.class);
         final List<ExceptionHandlerFunction> eh =
-                ImmutableList.<ExceptionHandlerFunction>builder().addAll(extensions.exceptionHandlers())
-                                                                 .addAll(ehFromAnnotated)
-                                                                 .addAll(baseExceptionHandlers)
-                                                                 .add(defaultExceptionHandler).build();
-
-        final List<RequestConverterFunction> reqFromAnnotated =
-                getAnnotatedInstances(method, clazz, RequestConverter.class, RequestConverterFunction.class);
+                getAnnotatedInstances(method, clazz, ExceptionHandler.class, ExceptionHandlerFunction.class)
+                        .addAll(baseExceptionHandlers).addAll(extensions.exceptionHandlers())
+                        .add(defaultExceptionHandler).build();
         final List<RequestConverterFunction> req =
-                ImmutableList.<RequestConverterFunction>builder().addAll(extensions.requestConverters())
-                                                                 .addAll(reqFromAnnotated)
-                                                                 .addAll(baseRequestConverters)
-                                                                 .build();
-
-        final List<ResponseConverterFunction> resFromAnnotated =
-                getAnnotatedInstances(method, clazz, ResponseConverter.class, ResponseConverterFunction.class);
+                getAnnotatedInstances(method, clazz, RequestConverter.class, RequestConverterFunction.class)
+                        .addAll(baseRequestConverters).addAll(extensions.requestConverters()).build();
         final List<ResponseConverterFunction> res =
-                ImmutableList.<ResponseConverterFunction>builder().addAll(extensions.responseConverters())
-                                                                  .addAll(resFromAnnotated)
-                                                                  .addAll(baseResponseConverters)
-                                                                  .build();
+                getAnnotatedInstances(method, clazz, ResponseConverter.class, ResponseConverterFunction.class)
+                        .addAll(baseResponseConverters).addAll(extensions.responseConverters()).build();
 
         final Optional<HttpStatus> defaultResponseStatus = findFirst(method, StatusCode.class)
                 .map(code -> {
@@ -665,15 +652,14 @@ public final class AnnotatedHttpServiceFactory {
      * Returns a {@link Builder} which has the instances specified by the annotations of the
      * {@code annotationType}. The annotations of the specified {@code method} and {@code clazz} will be
      * collected respectively.
-     * FIXME(heowc): Update javadoc.
      */
-    private static <T extends Annotation, R> List<R> getAnnotatedInstances(
+    private static <T extends Annotation, R> Builder<R> getAnnotatedInstances(
             AnnotatedElement method, AnnotatedElement clazz, Class<T> annotationType, Class<R> resultType) {
         final Builder<R> builder = new Builder<>();
         Stream.concat(findAll(method, annotationType).stream(),
                       findAll(clazz, annotationType).stream())
               .forEach(annotation -> builder.add(getInstance(annotation, resultType)));
-        return builder.build();
+        return builder;
     }
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -162,19 +162,19 @@ public final class AnnotatedHttpServiceFactory {
 
     /**
      * Returns the list of {@link AnnotatedHttpService} defined by {@link Path} and HTTP method annotations
-     * from the specified {@code object}, {@link ExceptionHandlerFunction}s, {@link RequestConverterFunction}s,
-     * {@link ResponseConverterFunction}s and {@link AnnotatedHttpServiceExtensions}.
+     * from the specified {@code object}, {@link RequestConverterFunction}s, {@link ResponseConverterFunction}s,
+     * {@link ExceptionHandlerFunction}s and {@link AnnotatedHttpServiceExtensions}.
      */
     public static List<AnnotatedHttpServiceElement> find(
             String pathPrefix, Object object,
-            List<ExceptionHandlerFunction> exceptionHandlerFunctions,
             List<RequestConverterFunction> requestConverterFunctions,
-            List<ResponseConverterFunction> responseConverterFunctions) {
+            List<ResponseConverterFunction> responseConverterFunctions,
+            List<ExceptionHandlerFunction> exceptionHandlerFunctions) {
         final List<Method> methods = requestMappingMethods(object);
         return methods.stream()
                       .flatMap((Method method) ->
-                                       create(pathPrefix, object, method, exceptionHandlerFunctions,
-                                              requestConverterFunctions, responseConverterFunctions).stream())
+                                       create(pathPrefix, object, method, requestConverterFunctions,
+                                              responseConverterFunctions, exceptionHandlerFunctions).stream())
                       .collect(toImmutableList());
     }
 
@@ -222,9 +222,9 @@ public final class AnnotatedHttpServiceFactory {
      */
     @VisibleForTesting
     static List<AnnotatedHttpServiceElement> create(String pathPrefix, Object object, Method method,
-                                                    List<ExceptionHandlerFunction> baseExceptionHandlers,
                                                     List<RequestConverterFunction> baseRequestConverters,
-                                                    List<ResponseConverterFunction> baseResponseConverters) {
+                                                    List<ResponseConverterFunction> baseResponseConverters,
+                                                    List<ExceptionHandlerFunction> baseExceptionHandlers) {
 
         final Set<Annotation> methodAnnotations = httpMethodAnnotations(method);
         if (methodAnnotations.isEmpty()) {
@@ -251,15 +251,15 @@ public final class AnnotatedHttpServiceFactory {
                                                 .build());
                 }).collect(toImmutableList());
 
-        final List<ExceptionHandlerFunction> eh =
-                getAnnotatedInstances(method, clazz, ExceptionHandler.class, ExceptionHandlerFunction.class)
-                        .addAll(baseExceptionHandlers).add(defaultExceptionHandler).build();
         final List<RequestConverterFunction> req =
                 getAnnotatedInstances(method, clazz, RequestConverter.class, RequestConverterFunction.class)
                         .addAll(baseRequestConverters).build();
         final List<ResponseConverterFunction> res =
                 getAnnotatedInstances(method, clazz, ResponseConverter.class, ResponseConverterFunction.class)
                         .addAll(baseResponseConverters).build();
+        final List<ExceptionHandlerFunction> eh =
+                getAnnotatedInstances(method, clazz, ExceptionHandler.class, ExceptionHandlerFunction.class)
+                        .addAll(baseExceptionHandlers).add(defaultExceptionHandler).build();
 
         final Optional<HttpStatus> defaultResponseStatus = findFirst(method, StatusCode.class)
                 .map(code -> {
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -169,15 +169,13 @@ public final class AnnotatedHttpServiceFactory {
             String pathPrefix, Object object,
             List<ExceptionHandlerFunction> exceptionHandlerFunctions,
             List<RequestConverterFunction> requestConverterFunctions,
-            List<ResponseConverterFunction> responseConverterFunctions,
-            AnnotatedHttpServiceExtensions extensions) {
+            List<ResponseConverterFunction> responseConverterFunctions) {
         final List<Method> methods = requestMappingMethods(object);
         return methods.stream()
                       .flatMap((Method method) ->
                                        create(pathPrefix, object, method, exceptionHandlerFunctions,
                                               requestConverterFunctions,
-                                              responseConverterFunctions,
-                                              extensions).stream())
+                                              responseConverterFunctions).stream())
                       .collect(toImmutableList());
     }
 
@@ -227,8 +225,7 @@ public final class AnnotatedHttpServiceFactory {
     static List<AnnotatedHttpServiceElement> create(String pathPrefix, Object object, Method method,
                                                     List<ExceptionHandlerFunction> baseExceptionHandlers,
                                                     List<RequestConverterFunction> baseRequestConverters,
-                                                    List<ResponseConverterFunction> baseResponseConverters,
-                                                    AnnotatedHttpServiceExtensions extensions) {
+                                                    List<ResponseConverterFunction> baseResponseConverters) {
 
         final Set<Annotation> methodAnnotations = httpMethodAnnotations(method);
         if (methodAnnotations.isEmpty()) {
@@ -257,14 +254,13 @@ public final class AnnotatedHttpServiceFactory {
 
         final List<ExceptionHandlerFunction> eh =
                 getAnnotatedInstances(method, clazz, ExceptionHandler.class, ExceptionHandlerFunction.class)
-                        .addAll(baseExceptionHandlers).addAll(extensions.exceptionHandlers())
-                        .add(defaultExceptionHandler).build();
+                        .addAll(baseExceptionHandlers).add(defaultExceptionHandler).build();
         final List<RequestConverterFunction> req =
                 getAnnotatedInstances(method, clazz, RequestConverter.class, RequestConverterFunction.class)
-                        .addAll(baseRequestConverters).addAll(extensions.requestConverters()).build();
+                        .addAll(baseRequestConverters).build();
         final List<ResponseConverterFunction> res =
                 getAnnotatedInstances(method, clazz, ResponseConverter.class, ResponseConverterFunction.class)
-                        .addAll(baseResponseConverters).addAll(extensions.responseConverters()).build();
+                        .addAll(baseResponseConverters).build();
 
         final Optional<HttpStatus> defaultResponseStatus = findFirst(method, StatusCode.class)
                 .map(code -> {
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -163,9 +163,8 @@ public final class AnnotatedHttpServiceFactory {
 
     /**
      * Returns the list of {@link AnnotatedHttpService} defined by {@link Path} and HTTP method annotations
-     * from the specified {@code object}, {@link ExceptionHandlerFunction}s,
-     * {@link RequestConverterFunction}s and {@link ResponseConverterFunction}s.
-     * FIXME(heowc): Update javadoc.
+     * from the specified {@code object}, {@link ExceptionHandlerFunction}s, {@link RequestConverterFunction}s,
+     * {@link ResponseConverterFunction}s and {@link AnnotatedHttpServiceExtensions}.
      */
     public static List<AnnotatedHttpServiceElement> find(
             String pathPrefix, Object object,
--- a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
@@ -174,8 +174,7 @@ public final class AnnotatedHttpServiceFactory {
         return methods.stream()
                       .flatMap((Method method) ->
                                        create(pathPrefix, object, method, exceptionHandlerFunctions,
-                                              requestConverterFunctions,
-                                              responseConverterFunctions).stream())
+                                              requestConverterFunctions, responseConverterFunctions).stream())
                       .collect(toImmutableList());
     }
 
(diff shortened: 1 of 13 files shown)


heowc:
This is the Draft.

I noticed it was different from the initial direction.
So, I have some questions.

1. I think we need clear priorities for `{ExceptionHandler/RequestConverter/ResponseConverter}Function`.
    - How about the following order? `global extension > annotation > method` (refer to [here](https://github.com/line/armeria/blob/3ff42dbeaff1d5ab823b2aa37ded6ab3e0278d0a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java#L254...L262))

2. how about creation time change for {VirtualHost}AnnotatedServiceBindingBuilder?
    -  I changed the position to apply  `{ExceptionHandler/RequestConverter/ResponseConverter}Function` of extentions.

Do you have another direction or opinion? Please tell me.

heowc:
This is the Draft.

I noticed it was different from the initial direction.
So, I have some questions. @trustin @minwoox @ikhoon 

1. I think we need clear priorities for `{ExceptionHandler/RequestConverter/ResponseConverter}Function`.
    - How about the following order? `global extension > annotation > method`
    - https://github.com/line/armeria/blob/3ff42dbeaff1d5ab823b2aa37ded6ab3e0278d0a/core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java#L254...L262

2. how about the creation time change for {VirtualHost}AnnotatedServiceBindingBuilder?
    -  I changed the position to apply  `{ExceptionHandler/RequestConverter/ResponseConverter}Function` of extentions.

3. Should I replace the already existing `{ExceptionHandler/RequestConverter/ResponseConverter}Function`s?
    - The implementation now is a way to increase the priority by adding in the future.

Do you have another direction or opinion? Please tell me.

trustin:
> How about the following order? `global extension > annotation > method`

The extensions must be evaluated in the following order:

1. The extensions specified as a method annotation.
2. The extensions specified as a class annotation.
3. The extensions specified when you build a route using `annotatedService(..., Iterable)`.
4. The extensions specified when you build a server or a virtual host using `annotatedServiceExtensions` :arrow_left: This is what you're doing in this PR.
5. Built-in extensions:
   - `AnnotatedValueResolver.defaultRequestConverters`
   - `AnnotatedHttpService.defaultResponseConverters`
   - `AnnotatedHttpServiceFactory.defaultExceptionHandler`

Therefore, you could simply insert the extensions specified with the `annotatedServiceExtensions()` builder method between (3) and (5).

> how about the creation time change for `{VirtualHost}AnnotatedServiceBindingBuilder`?

I think it's fine.

> Should I replace the already existing `{ExceptionHandler/RequestConverter/ResponseConverter}Functions`?

No, you have to keep them all to make sure the evaluation order is preserved.

trustin:
So.. I think you're going into the right direction. Please let us know if you are ready for a detailed review!

minwoox:
> how about the creation time change for {VirtualHost}AnnotatedServiceBindingBuilder?

I think we have to. 😄 Please keep going. 👍

heowc:
Thank you for the reply. 😄 @trustin @minwoox 

I am aware of the current direction and will continue further development. 🚀

trustin:
Please let us know once it's ready.

heowc:
> Please let us know once it's ready.

I'm ready! 👍

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1478,10 +1487,31 @@ public ServerBuilder contentPreview(int length) {
...
+     * and {@link ResponseConverterFunction} for creating a {@link AnnotatedHttpServiceExtensions}.
+     *
+     * @param exceptionHandlerFunctions the {@link ExceptionHandlerFunction}s
+     * @param requestConverterFunctions the {@link RequestConverterFunction}s
+     * @param responseConverterFunctions the {@link ResponseConverterFunction}s
+     */
+    public ServerBuilder annotatedHttpServiceExtensions(
+            Iterable<? extends ExceptionHandlerFunction> exceptionHandlerFunctions,
+            Iterable<? extends RequestConverterFunction> requestConverterFunctions,
+            Iterable<? extends ResponseConverterFunction> responseConverterFunctions) {
```
trustin: Please reorder parameters: `requestConverterFunctions`, `responseConverterFunctions` and `exceptionHandlerFunctions`.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1016,6 +1019,12 @@ ServerBuilder serviceConfigBuilder(ServiceConfigBuilder serviceConfigBuilder) {
         return this;
     }
 
+    ServerBuilder annotatedServiceBindingBuilder(
+            AnnotatedServiceBindingBuilder annotatedServiceBindingBuilder) {
+        annotatedServiceBindingBuilders.add(annotatedServiceBindingBuilder);
```
trustin: Shouldn't this be `virtualHostTemplate.addAnnotatedServiceBindingBuilders.add(...)`? `AnnotatedServiceBindingBuilder.applyToServiceConfigBuilder()` and `ServerBuilder.annotatedServiceBindingBuilder` could be then removed?

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceHandlersOrderTest.java
```diff
@@ -247,12 +286,12 @@ public void responseConverterOrder() throws Exception {
         // Converted from the ServiceLevelResponseConverter.
         assertThat(aRes.contentUtf8()).isEqualTo("hello foo");
 
-        // method level(+1) -> class level(+1) -> service level(+1)
+        // method level(+1) -> class level(+1) -> service level(+1) -> server level(+1)
         assertThat(responseCounter.get()).isEqualTo(3);
```
trustin: Shouldn't this be `isEquals(4)`, then?

core/src/test/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilderTest.java
```diff
@@ -58,7 +81,7 @@ void testWhenPathPrefixIsNotGivenThenUsesDefault() {
                                     .build(new TestService())
                                     .build();
 
-        assertThat(server.config().serviceConfigs()).hasSize(1);
+        assertThat(server.config().serviceConfigs()).hasSize(2);
```
trustin: Could you also add the assertions for the second service?

core/src/test/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilderTest.java
```diff
@@ -73,7 +96,7 @@ void testWhenPathPrefixIsGivenThenItIsPrefixed() {
...
```
trustin: Could you also add the assertions for the second service?

core/src/test/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilderTest.java
```diff
@@ -99,7 +122,7 @@ void testAllConfigurationsAreRespected() {
...
```
trustin: Could you also add the assertions for the second service?

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -259,18 +261,31 @@ public AnnotatedServiceBindingBuilder accessLogWriter(AccessLogWriter accessLogW
...
      */
     public ServerBuilder build(Object service) {
+        requireNonNull(service, "service");
+        this.service = service;
+        serverBuilder.annotatedServiceBindingBuilder(this);
+        return serverBuilder;
+    }
+
+    /**
+     * Apples the {@link ServiceConfigBuilder} created with the configured
```
minwoox: nit: Apples? 😆

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1478,10 +1488,32 @@ public ServerBuilder contentPreview(int length) {
         return contentPreview(length, ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET);
     }
 
+    /**
+     * Sets the {@link RequestConverterFunction}s, {@link ResponseConverterFunction}
+     * and {@link ExceptionHandlerFunction}s for creating a {@link AnnotatedHttpServiceExtensions}.
```
minwoox: nit: a -> an

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1478,10 +1488,32 @@ public ServerBuilder contentPreview(int length) {
...
+        return this;
+    }
+
     /**
      * Returns a newly-created {@link Server} based on the configuration properties set so far.
      */
     public Server build() {
+        final AnnotatedHttpServiceExtensions extensions =
+                virtualHostTemplate.getAnnotatedHttpServiceExtensions();
+        annotatedServiceBindingBuilders.forEach(builder -> builder.applyToServiceConfigBuilder(extensions));
```
minwoox: nit: need `assert extensions != null;`

core/src/main/java/com/linecorp/armeria/server/VirtualHostAnnotatedServiceBindingBuilder.java
```diff
@@ -275,11 +277,24 @@ public VirtualHostAnnotatedServiceBindingBuilder decorator(
      * @return {@link VirtualHostBuilder} to continue building {@link VirtualHost}
      */
     public VirtualHostBuilder build(Object service) {
+        this.service = service;
```
minwoox: nit: `requireNonNull(service, "service")`

core/src/main/java/com/linecorp/armeria/server/VirtualHostAnnotatedServiceBindingBuilder.java
```diff
@@ -275,11 +277,24 @@ public VirtualHostAnnotatedServiceBindingBuilder decorator(
...
+        virtualHostBuilder.addAnnotatedServiceBindingBuilder(this);
+        return virtualHostBuilder;
+    }
+
+    /**
+     * Apples the {@link ServiceConfigBuilder} created with the configured
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -839,6 +850,33 @@ public VirtualHostBuilder accessLogWriter(AccessLogWriter accessLogWriter, boole
         return this;
     }
 
+    /**
+     * Sets the {@link ExceptionHandlerFunction}s, {@link RequestConverterFunction}s
+     * and {@link ResponseConverterFunction} for creating a {@link AnnotatedHttpServiceExtensions}.
```
minwoox: nit: a -> an

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -892,12 +930,19 @@ VirtualHost build(VirtualHostBuilder template) {
...
 
+        final AnnotatedHttpServiceExtensions annotatedHttpServiceExtensions =
+                this.annotatedHttpServiceExtensions != null ?
+                this.annotatedHttpServiceExtensions : template.annotatedHttpServiceExtensions;
+
         assert requestContentPreviewerFactory != null;
         assert responseContentPreviewerFactory != null;
         assert rejectedRouteHandler != null;
         assert accessLogWriter != null;
         assert accessLoggerMapper != null;
```
minwoox: Let's add `assert annotatedHttpServiceExtensions != null` as well. 😉

core/src/main/java/com/linecorp/armeria/server/VirtualHostAnnotatedServiceBindingBuilder.java
```diff
@@ -275,11 +277,24 @@ public VirtualHostAnnotatedServiceBindingBuilder decorator(
...
+     * {@link AnnotatedHttpServiceExtensions} to the {@link VirtualHostBuilder}.
+     *
+     * @param extensions the {@link AnnotatedHttpServiceExtensions} at the virtual host level.
+     */
+    VirtualHostBuilder applyToServiceConfigBuilder(AnnotatedHttpServiceExtensions extensions) {
```
minwoox: This method does not have to return `VritualHostBuilder` as we do in `AnnotatedServiceBindingBuilder`.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
heowc: > Shouldn't this be virtualHostTemplate.addAnnotatedServiceBindingBuilders.add(...)?

`ServerBuilder.annotatedServiceBindingBuilder` and `VirtualHostBuilder.addAnnotatedServiceBindingBuilders` add different classes.

> AnnotatedServiceBindingBuilder.applyToServiceConfigBuilder() and ServerBuilder.annotatedServiceBindingBuilder could be then removed?

I changed the time of creation because I had to create a `ServiceConfigBuilder` including the `extensions`. So I think this is necessary.

Do you have any other opinions?

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
heowc: Oh! my mistake...

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
```diff
@@ -162,19 +162,22 @@
...
+     * from the specified {@code object}, {@link ExceptionHandlerFunction}s, {@link RequestConverterFunction}s,
+     * {@link ResponseConverterFunction}s and {@link AnnotatedHttpServiceExtensions}.
      */
     public static List<AnnotatedHttpServiceElement> find(
             String pathPrefix, Object object,
             List<ExceptionHandlerFunction> exceptionHandlerFunctions,
             List<RequestConverterFunction> requestConverterFunctions,
-            List<ResponseConverterFunction> responseConverterFunctions) {
+            List<ResponseConverterFunction> responseConverterFunctions,
+            AnnotatedHttpServiceExtensions extensions) {
```
ikhoon: I think `AnnotatedHttpServiceExtensions` structure is similar to `exceptionHandlerFunctions` + `requestConverterFunctions` + `responseConverterFunctions`. Shouldn't we accept one of them? 

How about merging `extensions.XXXFunctions` into `XXXFunctions` before calling this method.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
ikhoon: I think `virtualHostTemplate.getAnnotatedHttpServiceExtensions()` could be nullable. 🤔
https://github.com/line/armeria/blob/0b3e3fee81d7268922cb4a442e6d2571b55e1f6d/core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java#L154-L155

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
heowc: I also worried about this for a while. If you think so, I will correct it. 😃

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
trustin: Ah, OK, then. :bowing_man:

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceFactory.java
trustin: +1

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
heowc: Oh, I don't remember why I put it over there 😅

I'll fix it soon.

minwoox: This is somewhat unclear to me how it works because the `annotatedServiceBindingBuider` has the reference of the `serverBuilder` and call `addServiceConfigBuilder` inside in it.
How about just returning the list of `ServiceConfigBuilder` and add it here?
```java
List<ServiceConfigBuilder> toServiceConfigBuilders(AnnotatedHttpServiceExtensions extensions) {
    ...
    return elements.stream()
                   .map(element -> {
                       final HttpService decoratedService =
                               element.buildSafeDecoratedService(defaultServiceConfigSetters.decorator());
                       return defaultServiceConfigSetters.toServiceConfigBuilder(element.route(), decoratedService);
                   }).collect(toImmutableList());
}

...
// In ServerBuilder.build()
annotatedServiceBindingBuilders.forEach(annotatedBuilder -> {
    annotatedBuilder.toServiceConfigBuilders(extensions).forEach(
            virtualHostTemplate::addServiceConfigBuilder);
});
```

core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
```diff
@@ -259,18 +261,35 @@ public AnnotatedServiceBindingBuilder accessLogWriter(AccessLogWriter accessLogW
...
+                                                 requestConverterFunctions, responseConverterFunctions);
         elements.forEach(element -> {
             final HttpService decoratedService =
                     element.buildSafeDecoratedService(defaultServiceConfigSetters.decorator());
             final ServiceConfigBuilder serviceConfigBuilder =
                     defaultServiceConfigSetters.toServiceConfigBuilder(element.route(), decoratedService);
             serverBuilder.serviceConfigBuilder(serviceConfigBuilder);
         });
-        return serverBuilder;
     }
```
ikhoon: I think 1) this method call flow is complicated and 2) the role of this method is a little ambiguous.

1) The `build(Object service)` returns the `serverBulider` and it call this method and this method call `serverBuilder.serviceConfigBuilder()` again.
2) The method name is `applyToServiceConfigBuilder `, but creates `serviceConfigBuilder` and sets extension to `serviceConfigBuilder`

How about initiating `ServiceConfigBuilder` return it and change the method name to `buildServiceConfigBuilder`.

```diff
diff --git a/core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java b/core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
index a64e33fc9..34debe75c 100644
--- a/core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/AnnotatedServiceBindingBuilder.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.server;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.requireNonNull;
 
 import java.nio.charset.Charset;
@@ -273,7 +274,7 @@ public final class AnnotatedServiceBindingBuilder implements ServiceConfigSetter
      *
      * @param extensions the {@link AnnotatedHttpServiceExtensions} at the server level.
      */
-    void applyToServiceConfigBuilder(AnnotatedHttpServiceExtensions extensions) {
+    List<ServiceConfigBuilder> buildServiceConfigBuilder(AnnotatedHttpServiceExtensions extensions) {
         final List<ExceptionHandlerFunction> exceptionHandlerFunctions =
                 exceptionHandlerFunctionBuilder.addAll(extensions.exceptionHandlers()).build();
         final List<RequestConverterFunction> requestConverterFunctions =
@@ -284,12 +285,11 @@ public final class AnnotatedServiceBindingBuilder implements ServiceConfigSetter
         final List<AnnotatedHttpServiceElement> elements =
                 AnnotatedHttpServiceFactory.find(pathPrefix, service, exceptionHandlerFunctions,
                                                  requestConverterFunctions, responseConverterFunctions);
-        elements.forEach(element -> {
+
+        return elements.stream().map(element -> {
             final HttpService decoratedService =
                     element.buildSafeDecoratedService(defaultServiceConfigSetters.decorator());
-            final ServiceConfigBuilder serviceConfigBuilder =
-                    defaultServiceConfigSetters.toServiceConfigBuilder(element.route(), decoratedService);
-            serverBuilder.serviceConfigBuilder(serviceConfigBuilder);
-        });
+            return defaultServiceConfigSetters.toServiceConfigBuilder(element.route(), decoratedService);
+        }).collect(toImmutableList());
     }
 }
diff --git a/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java b/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
index fabfefbaa..7c41e60dd 100644
--- a/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
@@ -1512,7 +1512,11 @@ public final class ServerBuilder {
     public Server build() {
         final AnnotatedHttpServiceExtensions extensions =
                 virtualHostTemplate.getAnnotatedHttpServiceExtensions();
-        annotatedServiceBindingBuilders.forEach(builder -> builder.applyToServiceConfigBuilder(extensions));
+        for (AnnotatedServiceBindingBuilder bindingBuilder : annotatedServiceBindingBuilders) {
+            for (ServiceConfigBuilder configBuilder : bindingBuilder.buildServiceConfigBuilder(extensions)) {
+                serviceConfigBuilder(configBuilder);
+            }
+        }
 
         assert extensions != null;
 
```

heowc: @ikhoon 
good idea. What about the following rather than nested loops?

```java
annotatedServiceBindingBuilders.stream()
                               .flatMap(b -> b.buildServiceConfigBuilder(extensions).stream())
                               .forEach(this::serviceConfigBuilder)
```

ikhoon: Either is fine to me. 😀 Because this is not a hot path, executed once for build server.

core/src/test/java/com/linecorp/armeria/internal/annotation/AnnotatedHttpServiceHandlersOrderTest.java
```diff
@@ -22,10 +22,11 @@
 
 import javax.annotation.Nullable;
 
-import org.junit.ClassRule;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
```
ikhoon: 👍

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1478,10 +1488,37 @@ public ServerBuilder contentPreview(int length) {
...
+     * and {@link ExceptionHandlerFunction}s for creating an {@link AnnotatedHttpServiceExtensions}.
+     *
+     * @param requestConverterFunctions the {@link RequestConverterFunction}s
+     * @param responseConverterFunctions the {@link ResponseConverterFunction}s
+     * @param exceptionHandlerFunctions the {@link ExceptionHandlerFunction}s
+     */
+    public ServerBuilder annotatedHttpServiceExtensions(
+            Iterable<? extends RequestConverterFunction> requestConverterFunctions,
+            Iterable<? extends ResponseConverterFunction> responseConverterFunctions,
+            Iterable<? extends ExceptionHandlerFunction> exceptionHandlerFunctions) {
```
ikhoon: nit: Could you match the order of arguments of this method and  related methods with `AnnotatedHttpServiceExtensions` constructor’s arguments. 

Either way is fine if it does not make breaking change.

heowc: Ok!

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -892,11 +930,21 @@ VirtualHost build(VirtualHostBuilder template) {
...
+        assert annotatedHttpServiceExtensions != null;
+
+        virtualHostAnnotatedServiceBindingBuilders.stream()
+                                       .flatMap(b -> b.buildServiceConfigBuilder(annotatedHttpServiceExtensions)
```
minwoox: nit: indentation?

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -207,6 +209,8 @@ public ServerBuilder() {
         virtualHostTemplate.accessLogger(
                     host -> LoggerFactory.getLogger(defaultAccessLoggerName(host.hostnamePattern())));
         virtualHostTemplate.tlsSelfSigned(false);
+        virtualHostTemplate.annotatedHttpServiceExtensions(ImmutableList.of(), ImmutableList.of(),
```
minwoox: How about not setting `annotatedHttpServiceExtensions` here so that we do not call `buildServiceConfigBuilder` in `ServerBuilder.build()` when the `extenstions` is null?

heowc: `buildServiceConfigBuilder` should always be called. So I set a default value to reduce the complexity of `buildServiceConfigBuilder`.

Am I right about your question?

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
heowc: Fixed

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
minwoox: Ah, yes. I misread it. So it's okay as it is. 😉

Issue #2173, opened 2019-10-08
`RequestContextAwareCompletableFuture` callback not invoked when context mismatches.


trustin:
Let's say we have the following code:

```java
ServiceRequestContext ctx1 = ServiceRequestContext.of(...);
ServiceRequestContext ctx2 = ServiceRequestContext.of(...);
try (SafeCloseable sc1 = ctx1.push()) {
    CompletableFuture f = ctx2.makeContextAware(new CompletableFuture());
    AtomicBoolean callbackCalled = new AtomicBoolean();
    f.whenComplete((res, cause) -> callbackCalled.set(true));

    f.complete(null);    

    // The callback will never be invoked and
    // the following assertion will fail.
    assert callbackCalled.get(); // <-- Failure!
}
```

It seems like everything is working as expected. The callback (`BiConsumer`) was not invoked because it was decorated by `ctx2.makeContextAware(BiConsumer)` and the callback is invoked only when `ctx2.pushIfAbsent()` succeeds ([see](https://github.com/line/armeria/blob/armeria-0.94.0/core/src/main/java/com/linecorp/armeria/common/AbstractRequestContext.java#L138)), to protect our users from the accidental context conflict like the above example.

The problem is that a user does not have a way to know this happened. It is really confusing because the callback will *never* be invoked. It is neither possible to handle the `IllegalStateException` raised by `pushIfAbsent()`:

```java
f.whenComplete((res, cause) -> callbackCalled.set(true))
 .exceptionally(cause -> {
     // We never reach here, either, because this callback
     // will be decorated again and fail at `pushIfAbsent()` again.
     ...
 });
```

What would be the best way to let a user know about this problem?

trustin:
How about logging at WARN level than raising an exception for the future callbacks? If a user wraps something explicitly via `ctx.makeContextAware()`, we could raise an `IllegalStateException` as before, though, i.e.

```java
class AbstractRequestContext ... {
    // Stay same.
    BiConsumer makeContextAware(BiConsumer consumer) {
        return (t, u) -> {
            try (SafeCloseable ignored = pushIfAbsent()) {
                action.accept(t, u);
            }
        };
    }
}

class RequestContextAwareCompletableFuture ... {
    @Override
    public CompletableFuture<T> whenComplete(BiConsumer<? super T, ? super Throwable> action) {
        return ctx.makeContextAware(super.whenComplete(laxMakeContextAware(action)));
    }

[...]

PR #2322, opened 2019-12-13
Provide our own attribute mapping APIs for RequestContext

### LINKED ISSUE #1083: A user may get the wrong RequestContext when contexts are nested

minwoox:
Motivation:
- `Attribute` in Netty has deprecated APIs which causes a race condition. We don't have to implement it but just have our own APIs for attribute mapping.
- When a user writes a service that invokes a request using an Armeria client, the client cannot access the `ServiceRequestContext`.

Modifications
- `RequestContext` does not extend `AttributeMap` in Netty anymore.
  - It has now its own methods to retreive and set the value:
    - `T attr(AttributeKey<T> key)`
    - `void setAttr(AttributeKey<T> key, @Nullable T value)`
    - `T setAttrIfAbsent(AttributeKey<T> key, T value)`
    - `T computeAttrIfAbsent(AttributeKey<T> key, T value)`
    - `Iterator<Entry<AttributeKey<?>, Object>> attrs()`
    - `T ownAttr(AttributeKey<T> key)` (in `ClientRequestContext` only)
    - `Iterator<Entry<AttributeKey<?>, Object>> ownAttrs()` (in `ClientRequestContext` only)
- Add `ClientRequestContext.rootContext()`
  - The `rootContext` is from the service that invokes a client request which created the `ClientRequestContext`.
  - When `attr()` is called, the `ClientRequestContext` looks for the `rootContext` if it doesn't have the value.
  - `attrs()` returns the combined iterator of `ClientRequestContext` and `rootContext`.

Result:
- (Breaking) `RequestContext` does not extend `AttributeMap` anymore.
  ```java
  RequestContext ctx = ...
  AttributeKey<Integer> KEY = AttributeKey.valueOf(MyClass.class, "key");

  // before
  Attribute<Integer> attr = ctx.attr(KEY);
  attr.set(10);
  assert attr.get() == 10;

  // after
  ctx.setAttr(KEY, 11);
  assert ctx.attr(KEY) == 10;
  ```
- (Breaking) `RequestLog` does not extend `AttributeMap` anymore. You should use context:
  ```java
  RquestLog log = ...
  log.context().setAttr(...)
  ```
- You can get the `ServiceRequestContext` of a service that invokes the client's request and its attributes without copying.
- Close #1083


minwoox:
I have two issues to solve:
- The `ServiceRequestContext` is inserted when a `ClientRequestContext` is created if the thread has the `ServiceRequestContext` in its thread-local.
  - It's because I wanted to make the field final so we don't have to worry much about thread-safety.
  - However, this will make the `ClientRequestContext` has its `ServiceRequestContext` forever when even the context is popped out.
- The `Attribute` returned from the `ClientRequestContext.attr()` can change the value of the `Attribute` in `ServiceRequestcontext`.
  - I'm not sure if we need to make only gettable.

trustin:
> I'm wondering if we need to make it only gettable.

Perhaps we should set the attribute at `ClientRequestContext`? We will have to change `AttributMap` implementation a little bit for this though.

core/src/main/java/com/linecorp/armeria/client/AttributeMapMaybeWithParent.java
```diff
@@ -0,0 +1,46 @@
...
+
+import io.netty.util.Attribute;
+import io.netty.util.AttributeKey;
+import io.netty.util.AttributeMap;
+
+/**
+ * Holds {@link Attribute}s which can be accessed via {@link AttributeKey}.
+ */
+// TODO(minwoox) Better name for this?
+public interface AttributeMapMaybeWithParent extends AttributeMap {
```
trustin: Time to fork `AttributeMap` interface? I guess this breaking change will affect few users.

core/src/main/java/com/linecorp/armeria/client/AttributeMapMaybeWithParent.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    Iterator<Attribute<?>> attrs();
+
+    /**
+     * Returns the {@link Iterator} of all {@link Attribute}s this map contains.
+     *
+     * @param withAttrsInParent tells whether to iterate the parent's {@link Attribute}s after iterating this
+     *                          {@link AttributeMapMaybeWithParent}'s {@link Attribute}s
+     */
+    Iterator<Attribute<?>> attrs(boolean withAttrsInParent);
```
trustin: How about `ownAttrs()`? instead of using a flag? (Named after JavaScript `hasOwnPropertyNames()`)

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +164,43 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+    Iterator<Attribute<?>> attrs();
+
+    /**
+     * Returns the {@link Iterator} of all {@link Attribute}s set in this context.
+     *
+     * @param withAttrsInServiceContexts tells whether to iterate the {@link #serviceRequestContext()}'s
+     *                                   {@link Attribute}s after iterating this {@link ClientRequestContext}'s
+     *                                   {@link Attribute}s
+     */
+    Iterator<Attribute<?>> attrs(boolean withAttrsInServiceContexts);
```
trustin: ditto `ownAttrs()`?

core/src/main/java/com/linecorp/armeria/client/AttributeMapMaybeWithParent.java
minwoox: That's a good idea. So the `AttribuiteMap` will be:
```java
public interface AttributeMap {

    <T> Attribute<T> attr(AttributeKey<T> key);
    <T> Attribute<T> ownAttr(AttributeKey<T> key);
    <T> boolean hasAttr(AttributeKey<T> key);
    <T> boolean hasOwnAttr(AttributeKey<T> key);
    Iterator<Attribute<?>> attrs();
    Iterator<Attribute<?>> ownAttrs();
}
```
Is that right?

trustin: On second thought, how about adding the `own` methods to `DefaultAttributeMap` and `ClientRequestContext` only? No need to fork anything and introduce a new API.

minwoox: `ClientRequestContext` has to call `ServiceRequestContext.defaultAttribuiteMap()`. But `DefaultAttributeMap` is internal API so I can't add it to the public API because it violates the :site:build. That's why I introduce this new interface.

minwoox:
This is ready. 😉

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -83,7 +67,7 @@ public static void appendHttpStatusTag(ImmutableList.Builder<Tag> tagListBuilder
         tagListBuilder.add(Tag.of("httpStatus", status.codeAsText()));
     }
 
-    private static void onRequest(RequestLog log, MeterIdPrefixFunction meterIdPrefixFunction, boolean server) {
+    public static void onRequest(RequestLog log, MeterIdPrefixFunction meterIdPrefixFunction, boolean server) {
```
ikhoon: Add Javadoc?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,34 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
         return new ClientRequestContextBuilder(request, uri);
     }
 
+    /**
+     * Returns the {@link ServiceRequestContext} whose {@link Service} invokes the {@link Client}
+     * {@link Request} which created this {@link ClientRequestContext}.
```
anuraaga: `, or {@code null} if this client request was not made in the context of a server request.`

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,34 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     */
+    @Nullable
+    ServiceRequestContext serviceRequestContext();
+
+    /**
+     * {@inheritDoc} If {@link #serviceRequestContext()} is not {@code null} and the
+     * {@link ServiceRequestContext} has the {@link Attribute}, this will return the {@link Attribute}.
+     */
+    @Override
```
anuraaga: `@Nullable` I think

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,34 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     * {@link ServiceRequestContext} has the {@link Attribute}, this will return {@code true}.
+     */
+    @Override
+    <T> boolean hasAttr(AttributeKey<T> key);
+
+    /**
+     * {@inheritDoc} If {@link #serviceRequestContext()} is not {@code null}, then the {@link Iterator} will
+     * iterate the {@link ServiceRequestContext}'s {@link #attrs()}} afterward.
+     */
+    @Override
```
anuraaga: Ditto

core/src/main/java/com/linecorp/armeria/client/metric/AbstractMetricCollectingClient.java
```diff
@@ -42,7 +48,12 @@
 
     @Override
     public O execute(ClientRequestContext ctx, I req) throws Exception {
-        RequestMetricSupport.setup(ctx, meterIdPrefixFunction, false);
+        if (!ctx.hasAttr(ATTR_REQUEST_METRICS_SET)) {
```
anuraaga: This is fine too but I guess you could keep the helper and pass in `AttributeKey`.

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -316,4 +371,27 @@ public boolean hasNext() {
             }
         }
     }
+
+    private static final class CombinedIteratorImpl implements Iterator<Attribute<?>> {
```
anuraaga: Can use `Iterators.concat` from guava too

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: I think it's not nullable because we always make an empty `DefaultAttribute`

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: ditto `Collections.emptyIterator()` when empty.

core/src/main/java/com/linecorp/armeria/client/metric/AbstractMetricCollectingClient.java
minwoox: Let me revert it. 😄

core/src/main/java/com/linecorp/armeria/common/AttributeMap.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.Iterator;
+
+import io.netty.util.Attribute;
+import io.netty.util.AttributeKey;
+
+/**
+ * Holds {@link Attribute}s which can be accessed via {@link AttributeKey}.
+ */
+public interface AttributeMap extends io.netty.util.AttributeMap {
```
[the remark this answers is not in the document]
minwoox: Thanks!

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
minwoox: Thanks!

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
minwoox: I thought it was okay because it's in the internal package. but let me revert this anyway

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Thanks!

minwoox:
All fixed. PTAL 🙇

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -155,6 +172,24 @@ private DefaultClientRequestContext(
...
+            return (DefaultServiceRequestContext) current;
+        }
+        return null;
+    }
+
+    @Nullable
+    private static AttributeMap serviceContextAttributeMap(
+            @Nullable DefaultServiceRequestContext serviceContext) {
+        if (serviceContext != null) {
+            return serviceContext.attributeMap();
```
trustin: Isn't `RequestContext` an `AttributeMap`? I think we can use `ServiceRequestContext` instead of `DefaultServiceRequestContext` in this class.

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -191,19 +193,35 @@ public final MeterRegistry meterRegistry() {
         return meterRegistry;
     }
 
+    /**
+     * Returns {@link AttributeMap}.
+     */
+    public AttributeMap attributeMap() {
+        return attributeMap;
```
trustin: Not sure why we need this method. Doesn't this class already implement `AttributeMap` and delegate to `attributeMap`?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -155,19 +181,37 @@
...
+            return hasAttrInParent(key);
         }
     }
 
-    /**
-     * Returns all {@link Attribute}s this map contains.
-     */
+    private <T> boolean hasAttrInParent(AttributeKey<T> key) {
+        if (parentMap != null) {
+            // This happens at least only once, because the parentMap does not have its parent.
```
trustin: happens once at most?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -90,14 +111,19 @@
             }
         }
 
-        int i = index(key);
+        final int i = index(key);
         DefaultAttribute<?> head = attributes.get(i);
         if (head == null) {
+            if (parentMap != null && parentMap.hasAttr(key)) {
+                // This happens at least only once, because the parentMap does not have its parent.
```
trustin: happens once at most?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -70,18 +74,35 @@
...
     @Override
     public <T> Attribute<T> attr(AttributeKey<T> key) {
         if (key == null) {
             throw new NullPointerException("key");
         }
+
         AtomicReferenceArray<DefaultAttribute<?>> attributes = this.attributes;
         if (attributes == null) {
+            if (parentMap != null && parentMap.hasAttr(key)) {
+                // This happens at least only once, because the parentMap does not have its parent.
```
trustin: happens once at most?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -155,19 +181,37 @@
...
+        }
+        return false;
+    }
+
+    @Override
     public Iterator<Attribute<?>> attrs() {
+        return attrs(false);
+    }
+
+    public Iterator<Attribute<?>> attrs(boolean ownAttrs) {
```
trustin: Isn't it simpler to have two different implementations rather than using a boolean flag?

```java
public Iterator<Attribute<?>> attrs() {
    final AtomicReferenceArray<DefaultAttribute<?>> attributes = this.attributes;
    if (attributes == null) {
        return parentMap != null ? parentMap.attrs() : Collections.emptyIterator();
    }
    final Iterator<Attribute<?>> ownAttrsIt = new IteratorImpl(attributes);
    return parentMap != null ? ... : ownAttrsIt;
}

public Iterator<Attribute<?>> ownAttrs() {
    final AtomicReferenceArray<DefaultAttribute<?>> attributes = this.attributes;
    if (attributes == null) {
        return Collections.emptyIterator();
    }
    return new IteratorImpl(attributes);
}
```

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -155,19 +181,37 @@
...
         if (attributes == null) {
+            if (!ownAttrs && parentMap != null) {
+                // This happens at least only once, because the parentMap does not have its parent.
+                return parentMap.attrs();
+            }
             return Collections.emptyIterator();
         }
 
+        if (!ownAttrs && parentMap != null) {
+            return Iterators.concat(new IteratorImpl(attributes), parentMap.attrs());
```
trustin: You can't just concat. You have to skip the parent attributes if they exist in own attributes.

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -178,10 +222,13 @@ private static int index(AttributeKey<?> key) {
     @Override
     public String toString() {
         final ToStringHelper helper = MoreObjects.toStringHelper("");
-        for (Iterator<Attribute<?>> i = attrs(); i.hasNext();) {
+        for (final Iterator<Attribute<?>> i = attrs(); i.hasNext();) {
             final Attribute<?> a = i.next();
             helper.add(a.key().name(), a.get());
         }
+        if (parentMap != null) {
+            helper.add(", parent: ", parentMap);
```
trustin: How about skipping the attributes that exist in own attrs?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
minwoox: Couldn't think of it. Thanks!

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
minwoox: Oops I made big mistake. Just revert this because it's using `attrs()` which already implemented the feature.

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
minwoox: Looks much better, thanks!

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
minwoox: Yes, I didn't have to. 😄 Let me just use `ServiceRequestContext`, thanks!

minwoox:
@trustin All fixed. 🙇

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -70,18 +73,35 @@
     private static final int BUCKET_SIZE = 8;
     private static final int MASK = BUCKET_SIZE  - 1;
 
+    @Nullable
+    private final AttributeMap parentMap;
```
anuraaga: I'm trying to think of a word other than `parent` - the word almost always implies a chain of descendants, not just one as we have here (and have to explicitly state in comments). `root` is much more explicit and maybe clear enough? Or maybe `owner`, `creator`.

Not a big deal but want to bring up that `parent` here makes things harder to reason about.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
```diff
@@ -67,11 +65,6 @@
  */
 public interface RequestLog extends AttributeMap {
```
anuraaga: A thought for another PR - I wonder if it would simplify things to make `RequestLog` not be `AttributeMap`. Since `RequestLog.context()` always gives an attribute map, I'm confused between the relationship, and I guess it could provide some improved efficiency to not have two attribute maps that essentially represent a single concept.

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
minwoox: `root` sounds good. Thanks!

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
minwoox: Just a shortcut for `log.context().attr()`? Hm, 🤔 not sure which one is best. 🤣

trustin: Yeah, they are just a shortcut. They refer to the same map.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,35 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     * Returns the {@link ServiceRequestContext} whose {@link Service} invokes the {@link Client}
+     * {@link Request} which created this {@link ClientRequestContext}, or {@code null} if this client request
+     * was not made in the context of a server request.
+     */
+    @Nullable
+    ServiceRequestContext serviceRequestContext();
+
+    /**
+     * {@inheritDoc} If {@link #serviceRequestContext()} is not {@code null} and the
+     * {@link ServiceRequestContext} has the {@link Attribute}, this will return the {@link Attribute}.
```
trustin: I think you also need to say 'if the attribute is not in this context'? Perhaps better rephrasing to express the exact condition.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -81,6 +83,8 @@
     private Endpoint endpoint;
     @Nullable
     private final String fragment;
+    @Nullable
+    private final ServiceRequestContext serviceContext;
```
trustin: Hmm... `root` or `rootContext` like @anuraaga suggested? Feel free to ignore if you don't like it. :laughing:

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
anuraaga: Ah didn't notice it's just a shortcut :) I think it might be better not to extend `AttributeMap` and if the methods seem useful (quite a small shortcut...) define them directly with that explanation. I expected something that implements `AttributeMap` to have a map of attributes itself

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -75,9 +75,11 @@
     protected NonWrappingRequestContext(
             MeterRegistry meterRegistry, SessionProtocol sessionProtocol,
             RequestId id, HttpMethod method, String path, @Nullable String query,
-            @Nullable HttpRequest req, @Nullable RpcRequest rpcReq) {
+            @Nullable HttpRequest req, @Nullable RpcRequest rpcReq,
+            @Nullable AttributeMap parentAttributeMap) {
```
trustin: `rootAttributeMap`?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -70,18 +73,34 @@
...
     @Override
     public <T> Attribute<T> attr(AttributeKey<T> key) {
         if (key == null) {
             throw new NullPointerException("key");
         }
+
         AtomicReferenceArray<DefaultAttribute<?>> attributes = this.attributes;
         if (attributes == null) {
+            if (root != null && root.hasAttr(key)) {
+                return root.attr(key);
```
trustin: How about copying the attribute here, so that any attribute changes do not affect the root? e.g.

```java
DefaultAttributeMap root = new DefaultAttributeMap(null);
root.attr(FOO).set("bar");
DefaultAttributeMap map = new DefaultAttributeMap(map);
map.attr(FOO).set("baz");
assert root.attr(FOO).get()).equals("bar");
assert map.attr(FOO).get()).equals("baz");
```

Alternatively, we could return an `Attribute` wrapper whose `set()` will update `map` instead of `root`, but maybe that's more complex?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -90,14 +109,18 @@
             }
         }
 
-        int i = index(key);
+        final int i = index(key);
         DefaultAttribute<?> head = attributes.get(i);
         if (head == null) {
+            if (root != null && root.hasAttr(key)) {
+                return root.attr(key);
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -155,19 +178,34 @@
...
+        }
+        return false;
+    }
+
+    @Override
     public Iterator<Attribute<?>> attrs() {
         final AtomicReferenceArray<DefaultAttribute<?>> attributes = this.attributes;
         if (attributes == null) {
-            return Collections.emptyIterator();
+            return root != null ? root.attrs() : Collections.emptyIterator();
```
trustin: I just realized we need to copy on demand or copy all attributes here because otherwise the attribute changes will affect the root.

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -155,19 +178,34 @@
...
         }
 
+        final Iterator<Attribute<?>> ownAttrsIt = new IteratorImpl(attributes);
+        return root != null ? new ConcatenatedIteratorImpl(ownAttrsIt, root.attrs(), this)
```
trustin: - Same here. When `root.attrs()` is involved, we need to make sure any changes do not go to the root but to `this`.
- Instead of passing `this`, you could just make `ConcatenatedIteratorImpl` non-static.
- Maybe not worth, but we may want to invoke `root.attrs()` only after the iteration of `ownAttrsIt` is over.

core/src/main/java/com/linecorp/armeria/common/AttributeMap.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Holds {@link Attribute}s which can be accessed via {@link AttributeKey}.
+ */
+@SuppressWarnings("ClassNameSameAsAncestorName")
+public interface AttributeMap extends io.netty.util.AttributeMap {
+
+    /**
+     * Returns the {@link Iterator} of all {@link Attribute}s this map contains.
+     */
+    Iterator<Attribute<?>> attrs();
```
anuraaga: Should we add `AttributeMap ownAttrs()` so users can access attributes without the fallback to root? For some use cases I think this would be needed.

anuraaga: Unfortunately, this proposed method should return `AttributeMap` I think, which then creates a typing conflict with `attrs()` which returns `Iterator`

minwoox:
Yes, let me wrap the `Attribute` from the `root` using `attr()` and `attrs()`, so that a new value for the `Attribute` is set only to the current map, not the root. Thanks!

core/src/main/java/com/linecorp/armeria/common/AttributeMap.java
trustin: We could instead add `ownAttrs()`, `hasOwnAttrs()` and `ownAttr()` to `ClientRequestContext`?

anuraaga: Yup that works well too

minwoox: Will add those three methods to `ClientRequestContext`, thanks!

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
minwoox: Let me use `rootContext`. It's probably better. 😄

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
minwoox: All fixed except the last one. If we delay it we can:
- save one `findNext()` call
- `rootIt` can have attributes instead of being an `emptyIterator()` if there was an insertion before the iteration of `ownAttrsIt` is over. (which is rare)

I'm not sure if these have the advantage over having the final `rootIt` 🤔

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Added Thanks!

minwoox:
Let me squash commits to resolve conflicting. 🙏

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,82 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
         return new ClientRequestContextBuilder(request, uri);
     }
 
+    /**
+     * Returns the {@link ServiceRequestContext} whose {@link Service} invokes the {@link Client}
+     * {@link Request} which created this {@link ClientRequestContext}, or {@code null} if this client request
+     * was not made in the context of a server request.
+     */
+    @Nullable
+    ServiceRequestContext rootContext();
```
ikhoon: I guess the main reason for exposing `ServiceRequestContext` is to get attributes from the `rootContext`. Then don't we change `ServiceRequestContext` to `RequestContext`?
If I misunderstood, feel free to ignore this review. 😀

minwoox: It's for context aware and for https://github.com/line/armeria/issues/2173#issuecomment-559438479 🙏

minwoox:
All fixed. PTAL. 🙇

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
ikhoon: Thanks for the link.

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
trustin: > `rootIt` can have attributes instead of being an `emptyIterator()` if there was an insertion before the iteration of `ownAttrsIt` is over. (which is rare)

That's a good point. I guess it's not worth it.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
```
trustin: Sholuld we call it `root()` or not?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+
+    /**
+     * Returns the {@link Attribute} for the given {@link AttributeKey}. This method will never return
+     * {@code null}, but may return an {@link Attribute} which does not have a value set yet.
+     *
+     * <p>If the {@link Attribute} is not in this context, but {@link #rootContext()} has it,
+     * then this will return the {@link Attribute}.
```
trustin:
```
If the {@link Attribute} does not exist in this context but only in {@link #rootContext()},
this method will return the {@link Attribute} whose value is from the {@link #rootContext()}.
```

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     * Please note that if you change the {@link Attribute#get() value} from {@link #rootContext()} using
+     * setter methods (e.g. {@link Attribute#set(Object)}), a new {@link Attribute} for the given
+     * {@link AttributeKey} is created in this {@link ClientRequestContext} and the value will be set to
+     * the {@link Attribute}.
```
trustin: How about this:
```
Please note that any changes made to the {@link Attribute} returned by this method never
affects the {@link Attribute} owned by {@link #rootContext()}. For example:
<pre>{@code
ClientRequestContext ctx = ...;
assert ctx.rootContext().attr(KEY).get() == 42;
assert ctx.hasOwnAttr(KEY) == false;
assert ctx.attr(KEY).get() == 42;
// Overriding the root context attribute creates the client context's own attribute.
ctx.attr(KEY).set(4242);
assert ctx.hasOwnAttr(KEY) == true;
// root context attribute remains unaffected.
assert ctx.rootContext().attr(KEY).get() == 42;
}</pre>
```

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     * If you want to change the value from the root context, please call {@link #attr(AttributeKey)}
+     * from {@link #rootContext()}.
```
trustin: Could you add an example?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     *
+     * @see #attr(AttributeKey)
+     */
+    <T> Attribute<T> ownAttr(AttributeKey<T> key);
+
+    /**
+     * Returns {@code} true if and only if the given {@link Attribute} exists.
+     *
+     * <p>If the {@link Attribute} is not in this context, but {@link #rootContext()} has it,
+     * then this will return {@code true}.
```
trustin:
```
If the {@link Attribute} does not exist in this context but only in {@link #rootContext()},
this method will still return {@code true}.
```

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     * Returns the {@link Iterator} of all {@link Attribute}s this map contains.
+     *
+     * <p>If {@link #rootContext()} is not {@code null}, then the {@link Iterator} will iterate the
+     * {@link #rootContext()}'s {@link #attrs()} afterward.
+     * Please note that if you try to change the {@link Attribute#get() value} which is from the
+     * {@link Iterator} of {@link #rootContext()} using setter methods (e.g. {@link Attribute#set(Object)}),
+     * a new {@link Attribute} for the given {@link AttributeKey} is created in this
+     * {@link ClientRequestContext} and the value will be set to the {@link Attribute}.
+     * If you want to change the value from the root context while iterating, please call
+     * {@link #attrs()} from {@link #rootContext()}.
```
trustin: Could you update this part as well?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+    /**
+     * Returns {@code} true if and only if the given {@link Attribute} exists.
+     * Unlike {@link #hasAttr(AttributeKey)}, this does not search in {@link #rootContext()}.
+     *
+     * @see #hasAttr(AttributeKey)
+     */
+    <T> boolean hasOwnAttr(AttributeKey<T> key);
+
+    /**
+     * Returns the {@link Iterator} of all {@link Attribute}s this map contains.
```
trustin: this map -> this context

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     *
+     * <p>If {@link #rootContext()} is not {@code null}, then the {@link Iterator} will iterate the
+     * {@link #rootContext()}'s {@link #attrs()} afterward.
```
trustin: I think this is too implementation-specific. How about this:

```
The {@link Iterator} returned by this method will also yield the {@link Attribute}s from the
{@link #rootContext()} except those which exist already in this context, e.g.
<pre>{@code
... some example ...
}</pre>
```

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,83 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     * If you want to change the value from the root context while iterating, please call
+     * {@link #attrs()} from {@link #rootContext()}.
+     *
+     * @see #ownAttrs()
+     */
+    @Override
+    Iterator<Attribute<?>> attrs();
+
+    /**
+     * Returns the {@link Iterator} of all {@link Attribute}s this map contains.
```
trustin: this map -> this context

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Well, either is fine I guess. You think `root()` fits better?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -155,19 +193,34 @@
...
     public Iterator<Attribute<?>> attrs() {
         final AtomicReferenceArray<DefaultAttribute<?>> attributes = this.attributes;
         if (attributes == null) {
-            return Collections.emptyIterator();
+            return root != null ? root.attrs() : Collections.emptyIterator();
         }
 
+        final Iterator<Attribute<?>> ownAttrsIt = new IteratorImpl(attributes);
+        return root != null ? new ConcatenatedIteratorImpl(ownAttrsIt, root.attrs())
+                            : ownAttrsIt;
```
trustin: I guess this is not fixed yet, right? Please add a test case for this as well.

```java
AttributeMap root = ...;
root.attr(A).set(1);
root.attr(B).set(2);
AttributeMap attrs = ...;
assert attrs.isEmpty(); // attrs.iterator() will return `root.attrs()` because attrs has no attributes yet.
for (Attribute a : attrs) {
    a.set(a.get() * 10);
}
assert root.attr(A).get() == 1;
assert root.attr(B).get() == 2;
assert attrs.attr(A).get() == 10;
assert attrs.attr(B).get() == 20;
```

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -316,4 +464,58 @@ public boolean hasNext() {
...
+            if (thisIt.hasNext()) {
+                this.next = thisIt.next();
+            } else {
+                // Skip the attribute in rootIt if it's in thisAttributeMap.
+                for (;;) {
+                    if (rootIt.hasNext()) {
+                        final Attribute<?> tempNext = rootIt.next();
+                        if (!hasOwnAttr(tempNext.key())) {
+                            this.next = new CopyOnWriteAttribute<>(tempNext);
+                            break;
```
trustin: How about just returning here?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -316,4 +464,58 @@ public boolean hasNext() {
...
+                        }
+                    } else {
+                        this.next = null;
+                        break;
```
trustin: Need to throw a `NoSuchElementException`? Please add a test case.

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -316,4 +464,58 @@ public boolean hasNext() {
...
+
+        @Override
+        public Attribute<?> next() {
+            final Attribute<?> next = this.next;
+            if (next == null) {
+                throw new NoSuchElementException();
+            }
+
+            if (thisIt.hasNext()) {
+                this.next = thisIt.next();
```
trustin: How about returning here early and unwrapping the `else` block below?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
minwoox: This is for prefetching. `NoSuchElementException ` is raised here https://github.com/line/armeria/pull/2322/files#diff-74634bdb2b0b3d623e3378b6a6fa9037R498

trustin: Ahh, OK. Thanks!

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
minwoox: Do you think https://github.com/line/armeria/pull/2322/files#diff-2bd4763eb18d444219c2be4de4da9911R292 is not sufficient? I will update it then.

trustin: It's not enough. You return `root.attrs()` above.

minwoox: Oops Completely missed that. 😅

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: @anuraaga Do you have an idea on this? 😄

minwoox:
All fixed. 😉

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
ikhoon: minor nit: How about changing the values, which are assigned to the attribute, to meaningful values to make sure that the values belongs to.
```
// I'm not sure this is good. 😅
42 -> "ROOT"
4242 -> "CLIENT"
```

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -263,11 +338,99 @@ private void remove0() {
...
+        }
+
+        @Override
+        public void set(T value) {
+            if (attrInThis != null) {
+                attrInThis.set(value);
+            } else {
+                final Attribute<T> attrInThis = attr(attrInRoot.key(), true);
+                this.attrInThis = attrInThis;
+                attrInThis.set(value);
```
ikhoon: nit: Remove the temporary local variable?
```java
attrInThis = attr(attrInRoot.key(), true);
attrInThis.set(value);
```

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -263,11 +338,99 @@ private void remove0() {
...
     }
 
+    private final class CopyOnWriteAttribute<T> implements Attribute<T> {
+
+        private final Attribute<T> attrInRoot;
+
+        @Nullable
+        private Attribute<T> attrInThis;
+
+        CopyOnWriteAttribute(Attribute<T> attrInRoot) {
```
ikhoon: I think `root` concept does not make sense a little. 🧐
How about `attrInRoot` -> `original`, `attrInThis` -> `copied`?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -263,11 +338,99 @@ private void remove0() {
...
+                attrInThis.set(value);
+            } else {
+                final Attribute<T> attrInThis = attr(attrInRoot.key(), true);
+                this.attrInThis = attrInThis;
+                attrInThis.set(value);
+            }
+        }
+
+        @Override
+        public T getAndSet(T value) {
```
ikhoon: Add `@Nullable`? from the netty [Javadoc](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/Attribute.java#L40-L43):
```java
/**
 *  Atomically sets to the given value and returns the old value which may be {@code null} if non was set before.
 */
T getAndSet(T value);
```

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -263,11 +338,99 @@ private void remove0() {
...
+        @Override
+        public T get() {
+            if (attrInThis != null) {
+                return attrInThis.get();
+            }
+            return attrInRoot.get();
+        }
+
+        @Override
+        public void set(T value) {
```
ikhoon: `requireNonNull(value, "value")`?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -263,11 +338,99 @@ private void remove0() {
...
+
+        @Override
+        public boolean compareAndSet(T oldValue, T newValue) {
+            if (attrInThis != null) {
+                return attrInThis.compareAndSet(oldValue, newValue);
+            }
+            // Do I need to just return false here because this does not guarantee atomicity?
+            if (attrInRoot.get() == oldValue) {
+                final Attribute<T> attrInThis = attr(attrInRoot.key(), true);
+                this.attrInThis = attrInThis;
```
ikhoon: Can remove temporary local variable?

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
```diff
@@ -263,11 +338,99 @@ private void remove0() {
...
+            this.attrInThis = attrInThis;
+            return attrInThis.setIfAbsent(value);
+        }
+
+        @Override
+        public T getAndRemove() {
+            if (attrInThis != null) {
+                return attrInThis.getAndRemove();
+            }
+            return attrInRoot.get();
```
ikhoon: Need copy value here? I expect the following result:

```java
final DefaultAttributeMap root = new DefaultAttributeMap(null);
final AttributeKey<String> rootKey = AttributeKey.valueOf("root-key");
root.attr(rootKey).set("ROOT");
final DefaultAttributeMap child = new DefaultAttributeMap(root);
assert child.attrs() instanceof CopyOnWriteAttributeIterator;
final Attribute<String> attribute = (Attribute<String>) child.attrs().next();
assert attribute.getAndRemove() == "ROOT";
assert attribute.getAndRemove() == null;  // assertion fail, returns "ROOT"
```

ikhoon: Just call `getAndSet(null)` then own implementation?

minwoox: Nice catch, thanks!
Had a chat with @trustin and we need to revamp our attribute API.
Because the `AttributeMap` from netty has deprecated APIs such as `getAndRemove()` which will cause a race condition, we decided not to extend it anymore.
So we will add the following methods to `RequestContext`:
```
@Nullable
<T> T attr(AttributeKey<T> key);
<T> void setAttr(AttributeKey<T> key, T value);
Entry<AttributeKey<?>, ?> attrs();
```
and `ownX` methods to `ClientRequestContext`:
```
<T> T ownAttr(AttributeKey<T> key);
<T> void setOwnAttr(AttributeKey<T> key, T value);
Entry<AttributeKey<?>, ?> ownAttrs();
```
The reason why I don't define a new `AttributeMap` is that we are using the netty's `AttributeMap` in `ConnectionPoolListener` and didn't want to make a user get confused by using the same interface.

What do you think? @anuraaga

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -193,19 +195,50 @@ public final MeterRegistry meterRegistry() {
 
     @Override
     public <T> Attribute<T> attr(AttributeKey<T> key) {
-        return attrs.attr(key);
+        return attrs.attr(requireNonNull(key, "key"));
+    }
+
+    /**
+     * Returns the {@link Attribute} for the given {@link AttributeKey}. This method will never return
+     * {@code null}, but may return an {@link Attribute} which does not have a value set yet.
```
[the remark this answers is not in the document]
minwoox: Just override the description from the upstream, and I thought it's okay to just leave it as it is.

core/src/main/java/com/linecorp/armeria/internal/DefaultAttributeMap.java
anuraaga: Makes sense to me (less methods is always better when reasonable :) )

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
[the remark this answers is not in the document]
ikhoon: I didn't know that, sound good to me. :-)

minwoox:
This is ready for the another round. 😉

core/src/main/java/com/linecorp/armeria/common/DefaultAttributeMap.java
```diff
@@ -0,0 +1,453 @@
...
+            }
+        }
+    }
+
+    private static int index(AttributeKey<?> key) {
+        return key.id() & MASK;
+    }
+
+    @Nullable
+    <T> T setAttrIfAbsent(AttributeKey<T> key, @Nullable T value) {
```
trustin: How about adding `computeAttrIfAbsent` as well?

core/src/main/java/com/linecorp/armeria/common/DefaultAttributeMap.java
```diff
@@ -0,0 +1,453 @@
...
+        if (rootAttributeMap == null) {
+            return ownAttrsIt;
+        }
+
+        final Iterator<Entry<AttributeKey<?>, Object>> rootAttrs = rootAttributeMap.attrs();
+        if (!rootAttrs.hasNext()) {
+            return ownAttrsIt;
+        }
+
+        return new ConcatenatedIteratorImpl(ownAttrsIt, rootAttrs);
```
trustin: `ConcatenatedCopyOnWriteIterator`?

core/src/main/java/com/linecorp/armeria/common/DefaultAttributeMap.java
```diff
@@ -0,0 +1,453 @@
...
+            if (rootAttributeMap == null) {
+                return Collections.emptyIterator();
+            }
+
+            final Iterator<Entry<AttributeKey<?>, Object>> rootAttrs = rootAttributeMap.attrs();
+            if (!rootAttrs.hasNext()) {
+                return Collections.emptyIterator();
+            }
+
+            return new CopyOnWriteAttributeIterator(rootAttrs);
```
trustin: `CopyOnWriteIterator`? (this class is all about attributes, so..)

core/src/main/java/com/linecorp/armeria/common/DefaultAttributeMap.java
```diff
@@ -0,0 +1,453 @@
...
+        @Nullable
+        @Override
+        public T setValue(@Nullable T value) {
+            final T old = this.value;
+            this.value = value;
+            return old;
+        }
+    }
+
+    private static final class IteratorImpl implements Iterator<Entry<AttributeKey<?>, Object>> {
```
trustin: `OwnIterator`?

core/src/main/java/com/linecorp/armeria/common/NonWrappingRequestContext.java
```diff
@@ -42,9 +41,9 @@
 public abstract class NonWrappingRequestContext extends AbstractRequestContext {
 
     private final MeterRegistry meterRegistry;
-    private final DefaultAttributeMap attrs = new DefaultAttributeMap();
+    private final DefaultAttributeMap attrs;
     private final SessionProtocol sessionProtocol;
-    private RequestId id;
+    private final RequestId id;
```
trustin: Thanks :sweat_smile:

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -112,6 +112,36 @@
...
+
+    /**
+     * Associates the specified value with the given {@link AttributeKey} in this context only
+     * if this context does not contain a mapping for the {@link AttributeKey}.
+     *
+     * @return {@code null} if there was no mapping for the {@link AttributeKey} or the old value if there's
+     *         a mapping for the {@link AttributeKey}.
+     */
+    @Nullable
+    <V> V setAttrIfAbsent(AttributeKey<V> key, V value);
```
trustin: Could we also have `computeAttrIfAbsent()`?

site/src/sphinx/advanced-custom-attributes.rst
```diff
@@ -24,29 +24,18 @@ Then, you can access them via ``RequestContext.attr(AttributeKey)``:
...
-If you are only interested in checking whether an attribute exists or not, you should use
-``RequestContext.hasAttr(AttributeKey)``. It does not perform any allocation for a non-existent attribute
-unlike ``RequestContext.attr(AttributeKey)`` does:
-
-.. code-block:: java
-
-    if (ctx.hasAttr(INT_ATTR)) {
-        int i = ctx.attr(INT_ATTR).get();
-        ...
+    for (Entry<AttributeKey<?>, Object> a : ctx.attrs()) {
```
trustin: `attrs()` does not return `Iterable` anymore, so I guess this will not work. How about using `ctx.attrs().forEach()`?

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
minwoox: Added!

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -161,6 +165,114 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     * <pre>{@code
+     * ClientRequestContext ctx = ...;
+     * assert ctx.ownAttr(KEY_A) == "child_a";
+     * assert ctx.rootContext().attr(KEY_A) == "root_a";
+     * assert ctx.rootContext().attr(KEY_B) == "root_b";
+     *
+     * Iterator<Entry<AttributeKey<?>, Object>> attrs = ctx.attrs();
+     * assert attrs.next().getValue() == "child_a"; // KEY_A
+     * // Skip KEY_A in the root context.
+     * assert attrs.next().getValue() == "root_b"; // KEY_B
```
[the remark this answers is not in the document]
minwoox: Oops! fixed.

core/src/main/java/com/linecorp/armeria/common/DefaultAttributeMap.java
minwoox: Fixed. 😉

site/src/sphinx/advanced-custom-attributes.rst
minwoox: Oops, fixed to use `forEachRemaining` 😉

minwoox:
@anuraaga Would you mind taking a look at this again? 😄

brave/src/test/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContextTest.java
```diff
@@ -17,71 +17,50 @@
...
 import brave.propagation.CurrentTraceContext.Scope;
 import brave.propagation.TraceContext;
 import io.netty.channel.EventLoop;
-import io.netty.util.Attribute;
-import io.netty.util.AttributeKey;
 
-public class RequestContextCurrentTraceContextTest {
-    @Rule
-    public MockitoRule mockitoRule = MockitoJUnit.rule();
+@ExtendWith(MockitoExtension.class)
```
anuraaga: Shouldn't need this, `testing-internal` enables this extension automatically

brave/src/test/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContextTest.java
```diff
@@ -17,71 +17,50 @@
...
+class RequestContextCurrentTraceContextTest {
 
-    @Mock(answer = Answers.CALLS_REAL_METHODS)
-    RequestContext mockRequestContext;
-    @Mock(answer = Answers.CALLS_REAL_METHODS)
-    RequestContext mockRequestContext2;
+    @Nullable
```
anuraaga: I don't think we've ever used `@Nullable` in tests, probably overkill

brave/src/test/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContextTest.java
```diff
@@ -17,71 +17,50 @@
...
-        try (SafeCloseable requestContextScope = mockRequestContext.push()) {
-            try (Scope traceContextScope = currentTraceContext.newScope(traceContext)) {
-                TraceContextUtil.copy(mockRequestContext, mockRequestContext2);
-                assertThat(attrs1.attrs().next().get())
-                        .isEqualTo(traceContext)
-                        .isEqualTo(attrs2.attrs().next().get());
-            }
-        }
+    @BeforeEach
+    void setup() {
```
anuraaga: `setUp()` more conventional

brave/src/test/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContextTest.java
```diff
@@ -17,71 +17,50 @@
...
+        Mockito.lenient().when(eventLoop.inEventLoop()).thenReturn(true);
```
anuraaga: Better to add `@MockitoSettings` to the class to say to use lenient stubs

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
```diff
@@ -621,22 +609,4 @@ String toStringResponseOnly(Function<? super HttpHeaders, ?> headersSanitizer,
     String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersSanitizer,
                                 Function<Object, ?> contentSanitizer,
                                 Function<? super HttpHeaders, ?> trailersSanitizer);
-
-    /**
-     * {@inheritDoc}
-     *
-     * @deprecated Use {@code RequestLog.context().attr()}.
```
anuraaga: Did we already release armeria with these deprecations? If not, then this will be breaking with deprecation time. Maybe it's such a minor API it's ok to do that though.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
anuraaga: `root()` seems slightly better to me too.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
minwoox: We did not release yet. Perhaps we can release another version before merging this?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Let me use `root()`, then. Thanks!

brave/src/test/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContextTest.java
minwoox: Fixed all. 😄 Thanks!

brave/src/test/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContextTest.java
minwoox: Oops I didn't notice that. Thanks!

brave/src/test/java/com/linecorp/armeria/common/brave/RequestContextCurrentTraceContextTest.java
minwoox: Added!

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -52,15 +52,17 @@
...
+    /**
+     * Sets up request metrics.
+     */
+    public static void setup(RequestContext ctx, AttributeKey<Boolean> requestMetricsSetKey,
+                             MeterIdPrefixFunction meterIdPrefixFunction, boolean server) {
+        final Boolean isRequestMetricsSet = ctx.attr(requestMetricsSetKey);
 
-    public static void setup(RequestContext ctx, MeterIdPrefixFunction meterIdPrefixFunction, boolean server) {
-        if (ctx.hasAttr(ATTR_REQUEST_METRICS_SET)) {
+        if (isRequestMetricsSet != null && isRequestMetricsSet) {
```
ikhoon: nit: Simplify with:
```suggestion
        if (Boolean.TRUE.equals(isRequestMetricsSet)) {
```

minwoox: Fixed. Thank you! 😉

minwoox:
Thanks a lot @anuraaga, @ikhoon and @trustin for reviewing. 😉

anuraaga:
I had a random finding - should we return `root` for `ServiceRequestContext.current` if it's available?

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java#L70

I'm not entirely sure but I think it might be appropriate, the root is still active even though it's shadowed by a client context.

minwoox:
> should we return root for ServiceRequestContext.current if it's available?

Yes, I agree. Let me make a PR for that. Thanks!

PR #2323, opened 2019-12-13
Remove the word HTTP in HttpService and HttpClient classes

minwoox:
Following https://github.com/line/armeria/pull/2239#issuecomment-551433761

Modifications:
- Server-side
  - AnnotatedHttpDocServicePlugin -> AnnotatedDocServicePlugin
  - AnnotatedHttpService -> AnnotatedService
  - AnnotatedHttpServiceElement -> AnnotatedServiceElement
  - AnnotatedHttpServiceFactory -> AnnotatedServiceFactory
  - AnnotatedHttpServiceExtensions -> AnnotatedServiceExtensions
  - AnnotatedHttpServiceTypeUtil -> AnnotatedServiceTypeUtil (package-local)
  - HttpAuthService -> AuthService
  - HttpAuthServiceBuilder -> AuthServiceBuilder
  - HttpEncodingService -> EncodingService
  - HttpFileService -> FileService
  - HttpFileServiceConfig -> FileServiceConfig
  - HttpFileServiceBuilder -> FileServiceBuilder
    - HttpFile is not changed
  - WebOperationHttpService -> WebOperationService (package-local)
- Client-side
  - RetryingClient -> AbstractRetryingClient
  - RetryingHttpClient -> RetryingClient
  - RetryingClientBuilder -> AbstractRetryingClientBuilder
  - CircuitBreakerClient -> AbstractCircuitBreakerClient
  - CircuitBreakerHttpClient -> CircuitBreakerClient
  - CircuitBreakerHttpClientBuilder -> CircuitBreakerClientBuilder
  - HttpDecodingClient -> DecodingClient
  - ConcurrencyLimitingClient -> AbstractConcurrencyLimitingClient
  - ConcurrencyLimitingHttpClient -> ConcurrencyLimitingClient

Result:
- (Deprecated)
  - Use `AuthService` instead of `HttpAuthService`.
  - Use `AuthServiceBuilder` instead of `HttpAuthServiceBuilder`.
  - Use `EncodingService` instead of `HttpEncodingService`.
  - Use `FileService` instead of `HttpFileService`.
  - Use `FileServiceBuilder` instead of `HttpFileServiceBuilder`.
  - Use `RetryingClient` instead of `RetryingHttpClient`.
  - Use `RetryingClientBuilder` instead of `RetryingHttpClientBuilder`.
  - Use `CircuitBreakerClient` instead of `CircuitBreakerHttpClient`.
  - Use `CircuitBreakerClientBuilder` instead of `CircuitBreakerHttpClientBuilder`.
  - Use `ConcurrencyLimitingClient` instead of `ConcurrencyLimitingHttpClient`.
- (Breaking)
  - `RetryingHttpClientBuilder.build()` now returns `RetryingClient`.
    - `RetryingHttpClient.newDecorator(...)` now returns `Function<? super HttpClient, RetryingClient>`.
  - `CircuitBreakerHttpClietBuilder.build()` now returns `CircuitBreakerClient`.
    - `CircuitBreakerHttpCliet.newDecorator(...)` now returns `Function<? super HttpClient, CircuitBreakerClient>`.
  - `HttpAuthServiceBuilder.build()` now returns `AuthService`.
    - `HttpAuthService.newDecorator(...)` now returns `Function<? super HttpService, AuthService>`.


anuraaga:
Do you think there's a way to keep the old classes for a bit with inheritance? The previous migration was fairly involved so it feels a little like dark times being an Armeria user. This time it's much simpler though, so I'd only give it a small try if reasonable.

minwoox:
I thought we are going to remove the deprecated classes after 1.0 so chose the simpler way, but yes, I need to deprecate them first. Let me try that. Thanks!

minwoox:
@anuraaga Deprecate the classes except for the internal and package-private classes. PTAL. 🙇

minwoox:
I didn't change the inception year because it's just renaming.

anuraaga:
Thanks a lot! It's nice to have some time to migrate :)

trustin:
Shoudn't `ServiceRegistrationBean` be `HttpServiceRegistrationBean`, because it registers `HttpService`?

minwoox:
> Shoudn't ServiceRegistrationBean be HttpServiceRegistrationBean, because it registers HttpService?

Yes, let me revert that. Thanks!

minwoox:
> Shoudn't ServiceRegistrationBean be HttpServiceRegistrationBean, because it registers HttpService?

Done. PTAL. 🙇

minwoox:
Let me squash commits 🙏

core/src/main/java/com/linecorp/armeria/client/limit/ConcurrencyLimitingHttpClient.java
```diff
@@ -16,79 +16,43 @@
...
- * client = builder.build(...);
+ * WebClientBuilder builder = WebClient.builder(...);
+ * builder.decorator(ConcurrencyLimitingClient.newDecorator(16));
+ * WebClient client = builder.build();
  * }</pre>
  *
  */
-public final class ConcurrencyLimitingHttpClient extends ConcurrencyLimitingClient<HttpRequest, HttpResponse>
-        implements HttpClient {
+public final class ConcurrencyLimitingHttpClient extends ConcurrencyLimitingClient {
```
ikhoon: Need `@Deprecated`?

minwoox: Oops Thanks!

ikhoon: My pleasure.

core/src/main/java/com/linecorp/armeria/server/auth/HttpAuthService.java
```diff
@@ -20,79 +20,51 @@
...
     /**
      * Creates a new HTTP authorization {@link HttpService} decorator using the specified
      * {@link Authorizer}s.
      *
      * @param authorizers a list of {@link Authorizer}s.
+     * @deprecated Use {@link AuthService#newDecorator(Iterable)}.
      */
-    public static Function<? super HttpService, HttpAuthService> newDecorator(
+    @Deprecated
+    public static Function<? super HttpService, AuthService> newDecorator(
```
ikhoon: Can we change the return type without breaking change?

minwoox: I tried but in order to do that I need to copy the whole source or add another type parameter to the `AuthService`. So I have chosen this way because most of the users:
- do not declare a variable for the return value
- if they do that, they can easily notice because the IDE shows which one is wrong
So let me just do this way. 🙏

ikhoon: I understood. Then how about adding this to the breaking change list?

minwoox: Sure 😉

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
```diff
@@ -121,11 +121,11 @@ public Server armeriaServer(
         configureHttpServices(serverBuilder,
                               httpServiceRegistrationBeans.orElseGet(Collections::emptyList),
                               meterIdPrefixFuncFactory);
-        configureAnnotatedHttpServices(serverBuilder,
-                                       docServiceBuilder,
-                                       annotatedServiceRegistrationBeans.orElseGet(Collections::emptyList),
-                                       meterIdPrefixFuncFactory,
-                                       docsPath);
+        configureAnnotatedServices(server,
```
ikhoon: `server` -> `serverBuilder` maybe due to merge conflict.

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -242,7 +242,7 @@ private void configureArmeriaService(ServerBuilder sb, ArmeriaSettings settings)
         configureHttpServices(sb,
                               findBeans(HttpServiceRegistrationBean.class),
                               meterIdPrefixFunctionFactory);
-        configureAnnotatedHttpServices(sb,
+        configureAnnotatedServices(sb,
                                        docServiceBuilder,
                                        findBeans(AnnotatedServiceRegistrationBean.class),
                                        meterIdPrefixFunctionFactory,
```
ikhoon: nit: indent?

minwoox:
Let me just squash and rebase because it's hard to know what causes the build failure and it's also hard to recognize what's really changed from the previous commit in this case. 🙇

trustin:
@anuraaga Any other comments?

anuraaga:
LGTM

trustin:
Thanks a lot, @minwoox!

PR #2326, opened 2019-12-17
Small gRPC cleanups

anuraaga:
Just found these random points. One is removing an unused field (after we switched to pre-preparing the response headers)


trustin:
Thanks a lot, @anuraaga and reviewers!

Issue #2185, opened 2019-10-12
Allow customizing DocService when using Spring Boot integration


trustin:
I don't remember exactly where, but someone asked if it's possible to add an example via DocServiceBuilder when using Spring Boot integration.

It is currently not possible because ArmeriaAutoConfiguration currently builds a DocService by itself without exposing any customization point.

We could allow a user to inject a Consumer or Supplier of DocServiceBuilder so that the user can customize the DocService built by ArmeriaAutoConfiguration, and thus it's not necessary to to use registration beans.

Adding good-first-issue label since it should be fairly easy for anyone who's familiar with Spring's dependency injection mechanism.

yanghun0070:
If there is no working person, I want to work.

trustin:
Fixed via #2327 by @Charliocat. Thanks again for reporting the problem.

PR #2327, opened 2019-12-17
Allow customizing DocService when integrating with Spring Boot

Charliocat:
Hi I'd like to solve this issue #2185 , I made some changes but I don't know if what I'm doing is the right choice, could you help me?

### DIFF
--- a/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
+++ b/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
@@ -75,7 +75,8 @@ public class ArmeriaAutoConfiguration {
             Optional<List<ThriftServiceRegistrationBean>> thriftServiceRegistrationBeans,
             Optional<List<GrpcServiceRegistrationBean>> grpcServiceRegistrationBeans,
             Optional<List<HttpServiceRegistrationBean>> httpServiceRegistrationBeans,
-            Optional<List<AnnotatedServiceRegistrationBean>> annotatedServiceRegistrationBeans)
+            Optional<List<AnnotatedServiceRegistrationBean>> annotatedServiceRegistrationBeans,
+            Optional<List<DocServiceConfigurator>> docServiceConfigurators)
             throws InterruptedException {
 
         if (!armeriaServerConfigurators.isPresent() &&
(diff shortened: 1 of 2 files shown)


spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
```diff
@@ -60,6 +63,9 @@
     private static final Port DEFAULT_PORT = new Port().setPort(8080)
                                                        .setProtocol(SessionProtocol.HTTP);
 
+    @Autowired(required = false)
+    private DocServiceBuilderSupplier docServiceBuilderSupplier;
```
trustin: You could instead move this into a parameter of `armeriaServer()` method for consistency, because other components are injected there.

core/src/main/java/com/linecorp/armeria/server/docs/DocServiceBuilderSupplier.java
```diff
@@ -0,0 +1,22 @@
...
+distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+License for the specific language governing permissions and limitations
+under the License.
+*/
+
+package com.linecorp.armeria.server.docs;
+
+@FunctionalInterface
+public interface DocServiceBuilderSupplier {
```
trustin: - How about `DocServiceConfigurator`?
- How about implementing `Ordered`, like we did in https://github.com/line/armeria/blob/master/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaServerConfigurator.java ?
- How about just returning void? Is there a reason you wanted to return a `DocServiceBuilder`?

core/src/main/java/com/linecorp/armeria/server/docs/DocServiceBuilderImpl.java
```diff
@@ -0,0 +1,476 @@
...
+import com.google.common.collect.ListMultimap;
+
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+/**
+ * Builds a new {@link DocService}.
+ */
+public final class DocServiceBuilderImpl implements DocServiceBuilder {
```
trustin: Do we need to split `DocServiceBuilder` into an interface and an implementation? Do you have any specific use case for this? If not, it'd be nice if we could merge this back into `DocServiceBuilder`.

heowc:
I think I need to think briefly about this. 
If `DocServiceBuilder` exists to be injected, use it as is, otherwise, create a new instance. 

WDYT?

core/src/main/java/com/linecorp/armeria/server/docs/DocServiceBuilderSupplier.java
Charliocat: - I'll change the name for `DocServiceConfigurator` I put supplier because of the description in the issue.
- I'll check `Ordered` thanks for the info.
- I wanted to return a `DocServiceBuilder` so it'd be a wrapped `DocServiceBuilder` with the customized methods.

core/src/main/java/com/linecorp/armeria/server/docs/DocServiceBuilderImpl.java
Charliocat: I created the interface in order to inject the `DocServiceBuilder`, so it could be any custom implementation. But maybe it's not needed. I'll review it, and  I'll undo this change.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaAutoConfiguration.java
Charliocat: ok

Charliocat:
> I think I need to think briefly about this.
> If `DocServiceBuilder` exists to be injected, use it as is, otherwise, create a new instance.
> 
> WDYT?

Yes that was my first intention now I understand the Optional, it's very interesting.
 Then I checked the [ArmeriaServerConfigurator](https://github.com/line/armeria/blob/master/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaServerConfigurator.java) which extends Ordered and I copied it for the `DocServiceBuilder`.

trustin:
> If `DocServiceBuilder` exists to be injected, use it as is, otherwise, create a new instance. 

I'm not sure what's the advantage of this. Any specific use case that's impossible with `DocServiceCustomizer`?

heowc:
I agree to create a class like `DocServiceCustomizer`. This will give you the flexibility to deal with another problem later.

Charliocat:
Hi I made some changes but I didn't know how to add the spring dependencies on this module, could you tell me please?

heowc:
> Hi I made some changes but I didn't know how to add the spring dependencies on this module, could you tell me please?

I think core doesn't need the spring dependencies.

How about moving to [`boot-autoconfigure/com.linecorp.armeria.spring`](https://github.com/line/armeria/tree/master/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring)?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/DocServiceConfigurator.java
```diff
@@ -0,0 +1,48 @@
+/*
+Copyright 2019 LINE Corporation
```
trustin: Could you make the copyright header look same with other files? i.e.

```java
/*
 * Copyright ...
 *
 * ...
 */
```

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/DocServiceConfigurator.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.AnnotationAwareOrderComparator;
+import org.springframework.core.annotation.Order;
+
+import com.linecorp.armeria.server.docs.DocServiceBuilder;
+
+@FunctionalInterface
+public interface DocServiceConfigurator extends Ordered {
```
ikhoon: Add documentation for interface?

Charliocat: I added the documentation for interface, if it needs to be changed let me know

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/DocServiceConfigurator.java
Charliocat: Done!

trustin:
Thank you, @Charliocat and reviewers!

Charliocat:
Thank you all!

PR #2328, opened 2019-12-18
Propagate ClientRequestContext to Thrift client callback

minwoox:
Motivation:
`ClientRequestContext` is not propagate to `AsyncMethodCallback` and the callbacks registered to `RpcResponse`.

Modification:
- Introduce `RequestContextAwareRpcResponse`

Result:
- You can now get `ClientRequestContext` in `AsyncMethodCallback`
- Fix a problem in #760. (still have to solve https://github.com/line/armeria/issues/760#issuecomment-476503179)


core/src/main/java/com/linecorp/armeria/common/DefaultRpcResponse.java
```diff
@@ -84,27 +82,4 @@ public boolean cancel(boolean mayInterruptIfRunning) {
         final boolean updated = !isDone() && completeExceptionally(new CancellationException());
         return updated || isCancelled();
     }
-
-    @Override
-    public String toString() {
```
minwoox: didn't see any reason to override this. Please let me know if there's one. I will put this back.

core/src/main/java/com/linecorp/armeria/internal/RequestContextAwareRpcResponse.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.common.RequestContext;
+import com.linecorp.armeria.common.RpcResponse;
+
+final class RequestContextAwareRpcResponse extends RequestContextAwareCompletableFuture<Object>
```
minwoox: All callbacks (e.g `rpcResponse.thenApply(...)`) does not return `RequestContextAwareRpcResponse` but `RequestContextAwareCompletableFuture`.
I think it's right behavior not to return `RequestContextAwareRpcResponse`, but if I was wrong please let me know.

core/src/main/java/com/linecorp/armeria/common/DefaultRpcResponse.java
trustin: I think it's because `CompletableFuture.toString()` ..

- Adds object hash code, which is somewhat verbose.
- Does not add the value.

core/src/main/java/com/linecorp/armeria/internal/RequestContextAwareRpcResponse.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Nullable
+    @Override
+    public Throwable cause() {
+        return cause;
+    }
+
+    @Override
+    public boolean completeExceptionally(Throwable cause) {
+        causeUpdater.compareAndSet(this, null, requireNonNull(cause));
```
trustin: `, "cause"` ..?

trustin:
Nice to have this on the client side, but would it be possible to do the same on the server side? :thinking:

core/src/main/java/com/linecorp/armeria/internal/RequestContextAwareRpcResponse.java
```diff
@@ -0,0 +1,69 @@
...
+    public void obtrudeException(Throwable cause) {
+        this.cause = requireNonNull(cause);
+        super.obtrudeException(cause);
+    }
+
+    @Override
+    public boolean cancel(boolean mayInterruptIfRunning) {
+        final boolean updated = !isDone() && completeExceptionally(new CancellationException());
+        return updated || isCancelled();
+    }
```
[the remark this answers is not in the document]
ikhoon: I found the original `DefaultRpcReponse` overrides super method. 
but I'm still wondering the difference. 😄

core/src/main/java9/com/linecorp/armeria/internal/Java9RequestContextAwareRpcResponse.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public boolean completeExceptionally(Throwable cause) {
+        causeUpdater.compareAndSet(this, null, requireNonNull(cause));
+        return super.completeExceptionally(cause);
+    }
+
+    @Override
+    public void obtrudeException(Throwable cause) {
+        this.cause = requireNonNull(cause);
```
ikhoon: ditto.

core/src/main/java/com/linecorp/armeria/internal/RequestContextAwareRpcResponse.java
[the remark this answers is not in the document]
minwoox: It's for updating `cause` using `completeExceptionally`.

core/src/main/java/com/linecorp/armeria/internal/RequestContextAwareRpcResponse.java
```diff
@@ -0,0 +1,69 @@
...
+    @Nullable
+    @Override
+    public Throwable cause() {
+        return cause;
+    }
+
+    @Override
+    public boolean completeExceptionally(Throwable cause) {
+        causeUpdater.compareAndSet(this, null, requireNonNull(cause));
+        return super.completeExceptionally(cause);
```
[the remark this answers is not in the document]
minwoox: Yes, I need that. Thanks!

core/src/main/java/com/linecorp/armeria/common/DefaultRpcResponse.java
minwoox: Thanks for the explanation. Let me add it back. 😄

minwoox:
> Nice to have this on the client side, but would it be possible to do the same on the server side? 🤔

Let me add it. 😄 
I didn't add it because I though we don't need it in the server-side. But it turned out to be that I was wrong because a user can add an RPC decorator which is needed to be called with the thread-local in it. Thanks!

minwoox:
After adding it to the server-side, I realized that we have to solve #2173 first. I will work on this after #2173 is solve. 😉

thrift/src/main/java/com/linecorp/armeria/server/thrift/ThriftCallService.java
```diff
@@ -123,28 +124,29 @@ public RpcResponse serve(ServiceRequestContext ctx, RpcRequest call) throws Exce
...
-        return new DefaultRpcResponse(new TApplicationException(
+        reply.toCompletableFuture().completeExceptionally(new TApplicationException(
                 TApplicationException.UNKNOWN_METHOD, "unknown method: " + call.method()));
+        return reply;
     }
 
     private static void invoke(
             ServiceRequestContext ctx,
-            Object impl, ThriftFunction func, List<Object> args, DefaultRpcResponse reply) {
+            Object impl, ThriftFunction func, List<Object> args, CompletableFuture<Object> reply) {
```
ikhoon: I wonder that there are any reasons to define the type parameter of `CompletableFuture` as a `Object`?

or how about using `CompletableFuture<?>`?

minwoox:
This is not ready yet. 😉

thrift/src/main/java/com/linecorp/armeria/server/thrift/ThriftCallService.java
minwoox: I cannot complete the future here
```java
public void onComplete(Object response) {
    reply.complete(response);
}
```
So let me just use `Object`. 😉

minwoox:
The problem is gone after merging https://github.com/line/armeria/pull/2354.
😄

minwoox:
Thanks reviewer! 😉

PR #2329, opened 2019-12-18
Fix to return context aware future from Java9RequestContextAwareCompl…

minwoox:
…etableFuture

Found out that we forget to add `ctx.makeContextAware()` in `Java9RequestContextAwareCompletableFuture


anuraaga:
We're not supposed to need it because of `newIncompleteFuture`

https://github.com/line/armeria/blob/cc514074338d6bad8e92092ada242367c8373a39/core/src/main/java9/com/linecorp/armeria/internal/Java9RequestContextAwareCompletableFuture.java#L253

Do you an issue without `makeContextAware`?

minwoox:
> Do you see an issue without makeContextAware?

Oops, I completely miss that. Let me close this then, thanks!

ikhoon:
@minwoox The unit test look good to me. 😀

PR #2330, opened 2019-12-19
Delegate to log.context() instead of implementing client connec…

anuraaga:
…imings storage twice.

This is originally what made me think that `RequestLog` and `RequestContext` have different attributes :) Personally, having attribute methods on log that just delegate to context is confusing semantically and possibly overkill as a shortcut method, so if anyone else agrees with that, I could also remove those methods in this PR.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientConnectionTimings.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientConnectionTimings.java
@@ -75,10 +75,7 @@ public final class ClientConnectionTimings {
     @Nullable
     public static ClientConnectionTimings get(RequestLog log) {
         requireNonNull(log, "log");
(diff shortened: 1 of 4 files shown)


trustin:
> Personally, having attribute methods on log that just delegate to context is confusing semantically and possibly overkill as a shortcut method, so if anyone else agrees with that, I could also remove those methods in this PR.

Agreed. Deprecation first?

anuraaga:
Thanks - added deprecations

minwoox:
Could you update the documentation in advanced-custom-attributes as well? 😉

minwoox:
Found out that we use it in `InvocationUtil` as well. 😄

anuraaga:
Thanks for finding these!

trustin:
Thanks, @anuraaga !

PR #2333, opened 2019-12-20
Improve the performance of query string encoding/decoding

trustin:
Motivation:

As demonstrated by @franz1981, we can bypass JVM's array boundary checks
completely, by using `PlatformDependent.getByte()`.

Modifications:

- Optimize `QueryStringEncoder.isUnsafeOctet()` and
  `QueryStringDecoder.decodeHexNibble()` using
  `PlatformDependent.getByte()`.
- Fix an off-by-one error in `decodeHexNibble()`

Result:

- Another performance gain
- Fixes an off-by-one error in `decodeHexNibble()`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
@@ -267,16 +267,19 @@ final class QueryStringDecoder {
     }
 
     private static int decodeHexNibble(char c) {
-        if (c >= OCTETS_TO_HEX.length) {
+        // Widen explicitly so that JVM doesn't have to widen many times.
+        @SuppressWarnings("UnnecessaryLocalVariable")
+        final int index = c;
+        if (index >= OCTETS_TO_HEX.length) {
             return -1;
         }
 
         if (PlatformDependent.hasUnsafe()) {
             // Avoid boundary checks
-            return PlatformDependent.getByte(OCTETS_TO_HEX, c);
+            return PlatformDependent.getByte(OCTETS_TO_HEX, index);
         }
 
-        return OCTETS_TO_HEX[c];
+        return OCTETS_TO_HEX[index];
     }
 
     private static boolean isContinuation(int b) {
--- a/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
@@ -38,6 +38,8 @@ import com.google.common.annotations.VisibleForTesting;
 
 import com.linecorp.armeria.internal.TemporaryThreadLocals;
 
+import io.netty.util.internal.PlatformDependent;
+
 final class QueryStringDecoder {
 
     @SuppressWarnings("checkstyle:AvoidEscapedUnicodeCharacters")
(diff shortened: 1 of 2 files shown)


trustin:
This PR also fixes the off-by-one error :scream:

core/src/main/java/com/linecorp/armeria/common/QueryStringEncoder.java
```diff
@@ -198,15 +199,17 @@ private static int indexOfUnsafeOctet(String s, int start) {
...
 
-    // Do not inline `index` or remove `index < 0` check.
-    // They are specifically written to tell JVM to optimize away the array boundary check.
-    @SuppressWarnings({ "ConstantConditions", "UnnecessaryLocalVariable" })
     private static boolean isUnsafeOctet(char c) {
-        final int index = c;
-        if (index < 0 || index >= SAFE_OCTETS.length) {
-            return true;
+        if (c >= SAFE_OCTETS.length) {
+            return false;
```
ikhoon: The original logic is a bug?
It returns `true` if `index >= SAFE_OCTETS.length`. 🤔

ikhoon: > This PR also fixes the off-by-one error 😱

Is this? 🤣

core/src/main/java/com/linecorp/armeria/common/QueryStringEncoder.java
```diff
@@ -201,7 +201,7 @@ private static int indexOfUnsafeOctet(String s, int start) {
 
     private static boolean isUnsafeOctet(char c) {
         if (c >= SAFE_OCTETS.length) {
-            return false;
+            return true;
```
franz1981: sorry, missed that :O

core/src/main/java/com/linecorp/armeria/common/QueryStringEncoder.java
trustin: Nope. We had to return `true` here :scream:

core/src/main/java/com/linecorp/armeria/common/QueryStringEncoder.java
trustin: :rofl:

core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
```diff
@@ -264,15 +266,17 @@ private static int decodeHexByte(char c1, char c2) {
         return (hi << 4) | lo;
     }
 
-    // Do not inline `index` or remove `index < 0` check.
-    // They are specifically written to tell JVM to optimize away the array boundary check.
-    @SuppressWarnings({ "ConstantConditions", "UnnecessaryLocalVariable" })
     private static int decodeHexNibble(char c) {
-        final int index = c;
-        if (index < 0 || index > OCTETS_TO_HEX.length) {
+        if (c >= OCTETS_TO_HEX.length) {
```
trustin: Off-by-one error here

trustin:
Explicit widening from `char` to `int` seems unnecessary.

trustin:
Quick benchmarking tells me explicit widening helps a little bit actually. Let me update.

ikhoon:
Still LGTM

Issue #1078, opened 2018-03-12
Better interop with other JVM languages


trustin:
Most notably Scala and Kotlin. We could write a sample project that makes sure at least:

- there's no compilation error due to type parameter inference.
- there's no issues with nullness checker.

trustin:
@tobias- IIRC your Armeria-based project is written in Kotlin. It'd be awesome if you can tell us any issues you've encountered so far that makes your code unnecessarily verbose.

tobias-:
Most(all) of my kludges/enforced ugly code circle around lambda functions, especially with decorators. The syntax for them is slightly different in Kotlin, and the compiler seems less able to figure out what generics are needed. Thus the following generics nightmares:
```
interface GrpcDecorator<T> : Function<Service<HttpRequest, HttpResponse>, DecoratingService<HttpRequest, HttpResponse, HttpRequest, HttpResponse>>

class GrpcWebCors(kodein: Kodein) :
    GrpcDecorator<CorsService>,
    Function<Service<HttpRequest, HttpResponse>, DecoratingService<HttpRequest, HttpResponse, HttpRequest, HttpResponse>>
    by
    CorsServiceBuilder
...
        .newDecorator() as Function<Service<HttpRequest, HttpResponse>, DecoratingService<HttpRequest, HttpResponse, HttpRequest, HttpResponse>>
```

```
private fun ServerBuilder.addHeaders(httpsOnly: Boolean) {
    decorator { oldService: Service<HttpRequest, HttpResponse> ->
        object : SimpleDecoratingService<HttpRequest, HttpResponse>(oldService) {
            override fun serve(ctx: ServiceRequestContext, req: HttpRequest): HttpResponse {
                if (httpsOnly) {
                    ctx.addAdditionalResponseHeader("Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload")
                }
                ctx.addAdditionalResponseHeader("x-server-version", GitProperties.version)
                return delegate<Service<HttpRequest, HttpResponse>>().serve(ctx, req)
            }
        }
    }
}
```

I think it may be beyond the scope of Armeria to fix this, although a kotlin-project with just helper(extension) methods for kotlin would be nice. Kotlin excels at embellishing existing classes with new methods. I would prefer to write something like:
```
decorator {
    decorate { ctx, req ->
        if (httpsOnly) {
            ctx.addAdditionalResponseHeader("Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload")
        }
[...]

PR #2335, opened 2019-12-20
Add example:spring-boot-minimal-kotlin

heowc:
Related: #1078, #1880

Motivation:

- An example is provided to users of Kotlin.

Modifications:

- Make an example in Kotlin that looks like `spring-boot-minimal`.

Result:

- Partial fix for #1880
- User-friendliness


heowc:
```
#PROFILE=java8
Caused by: java.lang.UnsupportedClassVersionError: example/springframework/boot/minimal/kotlin/HelloApplicationIntegrationTest has been compiled by a more recent version of the Java Runtime (class file version 56.0), this version of the Java Runtime only recognizes class file versions up to 52.0

#PROFILE=java11
Caused by: java.lang.UnsupportedClassVersionError: example/springframework/boot/minimal/kotlin/HelloApplicationIntegrationTest has been compiled by a more recent version of the Java Runtime (class file version 56.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```
I'm looking for a reason why the test fails in AppVeyor.  :frowning:

heowc:
I didn't know to build only with JDK 12. (I thought it was building with each individual JDK in the AppVeyor.) So I had a problem with `JavaVersion.current()` being set to 12 only. :flushed:

This problem is a simple example provided to the user, so I think it's a good idea to fix it to 1.8 rather than modifying the various settings.

Note that JDK 13 is [not supported](https://kotlinlang.org/docs/reference/using-gradle.html).

examples/spring-boot-minimal-kotlin/build.gradle
```diff
@@ -0,0 +1,47 @@
+buildscript {
```
anuraaga: Since this is a Kotlin example, how about making the build script Kotlin too

examples/spring-boot-minimal-kotlin/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+        gradlePluginPortal()
+    }
+    dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-allopen:${managedVersions['org.jetbrains.kotlin:kotlin-allopen']}"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${managedVersions['org.jetbrains.kotlin:kotlin-gradle-plugin']}"
+        classpath "org.springframework.boot:spring-boot-gradle-plugin:${managedVersions['org.springframework.boot:spring-boot-gradle-plugin']}"
+    }
+}
+
+apply plugin: "org.jetbrains.kotlin.jvm"
```
anuraaga: Let's go ahead and use `plugins` block instead of old style.

examples/spring-boot-minimal-kotlin/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+        jvmTarget = JavaVersion.VERSION_1_8
+    }
+}
+
+compileTestKotlin {
+    kotlinOptions {
+        freeCompilerArgs = ['-Xjsr305=strict']
+        jvmTarget = JavaVersion.VERSION_1_8
+    }
+}
```
anuraaga: Have you applied the Armeria style from [here](https://github.com/line/armeria/tree/master/settings/intellij_idea)? The files are missing newlines and import order isn't our normal one (I'm hoping the code style still affects imports for Kotlin code)

examples/spring-boot-minimal-kotlin/build.gradle
heowc: Good idea :)

heowc: Fixed

examples/spring-boot-minimal-kotlin/build.gradle
heowc: Fixed

examples/spring-boot-minimal-kotlin/build.gradle
heowc: I'm trying to solve it. 🤔

examples/README.md
```diff
@@ -32,6 +32,9 @@
 - `spring-boot-minimal`
   - Learn how to use Armeria with the minimal Spring Boot dependencies.
 
+- `spring-boot-minimal-kotlin`
+  - Learn how to use Armeria with the minimal Spring Boot dependencies written in Kotlin.
```
trustin: `written in Kotlin` -> `(Kotlin)` ?

examples/spring-boot-minimal-kotlin/src/main/kotlin/example/springframework/boot/minimal/kotlin/HelloAnnotatedService.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    @Get("/")
+    fun defaultHello(): String = "Hello, world! Try sending a GET request to /hello/armeria"
+
+    /**
+     * An example in order to show how to use validation framework in an annotated HTTP service.
+     */
+    @Get("/hello/{name}")
+    fun hello(@Size(min = 3, max = 10, message = "name should have between 3 and 10 characters")
+              @Param
```
trustin: How about putting `@Param` first? It will look better.

examples/spring-boot-minimal-kotlin/src/main/kotlin/example/springframework/boot/minimal/kotlin/Main.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.runApplication
+
+@SpringBootApplication
+class Main
+
+fun main(args: Array<String>) {
+    runApplication<Main>(*args)
+}
```
trustin: Question: Is this conventional not to indent `fun main()` in Kotlin?

examples/spring-boot-minimal-kotlin/src/test/kotlin/example/springframework/boot/minimal/kotlin/HelloApplicationIntegrationTest.kt
```diff
@@ -0,0 +1,35 @@
...
+                  .isEqualTo("Hello, Spring! This message is from Armeria annotated service!")
+    }
+
+    @Test
+    fun failure() {
+        val response = client.get("/hello/a").aggregate().join()
+        Assertions.assertThat(response.status()).isEqualTo(HttpStatus.BAD_REQUEST)
+        JsonFluentAssert.assertThatJson(response.contentUtf8())
+                        .node("message")
+                        .isEqualTo("hello.name: name should have between 3 and 10 characters")
```
trustin: No static imports in Kotlin? What's the most popular assertion library in Kotlin ecosystem? Should we keep using AssertJ?

settings/checkstyle/checkstyle.xml
```diff
@@ -6,15 +6,15 @@
 <module name="Checker">
   <property name="charset" value="UTF-8"/>
   <property name="severity" value="error"/>
-  <property name="fileExtensions" value="java, properties, xml"/>
+  <property name="fileExtensions" value="java, kt, properties, xml"/>
```
trustin: Question: Does Checkstyle ever support Kotlin?

examples/spring-boot-minimal-kotlin/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+buildscript {
+    val managedVersions = extra["managedVersions"] as Map<*, *>
+    repositories {
+        gradlePluginPortal()
+    }
+    dependencies {
+        classpath("org.jetbrains.kotlin:kotlin-allopen:${managedVersions["org.jetbrains.kotlin:kotlin-allopen"]}")
```
anuraaga: I just noticed we use `managedVersions` here. @trustin does it make more sense to set the versions inline so the examples can be copy-pasted? It makes updating the plugin dependencies harder, but maybe not a big deal for examples even if some updates get missed.

examples/spring-boot-minimal-kotlin/src/main/kotlin/example/springframework/boot/minimal/kotlin/ValidationExceptionHandler.kt
```diff
@@ -0,0 +1,46 @@
...
+                        cause.message ?: "empty message")
+            }
+        } else ExceptionHandlerFunction.fallthrough()
+    }
+}
+
+/**
+ * A sample HTTP response which is sent to a client when a [ValidationException] is raised.
+ */
+data class ErrorResponse(val error: String,
```
ikhoon: Don't we need [jackson-module-kotlin](https://github.com/FasterXML/jackson-module-kotlin) to convert `data class` to json?

examples/spring-boot-minimal-kotlin/build.gradle.kts
trustin: We do that automatically when copying this to https://github.com/line/armeria-examples

trustin: We are gonna have to update `update-examples.sh` to support this project, though.

anuraaga: Ah thanks :) It even converts to `plugins` block so no problem

examples/spring-boot-minimal-kotlin/src/main/kotlin/example/springframework/boot/minimal/kotlin/ValidationExceptionHandler.kt
heowc: I added dependency [here](https://github.com/line/armeria/pull/2335/files#diff-d19512188cb37597acfa72d049a64f8aR25).

examples/spring-boot-minimal-kotlin/src/test/kotlin/example/springframework/boot/minimal/kotlin/HelloApplicationIntegrationTest.kt
heowc: > No static imports in Kotlin?
Can be used.
> What's the most popular assertion library in Kotlin ecosystem?
It seems that [the official test framework](https://github.com/kotlintest/kotlintest) contains assertions.
> Should we keep using AssertJ?
Most Kotlin assertion libraries have a library of assertj wrappers.

settings/checkstyle/checkstyle.xml
heowc: I have checked that Kotlin related files are verified with checkstyle.

But the main one seems to be [`ktlint`](https://ktlint.github.io/).

examples/spring-boot-minimal-kotlin/src/main/kotlin/example/springframework/boot/minimal/kotlin/Main.kt
heowc: Yes, Kotlin calls it 'top-level functions'. 😮

examples/spring-boot-minimal-kotlin/src/main/kotlin/example/springframework/boot/minimal/kotlin/ValidationExceptionHandler.kt
ikhoon: Does it work without registering module?
```java
val mapper = ObjectMapper().registerModule(KotlinModule())
```

heowc: Oh ... I didn't notice it. 😮 Thanks!

Each test passed, but I think I should check.

settings/checkstyle/checkstyle.xml
heowc: I wonder if it is right to apply checkstyle to the kotlin example. 🤔

minwoox: +1 for following ktlint.

heowc:
I will revise the review later. :blush:

:christmas_tree: Have a Merry Christmas!! :gift_heart:

examples/spring-boot-minimal-kotlin/src/main/kotlin/example/springframework/boot/minimal/kotlin/ValidationExceptionHandler.kt
heowc: I don't know the exact cause, but I fixed it as per the documentation guide. :sweat_smile:

settings/checkstyle/checkstyle.xml
heowc: I applied `ktlint`. However, since it is the [default](https://github.com/pinterest/ktlint#standard-rules), it may be different from the `checkstyle` setting. Do I have to fit it all?

From what I've noticed, it's limited to setting options, so I need to create a [custom ruleset](https://github.com/pinterest/ktlint#creating-a-ruleset).

settings.gradle
```diff
@@ -50,16 +50,17 @@ includeWithFlags ':docs-client'
...
-includeWithFlags ':examples:spring-boot-minimal',    'java'
-includeWithFlags ':examples:spring-boot-tomcat',     'java'
-includeWithFlags ':examples:spring-boot-webflux',    'java'
-includeWithFlags ':examples:static-files',           'java'
+includeWithFlags ':examples:grpc-reactor',                  'java', 'reactor-grpc'
+includeWithFlags ':examples:proxy-server',                  'java'
+includeWithFlags ':examples:saml-service-provider',         'java'
+includeWithFlags ':examples:server-sent-events',            'java'
+includeWithFlags ':examples:spring-boot-minimal',           'java'
+includeWithFlags ':examples:spring-boot-minimal-kotlin',    'java'
```
ikhoon: It is trivial but the `java` flag looks a little weird. If more Kotlin examples and DSL is added, then don't we need `kotlin` flag?

heowc: Do I need to clear the `java` flag and put `kotlin` flag?

To that end, I need to modify [`/gradle/scripts`](https://github.com/line/armeria/tree/master/gradle/scripts). But isn't this file related to [`line/gradle-scripts`](https://github.com/line/gradle-scripts)? 

Or is there another way?

ikhoon: I’m not 100% sure we need to add `kotlin` flag. Does it seems overkill? @trustin @minwoox

examples/spring-boot-minimal-kotlin/build.gradle
heowc: @anuraaga I found it difficult to remove old style. 
It seems possible that [gradle/gradle#9830](https://github.com/gradle/gradle/issues/9830) should be resolved.

settings.gradle
minwoox: I think it's fine because this is just an example and the build.gradle.kts are copied line by line by our script.
But I guess we need it someday because we are doing [this](https://github.com/line/gradle-scripts/blob/master/lib/java.gradle#L14). We might add kotlin.gradle in that directory. 😄

settings/checkstyle/checkstyle.xml
trustin: I think it's fine to use the ktlint default. Kotlin is not Java.

settings.gradle
ikhoon: > I think it's fine because this is just an example and the build.gradle.kts are copied line by line by our script.

That sounds good to me. Let leave it as it is. :-)

heowc:
Thanks for the passionate review of the new example! :laughing:

trustin:
Thanks, @heowc and reviewers! :tada:

trustin:
By the way, what about adding other Kotlin examples as well? :wink:

examples/spring-boot-minimal-kotlin/src/main/kotlin/example/springframework/boot/minimal/kotlin/Main.kt
tobias-: `main()` is NOT part of the `Main` class with this syntax. `main()` is part of the "generated" `<filename>Kt` class i.e. in this case `MainKt.class`.

heowc: Yes. So I wrote [this](https://github.com/line/armeria/blob/master/examples/grpc-kotlin/build.gradle.kts#L22) when creating `grpc-kotlin`. 😄 

I just written it as guided by [spring-boot](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-kotlin-api-runapplication).

PR #2336, opened 2019-12-20
Make `QueryString{Encoder,Decoder}` even faster

trustin:
Motivation:

@nitsanw and @franz1981 came up with an even faster way to convert a
character into a primitive value:

- Allocate an array as long as 65536.
- Use a character as an index.
  - No need for boundary check at all because a `char` never exceeds
    65535 by definition.
  - Increased cache miss may be a concern, but I guess it will not
    matter because most characters will be in ASCII range anyway.

Modifications:

- Use a 65536-byte long lookup table for decoding a hexdecimal nibble
  and determining if a character needs to be percent-encoded.

Result:

- Nice performance boost

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
@@ -42,7 +42,7 @@ final class QueryStringDecoder {
 
     @SuppressWarnings("checkstyle:AvoidEscapedUnicodeCharacters")
     private static final char UNKNOWN_CHAR = '\uFFFD';
-    private static final byte[] OCTETS_TO_HEX = new byte[256];
+    private static final byte[] OCTETS_TO_HEX = new byte[Character.MAX_VALUE + 1];
 
     static {
         Arrays.fill(OCTETS_TO_HEX, (byte) -1);
@@ -259,8 +259,8 @@ final class QueryStringDecoder {
     }
 
     private static int decodeHexByte(char c1, char c2) {
-        final int hi = OCTETS_TO_HEX[c1 & 0xFF];
-        final int lo = OCTETS_TO_HEX[c2 & 0xFF];
+        final int hi = OCTETS_TO_HEX[c1];
+        final int lo = OCTETS_TO_HEX[c2];
         return (hi << 4) | lo;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
@@ -42,7 +42,7 @@ final class QueryStringDecoder {
 
     @SuppressWarnings("checkstyle:AvoidEscapedUnicodeCharacters")
     private static final char UNKNOWN_CHAR = '\uFFFD';
-    private static final byte[] OCTETS_TO_HEX = new byte[Character.MAX_VALUE + 1];
+    private static final byte[] OCTETS_TO_HEX = new byte[256];
 
     static {
         Arrays.fill(OCTETS_TO_HEX, (byte) -1);
@@ -259,8 +259,8 @@ final class QueryStringDecoder {
     }
 
     private static int decodeHexByte(char c1, char c2) {
-        final int hi = OCTETS_TO_HEX[c1];
-        final int lo = OCTETS_TO_HEX[c2];
+        final int hi = OCTETS_TO_HEX[c1 & 0xFF];
+        final int lo = OCTETS_TO_HEX[c2 & 0xFF];
         return (hi << 4) | lo;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
+++ b/core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
@@ -38,13 +38,11 @@ import com.google.common.annotations.VisibleForTesting;
 
 import com.linecorp.armeria.internal.TemporaryThreadLocals;
 
-import io.netty.util.internal.PlatformDependent;
-
 final class QueryStringDecoder {
 
     @SuppressWarnings("checkstyle:AvoidEscapedUnicodeCharacters")
     private static final char UNKNOWN_CHAR = '\uFFFD';
-    private static final byte[] OCTETS_TO_HEX = new byte['f' + 1];
+    private static final byte[] OCTETS_TO_HEX = new byte[Character.MAX_VALUE + 1];
 
     static {
         Arrays.fill(OCTETS_TO_HEX, (byte) -1);
@@ -261,25 +259,9 @@ final class QueryStringDecoder {
     }
 
     private static int decodeHexByte(char c1, char c2) {
-        final int hi = decodeHexNibble(c1);
-        final int lo = decodeHexNibble(c2);
+        final int hi = OCTETS_TO_HEX[c1];
+        final int lo = OCTETS_TO_HEX[c2];
         return (hi << 4) | lo;
-    }
-
-    private static int decodeHexNibble(char c) {
-        // Widen explicitly so that JVM doesn't have to widen many times.
-        @SuppressWarnings("UnnecessaryLocalVariable")
-        final int index = c;
-        if (index >= OCTETS_TO_HEX.length) {
-            return -1;
-        }
-
-        if (PlatformDependent.hasUnsafe()) {
-            // Avoid boundary checks
-            return PlatformDependent.getByte(OCTETS_TO_HEX, index);
-        }
-
-        return OCTETS_TO_HEX[index];
     }
 
     private static boolean isContinuation(int b) {
--- a/core/src/main/java/com/linecorp/armeria/common/QueryStringEncoder.java
+++ b/core/src/main/java/com/linecorp/armeria/common/QueryStringEncoder.java
@@ -55,7 +55,7 @@ final class QueryStringEncoder {
 
     private static final char[] UTF_UNKNOWN = { '%', '3', 'F' }; // Percent encoded question mark
     private static final char[] UPPER_HEX_DIGITS = "0123456789ABCDEF".toCharArray();
-    private static final byte[] SAFE_OCTETS = new byte[256];
+    private static final byte[] SAFE_OCTETS = new byte[Character.MAX_VALUE + 1];
 
     static {
         final String safeOctetStr = "-_.*abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
@@ -76,7 +76,8 @@ final class QueryStringEncoder {
     private static StringBuilder encodeComponent(StringBuilder buf, String s) {
         final int len = s.length();
         for (int i = 0; i < len; i++) {
-            if (isUnsafeOctet(s.charAt(i))) {
+            final char c = s.charAt(i);
+            if (SAFE_OCTETS[c] == 0) {
                 if (i != 0) {
                     buf.append(s, 0, i);
                 }
@@ -182,16 +183,13 @@ final class QueryStringEncoder {
     private static int indexOfUnsafeOctet(String s, int start) {
         final int len = s.length();
         for (int i = start; i < len; i++) {
-            if (isUnsafeOctet(s.charAt(i))) {
+            final char c = s.charAt(i);
+            if (SAFE_OCTETS[c] == 0) {
                 return i;
             }
         }
 
         return -1;
-    }
-
-    private static boolean isUnsafeOctet(char c) {
-        return SAFE_OCTETS[c & 0xFF] == 0;
     }
 
     private QueryStringEncoder() {}
--- a/core/src/main/java/com/linecorp/armeria/common/QueryStringEncoder.java
+++ b/core/src/main/java/com/linecorp/armeria/common/QueryStringEncoder.java
@@ -55,7 +55,7 @@ final class QueryStringEncoder {
 
     private static final char[] UTF_UNKNOWN = { '%', '3', 'F' }; // Percent encoded question mark
     private static final char[] UPPER_HEX_DIGITS = "0123456789ABCDEF".toCharArray();
-    private static final byte[] SAFE_OCTETS = new byte[Character.MAX_VALUE + 1];
+    private static final byte[] SAFE_OCTETS = new byte[256];
 
     static {
         final String safeOctetStr = "-_.*abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
@@ -76,8 +76,7 @@ final class QueryStringEncoder {
     private static StringBuilder encodeComponent(StringBuilder buf, String s) {
         final int len = s.length();
         for (int i = 0; i < len; i++) {
-            final char c = s.charAt(i);
-            if (SAFE_OCTETS[c] == 0) {
+            if (isUnsafeOctet(s.charAt(i))) {
                 if (i != 0) {
                     buf.append(s, 0, i);
                 }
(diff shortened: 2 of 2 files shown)


trustin:
Before (unsafe array access):
```
Benchmark                                    Mode  Cnt       Score      Error  Units
QueryStringDecoderBenchmark.armeriaAscii    thrpt   50  377588.237 ± 5587.696  ops/s
QueryStringDecoderBenchmark.armeriaLong     thrpt   50   19582.624 ±  457.648  ops/s
QueryStringDecoderBenchmark.armeriaMixed    thrpt   50  119342.608 ± 2209.684  ops/s
QueryStringDecoderBenchmark.armeriaUnicode  thrpt   50   86064.748 ± 2097.511  ops/s
QueryStringEncoderBenchmark.armeriaAscii    thrpt   50  844493.147 ± 1773.086  ops/s
QueryStringEncoderBenchmark.armeriaLong     thrpt   50   21028.206 ±  927.736  ops/s
QueryStringEncoderBenchmark.armeriaMixed    thrpt   50  226818.284 ± 7212.655  ops/s
QueryStringEncoderBenchmark.armeriaUnicode  thrpt   50  216179.147 ± 2157.503  ops/s
```

After (64k lookup table):
```
Benchmark                                    Mode  Cnt       Score      Error  Units
QueryStringDecoderBenchmark.armeriaAscii    thrpt   50  385508.751 ± 5165.823  ops/s
QueryStringDecoderBenchmark.armeriaLong     thrpt   50   20252.521 ±  434.238  ops/s
QueryStringDecoderBenchmark.armeriaMixed    thrpt   50  129201.709 ± 4710.201  ops/s
QueryStringDecoderBenchmark.armeriaUnicode  thrpt   50   88696.118 ±  418.647  ops/s
QueryStringEncoderBenchmark.armeriaAscii    thrpt   50  890959.585 ± 7127.986  ops/s
QueryStringEncoderBenchmark.armeriaLong     thrpt   50   22346.938 ±  527.001  ops/s
QueryStringEncoderBenchmark.armeriaMixed    thrpt   50  227038.288 ±  926.122  ops/s
QueryStringEncoderBenchmark.armeriaUnicode  thrpt   50  212527.522 ± 2766.579  ops/s
```

Notes:
- `Decoder.armeriaAscii()` shouldn't be affected by this change.
- `Encoder.armeriaAscii()` should be most-affected by this change.

The result is somewhat noisy but `Decoder.armeriaUnicode()` and `Encoder.armeriaAscii()` show obvious win, so I think we can say this change improves throughput.

What's interesting is that we lost for `Encoder.armeriaUnicode()`, which might be due to increased cache miss.

core/src/main/java/com/linecorp/armeria/common/QueryStringDecoder.java
```diff
@@ -261,27 +259,11 @@ private static String decodeUtf8Component(TemporaryThreadLocals tempThreadLocals
     }
 
     private static int decodeHexByte(char c1, char c2) {
-        final int hi = decodeHexNibble(c1);
-        final int lo = decodeHexNibble(c2);
+        final int hi = OCTETS_TO_HEX[c1];
```
anuraaga: Don't you still want to use PlatformDependent throughout this PR to avoid automatic bounds check too? Maybe we can have a method that accepts char, `assert` array length is less than 64K, and then do the unsafe lookup

trustin: JVM already eliminates the automatic bounds check because it knows c1 and c2 is always < 65536 and `OCTETS_TO_HEX.length` is 65536.

anuraaga: Checked printed assembly and confirmed both indexing and getByte produce the same code!

anuraaga: Sorry I looked again - I think I see a bounds check at `0x000001ab07adaa2c:   cmp    $0x10000,%edx`. Unsafe version doesn't have cmp or any fallback codepath.

Array index

```
============================= C2-compiled nmethod ==============================
<nmethod compile_id='39' compiler='c2' entry='0x000002e5c6c28aa0' size='664' address='0x000002e5c6c28910' relocation_offset='360' insts_offset='400' stub_offset='496' scopes_data_offset='544' scopes_pcs_offset='576' dependencies_offset='656' oops_offset='520' metadata_offset='536' method='com.linecorp.armeria.jvm.ArraysBenchmark arrayIndex (C)B' bytes='6' count='5179' backedge_count='1' iicount='100161' stamp='0.340'/>
<print_nmethod compile_id='39' compiler='c2' stamp='0.340'>
----------------------------------- Assembly -----------------------------------

Compiled method (c2)     340   39             com.linecorp.armeria.jvm.ArraysBenchmark::arrayIndex (6 bytes)
 total in heap  [0x000002e5c6c28910,0x000002e5c6c28ba8] = 664
 relocation     [0x000002e5c6c28a78,0x000002e5c6c28a88] = 16
 main code      [0x000002e5c6c28aa0,0x000002e5c6c28b00] = 96
 stub code      [0x000002e5c6c28b00,0x000002e5c6c28b18] = 24
 oops           [0x000002e5c6c28b18,0x000002e5c6c28b28] = 16
 metadata       [0x000002e5c6c28b28,0x000002e5c6c28b30] = 8
 scopes data    [0x000002e5c6c28b30,0x000002e5c6c28b50] = 32
 scopes pcs     [0x000002e5c6c28b50,0x000002e5c6c28ba0] = 80
 dependencies   [0x000002e5c6c28ba0,0x000002e5c6c28ba8] = 8

--------------------------------------------------------------------------------
[Constant Pool (empty)]

--------------------------------------------------------------------------------

[Verified Entry Point]
  # {method} {0x000002e5ce762728} &apos;arrayIndex&apos; &apos;(C)B&apos; in &apos;com/linecorp/armeria/jvm/ArraysBenchmark&apos;
  # parm0:    rdx       = char
  #           [sp+0x20]  (sp of caller)
  0x000002e5c6c28aa0:   mov    %eax,-0x7000(%rsp)
  0x000002e5c6c28aa7:   push   %rbp
  0x000002e5c6c28aa8:   sub    $0x10,%rsp                   ;*synchronization entry
                                                            ; - com.linecorp.armeria.jvm.ArraysBenchmark::arrayIndex@-1 (line 55)
  0x000002e5c6c28aac:   cmp    $0x10000,%edx
  0x000002e5c6c28ab2:   jae    0x000002e5c6c28ad4
  0x000002e5c6c28ab4:   movabs $0xefc18840,%r10             ;   {oop([B{0x00000000efc18840})}
  0x000002e5c6c28abe:   movsbl 0x10(%r10,%rdx,1),%eax
  0x000002e5c6c28ac4:   add    $0x10,%rsp
  0x000002e5c6c28ac8:   pop    %rbp
  0x000002e5c6c28ac9:   mov    0x108(%r15),%r10
  0x000002e5c6c28ad0:   test   %eax,(%r10)                  ;   {poll_return}
  0x000002e5c6c28ad3:   retq
  0x000002e5c6c28ad4:   mov    %edx,%ebp
  0x000002e5c6c28ad6:   mov    $0xffffffe4,%edx
  0x000002e5c6c28adb:   callq  0x000002e5c6be9080           ; ImmutableOopMap {}
                                                            ;*baload {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - com.linecorp.armeria.jvm.ArraysBenchmark::arrayIndex@4 (line 55)
                                                            ;   {runtime_call UncommonTrapBlob}
  0x000002e5c6c28ae0:   ud2                                 ;*baload {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - com.linecorp.armeria.jvm.ArraysBenchmark::arrayIndex@4 (line 55)
  0x000002e5c6c28ae2:   hlt

```

Unsafe version

```
============================= C2-compiled nmethod ==============================
<nmethod compile_id='233' compiler='c2' entry='0x0000019c552aafa0' size='632' address='0x0000019c552aae10' relocation_offset='360' insts_offset='400' stub_offset='464' scopes_data_offset='528' scopes_pcs_offset='560' dependencies_offset='624' oops_offset='488' metadata_offset='496' method='com.linecorp.armeria.jvm.ArraysBenchmark unsafeArrayLookup (C)B' bytes='8' count='5400' backedge_count='1' iicount='60390' stamp='2.793'/>
<print_nmethod compile_id='233' compiler='c2' stamp='2.793'>
----------------------------------- Assembly -----------------------------------

Compiled method (c2)    2792  233             com.linecorp.armeria.jvm.ArraysBenchmark::unsafeArrayLookup (8 bytes)
 total in heap  [0x0000019c552aae10,0x0000019c552ab088] = 632
 relocation     [0x0000019c552aaf78,0x0000019c552aaf88] = 16
 main code      [0x0000019c552aafa0,0x0000019c552aafe0] = 64
 stub code      [0x0000019c552aafe0,0x0000019c552aaff8] = 24
 oops           [0x0000019c552aaff8,0x0000019c552ab000] = 8
 metadata       [0x0000019c552ab000,0x0000019c552ab020] = 32
 scopes data    [0x0000019c552ab020,0x0000019c552ab040] = 32
 scopes pcs     [0x0000019c552ab040,0x0000019c552ab080] = 64
 dependencies   [0x0000019c552ab080,0x0000019c552ab088] = 8

--------------------------------------------------------------------------------
[Constant Pool (empty)]

--------------------------------------------------------------------------------

[Verified Entry Point]
  # {method} {0x0000019c5c5428a8} &apos;unsafeArrayLookup&apos; &apos;(C)B&apos; in &apos;com/linecorp/armeria/jvm/ArraysBenchmark&apos;
  # parm0:    rdx       = char
  #           [sp+0x20]  (sp of caller)
  0x0000019c552aafa0:   sub    $0x18,%rsp
  0x0000019c552aafa7:   mov    %rbp,0x10(%rsp)              ;*synchronization entry
                                                            ; - com.linecorp.armeria.jvm.ArraysBenchmark::unsafeArrayLookup@-1 (line 64)
  0x0000019c552aafac:   movslq %edx,%r10
  0x0000019c552aafaf:   movabs $0xee300000,%r11             ;   {oop([B{0x00000000ee300000})}
  0x0000019c552aafb9:   movsbl 0x10(%r11,%r10,1),%eax       ;*invokevirtual getByte {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - sun.misc.Unsafe::getByte@5 (line 230)
                                                            ; - io.netty.util.internal.PlatformDependent0::getByte@10 (line 525)
                                                            ; - io.netty.util.internal.PlatformDependent::getByte@2 (line 503)
                                                            ; - com.linecorp.armeria.jvm.ArraysBenchmark::unsafeArrayLookup@4 (line 64)
  0x0000019c552aafbf:   add    $0x10,%rsp
  0x0000019c552aafc3:   pop    %rbp
  0x0000019c552aafc4:   mov    0x108(%r15),%r10
  0x0000019c552aafcb:   test   %eax,(%r10)                  ;   {poll_return}
  0x0000019c552aafce:   retq
  0x0000019c552aafcf:   hlt
  0x0000019c552aafd0:   hlt
  0x0000019c552aafd1:   hlt
  0x0000019c552aafd2:   hlt
```

Code
```
package com.linecorp.armeria.jvm;

import java.util.Arrays;
import java.util.Random;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;

import io.netty.util.internal.PlatformDependent;

/**
 * Benchmarks to check JVM optimizations of using arrays.
 */
@State(Scope.Benchmark)
public class ArraysBenchmark {

    private static final byte[] LARGER_THAN_SIXTEEN_BITS = new byte[Character.MAX_VALUE + 1];

    static {
        Arrays.fill(LARGER_THAN_SIXTEEN_BITS, (byte) -1);
    }

    private char index;

    @Setup
    public void setUp() {
        // Use fixed seed random to create an int, the JVM should have no way of constant-folding.
        index = (char) new Random().nextInt(Character.MAX_VALUE + 1);
    }

    @Benchmark
    public byte arrayIndex() {
        return arrayIndex(index);
    }

    private static byte arrayIndex(char index) {
        return LARGER_THAN_SIXTEEN_BITS[index];
    }

    @Benchmark
    public byte unsafeArrayLookup() {
        return unsafeArrayLookup(index);
    }

    private static byte unsafeArrayLookup(char index) {
        return PlatformDependent.getByte(LARGER_THAN_SIXTEEN_BITS, index);
    }

    @SuppressWarnings("checkstyle:UncommentedMain")
    public static void main(String[] args) {
        final ArraysBenchmark benchmark = new ArraysBenchmark();
        byte res = 0;
        for (int i = 0; i < 100_000_000; i++) {
            benchmark.setUp();
            res = benchmark.arrayIndex();
        }
        System.out.println(res);
    }
}
```

franz1981: Why not using `-prof perfasm` directly on JMH? It should be the same eh...just curious

franz1981: I see this too:
```
// Use fixed seed random to create an int, the JVM should have no way of constant-folding.
```
It isn't necessary because the JVM trust just `final static` fields to perform constant folding (or `final` on the stack, or inside special classes/captured effective-final on lambdas...)

anuraaga: I'm currently on Windows so it'd take some extra effort to set up perf, at least I think so 😆

franz1981: Let me take a look on `DecodeHexBenchmark.hexDigitsSafeNoChecks` and `DecodeHexBenchmark.hexDigitsUnsafeNoChecks` ASM (on https://github.com/netty/netty/pull/9896)  what's going on

anuraaga: You're probably right about the folding, I just wanted to make extra-sure for now since just looking at the assembly and not checking benchmark numbers yet.

FWIW, if I'm not mistaken, when indexing from a `char` _field_, there isn't a bounds check but when indexing from a `char` method _parameter_, there is. I initially thought the code was the same before splitting out a method with `char` param to better match the code, and then I got the above. It does seem to "know" the parameter is char (`# parm0:    rdx       = char`), but still has a jump.

anuraaga: By the way, I'm literally searching through `hotspot_pid*.log` in less to check the assembly as my jitwatch also tells me to enable class loading trace, even though it is enabled... So I could be misreading something for sure. But one takeaway may be that as JVM optimizations are a black box, if we have a more direct method `PlatformDependent.getByte` it probably doesn't hurt to just use it :) Actually, if `PlatformDependent.getByte` did the `hasUnsafe` check the code doesn't really get cluttered either, would that be a reasonable change (`static final boolean` should be removed completely even by C1)? Actually I'm not sure what's platform dependent about that method ;)

franz1981: That's strange: you can try verify this by not making `arrayIndex` to be inlined using `@CompilerControl(Mode.DONT_INLINE)` on it

franz1981: I think that Android miss `Unsafe` in the form we use it for other JVMs: my concern is that unsafe manipulation of arrays (or any field in an proper Java Object) *could* prevent certain nice optimizations related vectorization due to the introduction of compiler barriers: I don't remember the JEP and if has been solved, but I remember wasn't something 100% safe to be assumed. Anyway the numbers are not so different and is yet to be seen with the core cache filled of other stuff, in a real case, how much this optimization will make difference

franz1981:
@anuraaga any news re https://github.com/line/armeria/pull/2336#discussion_r360785367? I still don't know if you have found that we have unwanted bounds checks somewhere: in my experiments it doesn't seem the case....

anuraaga:
@franz1981 I haven't done any more investigation and am not planning to right now since not sure what else to try :)

I was just printing assembly for the code snippet I pasted, which matches the pattern for hex decide, and saw the branch. This doesn't surprise me since bytecode shouldn't see any difference between int and char for local variables on the stack, everything is just an int at the bytecode level (I'm guessing the char param comment I see in the output is from the classloader trace and not the JIT). 

As in practice the bounds check will always pass, I guess it's hard to write a benchmark since the branch will always be correctly predicted. That's probably enough reason to not worry too much.

Issue #2025, opened 2019-08-24
Propagate client TLS errors to users


anuraaga:
Currently, client TLS errors are logged but the user just gets a `ClosedSessionException` with no message or stack trace. TLS errors are generally not transient and are a result of a configuration error that the user needs to fix so we should propagate them as the cause of the exception.

Related https://github.com/openzipkin/zipkin/issues/2776

trustin:
@jrhee17 and this one

trustin:
Fixed via #2338

trustin:
Thanks again, @jrhee17 !

PR #2338, opened 2019-12-21
Propagate client TLS errors to user

jrhee17:
Fixes (#2025)

**Motivation**

Pointing https requests at a plaintext endpoint results in a general `ClosedSessionException`.
Explicitly throwing an exception may help clients handle configuration errors better

**Modifications**

- Introduce a `PENDING_EXCEPTION` channel attribute which can be thrown from `HttpSessionHandler`
- Remove TLS handshake logging

**Related PR** (#1297)


jrhee17:
I forgot to make this a draft -- will do next time :D

---

On a separate note, I have a question about future works

Can reviewers take a quick look at the current implementation?
I don't think there should be much performance overhead since the only additional cost is 1) some bit manipulations 2) adding and removing a handler.
Still, I don't like the idea of checking this for every http request (which I think will mostly be false anyways)

I have a couple ideas -- can you guys point me in the right direction?
1) leave as is -- check if ssl error code is returned for every client request
2) Make this specific to healthcheck requests only (I'll have to check if this is possible though..)
3) Introduce a new `debug` mode which checks enables this check

Let me know what you guys think! Thanks in advance:)

trustin:
I'm not sure if the current approach is right because:

- Another handler in Netty pipeline = more overhead
- It's more elegant to solve this at higher level.

IIUC, the initial connection-level errors will be handled in [this class](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java#L182) by catching an exception like `SSLException` (and maybe any other exception types as well?). Please try setting some breakpoints in `userEventTriggeres()` and `exceptionCaught()` and play with some connection failures to figure out what exceptions are raised but not propagated.

Let's assume that we captured the exception, then how could we pass that to a user client? I didn't look closely, but it all depends on:

- What exception a client currently gets.
- How the exception is raised, e.g. stack trace.

I'm not very certain, so you'll have to check by yourself, but if the exception is raised by `failUnfinishedResponses(ClosedSessionException.get());` in `HttpSessionHandler.channelInactive()`, we could pass the connection level exception via `Channel.attr(...)` and use that exception in lieu of `ClosedSessionException`.

Please feel free to let me know if you are stuck with playing with this idea.

jrhee17:
My bad -- I was thinking the other way around (http request to ssl protected endpoint) 😵 
I've made some changes following your advice! 🙇

core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
```diff
@@ -311,12 +315,17 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
             channelPool.connect(channel.remoteAddress(), H1C, sessionPromise);
         } else {
             // Fail all pending responses.
-            failUnfinishedResponses(ClosedSessionException.get());
-
+            final Throwable throwable;
+            if (ctx.channel().hasAttr(NESTED_THROWABLE)) {
+                throwable = ClosedSessionException.get(ctx.channel().attr(NESTED_THROWABLE).get());
```
trustin: Perhaps we can just use the exception as is rather than wrapping this something?

core/src/main/java/com/linecorp/armeria/common/ClosedSessionException.java
```diff
@@ -33,8 +33,20 @@ public static ClosedSessionException get() {
                new ClosedSessionException() : INSTANCE;
     }
 
+    /**
+     * Returns a {@link ClosedSessionException} with a cause.
+     */
+    public static ClosedSessionException get(Throwable cause) {
```
trustin: If we don't wrap, we don't need to add this method.

core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
```diff
@@ -64,6 +65,9 @@
      */
     private static final int MAX_NUM_REQUESTS_SENT = 536870912;
 
+    static final AttributeKey<Throwable> NESTED_THROWABLE =
+            AttributeKey.valueOf(HttpSessionHandler.class, "NESTED_THROWABLE");
```
trustin: How about `PENDING_EXCEPTION`?

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -767,6 +769,20 @@ void responseAbortWithException() throws InterruptedException {
                 .hasCause(badState);
     }
 
+    @Test
+    void httpsRequstToPlainTextEndpoint() throws Exception {
+        final ClientFactory clientFactory =
+                ClientFactory.builder()
+                             .sslContextCustomizer(
+                                     s -> s.trustManager(InsecureTrustManagerFactory.INSTANCE))
+                             .build();
```
trustin: Now you can just use `ClientFactory.insecure()`. :wink:

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -767,6 +769,20 @@ void responseAbortWithException() throws InterruptedException {
...
+        final WebClient client = WebClient.builder(SessionProtocol.HTTPS, newEndpoint())
+                                          .factory(clientFactory).build();
+        assertThatThrownBy(() -> client.get("/hello/world").aggregate().get())
+                .hasRootCauseInstanceOf(SSLException.class);
+        clientFactory.close();
```
trustin: and then we don't need to close anything here.

jrhee17:
done! thanks for the `kind && excellent` comments as always :D

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -767,6 +768,14 @@ void responseAbortWithException() throws InterruptedException {
                 .hasCause(badState);
     }
 
+    @Test
+    void httpsRequstToPlainTextEndpoint() throws Exception {
```
anuraaga: Request - Spelling

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
```diff
@@ -767,6 +768,14 @@ void responseAbortWithException() throws InterruptedException {
...
+        final WebClient client = WebClient.builder(SessionProtocol.HTTPS, newEndpoint())
+                                          .factory(ClientFactory.insecure()).build();
+        assertThatThrownBy(() -> client.get("/hello/world").aggregate().get())
```
anuraaga: Can we also test for hitting an ssl endpoint without a proper cert? I think making server self-signed and not setting the clientfactory to insecure should work.

trustin:
@jrhee17 Could you address @anuraaga's comments?

core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
```diff
@@ -231,7 +231,7 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
                 if (handshakeFailed &&
                     cause instanceof DecoderException &&
                     cause.getCause() instanceof SSLException) {
-                    // Swallow an SSLException raised after handshake failure.
+                    ctx.channel().attr(PENDING_EXCEPTION).set(cause);
```
trustin: Last but not least - How about `new UnprocessedRequestException(cause.getCause())`?

jrhee17: Q. Naively implementing this causes nested `UnprocessedRequestException`s due to the following line

https://github.com/line/armeria/blob/c260bfee7a68abbb5a6ca90814ed23d2e1f38f05/core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java#L458

Personally, I think it is ok to keep all exception wrapping logic at the `completeExceptionally()` layer in `HttpChannelPool`. 🤔  Let me know if I'm not understanding anything  correctly. 🙇

core/src/test/java/com/linecorp/armeria/client/HttpClientIntegrationTest.java
jrhee17: excellent point. done 👍

core/src/main/java/com/linecorp/armeria/client/HttpClientPipelineConfigurator.java
trustin: That sounds good to me.

trustin: `.set(cause.getCause()` because we are interested in `SSLException`, not `DecoderException`.

PR #2340, opened 2019-12-23
Add and clean up TLS configuration methods

### LINKED ISSUE #2325: SessionProtocolNegotiationException when HTTP/2 is enforced
Hi guys,

I tried to run 4 cases on the same private key and certificate - just a basic ping request / response over TLS.  Test setup is also straightforward - create client and server and then call remote 'ping' method in the loop 10*1000 times. 

1. Bare gRPC / Netty client and server takes ~ 1ms on average locally
2. Bare gRPC / Netty server with Armeria client (h2 protocol) also takes ~ 1ms on average locally
3. Armeria server with Armeria client (https protocol) takes ~ 3ms on average locally and 8ms in AWS us-east-1 - unacceptable performance for my case. I guess this is because the communication protocol is downgraded to h1. 
4. Armeria server with Armeria client (h2 protocol) fails with "SessionProtocolNegotiationException: expected: h2, actual: h1, reason: unexpected protocol negotiation result"  

Find it really surprising that Armeria can't talk to Armeria when H2 is enforced but Armeria client works fine with gRPC server and gives reasonable latency (2nd case). 

The server is configured like below: 

```java
 byte[] certPem = Files.readAllBytes(new File("/tmp/cert.pem").toPath());
 byte[] privPem = Files.readAllBytes(new File("/tmp/key.pem").toPath());
 ServerBuilder serverBuilder  =  Server.builder()
          .service(new GrpcServiceBuilder().addService(new PingService()).build());
 try (InputStream cert = new ByteArrayInputStream(certPem);
        InputStream priv = new ByteArrayInputStream(privPem))
 {
      serverBuilder = serverBuilder   
          .https(port)
          .tls(SslContextBuilder.forServer(cert, priv).build());
 }
 return serverBuilder.build();
```

Dependencies closure:
 - Armeria - 0.97
 - gRPC - 1.25
 - Netty - 4.1.43.Final
 - netty-tcnative-boringssl-static - 2.0.26.Final 

Sincerely,
Andrey

PS. Sorry for raising few issues recently - really like Armeria's model and want to use it in my projects.

trustin:
Motivation:

- `{Server,VirtualHost}Builder.tls(SslContext)` opens a hole which
  allows a user to specify a misconfigured `SslContext`, causing an
  unexpected server behavior, such as what's described in #2325.
- `tls()` methods do not cover all ways to build server-side `SslContext`,
  such as `SslContextBuilder.forServer(InputStream, InputStream)` and
  `forServer(PrivateKey, X509Certificate...)`.

Modifications:

- Add `{Server,VirtualHost}Builder.tlsCustomizer(Consumer)}`, so we do
  not have to overload all `tls()` methods with an additional `Consumer`
  parameter.
  - As a result, the instantiation of `SslContext` has been deferred
    again to the last `build()` call, and thus `tls()` methods do not
    have a `throws SSLException` clause anymore.
- Deprecate all `tls()` methods that require `Consumer tlsCustomizer`.
- Add more `tls()` methods to expose all ways to create an `SslContextBuilder`.
- Make sure a user can customize the `SslContext` even when it's built
  via `tlsSelfSigned()`
- Make client-side changes for consistency:
  - `ClientFactoryOption.SSL_CONTEXT_CUSTOMIZER` has been deprecated in
    favor of `TLS_CUSTOMIZER`.
  - `ClientFactoryBuilder.sslContextCustomizer()` has been deprecated in
    favor of `tlsCustomizer()`.
    - `tlsCustomizer()` is additive, unlike `sslContextCustomizer()`.

Result:

- A user is given with more safe ways to construct an `SslContext`.
- A user can customize an `SslContext` via `tlsCustomizer()` for both
  client and server.
- A user can now customize an `SslContext` via `tlsCustomizer()` even
  when it was created via `tlsSelfSigned()`.
- Closes #2325
- (Breaking) `{Server,VirtualHost}Builder.tls()` methods do not throw an
  `SSLException` anymore. `build()` will throw an `IllegalStateException`
  instead. As a result, any SSL configuration failure will be known when
  a user calls `build()`, rather than `tls()`.
- (Deprecation) `tls()` method that require `SslContext` has been
  deprecated without a replacement, because it's not safe.
  - This should be OK, because we provide the same set of `tls()`
    methods with what's provided by `SslContextBuilder.forServer()`.
    For example, a user can use `tls(InputStream, InputStream)` in #2325.
- (Deprecation) `tls()` methods that require `Consumer tlsCustomizer`
  have been deprecated in favor of `tlsCustomizer()`, which helps us
  keep the number of overloaded `tls()` methods under control.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryOption.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryOption.java
@@ -75,8 +75,18 @@ public final class ClientFactoryOption<T> extends AbstractOption<T> {
      * The {@link Consumer} which can arbitrarily configure the {@link SslContextBuilder} that will be
      * applied to the SSL session.
      */
+    public static final ClientFactoryOption<Consumer<? super SslContextBuilder>> TLS_CUSTOMIZER =
+            valueOf("TLS_CUSTOMIZER");
+
+    /**
+     * The {@link Consumer} which can arbitrarily configure the {@link SslContextBuilder} that will be
+     * applied to the SSL session.
+     *
+     * @deprecated Use {@link #TLS_CUSTOMIZER}.
+     */
+    @Deprecated
     public static final ClientFactoryOption<Consumer<? super SslContextBuilder>> SSL_CONTEXT_CUSTOMIZER =
-            valueOf("SSL_CONTEXT_CUSTOMIZER");
+            TLS_CUSTOMIZER;
 
     /**
      * The factory that creates an {@link AddressResolverGroup} which resolves remote addresses into
--- a/core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
@@ -31,6 +31,7 @@ import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.security.PrivateKey;
+import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.time.Duration;
 import java.util.ArrayList;
@@ -98,6 +99,8 @@ public final class VirtualHostBuilder {
     private SslContextBuilder sslContextBuilder;
     @Nullable
     private Boolean tlsSelfSigned;
+    @Nullable
+    private SelfSignedCertificate selfSignedCertificate;
     private final List<Consumer<? super SslContextBuilder>> tlsCustomizers = new ArrayList<>();
     private final List<RouteDecoratingService> routeDecoratingServices = new ArrayList<>();
     @Nullable
@@ -1018,11 +1021,16 @@ public final class VirtualHostBuilder {
         SslContext sslContext = null;
         boolean releaseSslContextOnFailure = false;
         try {
+            // Whether the `SslContext` came (or was created) from this `VirtualHost`'s properties.
+            boolean sslContextFromThis = false;
+
             // Build a new SslContext or use a user-specified one for backward compatibility.
             if (this.sslContext != null) {
                 sslContext = this.sslContext;
+                sslContextFromThis = true;
             } else if (sslContextBuilder != null) {
                 sslContext = buildSslContext(sslContextBuilder, tlsCustomizers);
+                sslContextFromThis = true;
                 releaseSslContextOnFailure = true;
             } else if (template.sslContext != null) {
                 sslContext = template.sslContext;
@@ -1038,6 +1046,7 @@ public final class VirtualHostBuilder {
                 if (this.tlsSelfSigned != null) {
                     tlsSelfSigned = this.tlsSelfSigned;
                     tlsCustomizers = this.tlsCustomizers;
+                    sslContextFromThis = true;
                 } else {
                     tlsSelfSigned = template.tlsSelfSigned;
                     tlsCustomizers = template.tlsCustomizers;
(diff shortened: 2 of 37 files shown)


trustin:
@andrey-tpt @anuraaga You might also wanna check this out.

trustin:
Another method I want to add is `ClientFactoryBuilder.tlsNoVerify()` which disables certificate chain verification, which is useful for testing.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -198,100 +165,176 @@ public VirtualHostBuilder hostnamePattern(String hostnamePattern) {
...
+        return tls(key, null, keyCertChain);
+    }
+
+    /**
+     * Configures SSL or TLS of this {@link VirtualHost} with the specified cleartext {@link PrivateKey} and
+     * {@link X509Certificate} chain.
+     *
+     * @see #tlsCustomizer(Consumer)
+     */
+    public VirtualHostBuilder tls(PrivateKey key, Iterable<X509Certificate> keyCertChain) {
```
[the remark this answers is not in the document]
trustin: Thanks! Fixed them all as well as other methods in `{Server,VirtualHost}Builder`.

trustin:
@jyblue You may be interested in this PR as well.

ikhoon:
Still LGTM.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -960,29 +1015,123 @@ VirtualHost build(VirtualHostBuilder template) {
...
-                tls(ssc.certificate(), ssc.privateKey());
+        SslContext sslContext = null;
+        boolean releaseSslContextOnFailure = false;
+        try {
+            // Build a new SslContext or use a user-specified one for backward compatibility.
+            if (this.sslContext != null) {
                 sslContext = this.sslContext;
-            } catch (Exception e) {
-                throw new RuntimeException("failed to create a self signed certificate", e);
+            } else if (sslContextBuilder != null) {
```
minwoox: What happens only if `tlsCustomizer(Consumer<? super SslContextBuilder> tlsCustomizer)` is called?
Should we raise an exception if `tlsCustomizer()` is called before `sslContextBuilder` or `tlsSelfSigned` is set ?

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -960,29 +1015,123 @@ VirtualHost build(VirtualHostBuilder template) {
...
+            return BouncyCastleKeyFactoryProvider.call(() -> SslContextUtil.createSslContext(
+                    sslContextBuilder, false, tlsCustomizers));
+        } catch (Exception e) {
+            // `createSslContext()` always throws an unchecked exception.
+            return Exceptions.throwUnsafely(e);
+        }
+    }
+
+    /**
+     * Validates the specified {@link SslContext} is configured properly. If configured as client context or
```
minwoox: Validates -> Validates whether

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOption.java
```diff
@@ -75,8 +75,18 @@
      * The {@link Consumer} which can arbitrarily configure the {@link SslContextBuilder} that will be
      * applied to the SSL session.
      */
+    public static final ClientFactoryOption<Consumer<? super SslContextBuilder>> TLS_CUSTOMIZER =
```
minwoox: Just wanted to make sure that we don't have to use same list as we do in the server-side
`List<Consumer<? super SslContextBuilder>>`?

trustin: I think it's fine as it is because it's more generic, and we do not expose TLS customizers in the server-side public API anyway.

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
trustin: Changed to `Makes sure`

core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
trustin: Because the following should work:

```java
sb.tlsCustomizer(...)
  .tlsSelfSigned()
  ...
```

we will have to do something at `build()`. Let me raise an exception if customizers were specified without `tls()` or `tlsSelfSigned()`.

trustin: Fixed

trustin:
Thanks for reviewing.

Issue #2306, opened 2019-12-10
Performance degradation when using JDK 11 and log4j2


ylgrgyq:
Hi folks,

Thanks for this great lib. I found an issue and would like to share it with you.

With JDK 11 and multi-release open, the `log4j2` lib will use the implementations with the Java9+ specific classes. This can cause a great performance degradation (roughly 30% in our service) because `Armeria` tries to [create a new `Logger`](https://github.com/line/armeria/blob/armeria-0.97.0/core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java#L194) on every request which will cause a slow deep stack walk initiated by `log4j2`. 

The typical stack trace is like this:
```
"armeria-common-worker-nio-2-3" #35 daemon prio=5 os_prio=31 cpu=5581.88ms elapsed=20.00s tid=0x00007fee5d0ea000 nid=0x7d03 runnable  [0x0000700010ab0000]
   java.lang.Thread.State: RUNNABLE
        at java.lang.StackStreamFactory$AbstractStackWalker.fetchStackFrames(java.base@11.0.2/Native Method)
        at java.lang.StackStreamFactory$AbstractStackWalker.fetchStackFrames(java.base@11.0.2/StackStreamFactory.java:386)
        at java.lang.StackStreamFactory$AbstractStackWalker.getNextBatch(java.base@11.0.2/StackStreamFactory.java:322)
        at java.lang.StackStreamFactory$AbstractStackWalker.peekFrame(java.base@11.0.2/StackStreamFactory.java:263)
        at java.lang.StackStreamFactory$AbstractStackWalker.hasNext(java.base@11.0.2/StackStreamFactory.java:351)
        at java.lang.StackStreamFactory$StackFrameTraverser.tryAdvance(java.base@11.0.2/StackStreamFactory.java:593)
        at java.util.stream.ReferencePipeline.forEachWithCancel(java.base@11.0.2/ReferencePipeline.java:127)
        at java.util.stream.AbstractPipeline.copyIntoWithCancel(java.base@11.0.2/AbstractPipeline.java:502)
        at java.util.stream.AbstractPipeline.copyInto(java.base@11.0.2/AbstractPipeline.java:488)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(java.base@11.0.2/AbstractPipeline.java:474)
        at java.util.stream.FindOps$FindOp.evaluateSequential(java.base@11.0.2/FindOps.java:150)
[...]

PR #2341, opened 2019-12-24
Remove per-service loggers

trustin:
Motivation:

- `ServiceConfig.loggerName()` and `ServiceRequestContext.logger()` have
  been deprecated since long ago.
- `LoggerFactory.getLogger()` is currently invoked on each request to
  create a new `Logger`, which may have performance impact on a certain
  logging framework, such as Log4J2, as reported in #2306.

Modifications:

- Remove:
  - `Server{Builder,Config}.serviceLoggerPrefix()`
  - `ServiceConfig.loggerName()`
  - `ServiceRequestContext.logger()`
- The default `Logger` of `LoggingService` is now:
  `LoggerFactory.getLogger(LoggingService.class)`
- Add `RequestContext.makeContextAware(Logger)`
  - Move `RequestContextAwareLogger` from `server` to `common`
- Make the logging implementations in `LoggingDecorators` log the
  current `RequestContext`.
- Remove unnecessary `RequestLogListener` instantiations in `Logging{Client,Service}`.

Result:

- A user can prepend the current `RequestContext` to log messages by
  using `RequestContext.makeContextAware(Logger)`.
- (Breaking) The following properties and/or methods have been removed:
  - `Server{Builder,Config}.serviceLoggerPrefix()`
  - `ServiceConfig.loggerName()`
  - `ServiceRequestContext.logger()`
  - Use your own logger or consider using `RequestContext.makeContextAware(Logger)`.
- `LoggingClient` now always prepend the current `RequestContext` to the
  log message.
- Slightly reduced memory footprint in `Logging{Client,Service}`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/RequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/common/RequestContext.java
@@ -16,6 +16,8 @@
 
 package com.linecorp.armeria.common;
 
+import static java.util.Objects.requireNonNull;
+
 import java.net.SocketAddress;
 import java.nio.ByteBuffer;
 import java.util.Iterator;
@@ -34,6 +36,7 @@ import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLSession;
 
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.linecorp.armeria.client.ClientRequestContext;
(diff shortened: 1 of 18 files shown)


trustin:
@ylgrgyq You might want to review this, too.

ylgrgyq:
Thanks! 👍 🎉

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -461,19 +464,28 @@ default ScheduledExecutorService makeContextAware(ScheduledExecutorService execu
...
     default <T> CompletableFuture<T> makeContextAware(CompletableFuture<T> future) {
         return makeContextAware((CompletionStage<T>) future).toCompletableFuture();
     }
 
+    /**
+     * Returns a {@link Logger} which prepends this {@link RequestContext} to the log message.
+     *
+     * @param logger the {@link Logger} to decorate.
+     */
+    default Logger makeContextAware(Logger logger) {
```
minwoox: Question: I'm wondering if we still need this methods because we add the `ctx` in `LoggingDecorator`. 🤔

trustin: This method is for a user who wants to make any logger context-aware. It's used for logging application-level logs.

PR #2342, opened 2019-12-24
Revamp RequestLog API

anuraaga:
This PR is still very much WIP - will finish it up by applying pattern for all methods and removing usage of deprecated, adding tests, etc after getting feedback on the API.

Currently, `RequestLog` has some issues for maintenance and usability

- `RequestLogAvailability` ties different properties together in an ad-hoc way. For example, `REQUEST_START` has start time and session information. If we find such a relationship is not what we want (e.g., for #2272 we want to separate start time from session information), it is tedious or sometimes impossible to update availabilities in a backwards compatible way.

- Understanding the properties available for any `RequestLogAvailability` requires a detailed look at javadoc, and can even require jumping through multiple different availabilities

This PR proposes some changes to help with this. We don't have to apply all of them, mostly just for brainstorming.

- Replace `RequestLogAvailability` with fine-grained `RequestLogProperty`. Advanced usage can continue to listen for availability of fine-grained properties
  - There is still some sharing of `RequestLogProperty` for fields that are obviously equivalent semantically. Different units for start time, or end time + duration, are always equivalent so share the same enum value. On the flip side, an exception and response content are considered different since they are totally different semantically.

- Define two large callbacks - `whenRequestComplete` and `whenComplete`. These two seem to be the most standard use cases
  - In #2272 @trustin suggested request identified and response end. From what I understand, request identified is just information from `context()` and doesn't need a callback, and response end seems to overlap with complete.

- Add `RequestRequestLog` which has the accessors for information available when request completes, so users have type-safe access to those properties when using `whenRequestComplete`

- Deprecates many shortcut methods. I feel as if having to deal with availability for these shortcut methods makes them less useful. I guess this is most controversial, no problem adding them back if this doesn't make sense.

One change I haven't made yet is that I would like all fields to be available when request is complete, even for those that didn't get populated (transferred byte times). They could probably return `-1` as a default. It seems tricky for just some fields to have a chance of never being available.


anuraaga:
Forgot to make it draft 😭

trustin:
> In #2272 @trustin suggested request identified and response end. From what I understand, request identified is just information from `context()` and doesn't need a callback, and response end seems to overlap with complete.

Request name may or may not be determined as soon as context is available. For RESTful requests, the name will be known immediately as soon as `RequestHeaders` is received. However, for RPC calls, some or whole request content must be retrieved.

anuraaga:
Right I thought of that but then figured request identification is user-specific so they need to set the listened properties themselves. Maybe the `whenAvailable` methods cover this case so we don't need a special group callback for it

trustin:
> Right I thought of that but then figured request identification is user-specific so they need to set the listened properties themselves. Maybe the `whenAvailable` methods cover this case so we don't need a special group callback for it 

Ah, that makes sense. Let me take a look today or tomorrow. :-)

core/src/main/java/com/linecorp/armeria/common/logging/RequestRequestLog.java
```diff
@@ -0,0 +1,267 @@
...
+import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.ServerBuilder;
+import com.linecorp.armeria.server.VirtualHostBuilder;
+
+import io.netty.channel.Channel;
+
+/**
+ * {@link RequestLog} properties available when request processing has completed.
+ */
+public interface RequestRequestLog {
```
trustin: I'm not sure I like this name :laughing:

anuraaga: You found my biggest trouble point ;)

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
```diff
@@ -83,17 +78,39 @@
...
+    @Deprecated
     Set<RequestLogAvailability> availabilities();
 
+    /**
+     * Returns an {@code int} representation of the currently available properties of this {@link RequestLog}.
+     * This can be useful when needing to quickly compare the availability of the {@link RequestLog} during
+     * the processing of the request. To actually check availability, use
+     * {@link #isAvailable(RequestLogProperty)}.
+     */
+    int availabilityFlags();
```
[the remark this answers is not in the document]
anuraaga: Yeah but with the larger number of properties, I was wary of precomputing all of the different values of the flags like we do right now in availability set.

Technically the fact that it matches the value of flags is the implementation detail but is intended to be used as an opaque value representing the current availability and tried to make that clear in the doc. But not sure this is the best way to solve the problem.

trustin:
> Add `RequestRequestLog` which has the accessors for information available when request completes, so users have type-safe access to those properties when using `whenRequestComplete`

I'm still in favor of merging `RequestRequestLog` back to `RequestLog` because 1) the name is weird and 2) a user may want to access the methods not available `RequestRequestLog`, e.g. accessing context or adding more listeners depending on what's available so far.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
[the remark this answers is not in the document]
trustin: I see. How about `availabilityStamp` then?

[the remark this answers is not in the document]
trustin: .. or `availabilityFingerprint`? Not sure which is better.

trustin:
@anuraaga Gentle ping :wink:

anuraaga:
Thanks for the comments! I'm currently in mostly web-only mode for solar new year for a couple of days but will address ASAP.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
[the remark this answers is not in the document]
anuraaga: `availabilityStamp` is what I was considering :)

trustin:
I like your idea of accessing the immediately unavailable properties via `context()`, so my only concern is naming.

I wonder it would be breaking too much if we rename `RequestLog` to `Log` and `RequestRequestLog` to `RequestLog` (and add `ResponseLog`). Thoughts?

anuraaga:
Yeah I was thinking of a rename, what came to mind was `RequestLog` -> `PartialLog`, though `Log` is probably fine too. I wouldn't use `ResponseLog` though since a log with complete response information also has complete request information. Perhaps `PartialLog`, `RequestLog`, `CompleteLog` (or `Log`)? Let's keep on brainstorming :)

Though indeed, it felt to me that renaming these wouldn't hurt so much, but when I tried an IntelliJ rename of `RequestLog` IntelliJ was hanging for a while and eventually applied to dozens of files, so even in the Armeria codebase it is indeed a big change. But maybe not so bad in user codebases, in my codebases I just see one reference to `RequestLog` total.

trustin:
I was thinking of having `RequestLog` and `ResponseLog` where each provides only request- or response- side properties only, and then we could have `Log` which extends the both interfaces. `ctx.log()` could return `Log` which raises an exception but the `RequestLog` or `Log` given to a listener would never raise an exception.

trustin:
Any progress on this?

anuraaga:
Sorry there's a lot of effect from the renaming and I haven't been able to get a big chunk of time for it yet. Think I'd get to it on Thursday, but if you have free time and want to take over to make sure that'd be fine too, I think we understand the general direction.

trustin:
OK. Let me leave a comment once I start to work on this, hopefully early next week. :bowing_man:

trustin:
Findings so far:

- I like the overall simplification of flags.
- I like the new shortcut methods - `whenComplete()` and `whenRequestComplete()`.
- Because `CompleteRequestLog` does not extend `RequestLog`, it is sometimes hard to share some common logic that works against both partially complete logs and complete logs, e.g. the default `MeterIdPrefixFunction` implementation.
  - I'm not sure if it's a good idea to force a user to fork the logic.
  - Perhaps we could replace `CompleteRequestLog` with `RequestLog` and just document that all properties will be available?
  - Same problem for `RequestCompleteRequestLog`.

Some idea: We could introduce a minimal type that only allows add listeners and expose it in `RequestContext`:

```java
ctx.log().requestHeaders(); // Compilation error
// ctx.log() is not a RequestLog but its subset.
RequestLog log;
log = ctx.log().ensureComplete(); // Throw exception if not complete
log = ctx.log().ensureRequestComplete(); // Throw exception if not complete
log = ctx.log().ensureAvailable(REQUEST_HEADERS, RESPONSE_HEADERS); // ...

// Or add a listener as usual.
ctx.log().whenComplete(...);
ctx.log().whenRequestComplete(...);
ctx.log().whenAvailable(...);
```

This way, we can give our users an impression that only the requested properties will be available, and then maybe continue using `RequestLog`?

anuraaga:
Ah yeah forcing forked logic doesn't seem so good. How about keeping `RequestCompleteLog` and `RequestLog extends RequestCompleteLog`? I think it allows the user to still have compiler checked availability for the simpler `whenRequestComplete` and `whenComplete` while we can use `RequestLog` in the other methods where it won't be full.

The thought of minimal `RequestLogListeners` type has come to mind. I think it works well but we will also need `ctx.log().partial()` for cases like logback integration that don't use listeners at all.

trustin:
> How about keeping `RequestCompleteLog` and `RequestLog extends RequestCompleteLog`?

SGTM.

trustin:
> The thought of minimal `RequestLogListeners` type has come to mind. I think it works well but we will also need `ctx.log().partial()` for cases like logback integration that don't use listeners at all.

Yeah, I guess varargs will do the job:

```java
RequestLog log;
log = ctx.log().ensureSome();
log = ctx.log().ensureSome(REQUEST_HEADERS);
```

trustin:
We have three interfaces now:

- `RequestLogAccess` - the minimal API for checking the availability and adding listeners.
- `RequestOnlyLog` (extends `RequestLogAccess`) - former `RequestCompleteRequestLog` 
- `RequestLog` (extends `RequestOnlyLog`) - replaces `CompleteRequestLog`

I'm not sure I like the type names. Any suggestions?

trustin:
Another idea, instead of `when{Available,Complete}(RequestLogListener)`, maybe we can just return a future? e.g.

```java
ctx.logFutures().complete().accept(log -> { .... });
log = ctx.logFutures().complete().join();
requestOnlyLog = ctx.logFutures().requestComplete().join();
ctx.logFutures().available(PROP_A, PROP_B).accept(log -> { .... });
```

Not sure about the incurred overhead though.

ikhoon:
Just a random idea.
How about making the log API similar to `HttpResponse`?
```
ctx.logFutures().aggregate().accept(log -> { .... });  //
ctx.log().subscribe(new Subscriber { 
  void onNext(LogElement log) 
}
log = ctx.log().aggregate().join();
// ... same as you mentioned.
```

trustin:
Had a chance to think more about using `CompletableFuture` and seems like the overhead will not be that much.

anuraaga:
Future sounds nice - we might prefer a callback if it's expected to be already completed in a majority of cases, but here I think it's almost always deferred.

trustin:
So far so good. It seems to me that `CompletableFuture` made the overall API much more understandable. Let me change the return type of `RequestContext.log()` from `RequestLog` to `RequestLogAccess` tomorrow. Expected changes:

```java
// Before:
RequestLog log = ctx.log();
await().untilAsserted(() -> assertThat(log.isComplete()).isTrue());
assertThat(log.requestCause()).isSameAs(resCause);
assertThat(log.responseCause()).isSameAs(resCause);

// After:
RequestLog log = ctx.log().future().join(); // or get(10, SECONDS)
assertThat(log.requestCause()).isSameAs(resCause);
assertThat(log.responseCause()).isSameAs(resCause);
```

trustin:
See `RequestLogAccess` for all the future methods.

minwoox:
`RequestLogAccess` looks good to me.

trustin:
`RequestContext.log()` now returns `RequestLogAccess`. The change looks fine to me. Let me know if you have ideas for more convenience methods.

trustin:
The only thing I don't like about the new API is that a user have to call `partial()` for the cases like the following:

```java
        final ResponseHeaders value;
        if (log.isAvailable(RequestLogProperty.RESPONSE_HEADERS)) {
            value = log.partial().responseHeaders();
        } else {
            value = null;
        }
```

anuraaga:
How about adding `ensureAvailableOrElse` next to `ensureAvailable`? I guess it's the JDK idiom for default values

trustin:
> How about adding `ensureAvailableOrElse` next to `ensureAvailable`? I guess it's the JDK idiom for default values

The problem is that `RequestLogProperty` does not always refer to a single property but sometimes to a group of properties, e.g. `REQUEST_START_TIME`.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Returns a {@link CompletableFuture} which will be completed when all the properties signified by
+     * the specified {@link RequestLogProperty}s are collected. Note that the completion of the returned
+     * {@link CompletableFuture} guarantees only the availability of the specified properties, which means
+     * any attempt to access other properties than specified may trigger
+     * a {@link RequestLogAvailabilityException}. If in doubt, use {@link #future()} or
+     * {@link #requestFuture()}.
+     */
+    CompletableFuture<RequestLog> partialFuture(RequestLogProperty... properties);
```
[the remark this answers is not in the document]
trustin: It throws an `IllegalArgumentException` at runtime, although it's not checked at compile time. What do you think?

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
```diff
@@ -16,276 +16,68 @@
...
-
 /**
  * A set of informational properties collected while processing a {@link Request} and its {@link Response}.
  * The properties provided by this class are not always fully available. Check the availability of each
- * property using {@link #isAvailable(RequestLogAvailability)} or {@link #availabilities()}. Attempting to
- * access the properties that are not available yet will cause a {@link RequestLogAvailabilityException}.
- * Use {@link #addListener(RequestLogListener, RequestLogAvailability)} to get notified when the interested
- * properties are available.
+ * property using {@link #isAvailable(RequestLogProperty)}. Attempting to access the properties that are
+ * not available yet will cause a {@link RequestLogAvailabilityException}. Use the following methods to
```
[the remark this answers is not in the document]
trustin: Yep. Need to rework the docs.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
```diff
@@ -0,0 +1,198 @@
...
+     * to access a unavailable property will trigger a {@link RequestLogAvailabilityException}.
+     * If in doubt, use {@link #future()} or {@link #requestFuture()}.
+     */
+    RequestLog partial();
+
+    /**
+     * Returns a {@link CompletableFuture} which will be completed when the {@link Request} has been processed
+     * completely and thus all properties of the {@link RequestLog} have been collected.
+     */
+    CompletableFuture<RequestLog> future();
```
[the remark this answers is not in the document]
trustin: Renamed

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
```diff
@@ -0,0 +1,198 @@
...
+        } while (i.hasNext());
+        return true;
+    }
+
+    /**
+     * Returns the {@link RequestLog} whose all properties may not be available. Note that an attempt
+     * to access a unavailable property will trigger a {@link RequestLogAvailabilityException}.
+     * If in doubt, use {@link #future()} or {@link #requestFuture()}.
+     */
+    RequestLog partial();
```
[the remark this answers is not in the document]
trustin: Moved down right below `ensurePartial()`

trustin:
`{request,response}FirstBytesTransferredTimeNanos` can still throw `RequestLogAvailabilityException` even after `RequestLog` is comeplete, because a request can fail without sending anything. I think there are a few options here:

- Document about this exception. This may not be user-friendly.
- Make the getters return `Long` or `null`. `null` means nothing was written. This incurs two extra tiny object allocations.
- Consider a special value such as `0` to mean nothing was written. If the timestamp is really 0, make it 1 or -1. This saves two tiny object allocations.

I'm leaning toward the option 2 or 3. Thoughts?

ikhoon:
> I'm leaning toward the option 2 or 3. Thoughts?

The option 3 sounds good to me. 👍
`{request,response}FirstBytesTransferredTimeNano` represents a specific time. I think users easily understand that `0` or `-1` can be treated as a special value, especially when representing time.

trustin:
Hmm.. `System.nanoTime()` can be any number between `Long.MIN_VALUE` and `Long.MAX_VALUE`, so I guess a user will not find any meaning from its value, as long as we do not define a constant. Although efficient, maybe it could be confusing to some users? "Why are these guys using 0 as 'nothing'? In nanoTime, 0 is a valid value!"

trustin:
The fourth option (suggested by @ikhoon) was to represent the `{request,response}FirstBytesTransferredTime` as a relative value against the `{request,response}StartTime`, e.g. `{request,response}FirstBytesTransferredAfterNanos` (any good names?).

trustin:
The problem with the fourth option is that no other properties in our API represent nano-precision timings that way, which may make our users even more confused. :scream:

ikhoon:
How about introducing a new `LogTiming` class to hide timing complex from `RequestLog`?
I am less sure that most users want to get `headers`, `cause` or `previewContent`

anuraaga:
Yeah my plan was to use `-1` since only recently I learned about negative nanotime in https://mail.openjdk.java.net/pipermail/net-dev/2020-January/013479.html :(

Though the review I got there from an openjdk dev seems to indicate using `-1` as an explicit default would be kosher (I found an alternative for that specific case though) - I guess since the probability of it being wrong is something like 1 in billions. So I'm ok with option 3. Option 2 seems ok too though it's unfortunate the values are so large and always change so they're never memoized.

trustin:
> So I'm ok with option 3. Option 2 seems ok too though it's unfortunate the values are so large and always change so they're never memoized.

How about keeping a timestamp as a pair of `(boolean, long)` and then creating a new `Long` only when the property is actually accessed? Escape analysis may help us (or not). :laughing:

trustin:
Other than the two properties, I think this pull request is ready for reviews.

anuraaga:
That sounds good too. Though since we have the flags I guess it's overkill to have a boolean too just to avoid the volatile read, I'd be fine with reading flags even for a complete request for those two only.

trustin:
> Though since we have the flags I guess it's overkill to have a boolean too just to avoid the volatile read

Realized that there's still no way to tell if the value is set or not because the value's flag will be set when `isComplete()` becomes true.

trustin:
Chose the 2nd option. Jolly happy review time! :see_no_evil:

trustin:
Let me rebase and squash so that Codecov finds the base commit well.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
```diff
@@ -0,0 +1,290 @@
...
+    boolean isComplete();
+
+    /**
+     * Returns {@code true} if the {@link Request} has been consumed completely and thus all properties of
+     * the {@link RequestOnlyLog} have been collected.
+     */
+    boolean isRequestComplete();
+
+    /**
+     * Returns {@code true} if the property signified by the specified {@link RequestLogProperty} is available.
```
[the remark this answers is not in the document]
trustin: I'm not sure which is better. Thoughts? @ikhoon @anuraaga

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
```diff
@@ -0,0 +1,290 @@
...
+     * completely and thus all properties of the {@link RequestLog} have been collected.
+     * <pre>{@code
+     * logAccess.completeFuture().thenAccept(log -> {
+     *     HttpStatus status = log.responseHeaders().status();
+     *     if (status == HttpStatus.OK) {
+     *         ...
+     *     }
+     * });
+     * }</pre>
+     */
```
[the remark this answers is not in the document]
trustin: Good idea. Added to all future-returning methods.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
```diff
@@ -0,0 +1,290 @@
...
+    RequestContext context();
+
+    /**
+     * Returns the {@link RequestId}. This method is a shortcut for {@code context().id()}.
+     *
+     * @deprecated Use {@code log.context().id()}.
+     */
+    @Deprecated
+    default RequestId id() {
```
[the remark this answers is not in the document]
trustin: Removed all deprecated fields and methods.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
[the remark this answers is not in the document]
ikhoon: `the properties signified by the specified {@link RequestLogProperty}s` have been used 4 times in this class. Unless we remove all of them, I'm good to use it. 😀

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
```diff
@@ -0,0 +1,281 @@
...
+     *              if (headers.path().startsWith("/foo/") &&
+     *                  resHeaders.status() == HttpStatus.OK) {
+     *                  ...
+     *              }
+     *          });
+     * }</pre>
+     *
+     * @throws IllegalArgumentException if {@code properties} is empty.
+     */
+    CompletableFuture<RequestLog> partialFuture(Iterable<RequestLogProperty> properties);
```
ikhoon: I'm not strong here, just summarizing what I was thinking while reviewing.

If users want to listen to all events and information of `RequestLogProperty.REQUEST_HEADERS`, `RequestLogProperty.REQUEST_HEADERS` and other properties immediately. And they want to make their own metric or logging.

I guess it is not impossible. It can be achieved by adding callback with event to `partialFuture(RequestLogProperty)` multiple times.
```java
logAcess.partialFuture(RequestLogProperty.REQUEST_START_TIME).thenAccept(log -> ...)
logAcess.partialFuture(RequestLogProperty.REQUEST_END_TIME).thenAccept(log -> ...)
logAcess.partialFuture(RequestLogProperty.RESPONSE_START_TIME).thenAccept(log -> ...)
logAcess.partialFuture(RequestLogProperty.RESPONSE_END_TIME).thenAccept(log -> ...)
```

Suggestion: How about adding `EventListener`, `ClientEventListener` which might need for DNS resolution timing information. Inspired by [okhttp - Events](https://square.github.io/okhttp/events/), [okhttp - EventListener](https://github.com/square/okhttp/blob/8511ee127f97f8cef2c382740b9cb6675d6ccae7/okhttp/src/main/java/okhttp3/EventListener.kt#L61)

```java
logAccess.addEventListener(new EventLister () {
	void onRequestHeaderSent(RequestHeaders) {
        ...
    }
    ...
})
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -90,22 +74,24 @@
     private static final ResponseHeaders DUMMY_RESPONSE_HEADERS = ResponseHeaders.of(HttpStatus.UNKNOWN);
 
     private final RequestContext ctx;
+    private final CompleteRequestLog notCheckingAccessor = new CompleteRequestLog();
```
anuraaga: By the way this was a brainstorming proposal - it seems good on paper, but we might want to benchmark before adding this class.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+    }
+
+    /**
+     * A {@link RequestLog} that delegates to {@link DefaultRequestLog} to read properties without
+     * checking availability.
+     */
+    private final class CompleteRequestLog implements RequestLog {
+        @Override
+        public boolean isComplete() {
+            return DefaultRequestLog.this.isComplete();
```
anuraaga:
```suggestion
            return true;
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public boolean isRequestComplete() {
+            return DefaultRequestLog.this.isRequestComplete();
```
anuraaga:
```suggestion
            return true;
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -177,192 +166,313 @@ public DefaultRequestLog(RequestContext ctx, ContentPreviewerFactory requestCont
...
+                return cast;
+            }
+
+            newFuture = new RequestLogFuture(interestedFlags);
+            futures.put(interestedFlags, newFuture);
         }
 
-        if (lastChild.isAvailable(RESPONSE_END)) {
-            propagateResponseEndData(lastChild);
+        if (isAvailable(flags, interestedFlags)) {
```
anuraaga: I guess it might be a tiny bit faster to special case so only one volatile read

```java
final int flags = this.flags;
if (flags == availableFlags && flags == FLAGS_REQUEST_COMPLETE) {
  newFuture.completeLog(notCheckingAccessor);
} else if (isAvailable(flags, interestedFlags)) {
  newFuture.completeLog(this);
}
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -177,192 +166,313 @@ public DefaultRequestLog(RequestContext ctx, ContentPreviewerFactory requestCont
...
     }
 
-    private boolean isAvailable(int interestedFlags) {
-        return (flags & interestedFlags) == interestedFlags;
+    @Override
+    public RequestLog ensurePartial(RequestLogProperty property) {
+        if (!isAvailable(property)) {
+            throw new RequestLogAvailabilityException(property.name());
+        }
+        return partial();
```
anuraaga: `return this` is probably fine for these partial methods, they'll usually access just a property or two anyways.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -177,192 +166,313 @@ public DefaultRequestLog(RequestContext ctx, ContentPreviewerFactory requestCont
...
-            satisfiedListeners = removeSatisfiedListeners();
-        }
-        if (satisfiedListeners != null) {
-            invokeOnRequestLog(satisfiedListeners, this);
+    private void updateAvailability(int flags) {
+        for (;;) {
+            final int oldAvailability = this.flags;
+            final int newAvailability = oldAvailability | flags;
+            if (flagsUpdater.compareAndSet(this, oldAvailability, newAvailability)) {
+                if (oldAvailability != newAvailability) {
```
anuraaga: Can this be outside the `compareAndSet`?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -177,192 +166,313 @@ public DefaultRequestLog(RequestContext ctx, ContentPreviewerFactory requestCont
...
-        return ctx;
+    public void endResponseWithLastChild() {
+        checkState(!hasLastChild, "last child is already added");
+        checkState(children != null && !children.isEmpty(), "at least one child should be already added");
+        hasLastChild = true;
+        final RequestLogAccess lastChild = children.get(children.size() - 1);
+        propagateResponseSideLog(lastChild.partial());
+    }
+
+    private void propagateResponseSideLog(RequestLog lastChild) {
```
anuraaga: By the way when I was reading this code I was confused by response checking availability but not request. I'm assuming the request will never be available, but adding a comment why would help

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public boolean isRequestComplete() {
+            return DefaultRequestLog.this.isRequestComplete();
+        }
+
+        @Override
+        public boolean isAvailable(RequestLogProperty property) {
+            return DefaultRequestLog.this.isAvailable(property);
```
anuraaga:
```suggestion
            return true;
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public RequestLog partial() {
+            return DefaultRequestLog.this.partial();
```
anuraaga:
```suggestion
            return this;
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public CompletableFuture<RequestLog> completeFuture() {
+            return DefaultRequestLog.this.completeFuture();
```
anuraaga: I wonder if there's any advantage to returning a new completed future GC-wise / escape-analysis-wise for all of these. Though I guess they're pretty rarely used anyways

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public CompletableFuture<RequestLog> partialFuture(Iterable<RequestLogProperty> properties) {
+            return DefaultRequestLog.this.partialFuture(properties);
+        }
+
+        @Override
+        public RequestLog ensureComplete() {
+            return DefaultRequestLog.this.ensureComplete();
```
anuraaga:
```suggestion
            return this;
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public RequestOnlyLog ensureRequestComplete() {
+            return DefaultRequestLog.this.ensureRequestComplete();
```
anuraaga:
```suggestion
            return this;
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public RequestLog ensurePartial(RequestLogProperty property) {
+            return DefaultRequestLog.this.ensurePartial(property);
```
anuraaga:
```suggestion
            return this;
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public RequestLog ensurePartial(RequestLogProperty... properties) {
+            return DefaultRequestLog.this.ensurePartial(properties);
```
anuraaga:
```suggestion
            return this;
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1349,314 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public RequestLog ensurePartial(Iterable<RequestLogProperty> properties) {
+            return DefaultRequestLog.this.ensurePartial(properties);
```
anuraaga:
```suggestion
            return this;
```

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
```diff
@@ -140,6 +140,9 @@ public static RequestContextExporterBuilder builder() {
      */
     public boolean containsAttribute(AttributeKey<?> key) {
         requireNonNull(key, "key");
+        if (attrs == null) {
+            return false;
+        }
         return Arrays.stream(attrs).anyMatch(e -> e.key.equals(key));
```
anuraaga: It's unrelated to this PR, but I noticed these methods seem fairly slow. I think the exporter is generally a singleton so object allocation shouldn't matter so much - should we use maps instead of arrays here?

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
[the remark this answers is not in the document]
anuraaga: I like `if the specified {@link RequestLogProperty} is available.` a little better for all

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
anuraaga: The API could be useful, but maybe we can add it later if there's a request for it. I think the use-case for grouped vs many partial listeners overwhelmingly leans towards grouped, in this PR there is basically only one location (on-child-request) that is setting many partial futures and in the wild it seems similarly rare.

core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefixFunction.java
```diff
@@ -71,25 +70,25 @@ static MeterIdPrefixFunction ofDefault(String name) {
...
             }
 
             @Override
-            public MeterIdPrefix apply(MeterRegistry registry, RequestLog log) {
+            public MeterIdPrefix completeRequestPrefix(MeterRegistry registry, RequestLog log) {
                 // hostNamePattern, httpStatus, method, route
                 final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(4);
-                // For optimal performance, add tags in order.
-                buildTags(tagListBuilder, log, true);
+                // For optimal performance, add tags in order, except 'http.status'.
```
anuraaga: Hmm - if we stop adding in order, maybe we don't need this comment ;)

site/src/sphinx/advanced-structured-logging.rst
```diff
@@ -451,13 +461,13 @@ You can retrieve the child logs using ``RequestLog.children()``.
 .. code-block:: java
 
     final RequestContext ctx = ...;
-    ctx.log().addListner(log -> {
+    ctx.log().completeFuture(log -> {
```
anuraaga:
```suggestion
    ctx.log().completeFuture().thenAccept(log -> {
```

core/src/main/java/com/linecorp/armeria/client/DefaultWebClient.java
```diff
@@ -95,7 +95,7 @@ private HttpResponse execute(EndpointGroup endpointGroup, HttpRequest req) {
         return execute(endpointGroup, req.method(),
                        pathAndQuery.path(), pathAndQuery.query(), null, req,
                        (ctx, cause) -> {
-                           if (ctx != null && !ctx.log().isAvailable(RequestLogAvailability.REQUEST_START)) {
+                           if (ctx != null && !ctx.log().isAvailable(RequestLogProperty.REQUEST_START_TIME)) {
```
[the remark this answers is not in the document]
trustin: Hmm.. I'm not sure, but maybe replacing `is` with `has been` helps?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: Will use `UnmodifiableFuture.completedFuture(null)` which returns a singleton once #2306 is merged.

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
trustin: I thought this method is used only for checking how the exporter is configured. What's your use case?

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -95,55 +95,48 @@ public void header(String name, String value) {
...
     }
 
-    static brave.http.HttpClientResponse asHttpClientResponse(ClientRequestContext ctx) {
-        return new HttpClientResponse(ctx);
+    static brave.http.HttpClientResponse asHttpClientResponse(RequestLog log) {
+        return new HttpClientResponse(log);
     }
 
     @SuppressWarnings("ClassNameSameAsAncestorName")
     private static final class HttpClientResponse extends brave.http.HttpClientResponse {
```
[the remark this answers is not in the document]
trustin: Done

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -118,18 +118,25 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
             }
         }
 
-        ctx.log().addListener(log -> SpanContextUtil.startSpan(span, log),
-                              RequestLogAvailability.REQUEST_START);
+        ctx.log().partialFuture(RequestLogProperty.REQUEST_START_TIME)
```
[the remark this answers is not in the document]
trustin: Good point. Inlined `startSpan()` as well.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: Bimorphic calls are almost as fast as monomorphic calls, so I guess it will be a win, although we allocate one extra object?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: Added some comments to both `propagate{Request,Response}SideLog()`

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: Fixed slightly differently, but basically one less volatile read. :-)

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: Just realized this is not possible because we return `notCheckingAccess` when `isRequestComplete() == true` as well. I think we need to choose between:

1. provide two different complete log implementations (`CompleteRequestLog` and `CompleteRequestOnlyLog`). Lazy-allocate `CompleteRequestOnlyLog`.
2. do not pass `notCheckingAccess` to `requestCompleteFuture` and optimize for `completeFuture`.
3. pass `notCheckingAccess` to both `requestCompleteFuture` and `completeFuture`.

We are currently at (3), but I find it somewhat confusing to work with. How about (1) or (2)? (1) sounds good to me. What do you think?

trustin:
Reposting since the comment thread has been collapsed:

Just realized that it's not possible to simply return `true` for `CompleteRequestLog.isAvailable()` or `this` for `CompleteRequestLog.isPartial()` because we complete the futures with `notCheckingAccess` when `isRequestComplete() == true` as well as when `isComplete() == true`. I think we need to choose between:

1. provide two different complete log implementations (`CompleteRequestLog` and `CompleteRequestOnlyLog`). Lazy-allocate `CompleteRequestOnlyLog`.
2. do not pass `notCheckingAccess` to `requestCompleteFuture` and optimize for `completeFuture`.
3. pass `notCheckingAccess` to both `requestCompleteFuture` and `completeFuture`.

We are currently at (3), but I find it somewhat confusing to work with. How about (1) or (2)? (1) sounds good to me. What do you think?

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
anuraaga: Not sure ;) But since these are public methods, figured it's worth considering it since doesn't seem that complicated to speed them up.

trustin:
> 1. provide two different complete log implementations (`CompleteRequestLog` and `CompleteRequestOnlyLog`). Lazy-allocate `CompleteRequestOnlyLog`.

This wasn't easy as initially thought, because we need to distinguish between 1) `whenAvailable(...all request properties...)` and 2) `whenRequestComplete()`. I'm not sure if `whenRequestComplete()` is worth its own future field or special flag/complication.

trustin:
> 2. do not pass `notCheckingAccess` to `requestCompleteFuture` and optimize for `completeFuture`.

Chose this option.

trustin:
Also found that there's some notification ordering issue, so added `Arrays.sort()` call and its corresponding test: 2c1735f011e781838f34662f95864122c79879d4 I'm somewhat concerned about performance but can't think of a good way to improve it.

trustin:
Now using a list instead of map, with same level of efficiency with the old implementation.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
anuraaga: Yeah it's my guess too, but generally better to benchmark than guess ;) But it's fine as is for now either way.

trustin:
Motivation:

- A user almost always is interested only in the complete log.
- When a user is interested in a certain property only, it may be notified later than it is actually set.
- The notion of getter flags and setter flags is complicated without much gain.
- A user can get `RequestLogAvailabilityException` too easily.

Modifications:

- Added `RequestLogAccess` which provides safe access to the `RequestLog` properties.
  - (Breaking) `RequestContext.log()` now returns `RequestLogAccess` instead of `RequestLog`.
- Added `RequestOnlyLog` which extends `RequestLogAccess`
- `RequestLog` extends `RequestOnlyLog`.
  - (Breaking) All deprecated methods in `RequestLog` have been removed.
  - (Breaking) `RequestLog.addListener()` has been replaced with `RequestLogAccess.when*()`.
  - (Breaking) `RequestLog.removeListener()` has been removed.
- (Breaking) `RequestLog.{request,response}FirstBytesTransferredTimeNanos()` now returns `Long` instead of `long` to represent the case where no byte was transferred.
- `RequestLogBuilder` extends `RequestLogAccess`.
- (Breaking) `RequestLogAvailability` has been replaced with `RequestLogProperty`
  - (Breaking) `RequestLogAvailabilitySet` has been removed.
- (Breaking) `MeterIdPrefixFunction.activeRequestPrefix()` now requires a `RequestOnlyLog`.
  - (Breaking) `MeterIdPrefixFunction.apply()` has been renamed to `completeRequestPrefix()`.
  - (Breaking) `MeterIdPrefixFunction` is not a functional interface anymore.
  - Added `MeterIdPrefixFunction.of(BiFunction)` instead.
- (Breaking) `LoggingClient` and `LoggingService` now use `RequestOnlyLog` for requests. `RequestLog` is used for responses.
- Miscellaneous:
  - Overall clean-up of `armeria-brave`
  - Fixed a bug where `RequestLogBuilder.endRequest()` is called too early in `HttpRequestSubscriber`, causing `requestFirstBytesTransferredTimeNanos()` to be recorded incorrectly.
  - Added `ServiceRequestContextBuilder.route()`.
  - (Breaking) `AccessLogWriter.log()` does not have the `throws Exception` clause.

Result:

- The new `RequestLog` API which is more user-friendly.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -177,192 +171,348 @@ public DefaultRequestLog(RequestContext ctx, ContentPreviewerFactory requestCont
...
         }
 
-        if (isAvailable(interestedFlags)) {
-            // No need to add to 'listeners'.
-            invokeOnRequestLog(listener, this);
-            return;
+        final CompletableFuture<RequestLog> future;
+
+        final int flags = this.flags;
+        final RequestLog log = partial(flags);
```
minwoox: nit: we can move this line down to 318

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -424,77 +576,76 @@ public long requestStartTimeMillis() {
...
         return requestStartTimeNanos;
     }
 
     @Override
-    public long requestFirstBytesTransferredTimeNanos() {
-        ensureAvailability(REQUEST_FIRST_BYTES_TRANSFERRED);
-        return requestFirstBytesTransferredTimeNanos;
+    public Long requestFirstBytesTransferredTimeNanos() {
+        ensureAvailable(RequestLogProperty.REQUEST_FIRST_BYTES_TRANSFERRED_TIME);
+        return requestFirstBytesTransferredTimeNanosSet ? requestFirstBytesTransferredTimeNanos : null;
```
minwoox: I think we can just return `requestFirstBytesTransferredTimeNanos`.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1389,323 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public boolean isRequestComplete() {
+            return true;
+        }
+
+        @Override
+        public boolean isAvailable(RequestLogProperty property) {
+            return true;
```
minwoox: I think we should treat `REQUEST_FIRST_BYTES_TRANSFERRED_TIME` and `RESPONSE_FIRST_BYTES_TRANSFERRED_TIME` specially.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1389,323 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+            // OK to cast because the future is unmodifiable.
+            @SuppressWarnings("unchecked")
+            final CompletableFuture<RequestOnlyLog> cast =
+                    (CompletableFuture<RequestOnlyLog>) (CompletableFuture<?>) whenComplete();
+            return cast;
+        }
+
+        @Override
+        public CompletableFuture<RequestLog> whenAvailable(RequestLogProperty property) {
+            return whenComplete();
```
minwoox: ditto. What if a user does like:
```
log.whenAvailable(RequestLogProperty. REQUEST_FIRST_BYTES_TRANSFERRED_TIME).thenApply(log -> {this log might not have the property.})
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1281,13 +1389,323 @@ public String toStringResponseOnly(Function<? super ResponseHeaders, ?> headersS
...
+        }
+
+        @Override
+        public long requestStartTimeNanos() {
+            return requestStartTimeNanos;
+        }
+
+        @Override
+        public Long requestFirstBytesTransferredTimeNanos() {
+            return requestFirstBytesTransferredTimeNanosSet ? requestFirstBytesTransferredTimeNanos : null;
```
minwoox: If we do `DefaultRequestLog.this.isAvailable(RequestLogProperty.REQUEST_FIRST_BYTES_TRANSFERRED_TIME)`
we can remove `requestFirstBytesTransferredTimeNanosSet` completely.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLogAccess.java
```diff
@@ -0,0 +1,277 @@
...
+     * }</pre>
+     */
+    CompletableFuture<RequestOnlyLog> whenRequestComplete();
+
+    /**
+     * Returns a {@link CompletableFuture} which will be completed when the specified
+     * {@link RequestLogProperty} is collected. The returned {@link CompletableFuture} is never completed
+     * exceptionally. Note that the completion of the returned {@link CompletableFuture} guarantees only
+     * the availability of the specified property, which means any attempt to access other properties than
+     * a {@link RequestLogAvailabilityException}.
```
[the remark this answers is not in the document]
trustin: :scream:

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
```diff
@@ -65,9 +65,9 @@ public static void setup(RequestContext ctx, AttributeKey<Boolean> requestMetric
         }
         ctx.setAttr(requestMetricsSetKey, true);
 
-        ctx.log().addListener(log -> onRequest(log, meterIdPrefixFunction, server),
-                              RequestLogAvailability.REQUEST_HEADERS,
-                              RequestLogAvailability.REQUEST_CONTENT);
+        ctx.log()
+           .whenAvailable(RequestLogProperty.REQUEST_HEADERS, RequestLogProperty.REQUEST_CONTENT)
+           .thenAccept(log -> onRequest(log, meterIdPrefixFunction, server));
```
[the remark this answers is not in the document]
trustin: I thought so initially but it seem to break the existing behavior/tests, mainly because not all services consume the request completely. I think we need to pick some essential properties. What else would be useful?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: No. A request can be complete without sending anything to the wire. In that case, `requestFirstBytesTransferredTimeNanosSet` is `false` and `null` has to be returned.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: We had to previously, but this PR introduces new flags `{request,response}FirstBytesTransferredTimeNanosSet`. When a request is complete, `{REQUEST,RESPONSE}_FIRST_BYTES_TRANSFERRED_TIME` becomes available now. As a result, no properties are treated specially.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: Again, `{REQUEST,RESPONSE}_FIRST_BYTES_TRANSFERRED_TIME` is not treated specially anymore, so the callback will be notified immediately.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: Yeah, but the point of this change is to make all properties behave in a consistent way.

When a certain property is not available, it means we do not know its value yet and we may know it later.

When available, it means we do know the value and it will not change.

Previously, `{request,response}FirstBytesTransferredTime` had no distinction between 'not written so far but may be written in the future' and 'not written so far and will never be written' unlike other properties. 

A user was able to tell by checking whether the request is complete, but it is indirect and a user was not able to write a simple listener like this:

```java
log.addListener(log -> {
    // Do this if written. Do that if not written.
}, REQUEST_FIRST_BYTES_TRANSFERRED);
```

.. because the callback will not always be invoked.

The new API guaranteed the callback will be invoked, which is less surprising:

```java
log.whenAvailable(REQUEST_FIRST_BYTES_TRANSFERRED_TIME).thenAccept(log -> {
    // Do this if written, Do that if not written.
});
```

core/src/main/java/com/linecorp/armeria/internal/metric/RequestMetricSupport.java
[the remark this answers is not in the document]
trustin: Added some Javadoc to `MeterIfPrefixFunction` about what properties will be available.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
minwoox: Oops, I didn't notice that the logic changed. My bad. 😅

trustin:
Thanks everyone!

PR #2344, opened 2019-12-26
Add `Clients.newContextCaptor()`

### LINKED ISSUE #1739: Provide a way to get `ClientRequestContext` without using a decorator.

trustin:
Motivation:

It'd be nice if a user can capture the `ClientRequestContext` of the
request being sent without using a decorator.

Modifications:

- Add `Clients.newContextCaptor()`
- Add `ClientThreadLocalState` which replaces the functional
  composition of `Consumer<ClientRequestContext>`s.
- Copy the customizer list when a `DefaultClientRequestContext` is
  constructed, so that thread-local customizers are evaluated even if
  an actual request starts from a different thread.
- Add `? super` to some type parameters for user convenience.
- Miscellaneous:
  - Fix a bug in the gRPC client where a wrong path (usually /) is set to the context.
  - Make `ClientRequestContext.ownAttr()` nullable.

Result:

- Closes #1739
- Fixed a bug where thread-local context customizers were called for derived contexts unintentionally.
- Fixed a bug where `Clients.withHttpHeaders()` and `withContextCustomizer()`
  don't work for gRPC clients.
- Fixed a bug in the gRPC client where a wrong path (usually /) is set to the context.


grpc/src/main/java/com/linecorp/armeria/client/grpc/ArmeriaChannel.java
```diff
@@ -166,8 +166,8 @@ private DefaultClientRequestContext newContext(HttpMethod method, HttpRequest re
                 sessionProtocol,
                 options().requestIdGenerator().get(),
                 method,
-                uri().getRawPath(),
-                uri().getRawQuery(),
+                req.path(),
+                null,
```
trustin: This fixes a bug in the gRPC client where a wrong path (usually `/`) is set to the context. /cc @anuraaga

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -96,6 +96,9 @@
     @Nullable
     private String strVal;
 
+    @Nullable
+    private volatile List<Consumer<? super ClientRequestContext>> customizers;
```
minwoox: Can we just make this final and not nullable by returning an empty list?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -282,11 +289,24 @@ private DefaultClientRequestContext(DefaultClientRequestContext ctx,
...
         maxResponseLength = ctx.maxResponseLength();
         additionalRequestHeaders = ctx.additionalRequestHeaders();
+        customizers = copyThreadLocalCustomizers();
 
         for (final Iterator<Attribute<?>> i = ctx.attrs(); i.hasNext();) {
             addAttr(i.next());
         }
-        runThreadLocalContextCustomizer();
+
+        runThreadLocalContextCustomizers();
```
minwoox: How about adding a small example that uses `RetryingClient` so that we can be sure that it works as expected?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: We clean it up after retrieving the list in `runThreaedLocalContextCustomizers` (just in case there's any big garbage lurking in there), so we can't make it final.

I thought about using an empty list instead of null, but I chose to use null because it's performance-critical path. I can use an empty list if you mind, though.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: Will do!

trustin: What's the expected behavior by the way when `RetryingClient` is involved?

trustin: I realized that we should not invoke the customizers for derived contexts. Just removed the call. Less code = no need to test. :rofl:

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
minwoox: Thak makes sense. It's fine as it is. 😄

core/src/test/java/com/linecorp/armeria/client/DefaultClientRequestContextTest.java
```diff
@@ -91,6 +86,36 @@ void deriveContext() {
         assertThat(derivedCtx.attr(bar).get()).isEqualTo(null);
     }
 
+    @Test
+    void derivedContextMustNotCallCustomizers() {
```
minwoox: Thanks!

core/src/main/java/com/linecorp/armeria/client/UserClient.java
```diff
@@ -126,26 +125,24 @@ protected final Endpoint endpoint() {
...
-        return execute(null, endpoint, method, path, query, fragment, req, fallback);
+        return execute(endpoint, method, path, query, fragment, req, fallback);
     }
 
     /**
      * Executes the specified {@link Request} via {@link #delegate()}.
-     *
-     * @param eventLoop the {@link EventLoop} to execute the {@link Request}
-     * @param endpoint the {@link Endpoint} of the {@link Request}
+     *  @param endpoint the {@link Endpoint} of the {@link Request}
```
ikhoon: nit: double space here. 
```suggestion
     * @param endpoint the {@link Endpoint} of the {@link Request}
```

core/src/main/java/com/linecorp/armeria/client/DefaultWebClient.java
```diff
@@ -48,10 +45,6 @@
 
     @Override
     public HttpResponse execute(HttpRequest req) {
-        return execute(null, req);
-    }
-
-    private HttpResponse execute(@Nullable EventLoop eventLoop, HttpRequest req) {
```
ikhoon: Nice! 👍

core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java
```diff
@@ -0,0 +1,83 @@
...
+        if (customizers != null) {
+            for (int i = customizers.size() - 1; i >= 0; i--) {
+                if (customizers.get(i) == customizer) {
+                    customizers.remove(i);
+                    return;
+                }
+            }
+        }
+
+        // Should not reach here, but may happen if a user tried to call from a wrong thread.
```
anuraaga: Is it actually possible? Isn't `close` in the `SafeCloseable` guaranteed to be on the right thread? If it's only possible by a bug in Armeria, we should say that instead of `user`.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
anuraaga: We may as well document it, I think it's something like `// It is more common to have no customizers than to have any. So instead of defaulting to an empty list and removing null checks, we use null checks which are faster than checking if a list is empty.`

core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java
```diff
@@ -0,0 +1,83 @@
...
+import javax.annotation.Nullable;
+
+import com.google.common.collect.ImmutableList;
+
+final class ClientRequestContextCustomizers {
+
+    @Nullable
+    private ArrayList<Consumer<? super ClientRequestContext>> customizers;
+
+    private boolean pendingContextCapture;
```
anuraaga: I think customizers and context capture don't seem related - do we use the same object to reduce some object allocations? If so, can probably document that.

thrift/src/test/java/com/linecorp/armeria/client/thrift/ThriftOverHttpClientTest.java
```diff
@@ -303,6 +305,50 @@ public void onError(Exception exception) {
...
+    @ParameterizedTest
+    @ArgumentsSource(ParametersProvider.class)
+    void contextCaptorSync(
+            ClientFactory clientFactory, SerializationFormat format, SessionProtocol protocol)
+            throws Exception {
+        withTimeout(() -> {
+            final HelloService.Iface client = Clients.newClient(clientFactory,
+                                                                uri(Handlers.HELLO, format, protocol),
+                                                                Handlers.HELLO.iface(), clientOptions);
+            Clients.captureNextContext();
```
anuraaga: Just brainstorming, would using a capturor be easier to understand than two static methods? (it would look more like mockito which is common knowledge in Java so it could be a good mental model reference point for users).

```java
ContextCaptor captor = Clients.newContextCaptor();
client.hello();
ctx = captor.get();
```

We could also consider this pattern - I personally like it the best since it gives explicit scoping to capturing, but it could feel a bit obtuse.

```java
final ClientRequestContext ctx1;
final ClientRequestContext ctx2;

try (ContextCaptor captor = Clients.newContextCaptor()) {
  client.hello("kukuman");
  ctx1 = captor.get();
  client.hello("fufuman");
  ctx2 = captor.get();

  for (ClientRequestContext ctx : captor.getAll()) {
    System.out.println(ctx);
  }
}
```
```

File: core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java (hunk 8 above)
Comment (anuraaga) [reply core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java]: Realized maybe it's just the class name, `ClientThreadLocalState` so it doesn't apply a semantic concept like customization only?

File: core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java (hunk 7 above)
Comment (anuraaga) [reply core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java]: Actually I guess it could be manually called on a different thread by user too - would be quite fun code ;) In that case, maybe

```
throw new IllegalStateException("Failed to remove... did you call SafeCloseable.close() manually on a different thread? Use try/resources or otherwise make sure SafeCloseable.close() is called on the same thread as the one that created it
```

File: thrift/src/test/java/com/linecorp/armeria/client/thrift/ThriftOverHttpClientTest.java (hunk 9 above)
Comment (trustin) [reply thrift/src/test/java/com/linecorp/armeria/client/thrift/ThriftOverHttpClientTest.java]: The second one was the initial prototype but it was way too verbose, so I ditched it. :sweat_smile: 

The first one looks nice. Let me give it a try.

File: core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java (hunk 7 above)
Comment (trustin) [reply core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java]: Yeah, a user can always choose to abuse `SafeCloseable` without `try`. :laughing:

File: thrift/src/test/java/com/linecorp/armeria/client/thrift/ThriftOverHttpClientTest.java (hunk 9 above)
Comment (trustin) [reply thrift/src/test/java/com/linecorp/armeria/client/thrift/ThriftOverHttpClientTest.java]: Done

File: core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java (hunk 8 above)
Comment (trustin) [reply core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java]: Done

File: core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java (hunk 7 above)
Comment (trustin) [reply core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java]: Done

File: core/src/main/java/com/linecorp/armeria/client/ClientRequestContextCustomizers.java [hunk 10]
```diff
@@ -0,0 +1,83 @@
...
+    void add(Consumer<? super ClientRequestContext> customizer) {
+        if (customizers == null) {
+            customizers = new ArrayList<>();
+        }
+        customizers.add(customizer);
+    }
+
+    void remove(Consumer<? super ClientRequestContext> customizer) {
+        if (customizers != null) {
+            for (int i = customizers.size() - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
trustin: Done

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: Done

trustin:
@anuraaga @minwoox @ikhoon PTAL again.

core/src/main/java/com/linecorp/armeria/client/UserClient.java
```diff
@@ -126,26 +125,24 @@ protected final Endpoint endpoint() {
...
-     * @param eventLoop the {@link EventLoop} to execute the {@link Request}
      * @param endpoint the {@link Endpoint} of the {@link Request}
      * @param method the method of the {@link Request}
      * @param path the path part of the {@link Request} URI
      * @param query the query part of the {@link Request} URI
      * @param fragment the fragment part of the {@link Request} URI
      * @param req the {@link Request}
      * @param fallback the fallback response {@link BiFunction} to use when
-     *                 {@link Client#execute(ClientRequestContext, Request)} of {@link #delegate()} throws
+     *        {@link Client#execute(ClientRequestContext, Request)} of {@link #delegate()} throws
```
minwoox: nit: revert?

thrift/src/test/java/com/linecorp/armeria/client/thrift/ThriftOverHttpClientTest.java
anuraaga: I realized if the client requests failed with an exception, there will be a context leaked to ThreadLocal. It's not huge since it gets reset next time but in a server with a lot of threads I guess it could be somewhat significant.

Are you sure with not using `SafeCloseable` to ensure the threadlocal is cleaned up?

core/src/main/java/com/linecorp/armeria/client/Clients.java
```diff
@@ -581,18 +582,51 @@ public static SafeCloseable withHttpHeaders(Function<HttpHeaders, HttpHeaders> h
...
+            boolean closed;
+
+            @Override
+            public void close() {
+                if (closed) {
+                    return;
+                }
+
+                closed = true;
+                customizers.remove(contextCustomizer);
```
anuraaga: Should we clean up threadlocal if no more customizers?

trustin: Good point. Yeah, we should because otherwise we may leak in a container.

trustin: Actually, that wouldn't prevent a leak. A user has to make sure to close all `SafeCloseable`s and make a call to clear a pending captor.

trustin: That being said, I think it's safe to leave the thread-local state.

trustin: Just realized that the thread will still have a reference to `ClientThreadLocalState` instance itself, so it's not really safe. :sweat_smile:

trustin: Done

core/src/main/java/com/linecorp/armeria/client/UserClient.java
trustin: Done

core/src/main/java/com/linecorp/armeria/client/Clients.java
anuraaga: It's normal in Java to require all closeables to be closed to prevent leaks - Java doesn't have proper RAII unfortunately and try/resources is just ok but is all we've got.

If we have a case here where `Captor.get` needs to be called to prevent a leak, I really feel instead it should be `SafeCloseable` instead of `Supplier` to make this explicit and allow users to use try/resources which is safe even with exceptions. It's a little more verbose sure, but it's just how Java is (Java 10+ help a lot with `var` and have far less of the burden anyways). 

Anyways, if you feel strongly about the current API I'm fine with merging but want to point this out.

trustin: > If we have a case here where `Captor.get` needs to be called to prevent a leak, I really feel instead it should be `SafeCloseable` instead of `Supplier` to make this explicit and allow users to use try/resources which is safe even with exceptions. 

Agreed. A user currently must call `Captor.get()` to prevent a leak in the following cases:

- A user specified an illegal request path.
- A user specified a request URI without an authority when a `WebClient` was created without the default authority.

We could 1) make it `SafeCloseable` or 2) make sure a context is created for all cases and consider it's a bug if a context is not created.

anuraaga: +1 for `SafeCloseable`. If I'm not mistaken, there are more cases for exceptions since it seems reasonable for a user to call aggregate.join on the same line, or they could be using a blocking stub.

trustin: Thread local will be cleaned up as soon as a new context is ceeated, so aggregate() or anything that happens after context creation is irrelevant.

trustin: I think it's better making it `SafeCloseable` because we can't have complete control over what exceptions are gonna be triggered before a context is created. Please check the new API: https://github.com/line/armeria/pull/2344/commits/a3e4be169cde43af4adb10f8144a93be4d3992c2

core/src/main/java/com/linecorp/armeria/client/ClientThreadLocalState.java
```diff
@@ -0,0 +1,149 @@
...
+
+import javax.annotation.Nullable;
+
+import com.google.common.collect.ImmutableList;
+
+import com.linecorp.armeria.common.util.SafeCloseable;
+
+final class ClientThreadLocalState {
+
+    private static final ThreadLocal<ClientThreadLocalState> threadLocalState = new ThreadLocal<>();
```
anuraaga: I guess it might be 50/50 whether this is called from ab event loop or not. Should we use `FastThreadLocal`?

trustin:
Added a few utility methods to `ClientRequestContextCaptor` - https://github.com/line/armeria/pull/2344/commits/1de2333480b358ce6a3cd6c695c9505c9ecd7c16

trustin:
Thanks for reviewing!

PR #2349, opened 2019-12-29
Migrate all tests in examples to JUnit 5

### LINKED ISSUE #2346: Migrate all tests in the examples to JUnit 5
The `spring-boot-minimal-kotlin` example started using JUnit 5, but our example conversion script still uses JUnit 5. It'd be nice to migrate all examples to JUnit 5 for consistency.

heowc:
- Remove JUnit4 dependency
- Migrate all tests in examples to JUnit 5
- Fix `checkstyle` configuration on `spring-boot-mininal-kotlin`
- Closes #2346


heowc:
This PR is simply migrated. 

Please mention if you need other modifications. I'll be happy to fix it. :smile:

examples/spring-boot-minimal/src/test/java/example/springframework/boot/minimal/HelloApplicationIntegrationTest.java
```diff
@@ -5,36 +5,33 @@
...
-    @Before
-    public void setUp() {
+    @BeforeEach
+    void setUp() {
         client = WebClient.of("http://localhost:" + server.activeLocalPort());
     }
 
     @Test
-    public void success() {
+    void success() {
```
ikhoon: nit: Remove `public` on the following method `failure`

examples/grpc/build.gradle
```diff
@@ -10,7 +10,6 @@ dependencies {
     runtime 'org.slf4j:slf4j-simple'
 
     testCompile 'jakarta.annotation:jakarta.annotation-api'
-    testCompile 'junit:junit'
```
minwoox: I think we should add JUnit 5 dependencies instead of this such as `org.junit.jupiter` and `org.junit.platform`. Please see https://github.com/line/armeria/blob/master/build.gradle#L118
Also could you run the script in `https://github.com/line/armeria-examples/blob/master/update-examples.sh` so that everything working as expected? Please, be sure to change [this](https://github.com/line/armeria-examples/blob/master/update-examples.sh#L8) and [this](https://github.com/line/armeria-examples/blob/master/update-examples.sh#L93) before running the script. 😄

heowc: I will modify `update-examples.sh` of `armeria-example` and come back again 🚀

trustin: `update-examples.sh` is not part of this repository and @heowc added other commits since the last review. @minwoox, could you take another look?

heowc: I add `jupiter-api` dependency from each example and sync master.

heowc: ... And I cloned `armeria-examples` to modify `update-examples.sh`. However, there is an exception to the operation and we are checking it.

minwoox: > update-examples.sh is not part of this repository

I agree. 😄 I just wanted to make sure that we do not miss any dependencies by running the script.

heowc: > I just wanted to make sure that we do not miss any dependencies by running the script.

I agree with this. In fact, I have a lot of changes and issues because I wrote in Kotlin.

I may be lacking or missing something, but if it's written in the legacy application style in the Kotlin build script, I can't use the `springBoot` scope. This is a problem on [this line](https://github.com/line/armeria-examples/blob/master/update-examples.sh#L114), so you have to find another way or problem.

minwoox: Ah yes, we can modify the script later. 😄

PR #2354, opened 2019-12-31
Integrate push and pushIfAbsent in RequestContext

minwoox:
Motivation:
After we added `ClientRequestContext.root()`, we can restrict the prerequisites when `RequestContext.push()` is called.

Modifications:
- Deprecate `RequestContext.pushIfAbsent()`
- Enforce the current thread-local state to call `RequestContext.push()` to be one of the following conditions:
  - For `ServiceRequestContext`
    - the thread-local does not have any `RequestContext` in it
    - the thread-local has the same `ServiceRequestContext` - reentrance
    - the thread-local has the `ClientRequestContext` whose `ClientRequestContext.root()` is the same `ServiceRequestContext`
  - For `ClientRequestContext`
    - the thread-local does not have any `RequestContext` in it
    - the thread-local has the same `ClientRequestContext` - reentrance
    - the thread-local has the `ServiceRequestContext` which is the same as `root()`
    - the thread-local has the `ClientRequestContext` whose `root()` is the same `root()`
    - the thread-local has the `ClientRequestContext` whose `root()` is `null` and this `root()` is `null`
- Remove `AbstraceReqeustContext`
  - Deprecated `RequestContext.isTimedOut()` is gone.
- `RequestContext.onChild()` is gone.
  - Use `ServiceRequestContext.onChild(BiConsumer<? super ServiceRequestContext, ? super ClientRequestContext> callback)` instead.
- Add `RequestContext.replace()` which replaces the current thread-local context without any valiation.

Result:
- (Breaking)
  - `RequestContext.onChild()` and `RequestContext.invokeOnChildCallbacks()` are gone.
    - Use `ServiceRequestContext.onChild(BiConsumer<? super ServiceRequestContext, ? super ClientRequestContext> callback)`.
    - Use `ServiceRequestContext.invokeOnChildCallbacks(ClientRequestContext newCtx)`
  - Deprecated `RequestContext.isTimedOut()` is gone.
    - Use `ServiceRequestContext.isTimedOut()`.
- (Deprecation)
  - `RequestContext.pushIfAbsent()` is deprecated. Use `RequestContext.push()`.
- Fix a bug where `RequestContext.push()` does not throw an exception when it's called when the status of the current thread-local is invalid.
- Fix a bug described in https://github.com/line/armeria/pull/2282


core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -289,6 +295,66 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
     @Override
     RpcRequest rpcRequest();
 
+    /**
+     * Pushes the specified context to the thread-local stack. To pop the context from the stack, call
```
trustin: the specified context -> this context

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -289,6 +295,66 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     * <ul>
+     *   <li>the thread-local does not have any {@link RequestContext} in it</li>
+     *   <li>the thread-local has the same {@link ClientRequestContext} as this - reentrance</li>
+     *   <li>the thread-local has the {@link ServiceRequestContext} which is the same as {@link #root()}</li>
+     *   <li>the thread-local has the {@link ClientRequestContext} whose {@link #root()}
+     *       is the same {@link #root()}</li>
+     *   <li>the thread-local has the {@link ClientRequestContext} whose {@link #root()} is {@code null}
+     *       and this {@link #root()} is {@code null}</li>
+     * </ul>
+     * If it doesn't, this will throw an {@link IllegalStateException}.
```
trustin: Otherwise, this method will

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -163,6 +169,77 @@ default InetAddress clientAddress() {
         return remoteAddress.getAddress();
     }
 
+    /**
+     * Pushes the specified context to the thread-local stack. To pop the context from the stack, call
```
trustin: the specified context -> this context

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -163,6 +169,77 @@ default InetAddress clientAddress() {
...
+     *
+     * <p>In order to call this method, the current thread-local state must meet one of the
+     * following conditions:
+     * <ul>
+     *   <li>the thread-local does not have any {@link RequestContext} in it</li>
+     *   <li>the thread-local has the same {@link ServiceRequestContext} as this - reentrance</li>
+     *   <li>the thread-local has the {@link ClientRequestContext} whose {@link ClientRequestContext#root()}
+     *       is the same {@link ServiceRequestContext} as this</li>
+     * </ul>
+     * If it doesn't, this will throw an {@link IllegalStateException}.
```
trustin: Otherwise, this method will

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -289,6 +295,66 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
     @Override
     RpcRequest rpcRequest();
 
+    /**
+     * Pushes the specified context to the thread-local stack. To pop the context from the stack, call
+     * {@link SafeCloseable#close()}, which can be done using a {@code try-with-resources} block:
+     * <pre>{@code
+     * try (PushHandle ignored = ctx.push(true)) {
```
anuraaga: `SafeCloseable` (seems to be all I talk about lately :) )

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -289,6 +295,66 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     *
+     * <p>In order to call this method, the current thread-local state must meet one of the
+     * following conditions:
+     * <ul>
+     *   <li>the thread-local does not have any {@link RequestContext} in it</li>
+     *   <li>the thread-local has the same {@link ClientRequestContext} as this - reentrance</li>
+     *   <li>the thread-local has the {@link ServiceRequestContext} which is the same as {@link #root()}</li>
+     *   <li>the thread-local has the {@link ClientRequestContext} whose {@link #root()}
+     *       is the same {@link #root()}</li>
+     *   <li>the thread-local has the {@link ClientRequestContext} whose {@link #root()} is {@code null}
```
anuraaga: Is this correct? They are two unrelated contexts

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -289,6 +295,66 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
...
+     *                     {@link #onExit(Consumer)} will be invoked when the context is pushed to and
+     *                     removed from the thread-local stack respectively.
+     *                     NOTE: In case of reentrance, the callbacks will never run.
+     */
+    @Override
+    default SafeCloseable push(boolean runCallbacks) {
+        final RequestContext oldCtx = RequestContextThreadLocal.getAndSet(this);
+        if (oldCtx == this) {
+            // Reentrance
+            return () -> { /* no-op */ };
```
anuraaga: Let's make a constant

core/src/main/java/com/linecorp/armeria/server/composition/AbstractCompositeService.java
```diff
@@ -165,8 +165,12 @@ public O serve(ServiceRequestContext ctx, I req) throws Exception {
...
             final ServiceRequestContext newCtx = new CompositeServiceRequestContext(
                     ctx, newRoute, result.routingResult().path());
-            try (SafeCloseable ignored = newCtx.push(false)) {
+            final ServiceRequestContext oldCtx = RequestContextThreadLocal.getAndSet(newCtx);
+            assert oldCtx == ctx;
+            try {
                 return result.value().serve(newCtx, req);
+            } finally {
+                RequestContextThreadLocal.set(oldCtx);
             }
```
trustin: I guess we forgot to consider this case of replacing an existing root context with another one. Should we provide another method dedicated to this scenario?

core/src/test/java/com/linecorp/armeria/client/ClientRequestContextTest.java
```diff
@@ -18,31 +18,45 @@
...
                                                          .hasMessageContaining("unavailable");
 
-        final ClientRequestContext ctx = ClientRequestContext.of(req);
+        final ClientRequestContext ctx = clientRequestContext();
         assertThat(ctx.id()).isNotNull();
         try (SafeCloseable unused = ctx.push()) {
+            assertThat(onEnterExitStack.size()).isOne();
             assertThat(ClientRequestContext.current()).isSameAs(ctx);
         }
+        assertThat(onEnterExitStack.size()).isZero();
```
trustin: `assertThat(onEnterExitStack).isEmpty()`

core/src/test/java/com/linecorp/armeria/client/ClientRequestContextTest.java
```diff
@@ -18,31 +18,45 @@
...
     @Test
     void current() {
         assertThatThrownBy(ClientRequestContext::current).isInstanceOf(IllegalStateException.class)
                                                          .hasMessageContaining("unavailable");
 
-        final ClientRequestContext ctx = ClientRequestContext.of(req);
+        final ClientRequestContext ctx = clientRequestContext();
         assertThat(ctx.id()).isNotNull();
         try (SafeCloseable unused = ctx.push()) {
+            assertThat(onEnterExitStack.size()).isOne();
```
trustin: `assertThat(onEnterExitStack).hasSize(1)`

core/src/test/java/com/linecorp/armeria/client/ClientRequestContextTest.java
```diff
@@ -70,16 +86,214 @@ void mapCurrent() {
...
         }
     }
+
+    @Test
+    void pushReentrance() {
+        final ClientRequestContext ctx = clientRequestContext();
+        try (SafeCloseable ignored = ctx.push()) {
+            assertThat(onEnterExitStack.size()).isOne();
+            try (SafeCloseable ignored2 = ctx.push()) {
+                assertThat(onEnterExitStack.size()).isOne();
```
trustin: Could you add assertions that make sure `RequestContext.current()` returns the right context for all new test cases? e.g.

    assertThat(RequestContext.current()).isSameAs(ctx);

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -163,6 +169,77 @@ default InetAddress clientAddress() {
         return remoteAddress.getAddress();
     }
 
+    /**
+     * Pushes the specified context to the thread-local stack. To pop the context from the stack, call
+     * {@link SafeCloseable#close()}, which can be done using a {@code try-with-resources} block:
+     * <pre>{@code
+     * try (PushHandle ignored = ctx.push(true)) {
```
anuraaga: Ditto

core/src/test/java/com/linecorp/armeria/client/retry/RetryingHttpClientWithContextAwareTest.java
```diff
@@ -0,0 +1,65 @@
...
+        final WebClient client =
+                WebClient.builder(server.uri("/"))
+                         .responseTimeoutMillis(100)
+                         .decorator(RetryingClient.builder(RetryStrategy.onServerErrorStatus())
+                                                  .maxTotalAttempts(2)
+                                                  .newDecorator())
+                         .build();
+
+        final ServiceRequestContext dummyCtx =
+                ServiceRequestContext.builder(HttpRequest.of(HttpMethod.GET, "/")).build();
```
trustin: `ServiceRequestContext.of(HttpRequest)`?

core/src/test/java/com/linecorp/armeria/server/ServiceRequestContextTest.java
```diff
@@ -69,16 +82,86 @@ void mapCurrent() {
...
-        try (SafeCloseable unused = ClientRequestContext.of(req).push()) {
+        try (SafeCloseable unused = clientRequestContext().push()) {
             assertThatThrownBy(() -> ServiceRequestContext.mapCurrent(c -> "foo", () -> "bar"))
                     .isInstanceOf(IllegalStateException.class)
                     .hasMessageContaining("not a server-side context");
         }
     }
+
+    @Test
+    void pushReentrance() {
```
trustin: Ditto - would be nice to have more assertions for all new test cases.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: I think this could happen in test cases or where the client call is not triggered by a service
```java
final WebClient client =
        WebClient.builder(server.uri("/"))
                 .decorator(RetryingHttpClient...)
                 .build();

final CompletableFuture<AggregatedHttpResponse> future = client.get("/").aggregate();
future.handle((msg, cause) -> {
    // This thread has a clientRequestContext in the thread-local. #2282
    client2.execute(...) // ctx2.push() happens while ctx1 is in the thread-local
    return null;
});
```
So I think we can allow this condition not to throw an exception. Could you have another case that you concerned about?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Oops didn't notice it. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: 🙆‍♂

core/src/test/java/com/linecorp/armeria/server/ServiceRequestContextTest.java
minwoox: Found a bug after adding more test cases. Thanks!

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Oops~

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Fixed Thanks!

core/src/main/java/com/linecorp/armeria/server/composition/AbstractCompositeService.java
minwoox: Added! 😄

minwoox:
Thanks a lot for reviewing! 😉

PR #2355, opened 2019-12-31
Overall cleanup and deprecation

trustin:
- Breaking changes
  - Remove `CorsConfig.DISABLED` and `isEnabled()`
    - Disabled `CorsConfig` was never used.
  - Do not use thread context classloader in `Version.identify()`.
    - `ClassLoader` in `Version.identify()` is now non-null.
- Deprecations
  - `new EventCount()` -> `of()`
  - `PrometheusMeterRegistries.defaultRegisty` -> `defaultRegistry()`
  - `AuthTokenExtractors.{BASIC,OAUTH1A,OAUTH2}` -> `{basic,oAuth1a,oAuth2}()`
  - `new SimpleCompositeService()` -> `of()`
  - `new SimpleCompositeRpcService()` -> `of()`
  - Deprecate the two resource-loading constructors in `KeyStoreCredentialResolverBuilder`
    - Add `KeyStoreCredentialResolverBuilder(ClassLoader, String)` for
      consistency with other resource-loading classes such as `HttpFile`.
- Miscallaneous:
  - Add `KeyStoreCredentialResolverBuilder(Path)`
  - Make `KeyStoreCredentialResolverBuilder.build()` return a
    `KeyStoreCredentialResolver` instead of `CredentialResolver`.
  - Remove `DefaultAddressResolverGroupFactory` which is unused
  - Do not use deprecated API in the examples, tests and Javadoc.
  - Add missing `static` in `ZooKeeperUpdatingListener.builder()`


core/src/main/java/com/linecorp/armeria/client/DefaultAddressResolverGroupFactory.java
```diff
@@ -1,55 +0,0 @@
...
-import io.netty.resolver.AddressResolverGroup;
-import io.netty.resolver.dns.DnsAddressResolverGroup;
-import io.netty.resolver.dns.DnsNameResolverBuilder;
-import io.netty.resolver.dns.DnsServerAddressStreamProviders;
-
-/**
- * The default {@link AddressResolverGroup} factory implementation, which enables asynchronous
- * DNS resolution and tracing by default.
- */
-final class DefaultAddressResolverGroupFactory
```
minwoox: Oops thanks!

core/src/main/java/com/linecorp/armeria/server/auth/AuthTokenExtractors.java
```diff
@@ -27,22 +27,52 @@
...
+     * @deprecated Use {@link #oAuth2()}.
      */
+    @Deprecated
     public static final Function<RequestHeaders, OAuth2Token> OAUTH2 =
             new OAuth2TokenExtractor(HttpHeaderNames.AUTHORIZATION);
 
+    /**
+     * Returns a {@link BasicToken} extractor function.
+     */
+    public static Function<RequestHeaders, BasicToken> basic() {
```
minwoox: How about `Function<? super RequestHeaders, ? extends BasicToken>`

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
```diff
@@ -58,20 +59,40 @@ public KeyStoreCredentialResolverBuilder(File file) {
...
      * Creates a builder with the specified {@code resourcePath} and {@link ClassLoader}.
+     *
+     * @deprecated Use {@link #KeyStoreCredentialResolverBuilder(ClassLoader, String)}.
      */
+    @Deprecated
     public KeyStoreCredentialResolverBuilder(String resourcePath, @Nullable ClassLoader classLoader) {
-        this.resourcePath = requireNonNull(resourcePath, "resourcePath");
-        this.classLoader = classLoader;
-        file = null;
+        this(firstNonNull(classLoader, KeyStoreCredentialResolverBuilder.class.getClassLoader()), resourcePath);
```
ikhoon: nit: `classLoader != null ? classLoader : KeyStoreCredentialResolverBuilder.class.getClassLoader()` 
to avoid calling of `getClassLoader()` if `classLoader` is not null?

core/src/main/java/com/linecorp/armeria/server/auth/AuthTokenExtractors.java
trustin: Let me add `? super`. `? extends` is not necessary because it's final.

saml/src/main/java/com/linecorp/armeria/server/saml/KeyStoreCredentialResolverBuilder.java
trustin: Good point. Fixed

ikhoon:
Still LGTM.

trustin:
Thanks, @minwoox and @ikhoon

PR #2356, opened 2019-12-31
Implement class aware retrofit meter function

### LINKED ISSUE #2345: Provide more than `Invocation` to `MeterIdPrefixFunction`
`armeria-retrofit` provides `InvocationUtil` which allows a user retrieve Retrofit's `Invocation` object when writing `MeterIdPrefixFunction`, notably `RetrofitMeterIdPrefixFunction`. It'd be even nicer if we can provide more than that, such as:

- The route (path pattern) of the current request.
  - Requires some annotation scanning to retrieve the values of method annotations like `@GET(...)`.
- ... and more?

arhont375:
Add new meter function class for retrofit clients that provide more information:
- Raw path from the Retrofit annotaiton
- Consistent service name based on class name, by default with `service` tag
- httpMethod and httpStatus

This changes does not replace previous RetrofitMeterIdFunction as it would be breaking changes and maybe some users will prefer other variant :thinking: 

In addition to this Tag names for metric may be adjusted as well to comply with Armeria server format (currently use `path` and `httpMethod` whether Armeria server metrics provide `route` that contains path and http method at the same time)
Closes #2345

### DIFF
--- a/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
+++ b/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
@@ -33,7 +33,6 @@ import com.google.common.collect.ImmutableMap;
 import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.common.metric.MeterIdPrefix;
-import com.linecorp.armeria.common.metric.MeterIdPrefixFunction;
 import com.linecorp.armeria.internal.metric.RequestMetricSupport;
 
 import io.micrometer.core.instrument.MeterRegistry;
@@ -64,7 +63,7 @@ import retrofit2.http.PUT;
  *     <li>{@code httpStatus} - {@link HttpStatus#code()}</li>
  * </ul>
  */
-public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFunction {
+final class RetrofitClassAwareMeterIdPrefixFunction extends RetrofitMeterIdPrefixFunction {
 
     private static final List<Class<?>> RETROFIT_ANNOTATIONS = ImmutableList.of(
             POST.class, PUT.class, PATCH.class, HEAD.class, GET.class, OPTIONS.class, HTTP.class, DELETE.class
@@ -86,7 +85,7 @@ public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFun
      * Returns a newly created {@link RetrofitClassAwareMeterIdPrefixFunction} with the specified {@code name}
      * and {@code serviceClass}.
      */
-    public static MeterIdPrefixFunction of(String name, Class<?> serviceClass) {
+    public static RetrofitMeterIdPrefixFunction of(String name, Class<?> serviceClass) {
         return builder(name, serviceClass).build();
     }
 
@@ -102,6 +101,8 @@ public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFun
     RetrofitClassAwareMeterIdPrefixFunction(String name,
                                             @Nullable String serviceTagName,
                                             Class<?> serviceClass) {
+        super(name, null, null);
+
         this.name = name;
         this.serviceTagName = firstNonNull(serviceTagName, "service");
         this.serviceName = serviceClass.getSimpleName();
@@ -175,7 +176,7 @@ public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFun
                 httpPath = (String) valueMethod.invoke(annotation);
             } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
                 // Should never happen on valid Retrofit client.
-                throw new IllegalStateException("Provided Retrofit client have unexpected annotation: " +
+                throw new IllegalStateException("Unexpected retrofit annotation: " +
                                                 annotation.annotationType(), e);
             }
             httpMethod = annotation.annotationType().getSimpleName();
--- a/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
+++ b/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
@@ -29,7 +29,6 @@ import javax.annotation.Nullable;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableMap.Builder;
 
 import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.logging.RequestLog;
@@ -55,9 +54,13 @@ import retrofit2.http.PUT;
  * client.
  * <ul>
  *     <li>{@code service(or serviceTagName} - Retrofit service interface name</li>
- *     <li>{@code path}   - Retrofit service interface method path taken from method annotation</li>
- *     <li>{@code method} - Retrofit service interface method</li>
- *     <li>{@code httpMethod} - HTTP method name from Retrofit service interface method annotation</li>
+ *     <li>{@code path}   - Retrofit service interface method path taken from method annotation
+ *                          or {@code UNKNOWN} if Retrofit service interface method available</li>
+ *     <li>{@code method} - Retrofit service interface method
+ *                          or {@code UNKNOWN} if Retrofit service interface method available</li>
+ *     <li>{@code httpMethod} - HTTP method name from Retrofit service interface method annotation
+ *                          or from {@link RequestLog#method()} if Retrofit service interface
+ *                          method not available</li>
  *     <li>{@code httpStatus} - {@link HttpStatus#code()}</li>
  * </ul>
  */
@@ -65,6 +68,13 @@ public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFun
 
     private static final List<Class<?>> RETROFIT_ANNOTATIONS = ImmutableList.of(
             POST.class, PUT.class, PATCH.class, HEAD.class, GET.class, OPTIONS.class, HTTP.class, DELETE.class
+    );
+    private static final String METHOD_TAG_NAME = "method";
+    private static final String PATH_TAG_NAME = "path";
+    private static final String HTTP_METHOD_TAG_NAME = "httpMethod";
+    private static final List<Tag> DEFAULT_METHOD_TAGS = ImmutableList.of(
+            Tag.of(METHOD_TAG_NAME, "UNKNOWN"),
+            Tag.of(PATH_TAG_NAME, "UNKNOWN")
     );
 
     private final Map<String, List<Tag>> methodNameToTags;
@@ -126,12 +136,13 @@ public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFun
             }
         }
 
-        tagsBuilder.add(Tag.of("method", log.method().name()));
+        tagsBuilder.add(Tag.of(HTTP_METHOD_TAG_NAME, log.method().name()));
+        tagsBuilder.addAll(DEFAULT_METHOD_TAGS);
     }
 
     @VisibleForTesting
     static Map<String, List<Tag>> defineTagsForMethods(Class<?> serviceClass) {
-        final Builder<String, List<Tag>> methodNameToTags = ImmutableMap.builder();
+        final ImmutableMap.Builder<String, List<Tag>> methodNameToTags = ImmutableMap.builder();
 
         final Method[] declaredMethods = serviceClass.getDeclaredMethods();
         for (final Method clientServiceMethod : declaredMethods) {
@@ -142,7 +153,7 @@ public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFun
 
                 final ImmutableList.Builder<Tag> tags = ImmutableList.builder();
                 tags.addAll(tagsFromAnnotation(annotation));
-                tags.add(Tag.of("method", clientServiceMethod.getName()));
+                tags.add(Tag.of(METHOD_TAG_NAME, clientServiceMethod.getName()));
                 methodNameToTags.put(clientServiceMethod.getName(), tags.build());
             }
         }
@@ -171,8 +182,8 @@ public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFun
         }
 
         return ImmutableList.of(
-            Tag.of("httpMethod", httpMethod),
-            Tag.of("path", httpPath)
+                Tag.of(HTTP_METHOD_TAG_NAME, httpMethod),
+                Tag.of(PATH_TAG_NAME, httpPath)
         );
     }
 }
--- a/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
+++ b/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2019 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.client.retrofit2;
+
+import static com.google.common.base.MoreObjects.firstNonNull;
+import static java.util.Objects.requireNonNull;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Map;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMap.Builder;
+
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.logging.RequestLog;
+import com.linecorp.armeria.common.metric.MeterIdPrefix;
+import com.linecorp.armeria.common.metric.MeterIdPrefixFunction;
+import com.linecorp.armeria.internal.metric.RequestMetricSupport;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import retrofit2.Invocation;
+import retrofit2.http.DELETE;
+import retrofit2.http.GET;
+import retrofit2.http.HEAD;
+import retrofit2.http.HTTP;
+import retrofit2.http.OPTIONS;
+import retrofit2.http.PATCH;
+import retrofit2.http.POST;
+import retrofit2.http.PUT;
+
+/**
+ * Returns the default function for retrofit that creates a {@link MeterIdPrefix} with the specified name and
+ * the {@link Tag}s derived from the {@link RequestLog} properties, {@link Invocation} and provided retrofit
+ * client.
+ * <ul>
+ *     <li>{@code service(or serviceTagName} - Retrofit service interface name</li>
+ *     <li>{@code path}   - Retrofit service interface method path taken from method annotation</li>
+ *     <li>{@code method} - Retrofit service interface method</li>
+ *     <li>{@code httpMethod} - HTTP method name from Retrofit service interface method annotation</li>
+ *     <li>{@code httpStatus} - {@link HttpStatus#code()}</li>
+ * </ul>
+ */
+public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFunction {
+    private static final List<Class> retrofitAnnotations = ImmutableList.of(
+            POST.class, PUT.class, PATCH.class, HEAD.class, GET.class, OPTIONS.class, HTTP.class, DELETE.class
+    );
+    private final Map<String, List<Tag>> methodNameToTags;
+    private final String name;
+    private final String serviceName;
+    private final String serviceTagName;
+
+    /**
+     * Returns a newly created {@link RetrofitClassAwareMeterIdPrefixFunction} with the specified {@code name}
+     * and {@code serviceClass}.
+     */
+    public static MeterIdPrefixFunction of(String name, Class serviceClass) {
+        return builder(name, serviceClass).build();
+    }
+
+    /**
+     * Returns a newly created {@link RetrofitMeterIdPrefixFunctionBuilder} with the specified {@code name}
+     * and {@code serviceClass}.
+     */
+    public static RetrofitMeterIdPrefixFunctionBuilder builder(String name, Class serviceClass) {
+        return new RetrofitMeterIdPrefixFunctionBuilder(requireNonNull(name, "name"))
+                .withServiceClass(serviceClass);
+    }
+
+    RetrofitClassAwareMeterIdPrefixFunction(String name,
+                                            Class serviceClass,
+                                            String serviceTagName) {
+        this.name = name;
+        this.serviceTagName = firstNonNull(serviceTagName, "service");
+        this.serviceName = serviceClass.getSimpleName();
+        this.methodNameToTags = defineTagsForMethods(serviceClass);
+    }
+
+    @Override
+    public MeterIdPrefix activeRequestPrefix(MeterRegistry registry, RequestLog log) {
+        final ImmutableList.Builder<Tag> tagsListBuilder = ImmutableList.builder();
+        buildTags(tagsListBuilder, log);
+        return new MeterIdPrefix(name, tagsListBuilder.build());
+    }
+
+    @Override
+    public MeterIdPrefix apply(MeterRegistry registry, RequestLog log) {
+        final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builder();
+        buildTags(tagListBuilder, log);
+        RequestMetricSupport.appendHttpStatusTag(tagListBuilder, log);
+        return new MeterIdPrefix(name, tagListBuilder.build());
+    }
+
+    private void buildTags(ImmutableList.Builder<Tag> tagsBuilder, RequestLog log) {
+        final Invocation invocation = InvocationUtil.getInvocation(log);
+
+        tagsBuilder.add(Tag.of(serviceTagName, serviceName));
+        if (invocation != null) {
+            final String methodName = invocation.method().getName();
+            final List<Tag> methodTags = methodNameToTags.get(methodName);
+            if (methodTags != null) {
+                tagsBuilder.addAll(methodTags);
+                return;
+            }
+        }
+
+        tagsBuilder.add(Tag.of("method", log.method().name()));
+    }
+
+    @VisibleForTesting
+    static Map<String, List<Tag>> defineTagsForMethods(Class serviceClass) {
+        final Builder<String, List<Tag>> methodNameToTags = ImmutableMap.builder();
+
+        final Method[] declaredMethods = serviceClass.getDeclaredMethods();
+        for (final Method clientServiceMethod : declaredMethods) {
+            final Annotation[] declaredAnnotations = clientServiceMethod.getDeclaredAnnotations();
+            if (declaredAnnotations.length == 0) {
+                continue;
+            }
+
+            for (final Annotation annotation : declaredAnnotations) {
+                if (!retrofitAnnotations.contains(annotation.annotationType())) {
+                    continue;
+                }
+
+                final ImmutableList.Builder<Tag> tags = ImmutableList.builder();
+                tags.addAll(tagsFromAnnotation(annotation));
+                tags.add(Tag.of("method", clientServiceMethod.getName()));
+                methodNameToTags.put(clientServiceMethod.getName(), tags.build());
+            }
+        }
+
+        return methodNameToTags.build();
+    }
+
+    private static List<Tag> tagsFromAnnotation(Annotation annotation) {
+        final String httpMethod;
+        final String httpPath;
+        if (annotation.annotationType().equals(HTTP.class)) {
+            final HTTP http = (HTTP) annotation;
+            httpMethod = http.method();
+            httpPath = http.path();
+        } else {
+            final Method valueMethod;
+            try {
+                valueMethod = annotation.getClass().getMethod("value");
+                httpPath = (String) valueMethod.invoke(annotation);
+            } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+                // Should never happen on valid Retrofit client.
+                throw new IllegalStateException("Provided Retrofit client have unexpected annotation: " +
+                                                annotation.annotationType(), e);
+            }
+            httpMethod = annotation.annotationType().getSimpleName();
+        }
+
+        return ImmutableList.of(
+            Tag.of("httpMethod", httpMethod),
+            Tag.of("path", httpPath)
+        );
+    }
+}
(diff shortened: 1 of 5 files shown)


retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,179 @@
...
+    @VisibleForTesting
+    static Map<String, List<Tag>> defineTagsForMethods(Class serviceClass) {
+        final Builder<String, List<Tag>> methodNameToTags = ImmutableMap.builder();
+
+        final Method[] declaredMethods = serviceClass.getDeclaredMethods();
+        for (final Method clientServiceMethod : declaredMethods) {
+            final Annotation[] declaredAnnotations = clientServiceMethod.getDeclaredAnnotations();
+            if (declaredAnnotations.length == 0) {
+                continue;
+            }
```
skkap: I wonder if this can be removed without affecting the logic and performance?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,179 @@
...
+ * <ul>
+ *     <li>{@code service(or serviceTagName} - Retrofit service interface name</li>
+ *     <li>{@code path}   - Retrofit service interface method path taken from method annotation</li>
+ *     <li>{@code method} - Retrofit service interface method</li>
+ *     <li>{@code httpMethod} - HTTP method name from Retrofit service interface method annotation</li>
+ *     <li>{@code httpStatus} - {@link HttpStatus#code()}</li>
+ * </ul>
+ */
+public class RetrofitClassAwareMeterIdPrefixFunction implements MeterIdPrefixFunction {
+    private static final List<Class> retrofitAnnotations = ImmutableList.of(
```
skkap: It looks like in Armeria it is common to add an empty line here as you did in `RetrofitClassAwareMeterIdPrefixFunctionTest`.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
trustin: +1 for removing for simplicity

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,179 @@
...
+        if (invocation != null) {
+            final String methodName = invocation.method().getName();
+            final List<Tag> methodTags = methodNameToTags.get(methodName);
+            if (methodTags != null) {
+                tagsBuilder.addAll(methodTags);
+                return;
+            }
+        }
+
+        tagsBuilder.add(Tag.of("method", log.method().name()));
```
trustin: A meter must have same set of tags regardless of whether `Invocation` is available or not, because otherwise Prometheus will complain.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,179 @@
...
+            httpPath = http.path();
+        } else {
+            final Method valueMethod;
+            try {
+                valueMethod = annotation.getClass().getMethod("value");
+                httpPath = (String) valueMethod.invoke(annotation);
+            } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+                // Should never happen on valid Retrofit client.
+                throw new IllegalStateException("Provided Retrofit client have unexpected annotation: " +
+                                                annotation.annotationType(), e);
```
trustin: Isn't it better skipping rather than raising an exception? A user may have added non-Retrofit annotations for other purposes.

trustin:
How about this:

- Make `RetrofitMeterIdPrefixFunctionBuilder.withServiceClass()` package-private to hide it from the public API
- Add `of(String name, Class serviceClass)` and `builder(String name, Class<?> serviceClass)` to `RetrofitMeterIdPrefixFunction`
  - and remove the factory methods in `RetrofitClassAwareMeterIdPrefixFunction`
- Make `RetrofitClassAwareMeterIdPrefixFunction` package-private to hide it from the public API

..?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
```diff
@@ -51,10 +54,23 @@ public RetrofitMeterIdPrefixFunctionBuilder withServiceTag(String serviceTagName
...
+        this.serviceClass = requireNonNull(serviceClass, "serviceClass");
+        return this;
+    }
+
     /**
      * Returns a newly created {@link RetrofitMeterIdPrefixFunction} with the properties specified so far.
      */
-    public RetrofitMeterIdPrefixFunction build() {
-        return new RetrofitMeterIdPrefixFunction(name, serviceTagName, defaultServiceName);
+    public MeterIdPrefixFunction build() {
```
trustin: We can continue returning `RetrofitMeterIdPrefixFunction` if we make `RetrofitClassAwareMeterIdPrefixFunction` extend `RetrofitMeterIdPrefixFunction`. Let's make `RetrofitMeterIdPrefixFunction` non-final with package-private constructor and let `RetrofitClassAwareMeterIdPrefixFunction` extend it.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
```diff
@@ -48,7 +48,7 @@
     /**
      * Returns a newly created {@link RetrofitMeterIdPrefixFunction} with the specified {@code name}.
      */
-    public static RetrofitMeterIdPrefixFunction of(String name) {
+    public static MeterIdPrefixFunction of(String name) {
```
trustin: We can continue returning `RetrofitMeterIdPrefixFunction` if we make `RetrofitClassAwareMeterIdPrefixFunction` extend `RetrofitMeterIdPrefixFunction`. Let's make `RetrofitMeterIdPrefixFunction` non-final with package-private constructor and let `RetrofitClassAwareMeterIdPrefixFunction` extend it.

trustin:
@kojilin PTAL as well.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,179 @@
...
+import retrofit2.http.PUT;
+
+/**
+ * Returns the default function for retrofit that creates a {@link MeterIdPrefix} with the specified name and
+ * the {@link Tag}s derived from the {@link RequestLog} properties, {@link Invocation} and provided retrofit
+ * client.
+ * <ul>
+ *     <li>{@code service(or serviceTagName} - Retrofit service interface name</li>
+ *     <li>{@code path}   - Retrofit service interface method path taken from method annotation</li>
+ *     <li>{@code method} - Retrofit service interface method</li>
```
kojilin: Looks like this may fall back to http method name if `Invocation` is null?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,190 @@
...
+                // Should never happen on valid Retrofit client.
+                throw new IllegalStateException("Unexpected retrofit annotation: " +
+                                                annotation.annotationType(), e);
+            }
+            httpMethod = annotation.annotationType().getSimpleName();
+        }
+
+        return ImmutableList.of(
+                Tag.of(HTTP_METHOD_TAG_NAME, httpMethod),
+                Tag.of(PATH_TAG_NAME, httpPath)
```
arhont375: How about 
> This decorator currently use `path` and `httpMethod` labels whether Armeria server metrics provide `route` label that contains path and http method at the same time

Makes sense to unify? Or just add another label :thinking:  I feel that it maybe easier to make queries when path and httpMethod separated

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
trustin: Ah I see. Thanks for explanation.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,190 @@
...
+import retrofit2.http.PATCH;
+import retrofit2.http.POST;
+import retrofit2.http.PUT;
+
+/**
+ * Returns the default function for retrofit that creates a {@link MeterIdPrefix} with the specified name and
+ * the {@link Tag}s derived from the {@link RequestLog} properties, {@link Invocation} and provided retrofit
+ * client.
+ * <ul>
+ *     <li>{@code service(or serviceTagName} - Retrofit service interface name</li>
```
trustin: `{@code service} (or {@code serviceTagName})`

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,190 @@
...
+ *     <li>{@code path}   - Retrofit service interface method path taken from method annotation
+ *                          or {@code UNKNOWN} if Retrofit service interface method available</li>
+ *     <li>{@code method} - Retrofit service interface method
+ *                          or {@code UNKNOWN} if Retrofit service interface method available</li>
+ *     <li>{@code httpMethod} - HTTP method name from Retrofit service interface method annotation
+ *                          or from {@link RequestLog#method()} if Retrofit service interface
+ *                          method not available</li>
```
trustin: Incorrect indentation at the 2nd and 3rd line

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
trustin: Agreed. Separate tags sounds good to me.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
trustin: This is not fixed yet.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
```diff
@@ -51,10 +53,24 @@ public RetrofitMeterIdPrefixFunctionBuilder withServiceTag(String serviceTagName
         return this;
     }
 
+    /**
+     * Adds a Retrofit interface service class that will allow to record actual path for each method in
+     * addition to method name.
+     *
+     * @param serviceClass the class defining client with Retrofit.
+     */
+    public RetrofitMeterIdPrefixFunctionBuilder withServiceClass(Class<?> serviceClass) {
```
trustin: What do you think about making this method package-private and adding `RetrofitMeterIdPrefixFunction.builder(String, Class)` and `of(String, Class)` which sets the service class at construction time?

arhont375: If I understood your idea correctly we don't need this method at all if will provide new package private constructor.

arhont375: Ah, I understood the way I changed will not work because class package private.
At the same time it feels strange to me to add such builder methods to the `RetrofitMeteridPrefixFunction` class because as a result we would return another specialization.

I feel that it would be better to make `RetrofitClassAwareMeterIdPrefixFunction` class either public or provide method to build this class in the builder (actually revert my last changes and add one more method to builder).

What do you think? Will you prefer way that you suggested initially or add more methods to builder?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @Override
+    public MeterIdPrefix apply(MeterRegistry registry, RequestLog log) {
+        final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builder();
+        buildTags(tagListBuilder, log);
+        RequestMetricSupport.appendHttpStatusTag(tagListBuilder, log);
+        return new MeterIdPrefix(name, tagListBuilder.build());
+    }
+
+    private void buildTags(ImmutableList.Builder<Tag> tagsBuilder, RequestLog log) {
```
ikhoon: How about changing `RetrofitMeterIdPrefixFunction.buildTags` to `protected`? 
And remove the overrided methods:
- `activeRequestPrefix(MeterRegistry, RequestLog)`
- `apply(MeterRegistry, RequestLog)`?

retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunctionTest.java
```diff
@@ -0,0 +1,241 @@
...
+import retrofit2.http.DELETE;
+import retrofit2.http.GET;
+import retrofit2.http.HEAD;
+import retrofit2.http.HTTP;
+import retrofit2.http.OPTIONS;
+import retrofit2.http.PATCH;
+import retrofit2.http.POST;
+import retrofit2.http.PUT;
+
+public class RetrofitClassAwareMeterIdPrefixFunctionTest {
```
ikhoon: nit
```suggestion 
class RetrofitClassAwareMeterIdPrefixFunctionTest {
```

retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunctionTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    private static final SimpleMeterRegistry meterRegistry = new SimpleMeterRegistry();
+    private static final ClientFactory clientFactory = ClientFactory.builder()
+                                                                    .meterRegistry(meterRegistry)
+                                                                    .build();
+
+    interface Example {
```
ikhoon: nit
```suggestion
    private interface Example {
```

retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunctionTest.java
```diff
@@ -0,0 +1,241 @@
...
+                @Override
+                protected HttpResponse doPost(ServiceRequestContext ctx, HttpRequest req) {
+                    return HttpResponse.of(HttpStatus.OK);
+                }
+
+                @Override
+                protected HttpResponse doTrace(ServiceRequestContext ctx, HttpRequest req) {
+                    return HttpResponse.of(HttpStatus.OK);
+                }
+            });
```
ikhoon: nit: 
1) If you want to handle all methods
```java
sb.service("/foo", (ctx, req) -> HttpResponse.of(HttpStatus.OK));
```
2) If you want to handle `GET`, `POST` and `TRACE` together
```java
sb.service(Route.builder().path("/foo")
                .methods(HttpMethod.GET, HttpMethod.POST, HttpMethod.TRACE)
                .build(),
           (ctx, req) -> HttpResponse.of(HttpStatus.OK));
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,190 @@
...
+    private final Map<String, List<Tag>> methodNameToTags;
+    private final String name;
+    private final String serviceName;
+    private final String serviceTagName;
+
+    /**
+     * Returns a newly created {@link RetrofitClassAwareMeterIdPrefixFunction} with the specified {@code name}
+     * and {@code serviceClass}.
+     */
+    public static RetrofitMeterIdPrefixFunction of(String name, Class<?> serviceClass) {
```
minwoox: A user cannot create this because `RetrofitClassAwareMeterIdPrefixFunction` is package-private. We should move this method upto`RetrofitMeterIdPrefixFunction`.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,190 @@
...
+        return builder(name, serviceClass).build();
+    }
+
+    /**
+     * Returns a newly created {@link RetrofitMeterIdPrefixFunctionBuilder} with the specified {@code name}
+     * and {@code serviceClass}.
+     */
+    public static RetrofitMeterIdPrefixFunctionBuilder builder(String name, Class<?> serviceClass) {
```
minwoox: ditto

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    RetrofitClassAwareMeterIdPrefixFunction(String name,
+                                            @Nullable String serviceTagName,
+                                            Class<?> serviceClass) {
+        super(name, null, null);
+
+        this.name = name;
+        this.serviceTagName = firstNonNull(serviceTagName, "service");
+        this.serviceName = serviceClass.getSimpleName();
```
minwoox: `this.serviceName` -> `serviceName`

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
```diff
@@ -33,11 +33,18 @@
...
+    @Nullable
+    private Class<?> serviceClass;
 
     RetrofitMeterIdPrefixFunctionBuilder(String name) {
         this.name = name;
     }
 
+    RetrofitMeterIdPrefixFunctionBuilder(String name, Class<?> serviceClass) {
+        this.name = name;
+        this.serviceClass = serviceClass;
```
minwoox: We should make sure that both `serviceClass` and `defaultServiceName` are not set together.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Returns a newly created {@link RetrofitMeterIdPrefixFunctionBuilder} with the specified {@code name}
+     * and {@code serviceClass}.
+     */
+    public static RetrofitMeterIdPrefixFunctionBuilder builder(String name, Class<?> serviceClass) {
+        return new RetrofitMeterIdPrefixFunctionBuilder(requireNonNull(name, "name"),
+                                                        requireNonNull(serviceClass, "serviceClass"));
+    }
```
trustin: These two classes could be removed?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
arhont375: Maybe I'm wrong, but I don't think it would affect usability:
If user use this builder then it's already clear that he want to use `serviceClass` but he still able to modify `serviceTag` with `withServiceTag` method.

Maybe comment on `withServiceTag` will help on this to clarify?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
arhont375: Usage of overriding methods gives us ability for a bit of microoptimization by providing expected list size. That by claims from guava may work faster then unbounded immutable list builder.

trustin: I guess it's fine as it is unless @ikhoon has a strong opinion on this.

ikhoon: > Usage of overriding methods gives us ability for a bit of microoptimization by providing expected list size. That by claims from guava may work faster then unbounded immutable list builder.

Sorry I missed this message. 

> I guess it's fine as it is unless @ikhoon has a strong opinion on this.

I don't have strong opinion here. Let's leave it as it is. :-)

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
trustin: Isn't `defaultServiceName` unused at all when `serviceClass` is used? How about:

- throwing an `IllegalStateException("defaultServiceName cannot be set for a builder created with serviceClass.")` for `withServiceTag(String, String)`,
- and adding `withServiceTag(String serviceTagName)` which works for both builder types.
  - If created without serviceClass, simply delegate to `withServiceTag(serviceTagName, "UNKNOWN")`

..?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,172 @@
...
+    private final String serviceTagName;
+
+    RetrofitClassAwareMeterIdPrefixFunction(String name,
+                                            @Nullable String serviceTagName,
+                                            Class<?> serviceClass) {
+        super(name, null, null);
+
+        this.name = name;
+        this.serviceTagName = firstNonNull(serviceTagName, "service");
+        serviceName = serviceClass.getSimpleName();
```
trustin: Is it possible for a user to use an anonymous class as a service class? In that case, `getSimpleName` will return an empty string. Perhaps we should strip only the package names in that case?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
arhont375: Why not to split this method into 2 and mark original `withServiceTag(String, String)` as deprecated?
- `withServiceTag(String serviceTag)`
- `withServiceName(String serviceName)` <-- I was curious why we need to provide `defaultServiceName` that used only in cases when `Invocation` not found (see [code](https://github.com/line/armeria/blob/master/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java#L106)). Let's just provide ability to use client provided service name that would be used for all cases and don't need to take name from the `Invocation` log then

arhont375: So this code from the link above:
```java
        if (invocation != null) {
            if (serviceTagName != null) {
                tagListBuilder.add(Tag.of(serviceTagName,
                                          invocation.method().getDeclaringClass().getSimpleName()));
            }
            tagListBuilder.add(Tag.of("method", invocation.method().getName()));
        } else {
            if (serviceTagName != null) {
                assert defaultServiceName != null;
                tagListBuilder.add(Tag.of(serviceTagName, defaultServiceName));
            }
            tagListBuilder.add(Tag.of("method", log.method().name()));
        }
```
Would be simplified to:
```
        if (serviceTagName != null) {
               tagListBuilder.add(Tag.of(serviceTagName, serviceName));
        }
        if (invocation != null) {
            tagListBuilder.add(Tag.of("method", invocation.method().getName()));
        } else {
            tagListBuilder.add(Tag.of("method", log.method().name()));
        }
```

arhont375: Unfortunately we can't make this ^ change now as it will break current metrics (as it may happen that provided serviceName would be different), but I think it should be way to go in the future

trustin: > Why not to split this method into 2 and mark original `withServiceTag(String, String)` as deprecated?

Sounds good to me!

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
minwoox: Oops, I found out that we don't use exact number for the expected size [here](https://github.com/line/armeria/blob/master/retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java#L89).
Could you change that as well please?
On my second thought, perhaps we can set the `builderWithExpectedSize` with the number + 1 in `RetrofitMeterIdPrefixFunction`, so that we can remove `activeRequestPrefix()` and `apply()` in `RetrofitClassAwareMeterIdPrefixFunction` and just overriding `buildTags()`.
Either way is fine for me though. 😉

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
minwoox: How about deprecating the current `withServiceTag` and adding Two methods:
- `withServiceName(String serviceTagName, String defaultServiceName)`
  - We should mention when the `defaultServiceName` is used in Javadoc
    - `@param defaultServiceName the default value of the tag, e.g. {@code "myService"}. It's only used when {@link InvocationUtil#getInvocation(RequestLog)} returns {@code null}.`
- `withServiceCalss(String serviceTagName, Class<?> serviceClass)`

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
arhont375: Is it possible to create Retrofit client with anonymous class?
May I ask to provide example?

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
arhont375: I think it's better to separate into `withServiceName(serviceName)` and `withServiceTag(serviceTagName)` methods as we may want to override service name for the case when class used. Also we may just introduce meaningful defaults to reduce code for users.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
trustin: It seems like `Retrofit.create(Class)` requires the class to be an interface. I guess it's not possible to create a Retrofit client with an anonymous class, because an anonymous class is never an interface. :sweat_smile:

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
minwoox: > we may want to override service name for the case when class used

Ah, yes I missed that point. 😄 

So, now we can set the property one by one, how about removing the prefix `with`?
Also, What do you think adding a setter for the serviceClass so that a user can do:
```
RetrofitMeterIdPrefixFunction.of(name)
                             .serviceTag(serviceTagName)
                             .serviceClass(serviceClass) // I'm not sure if this is a good name though. ;)
                             .build()
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
```diff
@@ -52,34 +53,58 @@ public static RetrofitMeterIdPrefixFunction of(String name) {
...
-            this.defaultServiceName = null;
         }
+        this.serviceName = serviceName;
+        this.defaultServiceName = defaultServiceName;
     }
 
     @Override
     public MeterIdPrefix activeRequestPrefix(MeterRegistry registry, RequestLog log) {
-        final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(3);
+        final ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(2);
```
minwoox: Thanks. 😄

arhont375:
Rebased to fix failed checks, also I moved `RetrofitMeterIdPefixFunctionTest` from `.meter` package as no other classes there including just `RetrofitMeterIdPrefixFunction` and now it's late to change as it would be breaking change

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunctionBuilder.java
```diff
@@ -33,28 +33,77 @@
...
+    }
+
+    /**
+     * Adds retrofit client service class that would be used to provide additional tags for metrics
+     * based on retrofit annotations. See {@link RetrofitClassAwareMeterIdPrefixFunction}.
+     *
+     * @param serviceClass class that defines retrofit client service.
+     */
+    public RetrofitMeterIdPrefixFunctionBuilder serviceClass(Class<?> serviceClass) {
+        this.serviceClass = serviceClass;
```
arhont375: Forgot to add here `requireNonNull` 😢

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitClassAwareMeterIdPrefixFunction.java
```diff
@@ -0,0 +1,174 @@
...
+        final String httpPath;
+        if (annotation.annotationType().equals(HTTP.class)) {
+            final HTTP http = (HTTP) annotation;
+            httpMethod = http.method();
+            httpPath = http.path();
+        } else {
+            final Method valueMethod;
+            try {
+                valueMethod = annotation.getClass().getMethod("value");
+                httpPath = (String) valueMethod.invoke(annotation);
```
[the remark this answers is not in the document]
arhont375: I was thinking about switch, but it would look ugly also increase repetition of code that will may lead to errors 🌵

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/RetrofitMeterIdPrefixFunction.java
ikhoon: It is trivial, however, if we want to optimize this:

```java
int expectedSize = serviceTagName != null ? 3 : 2;
ImmutableList.Builder<Tag> tagListBuilder = ImmutableList.builderWithExpectedSize(expectedSize);
```

minwoox: It's only two tags in this case, so I think it's `2`. 😄

ikhoon: I misread the code.

arhont375:
Rebased, fixed after merge.

trustin:
Thanks a lot for your patience, @arhont375 !

PR #2358, opened 2020-01-01
Fix a bug where causes cache misses from RouteCache

hyangtack:
Motivation:
While I was making #2102, I fixed `hashCode()` and `equals()` implementation
of `DefaultRoutingContext` class as using `query()` and `headers()` rather
than `contentType()` and `acceptTypes()`. It seemed okay because I added
the properties `params()` and `headers()` to `RoutingContext`, but it was
a mistake from the `RouteCache`'s view. It would cause cache misses from
`RouteCache` because the variation of cache key becomes wide.

Modifications:
- Use `contentType()` and `acceptTypes()` from `hashCode()` and `equals()`
  of `DefaultRoutingContext` class.

Result:
- `RouteCache` can cache routes as the same as before.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/DefaultRoutingContext.java
+++ b/core/src/main/java/com/linecorp/armeria/server/DefaultRoutingContext.java
@@ -158,9 +158,12 @@ final class DefaultRoutingContext implements RoutingContext {
     // 0 : VirtualHost
     // 1 : HttpMethod
     // 2 : Path
-    // 3 : Query
-    // 4 : Headers
-    // 5 : Boolean properties
+    // 3 : Content-Type
+    // 4 : Accept
+    //
+    // Note that we don't use query(), params() and header() for generating hashCode and comparing objects,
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/server/DefaultRoutingContext.java
```diff
@@ -158,9 +158,12 @@ public HttpStatusException deferredStatusException() {
...
     // 2 : Path
-    // 3 : Query
-    // 4 : Headers
-    // 5 : Boolean properties
+    // 3 : Content-Type
+    // 4 : Accept
+    //
+    // Note that we don't use query(), params() and header() for generating hashCode and comparing objects,
+    // because this class can be cached in RouteCache class. Using all properties may cause cache misses
+    // from RouteCache so it would be better if we choose the properties which can be a cache key.
```
ikhoon: Thanks for the education. 🙏

trustin:
Thanks, @hyangtack for a follow up.

PR #2359, opened 2020-01-02
Make sure to close `HttpResponseWriter` and release content

trustin:
Motivation:

`HttpResponseWriter.close(AggregatedHttpResponse)` and `StreamWriter.close(Object)`
raise an exception when 1) a user gave an invalid input or 2) the stream
is closed already while writing.

Such behavior can lead to a stream left open or a content left
unreleased.

Modifications:

- Deprecate `StreamWriter.close(Object)` which has an ambiguous
  contract.
- Always close the stream for `close(AggregatedHttpResponse)` and its
  cousins in `HttpResponseWriter`.
- Use `tryWrite()` and `close()` instead of `close(Object)` in our code.

Result:

- Less leak
- Less ambiguity

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
@@ -143,7 +143,10 @@ public interface StreamWriter<T> {
      */
     @Deprecated
     default void close(T obj) {
-        write(obj);
-        close();
+        try {
+            write(obj);
+        } finally {
+            close();
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/StreamWriter.java
```diff
@@ -132,8 +133,15 @@ default boolean tryWrite(Supplier<? extends T> o) {
...
+     *
+     * @throws ClosedPublisherException if the stream was already closed.
+     * @see <a href="#reference-counted">Life cycle of reference-counted objects</a>
+     *
+     * @deprecated Use {@link #tryWrite(Object)} and {@link #close()}.
      */
+    @Deprecated
     default void close(T obj) {
         write(obj);
         close();
```
minwoox: Don't we need to update this as well?
```java
    try {
        write(obj);
    } finally {
        close();
    }
```

trustin: Just wanted to preserve the behavior, but I guess you're right.

trustin:
Thanks for reviweing 😄

PR #2361, opened 2020-01-02
Add `Clients.builder()` / Deprecate methods with optional parameters

### LINKED ISSUE #1719: Using 'builder()' factory method instead of constructor

trustin:
Motivation:

- For consistency, a user should be able to create a `ClientBuilder` via
  `builder()` factory methods.
- The factory methods in `Clients` and `WebClient` are overly crowded.

Modifications:

- Add `Clients.builder()` and deprecated their corresponding
  `ClientBuilder` constructors.
- Deprecate the factory methods that require optional parameters in
  `Clients` and `WebClients`.
- Miscellaneous improvement:
  - `WebClient` now accepts a URI that starts with `none+`.

Result:

- Closes #1719
- Less cognitive load (once we remove the deprecated ones)

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/WebClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/WebClientBuilder.java
@@ -88,9 +88,11 @@ public final class WebClientBuilder extends AbstractClientOptionsBuilder<WebClie
             final String givenScheme = requireNonNull(uri, "uri").getScheme();
             final Scheme scheme = validateScheme(givenScheme);
             if (scheme.uriText().equals(givenScheme)) {
+                // No need to replace the user-specified scheme because it's already in its normalized form.
                 this.uri = uri;
             } else {
                 // Replace the user-specified scheme with the normalized one.
+                // e.g. http://foo.com/ -> none+http://foo.com/
                 this.uri = URI.create(scheme.uriText() +
                                       uri.toString().substring(givenScheme.length()));
             }
--- a/core/src/main/java/com/linecorp/armeria/client/WebClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/WebClientBuilder.java
@@ -19,11 +19,14 @@ package com.linecorp.armeria.client;
 import static java.util.Objects.requireNonNull;
 
 import java.net.URI;
+import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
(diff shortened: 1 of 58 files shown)


grpc-protocol/src/main/java/com/linecorp/armeria/common/grpc/protocol/UnaryGrpcClient.java
```diff
@@ -59,7 +59,7 @@
      * Constructs a {@link UnaryGrpcClient} for the given {@link WebClient}.
      */
     // TODO(anuraaga): We would ideally use our standard client building pattern, i.e.,
-    // new ClientBuilder(...).build(UnaryGrpcClient.class), but that requires mapping protocol schemes to media
+    // Client.builder(...).build(UnaryGrpcClient.class), but that requires mapping protocol schemes to media
```
minwoox: Clients?

core/src/main/java/com/linecorp/armeria/client/WebClientBuilder.java
```diff
@@ -77,8 +85,15 @@ static boolean isUndefinedUri(URI uri) {
...
-            validateScheme(requireNonNull(uri, "uri").getScheme());
-            this.uri = URI.create(SerializationFormat.NONE + "+" + uri);
+            final String givenScheme = requireNonNull(uri, "uri").getScheme();
+            final Scheme scheme = validateScheme(givenScheme);
+            if (scheme.uriText().equals(givenScheme)) {
+                this.uri = uri;
+            } else {
+                // Replace the user-specified scheme with the normalized one.
+                this.uri = URI.create(scheme.uriText() +
+                                      uri.toString().substring(givenScheme.length()));
```
minwoox: Could you add a small example for this case please?

trustin:
Thanks for reviewing, folks. :bowing_man:

Issue #2347, opened 2019-12-27
Automatic propagation of RequestContext


minwoox:
One of the most complaints from the users is that they are struggling to propagate a `RequestContext`. Check out this snippet:
```java
@Override
public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
    backend1.execute(req).aggregate().handle((aggregatedRes, cause) -> {
        // ctx is not in the thread-local. You should use ctx.makeContextAware() or contextAwareExecutor
        backend2.execute(req)...
    });
    ...             
    return res;
}
```

If we return a `RequestContextAwareCompletebleFuture` from `aggregate()` call, most of the situations are solved because `RequestContextAwareCompletebleFuture` keeps spwaning the future itself when adding callbacks.

Of course, we cannot propagate the `ReqeustContext` all of the cases. For example, if a user creates a new `CompletableFuture` instance and completes it with a value, the `RequestContextAwareCompletebleFuture` will stop spawning.
So he/she has to understand how the propagation works to use Armeria properly.

However, I think Automatic propagation will still reduce the burden of users who are not familiar with the asynchronous server and can lower the entry point.

minwoox:
@anuraaga I wanted to listen to your opinion before I go ahead because I feel you are reluctant Armeria to have this feature. 😄

anuraaga:
> However, I think Automatic propagation will still reduce the burden of users who are not familiar with the asynchronous server and can lower the entry point.

I think you know my opinion ;) 

If relying on context and writing an asynchronous server, knowing how context works is a requirement - I find bandaids to be worse since the context bug will inevitably appear, but later in the lifecycle of the codebase and has an even higher chance of affecting production vs development servers. Being confused in the beginning is better than later IMO.

Also, IMO writing asynchronous code effectively basically requires a framework on top of Armeria for a reasonable sized app. I personally know Dagger makes it trivial to solve the context problem completely where it would be harder for a user to make a mistake than do the right thing. I have heard of rxjava users forgetting to subscribe correctly - perhaps enforcing use of something like `ArmeriaFlux` that automatically subscribes would be harder to forget. I was a little interested in seeing what approaches could help with RxJava but currently have no plans of using it so not sure if I could come up with anything but it seems like a better layer to deal with asynchronous issues than our custom `CompletableFuture` implementation.

[...]

PR #2363, opened 2020-01-02
Add examples for context propagation

anuraaga:
…ger and rxjava.

I don't know if these examples are relevant enough to have in the main repo, but wanted to give examples of what I think regards to showing good patterns for context propagation #2347. We don't have to merge this if it doesn't seem useful just wanted to throw it out there.

The same pattern here can be applied regardless of if there are other libraries like MySQL / Redis involved and hopefully can be helpful for understanding context in complex server codebases.

I'm hoping someone familiar with RxJava and Kotlin can add similar examples showing best practices for them :)


trustin:
How about making the examples more interesting? `contextpropagation` sounds too honest and dry :laughing: 

I used to use recursive Fibonacci server to show how to talk to multiple backends, but I'm not 100% sure if it's suitable here.. How about defining some simple yet familiar problem and solving it with async calls?

minwoox:
> I'm hoping someone familiar with RxJava and Kotlin can add similar examples showing best practices for them :)

Isn't it just calling `RequestContextAssembly.enable();` for RxJava? 😆 
> How about defining some simple yet familiar problem and solving it with async calls?

That sounds more fun. 😆 I'm a little bit worried about though because that could harm how simple it is to propagate the context which this example shows.

anuraaga:
> Isn't it just calling RequestContextAssembly.enable(); for RxJava? 😆

That sounds pretty good - I wonder why we get so many reports of trouble with context propagation then though. I remember @kojilin mentioning that it's still possible to forget to `subscribeOn` the context executor, but maybe I misunderstood.

> How about making the examples more interesting?

Probably still too dry ;) But switched to computing squares of input numbers. It might have been nice though, I found that even for such a simple example, automatic context propagation wouldn't work since I have to use `CompletableFuture.allOf`.

minwoox:
> I wonder why we get so many reports of trouble with context propagation then though.

It's because the number of users is increasing? 😆 It's the case that using `webClient.execute()` and all from different users.

examples/context-propagation/manual/src/main/java/example/armeria/contextpropagation/manual/MainService.java
```diff
@@ -0,0 +1,109 @@
...
+                // use ctxExecutor for them. But if in doubt, it doesn't hurt too much to use it everywhere.
+                fetchFromBackend.thenApply(CompletableFutures::allAsList)
+                                .thenCompose(u -> u)
+                                .thenApplyAsync(
+                                        (backendResponse) -> {
+                                            // The context is mounted in a thread-local, meaning it is
+                                            // available to all logic such as tracing.
+                                            checkState(ServiceRequestContext.current() == ctx);
+                                            return HttpResponse.of(backendResponse.stream()
+                                                                   .map(AggregatedHttpResponse::contentUtf8)
```
minwoox: nit: indentation.

anuraaga:
Oh yeah - done

examples/README.md
```diff
@@ -4,6 +4,12 @@
   - Learn how to write an HTTP service using annotations.
   - See [Annotated services](https://line.github.io/armeria/server-annotated-service.html).
 
+- `context-propagation`
+  - Learn how to propagate Armeria's `RequestContext` for use in scenarios like tracing.
+  - [`dagger`](https://dagger.dev) provides an example using the Dagger asynchronous framework for
```
trustin: Better hyperlinking for `the Dagger asynchronous framework`?

anuraaga: Yup

examples/context-propagation/rxjava/src/main/java/example/armeria/contextpropagation/rxjava/MainService.java
```diff
@@ -0,0 +1,88 @@
...
+                .mergeWith(FutureConverter.toSingle(fetchFromFakeDb)
+                                          .flatMapObservable(Observable::fromIterable))
+                .flatMapSingle(num -> {
+                    // The context is mounted in a thread-local, meaning it is available to all logic
+                    // such as tracing.
+                    checkState(ServiceRequestContext.current() == ctx);
+
+                    return FutureConverter.toSingle(backendClient.get("/square/" + num).aggregate());
+                })
+                .map(AggregatedHttpResponse::contentUtf8)
```
anuraaga: One aspect of rxjava integration I noticed is methods like these are fairly ridiculously slow due to context mounting. An inevitable downside of having the integration be a single static method invocation in `main` I guess.

anuraaga:
I added assertions for the current thread to dagger and manual too. I realized that one disadvantage of the rxjava assembly is it doesn't control the callback thread so callbacks will be on a random thread - this isn't terrible but means user code needs to be thread-safe and I find it tends to be a better pattern to use a single thread for all callbacks.

In https://github.com/line/armeria/pull/2363/commits/10c61933509bf0397e60098d026c0df3f4621076 I tried switching from the assembly to just use `subscribeOn`, this does make sure the thread is correct. It is of course much easier to accidentally forget context here so this might be worse. Let me know which pattern we should recommend. @kojilin especially as I guess you use these APIs and might know some practices.

kojilin:
Currently, my preferred pattern is like above, do `observeOn` before next async call to avoid putting at callee side.

```
    @Override
    public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) {
        final Scheduler contextAwareScheduler = Schedulers.from(ctx.contextAwareExecutor());
        // This logic mimics using a blocking method, which would usually be something like a MySQL
        // database query using JDBC.
        final Flowable<Long> fetchFromFakeDb =
                Single.fromCallable(
                        () -> {
                            // The context is mounted in a thread-local, meaning it is available to all
                            // logic such as tracing.
                            checkState(ServiceRequestContext.current() == ctx);
                            checkState(!ctx.eventLoop().inEventLoop());

                            Uninterruptibles.sleepUninterruptibly(Duration.ofMillis(50));
                            return ImmutableList.of(23L, -23L);
                        })
                      // Always run blocking logic on the blocking task executor. By using
                      // ServiceRequestContext.blockingTaskExecutor, you also ensure the context is mounted
                      // inside the logic (e.g., your DB call will be traced!).
                      .subscribeOn(Schedulers.from(ctx.blockingTaskExecutor()))
                      .flattenAsFlowable(l -> l);

        Flowable<Long> fetchFromBackend =
                FutureConverter.toSingle(req.aggregate())
                               .flatMapPublisher(request -> {
                                   // The context is mounted in a thread-local, meaning it is available to all logic
                                   // such as tracing.
                                   checkState(ServiceRequestContext.current() == ctx);
                                   checkState(ctx.eventLoop().inEventLoop());

                                   final List<Long> nums = new ArrayList<>();
                                   for (String token : Iterables.concat(
                                           NUM_SPLITTER.split(request.path().substring(1)),
                                           NUM_SPLITTER.split(request.contentUtf8()))) {
                                       nums.add(Long.parseLong(token));
                                   }
                                   return Flowable.fromIterable(nums);
                               });

        final Single<HttpResponse> response =
                Flowable.concatArrayEager(fetchFromBackend, fetchFromFakeDb)
                        .observeOn(contextAwareScheduler)
                        .flatMapSingle(num -> {
                            // The context is mounted in a thread-local, meaning it is available to all logic
                            // such as tracing.
                            checkState(ServiceRequestContext.current() == ctx);
                            checkState(ctx.eventLoop().inEventLoop());
                            return FutureConverter.toSingle(backendClient.get("/square/" + num).aggregate());
                        })
                        .map(AggregatedHttpResponse::contentUtf8)
                        .collectInto(new StringBuilder(),
                                     (current, item) -> current.append(item)
                                                               .append('\n'))
                        .map(content -> HttpResponse.of(content.toString()))
                        .onErrorReturn(HttpResponse::ofFailure)
                        // Unless you know what you're doing, always use subscribeOn with the context executor to have
                        // the context mounted and stay on a single thread to reduce concurrency issues.
                        .subscribeOn(contextAwareScheduler);

        return HttpResponse.from(FutureConverter.toCompletableFuture(response));
    }
```

examples/context-propagation/rxjava/src/main/java/example/armeria/contextpropagation/rxjava/MainService.java
```diff
@@ -0,0 +1,100 @@
...
+                    final List<Long> nums = new ArrayList<>();
+                    for (String token : Iterables.concat(
+                            NUM_SPLITTER.split(request.path().substring(1)),
+                            NUM_SPLITTER.split(request.contentUtf8()))) {
+                        nums.add(Long.parseLong(token));
+                    }
+
+                    return Observable.fromIterable(nums);
+                })
+                .mergeWith(fetchFromFakeDb)
```
kojilin: Maybe `Flowable.concatEager` is more suitable comapring to other example.

examples/context-propagation/rxjava/src/main/java/example/armeria/contextpropagation/rxjava/MainService.java
```diff
@@ -0,0 +1,100 @@
...
+
+                            Uninterruptibles.sleepUninterruptibly(Duration.ofMillis(50));
+                            return ImmutableList.of(23L, -23L);
+                        })
+                      // Always run blocking logic on the blocking task executor. By using
+                      // ServiceRequestContext.blockingTaskExecutor, you also ensure the context is mounted
+                      // inside the logic (e.g., your DB call will be traced!).
+                      .subscribeOn(Schedulers.from(ctx.blockingTaskExecutor()))
+                      // Make sure callbacks still stay on the context executor using observeOn.
+                      .observeOn(contextAwareScheduler)
```
kojilin: Maybe put at chaining phase?

anuraaga: Can you provide your reasoning for that? I can add as a comment and update :) I figured it's better to be here since the reason we need to `observeOn` is because of the blocking subscription, so having the two close to each other seems more readable.

kojilin: I usually let storage call/remote call layer does not include subscribeOn&observeOn, I think they usually don't need to care about the `trigger` thread or ctx and we can reuse it in non armeria-context too.
So in my situation we force write all `observeOn` at service/facade layer and check when we code do code review....

examples/context-propagation/rxjava/src/main/java/example/armeria/contextpropagation/rxjava/MainService.java
anuraaga: By the way, I was able to confirm that the pattern with `subscribeOn` instead of using `RequestContextAssembly` does not have the performance issue. The key point I guess is `RequestContextSubscriber` mounts on every call to `onNext`

https://github.com/line/armeria/blob/master/rxjava/src/main/java/com/linecorp/armeria/common/rxjava/RequestContextSubscriber.java#L38

which ends up being called from every single callback in the flow.

Subscribing on the executor though mounts the context when the thread starts to process the flow, which ends up processing as many items as it can at the time. So if the buffer has 10 items, it will go through all 10 in a single invocation of the executor, meaning just a single threadlocal push/pop for 10 invocations of `onNext`.

Just an observation, I think the assembly is still useful since it is quite nice for a single method call to be able to guarantee callbacks have context, but we do have to be aware of the cost of the convenience.

trustin:
@anuraaga @kojilin Thanks for your collaboration. Please let me know if this pull request is ready for merge. (No rush)

trustin:
@kojilin Gentle ping

kojilin:
LGTM, maybe due to build failed, I can't approve.

trustin:
Thanks @anuraaga and reviewers.

adriancole:
this is really teaching. nice work

PR #2365, opened 2020-01-03
Reorganize the classes in `grpc-protocol`

trustin:
Motivation:

Realized that both client and server classes in `grpc-protocol` are
placed under `common.grpc.protocol`.

Modifications:

- Move the client and server side classes to `client.grpc.protocol` and
  `server.grpc.protocol` respectively.
- Miscellaneous:
  - Add @anuraaga to the owners of `/grpc-protocol/`

Result:

Consistency

### DIFF
--- a/grpc-protocol/src/main/java/com/linecorp/armeria/client/grpc/protocol/package-info.java
+++ b/grpc-protocol/src/main/java/com/linecorp/armeria/client/grpc/protocol/package-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 LINE Corporation
+ * Copyright 2020 LINE Corporation
  *
  * LINE Corporation licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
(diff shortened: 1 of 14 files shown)


grpc-protocol/src/main/java/com/linecorp/armeria/client/grpc/protocol/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+/**
+ * Unary gRPC client without support for gRPC generated code stubs. This package is separated for advanced
+ * users that would like to use the gRPC wire protocol without depending on gRPC itself.
+ * This package must not depend on any dependencies outside of {@code armeria-core}.
+ *
+ * <p>Don't use this package unless you know what you're doing, it is generally recommended to use a normal
+ * gRPC client provided by the {@code armeria-grpc} module.
```
anuraaga: While we're here, I think we can add that we don't guarantee backwards compatibility for this package either since it's an advanced API if we don't mind having such packages in Armeria.

trustin: Added warnings :smile:

trustin:
Thanks for reviewing :-)

PR #2366, opened 2020-01-03
Warn when pushing a wrong context using ContextFuture

### LINKED ISSUE #2173: `RequestContextAwareCompletableFuture` callback not invoked when context mismatches.

minwoox:
Motivation:
Related #2173

In that case, which is trying to push a wrong context when the current thread had an unrelated context already, the `RequestContextAwareCompletableFuture` just swallows the exception and a user has no way what is going wrong.

Modification:
- Leave a warning log every time a wrong context is pushed to let the user something is going wrong.

Result:
- You can now easily get noticed when pushing the wrong context.
- Close #2173


core/src/test/java/com/linecorp/armeria/internal/ContextFutureCallbackArgumentsProvider.java
```diff
@@ -0,0 +1,263 @@
...
+
+    @Override
+    public Stream<? extends Arguments> provideArguments(ExtensionContext context) throws Exception {
+        final Arguments thenApply = Arguments.of(
+                (BiConsumer<CompletableFuture<?>, AtomicBoolean>) (future, called) -> {
+                    future.thenApply(res -> {
+                        called.set(true);
+                        return null;
+                    }).exceptionally(cause -> {
+                        called.set(true);
```
minwoox: This is not called because in `exceptionally`, pushing with a wrong context happens as well.
I'm wondering we should mitigate this by not pushing the context in `exceptionally`. 🤔

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -545,7 +545,10 @@ public String toString() {
             return strVal;
         }
 
-        final StringBuilder buf = new StringBuilder(96);
+        final StringBuilder buf = new StringBuilder(117);
+        buf.append('[')
+           .append(ClientRequestContext.class.getSimpleName())
+           .append(']');
```
trustin: Could we make it much shorter? It will increase the amount of our logs by large margin.

core/src/main/java/com/linecorp/armeria/internal/AbstractRequestContextAwareCompletableFuture.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    protected Runnable makeContextAwareLoggingException(Runnable runnable) {
+        requireNonNull(runnable, "runnable");
+        return () -> {
+            try (SafeCloseable ignored = ctx.push()) {
+                runnable.run();
+            } catch (Throwable th) {
+                logger.warn("An error occurred when pushing a context", th);
+                throw th;
```
trustin: Could extract into a new method and deduplicate

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -584,7 +584,10 @@ public String toString() {
             return strVal;
         }
 
-        final StringBuilder buf = new StringBuilder(96);
+        final StringBuilder buf = new StringBuilder(119);
+        buf.append('[')
+           .append(ServiceRequestContext.class.getSimpleName())
+           .append(']');
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
minwoox: Then how about `[ClientCtx]`? or `[Client]`? I really wish to know whether the context is `ClientRequestContext` or `ServiceRequestContext` from the logs. 🤔

core/src/test/java/com/linecorp/armeria/internal/ContextFutureCallbackArgumentsProvider.java
trustin: That's a good question. :thinking: 

It is also possible that a user tries to handle an exception using `handle` or `whenComplete`, so I think it's not a good idea to handle `exceptionally` exceptionally. (pun intended :laughing:) i.e. I can't think of a good solution other than just logging.

minwoox: 😆

core/src/main/java/com/linecorp/armeria/internal/AbstractRequestContextAwareCompletableFuture.java
minwoox: Fixed. 😉

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
minwoox: Changed to `[ClientCtx]` and `[ServiceCtx]`

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -351,10 +352,7 @@ default SafeCloseable push(boolean runCallbacks) {
 
         // Put the oldCtx back before throwing an exception.
         RequestContextThreadLocal.set(oldCtx);
-        throw new IllegalStateException(
-                "Trying to call object wrapped with context " + this + ", but context is currently " +
-                "set to " + oldCtx + ". This means the callback was called from " +
-                "unexpected thread or forgetting to close previous context.");
+        return throwIllegalContextPushing(this, oldCtx);
```
anuraaga: Ah here's where I first saw the pattern - see my comment here https://github.com/line/armeria/pull/2371#discussion_r363069843

Using `return` instead of `throw` doesn't seem to improve verbosity but makes the code much less readable.

minwoox: Yes, it's now much readable. Thanks!

core/src/main/java/com/linecorp/armeria/internal/AbstractRequestContextAwareCompletableFuture.java
```diff
@@ -0,0 +1,102 @@
...
+    protected <I, U, V> BiFunction<I, U, V> makeContextAwareLoggingException(BiFunction<I, U, V> function) {
+        requireNonNull(function, "function");
+        return (t, u) -> makeContextAwareLoggingException0(() -> function.apply(t, u));
+    }
+
+    private void makeContextAwareLoggingException0(Runnable runnable) {
+        try (SafeCloseable ignored = ctx.push()) {
+            runnable.run();
+        } catch (Throwable th) {
+            logger.warn("An error occurred when pushing a context", th);
```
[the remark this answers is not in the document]
minwoox: That's a good point. Let me update the message. 😄

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: How about shortening even more, like `[C]` and `[S]`?

minwoox: It's fine for me. Let me change that. 😄

core/src/main/java/com/linecorp/armeria/internal/AbstractRequestContextAwareCompletableFuture.java
```diff
@@ -67,35 +67,35 @@ protected Runnable makeContextAwareLoggingException(Runnable runnable) {
...
-    private void makeContextAwareLoggingException0(Runnable runnable) {
+    private void makeContextAwareLoggingException0(Runnable action) {
         try (SafeCloseable ignored = ctx.push()) {
-            runnable.run();
+            action.run();
         } catch (Throwable th) {
-            logger.warn("An error occurred when pushing a context", th);
+            logger.warn("An error occurred when pushing a context and executing the action", th);
             throw th;
         }
```
trustin: We should not log exceptions raised by `.run()`

```java
final SafeCloseable handle;
try {
    handle = ctx.push();
} catch (Throwable t) {
    logger.warn(...);
    Exceptions.throwUnsafely(t);
    return;
}

try {
    action.run();
} finally {
    handle.close();
}
```

core/src/main/java/com/linecorp/armeria/internal/AbstractRequestContextAwareCompletableFuture.java
```diff
@@ -67,35 +67,35 @@ protected Runnable makeContextAwareLoggingException(Runnable runnable) {
...
-    private <V> V makeContextAwareLoggingException0(Supplier<? extends V> action) {
+    private <V> V makeContextAwareLoggingException0(Supplier<? extends V> fn) {
         try (SafeCloseable ignored = ctx.push()) {
-            return action.get();
+            return fn.get();
         } catch (Throwable th) {
-            logger.warn("An error occurred when pushing a context", th);
+            logger.warn("An error occurred when pushing a context and executing the fn", th);
             throw th;
         }
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
```diff
@@ -87,5 +87,20 @@ public static SafeCloseable pushWithRootAndOldCtx(ClientRequestContext currentCt
         }
     }
 
+    /**
+     * An {@link IllegalStateException} which is raised when pushing a context from the unexpected thread
+     * or forgetting to close the previous context.
+     */
+    public static final class IllegalContextPushingException extends IllegalStateException {
```
trustin: Could you make it a top-level class and make it part of `common` package?

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
```diff
@@ -87,5 +87,20 @@ public static SafeCloseable pushWithRootAndOldCtx(ClientRequestContext currentCt
...
+
+        private static final long serialVersionUID = 5431942355463120798L;
+
+        public IllegalContextPushingException(RequestContext newCtx, RequestContext oldCtx) {
+            super("Trying to call object wrapped with context " + newCtx + ", but context is currently " +
+                  "set to " + oldCtx + ". This means the callback was called from " +
+                  "unexpected thread or forgetting to close previous context.");
```
trustin: Now that we have a dedicated exception type, we could add getters for the two contexts? We could also simplify the exception message and add more details to the class Javadoc.

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
```diff
@@ -87,5 +87,20 @@ public static SafeCloseable pushWithRootAndOldCtx(ClientRequestContext currentCt
...
 
+    /**
+     * An {@link IllegalStateException} which is raised when pushing a context from the unexpected thread
+     * or forgetting to close the previous context.
+     */
+    public static final class IllegalContextPushingException extends IllegalStateException {
+
+        private static final long serialVersionUID = 5431942355463120798L;
+
+        public IllegalContextPushingException(RequestContext newCtx, RequestContext oldCtx) {
```
anuraaga: This type looks good but just to clarify, I was just expecting the old factory method to return `IllegalStateException` instead of throwing it ;)

minwoox: Ah, I misunderstood. 😅

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
minwoox: So If I change just to return the `IllegalStateException` from the previous method, I don't have to declare this new exception type. And I think we don't have to have this another exception, especially for this case. (Sorry I made this class because I misunderstood the comment. 🙏) Do you think we still need this?

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
```diff
@@ -87,5 +87,19 @@ public static SafeCloseable pushWithRootAndOldCtx(ClientRequestContext currentCt
...
 
+    /**
+     * Returns an {@link IllegalStateException} which is raised when pushing a context from
+     * the unexpected thread or forgetting to close the previous context.
+     */
+    public static IllegalStateException newIllegalContextPushingException(
+            RequestContext newCtx, RequestContext oldCtx) {
+        requireNonNull(newCtx, "newCtx");
+        requireNonNull(oldCtx, "oldCtx");
+        throw new IllegalStateException(
```
anuraaga:
```suggestion
        return new IllegalStateException(
```

minwoox: Oops forgot to change that.

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
trustin: Nope. Let's continue using `IllegalStateException`.

minwoox:
Thanks reviewers!

PR #2367, opened 2020-01-03
Follow Micrometer's naming convention

trustin:
Co-authored-by: Per Lundberg <perlun@gmail.com>

Motivation:

The Micrometer naming conventions (http://micrometer.io/docs/concepts#_naming_meters)
suggest that dotted notation (`foo.bar.baz`) should be used when naming
meters and tags. This has the advantage of letting different Micrometer
monitoring implementations be able to convert it to the naming
convention which is native for it (e.g. JMX, Prometheus etc).

Modifications:

- Switch to Micrometer's naming convention everywhere.
- Add a new flag `com.linecorp.armeria.useLegacyMeterNames` which
  switches back to the old meter names.
- Deprecate `MoreNamingConventions`
  - The `NamingConvention`s returned by `MoreNamingConventions` now return
    Micrometer's default `NamingConvention` for corresponding backend.
  - `MoreNamingConventions.configure()` is now no-op.
- Miscellaneous:
  - `{Dropwizard,Prometheus}MeterRegistries.configureRegistry()` does
    not override the pause detector anymore because Micrometer now uses
    `NoPauseDetector` by default.

Result:

- Armeria is now a good citizen of Micrometer ecosystem.
- Users have time to migrate to the new naming thanks to the new flag.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistries.java
+++ b/core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistries.java
@@ -24,6 +24,8 @@ import javax.annotation.Nullable;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
 
+import com.linecorp.armeria.common.Flags;
+
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.Tag;
@@ -159,8 +161,8 @@ public final class DropwizardMeterRegistries {
                 return MeterFilterReply.NEUTRAL;
             }
         });
+
         meterRegistry.config().namingConvention(MoreNamingConventions.dropwizard());
-        meterRegistry.config().pauseDetector(new NoPauseDetector());
         return meterRegistry;
     }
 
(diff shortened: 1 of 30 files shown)


trustin:
@perlun @taicki @julnamoo @kshgizmo @renaudb PTAL as well.

trustin:
@huydx @linxGnu You might wanna take a look as well.

linxGnu:
LGTM

core/src/main/java/com/linecorp/armeria/common/metric/DropwizardMeterRegistries.java
```diff
@@ -159,8 +158,8 @@ public MeterFilterReply accept(Meter.Id id) {
                 return MeterFilterReply.NEUTRAL;
             }
         });
+
         meterRegistry.config().namingConvention(MoreNamingConventions.dropwizard());
-        meterRegistry.config().pauseDetector(new NoPauseDetector());
```
ikhoon: 👍/cc @minwoox FYI https://github.com/micrometer-metrics/micrometer-docs/commit/53698ab25fda071676d45dd53a3e2877b006b35d

trustin:
Thanks for reviewing!

PR #2369, opened 2020-01-04
Support reactor-grpc for client-side

trustin:
- Update `HelloServiceTest` in `:examples:grpc-reactor` to use
  reactor-grpc for the client side.
- Fix a bug where `ArmeriaClientCall` does not call `onReady()`, which made a `Flux` request not sent at all.
  - Fix provided by @cbornet

### DIFF
--- a/examples/grpc-reactor/src/test/java/example/armeria/grpc/reactor/HelloServiceTest.java
+++ b/examples/grpc-reactor/src/test/java/example/armeria/grpc/reactor/HelloServiceTest.java
@@ -1,40 +1,37 @@
 package example.armeria.grpc.reactor;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.awaitility.Awaitility.await;
 
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.platform.commons.logging.LoggerFactory;
 
 import com.google.common.base.Stopwatch;
-import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.MoreExecutors;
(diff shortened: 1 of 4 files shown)


heowc:
This PR will help me write grpc-kotlin. Looking forward to :)

trustin:
Let me separate this into two PRs - 1) generalization of stub factory detection 2) client-side reactor-grpc

trustin:
First part: https://github.com/line/armeria/pull/2376 I'll update this pull request once it's merged.

examples/grpc-reactor/src/test/java/example/armeria/grpc/reactor/HelloServiceTest.java
```diff
@@ -50,179 +47,70 @@ private static String uri() {
...
-                    }
-                });
-
-        for (String name : names) {
-            request.onNext(HelloRequest.newBuilder().setName(name).build());
-        }
-        request.onCompleted();
-        await().untilTrue(completed);
+        final Flux<HelloRequest> request = Flux.just(names).log()
+                                               .map(name -> HelloRequest.newBuilder().setName(name).build());
```
cbornet: You could write
```java
final String message = Flux.just(names).log()
                             .map(name -> HelloRequest.newBuilder().setName(name).build())
                             .as(helloService::lotsOfGreetings)
                             .map(HelloReply::getMessage)
                             .block();
```

cbornet:
The code seems OK. I'm not sure why it blocks. Maybe something to do with the executors.

cbornet:
I tested with a regular gRPC server + managed channel and it worked fine. So it surely has something to do with how Armeria handles things.

cbornet:
Also an Armeria server with a regular managed channel works fine. So it seems to come from the Armeria client.

trustin:
Yeah, that's my finding so far as well. One interesting thing is that gRPC doesn't call `ArmeriaClientCall.sendMessage()` (which is an override of `ClientCall.sendMessage()`). As a result, nothing is being sent.

cbornet:
One of the differences I see is that `grpc-netty` sends a `listener.onReady()` during its init sequence (in `NettyClientStream::setHttp2Stream`).
So I added `responseListener.onReady();` at the end of `ArmeriaClientCall::start` and now the tests are passing.
I don't know if it's the appropriate place to put this. Is the listener completely ready at this stage ?

anuraaga:
Thanks for the debugging! I suspect it's fine - most of the client-side lifecycle matching had been trial and error since unlike the server, we aren't responsible for it and need to understand it through practice.

Do you mind sending a PR for it?

cbornet:
I did a PR on trustin's repo : https://github.com/trustin/armeria/pull/1

trustin:
Thanks all for reviewing.

PR #2370, opened 2020-01-04
Add examples:grpc-kotlin

heowc:
Related: #1078, #1880

Motivation:

- An example is provided to users of Kotlin.

Modifications:

- Make an example in Kotlin that looks like `grpc-example`.

Result:

- Partial fix for #1880
- User-friendliness


minwoox:
Are we going to update this following #2369? 🤔

heowc:
> Are we going to update this following #2369? 🤔

I have no direct connection with #2369. I was interested in internal logic changes. 
(I will edit the comment.)

Or do you need to make changes as simple as the test case for the issue?

minwoox:
Ah forget about it. I was confused. 😄

examples/grpc-kotlin/src/main/kotlin/example/armeria/grpc/kotlin/HelloExtensions.kt
```diff
@@ -0,0 +1,10 @@
+package example.armeria.grpc.kotlin
+
+import example.armeria.grpc.kotlin.Hello.HelloReply
+import example.armeria.grpc.kotlin.Hello.HelloRequest
+
+fun String.buildHelloReply(): HelloReply = HelloReply.newBuilder().setMessage(this).build()
+
+fun String.buildHelloRequest(): HelloRequest = HelloRequest.newBuilder().setName(this).build()
+
+fun String.hello(): String = "Hello, $this!"
```
ikhoon: This is not related to the Armeria core feature. However, I think string extension methods are not a good example to build a Message because `String` is a common class that represents char sequences. It looks fancy but I feel it is a bit abused.

How about remove the string extension methods and add a class like `HelloMessageExtensions` or `HelloMessageFatory`?

```kt
class HeloMessageFatory {
  fun newHelloReply(message: String): HelloReply = ...
  fun newHelloRequest(message: String): HelloRequest = ...
  ...
}
```

heowc: I agree. Usually in Kotlin, adding `Extension` to the end of a class name has a different purpose, so we'll change it to `HelloMessageFatory`.

examples/grpc-kotlin/src/main/kotlin/example/armeria/grpc/kotlin/HelloMessageFactory.kt
```diff
@@ -0,0 +1,14 @@
+package example.armeria.grpc.kotlin
+
+import example.armeria.grpc.kotlin.Hello.HelloReply
+import example.armeria.grpc.kotlin.Hello.HelloRequest
+
+class HelloMessageFactory {
```
anuraaga: How about using private methods like the other examples?

https://github.com/line/armeria/blob/master/examples/grpc-reactor/src/main/java/example/armeria/grpc/reactor/HelloServiceImpl.java#L104

Unless that's a total antipattern in Kotlin, I think it's better for the examples to be as close to each other as possible for comparison

ikhoon: > I think it's better for the examples to be as close to each other as possible for comparison

That makes sense. 
@heowc Sorry for the interrupting multiple times. 😅 I think @anuraaga ‘s suggestion is better than mine.

heowc: Fine. 😄I will reflect on the code soon.

heowc:
It's all thanks to you. 😄

trustin:
Thanks, @heowc !

PR #2371, opened 2020-01-04
Replace `Optional` with a nullable value

### LINKED ISSUE #2362: Reconsider the usage of `Optional<T>` in our API
The early API such as `Scheme.tryParse()` and `Server.activePort()` returns an `Optional` while more recent APIs just return `null`.

Given that `Optional` tends to make user code verbose and JDK will nor make `Optional` a value type due to backward compatibility (which is a shame..), I think it's better just returning a nullable value.

This guarantees breaking changes, but migration should be as simple as wrapping with `Optional.ofNullable()`, so ... :sweat_smile:

trustin:
Motivation:

The early API such as `Scheme.tryParse()` and `Server.activePort()`
returns an `Optional` while more recent APIs just return `null`.

Given that `Optional` tends to make user code verbose and JDK will not
make `Optional` a value type due to backward compatibility, I think it's
better just using a nullable value in our API.

Modifications:

- (Breaking) Replace all `Optional` usages with `@Nullable`.
- Add `MediaType.charset(Charset defaultCharset)` because
  `MediaType.charset().orElse(UTF_8)` seems pretty common.
- (Breaking) `ClientOptions.get()` now throws a `NoSuchElementException`
  when there's no option value set.
  - Added `ClientOptions.getOrNull()` as an alternative.
- Add `ClientFactoryOptions.get()`, `getOrNull()` and `getOrElse()`,
  which seems like we forgot to add before.

Result:

- Fixes #2362
- (Breaking) Our public API does not expose `Optional` anymore.
- `ClientOptions.getOrNull()`
- `ClientFactoryOptions.get()`, `getOrNull()` and `getOrElse()`


core/src/main/java/com/linecorp/armeria/client/Client.java
```diff
@@ -59,22 +57,22 @@
...
      *
      * // You can also use Clients.unwrap(), which is useful especially for
      * // Thrift and gRPC where the client object does not implement the 'as()' method.
-     * LoggingClient unwrapped2 = Clients.unwrap(client, LoggingClient.class).get();
+     * LoggingClient unwrapped2 = Clients.unwrap(client, LoggingClient.class);
      * }</pre>
      *
      * @param type the type of the object to return
-     * @return the object of the specified {@code type} if found. {@link Optional#empty()} if not found.
+     * @return the object of the specified {@code type} if found. {@code null} if not found.
```
anuraaga:
```suggestion
     * @return the object of the specified {@code type} if found, or {@code null} if not found.
```

While we're here

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -254,26 +254,27 @@ static void disableShutdownHook() {
...
      *
      * // If the client implements Unwrappable, you can just use the 'as()' method.
-     * LoggingClient unwrapped2 = client.as(LoggingClient.class).get();
+     * LoggingClient unwrapped2 = client.as(LoggingClient.class);
      * }</pre>
      *
      * @param client the client object
      * @param type the type of the object to return
-     * @return the object of the specified {@code type} if found. {@link Optional#empty()} if not found.
+     * @return the object of the specified {@code type} if found. {@code null} if not found.
```
anuraaga:
```suggestion
     * @return the object of the specified {@code type} if found, or {@code null} if not found.
```

core/src/main/java/com/linecorp/armeria/client/Clients.java
```diff
@@ -587,26 +588,27 @@ private static ClientBuilderParams builderParams(Object client) {
...
+     * LoggingClient unwrapped = Clients.unwrap(client, LoggingClient.class);
      *
      * // If the client implements Unwrappable, you can just use the 'as()' method.
-     * LoggingClient unwrapped2 = client.as(LoggingClient.class).get();
+     * LoggingClient unwrapped2 = client.as(LoggingClient.class);
      * }</pre>
      *
      * @param type the type of the object to return
-     * @return the object of the specified {@code type} if found. {@link Optional#empty()} if not found.
+     * @return the object of the specified {@code type} if found. {@code null} if not found.
```
anuraaga:
```suggestion
     * @return the object of the specified {@code type} if found, or {@code null} if not found.
```

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
```diff
@@ -149,6 +139,27 @@ protected AbstractRequestContextBuilder(boolean server, RpcRequest rpcReq, URI u
         query = pathAndQuery.query();
     }
 
+    private static SessionProtocol getSessionProtocol(URI uri) {
+        final String schemeStr = uri.getScheme();
+        if (schemeStr != null && schemeStr.indexOf('+') < 0) {
+            final SessionProtocol parsed = SessionProtocol.find(schemeStr);
+            if (parsed == null) {
+                return reportInvalidScheme(uri);
```
anuraaga: I think I saw this pattern in another PR and found it a bit weird, and now am reconsidering it again. Is it clearer to use `throw newInvalidSchemeException(uri)` so it's easy to see that this line is throwing an exception, without much change in verbosity (actually guess less characters)?

core/src/main/java/com/linecorp/armeria/server/Service.java
```diff
@@ -69,12 +68,12 @@ default void serviceAdded(ServiceConfig cfg) throws Exception {}
      * }</pre>
      *
      * @param type the type of the object to return
-     * @return the object of the specified {@code type} if found. {@link Optional#empty()} if not found.
+     * @return the object of the specified {@code type} if found. {@code null} if not found.
```
anuraaga:
```suggestion
     * @return the object of the specified {@code type} if found, or {@code null} if not found.
```

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -171,11 +170,12 @@ public String defaultHostname() {
      * Returns the primary {@link ServerPort} that this {@link Server} is listening to. This method is useful
      * when a {@link Server} listens to only one {@link ServerPort}.
      *
-     * @return {@link Optional#empty()} if this {@link Server} did not start
+     * @return {@code null} if this {@link Server} did not start
```
anuraaga: Feels a bit odd to have two `Return` / `@return` in a single javadoc.

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/TomcatServiceConfig.java
```diff
@@ -106,10 +106,11 @@ Path docBase() {
     /**
      * Returns the path to the root directory of a web application inside a JAR/WAR.
      *
-     * @return {@link Optional#empty()} if {@link #docBase()} is not a JAR/WAR file
+     * @return {@code null} if {@link #docBase()} is not a JAR/WAR file
```
anuraaga: Here especially, I'd expect just `, or {@code null}`... inlined with the first line.

thrift/src/test/java/com/linecorp/armeria/client/thrift/ThriftOverHttpClientTServletIntegrationTest.java
```diff
@@ -191,20 +189,20 @@ public static void destroyServer() throws Exception {
...
     @Test
-    public void sendHelloViaHttp1() throws Exception {
+    void sendHelloViaHttp1() throws Exception {
         final AtomicReference<SessionProtocol> sessionProtocol = new AtomicReference<>();
         final HelloService.Iface client = newSchemeCapturingClient(http1uri(HTTP), sessionProtocol);
 
         for (int i = 0; i <= MAX_RETRIES; i++) {
             try {
-                assertEquals("Hello, old world!", client.hello("old world"));
+                Assertions.assertEquals("Hello, old world!", client.hello("old world"));
```
anuraaga: I think IntelliJ refactor might have un-static imported the assertions. Though is migrating to junit5 also the time to migrate to assertj? ;)

trustin: Fixed :-)

core/src/main/java/com/linecorp/armeria/server/Server.java
trustin: Seems to me that the Javadoc body gets too long if I move what's written in `@return`..

core/src/main/java/com/linecorp/armeria/common/AbstractRequestContextBuilder.java
trustin: Fixed :-)

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/EventCounter.java
```diff
@@ -28,16 +28,16 @@
     /**
      * Counts success events.
      *
-     * @return An {@link Optional} containing the current {@link EventCount} if it has been updated,
-     *         or else an empty {@link Optional}.
+     * @return the current {@link EventCount} if it has been updated, or else {@code null}.
```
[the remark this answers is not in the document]
trustin: Fixed

core/src/main/java/com/linecorp/armeria/common/MediaType.java
```diff
@@ -648,31 +667,49 @@ public String subtype() {
...
+
+    /**
+     * Returns a {@link Charset} for the value of the charset parameter if it is specified.
+     *
+     * @return the {@link Charset}, or {@code defaultCharset} if the charset parameter is not specified.
+     *
+     * @throws IllegalStateException if multiple charset values have been set for this media type
+     * @throws IllegalCharsetNameException if a charset value is present, but illegal
+     * @throws UnsupportedCharsetException if a charset value is present, but no support is available
+     *     in this instance of the Java virtual machine
```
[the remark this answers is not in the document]
trustin: Fixed, although it was forked from Guava :-)

PR #2375, opened 2020-01-06
Remove onEnter() and onExit() from RequestContext

minwoox:
Motivation:
I think `onEnter` has two drawbacks:
- `onEnter` can't be invoked when the context is first pushed into the thread-local
  - A user attaches the hook to the context which is already pushed.
- Most of the users do not know when `onEnter` is called. They just guessed. These are a few examples that I've recently heard and seen:
  - A user tried to handle `RequestTimeout` using `isTimedOut` and `onExit()`.
    - I guided him to use `setRequestTimeoutHandler`
  - A user puts a heavy initialization logic in `onEnter` and clearing logic in `onExit`
    - He could initialize once and clear it using `RequestLogAvailability.COMPLETE`
  - A user tried to send a different `HttpStatus` than `200` in the Thrift service.
    - What????

These APIs can give another option to users when they want to do something with a `RequestContext`. But most of the time, it's just confusing and no obsolete.
I think we can do what we want to do without `onEnter` and `onExit`. Therefore, I hope to get rid of `onEnter` and `onExit`.
Also, after we support looking up attributes from the `root`, we don't need `onChild` anymore.

Modification:
- Remove those APIs

Results:
- (Breaking)
  - `RequestContext.onEnter()` is gone.
  - `RequestContext.onExit()` is gone.
  - `RequestContext.onChild()` is gone.
- (Deprecated)
  - Use `RequestContext.push()` instead of `RequestContext.push(runCallbacks)` because you cannot attach callbacks to a `RequestContext`.


trustin:
@delegacy IIRC you are using these callbacks extensively. Could you tell us if there is anything impossible (or very inconvenient) without them?

anuraaga:
We originally added these methods to interface with brave's scope management, which is by default implemented as a `ThreadLocal`. But we found that brave allows customizing scope management via `CurrentRequestContext` and we could just do it directly.

I surveyed some libraries (opentelemetry, opencensus, jaeger, datadog, logback, log4j2) and they all provide a way to customize the scope management and not rely on `ThreadLocal`. So it seems like in general these methods shouldn't be needed. But we may run into a library that didn't provide such an extension point and users would have trouble dealing with context in such a case - actually, we do know of such a library that doesn't provide the extension point for context management, it's name is Armeria :) 

We can take inspiration from grpc-context. It doesn't have `on*` callbacks

https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html

But instead provides the ability to customize `Context.Storage` - https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.Storage.html

This customization can be used to implement `onEnter/Exit` functionality, or it can be used to replace the storage completely. Armeria should probably provide a similar interface to match the best practice of all these other libraries. If a user does need `onEnter/Exit`, they can override the storage and wrap the default, and hopefully if documented as `Storage`, it reduces foot-shooting.

minwoox:
> This customization can be used to implement onEnter/Exit functionality,

So you think we still need `onEnter` and `onExit` functionality? The reason that I made this draft is that it was hard for me to find the reason that we need `onEnter` and `onExit`. If we still need those, I think I can just close this PR and use them.

> it can be used to replace the storage completely.

Do you mean replacing `attr`?

anuraaga:
While the common libraries I surveyed do allow customization that works well even without `onEnter/Exit` I can definitely imagine finding something that doesn't. So it's probably good to have the functionality in some form, but I think we can definitely provide a better API.

> Do you mean replacing attr?

Nah, that's the map within the context, I mean where the context itself is stored when pushed, so to replace the usage of our `ThreadLocal` - for example, `RequestContext` can be stored in gRPC `Context` instead of our own `ThreadLocal` if that's convenient for some users. This is exactly the sort of customization we take advantage of for our brave integration, and I think it's good if we provide it in Armeria as well.

> onEnter can't be invoked when the context is first pushed into the thread-local

This is actually a pretty big limitation of the current API, and it would also be solved if instead we allow customization of `Storage` similar to grpc-context.

minwoox:
> So it's probably good to have the functionality in some form

So it's not the callback executed when the context is pushed in and out, but something working differently such as how gRPC handles it, right?

> This is exactly the sort of customization we take advantage of for our brave integration, and I think it's good if we provide it in Armeria as well.

Yes, I agree.

> This is actually a pretty big limitation of the current API, and it would also be solved if instead we allow customization of Storage similar to grpc-context.

It's like the first one answered right? We need a different API for that.

So you think it's time to introduce the `storage` thing to Armeria before we release 1.0? or Can we just remove `onEnter` and `onExit`, and add a new API after 1.0?

anuraaga:
> So it's not the callback executed when the context is pushed in and out, but something working differently such as how gRPC handles it, right?

Yeah, I think it'd be better in that form since it's more flexible.

> So you think it's time to introduce the storage thing to Armeria before we release 1.0? or Can we just remove onEnter and onExit, and add a new API after 1.0?

I think it'd be fine to remove first and add after 1.0 since it would be a compatible change, unless we hear from a user they currently require the functionality for their server.

minwoox:
> I think it'd be fine to remove first and add after 1.0 since it would be a compatible change, unless we hear from a user they currently require the functionality for their server.

OK, then let's keep this open a little bit if there's some issue with this.
Thank you as always. 😄

anuraaga:
I found some usage by @tobias- which looks quite valid, saving the request context into a DI scope

https://github.com/SourceForgery/tachikoma/blob/6a1d8e3d3e057a8c8c15ceed377c1e0519da5853/tachikoma-webserver/src/main/kotlin/com/sourceforgery/tachikoma/webserver/grpc/HttpRequestScopedDecorator.kt#L31

@tobias- just out of curiosity, is it possible to provide `RequestContext.current()` in the DI container instead of using the callbacks, similar to [this](https://github.com/line/armeria/pull/2363/files#diff-d974e102ab978b10f178e799f7639fdcR46)? So go inside-out instead of outside-in? I'm not familiar with `hk2` so just a random idea but don't know if it makes sense.

minwoox:
> SourceForgery/tachikoma:tachikoma-webserver/src/main/kotlin/com/sourceforgery/tachikoma/webserver/grpc/HttpRequestScopedDecorator.kt@6a1d8e3#L31

I'm wondering it this works well. 🤔 Because the ctx is already in the thread-local so the first `onEnter` is not called... hm..

anuraaga:
> I'm wondering it this works well. 🤔 Because the ctx is already in the thread-local so the first onEnter is not called... hm..

`runInScope` also calls `setCurrent` :)

minwoox:
> runInScope also calls setCurrent :)

Ah, I missed that. 😅

minwoox:
> This is actually a pretty big limitation of the current API, and it would also be solved if instead we allow customization of Storage similar to grpc-context.

Had a chat with @trustin, how about adding a callback to the `ServiceBindingBuilder` such as `onContextEnter(Consumer<? super ServiceRequestContext> callback) {...}`. (Of course, `VirtualHostBuilder` and `ServerBuilder` as well as we always do.)
If we do that, we can remove the limitation.

For the client-side, I'm not sure if we need that. It can be handled using `withContextCustomizer` for the first time and there's not so much chance that the `ClientRequestContext` is pushed twice.

anuraaga:
I sent https://github.com/SourceForgery/tachikoma/pull/191 to migrate that class to the same pattern we use for brave, if I understand correctly it should work well there instead of threadlocal-syncing.

anuraaga:
Maybe `contextPushHandler` or something? Ideally this allows customizing the storage of the context itself, so it implements the logic of `push` itself. This would provide parity with similar APIs like `grpc-context`, brave, etc.

Unfortunately, using an SPI-type of interface like `grpc-context` or `opentelemetry` does might be best for it though as if we do allow customizing push storage, than server and client need to be using the same implementation all the time.

Of course, we can also decide to just not allow customization of the storage, but it does seem like a nice feature and sort of an industry standard for all these libraries that have the concept of context.

minwoox:
> I sent SourceForgery/tachikoma#191 to migrate that class to the same pattern we use for brave, if I understand correctly it should work well there instead of threadlocal-syncing.

Wow, thanks for that. All users can do the same change if they are using `onEnter` already.

> Of course, we can also decide to just not allow customization of the storage, but it does seem like a nice feature and sort of an industry standard for all these libraries that have the concept of context.

Yes it seems like a nice feature. 😄 So we go back to removing `onEnter` and `onExit` before 1.0 and adding that feature after 1.0. Right?

anuraaga:
> Yes it seems like a nice feature. 😄 So we go back to removing onEnter and onExit before 1.0 and adding that feature after 1.0. Right?

Yup that's the recommendation I lean towards - hopefully we don't hear countercases 😄

minwoox:
This is ready. 😄

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -242,8 +235,7 @@ default SafeCloseable push(boolean runCallbacks) {
      * <pre>{@code
      * ctx.onChild((curCtx, newCtx) -> {
      *     assert ctx == curCtx && curCtx != newCtx;
-     *     // Add a callback to the child context.
-     *     newCtx.onExit(() -> { ... });
+     *     ...
```
minwoox: It's kinna difficult for me to show a proper example for this right now. Do you have any thing recommended?
Or can we remove this method as well? 🤔

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -596,60 +588,61 @@ default Logger makeContextAware(Logger logger) {
...
-     * Registers {@code callback} to be run when re-entering this {@link RequestContext}, usually when using
-     * the {@link #makeContextAware} family of methods. Any thread-local state associated with this context
-     * should be restored by this callback.
+     * <p>This method has a limitation which is that the callback cannot be called when this
+     * {@link RequestContext} is pushed first. So we decided to remove this method and provide
+     * better API after 1.0. If you are using this, please find another way to achieve the same behavior.
+     * Most of the cases, you might be able to just use {@link RequestLogListener}.
+     * Also, you might follow this approach which is using your own thread-local.
+     * https://github.com/SourceForgery/tachikoma/pull/191
+     * If you're still struggle to replace this method, please come to our slack workspace and let us know.
```
trustin: If you're still struggling to

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -596,60 +588,61 @@ default Logger makeContextAware(Logger logger) {
...
+     * <p>Please check out https://github.com/line/armeria/pull/2375
      *
-     * @deprecated Use {@link #onEnter(Consumer)} instead.
+     * @deprecated Do not use this method.
      */
     @Deprecated
-    default void onEnter(Runnable callback) {
-        onEnter(ctx -> callback.run());
+    default void onEnter(Consumer<? super RequestContext> callback, String doNotUseThisMethod) {
+        /* no-op */
```
trustin: How about just throwing `UnsupportedOperationException` with some message instead of adding a parameter?

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
trustin: I'm not sure if we can remove this method or not. It seems like it's not used anywhere in our code, though. Slightly in favor of removal.

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
ikhoon: `UnsupportedOperationException` is a good choice.
Or how about making breaking changes in this case by removing these methods? 
Because 1) I guess @minwoox added `String doNotUseThisMethod` for making breaking changes on purpose to produce compile errors and 2) Users can easily detect that it does not work anymore.

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
```diff
@@ -37,54 +37,20 @@ public static SafeCloseable noopSafeCloseable() {
...
-     * Runs callbacks if {@code runCallbacks} is {@code true} and returns the {@link SafeCloseable}
-     * which will set the root in the thread-local when {@link SafeCloseable#close()} is invoked.
-     */
-    public static SafeCloseable pushWithRootCtx(ClientRequestContext currentCtx, ServiceRequestContext root,
-                                                boolean runCallbacks) {
-        return pushWithRootAndOldCtx(currentCtx, root, root, runCallbacks);
-    }
-
     /**
      * Runs callbacks in {@code currentCtx} and {@code root} if {@code runCallbacks} is {@code true} and
```
ikhoon: Remove this condition? `if {@code runCallbacks} is {@code true}`

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -596,60 +588,61 @@ default Logger makeContextAware(Logger logger) {
...
-
-    /**
-     * Registers {@code callback} to be run when re-entering this {@link RequestContext}, usually when using
-     * the {@link #makeContextAware} family of methods. Any thread-local state associated with this context
-     * should be restored by this callback.
+     * <p>This method has a limitation which is that the callback cannot be called when this
+     * {@link RequestContext} is pushed first. So we decided to remove this method and provide
+     * better API after 1.0. If you are using this, please find another way to achieve the same behavior.
+     * Most of the cases, you might be able to just use {@link RequestLogListener}.
+     * Also, you might follow this approach which is using your own thread-local.
```
anuraaga: I guess it's opposite, uses `RequestContext` instead of own thread-local.

Maybe looking at the PR is too confusing to make it worth linking to.

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
```diff
@@ -596,60 +588,61 @@ default Logger makeContextAware(Logger logger) {
...
+     * https://github.com/SourceForgery/tachikoma/pull/191
+     * If you're still struggle to replace this method, please come to our slack workspace and let us know.
+     * We might be a help to you.
```
anuraaga: We'll help you

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
anuraaga: I agree that if making a behavior breaking change anyways it's better to just remove the API. We want to deprecate only where users will be able to continue to use their unmodified code.

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
anuraaga: We used it before adding attr lookup from root context only for copying attributes. I can't think of a good use case not better using the root lookup so in favor of removing

core/src/main/java/com/linecorp/armeria/common/RequestContext.java
minwoox: Just wanted to give more information from Javadoc, but I guess it's too much. Let me remove this. Thanks!

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
minwoox: Let me remove this as well. 😄

minwoox:
Removed `onEnter`, `onExit` and `onChild`. 😄

trustin:
Thanks, @minwoox and reviewers.

minwoox:
Thank you all!

PR #2376, opened 2020-01-06
Generalize gRPC stub factory lookup mechanism

trustin:
Motivation:

`GrpcClientFactory` currently assumes that the stub factory methods
are always one of `newStub()`, `newBlockingStub()` and `newAsyncStub()`.

However, by using a plugin like `grpc-reactor`, the stub factory method
can be `newReactorStub()`

Modifications:

- Generalize the stub factory lookup mechanism in `GrpcClientFactory`.
- Miscellaneous:
  - Clean-up `logback-test.xml` so that the log messages from Project
    Reactor are logged.

Result:

Partial support for the gRPC client stubs generated with `grpc-reactor`.

### DIFF
--- a/grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactory.java
+++ b/grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactory.java
@@ -19,6 +19,7 @@ import static com.google.common.collect.ImmutableSet.toImmutableSet;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactory.java
```diff
@@ -101,61 +102,78 @@
...
+        Method newStubMethod = null;
+        for (Method method : stubClass.getDeclaredMethods()) {
+            final int methodModifiers = method.getModifiers();
+            if (!(Modifier.isPublic(methodModifiers) && Modifier.isStatic(methodModifiers))) {
+                // Must be public and static.
+                continue;
+            }
+
+            final String methodName = method.getName();
+            if (!(methodName.startsWith("new") && methodName.endsWith("Stub"))) {
```
anuraaga: Cool check :)

trustin:
Thanks for reviewing :-)

PR #2377, opened 2020-01-06
Return ClientRequestContext.root() when invoking ServiceRequest…

minwoox:
…t.current()

Motivation:
It makes sense that returning `ClientRequestContext.root()` if the client call is made by a service  when invoking `ServiceRequestContext.current()` and its family methods.

Modification:
- Return `ClientRequestContext.root()` if it exists when calling `ServiceRequestContext.current()` and `currentOrNull()`.
  - It still throws an exception if the `ClientRequestContext` does not have the root.

Result:
- More flexible API.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
@@ -16,7 +16,6 @@
 
 package com.linecorp.armeria.server;
 
-import static com.google.common.base.Preconditions.checkState;
 import static com.linecorp.armeria.internal.RequestContextUtil.noopSafeCloseable;
 import static com.linecorp.armeria.internal.RequestContextUtil.pushWithoutRootCtx;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -61,33 +60,54 @@
...
+        }
+
+        final ServiceRequestContext root = ((ClientRequestContext) ctx).root();
+        if (root != null) {
+            return root;
+        }
+
+        throw new IllegalStateException(
+                "The current context is not a server-side context and does not have a root " +
+                "which means that the context is not invoked by a server request. ctx:" + ctx);
```
trustin: Space after `:`?

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -61,33 +60,54 @@
...
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
minwoox: Oops~ 😅

core/src/test/java/com/linecorp/armeria/server/ServiceRequestContextTest.java
```diff
@@ -67,6 +72,11 @@ void currentOrNull() {
         try (SafeCloseable unused = ctx.push()) {
             assertThat(onEnterExitStack).hasSize(1);
             assertThat(ServiceRequestContext.currentOrNull()).isSameAs(ctx);
+            try (SafeCloseable unused1 = clientRequestContext().push()) {
+                assertThat(onEnterExitStack).hasSize(2);
+                assertThat(ServiceRequestContext.currentOrNull()).isSameAs(ctx);
```
anuraaga: I think we can add readability and some documentation by asserting `ClientRequestContext.current` and `RequestContext.current` too.

core/src/test/java/com/linecorp/armeria/server/ServiceRequestContextTest.java
```diff
@@ -87,6 +97,12 @@ void mapCurrent() {
             assertThat(onEnterExitStack).hasSize(1);
             assertThat(ServiceRequestContext.mapCurrent(c -> "foo", () -> "bar")).isEqualTo("foo");
```
anuraaga: I think these tests are too loose, and especially with the double-stack, the new test code is not testing anything interesting - we only assert that any context is mounted and mapped, but it may not necessarily be the right context. `c == ctx ? "foo" : "bar"`? Or add/read attributes? Not sure which is less tedious.

core/src/test/java/com/linecorp/armeria/server/ServiceRequestContextTest.java
minwoox: That's a good suggestion. Let me add it.

core/src/test/java/com/linecorp/armeria/server/ServiceRequestContextTest.java
minwoox: That's a good point. `c == ctx ? "foo" : "bar"` is fine for me. Thanks!

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -61,33 +60,54 @@
...
+            return (ServiceRequestContext) ctx;
+        }
+
+        final ServiceRequestContext root = ((ClientRequestContext) ctx).root();
+        if (root != null) {
+            return root;
+        }
+
+        throw new IllegalStateException(
+                "The current context is not a server-side context and does not have a root " +
```
ikhoon: nit: The Javadoc said `the current context is a {@link ClientRequestContext} and`. Don't we need to synchronize this too or revert the Javadoc?

minwoox: I synced it. 😄

minwoox: Well let me revert this, because the context is not a client request context as well. It could be a fakeContext in a unit test.

minwoox: In that case, the Javadoc still makes sense I guess.

ikhoon: Ah, I misunderstood. Let revert this. :-)

PR #2381, opened 2020-01-06
Remove `EndpointGroupRegistry`, allow specifying `EndpointGroup`

trustin:
…when building a client.

Related: #1084

Motivation:

Using a 'string' as a way to specify a target endpoint group makes it hard to build a clear dependency between an `EndpointGroup` and a `Client`.

Modifications:

- (Breaking) Remove `EndpointGroupRegistry`
- Change the input parameter from `Endpoint` to `EndpointGroup` wherever applicable, to allow a user to specify an `EndpointGroup` when building a client, e.g. `Clients.newClient(SessionProtocol, EndpointGroup)`.
- `ClientBuilderParams` now provides all properties to required for creating a derived `Client`.
  - (Breaking) `ClientFactory.newClient(ClientBuilderParams)` replaced all other `newClient()` methods.
    - A user was never supposed to use these methods but `Clients.newClient()` or `Clients.builder()` anyway.
  - Add various public validation methods to `ClientFactory`, which is used by other `ClientFactory` implementations and `DefaultClientBuilderParams`.
- (Breaking) `Endpoint` now always refers to a single host.
  - 'group:groupName' notation is not used anymore.
  - Remove `Endpoint.ofGroup()`
  - Remove `Endpoint.isGroup()`
  - Remove `Endpoint.groupName()`
  - Remove `Endpoint.resolve()`
- `EndpointGroup` is now an `EndpointGroupSelector`.
  - (Breaking) `ClientRequestContext.endpointSelector()` has been replaced with `endpointGroup()`.
  - When creating most `EndpointGroup`s, a user can specify an `EndpointSelectionStrategy`, which is weighted round-robin by default.
  - Add `EndpointGroup.selectionStrategy()`
  - Add `EndpointGroup.of()` which requires an `EndpointSelectionStrategy`
- (Breaking) Overhaul `armeria-retrofit`
  - Add `ArmeriaRetrofit` which provides the factory methods for `Retrofit` and `ArmeriaRetrofitBuilder`
  - A user can now specify a `WebClient` instead of specifying `ClientOption`s.
  - A user now has much more control over how a `WebClient` for a non-base URL is created.
- Miscellaneous:
  - Add `Clients.isUndefinedUri()`
  - Add `EmptyEndpointGroupException`
  - (Deprecation) `EndpointSelectionStrategy.ROUND_ROBIN` and `WEIGHT_ROUND_ROBIN` in favor or `unweightedRoundRobin()` and `weightRoundRobin()`
  - (Deprecation) Some `PropertiesEndpointGroup.of()` in favor of `PropertiesEndpointGroupBuilder`.
  - (Deprecation) `ZooKeeperEndpointGroup` constructors in favor of `ZooKeeperEndpointGroup.of()` and `ZooKeeperEndpointGroupBuilder`.
  - (Breaking) `StaticEndpointGroup` is not public anymore.

Result:

- Much cleaner API
- No group name clashes
- Lots of breaking changes if a user was using `EndpointGroupRegistry`
  extensively.


trustin:
Meant to create a draft PR but that big green button was so tempting. :rofl: Added `[WIP]` header.

trustin:
@anuraaga @minwoox @ikhoon How does it look? I think most stuff works, except `armeria-retrofit` which requires a base URL.

trustin:
@kojilin Because this change removed `EndpointGroupRegistry` completely, `armeria-retrofit` required quite a bit of breaking changes as well. Here's the test case that shows the new API that allows:

- using client-side load-balancing: https://github.com/trustin/armeria/blob/endpoint_group_meets_strategy/retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactoryTest.java#L447-L454
- using non-default `WebClient`
  - Enforcing H1C: https://github.com/trustin/armeria/blob/endpoint_group_meets_strategy/retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactoryTest.java#L520-L528
  - Setting timeout: https://github.com/trustin/armeria/blob/endpoint_group_meets_strategy/retrofit2/src/test/java/com/linecorp/armeria/client/retrofit2/ArmeriaCallFactoryLargeStreamTest.java#L96-L105

What do you think? Is there any use case I did not cover yet? Do you need any convenience methods in `ArmeriaRetrofitBuilder`?

trustin:
Build now passes with `-PnoCheckstyle` option. Will remove WIP after fixing Checkstyle violations and documentation.

ikhoon:
Oops... I tapped Approved. 😅

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
```diff
@@ -77,91 +84,47 @@
...
-        requireNonNull(baseUrl, "baseUrl");
-        checkArgument(SessionProtocol.find(baseUrl.getScheme()) != null,
-                      "baseUrl must have an HTTP scheme: %s", baseUrl);
-        final String path = baseUrl.getPath();
-        if (!path.isEmpty() && !SLASH.equals(path.substring(path.length() - 1))) {
-            throw new IllegalArgumentException("baseUrl must end with /: " + baseUrl);
-        }
-        this.baseUrl = baseUrl.toString();
+    public ArmeriaRetrofitBuilder endpointGroup(
+            Function<? super HttpUrl, ? extends EndpointGroup> endpointGroupMapping) {
```
anuraaga: I'm not quite understanding, why do we need to map URL to endpoint group instead of just accepting a single endpoint group used for the retrofit client, similar to the normal web client?

trustin: Retrofit has an interesting annotation called `@Url` which allows a user send a request to an arbitrary URL, and a user should be able to choose an `EndpointGroup` in that case, too. See `ArmeriaCallFactoryTest.urlAnnotation*()` for examples.

anuraaga: I think it's similar to webclient as well when issuing a request to absolute URL. I guess it would still send to the registered endpoint group and just change the host header. I don't know if we really need to allow `@Url` to resolve to an endpoint group - if the client has no endpoint group registered the URL can be resolved dynamically, if it does we just send to the registered endpoint group changing the host header. This seems simpler than using a mapping and I think preserves retrofit behavior well in the case of no endpoint group.

trustin: Not sure I understood what you suggest. Could you elaborate?

trustin: > if the client has no endpoint group registered, the URL can be resolved dynamically, if it does, we just send to the registered endpoint group changing the host header.

There can be more than one `EndpointGroup` involved here, so some kind of mapping is inevitable:

```java
EndpointGroup groupA = ...;
EndpointGroup groupB = ...;

ArmeriaRetrofit.builder("https://group-a/")
               .endpointGroup(url -> {
                   switch(url.host()) {
                       case "group-a": return groupA;
                       case "group-b": return groupB;
                       default: Endpoint.of(url.host(), url.port());
                   }
               })
               .build();

service.fullUrl("https://group-b/foo");
```

trustin: I think I can simplify this at least for the case when no `@Url` annotation is used, like making it unnecessary to call `.endpointGroup()` at all.

trustin: Alternatively, we could provide a `WebClient` wrapper which intercepts a request and translates a certain host to an `EndpointGroup`, instead of providing this functionality only in `armeria-retrofit`.

anuraaga: I'm still finding that having multiple endpoint groups at this layer seems like overkill. URLs are basically host header / TLS but are somewhat independent from endpoint group - we resolve them to an endpoint when a user doesn't provide their own, but otherwise should just use the user specified endpoint independent of host. I think in common usage it will almost always be one endpoint group per client, and worse case a user can make multiple clients.

Also realized that it seems like if a user really needs host-aware endpoint selection, it could be fairly simply implemented as an endpoint selector that delegates to group selectors based on host?

trustin: It should look somewhat better now. What do you think?

- A user can now specify the base `WebClient` when building `Retrofit` with `ArmeriaRetrofit`.
  - In most cases, a user can specify a customized `WebClient` here.
- Removed `ArmeriaRetrofitBuilder.endpointGroup()`
- Replaced `ArmeriaRetrofitBuilder.webClient()` with `nonBaseWebClientFactory()`.
  - The biggest difference from the previous version is that it is used only for non-base URLs, i.e. a user does not need to call this in most cases.
  - A user can still replace a certain host with an `EndpointGroup` to leverage client-side load-balancing for non-base URLs. (See the example in `nonBaseClientFactory()` Javadoc)

trustin:
This pull request is now ready for full reviews.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
```diff
@@ -41,127 +39,118 @@
...
+     * ArmeriaRetrofit.of(WebClient.builder("http://example.com/")
+     *                             .setHttpHeader(HttpHeaderNames.AUTHORIZATION,
+     *                                            "bearer my-access-token")
+     *                             .build());
+     * }</pre></p>
      */
-    public ArmeriaRetrofitBuilder withClientOptions(
-            BiFunction<String, ? super ClientOptionsBuilder, ClientOptionsBuilder> configurator) {
-        this.configurator = requireNonNull(configurator, "configurator");
+    public ArmeriaRetrofitBuilder nonBaseClientFactory(
```
anuraaga: Since normal usage doesn't need to worry about this, I guess it's ok. But doesn't it seem like too advanced functionality? The asymmetry vs `WebClient` (the one I pointed to above) seems unnatural too - if we really want this functionality, we could move it to a `ClientOption` so it applies to both?

And while it doesn't actually solve the asymmetry problem by itself, I'm wondering whether we could just let the `EndpointSelector` layer take care of it, it seems simpler to reason about and don't even need to worry about a client cache

```java
class HostAwareEndpointGroup implements EndpointGroup {
  public Endpoint select(ClientRequestContext ctx) {
    switch (ctx.request().authority()) {
      case "group-bar":
        return groupBarEndpointGroup.select(ctx);
      ...
```

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
```diff
@@ -41,127 +39,118 @@
...
-     * Sets the {@link BiFunction} that customizes the underlying {@link WebClient}.
+     * Specifies the {@link BiFunction} that creates a new non-base {@link WebClient}, which is used for
+     * sending requests to other authorities than that of base URL. If not specified, the non-base
+     * {@link WebClient} will have the same options with the base {@link WebClient}, which was specified
+     * with {@link ArmeriaRetrofit#of(WebClient)} or {@link ArmeriaRetrofit#builder(WebClient)}.
+     *
+     * <p>To avoid the overhead of repetitive instantiation of {@link WebClient}s, the {@link WebClient}s
+     * returned by the specified {@link BiFunction} will be cached for each combination of:
+     * <ul>
+     *   <li>Whether the connection is secured (HTTPS or HTTPS)</li>
```
anuraaga: If this is the cache key, I guess it should only be these three fields that are passed to the callback, not `HttpUrl` - it's possible to create a broken callback if inspecting other fields of `HttpUrl`.

core/src/main/java/com/linecorp/armeria/client/DefaultWebClient.java
```diff
@@ -69,7 +67,7 @@ public HttpResponse execute(HttpRequest req) {
             return execute(endpoint, newReq);
```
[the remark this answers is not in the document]
trustin: Yeah, I thought about this briefly but wasn't sure which is a better place. Let me play with this idea.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
trustin: I didn't want to bother adding a factory interface, so... :laughing: Let me try to clean it up.

trustin: Just fixed by passing `Endpoint` instead of `HttpUrl`.

core/src/main/java/com/linecorp/armeria/client/DefaultWebClient.java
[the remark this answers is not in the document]
trustin: Added `WebClientOptions.ENDPOINT_REMAPPER` and moved endpoint remapping to `DeaultWebClient`.

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofitBuilder.java
trustin: I think `ENDPOINT_REMAPPER` options handles this better. What do you think?

core/src/main/java/com/linecorp/armeria/client/DecoratingClientFactory.java
```diff
@@ -49,10 +49,28 @@ protected DecoratingClientFactory(ClientFactory delegate) {
...
 
+    /**
+     * Creates a new {@link HttpClient} which uses the same {@link SessionProtocol}, {@link EndpointGroup} and
+     * {@link ClientOptions} with the specified {@link ClientBuilderParams}. Note that {@code path} and
+     * {@link SerializationFormat} are always {@code "/"} and {@link SerializationFormat#NONE}.
+     */
+    protected final HttpClient newHttpClientDelegate(ClientBuilderParams params) {
+        final URI uri = params.uri();
+        if (Clients.isUndefinedUri(uri)) {
+            return (HttpClient) delegate().newClient(
```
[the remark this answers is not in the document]
trustin: I think `delegate()` is correct because a `DecoratingClientFactory` usually has `HttpClientFactory` as its delegate. It's not very clear, though. Let me make this method require a `ClientFactory` explicitly.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -173,24 +176,29 @@ public ClientBuilder serializationFormat(SerializationFormat format) {
...
+        return cast;
     }
 
     private Scheme scheme() {
         return scheme == null ? Scheme.of(format, protocol) : scheme;
     }
 
-    private void ensureEndpoint() {
-        if (endpoint == null) {
+    private void ensureEndpointGroup() {
```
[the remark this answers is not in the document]
trustin: Good point. We will also need the same for `WebClientBuilder`. Maybe it's too much? :thinking:

core/src/main/java/com/linecorp/armeria/client/WebClientOptions.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.linecorp.armeria.client.endpoint.EndpointGroup;
+
+/**
+ * {@link ClientOption}s to control {@link WebClient}-specific behavior.
+ */
+public final class WebClientOptions {
+
+    /**
+     * A {@link Function} that remaps an {@link Endpoint} or an absolute URL's authority into
```
anuraaga: Looks nice

core/src/main/java/com/linecorp/armeria/client/endpoint/EndpointSelectionStrategy.java
```diff
@@ -26,13 +26,37 @@
...
+    static EndpointSelectionStrategy weightedRoundRobin() {
+        return WeightedRoundRobinStrategy.INSTANCE;
+    }
+
+    /**
+     * Returns an unweighted round-robin strategy, which ignores {@link Endpoint#weight()}.
+     *
+     * @see #weightedRoundRobin()
+     */
+    static EndpointSelectionStrategy unweightedRoundRobin() {
```
anuraaga: I think it's fine to be `roundRobin` and `weightedRoundRobin`, I don't know if I commonly hear the word unweighted round robin

core/src/main/java/com/linecorp/armeria/client/DefaultWebClient.java
```diff
@@ -40,9 +41,29 @@
...
+    }
+
+    private static ClientBuilderParams fillDefaultOptions(ClientBuilderParams params) {
+        final ClientOptions options = params.options();
+        if (options.getOrNull(WebClientOptions.ENDPOINT_REMAPPER) != null) {
+            return params;
+        }
+
+        final ClientOptionsBuilder optionsBuilder = options.toBuilder();
+        optionsBuilder.option(WebClientOptions.ENDPOINT_REMAPPER, Function.identity());
```
minwoox: Can we just set this to `ClientOptions.DEFAULT_OPTIONS` so that we don't have to call this method?

core/src/main/java/com/linecorp/armeria/client/endpoint/EmptyEndpointGroupException.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2016 LINE Corporation
```
minwoox: 😆

core/src/main/java/com/linecorp/armeria/client/endpoint/EmptyEndpointGroupException.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package com.linecorp.armeria.client.endpoint;
+
+import com.linecorp.armeria.client.Endpoint;
+import com.linecorp.armeria.common.Flags;
+
+/**
+ * An {@link EndpointGroupException} raised when the resolution of an {@link EndpointGroup} fails
+ * because there are no {@link Endpoint}s in the {@link EndpointGroup}..
```
minwoox: nit: two dots.

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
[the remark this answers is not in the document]
trustin: Not? :sweat_smile:

core/src/main/java/com/linecorp/armeria/client/DefaultWebClient.java
trustin: This option is applicable only to `WebClient`, so I think it doesn't belong there, like gRPC options are not there.

core/src/main/java/com/linecorp/armeria/client/endpoint/EmptyEndpointGroupException.java
trustin: :scream:

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -153,86 +156,10 @@ static void disableShutdownHook() {
...
-     *
-     * @param scheme the {@link Scheme} for the {@code endpoint}
-     * @param endpoint the server {@link Endpoint}
-     * @param path the service {@code path}
-     * @param clientType the type of the new client
-     * @param options the {@link ClientOptions}
-     */
-    <T> T newClient(Scheme scheme, Endpoint endpoint, @Nullable String path, Class<T> clientType,
-                    ClientOptions options);
+    Object newClient(ClientBuilderParams params);
```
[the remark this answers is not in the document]
trustin: I tried but it seems to make the overall type signature more complex without much gain, like adding another type parameter to `UserClient`.

Perhaps it's not worth it given a user is not supposed to create a client using a `ClientBuilderParams`?

core/src/main/java/com/linecorp/armeria/client/DefaultWebClient.java
trustin: Just realized this option mght be applicable for any clients. Let me play with this idea a little bit.

trustin: Moved `ENDPOINT_MAPPER` to `ClientOptions` and it looks better. This will be useful to non-`WebClient`s when:

- A user is given with a client that talks to a single `Endpoint` and wants to override that.
- We provide a way to send a request to an arbitrary `Endpoint` by using a single client instance in the future.

trustin:
Moved `ENDPOINT_MAPPER` to `ClientOptions` and it looks better. This will be useful to non-`WebClient`s when:

- A user is given with a client that talks to a single `Endpoint` and wants to override that.
- We provide a way to send a request to an arbitrary `Endpoint` by using a single client instance in the future.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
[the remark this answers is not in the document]
minwoox: Ah, thanks! It's fine as it is. 😄

PR #2382, opened 2020-01-07
Revert RpcResponse not to extend ContextAwareFuture

minwoox:
Motivation:
As discussed in #2347, we don't return `RequestContextAwareCompletableFuture` from `client.execute().aggregate().
For the same behavior, we should not return `RequestContextAwareRpcResponse` when invoking Thrift call.

Modifications:
- `DefaultRpcResponse` does not extend `RequestContextAwareCompletableFuture`.
  - Revert most of the work in #2328
- `DefaultRpcResponse` and `RequestContextAwareCompletableFuture` now warns when calling its blocking methods from a event loop.

Result:
- Revert #2328
  - But You can still get `ClientRequestContext` in `AsyncMethodCallback`
- Warn if calling blocking operation on `RpcResponse` and `RequestContextAwareCompletableFuture`.


minwoox:
It's all thanks to you. :smile:

PR #2383, opened 2020-01-08
Fix to have context in the thread-local when calling AsyncMetho…

minwoox:
Motivations:
If the `reply` returned from Thrift client call is complete before calling `future.handle()`, the context is not pushed into thread-local when calling `AsyncMethodCallback`.

Modifications:
- Fix to push context right before calling `future.handle()`
- `ClientRequestContextCaptor` now captures nested contexts.
  - This is not a breaking change because we didn't publish a new version yet.
  - Add `ClientRequestContextCaptor.getOrNull()`

Result:
- Fix test failure

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientThreadLocalState.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientThreadLocalState.java
@@ -123,7 +123,7 @@ final class ClientThreadLocalState {
         final List<ClientRequestContext> captured = new ArrayList<>();
 
         @Nullable
-        private final DefaultClientRequestContextCaptor oldCaptor;
+        private DefaultClientRequestContextCaptor oldCaptor;
 
         DefaultClientRequestContextCaptor(@Nullable DefaultClientRequestContextCaptor oldCaptor) {
             this.oldCaptor = oldCaptor;
@@ -166,6 +166,7 @@ final class ClientThreadLocalState {
         @Override
         public void close() {
             pendingContextCaptor = oldCaptor;
+            oldCaptor = null;
             maybeRemoveThreadLocal();
         }
     }
--- a/core/src/main/java/com/linecorp/armeria/client/ClientThreadLocalState.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientThreadLocalState.java
@@ -82,7 +82,7 @@ final class ClientThreadLocalState {
 
     void addCapturedContext(ClientRequestContext ctx) {
         if (pendingContextCaptor != null) {
-            pendingContextCaptor.captured.add(ctx);
+            pendingContextCaptor.add(ctx);
         }
     }
 
@@ -129,12 +129,28 @@ final class ClientThreadLocalState {
             this.oldCaptor = oldCaptor;
         }
 
+        void add(ClientRequestContext ctx) {
+            captured.add(ctx);
+            if (oldCaptor != null) {
+                oldCaptor.add(ctx);
+            }
+        }
+
         @Override
         public ClientRequestContext get() {
             if (captured.isEmpty()) {
                 throw new NoSuchElementException("No context was captured; no request was made?");
             }
             return captured.get(0);
+        }
+
+        @Nullable
+        @Override
+        public ClientRequestContext getOrNull() {
+            if (!captured.isEmpty()) {
+                return captured.get(0);
+            }
+            return null;
         }
 
         @Override
--- a/thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientInvocationHandler.java
+++ b/thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientInvocationHandler.java
@@ -16,6 +16,8 @@
 
 package com.linecorp.armeria.client.thrift;
 
+import static com.linecorp.armeria.common.thrift.AsyncMethodCallbacks.invokeOnError;
+
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.net.URI;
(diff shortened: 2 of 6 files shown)


core/src/main/java/com/linecorp/armeria/client/ClientThreadLocalState.java
```diff
@@ -129,6 +129,13 @@ private static void reportThreadSafetyViolation() {
             this.oldCaptor = oldCaptor;
         }
 
+        void add(ClientRequestContext ctx) {
+            captured.add(ctx);
+            if (oldCaptor != null) {
+                oldCaptor.add(ctx);
+            }
+        }
```
trustin: How about not adding after `close()`?

thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientInvocationHandler.java
```diff
@@ -119,31 +124,51 @@ private Object invokeClientMethod(Method method, @Nullable Object[] args) throws
...
-                AsyncMethodCallbacks.transfer(reply, callback);
-                return null;
-            } else {
-                try {
-                    return reply.get();
-                } catch (ExecutionException e) {
-                    throw Exceptions.peel(e);
+                final ClientRequestContext ctx = captor.get();
+                if (callback != null) {
+                    reply.handle(ctx.makeContextAware((res, cause) -> {
```
anuraaga: I think we can directly mount the context here and below instead of using `makeContextAware`.

minwoox:
> Is it possible to add a unit test for the thrift behavior change?

I thought `ClientRequestContextPushedOnCallbackTest` handles the cases, but it seems that I missed one. Will add it. 😄

thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientInvocationHandler.java
minwoox: https://github.com/line/armeria/pull/2383/files#diff-6ab8399d40ca91dd25785dc822606306R167

This line should be called by the same executor so I chose this way which is let the thread which complete the `reply` also does the jobs in callback.
But I'm not so strong this and I can put the job in the executor.

core/src/main/java/com/linecorp/armeria/client/ClientThreadLocalState.java
minwoox: Oops. 😅

thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientInvocationHandler.java
anuraaga: I'm not sure I understand, probably since I wasn't clear either - I didn't mean use the executor, but just something like this to directly mount the context without creating a wrapped callable

```java
reply.handle((res, cause) -> {
    if (cause != null) {
                     invokeOnError(callback, cause);
                } else {
                    callback.onComplete(res);
                            }
                        } catch (Exception e) {
                            CompletionActions.log(e);
                        }
                        return null;
                    }));
```

minwoox: Well actually, I can't get an `executor` in that case. Is it okay to execute the callback`(invokeOnError(callback, cause);)` with the different thread in this case?

core/src/main/java/com/linecorp/armeria/client/ClientThreadLocalState.java
minwoox: Changed. But I couldn't make a test case because `oldCaptor.add()` is never called...

thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientInvocationHandler.java
anuraaga: By the way, as far as the thread, from what I understand, the thread running the `handle` callback will be the thread in `delegate().execute(` and isn't related to this code directly. Though I guess this code, `delegate().execute` will in practice be the same thread, which happens to be `ctx.contextAwareExecutor` too. JFYI, I don't think it affects how we mount the context here.

anuraaga:
Cool if there's another test to add that's nice too, but I was also just misunderstanding that this is a new fix, didn't realize it's for the test failures that are already happening on `HEAD`

thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientInvocationHandler.java
minwoox: Well, it could be the thread which calls `reply.handle()` (i,e, reply is completed when attaching callback. 😭  This is why previous [PR](https://github.com/line/armeria/commit/845f482e6c39e1c55b4434473466ffd7af4409d5#commitcomment-36690155) was not working )

minwoox: > I'm not sure I understand, probably since I wasn't clear either - I didn't mean use the executor, but just something like this to directly mount the context without creating a wrapped callable

Yeah, I can do that and I misunderstood it. 😅

minwoox:
> Cool if there's another test to add that's nice too, but I was also just misunderstanding that this is a new fix, didn't realize it's for the test failures that are already happening on HEAD

Thanks to you, I found one that I missed. 😆

thrift/src/main/java/com/linecorp/armeria/client/thrift/THttpClientInvocationHandler.java
anuraaga: Aha - thanks I understand the threading issue now. I never realized the non-async callback methods had such a dangerous caveat to them

trustin:
Thanks, @minwoox :pray:

minwoox:
Thanks a lot for reviewing!

PR #2387, opened 2020-01-09
Preserve default `CHANNEL_OPTION`s in  `ClientFactoryBuilder`

### LINKED ISSUE #2385: Using ClientFactoryOptions causes NPE during client connection
`ClientFactoryOptions.DEFAULT` overrides the defaults set by `ClientFactoryBuilder` which causes NPE (at `HttpChannelPool` constructor) during client connection, which makes `ClientFactoryOptions` impossible to use.

**The code to reproduce the issue:**

```
  ClientFactoryBuilder clientFactoryBuilder = ClientFactory.builder();
  clientFactoryBuilder.options(ClientFactoryOptions.of()); // or any other "of" method
  clientBuilder.factory(clientFactoryBuilder.build());

```
**The exception thrown during client connect:**

```
    at com.linecorp.armeria.client.HttpChannelPool.<init>(HttpChannelPool.java:109)

```
**The code that throws NPE (`HttpChannelPool.<init>`):**

```
  this.connectTimeoutMillis = (Integer)baseBootstrap.config().options().get(ChannelOption.CONNECT_TIMEOUT_MILLIS);
```

**Analysis:**

`ClientFactoryBuilder` sets some defaults during the construction [`ClientFactory.builder()` --> `new ClientFactoryBuilder()`].
More specifically it sets `CHANNEL_OPTIONS` with a default `CONNECT_TIMEOUT_MILLIS` as following:

```
  connectTimeoutMillis(Flags.defaultConnectTimeoutMillis());
```

In the debugger I see the following after executing `ClientFactory.builder()`:
```
  options = {java.util.LinkedHashMap@2624}  size = 1
   {com.linecorp.armeria.client.ClientFactoryOption@2629} "CHANNEL_OPTIONS" -> {com.linecorp.armeria.client.ClientFactoryOptionValue@2630} "CHANNEL_OPTIONS={CONNECT_TIMEOUT_MILLIS=>3200}"
```

On a contrary, `ClientFactoryOptions` defines its own set of defaults (via its `DEFAULT_OPTIONS`) which defines empty `CHANNEL_OPTIONS` as following:

```
  ClientFactoryOption.CHANNEL_OPTIONS.newValue(new Object2ObjectArrayMap<>()),
```

When `ClientFactoryBuilder#options()` method invoked with any `ClientFactoryOptions` it will override previous `CHANNEL_OPTIONS` default.
Then, it empties `CHANNEL_OPTIONS`, while `CONNECT_TIMEOUT_MILLIS` expected by `HttpChannelPool` and possibly by other places.

In the debugger I see the following after executing `clientFactoryBuilder.options(ClientFactoryOptions.of())`:
```
  {com.linecorp.armeria.client.ClientFactoryOption@2629} "CHANNEL_OPTIONS" -> {com.linecorp.armeria.client.ClientFactoryOptionValue@2688} "CHANNEL_OPTIONS={}"
```

**Expectation:**

1. The use of `ClientFactoryOptions` should not cause any failure. `ClientFactoryOptions` defaults should not override any defaults set by `ClientFactoryBuilder`.
2. The solution should only take those options defined by the user and apply those to `ClientFactoryBuilder`.
3. `HttpChannelPool` should not fail with NPE when some expected configurations are missing. It should provide reasonable problem indication or/and fall back to a default.

**Current Workaround:**

Avoid using `ClientFactoryOptions` and use individual user defined options as a collection of `ClientFactoryOptions` and add those to `ClientFactoryBuilder` via `ClientFactoryBuilder#option()` method as following:

```
  Collection<ClientFactoryOptionValue<?>> clientFactoryOptions = ...
  if (clientFactoryOptions != null && !clientFactoryOptions.isEmpty()) {
    clientFactoryOptions.forEach(clientFactoryBuilder::option);
  }
  clientBuilder.factory(clientFactoryBuilder.build());
```

ikhoon:
Motivation:
When the default ClientFactoryOptions is passed to ClientFactoryBulider via `options(ClientFactoryOptionsptions)`,
ChannelOption.CONNECT_TIMEOUT_MILLIS is set to null.
It causes `NullPointException` in initialization of  `HttpChannelPool`. #2385
https://github.com/line/armeria/blob/f0d870fde1088114070be31b67f7df0a21e835c6/core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java#L109-L110

Modifications:
* Make `ClientFactoryOption.CHANNEL_OPTIONS` immutable.
* Copy new CHANNEL_OPTIONS and preserve the existing one.

Result:
No more `NullPointException` with `ClientFactoryOptions.of()` during client connection.
Fixes #2385

/cc @delegacy @max904-github

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
@@ -40,6 +40,7 @@ import javax.net.ssl.TrustManagerFactory;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.base.MoreObjects.ToStringHelper;
+import com.google.common.collect.ImmutableMap;
 
 import com.linecorp.armeria.common.CommonPools;
 import com.linecorp.armeria.common.Flags;
@@ -55,7 +56,6 @@ import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.resolver.AddressResolverGroup;
 import io.netty.resolver.DefaultAddressResolverGroup;
 import io.netty.resolver.dns.DnsNameResolverBuilder;
-import it.unimi.dsi.fastutil.objects.Object2ObjectArrayMap;
 
 /**
  * Builds a new {@link ClientFactory}.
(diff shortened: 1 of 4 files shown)


trustin:
Could we make sure the resulting value of `CHANNEL_OPTIONS` is always immutable?

- `ClientFactoryOptions.DEFAULT_OPTIONS[CHANNEL_OPTIONS]` must have an immutable map. The current implementation is not safe because a user can get it and modify the default.
- When building or creating `ClientFactoryOptions`, we must make an immutable copy of `CHANNEL_OPTIONS` value.
  - See `ClientOptions.filterValue()` to learn how to filter user-specified option values.
  - Perhaps we should replace `Object2ObjectArrayMap` with `ImmutableMap`?

ikhoon:
> Could we make sure the resulting value of CHANNEL_OPTIONS is always immutable?

That is a good idea. I didn’t think of that. I will address comment soon. 😀

Issue #1263, opened 2018-06-18
Add an option to send a PING frame periodically for HTTP/2 connections


trustin:
.. both on client and server side. See https://httpwg.org/specs/rfc7540.html#PING

sivaalli:
Hi @trustin, Can I look into this? Or is it already implemented?

trustin:
Sure! Thanks for looking into it. It's not implemented yet.

PR #2388, opened 2020-01-10
Http2 Ping

sivaalli:
Issue: https://github.com/line/armeria/issues/1263

I've been working on this for a few days. And reached to a point to see if this code makes any sense. I've tested it a bit but still needs a lot of testing and also a **lot** of code changes to make it look decent enough.

@trustin @ikhoon @minwoox Please let me know if this code makes any sense. Also, please feel free to suggest anything that I may need to consider.


core/src/main/java/com/linecorp/armeria/common/DefaultPingHandler.java
```diff
@@ -0,0 +1,141 @@
...
+                                               ctx.fireUserEventTriggered(
+                                                       new Http2PingTimeoutEvent()
+                                               );
+                                           }, pingAckTimeout, TimeUnit.SECONDS);
+                                       } else {
+                                           // TODO send WriteFailed event instead of Object
+                                           ctx.fireUserEventTriggered(new Object());
+                                       }
+                                   });
+    }
```
sivaalli: But mostly this is the important piece. The flow is:

1. We schedule and ping 
2. When the PING write is success full we schedule an another runnable to notify for `Http2PingTimeoutEvent `
3. If we indeed get PING ACK back then cancel the future created in (2).

Any suggestions that you can recommend to make this not look like a callback hell?

sivaalli: I've thought of another approach that may look more simpler. Idea is:

1. We scheduleWithFixedDelay a PING write periodically (`pingDelay`).
2. We send PING if it's the time
3. On next schedule, we see if we got ACK for the ping sent in (2).

But the drawback for this approach is it takes a maximum of `pingDelay` to discover that we did not get ACK for the PING back.

sivaalli: Another idea is:

When we receive `IdleStateEvent` emitted by  `IdleTimeoutHandler`, then we send a PING to see if the connection is still alive. If we do not get an ACK back or write failed then we can close the connection or do any other necessary action.

trustin: How about using `scheduleWithFixedDelay()`, counting the number of send PINGs and received ACKs, and then disconnect if `Math.abs(numPingAcks - numPings)` exceeds a certain value?

core/src/main/java/com/linecorp/armeria/common/DefaultPingHandler.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Override
+    public void onPingRead(final ChannelHandlerContext ctx, final long data) throws Http2Exception {
+        // Notes: Http2ConnectionHandler by default send ACK when PING is received.
+        // So do not have write PING ACK here. But usefult for debugging purpose.
+        logger.trace("received a PING {} from client and " +
+                     "responding with PING for channel: {}",
+                     data,
+                     ctx.channel().id());
```
trustin: Could be:

    "{} Received a PING {} ....", ctx.channel(), data

..?

core/src/main/java/com/linecorp/armeria/common/DefaultPingHandler.java
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Default implementation of {@link PingHandler}.
+ */
+@NotThreadSafe
+public class DefaultPingHandler implements PingHandler {
+
+    private static final Logger logger = LoggerFactory.getLogger(DefaultPingHandler.class);
+
+    private final Http2FrameWriter frameWriter;
+    private final Http2PingRequestResponsePair pair = new Http2PingRequestResponsePair();
```
trustin: Perhaps we can just count the number of received/sent acks/pings?

core/src/main/java/com/linecorp/armeria/common/DefaultPingHandler.java
```diff
@@ -0,0 +1,141 @@
...
+    private final Random random = new Random();
```
trustin: Could use thread-local random?

core/src/main/java/com/linecorp/armeria/common/DefaultPingHandler.java
trustin: > When we receive `IdleStateEvent` emitted by  `IdleTimeoutHandler`, then we send a PING to see if the connection is still alive. If we do not get an ACK back or write failed then we can close the connection or do any other necessary action.

This also sounds good to me. How about combining `IdleTimeoutHandler` and PING/ACK counters?

trustin:
Thank you so much for this PR by the way. :+1:

anuraaga:
Just a thought it could be good to reference gRPC's pings for inspiration and to find any gotchas they may have experienced already.

https://github.com/grpc/grpc-java/search?utf8=%E2%9C%93&q=ping&type=

https://github.com/grpc/grpc-java/blob/c8a0af572c561d28f61e9fb4f6097a72d4ebe8e9/core/src/main/java/io/grpc/internal/KeepAliveManager.java

One thing I notice is they do use traffic received to control ping scheduling, while have a separate flag for whether to ping during idle - didn't look for the motivation for that flag yet.

And one gotcha-type thing I found is they only apply ping acks to flow control.

https://github.com/grpc/grpc-java/blob/9fcfb5b4f80b832d48665dc7ec2edfc15ed873dc/netty/src/main/java/io/grpc/netty/Http2ControlFrameLimitEncoder.java#L72

sivaalli:
I will look at gRPC's implementation. And will write here on how would I proceed so that we reach consensus. Thanks for your suggestions @anuraaga @trustin 🙇

sivaalli:
@trustin @anuraaga here are my thoughs on gRPC's PING(keepalive) implementation:

1. As @anuraaga  said, gRPC server uses traffic to control PING scheduling. The PING is scheduled initially and is re-scheduled where there is new data received. I think this basically is doing `send ping only when idle`. The default delay is 2 Hours. But for client-side this is controlled by a flag that enables/disables PINGs during idle.

2. And there are checks to make sure that client does not send too many pings to sever within a time limit which is a nice defensive protection. I think its good to have this feature.

3. Also PING acks are applied to flow control to limit number of flow control frames currently active. see [here](https://nvd.nist.gov/vuln/detail/CVE-2019-9515#vulnCurrentDescriptionTitle) for more details.

When a PING write has failed or peer does not ACK within configured timeout, the channel is closed assuming that peer has disconnected.

Also, I feel like some implementation in gRPC may be redundant (I could be totally wrong as well) may be because specification might have changed from when it was developed or may be gRPC extended http2 protocol to achieve some use case. For ex: if the PING payload is `0x97ACEF001L` (GRACEFUL_SHUTDOWN_PING)  then the channel is closed, but I could not find this in specification: https://http2.github.io/http2-spec/.

But overall I think we can send PING when there are no streams active so we maintain the liveliness of connection. Also making sure that client does not send too many ping within a time limit. Please let me know what do you think and also share any suggestions.

trustin:
> But overall I think we can send PING when there are no streams active so we maintain the liveliness of connection. Also making sure that client does not send too many ping within a time limit. Please let me know what do you think and also share any suggestions.

I guess we should send PINGs regardless of the number of open streams, because there may be many open yet idle streams in a connection. Did you mean 'active streams' by that, then we're on the same page. :smile: 

Sounds good to me besides that!

sivaalli:
Sorry for the verbiage. I meant `active streams`. Thanks for your suggestions and I resume my work.

sivaalli:
Will be closing this and will implement the changes as discussed and create a new PR.

PR #2391, opened 2020-01-10
Clean up TimeoutController implementation

ikhoon:
Motivation:
When a past time is passed to `ctx.set{Response,Request}TimeoutAt()`
it throws `IllegalArgumentException`.
I think it would be more natural to trigger timeout immediately.
And clean up TimeoutController implementation.

Modifications:
* Add `timeoutNow()` to `TimeoutController` for invoking timeout directly.
* Add `isTimedOut()` for checking whether the timeout has been timed out or not.
* Add `State` to `DefaultTimeoutController` for expressing the current timeout status.
* Rename `initTimeout` to `scheduleTimeout`.
* Change `void xxxTimeout(...)` to `boolean xxxTimeout(...)`.
* Allow to execute the pending task when the `TimeoutController` is set to the `{Client,Service}RequestContext`.
* Invoke `timeoutNow()` when the given time is before now in `ctx.set{Response,Request}TimeoutAt()`.
* Remove order of precedence for `xxxTimeout`.
  * Can call `resetTimeout` or `exendTimeout` before calling `scheduleTimeout`.

Result:
* A user can trigger timeout immediately by setting past time.
* More clarified implementations.


core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -149,6 +149,13 @@ public void resetTimeout(long newTimeoutMillis) {
         }
     }
 
+    @Override
+    public void timeoutNow() {
+        ensureInitialized();
+        cancelTimeout();
+        timeoutTask.run();
```
ikhoon: Does it need a guard here? 
```java
if (cancelTimeout()) {
    timeoutTask.run();
}
```

trustin: I guess yes, unless there's anything that must be done even after cancellation.

core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -497,6 +498,7 @@ default MediaType negotiatedProduceType() {
     /**
      * Schedules the request timeout that is triggered at the specified time represented
      * as the number since the epoch ({@code 1970-01-01T00:00:00Z}).
+     * Note that the request timeout is triggered immediately if the specified time is before now.
```
trustin: How about: the request will be timed out immediately

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -456,12 +456,20 @@ public void setRequestTimeoutAfter(Duration requestTimeout) {
...
+
+        if (requestTimeoutAfter <= 0) {
+            final TimeoutController requestTimeoutController = this.requestTimeoutController;
+            if (requestTimeoutController != null) {
+                if (eventLoop().inEventLoop()) {
+                    requestTimeoutController.timeoutNow();
+                } else {
+                    eventLoop().execute(requestTimeoutController::timeoutNow);
+                }
+            }
```
minwoox: Don't we need to add `pendingTimeoutTask` here?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -26,7 +26,7 @@
     /**
      * Initializes the timeout scheduler with the specified {@code timeoutMillis}.
```
minwoox: Initializes -> Sets?
Do you have a reason that you changed the name? I think I'd like `initTimeout`. 🤔

ikhoon: I didn’t update Javadoc yet. I’ll let you know when ready to review. 🙇‍♂️

ikhoon:
PTAL. 🙇‍♂️ 
/cc @minwoox

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
ikhoon: Indeed! Updated. 😀

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -78,74 +79,98 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
         }
+        lastExecutionTimeNanos = nanoTime;
+        timeoutFuture = eventLoop.schedule(timeoutTask, timeoutMillis, TimeUnit.MILLISECONDS);
+        return true;
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Note that the {@link TimeoutTask} should be set via the {@link #setTimeoutTask(TimeoutTask)} or
```
minwoox: It seems like there's not much chance that a new `TimeoutTask` it set.
How about removing the setter and setting it in the constructor?
```
// In HttpResponseDecoder
HttpResponseWrapper(...) {
    super(eventLoop, newTimeoutTask(delegate, ctx));
    this.delegate = delegate;
    this.ctx = ctx;
    this.maxContentLength = maxContentLength;
    this.responseTimeoutMillis = responseTimeoutMillis;
}

// In HttpResponseSubscriber
HttpResponseSubscriber(ChannelHandlerContext ctx, HttpObjectEncoder responseEncoder,
                       DefaultServiceRequestContext reqCtx, DecodedHttpRequest req,
                       boolean enableServerHeader, boolean enableDateHeader) {
    super(ctx.channel().eventLoop(), newTimeoutTask(reqCtx));
    this.ctx = ctx;
    this.responseEncoder = responseEncoder;
    this.req = req;
    this.reqCtx = reqCtx;
    this.enableServerHeader = enableServerHeader;
    this.enableDateHeader = enableDateHeader;
}
```

ikhoon: That is a nice suggestion. I tried to make `TimeoutTask` immutable in `DefaultTimeoutController` initially. We have to make `newTimeoutTask` static to call `newTimeoutTask` in `super`. But it is difficult because it refers to the `state` that is non-static field.

Or there is something I missed, let me know :-)

ikhoon:
> Could you leave a todo comment that how we're going to remove the duplicate code in DefaultClientRequestContext and DefaultServiceRequestContext as we discussed?

Sure!

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -24,27 +24,38 @@
...
      * @param newTimeoutMillis the new timeout value in milliseconds. {@code 0} if disabled.
+     * @return {@code true} if the current timeout is reset by the specified {@code newTimeoutMillis}.
      */
-    void resetTimeout(long newTimeoutMillis);
+    boolean resetTimeout(long newTimeoutMillis);
+
+    /**
+     * Trigger the current timeout immediately.
+     */
+    void timeoutNow();
```
trustin: Question, does it makes sense to make this method return a boolean value?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -24,27 +24,38 @@
 public interface TimeoutController {
 
     /**
-     * Initializes the timeout scheduler with the specified {@code timeoutMillis}.
+     * Schedules a new timeout with the specified {@code timeoutMillis}.
+     * If a timeout is scheduled already, this method will not start a new timeout.
+     * @return {@code true} if the timeout is scheduled.
```
trustin: Could you also document when `false` is returned, e.g. `{@code false} if the timeout could not be scheduled because .... or ....` (Please fix all methods.)

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -78,74 +79,98 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
+        return true;
+    }
+
+    @Override
+    public void timeoutNow() {
+        checkState(timeoutTask != null,
+                   "setTimeoutTask(timeoutTask) is not called yet.");
+        if (cancelTimeout()) {
+            timeoutTask.run();
         }
```
trustin: Could return whether the timeout task was run or not?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
ikhoon: That sounds good!

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -37,6 +39,8 @@
      *
      * @param adjustmentMillis the adjustment of time amount value in milliseconds.
      * @return {@code true} if the current timeout is extended by the specified {@code adjustmentMillis}.
+     *         {@code false} if the current timeout could not be extended
+     *         because of some constraints in the class that implements this interface.
```
trustin: What are the 'some constraints'?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -27,6 +27,8 @@
      * Schedules a new timeout with the specified {@code timeoutMillis}.
      * If a timeout is scheduled already, this method will not start a new timeout.
      * @return {@code true} if the timeout is scheduled.
+     *         {@code false} if the timeout could not be scheduled because the timeout is currently being
+     *         scheduled.
```
trustin: .. if the timeout has been scheduled already?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -45,17 +49,24 @@
      *
      * @param newTimeoutMillis the new timeout value in milliseconds. {@code 0} if disabled.
      * @return {@code true} if the current timeout is reset by the specified {@code newTimeoutMillis}.
+     *         {@code false} if the current timeout could not be reset
+     *         because of some constraints in the class that implements this interface.
```
trustin: What are the 'some constraints'?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -45,17 +49,24 @@
...
      */
     boolean resetTimeout(long newTimeoutMillis);
 
     /**
      * Trigger the current timeout immediately.
+     * @return {@code true} if the current timeout is triggered successfully.
+     *         {@code false} if the current timeout could not be triggered now because
+     *         the timeout schedule has already completed normally.
```
trustin: .. if the timeout has been triggered already?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -45,17 +49,24 @@
...
      */
-    void timeoutNow();
+    boolean timeoutNow();
 
     /**
      * Cancels the current timeout scheduled.
-     * @return {@code true} if the current timeout is canceled.
+     * @return {@code true} if the current timeout is canceled or no timeout is being scheduled.
+     *         {@code false} if the current timeout could not be cancelled typically because
+     *         the timeout schedule has already completed normally.
```
trustin: .. if the timeout has been triggered already?

What happens if no timeout was scheduled previously? What happens if a user calls this method and then scheduled a new timeout?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
ikhoon: The `resetTimeout` is failed when `TimeoutTask` can not schedule.
I've tried to explain the exact cause... but `TimeoutTask` is only used in `DefaultTimeoutController`.
In this case, do I need to remove this sentence and simplify with: 
```
{@code false} if the current timeout failed to reset
```
or remove the ambiguous `{@code false} ...` sentence?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
ikhoon: > What happens if no timeout was scheduled previously?

It should return `true`. Let me rewrite the `no timeout is being scheduled` to `no timeout was scheduled previously`.

> What happens if a user calls this method and then scheduled a new timeout?

A user can schedule a new timeout. I missed that point. Will update it to the `Javadoc`.

ikhoon:
Updated: Had chat with @trustin, and he suggested to add `State` for expressing the current timeout controller status. It was added at https://github.com/line/armeria/pull/2391/commits/b0c03af8b844183dcc70d665a7d03feb4ad77533

The `State` of `DefalutTimeoutController` will be transited to:
* `scheduleTimeout`
  - `DISABLED` -> Schedule a new timeout -> `SCHEDULE`
  - `!DISABLED -> No operation
* `extendTimeout`
  - `SCHEDULE` -> Extend the timeout -> `SCHEDULE`
  - `!SCHEDULE` -> No operation
* `resetTimeout`
  - `SCHEDULED` -> Reset the timeout -> `SCHEDULED` if the timeoutMillis > `0` otherwise `DISABLED`
  - `DISABLED` -> Schedule a new timeout -> `SCHEDULED` if the timeoutMillis > `0` otherwise `DISABLED`
  - `TIMED_OUT` -> No operation
* `timeoutNow`
  - `!TIMED_OUT` -> Timed out -> `TIMEDOUT`
  - `TIMED_OUT` -> No operation
* `cancelTimeout`
  - `!TIMED_OUT` -> Cancel timeout -> `DISABLED`
  - `TIMED_OUT` -> No operation

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -24,27 +24,47 @@
 public interface TimeoutController {
 
     /**
-     * Initializes the timeout scheduler with the specified {@code timeoutMillis}.
+     * Schedules a new timeout with the specified {@code timeoutMillis}.
+     * If a timeout is scheduled already, this method will not start a new timeout.
+     *
+     * @param timeoutMillis a positive time amount value in milliseconds.
+     * @return {@code true} if the timeout is scheduled.
+     *         {@code false} if the timeout has been scheduled or the timeout has been triggered already.
```
trustin: has been scheduled or triggered already

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +84,166 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
      *
      * <p>Note that the {@link TimeoutTask} should be set via the {@link #setTimeoutTask(TimeoutTask)} or
      * the {@link #DefaultTimeoutController(TimeoutTask, EventLoop)} before calling this method.
+     *
+     * @return {@code true} if the timeout is scheduled.
+     *         {@code false} if the timeout has been scheduled, the timeout has been triggered already
+     *         or the {@link TimeoutTask} could not be scheduled now.
```
trustin: Perhaps we can remove this, because it's already in the super class Javadoc?

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +84,166 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Note that the {@link TimeoutTask} should be set via the {@link #setTimeoutTask(TimeoutTask)} or
+     * the {@link #DefaultTimeoutController(TimeoutTask, EventLoop)} before calling this method.
+     *
+     * @return {@code true} if the current timeout is extended by the specified {@code adjustmentMillis}.
+     *         {@code false} if no timeout was scheduled previously, the timeout has been triggered already
+     *         or the {@link TimeoutTask} could not be scheduled now.
```
trustin: Ditto - already in the super class Javadoc

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +84,166 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Note that the {@link TimeoutTask} should be set via the {@link #setTimeoutTask(TimeoutTask)} or
+     * the {@link #DefaultTimeoutController(TimeoutTask, EventLoop)} before calling this method.
+     *
+     * @return {@code true} if the current timeout is reset by the specified {@code newTimeoutMillis}.
+     *         {@code false} if the timeout has been triggered already or
+     *         the {@link TimeoutTask} could not be scheduled now.
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +84,166 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
+     * @return {@code true} if the current timeout is triggered successfully.
+     *         {@code false} if no timeout was scheduled previously, the timeout has been triggered already
+     *         or the {@link TimeoutTask} could not be scheduled now.
+     */
+    @Override
+    public boolean timeoutNow() {
+        checkState(timeoutTask != null,
+                   "setTimeoutTask(timeoutTask) is not called yet.");
+
+        if (State.SCHEDULED != state || !timeoutTask.canSchedule()) {
```
trustin: Global comment - `state != ... or state == ...` for readability

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +84,166 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
+        if (firstExecutionTimeNanos == 0) {
+            firstExecutionTimeNanos = lastExecutionTimeNanos = System.nanoTime();
+        }
+    }
+
+    @Nullable
+    private void invokeTimeoutTask() {
+        if (timeoutTask != null) {
+            timeoutTask.run();
+            state = State.TIMED_OUT;
```
trustin: Could set in `finally` block, just in case `run()` throws an exception

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +84,166 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
+        final long currentNanoTime = System.nanoTime();
+        final long passedTimeMillis = TimeUnit.NANOSECONDS.toMillis(currentNanoTime - lastExecutionTimeNanos);
+        final long newTimeoutMillis = LongMath.saturatedAdd(
+                LongMath.saturatedSubtract(timeoutMillis, passedTimeMillis), adjustmentMillis);
+        timeoutMillis = newTimeoutMillis;
+        lastExecutionTimeNanos = currentNanoTime;
+
+        if (newTimeoutMillis <= 0) {
+            invokeTimeoutTask();
+            state = State.TIMED_OUT;
```
trustin: This is redundant because `invokeTimeoutTask` updates `state`.

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +84,166 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
     @Override
     public boolean cancelTimeout() {
-        final ScheduledFuture<?> timeoutFuture = this.timeoutFuture;
-        if (timeoutFuture == null) {
+        if (State.TIMED_OUT == state) {
+            return false;
+        }
+        if (State.DISABLED == state) {
             return true;
         }
```
trustin: nit: Could use `switch-case`?

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -35,16 +36,23 @@
...
+
     @Nullable
     private TimeoutTask timeoutTask;
     private final EventLoop eventLoop;
 
     private long timeoutMillis;
-    private long firstStartTimeNanos;
-    private long lastStartTimeNanos;
+    private long firstExecutionTimeNanos;
+    private long lastExecutionTimeNanos;
```
trustin: `0` has a meaning in nano time. Need to introduce a boolean?

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
ikhoon: Sure! 😀

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
ikhoon: Nice tip, thanks!

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
ikhoon: Ahh... I forget it. 😱

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
ikhoon: The `TimeoutTask` is only used for `DefaultTimeoutController`.
I will add a more general expression to describe why it could return `false`. 😀

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -258,11 +261,13 @@ public void onError(Throwable cause) {
 
     @Override
     public void onComplete() {
-        if (!cancelTimeout() && reqCtx.requestTimeoutHandler() == null) {
+        if (isTimedOut() && reqCtx.requestTimeoutHandler() == null) {
```
ikhoon: The `cancelTimeout` returns `true` if the timeout is scheduled now.

core/src/main/java/com/linecorp/armeria/client/HttpResponseDecoder.java
```diff
@@ -294,14 +294,19 @@ private void cancelAction(@Nullable Throwable cause) {
...
+                cancelTimeout();
                 // There's no timeout or the response has not been timed out.
-                actionOnTimeoutCancelled.accept(cause);
+                actionOnNotTimedOut.accept(cause);
                 return;
             }
 
+            if (delegate.isOpen()) {
+                closeAction(cause);
+            }
```
ikhoon: I changed due to the changed `cancelTimeout` return value.

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -93,14 +94,15 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
+        if (state != State.DISABLED || !timeoutTask.canSchedule()) {
             return false;
         }
 
         cancelTimeout();
         this.timeoutMillis = timeoutMillis;
         final long nanoTime = System.nanoTime();
-        if (firstExecutionTimeNanos == 0) {
+        if (!isExecutedAtLeastOnce) {
+            isExecutedAtLeastOnce = true;
```
trustin: How about `scheduledAtLeastOnce`?

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -196,36 +190,60 @@ public boolean timeoutNow() {
...
-        if (cancelTimeout()) {
-            invokeTimeoutTask();
-            return true;
+        switch (state) {
+            case TIMED_OUT:
+                return false;
+            case DISABLED:
+                invokeTimeoutTask();
+                return true;
+            default: // state == State.SCHEDULED
```
trustin:
```
case SCHEDULED:
    ...

default:
    throw new Error();
```

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -69,11 +69,32 @@
...
      */
     boolean cancelTimeout();
 
+    /**
+     * Returns whether the timeout is scheduled or not.
+     *
+     * @return {@code true} the timeout is scheduled now.
+     *         {@code false} if the timeout has been triggered already or no timeout was scheduled previously.
+     */
+    boolean isScheduled();
```
trustin: It seems like this method is not used anywhere? Maybe OK to remove?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -95,6 +95,8 @@
     private Runnable responseTimeoutHandler;
     @Nullable
     private TimeoutController responseTimeoutController;
+    @Nullable
+    private Consumer<TimeoutController> pendingTimeoutTask;
```
trustin: Is there a chance a user calls timeout methods multiple times? Probably better chaining with `andThen()`

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
```diff
@@ -24,32 +26,79 @@
...
+     *
+     * @return {@code true} if the timeout has been triggered already.
+     *         {@code false} if the timeout is scheduled now or no timeout was scheduled previously.
+     */
+    boolean isTimedOut();
+
+    /**
+     * Returns {@code true} if no timeout was scheduled previously.
+     */
+    boolean isDisabled();
```
trustin: Seems like this is only used for testing.

I'm not sure, but if (`startTimeNanos()` is `null` if and only if `state` is `DISABLED`), then we could:

- Remove `isDisabled()` and use `startTimeNanos() == null` in the tests
- Replace `isExecutedAtLeastOnce` with `state != DISABLED`
  - We could set `firstExecutionTime` when the current state is `DISABLED`.

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
ikhoon: Yes, it don't use anywhere. I was considering to add this or not. I've already added other boolean getters for `TIMED_OUT` and `DISABLED`.
So I thought it would be more natural for adding boolean getters for `SCHEDULED`.
But it seems overkill. Let me remove the unused code. 😀

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
ikhoon: That is a good name. 😀

core/src/main/java/com/linecorp/armeria/common/util/TimeoutController.java
ikhoon: The `DISABLED` state can be reached even if `isExecutedAtLeastOnce = true` by calling `cancelTimeout()`
```java
TimeoutContoller tc = new DefaultTimeoutController() // isExecutedAtLeastOnce == false, DISABLED
tc.scheduleTimeout(1000) // isExecutedAtLeastOnce == true, SCHEDULED
// cancel timeout before the deadline.
tc.cancelTimeout() // isExecutedAtLeastOnce == true, DISABLED
```

By the way, I think we can make `startTimeNanos() == null` after calling `cancelTimeout()`.
It is a nicer approach.

ikhoon: The `responseTimeoutMillis` in `DefaultClientContext` is used to represent the amount of time since the start time of receiving the response.

```java
ctx.setResponseTimeoutMillis(1000)
// 1 seconds after
ctx.setResponseTimeoutAfterMillis(1000) 
// We need startTimeNanos() to get new 'responseTimeoutMillis' 
assert ctx.responseTimeoutMillis == System.nanoTime() - statTimeNanos() + responseTimeoutAfterMillis
```

I think we still need `isExecutedAtLeastOnce()` or add a new state like `WAITING` to express not never scheduled?

trustin: +1 for adding another state. Probably `INIT`?

ikhoon: Good, thought `INIT` too. 👍

core/src/test/java/com/linecorp/armeria/internal/DefaultTimeoutControllerTest.java
```diff
@@ -281,12 +282,12 @@ public boolean scheduleTimeout(long timeoutMillis) {
...
             if (result) {
                 // Previous: DISABLED
-                assertThat(prevState).isEqualTo(State.DISABLED);
+                assertThat(prevState == State.INIT || prevState == DISABLED).isTrue();
                 // Transition to: SCHEDULE
                 assertThat(delegate.state()).isEqualTo(State.SCHEDULED);
             } else {
                 // Previous: !DISABLED
-                assertThat(prevState).isNotEqualTo(State.DISABLED);
+                assertThat(prevState == State.INIT || prevState == DISABLED).isFalse();
```
trustin: `assertThat(prevState).isNotIn(State.INIT, State.DISABLED)`

core/src/test/java/com/linecorp/armeria/internal/DefaultTimeoutControllerTest.java
```diff
@@ -281,12 +282,12 @@ public boolean scheduleTimeout(long timeoutMillis) {
             final boolean result = delegate.scheduleTimeout(timeoutMillis);
             if (result) {
                 // Previous: DISABLED
-                assertThat(prevState).isEqualTo(State.DISABLED);
+                assertThat(prevState == State.INIT || prevState == DISABLED).isTrue();
```
trustin: `assertThat(prevState).isIn(State.INIT, State.DISABLED)`

core/src/test/java/com/linecorp/armeria/internal/DefaultTimeoutControllerTest.java
```diff
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.internal;
 
+import static com.linecorp.armeria.internal.DefaultTimeoutController.State.DISABLED;
```
trustin: Static-import all or none

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +85,184 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
-    public void initTimeout(long timeoutMillis) {
-        if (timeoutFuture != null || timeoutTask == null || !timeoutTask.canSchedule()) {
-            // No need to schedule a response timeout if:
-            // - the timeout has been scheduled already,
-            // - the timeout task is not set yet or
-            // - the timeout task cannot schedule yet or
-            return;
+    public boolean scheduleTimeout(long timeoutMillis) {
+        checkArgument(timeoutMillis > 0,
+                      "timeoutMillis: " + timeoutMillis + " (expected: > 0)");
```
trustin: Use `%s`. `+` will be done even if `IllegalArgumentException` is not raised.

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +85,184 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Note that the {@link TimeoutTask} should be set via the {@link #setTimeoutTask(TimeoutTask)} or
+     * the {@link #DefaultTimeoutController(TimeoutTask, EventLoop)} before calling this method.
+     *
+     * @return {@code true} if the current timeout is extended by the specified {@code adjustmentMillis}.
+     *         {@code false} if no timeout was scheduled previously, the timeout has been triggered already
+     *         or the {@link TimeoutTask} could not be scheduled now.
```
trustin: How about: `{@link TimeoutTask#canScheduled()} returned {@code false}.` to make it more direct and clear?

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +85,184 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Note that the {@link TimeoutTask} should be set via the {@link #setTimeoutTask(TimeoutTask)} or
+     * the {@link #DefaultTimeoutController(TimeoutTask, EventLoop)} before calling this method.
+     *
+     * @return {@code true} if the current timeout is reset by the specified {@code newTimeoutMillis}.
+     *         {@code false} if the timeout has been triggered already or
+     *         the {@link TimeoutTask} could not be scheduled now.
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +85,184 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
+        timeoutFuture = eventLoop.schedule(this::invokeTimeoutTask, newTimeoutMillis, TimeUnit.MILLISECONDS);
+        return true;
+    }
+
+    /**
+     * Trigger the current timeout immediately.
+     *
+     * @return {@code true} if the current timeout is triggered successfully.
+     *         {@code false} the timeout has been triggered already
+     *         or the {@link TimeoutTask} could not be scheduled now.
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -427,12 +446,22 @@ public void setRequestTimeoutAfter(Duration requestTimeout) {
     public void setRequestTimeoutAtMillis(long requestTimeoutAtMillis) {
         checkArgument(requestTimeoutAtMillis >= 0,
                       "requestTimeoutAtMillis: " + requestTimeoutAtMillis + " (expected: >= 0)");
```
trustin: `%s` please

core/src/main/java/com/linecorp/armeria/internal/DefaultTimeoutController.java
```diff
@@ -76,100 +85,184 @@ public void setTimeoutTask(TimeoutTask timeoutTask) {
...
+        final long currentNanoTime = System.nanoTime();
+        final long passedTimeMillis = TimeUnit.NANOSECONDS.toMillis(currentNanoTime - lastExecutionTimeNanos);
+        final long newTimeoutMillis = LongMath.saturatedAdd(
+                LongMath.saturatedSubtract(timeoutMillis, passedTimeMillis), adjustmentMillis);
+        timeoutMillis = newTimeoutMillis;
+        lastExecutionTimeNanos = currentNanoTime;
+
+        if (newTimeoutMillis <= 0) {
+            invokeTimeoutTask();
+            return false;
```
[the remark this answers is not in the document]
ikhoon: Ah, that makes sense! It seems better to return `true`.

PR #2392, opened 2020-01-10
Send periodic `102 Processing` headers for long-polling health…

### LINKED ISSUE #2386: Long-polling health check cannot detect unresponsive servers
With long-polling health check, it is currently not possible to distinguish unresponsive servers who just accepts sockets doing nothing. `HealthCheckService` must send some informational headers periodically, such as `102 Processing`.

For compatibility with the old long-polling `HealthCheckedEndpointGroup`, we're gonna have to update the format of the `armeria-lphc` header value.

trustin:
Motivation:

With long-polling health check, it is currently not possible to
distinguish unresponsive servers who just accepts sockets doing nothing.
`HealthCheckService` must send some informational headers periodically,
such as `102 Processing`.

Modifications:

- Change the format of the `armeria-lphc` header value to contain two
  values:
  - `maxLongPollingTimeoutSeconds`
  - `pingIntervalSeconds`
- Change the behavior of `HealthCheckService` to send `102 Processing`
  informational headers every `pingIntervalSeconds`.
  - Updated `HealthCheckServiceBuilder.longPolling()`
- Change the behavior of `HealthCheckedEndpointGroup` to mark an
  `Endpoint` as unhealthy when not receiving a ping for last
  `pingIntervalSeconds * 2` seconds.
- Miscellaneous:
  - Removed `Http` from the test classes that test `HealthCheckEndpointGroup`

Result:

- Fixes #2386
- `HealthCheckedEndpointGroup` now marks an unresponsive endpoints as
  soon as possible when long-polling is enabled.
- The old Armeria clients will automatically fall back to traditional
  periodic health check, because it does not understand the new header
  format.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
@@ -22,25 +22,42 @@ import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
 
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.math.LongMath;
+
 import com.linecorp.armeria.client.ClientOptions;
 import com.linecorp.armeria.client.ClientRequestContext;
+import com.linecorp.armeria.client.ClientRequestContextCaptor;
+import com.linecorp.armeria.client.Clients;
 import com.linecorp.armeria.client.Endpoint;
 import com.linecorp.armeria.client.HttpClient;
+import com.linecorp.armeria.client.ResponseTimeoutException;
 import com.linecorp.armeria.client.SimpleDecoratingHttpClient;
 import com.linecorp.armeria.client.WebClient;
-import com.linecorp.armeria.common.AggregatedHttpResponse;
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpObject;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.HttpStatusClass;
 import com.linecorp.armeria.common.RequestHeaders;
 import com.linecorp.armeria.common.RequestHeadersBuilder;
+import com.linecorp.armeria.common.ResponseHeaders;
+import com.linecorp.armeria.common.stream.SubscriptionOption;
 import com.linecorp.armeria.common.util.AsyncCloseable;
 
 import io.netty.util.AsciiString;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.concurrent.ScheduledFuture;
 
 final class HttpHealthChecker implements AsyncCloseable {
+
+    private static final Logger logger = LoggerFactory.getLogger(HttpHealthChecker.class);
 
     private static final AsciiString ARMERIA_LPHC = HttpHeaderNames.of("armeria-lphc");
 
@@ -51,6 +68,7 @@ final class HttpHealthChecker implements AsyncCloseable {
     private final boolean useGet;
     private boolean wasHealthy;
     private long maxLongPollingSeconds;
+    private long pingIntervalSeconds;
     @Nullable
     private HttpResponse lastResponse;
     private boolean closed;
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -89,35 +107,188 @@ private synchronized void check() {
...
+        private void updateLongPollingSettings(ResponseHeaders headers) {
+            final String longPollingSettings = headers.get(ARMERIA_LPHC);
+            if (longPollingSettings == null) {
                 maxLongPollingSeconds = 0;
+                pingIntervalSeconds = 0;
+                return;
             }
 
+            final int commaPos = longPollingSettings.indexOf(',');
+            long maxLongPollingSeconds = 0;
```
anuraaga: Don't think these need to be `long`, especially the ping interval

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -89,35 +107,188 @@ private synchronized void check() {
...
+    private final class ResponseTimeoutUpdater extends SimpleDecoratingHttpClient {
+        ResponseTimeoutUpdater(HttpClient delegate) {
+            super(delegate);
+        }
+
+        @Override
+        public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Exception {
+            if (maxLongPollingSeconds > 0) {
+                final long responseTimeoutMillis = ctx.responseTimeoutMillis();
+                if (responseTimeoutMillis > 0) {
```
anuraaga: Should extend itself be a no-op if the timeout is 0?

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -89,35 +107,188 @@ private synchronized void check() {
...
+
+    @Override
+    public synchronized CompletableFuture<?> closeAsync() {
+        if (lastResponse != null) {
+            if (!closed) {
+                closed = true;
+                lastResponse.abort();
+            }
+            return lastResponse.completionFuture().handle((unused1, unused2) -> null);
+        } else {
```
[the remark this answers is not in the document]
trustin: Because it's a bug? :laughing:

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
trustin: Good point. No need for `if`.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
trustin: All changed to `int`

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -89,35 +107,188 @@ private synchronized void check() {
...
+            if (pingCheckFuture != null) {
+                return;
+            }
+
+            final long pingIntervalSeconds = HttpHealthChecker.this.pingIntervalSeconds;
+            final long pingTimeoutNanos = LongMath.saturatedMultiply(
+                    TimeUnit.SECONDS.toNanos(pingIntervalSeconds), 2);
+            if (pingIntervalSeconds <= 0) {
+                return;
+            }
```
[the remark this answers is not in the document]
trustin: Indeed.

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -89,35 +107,189 @@ private synchronized void check() {
...
                     default:
-                        if (res.status() == HttpStatus.NOT_MODIFIED) {
-                            maxLongPollingSeconds = getMaxLongPollingSeconds(res);
+                        if (headers.status() == HttpStatus.NOT_MODIFIED) {
                             isHealthy = wasHealthy;
+                            receivedExpectedResponse = true;
                         } else {
                             // Do not use long polling on an unexpected status for safety.
                             maxLongPollingSeconds = 0;
+                            logger.warn("{} Unexpected health check response: {}", reqCtx, headers);
```
anuraaga: Just noticed this is probably for #1913.
Do you think we can special case 4XX with something like "Received 4XX health check response. This generally indicates a misconfiguration of the client. Did you happen to forget to configure the endpoint group option with those of the client?"

trustin:
@anuraaga Ah yeah, sure. Let me send a follow up PR soon.

Issue #2394, opened 2020-01-13
Add advanced throttling mechanism based on Token-Bucket algorithm


max904-github:
The [Token Bucket](https://en.wikipedia.org/wiki/Token_bucket) is an algorithm used in packet switched computer networks and telecommunications networks. It can be used to check that data transmissions, in the form of packets, conform to defined limits on bandwidth and burstiness (a measure of the unevenness or variations in the traffic flow).
I have developed a server-side throttling integration based on Token Bucket algorithm and [Bucket4j](https://github.com/vladimir-bukhtoyarov/bucket4j) library. This is more advanced throttling facility comparing to Guava's basic `RateLimiter`. It can even be extended to support distributed throttling across multiple microservices. The library supports asynchronous API and `CompletableFuture`.

PR #2395, opened 2020-01-13
Implemented new advanced throttling based on Bucket4j (a rate-limitin…

max904-github:
…g library based on token-bucket algorithm) library:

- added new dependency for Bucket4j library to the core project.
- added new package com.linecorp.armeria.server.throttling.tokenbucket.
- added two new classes RetryThrottlingStrategy and RetryThrottlingService to com.linecorp.armeria.server.throttling to be able to customize HTTP response status and specify Retry-After header.
- added BandwidthLimit, TokenBucketConfig and TokenBucketConfigBuilder to handle TokenBucket algorithm configuration.
- implemented TokenBucketThrottlingStrategy extending RetryThrottlingStrategy to complete the feature.
- added junit tests to cover the work.

This is made in the context of Issue #2394:
"Add advanced throttling mechanism based on Token-Bucket algorithm"

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/BandwidthLimit.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/BandwidthLimit.java
@@ -37,16 +37,6 @@ public class BandwidthLimit {
     @Nonnull
     private final Duration period;
 
-    /**
-     * Specifies limitation in
-     *   <a href="https://github.com/vladimir-bukhtoyarov/bucket4j/blob/1.3/doc-pages/token-bucket-brief-overview.md#token-bucket-algorithm">classic interpretation</a>
-     *   of token-bucket algorithm.
-     * @param limit the bucket size - defines the count of tokens which can be held by bucket
-     *              and defines the speed at which tokens are regenerated in bucket
-     * @param overdraftLimit defines the maximum overdraft count of tokens which can be held by bucket,
-     *                       this must exceed the {@code limit}
-     * @param period the time window, during which the tokens will be regenerated
-     */
     BandwidthLimit(long limit, long overdraftLimit, @Nonnull Duration period) {
         // validate limit
         if (limit <= 0L) {
@@ -69,13 +59,33 @@ public class BandwidthLimit {
     }
 
     /**
+     * Creates new {@link BandwidthLimit}.
+     * Specifies limitation in
+     *   <a href="https://github.com/vladimir-bukhtoyarov/bucket4j/blob/1.3/doc-pages/token-bucket-brief-overview.md#token-bucket-algorithm">classic interpretation</a>
+     *   of token-bucket algorithm.
+     * @param limit the bucket size - defines the count of tokens which can be held by bucket
+     *              and defines the speed at which tokens are regenerated in bucket
+     * @param overdraftLimit defines the maximum overdraft count of tokens which can be held by bucket,
+     *                       this must exceed the {@code limit}
+     * @param period the time window, during which the tokens will be regenerated
+     */
+    public static BandwidthLimit of(long limit, long overdraftLimit, @Nonnull Duration period) {
+        return new BandwidthLimit(limit, overdraftLimit, period);
+    }
+
+    BandwidthLimit(long limit, @Nonnull Duration period) {
+        this(limit, 0, period);
+    }
+
+    /**
+     * Creates new {@link BandwidthLimit}.
      * Specifies easy limitation of {@code limit} tokens per {@code period} time window.
      * @param limit the bucket size - defines the maximum count of tokens which can be held by bucket
      *              and defines the speed at which tokens are regenerated in bucket
      * @param period the time window, during which the tokens will be regenerated
      */
-    BandwidthLimit(long limit, @Nonnull Duration period) {
-        this(limit, 0, period);
+    public static BandwidthLimit of(long limit, @Nonnull Duration period) {
+        return new BandwidthLimit(limit, period);
     }
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/BandwidthLimit.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/BandwidthLimit.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2020 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.server.throttling.tokenbucket;
+
+import static java.util.Objects.requireNonNull;
+
+import java.time.Duration;
+
+import javax.annotation.Nonnull;
+
+import com.google.common.base.MoreObjects;
+
+import io.github.bucket4j.Bandwidth;
+import io.github.bucket4j.Refill;
+
+/**
+ * Stores configurations of a Token-Bucket bandwidth limit.
+ */
+public class BandwidthLimit {
+
+    private final long limit;
+    private final long overdraftLimit;
+    @Nonnull
+    private final Duration period;
+
+    /**
+     * Specifies limitation in
+     *   <a href="https://github.com/vladimir-bukhtoyarov/bucket4j/blob/1.3/doc-pages/token-bucket-brief-overview.md#token-bucket-algorithm">classic interpretation</a>
+     *   of token-bucket algorithm.
+     * @param limit the bucket size - defines the count of tokens which can be held by bucket
+     *              and defines the speed at which tokens are regenerated in bucket
+     * @param overdraftLimit defines the maximum overdraft count of tokens which can be held by bucket,
+     *                       this must exceed the {@code limit}
+     * @param period the time window, during which the tokens will be regenerated
+     */
+    BandwidthLimit(long limit, long overdraftLimit, @Nonnull Duration period) {
+        // validate limit
+        if (limit <= 0L) {
+            throw new IllegalArgumentException("Bandwidth Limit must be positive");
+        }
+        this.limit = limit;
+
+        // validate overdraftLimit
+        if (overdraftLimit > 0L && overdraftLimit <= limit) {
+            throw new IllegalArgumentException("Overdraft Limit has to exceed Bandwidth Limit");
+        }
+        this.overdraftLimit = overdraftLimit;
+
+        requireNonNull(period, "period");
+        if (period.isNegative() || period.isZero()) {
+            throw new IllegalArgumentException(
+                    "period: " + period + " (expected: > 0)");
+        }
+        this.period = period;
+    }
+
+    /**
+     * Specifies easy limitation of {@code limit} tokens per {@code period} time window.
+     * @param limit the bucket size - defines the maximum count of tokens which can be held by bucket
+     *              and defines the speed at which tokens are regenerated in bucket
+     * @param period the time window, during which the tokens will be regenerated
+     */
+    BandwidthLimit(long limit, @Nonnull Duration period) {
+        this(limit, 0, period);
+    }
+
+    /**
+     * The bucket size - defines the count of tokens which can be held by bucket
+     * and defines the speed at which tokens are regenerated in bucket.
+     * @return Bucket size.
+     */
+    public long limit() {
+        return limit;
+    }
+
+    /**
+     * The maximum overdraft count of tokens which can be held by bucket,
+     * this must exceed the {@code BandwidthLimit#limit()}.
+     * @return Bucket maximum overdraft count.
+     */
+    public long overdraftLimit() {
+        return overdraftLimit;
+    }
+
+    /**
+     * The time window, during which the tokens will be regenerated.
+     * @return Time window for the limit.
+     */
+    @Nonnull
+    public Duration period() {
+        return period;
+    }
+
+    @Nonnull
+    Bandwidth bandwidth() {
+        if (overdraftLimit > limit) {
+            // overdraft has been defined
+            return Bandwidth.classic(overdraftLimit, Refill.greedy(limit, period));
+        } else {
+            return Bandwidth.simple(limit, period);
+        }
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects
+                .toStringHelper(this)
+                .add("limit", limit)
+                .add("overdraftLimit", overdraftLimit)
+                .add("period", period)
+                .toString();
+    }
+}
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketConfig.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketConfig.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2020 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.server.throttling.tokenbucket;
+
+import static java.util.Objects.requireNonNull;
+
+import java.time.Duration;
+
+import javax.annotation.Nonnull;
+
+import com.google.common.base.MoreObjects;
+
+import com.linecorp.armeria.common.HttpHeaderNames;
+import com.linecorp.armeria.common.HttpStatus;
+
+/**
+ * Stores configuration of the Token-Bucket algorithm.
+ */
+public class TokenBucketConfig {
+
+    @Nonnull
+    private final BandwidthLimit[] limits;
+    @Nonnull
+    private final Duration retryAfterTimeout;
+
+    /**
+     * Defines throttling configuration comprised of one or more bandwidth limits in accordance to
+     * token-bucket algorithm.
+     *
+     * <h3>Multiple bandwidths:</h3>
+     * It is possible to specify more than one bandwidth per bucket,
+     * and bucket will handle all bandwidth in strongly atomic way.
+     * Strongly atomic means that token will be consumed from all bandwidth or from nothing,
+     * in other words any token can not be partially consumed.
+     * <br> Example of multiple bandwidth:
+     * <pre>{@code
+     * // Adds bandwidth that restricts to consume
+     * // not often than 1000 tokens per 1 minute and
+     * // not often than 100 tokens per second.
+     * TokenBucketConfig config = TokenBucketConfig.builder()
+     *      .limit(1000L, Duration.ofMinutes(1))
+     *      .limit(100L, Duration.ofSeconds(1))
+     *      .retryAfterTimeout(Duration.ofSeconds(10L))
+     *      .build()
+     * }</pre>
+     *
+     * @param retryAfterTimeout the client timeout to retry the attempt
+     * @param limits one or more bandwidth limits to be used by token-bucket algorithm
+     */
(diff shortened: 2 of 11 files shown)


max904-github:
> 
> 
> [CLA assistant check](https://cla-assistant.io/line/armeria?pullRequest=2395)
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/line/armeria?pullRequest=2395) before we can accept your contribution.
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/line/armeria?pullRequest=2395) it.

Signed

trustin:
@max904-github Seems like CLA assistant did not recognize the e-mail addresses in your commits. Could you double-check your e-mail addresses are in your GitHub settings?

trustin:
Some ideas:

Are token bucket algorithm and `retry-after` inseparable from each other? Is it possible for a user to use token buckets but not `retry-after`? Then perhaps we should make `TokenBucketThrottlingStrategy` not extend `RetryThrottlingStrategy`.

How about employing the builder pattern for `TokenBucketThrottlingStrategy` rather than `TokenBuckerConfig`? e.g.

```java
// Before
new TokenBucketThrottlingStrategy(TokenBucketConfig.builder()...build());
// After:
TokenBucketThrottlingStrategy.builder().....build();
```

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketThrottlingStrategy.java
```diff
@@ -0,0 +1,116 @@
...
+        // reconfigure the bucket
+        return bucket.replaceConfiguration(builder.build());
+    }
+
+    /**
+     * Registers a request with the bucket.
+     */
+    @Override
+    public CompletionStage<Boolean> accept(ServiceRequestContext ctx, T request) {
+        return bucket.tryConsume(1L);
```
anuraaga: FWIU, token bucket is designed for managing connections with varying bandwidth. Can you describe the advantages of this over Guava's algorithm when using a fixed bandwidth (`1`).

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/BandwidthLimit.java
```diff
@@ -0,0 +1,127 @@
...
+import io.github.bucket4j.Bandwidth;
+import io.github.bucket4j.Refill;
+
+/**
+ * Stores configurations of a Token-Bucket bandwidth limit.
+ */
+public class BandwidthLimit {
+
+    private final long limit;
+    private final long overdraftLimit;
```
adriancole: do you want to start with overdraft support? If not, you can probably vendor the dependency as bucket limiting implementation can maybe be shared from other code if not for this.

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/BandwidthLimit.java
```diff
@@ -0,0 +1,127 @@
...
+        this.period = period;
+    }
+
+    /**
+     * Specifies easy limitation of {@code limit} tokens per {@code period} time window.
+     * @param limit the bucket size - defines the maximum count of tokens which can be held by bucket
+     *              and defines the speed at which tokens are regenerated in bucket
+     * @param period the time window, during which the tokens will be regenerated
+     */
+    BandwidthLimit(long limit, @Nonnull Duration period) {
```
adriancole: I might suggest something a bit more constrained, like seconds. That way, excessively small periods needn't be handled.

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketConfig.java
```diff
@@ -0,0 +1,109 @@
...
+     * and bucket will handle all bandwidth in strongly atomic way.
+     * Strongly atomic means that token will be consumed from all bandwidth or from nothing,
+     * in other words any token can not be partially consumed.
+     * <br> Example of multiple bandwidth:
+     * <pre>{@code
+     * // Adds bandwidth that restricts to consume
+     * // not often than 1000 tokens per 1 minute and
+     * // not often than 100 tokens per second.
+     * TokenBucketConfig config = TokenBucketConfig.builder()
+     *      .limit(1000L, Duration.ofMinutes(1))
```
adriancole: similarly, I wonder if we can't start with a simple single limit per config which makes the implementation simpler and less coupled to a specific library

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketConfig.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * When the bucket limits reached, the incoming requests will be replied
+     * with {@link HttpStatus#TOO_MANY_REQUESTS} that carries {@link HttpHeaderNames#RETRY_AFTER} with
+     * the value returned by this method.
+     * @return A {@link Duration} after which the client may retry again.
+     */
+    @Nonnull
+    public Duration retryAfterTimeout() {
```
adriancole: agree with something I think said before, which is maybe retryAfter can be compositional

max904-github:
> Are token bucket algorithm and `retry-after` inseparable from each other? Is it possible for a user to use token buckets but not `retry-after`? Then perhaps we should make `TokenBucketThrottlingStrategy` not extend `RetryThrottlingStrategy`.
> 

Token Bucket and Retry-After are not inseparable. In fact Retry-After is generic enough to be part of general ThrottlingService/Strategy. I have introduced RetryThrottlingStrategy in order to avoid changes to base ThrottlingStrategy.

> How about employing the builder pattern for `TokenBucketThrottlingStrategy` rather than `TokenBuckerConfig`? e.g.
> 
> ```java
> // Before
> new TokenBucketThrottlingStrategy(TokenBucketConfig.builder()...build());
> // After:
> TokenBucketThrottlingStrategy.builder().....build();
> ```

I was following the implementation of ThrottlingService/Strategy and RateLimitingThrottlingStrategy which do not employ a builder pattern for the Strategy.
But If you think it's a better approach, I can change it.

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketThrottlingStrategy.java
max904-github: I don't have the details of Guava's algorithm. In their documentation they just say it's "simple". I would say it's too simple to match the needs of our systems. On the contrary, [Token Bucket](https://en.wikipedia.org/wiki/Token_bucket) is an established algorithm serving wide variety use cases. Token Bucket and Leaky Bucket algorithms are standard to the industry I work for (telco) primarily used for traffic shaping and congestion control. There are multiple implementations available. I've picked the most popular one for JVM.
Now, comparing the rate limiting implementations of Guava and Bucket4j. Here are some highlights:
- Precision: `Bucket4j.builder().withNanosecondPrecision()` - allows to go up to nanosecond precision.
- Refill strategy: `Refill.intervally(10, Duration.ofSeconds(1))` - it's possible to customize refill strategy. The refill can be "greedy" (asap), interval based or take place only at reaching the end of the interval.
- Multiple limits: This is invaluable to be able to control the bursts of traffic.
- Returning tokens back to bucket: `bucket.addTokens(50)` - tokens could be returned to the bucket.
- Bucket configuration can change on the fly: `bucket.replaceConfiguration(newConfiguration)` - so it's possible to implement several traffic shaping schemes, weekday vs weekend, etc.
- Bucket can be distributed when configured to use distributed cache like Apache Ignite. This way it's possible to do throttling across the cluster.
- Bucket4j provides asynchronous API: compare `return bucket.tryConsume(1L)` with return `completedFuture(rateLimiter.tryAcquire())` from `RateLimitingThrottlingStrategy`.
- Bucket4j supports monitoring via bucket events.

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/BandwidthLimit.java
max904-github: Overdraft is very important for our use cases. There are other important things which I did not (yet) add, like refill mode.

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/BandwidthLimit.java
max904-github: Fine precision is one of the great advantages of this implementation!

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketConfig.java
max904-github: Single limit is too simple. If the goal is to stay within a single limit then integration with the Token Bucket is quite useless indeed.

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketConfig.java
max904-github: Could you please clarify?
[Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) is a very important mechanism of throttling which Armeria's ThrottlingService/Strategy have not utilized yet. In fact, it should be consistent with client-side circuit breaker and retry policies. E.g. the client-side circuit breaker should react to this header and open the breaker for the specified period of time automatically. Similarly it could be done for the retry.
It's indeed unrelated to Token-Bucket, but it's required for many use cases I have to deal with.

adriancole: I think in general this issue is more advocating on the premise that everything in bucket4j is relevant here vs mentioning which things you need and how you decide to perform things. for example, retry after time implies you have a sensible value to return. this could be throttle specific or otherwise, and how to calculate that value could be parameterized also. while it may be that the team merge your requests without asking how to simplify, how it is used, how less could be viable, it isnt more than just normal analysis to ask these sort of questions. concretely here.. maybe you can provide some rationale for why it is always relevant a time based back off value? you can choose not to answer of course. someone asked me to look at this issue and I did as a favor to a friend. personally I have no dog in this race.

adriancole:
my final drive by comment here would be to explicitly package/name this based on the library in use. as far as I can gather the intent is probably more about bucket4j specifically vs generic decoupled throttling. in such case sooner or later we would need to package accordingly.

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketConfig.java
max904-github: I appreciate your comments. I did not try to undermine your feedback in any way. I just tried to clarify what you meant as "compositional".
Regarding the time-based backoff. I did not say it's absolutely always relevant. The implementation I proposed adds this as an additional layer over base ThrottlingService/Strategy. The user may chose to use this layer or not. Besides, the backoff is optional:
```
        if (retryAfterSeconds != null) {
            return ResponseHeaders.of(status, HttpHeaderNames.RETRY_AFTER, retryAfterSeconds);
        } else {
            return ResponseHeaders.of(status);
        }
```
I also gave you example of Armeria's own CircuitBreaker and Retry implementations where it makes a lot of sense to coordinate better with the server side. If the server does not inform of any particular backoff period and the client relies on its own disconnected configuration than the throttling goals may not be achieved by the end-to-end system.
Now regarding calculating the backoff period dynamically. The API I proposed in RetryThrottlingStrategy allows doing that. The value can change dynamically or even become NULL at some point. Although, my current implementation of TokenBucketConfig does not allow such dynamism. I could use some kind of Supplier for that, but at this point it may just overcomplicate things.

max904-github:
> my final drive by comment here would be to explicitly package/name this based on the library in use. as far as I can gather the intent is probably more about bucket4j specifically vs generic decoupled throttling. in such case sooner or later we would need to package accordingly.

I'm following the other examples I find in Armeria:
RateLimitingThrottlingStrategy - is not GuavaRateLimiterThrottlingStrategy,
MetricCollectingService - is not MicrometerCollectionService,
etc.

trustin:
I think we can shade bucket4j once we hide its classes from the public API, like we did for Caffeine. Then, perhaps we don't need to worry much about naming :-) I might have missed some, but the proposed public API seemed to have no direct references to the classes in bucket4j. Correct?

max904-github:
> 
> 
> I think we can shade bucket4j once we hide its classes from the public API, like we did for Caffeine. Then, perhaps we don't need to worry much about naming :-) I might have missed some, but the proposed public API seemed to have no direct references to the classes in bucket4j. Correct?

Correct

trustin:
> Token Bucket and Retry-After are not inseparable. In fact Retry-After is generic enough to be part of general ThrottlingService/Strategy. I have introduced RetryThrottlingStrategy in order to avoid changes to base ThrottlingStrategy.
>
> From my other response...
> [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) is a very important mechanism of throttling which Armeria's ThrottlingService/Strategy have not utilized yet. In fact, it should be consistent with client-side circuit breaker and retry policies. E.g. the client-side circuit breaker should react to this header and open the breaker for the specified period of time automatically. Similarly it could be done for the retry.

That sounds nice. I like the idea of adding `Retry-After` header when a request is throttled. Do you think it will be more useful to modify `ThrottlingStrategy.accept()` to return `CompletionStage<Duration>` (instead of `CompletionStaget<Boolean>` for more accurate control? (and this will make `RetryThrottlingStrategy` unnecessary I guess?)

> Actually, the reason why TokenBucketConfig had to be exposed is the ability to change the config on the strategy TokenBucketThrottlingStrategy:
>
>     public CompletableFuture<Void> replaceConfiguration(@Nonnull TokenBucketConfig config)

Ah, then it's fine as it is.

trustin:
> Correct

Looking from the bucket4j documentation, it seems like it has some advanced features like clustered rate-limiting. If there's any chance we will have to expose Bucket4j's classes in our public API in the future, then perhaps it's better adding a separate module like `armeria-bucket4j`?

We could also:

- Provide an essential version (like what's in this PR) which does not expose any Bucket4j API in Armeria core, and shaded bucket4j.
- Provide a separate module `armeria-bucket4j` which provides an alternative `ThrottlingStrategy` implementation that exposes bucket4j classes in the public API. (Maybe not the scope of this PR)

What do you think?

trustin:
> `CompletionStage<Duration>`

Not as simple as this but we will have to define some result type which is probably a tuple of `(Boolean, Duration)`, where the duration is always null when the boolean is true.

max904-github:
> 
> 
> > Correct
> 
> Looking from the bucket4j documentation, it seems like it has some advanced features like clustered rate-limiting. If there's any chance we will have to expose Bucket4j's classes in our public API in the future, then perhaps it's better adding a separate module like `armeria-bucket4j`?
> 
> We could also:
> 
>     * Provide an essential version (like what's in this PR) which does not expose any Bucket4j API in Armeria core, and shaded bucket4j.
> 
>     * Provide a separate module `armeria-bucket4j` which provides an alternative `ThrottlingStrategy` implementation that exposes bucket4j classes in the public API. (Maybe not the scope of this PR)
> 
> 
> What do you think?

As the PR shades bucket4j, I think it's safe to have only essential version for now. It's functionality could be expanded later when needed up to some level. If someone requires an advanced set of Bucket4j features they better of implementing their own Throttling strategy (like we do currently, thanks to Armeria open architecture).
The reason I proposed this enhancement is that the current RateLimiting implementation is a way too limited even for reasonably complex use cases.

max904-github:
> 
> 
> > `CompletionStage<Duration>`
> 
> Not as simple as this but we will have to define some result type which is probably a tuple of `(Boolean, Duration)`, where the duration is always null when the boolean is true.

It's probaly a good idea, but it seems a much bigger refactoring of the current code than I envisioned.
If this is to be done, I suggest to take Retry-After as separate issue and deal with it first, including all the refactoring.
Also, It would make sense to create connected issues for the Circuit Breaker and Retry to handle Retry-After header.

anuraaga:
> The reason I proposed this enhancement is that the current RateLimiting implementation is a way too limited even for reasonably complex use cases.

Can you describe more some of the use cases that benefit from this over a simple implementation?

I understand the application of this sort of complex rate limiting for packet routing where sizes vary and latency is low, but Armeria is a server framework. It is almost always unary RPC, with businesses logic that in practice should end up with milliseconds of latency. If it was streams, the rate limiting doesn't apply anyways and HTTP/2 flow control is probably more apt. So I'm concerned this will add confusion if added to core, especially with claims of existing approaches being way too limited. Better would be to have real examples to teach users when they should use the approach over others, then we can add them to documentation.

+1 to separate artifact personally but not strong if we can provide clear documentation of real-world use cases that indicate core users would want to use this

max904-github:
> 
> 
> > The reason I proposed this enhancement is that the current RateLimiting implementation is a way too limited even for reasonably complex use cases.
> 
> Can you describe more some of the use cases that benefit from this over a simple implementation?
> 
> I understand the application of this sort of complex rate limiting for packet routing where sizes vary and latency is low, but Armeria is a server framework. It is almost always unary RPC, with businesses logic that in practice should end up with milliseconds of latency. If it was streams, the rate limiting doesn't apply anyways and HTTP/2 flow control is probably more apt.

With the introduction of 5G most of Telco specific protocols like Diameter are being replaced with HTTP/2 as part of [5G SBA](https://www.etsi.org/deliver/etsi_ts/123500_123599/123501/15.04.00_60/ts_123501v150400p.pdf). Those protocols have very strict latency and congestion control requirements which have to be implemented on top of HTTP/2.

> So I'm concerned this will add confusion if added to core, especially with claims of existing approaches being way too limited. Better would be to have real examples to teach users when they should use the approach over others, then we can add them to documentation.
> 

If you decide that Guava based throttling is sufficient, it's fine by me. I already have even more complex throttling implemented on top of Armeria. This PR is my attempt to create something simple enough but still useful and I hope we can find the right way to do it.
I don't have an intention to confuse anybody. But I know a lot of "users" in my industry who would benefit from having more complex throttling mechanism integrated directly into the framework.

> +1 to separate artifact personally but not strong if we can provide clear documentation of real-world use cases that indicate core users would want to use this
>

Explaining my use cases is going to be confusing to the user, but I'm sure we can find some reasonable example for a sake of documentation.

max904-github:
Some additional piece of information on the subject: [RateLimit Header Fields for HTTP - draft](https://tools.ietf.org/id/draft-polli-ratelimit-headers-00.html). Perhaps, in addition to standard `Retry-After` header, we may want to use additional `X-RateLimit-Limit`, `X-RateLimit-Remaining`, `X-RateLimit-Reset` as more fine-grained throttling mechanism, at least until the new standard gets passed.

max904-github:
Example: Rate limiting capability exposed by Bitbucket online services [Improving instance stability with rate limiting](https://confluence.atlassian.com/bitbucketserver/improving-instance-stability-with-rate-limiting-976171954.html) based on Token Bucket.
When Rate limiting enabled, they seem to use optional X-headers with OK responses:
```
HTTP/1.1 200 OK
X-RateLimit-Limit: 5000
X-RateLimit-Remaining: 4999
X-RateLimit-Reset: 1366037820
```

core/src/main/java/com/linecorp/armeria/server/throttling/tokenbucket/TokenBucketThrottlingStrategy.java
anuraaga: Thanks for the extra links. It made me think that we might need a `BucketSelector`? It seems like this is most effective when each client has it's own bucket, to prevent one bad client from affecting others. With a global bucket, it seems less clear what a client should do if returned the rate limit information like the bitbucket example since other clients will also be using the bucket.

I guess we could add a client decorator too that attaches a client id (uuid I guess) and one bucket selector would be to use the header. Or for HTTP/2, a reasonable one would be bucket per connection. Though that seems to overlap with HTTP/2 flow control so maybe not as useful.

trustin:
(Sorry for a late reply!) Thanks for your investigation, @max904-github. We are currently focusing on releasing 1.0, our first release with stable API, so we are somewhat concerned about adding a new feature to the core module. What do you think about adding `armeria-contrib` module and putting your work there until it's stabilized enough?

trustin:
I didn't mean that this pull request has to be closed. How about just creating a new module and moving the classes there? We could mark the package as unstable until we have more experience and then move it into the core module when we do a minor version bump.

max904-github:
@trustin I understand that bucket4j integration does not have to be added the core. Although, it's not worth having it in a separate bundle either.
What I strongly suggest to change in the core is adding an ability to supply custom headers like `Retry-After` or/and `X-RateLimit-***`. This behavior should be provided as part of the throttling strategy implementation. Moreover, as per [RateLimit Header Fields for HTTP - draft](https://tools.ietf.org/id/draft-polli-ratelimit-headers-00.html), the headers might need to be provided even for the positive unthrottled responses.
This core modification should allow adding bucket4j or other throttling implementations with just a single class implementing such as custom throttling strategy. Otherwise, people will be forced messing with a lot of internals or end up re-implementing whole throttling part of the framework.
Unfortunately, my PR does not solve this problem of the core, or actually solves it partially in not so generic fashion. I suggest to create a new Issue for that and address it in the core.
Do you want me to create such new issue?

PR #2396, opened 2020-01-14
Make `ClientFactory` a part of `ClientOptions`

### LINKED ISSUE #1913: Disparity between HttpHealthCheckedEndpoint and HttpClient can be confusing.

### LINKED ISSUE #2384: Make `ClientFactory` a part of `ClientOptions`.
When creating a new client derived from an existing client, a user can write the following code:

```java
WebClient client = WebClient.builder(...)...build();
WebClient derived = WebClient.builder(client.scheme().sessionProtocol(),
                                      client.endpointGroup())
                             .path(client.absolutePathRef())
                             .factory(client.factory())
                             .options(client.options())
                             .build();
```

Perhaps we could make `ClientFactory` a part of `ClientOptions`, so that a user doesn't have to specify it manually?

I don't think we have do this for `path` though, because it's part of the endpoint URI.

trustin:
Motivation:

When creating a new client derived from an existing client, a user can
write the following code:

```java
WebClient client = WebClient.builder(...)...build();
WebClient derived =
    WebClient.builder(client.scheme().sessionProtocol(),
                      client.endpointGroup())
             .path(client.absolutePathRef())
             .factory(client.factory())
             .options(client.options())
             .build();
```

We could make `ClientFactory` a part of `ClientOptions`, so that a user
doesn't have to specify it manually:

```java
WebClient derived =
    WebClient.builder(client.scheme().sessionProtocol(),
                      client.endpointGroup())
             .path(client.absolutePathRef())
             .options(client.options())
             .build();
```

Modifications:

- Add `ClientOption.FACTORY`.
- Add `AbstractClientOptionsBuilder.factory(ClientFactory)`
  - Remove the `factory(ClientFactory)` methods in the subclasses
- (Breaking) `ClientBuilderParams.of()` does not require a
  `ClientFactory` anymore.
  - Hide `DefaultClientBuilderParams` from the public API
- (Deprecation) Deprecate `ClientBuilderParams.factory()`
- (Breaking) `DefaultClientRequestContext()` does not require a
  `ClientFactory` anymore.
- Clean up the client configuration in `AbstractHealthCheckedEndpointGroupBuilder`.
  - (Breaking) `HealthCheckerContext.clientConfigurator()` and
    `clientFactory()` has been replaced with `clientOptions()`.
- Miscellaneous:
  - Remove unused fields in `HealthCheckedEndpointGroupTest`

Result:

- Fixes #1913
- Fixes #2384
- `ClientBuilder.options(ClientOptions)` now sets `ClientFactory` as
  well as other options, which is more convenient when building a
  derived client.
  - However, this may break the following code:
    ```java
    builder.factory(myFactory)
           .options(myOptions)
    ```
    .. because `myFactory` will be ignored and `myOptions.factory()`
    will be used instead.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
@@ -119,6 +119,14 @@ class AbstractClientOptionsBuilder<B extends AbstractClientOptionsBuilder<B>> {
     }
 
     /**
+     * Sets the {@link ClientFactory} used for creating a client.
+     * The default is {@link ClientFactory#ofDefault()}.
+     */
+    public B factory(ClientFactory factory) {
+        return option(ClientOption.FACTORY, requireNonNull(factory, "factory"));
+    }
+
+    /**
      * Sets the timeout of a socket write attempt.
      *
      * @deprecated Use {@link #writeTimeout(Duration)}.
--- a/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
@@ -70,8 +70,6 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder<ClientBuil
 
     private SerializationFormat format = SerializationFormat.NONE;
 
-    private ClientFactory factory = ClientFactory.ofDefault();
-
     /**
      * Creates a new {@link ClientBuilder} that builds the client that connects to the specified {@code uri}.
      *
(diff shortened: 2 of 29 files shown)


ikhoon:
> However, this may break the following code:
> ```
> builder.factory(myFactory)
>        .options(myOptions)
> .. because myFactory will be ignored and myOptions.factory()
> will be used instead.

Everything is fine except this point. Is there any change we notify to a user by 1) `throw Exception` or 2) Add a `@Deprecation`?

trustin:
> Everything is fine except this point. Is there any chance we can notify a user by 1) `throw Exception` or 2) Add a `@Deprecation`?

This was already happening for other setters such as `requestTimeout()`. I think it's hard to tell if it's by intention or mistake..

ikhoon:
> This was already happening for other setters such as requestTimeout(). I think it's hard to tell if it's by intention or mistake..

I see, then SGTM again! 😀

trustin:
Thanks for reviewing!

PR #2398, opened 2020-01-15
Don't fail to start server when version info can't be found.

anuraaga:
I had a report of this stack trace

```
Exception in thread "main" java.lang.NullPointerException
    at com.linecorp.armeria.server.Server.setupVersionMetrics(Server.java:319)
    at com.linecorp.armeria.server.Server.<init>(Server.java:128)
    at com.linecorp.armeria.server.ServerBuilder.build(ServerBuilder.java:1543)
```

Currently we always export version information into the meter registry based on a classpath resource. While the resource should almost always be present, it's conceivable that some scenario will not have the files (old fat jar methods are notorious for only having one copy of service files in the jar) and the server should still start up.

I also modified
- Cache property loading
- Return immutable map

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/Version.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/Version.java
@@ -95,7 +95,7 @@ public final class Version {
      * with arbitrary {@code unknown} values.
      */
     public static Version get(String artifactId, ClassLoader classLoader) {
-        final Version version = identify(classLoader).get(artifactId);
+        final Version version = getAll(classLoader).get(artifactId);
         if (version != null) {
             return version;
         }
@@ -114,8 +114,8 @@ public final class Version {
      *
      * @return A {@link Map} whose keys are Maven artifact IDs and whose values are {@link Version}s
      */
-    public static Map<String, Version> identify() {
-        return identify(Version.class.getClassLoader());
+    public static Map<String, Version> getAll() {
+        return getAll(Version.class.getClassLoader());
     }
 
     /**
@@ -123,7 +123,7 @@ public final class Version {
      *
      * @return A {@link Map} whose keys are Maven artifact IDs and whose values are {@link Version}s
      */
-    public static Map<String, Version> identify(ClassLoader classLoader) {
+    public static Map<String, Version> getAll(ClassLoader classLoader) {
         requireNonNull(classLoader, "classLoader");
 
         return VERSIONS.computeIfAbsent(classLoader, cl -> {
@@ -196,6 +196,31 @@ public final class Version {
 
             return versions.build();
         });
+    }
+
+    /**
+     * Retrieves the version information of Armeria artifacts.
+     * This method is a shortcut for {@code identify(Version.class.getClassLoader())}.
+     *
+     * @return A {@link Map} whose keys are Maven artifact IDs and whose values are {@link Version}s
+     *
+     * @deprecated Use {@link #getAll()}.
+     */
+    @Deprecated
+    public static Map<String, Version> identify() {
+        return getAll(Version.class.getClassLoader());
+    }
+
+    /**
+     * Retrieves the version information of Armeria artifacts using the specified {@link ClassLoader}.
+     *
+     * @return A {@link Map} whose keys are Maven artifact IDs and whose values are {@link Version}s
+     *
+     * @deprecated Use {@link #getAll(ClassLoader)}.
+     */
+    @Deprecated
+    public static Map<String, Version> identify(ClassLoader classLoader) {
+        return getAll(classLoader);
     }
 
     private static long parseIso8601(String value) {
--- a/core/src/main/java/com/linecorp/armeria/common/util/Version.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/Version.java
@@ -152,14 +152,14 @@ public final class Version {
                         "Could not find any property files at " +
                         "META-INF/com.linecorp.armeria.versions.properties. " +
                         "This usually indicates an issue with your application packaging, for example using " +
-                        "a fat-jar method that only keeps one copy of any file. For maximum functionality, " +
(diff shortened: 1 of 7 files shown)


ikhoon:
Nice catch. I found another `Version.identify()` usage. Could you add a null check for the Armeria server header? 😉
https://github.com/line/armeria/blob/32f45dd6b3587f1087c090dcd1280ba8a55bc424/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java#L71-L73

core/src/main/java/com/linecorp/armeria/common/util/Version.java
```diff
@@ -89,61 +99,76 @@
...
-
-            final int dotIndex = k.indexOf('.');
-            if (dotIndex <= 0) {
-                continue;
+            if (!foundProperties) {
+                logger.info(
+                        "Could not find any property files at " +
+                        "META-INF/com.linecorp.armeria.versions.properties. " +
+                        "This usually indicates an issue with your application packaging, for example using " +
+                        "a fat-jar method that only keeps one copy of any file. For maximum functionality, " +
```
trustin: jar -> JAR?

core/src/main/java/com/linecorp/armeria/server/Server.java
```diff
@@ -317,21 +317,23 @@ public int numConnections() {
...
-                                                       repositoryStatus));
-        Gauge.builder("armeria.build.info", () -> 1)
-             .tags(tags)
-             .description("A metric with a constant '1' value labeled by version and commit hash" +
-                          " from which Armeria was built.")
-             .register(meterRegistry);
+        if (versionInfo != null) {
+            final String version = versionInfo.artifactVersion();
+            final String commit = versionInfo.longCommitHash();
+            final String repositoryStatus = versionInfo.repositoryStatus();
```
trustin: Should we fall back to the values like `UNKNOWN`?

core/src/main/java/com/linecorp/armeria/common/util/Version.java
```diff
@@ -89,61 +99,76 @@
...
+                        "it is recommended to fix your packaging to include these files.");
+                return ImmutableMap.of();
             }
 
-            final String artifactId = k.substring(0, dotIndex);
+            // Collect all artifactIds.
+            final Set<String> artifactIds = new HashSet<>();
+            for (Object o: props.keySet()) {
```
minwoox: nit: space between o:

jcchavezs:
Thanks @anuraaga. Does it need a test to cover the change?

anuraaga:
I want a test but this is based on presence of classpath resource that usually is there. Not sure of how to test without a huge amount of engineering (creating a Gradle project that filters the classpath)

jcchavezs:
If it adds any value, the setup of my project is pretty basic and it uses kotlin.

anuraaga:
@jcchavezs You don't use shade plugin or anything? Hmm, I wonder if Kotlin also does some classloader magic that affects us.

core/src/main/java/com/linecorp/armeria/server/Server.java
anuraaga: Ah good point, went ahead and added methods that will return a default value when not found.

jcchavezs:
@anuraaga now I am confused. This are the `pom`s for the failing one: https://gist.github.com/jcchavezs/2c9d006c92b63455bb20539070649855. Notice this is multimodule.

Then I have another example which is fairly standard, using kotlin with this pom https://gist.github.com/jcchavezs/2e9bf06e53bc9b87654ca3537878b1a8 and **here it does not fail**

anuraaga:
@jcchavezs Yup, I'm suspecting the `maven-assembly-plugin` in the failing version, I believe that creates a fat jar and probably doesn't handle multiple files with the same name correctly.

I guess to fix it you'd use `file-aggregator` with META-INF/com.linecorp.armeria.versions.properties as input and output so all the values are merged into a single file.

https://maven.apache.org/plugins/maven-assembly-plugin/examples/single/using-container-descriptor-handlers.html

core/src/main/java/com/linecorp/armeria/common/util/Version.java
```diff
@@ -71,6 +78,36 @@
...
+        if (version != null) {
+            return version;
+        }
+        return new Version(
+                artifactId,
+                "unknown",
+                0,
+                "unknown",
+                "unknown",
+                "unknown");
```
ikhoon: nit: make it singleton?

core/src/main/java/com/linecorp/armeria/common/util/Version.java
```diff
@@ -71,6 +78,36 @@
...
+    public static Version get(String artifactId) {
+        return get(artifactId, Version.class.getClassLoader());
+    }
+
+    /**
+     * Returns the version information for the Armeria artifact named {@code artifactId} using the specified
+     * {@link ClassLoader}. If information for the artifact can't be found, a default value is returned
+     * with arbitrary {@code unknown} values.
+     */
+    public static Version get(String artifactId, ClassLoader classLoader) {
```
ikhoon: Nice! 👍

core/src/main/java/com/linecorp/armeria/common/util/Version.java
anuraaga: Unfortunately, it has `artifactId` mainly for use in DocService. I could special-case docservice itself but since this is an unexpected case, I figured it's fine to keep it simple like this.

anuraaga:
I went ahead and renamed `identify` to `getAll` which seems nicer next to `get`

core/src/main/java/com/linecorp/armeria/common/util/Version.java
ikhoon: This is mainly used for DocsService, LGTM 👍

trustin:
Thanks, @anuraaga !

PR #2399, opened 2020-01-15
Add examples for context propagation using kotlin coroutine

kojilin:
his changeset adds the examples for context propagation using kotlin coroutines.


examples/context-propagation/kotlin/build.gradle
```diff
@@ -0,0 +1,25 @@
+plugins {
+    id 'application'
+    id "org.jetbrains.kotlin.jvm"
+}
+
+dependencies {
+    implementation project(':core')
+
+    implementation 'io.reactivex.rxjava2:rxjava'
```
anuraaga: Remove

examples/context-propagation/kotlin/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+    implementation 'org.jetbrains.kotlin:kotlin-stdlib-jdk8'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-jdk8'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-rx2'
```
anuraaga: Remove

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/Main.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import com.linecorp.armeria.client.WebClient
+import com.linecorp.armeria.common.HttpResponse
+import com.linecorp.armeria.common.HttpStatus
+import com.linecorp.armeria.server.Server
+import com.linecorp.armeria.server.ServiceRequestContext
+
+fun main(args: Array<String>) {
+    val backend = Server.builder()
+        .service("/square/{num}") { ctx: ServiceRequestContext, _ ->
```
anuraaga: Do you need `: ServiceRequestContext`? Or is it just more idiomatic in kotlin?

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/Main.kt
```diff
@@ -0,0 +1,52 @@
...
+            val num = try {
+                ctx.pathParam("num")?.toLong()
+            } catch (e: NumberFormatException) {
```
anuraaga: We're not bothering being safe about this check this in other examples, probably fine not to here (or otherwise we could update the other examples

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/Main.kt
```diff
@@ -0,0 +1,52 @@
...
+            }
+        }
+        .http(8081)
+        .build()
+    val backendClient = WebClient.of("http://localhost:8081")
+    val frontend = Server.builder()
+        .http(8080)
+        .serviceUnder("/", MainService(backendClient))
+        .build()
+    Runtime.getRuntime().addShutdownHook(Thread(Runnable {
```
anuraaga: Do you need `Runnable`?

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/MainService.kt
```diff
@@ -0,0 +1,100 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package example.armeria.contextpropagation.manual
+
+import com.google.common.base.Splitter
```
anuraaga: I think we should be able to use Armeria's code style at least for the imports (we do in `grpc-kotlin` example)

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/MainService.kt
```diff
@@ -0,0 +1,100 @@
...
+            HttpResponse.of(
+                backendResponses.stream()
+                    .map(Function { obj: AggregatedHttpResponse -> obj.contentUtf8() })
+                    .collect(Collectors.joining("\n"))
+            )
+        }
+        return HttpResponse.from(response)
+    }
+
+    private suspend fun fetchFromRequest(req: HttpRequest): List<Long> {
```
anuraaga:
```suggestion
    private suspend fun fetchFromRequest(req: HttpRequest): List<Long> {
      // The context is mounted in a thread-local, meaning it is available to all logic such as tracing.
      require(ServiceRequestContext.current() === ctx)
      require(ctx.eventLoop().inEventLoop())
```

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/MainService.kt
```diff
@@ -0,0 +1,100 @@
...
+        val aggregatedHttpRequest = req.aggregate().await()
+        val nums = mutableListOf<Long>()
```
anuraaga:
```suggestion
        // The context is mounted in a thread-local, meaning it is available to all logic such as tracing.
        require(ServiceRequestContext.current() === ctx)
        require(ctx.eventLoop().inEventLoop())
        val nums = mutableListOf<Long>()
```

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/MainService.kt
```diff
@@ -0,0 +1,100 @@
...
+                    require(ctx.eventLoop().inEventLoop())
+                    backendClient.get("/square/$num").aggregate().asDeferred()
+                }.toTypedArray()
+            ).toList()
+            // The context is mounted in a thread-local, meaning it is available to all logic such as tracing.
+            require(ServiceRequestContext.current() === ctx)
+            require(ctx.eventLoop().inEventLoop())
+            HttpResponse.of(
+                backendResponses.stream()
+                    .map(Function { obj: AggregatedHttpResponse -> obj.contentUtf8() })
```
anuraaga: Doesn't Kotlin have method reference too? `obj::contentUtf8`?

kojilin:
Updated. (sorry that I just using IDE to help convert and miss many bad styles..)

kojilin:
so sad that IDEA didn't provide import layout for kotlin. (I just re-ordered manually)

examples/context-propagation/kotlin/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
```
anuraaga: Can we write this file in kotlin since it's a kotlin project?

examples/context-propagation/kotlin/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-jdk8'
+}
+
+application {
+    mainClassName = 'example.armeria.contextpropagation.kotlin.Main'
+}
+
+compileKotlin {
+    kotlinOptions.jvmTarget = "1.8"
```
anuraaga: I think this is required in our build because of the way we run tests. Maybe we should apply in top level build file instead?

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/MainService.kt
```diff
@@ -0,0 +1,104 @@
...
+            )
+        }
+        return HttpResponse.from(response)
+    }
+
+    private suspend fun fetchFromRequest(ctx: ServiceRequestContext, req: HttpRequest): List<Long> {
+        // The context is mounted in a thread-local, meaning it is available to all logic such as tracing.
+        require(ServiceRequestContext.current() === ctx)
+        require(ctx.eventLoop().inEventLoop())
+        val aggregatedHttpRequest = req.aggregate().await()
```
anuraaga: Can we check context / event loop after await too? I guess that's the equivalent of a callback in future style and it would be incorrect if the context was wrong here and we needed to make another backend call.

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/MainService.kt
```diff
@@ -0,0 +1,104 @@
...
+        return nums
+    }
+
+    private suspend fun fetchFromFakeDb(ctx: ServiceRequestContext): List<Long> {
+        // The context is mounted in a thread-local, meaning it is available to all logic such as tracing.
+        require(ServiceRequestContext.current() === ctx)
+        require(ctx.eventLoop().inEventLoop())
+        // This logic mimics using a blocking method, which would usually be something like a MySQL
+        // database query using JDBC.
+        return withContext(ctx.blockingTaskExecutor().asCoroutineDispatcher()) {
```
anuraaga: What do you think about `withContext` vs `CompletableFuture.supplyAsync().await()`? I'm a bit worried about using a suspending function inside this block which I think would change threads and possibly break something like a transaction. I guess suspension should generally be prevented for uses of blocking executor, do you know if there's a simpler way to do that?

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/Main.kt
```diff
@@ -0,0 +1,47 @@
...
+    val frontend = Server.builder()
+        .http(8080)
+        .serviceUnder("/", MainService(backendClient))
+        .build()
+    Runtime.getRuntime().addShutdownHook(Thread {
+        backend.stop().join()
+        frontend.stop().join()
+    })
+    backend.start().join()
+    frontend.start().join()
```
ikhoon: nit: Could you add empty lines for improving readability?

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/MainService.kt
kojilin: You mean user may do like below?
```
return withContext(ctx.blockingTaskExecutor().asCoroutineDispatcher()) {
            ..
            // begin tx
            ..
            anotherSuspendFunction();
            ..
            // end tx
        }
```

If so, maybe as you said we should wrap whole tx in one `CompletableFuture.supplyAsync()`.

anuraaga: Yeah that's what I'm a bit worried about - maybe it's too much worrying ;) But I wish there were a good pattern to prevent suspension within a certain block of code, couldn't find anything with some googling but it seems important enough to be in Kotlin.

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/kotlin/MainService.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    private suspend fun fetchFromFakeDb(ctx: ServiceRequestContext): List<Long> {
+        // The context is mounted in a thread-local, meaning it is available to all logic such as tracing.
+        require(ServiceRequestContext.current() === ctx)
+        require(ctx.eventLoop().inEventLoop())
+
+        // This logic mimics using a blocking method, which would usually be something like a MySQL
+        // database query using JDBC.
+        return CompletableFuture.supplyAsync(
+            Supplier<List<Long>> {
```
anuraaga: Maybe this type came automatically from intellij

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/kotlin/MainService.kt
```diff
@@ -0,0 +1,126 @@
...
+                // The context is mounted in a thread-local, meaning it is available to all logic such
+                // as tracing.
+                require(ServiceRequestContext.current() === ctx)
+                require(!ctx.eventLoop().inEventLoop())
+
+                Uninterruptibles.sleepUninterruptibly( Duration.ofMillis( 50 ) )
```
anuraaga: Static import was lost (not a big deal) and spaces around parens (bigger deal)

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/manual/MainService.kt
kojilin: Agree, anyway I have updated the code. I'm not familiar with reactive transaction https://spring.io/blog/2019/05/16/reactive-transactions-with-spring. Looks like it can allow us to suspend during tx.

examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/kotlin/MainService.kt
kojilin: Looks like kotlin failed to infer the type, so I need to add it explicitly.

kojilin:
updated :)

PR #2400, opened 2020-01-15
Fix Kotlin plugin warning from Gradle

### LINKED ISSUE #2397: Fix Kotlin plugin warning from Gradle
Getting the following warning when building Armeria:

```
The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build. 
This might happen in subprojects that apply the Kotlin plugins with the Gradle 'plugins { ... }' DSL if they specify explicit versions, even if the versions are equal.
Please add the Kotlin plugin to the common parent project or the root project, then remove the versions in the subprojects.
If the parent project does not need the plugin, add 'apply false' to the plugin line.
See: https://docs.gradle.org/current/userguide/plugins.html#sec:subprojects_plugins_dsl
The Kotlin plugin was loaded in the following projects: ':examples:grpc-kotlin', ':examples:spring-boot-minimal-kotlin'
```

@heowc, would you mind taking a look?

heowc:
Motivation:
- Fix Kotlin plugin warning from Gradle

Modifications:
- Move `org.jetbrains.kotlin.jvm` to root script

Result:
- Fixes #2397

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -417,7 +417,6 @@ org.jctools:
 
 org.jetbrains.kotlin:
   kotlin-allopen: { version: &KOTLIN_VERSION '1.3.61' }
-  kotlin-gradle-plugin: { version: *KOTLIN_VERSION }
 
 org.jlleitschuh.gradle:
   ktlint-gradle: { version: '9.1.1' }
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -415,6 +415,8 @@ org.jctools:
     - from: org.jctools
       to: com.linecorp.armeria.internal.shaded.jctools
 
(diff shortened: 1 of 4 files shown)


dependencies.yml
```diff
@@ -417,7 +417,6 @@ org.jctools:
 
 org.jetbrains.kotlin:
   kotlin-allopen: { version: &KOTLIN_VERSION '1.3.61' }
```
trustin: Could you add the same comment here? So that a dev does not forget to update `build.gradle`?

heowc: Good Idea!

trustin:
Let me merge this away, because it seems trivial and the build passes.

PR #2401, opened 2020-01-15
Helpful message on a 4xx health check response

trustin:
Motivation:

A 4xx health check response usually indicates the client configuration
error such as missing access token.

Modifications:

- Log more friendly message when receiving a 4xx health check response.
- Remove redundant `saturatedMultiply()`

Result:

User-friendliness


trustin:
Thanks, @anuraaga for the heads up.

trustin:
Thanks for quick reviews.

Issue #2404, opened 2020-01-16
Schedulers.isInNonBlockingThread() should return true on EventLoop


minwoox:
In #1665, we made our `EventLoop` implement `NonBlocking` interface so that a user who uses Project Reactor gets a warning log when he/she is using a blocking call on our `EventLoop`s.
We didn't use `NonBlocking` in Project Reactor, but added it under our own `reactor.core.scheduler` directory because we didn't want to add the Reactor dependency to the core.
However, it doesn't seem to work so we should fix it.

Sample snippet from @joonhaeng:
```java
@Slf4j
@RestController
public class DemoController {
    @RequestMapping("/test")
    public String test() {
        log.debug("{}",Thread.currentThread().getName());
        log.debug("{}",Thread.currentThread().getClass());
        String test[] = new String[] {"1", "2", "3", "4", "5"};
        return Flux.fromArray(test).blockFirst(); // A warning log should appear.
    }
}
```

anuraaga:
This looks like it was a proguard issue up to 0.97.0. The `NonBlocking` interface is empty and wasn't used anywhere in our code so got stripped away as being a no-op. If you look at the published JAR, it doesn't have the `reactor` package and the bytecode for `EventLoopThread` does not reference `NonBlocking`.

But the issue should already be fixed at HEAD - now we use the `NonBlocking` interface in our own code

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/eventloop/EventLoopCheckingUtil.java#L47

so proguard keeps it.

minwoox:
Ah, I couldn't even imagine that. Thanks!

ikhoon:
Thanks for the inspection! The mystery is solved. 🕵️‍♀️

ikhoon:
Already Fixed by #2275

PR #2405, opened 2020-01-16
Add test for #2404 (Reactor non-blocking integration)

anuraaga:
Seems good to me. I guess there might be an issue if proguard strips out `NonBlocking` from our jar? ~~Armeria and reactor are in different classloaders? It'd be interesting to hear more details of the deployment setup to add to #2404~~


minwoox:
@anuraaga To reproduce, I just put `Schedulers.isInNonBlockingThread()` in [here](https://github.com/line/armeria-examples/blob/master/server-sent-events/src/main/java/example/armeria/server/sse/Main.java#L52) (Please note that it's `armeria-examples` not `armeria/examples`)

trustin:
Thanks, @anuraaga for looking into the cause.

PR #2406, opened 2020-01-16
Add more useful methods to `AsyncCloseable`

trustin:
Motivation:

When a user closes something asynchronously, it is likely that he or she
wants the following functionalities:

- Getting notified when it is closed, sometimes even without closing it.
- Using with `try`-with-resources.

Modifications:

- Make `AsyncCloseable` extend `AutoCloseable`.
- Add `AsyncCloseable.closeFuture()` which allows a user to get the
  close future without closing.
- Add `AsyncCloseableSupport` which helps a user implement `AsyncCloseable`.
- Made the following classes implement `AsyncCloseable`:
  - `ClientFactory`
  - `HttpChannelPool`
  - `Server`
  - `StartStopSupport`
- Stopped using `Mockito.spy()` in `StartStopSupportTest` because can't
  test with spies.
- Fixed a dead lock in `HealthCheckedEndpointGroupTest.delegateUpdateCandidatesWhileCreatingHealthCheckedEndpointGroup()`

Result:

- `ClientFactory` and `Server` can be closed asynchronously.
- A user can do something when a `ClientFactory` closes, which will be
  useful for https://github.com/line/armeria/issues/1084#issuecomment-572481160

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
@@ -24,10 +24,14 @@ import java.lang.reflect.Proxy;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Set;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Supplier;
 
 import javax.annotation.Nullable;
 
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Strings;
@@ -35,12 +39,15 @@ import com.google.common.base.Strings;
 import com.linecorp.armeria.common.Scheme;
 import com.linecorp.armeria.common.SerializationFormat;
 import com.linecorp.armeria.common.SessionProtocol;
+import com.linecorp.armeria.common.util.AsyncCloseable;
+import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.common.util.ReleasableHolder;
 import com.linecorp.armeria.common.util.Unwrappable;
 
 import io.micrometer.core.instrument.MeterRegistry;
 import io.netty.channel.EventLoop;
 import io.netty.channel.EventLoopGroup;
+import reactor.core.scheduler.NonBlocking;
 
 /**
  * Creates and manages clients.
@@ -60,7 +67,7 @@ import io.netty.channel.EventLoopGroup;
  * {@link ClientFactory}, use {@link #closeDefault()}.
  * </p>
  */
-public interface ClientFactory extends AutoCloseable {
+public interface ClientFactory extends AsyncCloseable {
 
     /**
      * The default {@link ClientFactory} implementation.
@@ -96,12 +103,38 @@ public interface ClientFactory extends AutoCloseable {
      * Closes the default {@link ClientFactory}.
      */
     static void closeDefault() {
-        LoggerFactory.getLogger(ClientFactory.class).debug(
-                "Closing the default {}", ClientFactory.class.getSimpleName());
-        try {
-            DefaultClientFactory.DEFAULT.doClose();
-        } finally {
-            DefaultClientFactory.INSECURE.doClose();
+        final Logger logger = LoggerFactory.getLogger(ClientFactory.class);
+        logger.debug("Closing the default client factories");
+        final CompletableFuture<Void> closeFuture = CompletableFuture.allOf(
+                DefaultClientFactory.DEFAULT.closeAsync(false),
+                DefaultClientFactory.INSECURE.closeAsync(false)).handle((unused1, cause) -> {
+            if (cause == null) {
+                logger.debug("Closed the default client factories");
+            } else {
+                logger.warn("Failed to close the default client factories:", Exceptions.peel(cause));
+            }
+            return null;
+        });
+
+        if (!(Thread.currentThread() instanceof NonBlocking)) {
+            boolean interrupted = false;
+            try {
+                for (;;) {
+                    try {
+                        closeFuture.get();
+                        break;
+                    } catch (InterruptedException e) {
+                        interrupted = true;
+                    } catch (ExecutionException | CancellationException ignored) {
+                        // Will be logged by the callback we added above.
+                        break;
+                    }
+                }
+            } finally {
+                if (interrupted) {
+                    Thread.currentThread().interrupt();
+                }
+            }
         }
     }
 
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
@@ -39,8 +39,8 @@ import com.google.common.base.Strings;
 import com.linecorp.armeria.common.Scheme;
 import com.linecorp.armeria.common.SerializationFormat;
 import com.linecorp.armeria.common.SessionProtocol;
-import com.linecorp.armeria.common.util.AsyncCloseable;
 import com.linecorp.armeria.common.util.Exceptions;
+import com.linecorp.armeria.common.util.ListenableAsyncCloseable;
 import com.linecorp.armeria.common.util.ReleasableHolder;
 import com.linecorp.armeria.common.util.Unwrappable;
 
@@ -67,7 +67,7 @@ import reactor.core.scheduler.NonBlocking;
  * {@link ClientFactory}, use {@link #closeDefault()}.
  * </p>
  */
-public interface ClientFactory extends AsyncCloseable {
+public interface ClientFactory extends ListenableAsyncCloseable {
 
     /**
      * The default {@link ClientFactory} implementation.
@@ -352,11 +352,4 @@ public interface ClientFactory extends AsyncCloseable {
         }
         return params;
     }
-
-    /**
-     * Closes all clients managed by this factory and shuts down the {@link EventLoopGroup}
-     * created implicitly by this factory.
-     */
-    @Override
-    void close();
 }
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.Supplier;
 
 import org.slf4j.Logger;
@@ -34,6 +35,7 @@ import com.google.common.collect.Streams;
 
 import com.linecorp.armeria.common.Scheme;
 import com.linecorp.armeria.common.SessionProtocol;
+import com.linecorp.armeria.common.util.AsyncCloseableSupport;
 import com.linecorp.armeria.common.util.ReleasableHolder;
 
 import io.micrometer.core.instrument.MeterRegistry;
@@ -79,6 +81,7 @@ final class DefaultClientFactory implements ClientFactory {
     private final HttpClientFactory httpClientFactory;
     private final Map<Scheme, ClientFactory> clientFactories;
     private final List<ClientFactory> clientFactoriesToClose;
+    private final AsyncCloseableSupport closeable = AsyncCloseableSupport.of(this::doCloseAsync);
 
     DefaultClientFactory(HttpClientFactory httpClientFactory) {
         this.httpClientFactory = httpClientFactory;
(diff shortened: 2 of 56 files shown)


trustin:
Build keeps timing out for a unknown reason. Build passes locally, though.

trustin:
Force-pushed to launder the ugly commits.

trustin:
Build is now passing.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -96,12 +103,38 @@ static ClientFactoryBuilder builder() {
...
     static void closeDefault() {
-        LoggerFactory.getLogger(ClientFactory.class).debug(
-                "Closing the default {}", ClientFactory.class.getSimpleName());
-        try {
-            DefaultClientFactory.DEFAULT.doClose();
-        } finally {
-            DefaultClientFactory.INSECURE.doClose();
+        final Logger logger = LoggerFactory.getLogger(ClientFactory.class);
+        logger.debug("Closing the default client factories");
+        final CompletableFuture<Void> closeFuture = CompletableFuture.allOf(
```
minwoox: Shouldn't we use `CompletableFutures.succesfulAsList(...)` (log the cause using `defaultValueMapper`) because we have to wait for all futures to be complete?

core/src/main/java/com/linecorp/armeria/client/HttpClientFactory.java
```diff
@@ -256,22 +266,56 @@ public Object newClient(ClientBuilderParams params) {
...
+            dependencies.add(i.next().closeAsync());
             i.remove();
         }
 
         addressResolverGroup.close();
-        if (shutdownWorkerGroupOnClose) {
-            workerGroup.shutdownGracefully().syncUninterruptibly();
-        }
+
+        CompletableFuture.allOf(dependencies.toArray(EMPTY_FUTURES)).handle((unused, cause) -> {
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseable.java
```diff
@@ -22,12 +22,24 @@
  * method releases the resources asynchronously, returning the {@link CompletableFuture} which is completed
```
minwoox: need to change the doc a little bit?
Unlike {@link AutoCloseable}, -> In addition to the funcationality of closing synchronously with {@link AutoCloseable},

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseableSupport.java
```diff
@@ -0,0 +1,203 @@
...
+     *                    completes the given {@link CompletableFuture}.
+     */
+    public static AsyncCloseableSupport of(Consumer<CompletableFuture<?>> closeAction) {
+        return new AsyncCloseableSupport(requireNonNull(closeAction, "closeAction"));
+    }
+
+    /**
+     * Returns the {@link AsyncCloseableSupport} which has been closed already.
+     */
+    public static AsyncCloseableSupport closed() {
```
minwoox: Not sure if we need this method, but if we do, how about making a static variable for this?

trustin:
> I'm not sure but is it worth to make `EndpointGroup` as `AsyncCloseable` as well?

SGTM.

trustin:
`EndpointGroup` is now `AsyncCloseable`. `closeFuture()` will always return a completed future, which may or may not make sense. Should we remove `AsyncCloseable.closeFuture()` and let a user choose to provide `closeFuture()`? e.g. `ClientFactory` or `Server` will have `closeFuture()` but `EndpointGroup` won't.

trustin:
Thought `UnmodifiableFuture` deserves to be a part of `completable-futures`, so create an issue: https://github.com/spotify/completable-futures/issues/65

minwoox:
> always return a completed future for static endpoint groups

I think it is okay. But I sligtly prefer `AsyncCloseable` not to have `closeFuture`.

> Thought UnmodifiableFuture deserves to be a part of completable-futures, so create an issue: spotify/completable-futures#65

That's a good idea. Also I thought some of our APIs should return and use that such as `streamMessage.completionFuture()`, `RequestContextAwareCompletableFuture`, etc.

minwoox:
Actually, most of the futures returned from our API could be unuodatable and EventLoopChecking.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
```diff
@@ -96,12 +103,38 @@ static ClientFactoryBuilder builder() {
...
+            } else {
+                logger.warn("Failed to close the default client factories:", Exceptions.peel(cause));
+            }
+            return null;
+        });
+
+        if (!(Thread.currentThread() instanceof NonBlocking)) {
+            boolean interrupted = false;
+            try {
+                for (;;) {
```
anuraaga: Can you add a comment why we don't use `join()` here? I think this is basically only called from a shutdown hook ever, so interruption probably doesn't matter so much but not sure.

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseable.java
```diff
@@ -18,16 +18,28 @@
...
-@FunctionalInterface
-public interface AsyncCloseable {
+public interface AsyncCloseable extends AutoCloseable {
     /**
-     * Releases the resources held by this object asynchronously.
+     * Returns the {@link CompletableFuture} which is completed after the resources are released. Note that
+     * you must use {@link #close()} or {@link #closeAsync()} to release the resources actually. This method
+     * merely returns the {@link CompletableFuture}.
+     */
+    CompletableFuture<?> closeFuture();
```
anuraaga: Having two similar names with same return type is making me scratch my head a little. It looks like future is a way of closing out of these three

- (Implicitly implied) Blocking close
- Async close
- Future close

I know this matches our general pattern, but shall we consider renaming all `fooFuture` to `whenFoo`? `ctx.log().whenComplete().thenAccept()` reads like English and seems quite nice to me.

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
```diff
@@ -162,20 +165,55 @@ public Object newClient(ClientBuilderParams params) {
...
+            return closeFuture();
+        }
+        return closeable.closeAsync();
+    }
+
+    private void closeAsync(CompletableFuture<?> future) {
+        final CompletableFuture<?>[] delegateCloseFutures =
+                clientFactoriesToClose.stream()
+                                      .map(ClientFactory::closeAsync)
+                                      .toArray(CompletableFuture[]::new);
```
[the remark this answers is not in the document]
trustin: `Steam.toArray()` does not accept it

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseable.java
```diff
@@ -18,16 +18,28 @@
...
- * An object that may hold resources until it is closed. Unlike {@link AutoCloseable}, the {@link #closeAsync()}
- * method releases the resources asynchronously, returning the {@link CompletableFuture} which is completed
- * after the resources are released.
+ * An object that may hold resources until it is closed. In addition to {@link AutoCloseable#close()},
+ * this interface provides {@link #closeAsync()} which releases the resources asynchronously, returning
+ * a {@link CompletableFuture} which is completed after the resources are released.
  */
-@FunctionalInterface
-public interface AsyncCloseable {
+public interface AsyncCloseable extends AutoCloseable {
```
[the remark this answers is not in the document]
trustin: Added the label.

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseableSupport.java
```diff
@@ -0,0 +1,211 @@
...
+     * >     public CompletableFuture<?> close() {
+     * >         return closeableSupport.close();
+     * >     }
+     * > }
+     * }</pre>
+     *
+     * @param closeAction the {@link Consumer} which performs the task that release the resources and
+     *                    completes the given {@link CompletableFuture}.
+     */
+    public static AsyncCloseableSupport of(Consumer<CompletableFuture<?>> closeAction) {
```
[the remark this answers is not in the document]
trustin: Thought about it, but `? super` does not make sense here, because the consumer must call `CompletableFuture.complete*()`. A consumer that accepts a `CompletionStage` or `Object` cannot do that.

core/src/main/java/com/linecorp/armeria/client/ClientFactory.java
trustin: It's because it may not be invoked from a shutdown hook? Nothing stops a user from calling this method manually (or automatically via some framework).

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseable.java
trustin: That's a good idea. Let me rename. `whenClosed()`, then?

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseableSupport.java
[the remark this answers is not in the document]
ikhoon: Either way is fine. Sometimes contravariance makes me confused.🤪 
Because we can call `CompletableFuture.complete*()` even it is defined with `Consumer<? super CompletableFuture<?>>`. 🧐
```java
// Compiles
Consumer<? super CompletableFuture<Integer>> consumerSuper = x -> x.complete(10);
```

core/src/main/java/com/linecorp/armeria/client/DefaultClientFactory.java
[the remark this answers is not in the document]
ikhoon: Oops! Thanks.

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseableSupport.java
```diff
@@ -80,27 +77,24 @@ public static AsyncCloseableSupport of() {
      * Returns a new {@link AsyncCloseableSupport} which calls the specified {@link Consumer} on
      * {@link #close()} or {@link #closeAsync()}.
      * <pre>{@code
-     * > class MyClass implements AutoCloseable {
-     * >     final AsyncCloseableSupport closeableSupport = AsyncCloseableSupport.of(f -> {
+     * > class MyClass implements AsyncCloseable {
```
minwoox: nit: This one should be `ListenableAsyncCloseable` as well.

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseableSupport.java
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Returns the {@link AsyncCloseableSupport} which has been closed already.
+     */
+    public static AsyncCloseableSupport closed() {
+        return CLOSED;
+    }
+
+    private final Consumer<CompletableFuture<?>> closeAction;
+    private final CompletableFuture<?> closeFuture = new CompletableFuture<>();
+    private final UnmodifiableFuture<?> unupdatableCloseFuture =
```
minwoox: nit: unmodifiable?

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseableSupport.java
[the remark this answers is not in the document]
trustin: Yeah it compiles, but what I'm saying is this doesn't make sense:

```java
Consumer<CompletionStage<?>> consumer = stage -> ...; // Can't call complete()
AsyncCloseableSupport.of(consumer); // This doesn't make sense. Consumer must call .complete().
```

core/src/main/java/com/linecorp/armeria/common/util/UnmodifiableFuture.java
```diff
@@ -0,0 +1,139 @@
...
+
+    static {
+        NIL = new UnmodifiableFuture<>();
+        NIL.doComplete(null);
+    }
+
+    /**
+     * Returns an {@link UnmodifiableFuture} which has been completed with the specified {@code value}.
+     */
+    public static <U> UnmodifiableFuture<U> completedFuture(@Nullable U value) {
```
minwoox: Shouldn't this return `CompletableFuture`?

core/src/main/java/com/linecorp/armeria/common/util/UnmodifiableFuture.java
```diff
@@ -0,0 +1,139 @@
...
+        final UnmodifiableFuture<U> future = new UnmodifiableFuture<>();
+        future.doComplete(value);
+        return future;
+    }
+
+    /**
+     * Returns an {@link UnmodifiableFuture} which has been completed exceptionally with the specified
+     * {@link Throwable}.
+     */
+    public static <U> UnmodifiableFuture<U> exceptionallyCompletedFuture(Throwable cause) {
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/util/UnmodifiableFuture.java
```diff
@@ -0,0 +1,139 @@
...
+        final UnmodifiableFuture<U> future = new UnmodifiableFuture<>();
+        future.doCompleteExceptionally(cause);
+        return future;
+    }
+
+    /**
+     * Returns an {@link UnmodifiableFuture} which will be completed when the specified
+     * {@link CompletableFuture} is completed.
+     */
+    public static <U> UnmodifiableFuture<U> wrap(CompletableFuture<U> future) {
```
minwoox: ditto

core/src/main/java/com/linecorp/armeria/common/util/AsyncCloseableSupport.java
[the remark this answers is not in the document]
ikhoon: > Consumer must call .complete().

I understood. You may want to constrain consumer implementation. It makes sense to me. 👍

core/src/main/java/com/linecorp/armeria/common/util/UnmodifiableFuture.java
minwoox: On second thought, it's better to return `UnmodifiableFuture`. 😉 Sorry about the noise.

trustin: I'm not sure. It's like `ImmutableList.of()` returns an `ImmutableList` instead of a `List`. We might add more public API to it later as well.

adriancole:
would be cool to have a before/after code example showing how this makes some call simpler or more robust, but happy to wait for the release notes

Issue #2360, opened 2020-01-02
Check for over-exposed APIs


anuraaga:
We noticed in https://github.com/line/armeria/pull/2343 that there is a public interface for controlling timeout marked as internal-only in Javadoc. Generally, we should be able to avoid that by using the `internal` package, but we may have some APIs that are exposed that shouldn't be.

A couple I found are
- DefaultServiceRequestContext
- DefaultClientRequestContext

I think users are supposed to use request context through their interfaces so do these need to be public classes? If we need to allow users to instantiate these objects, it follow our pattern better to have factory methods on e.g. `ServiceRequestContext` that returns `ServiceRequestContext` by creating `DefaultServiceRequestContext` instead of directly exposing the class I think. This would also allow hiding classes like `TimeoutController`

minwoox:
I agree. We can even make `DefaultServiceRequestContext` package-local.

trustin:
+1

trustin:
For the classes that cannot be hidden, we could introduce annotations like `@UnstableApi` (similar to Guava's `@Beta`).

trustin:
@anuraaga I found we expose protobuf-generated classes such as `Debug`, `StackElementProto` and `ThrowableProto` in `common.grpc`. Should we hide it from the public API?

anuraaga:
I don't think we need to hide generated protos unless we actually hid protobuf itself too, which I guess isn't possible for gRPC.

In addition, for those classes I actually want us to commit to the protocol itself - Armeria transparently converts to domain objects but I think it's good if non-Java languages also can matshal those structs. For best UX I guess we would want to extract a module with only protos but in the meantime I hope we can treat the protos as public API.

trustin:
OK. I think the name `Debug` sounds somewhat vague. Do you have a better name? I also would like to make protoc generate some Javadoc for `Debug` class, which is left empty at the moment.

trustin:
The classes that could be hidden or annotated with `@UnstableApi`:

- `DefaultClientRequestContext` and `DefaultServiceRequestContext`
- `*Provider`
- `DefaultRpcResponse...`
- `DefaultRequestLog`
- `DocServicePlugin` and its related classes such as `*Info`

The classes that could be removed completely due to their dubious value:

- Everything in `server.logging.structured{,.kafka}`

anuraaga:
[...]

Issue #1409, opened 2018-10-25
Change FixedStreamMessage not to discard its stream message


minwoox:
Currently, `FixedStreamMessage` releases its elements as soon as the element was subscribed. But if we change it to release after all of the elements are subscribed, we can duplicate the `FixedStreamMessage` without using `StreamMessageDuplicator`.
But we have to consider that there's a tradeoff because we have to hole it in memory litter longer.

anuraaga:
I agree it should be fine to release after subscription, the extra time the objects stay around should be insignificant.

minwoox:
The duplicator has to enable a user to
- Abort it
  - All subscriptions are aborted and all elements are discarded
- Close it
  - Close, here, means that the duplicator is not subscribable and all elements are discarded after all subscriptions are done.
- Make it closed after one more subscription
  - Closed means the same status described above.

The `FixedStreamMessage` has to be a duplicator to be subscribed multiple times. We need an API for that.
```java
HttpResponse res = HttpResponse.of(...); // FixedStreamMessage
RepeatableHttpResponse repeatable = res.toRepeatable();
// res.subscribe(...)  throws an IllegalStateException

repeatable.subscribe(...)
repeatable.subscribe(...)
repeatable.subscribe(..., SubscriptionOption.LAST_SUBSCRIPTION)
// repeatable.subscribe(...) throws an IllegalStateException
// repeatable is now closed status. All elements are discarded after subscriptions are done.
```
The `toRepeatable()` is also applicable to other stream messages:
```java
HttpResponse streaming = HttpResponse.streaming();
RepeatableHttpResponse repeatable = streaming.toRepeatable();
// streaming.subscribe(...)  throws an IllegalStateException

repeatable.subscribe(...)
repeatable.subscribe(...)
repeatable.closeAfterAnotherSubscription(); // Do we have a better name for this?
// If we called repeatable.close() here, no more subscription.
repeatable.subscribe(...) // No need to specify SubscriptionOption.LAST_SUBSCRIPTION
// repeatable.subscribe(...) throws an IllegalStateException
```
The changes are that the duplicator has to call `duplicateStream` before a subscriber subscribes to.
```
duplicator.duplicateStream().subscribe(...);
```
The name is also changed from `HttpResponseDuplicator` to `RepeatableHttpResponse`.

I will do the wrapping `FixedStreamMessage` after 1.0 because it does not bring any breaking changes, but I have to do the rest of it before 1.0. 😄

[...]

Issue #2189, opened 2019-10-15
Content preview and compressed stream


trustin:
When sending or receiving a compressed stream, `RequestLog.contentPreview()` will preview the compressed content, which is not very human friendly.

It would be nice if we can push the uncompressed content preview to `RequestLog`, with low overhead. Currently, a user can write a content previewer that performs decompression, but it means the content will be decompressed twice, once for previewing and once for actual content decoding.

snaiper80:
I also need this exactly. Currently, I implemented a simple content previewer that is naive approach using `ContentPreviewerFactory` to decompress the gzipped body stream. But this codes have been running in debug mode only due to performance problem as you're already said.

trustin:
@anuraaga Any thoughts on this? I can't think of a way to handle this nicely..

anuraaga:
@trustin I guess instead of being called in-band by `RequestLog`, the functionality can be moved to a decorator? Then a user gets to decide too whether they want to preview the encoded or decoded content. I think the main APIs are already there, `RequestContext.logBuilder().requestContentPreview(String)` so just need to move the logic to a decorator. It's similar to the decorator linked in #1997 except that used `aggregate` for ease of implementation, while it's not that much worse to use a normal subscription.

anuraaga:
@trustin Probably good to target this for 1.0.0

trustin:
Yeah, I wonder how it will impact our current service binding builder API. It would result in a surprise if a user specifies previewers there and forgets to add a previewing decorator. Any idea for preventing this? Just checking with `as()` ?

minwoox:
>  I guess instead of being called in-band by RequestLog, the functionality can be moved to a decorator?

That's a good idea. Let me work on this. I think we can add `ContentPreviewingService` and `ContentPreviewingClient` with their builders. Inside the decorator, we can use `FilteredStreamMessage` to subscribe once.

anuraaga:
@trustin Sorry for not replying to your question. My idea was just to remove the current service binding content preview flags since the decorator itself could have all the configuration.

minwoox:
Yes, we can completely move the logic to the decorators. 😉

PR #2410, opened 2020-01-20
Add StreamMessage.toDuplicator

minwoox:
Motivation:
Before I get into #1409, I need to clean up the duplicator so that `FixedStreamMessage` and other `StreamMessage`s can have the same APIs to return the duplicator.

Modifications:
- Add `StreamMessage.toDuplicator()`
- Add `StreamMessageDuplicator`, `HttpRequestDuplicator` and `HttpResponseDuplicator` interfaces

Result:
- (Breaking)
  - `HttpRequestDuplicator` and `HttpResponseDuplicator` are now interfaces.
    - You should use `HttpRequest.toDuplicator()` and `HttpResponse.toDuplicator()` to use the duplicators.

To-dos:
- Implement a new duplicator that optimizes `FixedStreamMessage` by wrapping it.
- Move `DefaultStreamMessageDuplicator` under the internal package. Related #2360

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/DefaultHttpRequestDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/DefaultHttpRequestDuplicator.java
@@ -18,8 +18,6 @@ package com.linecorp.armeria.common;
 
 import static java.util.Objects.requireNonNull;
 
-import javax.annotation.Nullable;
-
 import com.google.common.base.MoreObjects;
 
 import com.linecorp.armeria.common.stream.DefaultStreamMessageDuplicator;
--- a/core/src/main/java/com/linecorp/armeria/common/DefaultHttpRequestDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/DefaultHttpRequestDuplicator.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2020 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.common;
+
+import static java.util.Objects.requireNonNull;
+
+import javax.annotation.Nullable;
+
+import com.google.common.base.MoreObjects;
+
+import com.linecorp.armeria.common.stream.DefaultStreamMessageDuplicator;
+import com.linecorp.armeria.common.stream.StreamMessage;
+import com.linecorp.armeria.common.stream.StreamMessageWrapper;
+
+import io.netty.util.concurrent.EventExecutor;
+
+final class DefaultHttpRequestDuplicator
+        extends DefaultStreamMessageDuplicator<HttpObject> implements HttpRequestDuplicator {
+
+    private final RequestHeaders headers;
+
+    DefaultHttpRequestDuplicator(HttpRequest req, EventExecutor executor, long maxRequestLength) {
+        super(requireNonNull(req, "req"), obj -> {
+            if (obj instanceof HttpData) {
+                return ((HttpData) obj).length();
+            }
+            return 0;
+        }, executor, maxRequestLength);
+        headers = req.headers();
+    }
+
+    @Override
+    public HttpRequest duplicate() {
+        return duplicate(headers);
+    }
+
+    @Override
+    public HttpRequest duplicate(boolean lastStream) {
+        return duplicate(headers, lastStream);
+    }
+
+    @Override
+    public HttpRequest duplicate(RequestHeaders newHeaders) {
+        return duplicate(newHeaders, false);
+    }
+
+    @Override
+    public HttpRequest duplicate(RequestHeaders newHeaders, boolean lastStream) {
+        requireNonNull(newHeaders, "newHeaders");
+        return new DuplicatedHttpRequest(super.duplicate(lastStream), newHeaders);
+    }
+
+    private class DuplicatedHttpRequest
+            extends StreamMessageWrapper<HttpObject> implements HttpRequest {
+
+        private final RequestHeaders headers;
+
+        DuplicatedHttpRequest(StreamMessage<? extends HttpObject> delegate,
+                              RequestHeaders headers) {
+            super(delegate);
+            this.headers = headers;
+        }
+
+        @Override
+        public RequestHeaders headers() {
+            return headers;
+        }
+
+        // Override to return HttpRequestDuplicator.
+
+        @Override
+        public HttpRequestDuplicator toDuplicator() {
+            return toDuplicator(duplicatorExecutor());
+        }
+
+        @Override
+        public HttpRequestDuplicator toDuplicator(EventExecutor executor) {
+            return HttpRequest.super.toDuplicator(executor);
+        }
+
+        @Override
+        public EventExecutor defaultSubscriberExecutor() {
+            return duplicatorExecutor();
+        }
+
+        @Override
+        public String toString() {
+            return MoreObjects.toStringHelper(this)
+                              .add("headers", headers)
+                              .toString();
+        }
+    }
+}
--- a/core/src/main/java/com/linecorp/armeria/common/DefaultHttpRequestDuplicator.java
+++ b/core/src/main/java/com/linecorp/armeria/common/DefaultHttpRequestDuplicator.java
@@ -47,19 +47,9 @@ final class DefaultHttpRequestDuplicator
     }
 
     @Override
-    public HttpRequest duplicate(boolean lastStream) {
-        return duplicate(headers, lastStream);
-    }
-
-    @Override
     public HttpRequest duplicate(RequestHeaders newHeaders) {
-        return duplicate(newHeaders, false);
-    }
-
-    @Override
-    public HttpRequest duplicate(RequestHeaders newHeaders, boolean lastStream) {
         requireNonNull(newHeaders, "newHeaders");
-        return new DuplicatedHttpRequest(super.duplicate(lastStream), newHeaders);
+        return new DuplicatedHttpRequest(super.duplicate(), newHeaders);
     }
 
     private class DuplicatedHttpRequest
@@ -67,8 +57,7 @@ final class DefaultHttpRequestDuplicator
 
         private final RequestHeaders headers;
 
-        DuplicatedHttpRequest(StreamMessage<? extends HttpObject> delegate,
-                              RequestHeaders headers) {
+        DuplicatedHttpRequest(StreamMessage<? extends HttpObject> delegate, RequestHeaders headers) {
             super(delegate);
             this.headers = headers;
         }
--- a/core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
+++ b/core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
@@ -427,10 +427,7 @@ public interface HttpRequest extends Request, StreamMessage<HttpObject> {
      * the trailers of the request is received fully.
      */
     default CompletableFuture<AggregatedHttpRequest> aggregate() {
-        final CompletableFuture<AggregatedHttpRequest> future = new EventLoopCheckingCompletableFuture<>();
-        final HttpRequestAggregator aggregator = new HttpRequestAggregator(this, future, null);
-        subscribe(aggregator);
-        return future;
+        return aggregate(defaultSubscriberExecutor());
     }
 
     /**
@@ -452,11 +449,7 @@ public interface HttpRequest extends Request, StreamMessage<HttpObject> {
      * use {@link #aggregate()}.
      */
     default CompletableFuture<AggregatedHttpRequest> aggregateWithPooledObjects(ByteBufAllocator alloc) {
-        requireNonNull(alloc, "alloc");
-        final CompletableFuture<AggregatedHttpRequest> future = new EventLoopCheckingCompletableFuture<>();
-        final HttpRequestAggregator aggregator = new HttpRequestAggregator(this, future, alloc);
-        subscribe(aggregator, SubscriptionOption.WITH_POOLED_OBJECTS);
-        return future;
+        return aggregateWithPooledObjects(defaultSubscriberExecutor(), alloc);
     }
 
     /**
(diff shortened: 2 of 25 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -317,6 +318,21 @@ default void subscribe(Subscriber<? super T> subscriber, SubscriptionOption... o
...
 
+    /**
+     * To duplicator.
+     */
+    default StreamMessageDuplicator<T> toDuplicator() {
+        return toDuplicator(defaultSubscriberExecutor());
+    }
+
+    /**
+     * To duplicator.
```
trustin: Javadoc?

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -319,14 +319,24 @@ default void subscribe(Subscriber<? super T> subscriber, SubscriptionOption... o
     CompletableFuture<List<T>> drainAll(EventExecutor executor, SubscriptionOption... options);
 
     /**
-     * To duplicator.
+     * Returns a new {@link StreamMessageDuplicator} that duplicates multiple {@link StreamMessage}s which
+     * publish the same elements with this {@link StreamMessage}.
```
trustin: `that duplicates this {@link StreamMessage} into one ore more {@link StreamMessage}s, which publish the same elements.` ?

core/src/main/java/com/linecorp/armeria/common/HttpRequest.java
```diff
@@ -474,4 +467,62 @@ default HttpRequest withHeaders(RequestHeadersBuilder newHeadersBuilder) {
         subscribe(aggregator, executor, SubscriptionOption.WITH_POOLED_OBJECTS);
         return future;
     }
+
+    /**
+     * Returns a new {@link HttpRequestDuplicator} that duplicates multiple {@link HttpRequest}s which publish
+     * the same elements with this {@link HttpRequest}.
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -437,4 +430,62 @@ static HttpResponse ofFailed(Throwable cause) {
         subscribe(aggregator, executor, SubscriptionOption.WITH_POOLED_OBJECTS);
         return future;
     }
+
+    /**
+     * Returns a new {@link HttpResponseDuplicator} that duplicates multiple {@link HttpResponse}s which publish
+     * the same elements with this {@link HttpResponse}.
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
minwoox: Fixed. 😉

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
```diff
@@ -437,4 +430,62 @@ static HttpResponse ofFailed(Throwable cause) {
...
+     * Note that you cannot subscribe to this {@link HttpResponse} anymore after you call this method.
+     * To subscribe, call {@link HttpResponseDuplicator#duplicate()} from the returned
+     * {@link HttpResponseDuplicator}.
+     *
+     * @param maxResponseLength the maximum response length that the duplicator can hold in its buffer.
+     *                         {@link ContentTooLargeException} is raised if the length of the buffered
+     *                         {@link HttpData} is greater than this value.
+     */
+    default HttpResponseDuplicator toDuplicator(long maxResponseLength) {
```
ikhoon: nit: How about narrowing down the maxSingalLength to `int` here than truncating the `maxResponseLength` in `StreamMessageProcessor` constructor?

core/src/main/java/com/linecorp/armeria/common/HttpRequestDuplicator.java
```diff
@@ -16,128 +16,69 @@
...
+ * A duplicator that duplicates a {@link HttpRequest} into one or more {@link HttpRequest}s,
+ * which publish the same elements.
  *
  * <pre>{@code
- * final HttpRequest originalReq = ...
- * final HttpRequestDuplicator reqDuplicator = new HttpRequestDuplicator(originalReq);
- *
- * final HttpRequest dupReq1 = reqDuplicator.duplicateStream();
- * final HttpRequest dupReq2 = reqDuplicator.duplicateStream(true); // the last stream
+ * HttpRequest<String> httpRequest = ...
```
ikhoon: `c.l.a.c.HttpRequest` doesn't seem to have type parameter. 😀

core/src/main/java/com/linecorp/armeria/common/HttpRequestDuplicator.java
```diff
@@ -16,128 +16,69 @@
...
+ * HttpRequestDuplicator<String> duplicator = httpRequest.toDuplicator();
+ * // httpRequest.subscribe(...) will throw an exception. You cannot subscribe to httpRequest anymore.
  *
- * dupReq1.subscribe(new FooSubscriber() {
- *     ...
- *     // Do something according to the first few elements of the request.
- * });
+ * // Duplicate the stream as many as you want to subscribe.
+ * HttpRequest<String> duplicatedHttpRequest1 = duplicator.duplicate();
```
ikhoon: Ditto.

core/src/main/java/com/linecorp/armeria/common/HttpResponseDuplicator.java
```diff
@@ -16,100 +16,54 @@
...
- * >     public void onNext(Object o) {
- * >     ...
- * >     // Do something according to the header's status.
- * >     }
- * > });
- * >
- * > dupRes2.aggregate().handle((aRes, cause) -> {
- * >     // Do something with the message.
- * > });
+ * HttpResponse<String> httpResponse = ...
```
ikhoon: Ditto

core/src/main/java/com/linecorp/armeria/common/HttpRequestDuplicator.java
minwoox: Oops. 😅

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessage.java
```diff
@@ -326,7 +352,10 @@ default void subscribe(Subscriber<? super T> subscriber, SubscriptionOption... o
      * different depending on this {@link StreamMessage} implementation.
      */
     default EventExecutor defaultSubscriberExecutor() {
-        return RequestContext.mapCurrent(RequestContext::eventLoop, () -> CommonPools.workerGroup().next());
+        final EventLoop eventExecutor = RequestContext.mapCurrent(RequestContext::eventLoop,
+                                                                  () -> CommonPools.workerGroup().next());
```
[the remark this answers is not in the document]
minwoox: Oh, I missed this. Thanks!

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
minwoox: I used long because `Flags.defaultMaxResponseLength()` uses long. Do you want me to change that as well?

trustin: `defaultMaxResponseLength` must be long.

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageDuplicator.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public interface StreamMessageDuplicator<T> {
+
+    /**
+     * Returns a new {@link StreamMessage} that publishes the same elements with the {@link StreamMessage}
+     * that this duplicator is created from.
+     *
+     * @param lastStream whether to prevent further duplication
+     */
+    StreamMessage<T> duplicate(boolean lastStream);
```
anuraaga: Do we need this? From what I understand from the javadocs, these two are equivalent

```java
duplicator.duplicate(true);
duplicator.close();
```

```java
duplicator.duplicate();
duplicator.close();
```

minwoox: Yes, they are equivalent.
Actually, we don't have to call `close` in the first example.
```java
duplicator.duplicate(true).subscribe(...);
// Equivalent to 
duplicator.duplicate().subscribe(...);
duplicator.close();
```

We need the first code for the response in `RetryingClient`. We have to return the last `HttpResonse` to the user API, so when the last response is subscribed and complete, the resources are cleaned up. (We don't know when we need to call `close()` because we don't know the duplicate `HttpResponse` is subscribed by the user or not.)

We need the second one for the request. The duplicated `HttpRequest` is already subscribed by the `HttpRequestSubscriber` and we don't want to duplicate anymore so just call `close()`.

core/src/main/java/com/linecorp/armeria/common/HttpResponse.java
minwoox: So it's good to keep it as it is? @ikhoon

ikhoon: > I used long because `Flags.defaultMaxResponseLength()` uses long.

The `Flags.defaultMaxResponseLength()` enventually cast `long` to `int`. So I thought the following because it can expose stricter type.
```java
@Override
default HttpResponseDuplicator toDuplicator(EventExecutor executor, int maxResponseLength) 
   return new DefaultHttpResponseDuplicator(this, executor, maxResponseLength);
}

@Override
default HttpResponseDuplicator toDuplicator(EventExecutor executor) 
    int maxResponseLength = Flags.defaultMaxResponseLength() > Integer.MAX_VALUE ? Integer.MAX_VAUE : Flags.defaultMaxResponseLength();
    return toDuplicator(executor, maxResponseLength);
}
```

But I'm not strong here. Let keep it!

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageDuplicator.java
anuraaga: But `close` javadoc says that duplicated streams are not cleaned up until they finish publication. So shouldn't it be fine to replace any instance of `duplicate(true)` with `duplicate()` and `close()`, even for `RetryingClient`?

The class-level javadoc recommends always calling `close` to make sure resources are cleaned up, which makes sense, but in that case, if `duplicate(true)` is equivalent, I think we should remove it if it's redundant.

minwoox: > duplicate(true) with duplicate() and close(), even for RetryingClient response?

The problem is this throws an exception:
```java
StreamMessage message = duplicator.duplicate();
duplicator.close();
message.subscribe(...) // This throw an exception because the duplicator is closed already.
```
So I have to change that a subscriber is able to subscribe to the duplicated stream which is created before `duplicator.close()` is called.

Let me see what I can do. 🤔

anuraaga: If it's possible to support that, I think that could be cleanest, and it seems doable - it's normal for a stream that's not subscribed or aborted to be leaked, so it's not necessary for `close` to be cleaning up unsubscribed streams. 

But otherwise, now I understand the example better - so if we can't fix that behavior, the javadoc for close needs to be extended to mention that any streams that haven't been subscribed to will be cleaned up right away, which isn't the impression I got from the current javadoc.

minwoox: > If it's possible to support that, I think that could be cleanest, and it seems doable

Yes, it's much cleaner and I think I need to fix that. 😄 Thanks!

minwoox: @anuraaga Changed the logic as you suggested. PTAL 🙇

ikhoon: I guess some users might want to release the duplicator automatically. How about extending `SafeCloseable`?
```java
public interface HttpRequestDuplicator extends StreamMessageDuplicator<HttpObject>, SafeCloseable {
    ...
    @Override
    void close();
}
```

core/src/main/java/com/linecorp/armeria/common/DefaultHttpRequestDuplicator.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    private class DuplicatedHttpRequest
+            extends StreamMessageWrapper<HttpObject> implements HttpRequest {
+
+        private final RequestHeaders headers;
+
+        DuplicatedHttpRequest(StreamMessage<? extends HttpObject> delegate, RequestHeaders headers) {
+            super(delegate);
+            this.headers = headers;
```
anuraaga: Is it ok not to filter the stream so these are published?

core/src/main/java/com/linecorp/armeria/common/stream/StreamMessageDuplicator.java
minwoox: That's a good idea! Fixed. Thanks!

core/src/main/java/com/linecorp/armeria/common/HttpRequestDuplicator.java
```diff
@@ -16,128 +16,51 @@
...
-    public HttpRequest duplicateStream(RequestHeaders newHeaders) {
-        return duplicateStream(newHeaders, false);
-    }
+    HttpRequest duplicate();
 
     /**
-     * Creates a new {@link HttpRequest} instance that publishes data from the {@code publisher} you create
-     * this factory with. If you specify the {@code lastStream} as {@code true}, it will prevent further
-     * creation of duplicate stream.
+     * Returns a new {@link HttpRequest} with the specified {@link RequestHeaders} that publishes the same
```
anuraaga: Related to above it's not clear what it means to have specified headers but publish same objects

core/src/main/java/com/linecorp/armeria/common/DefaultHttpRequestDuplicator.java
minwoox: The elements of the `delegate` are from the `HttpRequest` that this duplicator is created from by the contract. `HttpRequest` does not publish `RequestHeaders`, (`HttpRequest.requestHeaders()`) so I think this is OK. 😉

core/src/main/java/com/linecorp/armeria/common/HttpRequestDuplicator.java
minwoox: How about this?
```java
* Returns a new {@link HttpRequest} with the specified {@link RequestHeaders} that publishes the same
* {@link HttpData}s and {@linkplain HttpHeaders trailers} with the {@link HttpRequest} that
* this duplicator is created from.
```

anuraaga: Yeah I think that's clearer and nice, by the way `with the` -> `as the`

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessageDuplicator.java
```diff
@@ -44,129 +44,79 @@
...
- * you do not need the contents anymore, otherwise memory leak might happen.</p>
+ * A default duplicator.
  *
  * @param <T> the type of elements
- * @param <U> the type of the upstream {@link StreamMessage} and duplicated {@link StreamMessage}s
+ * @see StreamMessageDuplicator
  */
-public abstract class AbstractStreamMessageDuplicator<T, U extends StreamMessage<T>>
-        implements SafeCloseable {
+public class DefaultStreamMessageDuplicator<T> implements StreamMessageDuplicator<T>, SafeCloseable {
```
anuraaga: Do we need the `Default*` to be public?

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessageDuplicator.java
```diff
@@ -388,8 +331,8 @@ private void doUnsubscribe(DownstreamSubscription<T> subscription, @Nullable Thr
         }
 
         private void doCleanupIfLastSubscription() {
-            if (isLastDownstreamAdded() && downstreamSubscriptions.isEmpty()) {
-                state = State.CLOSED;
+            if (isClosed() && duplicator.unsubscribed == 0 && downstreamSubscriptions.isEmpty()) {
```
anuraaga: The thread safety of this `unsubscribed == 0` check isn't so obvious so I think we can add a comment. Maybe

`// Because the duplicator is closed, we know that unsubscribed will not be incremented anymore and are guaranteed that the last unsubscribed downstream will run this cleanup logic.`

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessageDuplicator.java
minwoox: Yes, `HttpRequestDuplicator` and `HttpResponseDuplicator` are in different package.
So I was going to move this to the internal package but a lot of other files in the stream package have to come together. I will do it after this is merged. 😄

core/src/main/java/com/linecorp/armeria/common/HttpRequestDuplicator.java
minwoox: Oops Thank!

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessageDuplicator.java
minwoox: That's a good suggestion. Added. 😄

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessageDuplicator.java
anuraaga: Thanks - sounds like you're signing yourself up for https://github.com/line/armeria/issues/2360 ? 😆

minwoox: Well, I was actually just going to handle the classes under `stream` package.😆
If you want I can take it. 😄 Or maybe we can divide the packages and each takes care of them. Either way is fine for me. 😎

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessageDuplicator.java
minwoox: Anyway, I am working on #2189 and it will be done today or tomorrow. (I hope 😆 )

minwoox:
Thank you all for reviewing this. ❤️

PR #2411, opened 2020-01-20
Prevent ProGuard from pruning Reactor `NonBlocking`

trustin:
Related: #2404
Motivation:

We allowed Proguard to prune all classes that are not under
`com.linecorp.armeria` (except the internal ones). This can make our own
forked tag interface `reactor.core.scheduler.NonBlocking` to be pruned
by ProGuard, although it is currently kept indirectly and automatically.
It'll be safer if we configure ProGuard explicitly.

Modifications:

- Add ProGuard `keep` for `reactor.core.scheduler.NonBlocking`

Result:

- No chance of regression


trustin:
Thanks!

PR #2412, opened 2020-01-21
Auto-generate gRPC examples with default fields

trustin:
Motivation:

A user sometimes uses a default gRPC `Message` instance as an example
request when building a `DocService`:

    DocService.builder()
              .exampleRequestForMethod(MyServiceGrpc.SERVICE_NAME,
                                       "DoSomething",
                                       SomeRequest.getDefaultInstance())

.. which is a tedious job that can be automated.

Modifications:

- Make `GrpcDocServicePlugin` append an example request generated from
  the default instance.
  - The frontend will show the user-specified example first.

Result:

- A user does not need to set the default gRPC examples.


trustin:
Needs to fix a test case but the implementation is working as expected.
I wonder if this is useful when no example is given by a user. Thoughts?

core/src/main/java/com/linecorp/armeria/server/docs/DocService.java
```diff
@@ -318,14 +318,20 @@ private ServiceInfo addServiceExamples(ServiceInfo service) {
...
+                        concatAndDedup(exampleHttpHeaders.get(m.name()), m.exampleHttpHeaders()),
+                        concatAndDedup(exampleRequests.get(m.name()), m.exampleRequests()),
                         m.httpMethod(), m.docString()))::iterator,
                 Iterables.concat(service.exampleHttpHeaders(),
                                  exampleHttpHeaders.get("")),
                 service.docString());
     }
 
+    private static <T> Iterable<T> concatAndDedup(Iterable<T> first, Iterable<T> second) {
+        return Streams.stream(Iterables.concat(first, second)).distinct()::iterator;
```
anuraaga: A bit more characters, but guess we may just stick to Stream API for simplicity `Stream.concat(Streams.stream(first), Streams.stream(second)).distinct()`

trustin:
@anuraaga Do you think we should serialize the default fields for user-specified examples? Not sure which is a sensible behavior. Would it be nice if a user can specify an option for a plugin?

anuraaga:
Yeah I don't think I actually thought through that part of the behavior - serializing default values for all examples seems like a good behavior to me.

trustin:
Ready for reviews

trustin:
@ikhoon Should be fixed now.

ikhoon:
LGTM 👍

PR #2413, opened 2020-01-22
Add `RequestLogProperty.NAME`

### LINKED ISSUE #1931: Add RequestLog.setName or similar
In https://github.com/openzipkin/zipkin/pull/2709 I found it's very tedious to set a name to the current request in a client, since it requires concepts like `RpcRequest` and `originalRequest`. Having a simple way to set the name of the current request might be nice.

trustin:
Related: #1931
Motivation:

A user may want to specify the name of a request manually, rather than
relying on `RpcRequest.method()`.

Modifications:

- Add `RequestLogProperty.NAME`, `RequestOnlyLog.name()` and
  `RequestLogBuilder.name(String)`.
- Add `BuiltInProperty.REQ_NAME`
- Set `RequestLog.name()` for:
  - Thrift clients and services
  - gRPC clients and services
  - Retrofit clients
  - Annotated services
- Use `RequestLog.name()` instead of `RpcRequest.method()` wherever
  applicable.
  - `MeterIdPrefixFunction`
  - `AccessLogComponent`
  - `{Client,Service}RequestContextAdapter`
- Make `MeterIdPrefixFunction.activeRequestPrefix()` is invoked when
  `REQUEST_START_TIME`, `REQUEST_HEADERS`, `SESSION` and `NAME` are
  available.
  - `REQUEST_CONTENT` is not available anymore.
- Miscellaneous:
  - Give up counting `StringBuilder` capacity and use a thread-local
    buffer in `DefaultRequestLog`.
  - Fix a regression in `Brave{Client,Service}` where thread-local
    context is not pushed.
  - Fix blocking operation warning in `BraveIntegrationTest`

Result:

- Closes #1931
- A user can specify a custom request name.


core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -642,6 +646,25 @@ public Scheme scheme() {
...
+    @Nullable
+    public String name() {
+        ensureAvailable(RequestLogProperty.NAME);
+        return name;
+    }
+
+    @Override
+    public void name(String name) {
+        requireNonNull(name, "name");
+        checkArgument(!name.isEmpty(), "name is empty.");
```
minwoox: nit: we can combine these two using `isNullOrEmpty`

core/src/test/java/com/linecorp/armeria/internal/metric/RequestMetricSupportTest.java
```diff
@@ -56,18 +51,21 @@ void httpSuccess() {
         final ClientRequestContext ctx = setupClientRequestCtx(registry);
 
         setConnectionTimings(ctx);
-        ctx.logBuilder().requestHeaders(RequestHeaders.of(HttpMethod.POST, "/foo"));
-        ctx.logBuilder().requestFirstBytesTransferred();
-        ctx.logBuilder().requestContent(null, null);
-        ctx.logBuilder().requestLength(123);
+        // FIXME(trustin): In reality, most HTTP requests will not have any name.
+        //                 As a result, `activeRequestMeterId()` will be invoked only after
```
minwoox: You mean `activeRequestMetrics.decrement()`?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
trustin: Nope. Can't throw an NPE with it.

core/src/test/java/com/linecorp/armeria/internal/metric/RequestMetricSupportTest.java
trustin: I meant `MeterIdPrefixFunction.activeRequestMeterId()`.

minwoox: Then it's `activeRequestPrefix`? 😆

trustin: Ah yeah :-)

trustin:
Thanks for reviewing.

PR #2415, opened 2020-01-22
Fix a bug where callbacks are invoked with the wrong context

### LINKED ISSUE #2414: Unexpected Exception when pushing`RequestContext`
While testing gRPC server & client, I saw the following exception with `WARN` message. 
```java
[armeria-common-worker-nio-2-1] WARN  i.n.util.concurrent.DefaultPromise - An exception was thrown by com.linecorp.armeria.server.HttpResponseSubscriber$$Lambda$630/0x000000080153f040.operationComplete()
java.lang.IllegalStateException: Trying to call object wrapped with context [S][id: 0x36589007, L:/127.0.0.1:54739 - R:/127.0.0.1:54740][h2c://ihi.local:54739/armeria.GithubService/Simple#POST], but context is currently set to [S][id: 0x36589007, L:/127.0.0.1:54739 - R:/127.0.0.1:54740][h2c://ihi.local:54739/armeria.GithubService/Simple#POST]. This means the callback was called from unexpected thread or forgetting to close previous context.
	at com.linecorp.armeria.internal.RequestContextUtil.newIllegalContextPushingException(RequestContextUtil.java:46)
	at com.linecorp.armeria.server.ServiceRequestContext.push(ServiceRequestContext.java:228)
	at com.linecorp.armeria.common.logging.RequestLogListenerInvoker.invokeOnRequestLog(RequestLogListenerInvoker.java:58)
	at com.linecorp.armeria.common.logging.DefaultRequestLog.updateAvailability(DefaultRequestLog.java:985)
	at com.linecorp.armeria.common.logging.DefaultRequestLog.endResponse0(DefaultRequestLog.java:961)
	at com.linecorp.armeria.common.logging.DefaultRequestLog.endResponse0(DefaultRequestLog.java:940)
	at com.linecorp.armeria.common.logging.DefaultRequestLog.endResponse(DefaultRequestLog.java:921)
	at com.linecorp.armeria.server.HttpResponseSubscriber.lambda$write$1(HttpResponseSubscriber.java:325)
	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:577)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners0(DefaultPromise.java:570)
	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:549)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:490)
	at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:615)
	at io.netty.util.concurrent.DefaultPromise.setSuccess0(DefaultPromise.java:604)
	at io.netty.util.concurrent.DefaultPromise.trySuccess(DefaultPromise.java:104)
	at io.netty.handler.codec.http2.Http2CodecUtil$SimpleChannelPromiseAggregator.tryPromise(Http2CodecUtil.java:383)
	at io.netty.handler.codec.http2.Http2CodecUtil$SimpleChannelPromiseAggregator.trySuccess(Http2CodecUtil.java:349)
	at io.netty.handler.codec.http2.Http2CodecUtil$SimpleChannelPromiseAggregator.trySuccess(Http2CodecUtil.java:261)
	at io.netty.util.internal.PromiseNotificationUtil.trySuccess(PromiseNotificationUtil.java:48)
	at io.netty.channel.ChannelOutboundBuffer.safeSuccess(ChannelOutboundBuffer.java:715)
	at io.netty.channel.ChannelOutboundBuffer.remove(ChannelOutboundBuffer.java:270)
	at io.netty.channel.ChannelOutboundBuffer.removeBytes(ChannelOutboundBuffer.java:350)
	at io.netty.channel.socket.nio.NioSocketChannel.doWrite(NioSocketChannel.java:431)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush0(AbstractChannel.java:931)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.flush0(AbstractNioChannel.java:354)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush(AbstractChannel.java:898)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.flush(DefaultChannelPipeline.java:1384)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:749)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:741)
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:727)
	at io.netty.handler.flush.FlushConsolidationHandler.flushNow(FlushConsolidationHandler.java:206)
	at io.netty.handler.flush.FlushConsolidationHandler.flush(FlushConsolidationHandler.java:140)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:749)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:741)
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:727)
	at com.linecorp.armeria.internal.TrafficLoggingHandler.flush(TrafficLoggingHandler.java:49)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:749)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:741)
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:727)
	at io.netty.handler.codec.http2.Http2ConnectionHandler.flush(Http2ConnectionHandler.java:189)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:749)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:741)
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:727)
	at com.linecorp.armeria.server.HttpResponseSubscriber.write(HttpResponseSubscriber.java:342)
	at com.linecorp.armeria.server.HttpResponseSubscriber.onNext(HttpResponseSubscriber.java:225)
	at com.linecorp.armeria.server.HttpResponseSubscriber.onNext(HttpResponseSubscriber.java:59)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.notifySubscriberWithElements(DefaultStreamMessage.java:341)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.notifySubscriber0(DefaultStreamMessage.java:319)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.notifySubscriber(DefaultStreamMessage.java:252)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.addObjectOrEvent(DefaultStreamMessage.java:238)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.addObject(DefaultStreamMessage.java:158)
	at com.linecorp.armeria.common.stream.AbstractStreamMessageAndWriter.tryWrite(AbstractStreamMessageAndWriter.java:74)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.tryWrite(DefaultStreamMessage.java:63)
	at com.linecorp.armeria.common.stream.StreamWriter.write(StreamWriter.java:71)
	at com.linecorp.armeria.server.grpc.ArmeriaServerCall.doSendHeaders(ArmeriaServerCall.java:242)
	at com.linecorp.armeria.server.grpc.ArmeriaServerCall.sendHeaders(ArmeriaServerCall.java:208)
	at io.grpc.stub.ServerCalls$ServerCallStreamObserverImpl.onNext(ServerCalls.java:348)
	at com.linecorp.armeria.it.grpc.GithubApiService.simple(GithubApiService.java:47)
	at com.linecorp.armeria.grpc.GithubServiceGrpc$MethodHandlers.invoke(GithubServiceGrpc.java:277)
	at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:172)
	at com.linecorp.armeria.server.grpc.ArmeriaServerCall.invokeHalfClose(ArmeriaServerCall.java:436)
	at com.linecorp.armeria.server.grpc.ArmeriaServerCall.endOfStream(ArmeriaServerCall.java:429)
	at com.linecorp.armeria.common.grpc.protocol.ArmeriaMessageDeframer.close(ArmeriaMessageDeframer.java:313)
	at com.linecorp.armeria.common.grpc.protocol.ArmeriaMessageDeframer.closeWhenComplete(ArmeriaMessageDeframer.java:293)
	at com.linecorp.armeria.internal.grpc.HttpStreamReader.closeDeframer(HttpStreamReader.java:204)
	at com.linecorp.armeria.internal.grpc.HttpStreamReader.apply(HttpStreamReader.java:183)
	at com.linecorp.armeria.internal.grpc.HttpStreamReader.apply(HttpStreamReader.java:48)
	at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:930)
	at java.base/java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:907)
	at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:478)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:510)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:518)
	at io.netty.util.concurrent.SingleThreadEventExecutor$6.run(SingleThreadEventExecutor.java:1050)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at java.base/java.lang.Thread.run(Thread.java:835)
```

The error can be reproduced by executing the `simpleNonBlocking` test with 20% chance.
https://github.com/line/armeria/blob/08afa31fbe512e65a2e38bfcd6a3315ed313c241/grpc/src/test/java/com/linecorp/armeria/it/grpc/DownstreamSimpleBenchmarkTest.java#L95-L113

minwoox:
Motivation:
When we send a data, we don't buffer but flush them to the socket one by one.
However, under heavy load, especially when using HTTP/2, there's a chance that the data is not fully written(partial write) or nor written at all because of the lack of the space in sending buffer. http://man7.org/linux/man-pages/man2/write.2.html
Even though we try to push the data to the socket while spinning (16 times by default), it could fail.
In that case, we just store data to `ChannelOutboundBuffer` and try it next time when writing other data.

This is where the trouble begins. The next write that would be handling a different request, which means different `RequestContext` in the thread-local, completes the writing and executes all callbacks. Some of the callback should not be invoked with the `RequestContext`.

This could happen to some of the other `ChannelFuture`s used, so I updated them as well.

Modification:
- Remove the `RequestContext` if exists in `ChannelFuture` listener.

Result:
- Callbacks are not invoked with the wrong context in the thread-local
- Close #2414

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
@@ -49,5 +49,18 @@ public final class RequestContextUtil {
                 "unexpected thread or forgetting to close previous context.");
     }
 
+    /**
+     * Removes the {@link RequestContext} in the thread-local if exists and returns {@link SafeCloseable} which
+     * pushes it back to the thread-local.
+     */
+    public static SafeCloseable pop() {
+        final RequestContext oldCtx = RequestContextThreadLocal.getAndRemove();
+        if (oldCtx == null) {
+            return noopSafeCloseable();
+        }
+
+        return () -> RequestContextThreadLocal.set(oldCtx);
+    }
+
     private RequestContextUtil() {}
 }
--- a/core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
@@ -21,6 +21,9 @@ import static java.util.Objects.requireNonNull;
 import com.linecorp.armeria.common.RequestContext;
 import com.linecorp.armeria.common.util.SafeCloseable;
 
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
+
 /**
  * Utilities for {@link RequestContext}.
  */
@@ -51,7 +54,12 @@ public final class RequestContextUtil {
 
     /**
      * Removes the {@link RequestContext} in the thread-local if exists and returns {@link SafeCloseable} which
-     * pushes it back to the thread-local.
+     * pushes the {@link RequestContext} back to the thread-local.
+     *
+     * <p>Because this method pops the {@link RequestContext} arbitrarily, it shouldn't be used in
+     * most cases. One of the examples this can be used is in {@link ChannelFutureListener}.
+     * The {@link ChannelFuture} can be complete when the eventloop handles the different request. The
+     * eventloop might have the wrong {@link RequestContext} in the thread-local, so we should pop it.
      */
     public static SafeCloseable pop() {
         final RequestContext oldCtx = RequestContextThreadLocal.getAndRemove();
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
@@ -31,6 +31,7 @@ import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.logging.RequestLogBuilder;
+import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.armeria.internal.PathAndQuery;
 
 import io.netty.channel.Channel;
@@ -203,10 +204,15 @@ final class HttpClientDelegate implements HttpClient {
 
     private static void handleEarlyRequestException(ClientRequestContext ctx,
                                                     HttpRequest req, Throwable cause) {
-        req.abort(cause);
-        final RequestLogBuilder logBuilder = ctx.logBuilder();
-        logBuilder.endRequest(cause);
-        logBuilder.endResponse(cause);
+        // This can be executed by the same eventloop which is holding a different context
+        // because the future can be complete while the eventloop is dealing another request.
+        // So we should set the reqCtx explicitly.
+        try (SafeCloseable ignored = ctx.replace()) {
+            req.abort(cause);
+            final RequestLogBuilder logBuilder = ctx.logBuilder();
+            logBuilder.endRequest(cause);
+            logBuilder.endResponse(cause);
+        }
     }
 
     private void doExecute(PooledChannel pooledChannel, ClientRequestContext ctx,
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
@@ -205,9 +205,6 @@ final class HttpClientDelegate implements HttpClient {
 
     private static void handleEarlyRequestException(ClientRequestContext ctx,
                                                     HttpRequest req, Throwable cause) {
-        // This can be executed by the same eventloop which is holding a different context
-        // because the future can be complete while the eventloop is dealing another request.
-        // So we should pop the ctx temporarily.
         try (SafeCloseable ignored = RequestContextUtil.pop()) {
             req.abort(cause);
             final RequestLogBuilder logBuilder = ctx.logBuilder();
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
@@ -33,6 +33,7 @@ import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.logging.RequestLogBuilder;
 import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.armeria.internal.PathAndQuery;
+import com.linecorp.armeria.internal.RequestContextUtil;
 
 import io.netty.channel.Channel;
 import io.netty.channel.EventLoop;
@@ -206,8 +207,8 @@ final class HttpClientDelegate implements HttpClient {
                                                     HttpRequest req, Throwable cause) {
         // This can be executed by the same eventloop which is holding a different context
         // because the future can be complete while the eventloop is dealing another request.
-        // So we should set the ctx explicitly.
-        try (SafeCloseable ignored = ctx.replace()) {
+        // So we should pop the ctx temporarily.
+        try (SafeCloseable ignored = RequestContextUtil.pop()) {
             req.abort(cause);
             final RequestLogBuilder logBuilder = ctx.logBuilder();
             logBuilder.endRequest(cause);
--- a/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpClientDelegate.java
@@ -206,7 +206,7 @@ final class HttpClientDelegate implements HttpClient {
                                                     HttpRequest req, Throwable cause) {
         // This can be executed by the same eventloop which is holding a different context
         // because the future can be complete while the eventloop is dealing another request.
-        // So we should set the reqCtx explicitly.
+        // So we should set the ctx explicitly.
         try (SafeCloseable ignored = ctx.replace()) {
             req.abort(cause);
             final RequestLogBuilder logBuilder = ctx.logBuilder();
--- a/core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
@@ -46,6 +46,7 @@ import com.linecorp.armeria.common.stream.ClosedPublisherException;
 import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.armeria.internal.HttpObjectEncoder;
+import com.linecorp.armeria.internal.RequestContextUtil;
 
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
@@ -111,8 +112,8 @@ final class HttpRequestSubscriber implements Subscriber<HttpObject>, ChannelFutu
 
         // This can be executed by the same eventloop which is holding a different context
         // because the future can be complete while the eventloop is dealing another request.
-        // So we should set the reqCtx explicitly.
-        try (SafeCloseable ignored = reqCtx.replace()) {
+        // So we should pop the ctx temporarily.
+        try (SafeCloseable ignored = RequestContextUtil.pop()) {
             if (future.isSuccess()) {
                 // The first write is always the first headers, so log that we finished our first transfer
                 // over the wire.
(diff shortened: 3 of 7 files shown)


minwoox:
Will update the description soon.

minwoox:
Just added `ctx.replace()` where it has to be. I'm not so sure this is the best I can do. 🤔

minwoox:
The exception is raised when notifying log listeners. (Even though we changed the API to return `CompletableFuture`, we should fix that.) I have thought about pushing the context before notifying listeners but we still have to push explicitly in some situations, so I gave up.

anuraaga:
If I'm not mistaken, I guess this only happens when using `makeContextAware` as opposed to `contextAwareExecutor` since the latter will be a fresh callback on the event loop and not invoked inline.

It's a bit unfortunate we have to mount and unmount for the latter case to fix the former but guess it's the best we can do.

Instead of putting the mount all over the place, is it possible to only do it once when handling the channel write future?

minwoox:
> If I'm not mistaken, I guess this only happens when using makeContextAware as opposed to contextAwareExecutor since the latter will be a fresh callback on the event loop and not invoked inline.

Well, it's nothing to do with those. The first context is pushed [here](https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java#L435).
And it triggers to push another context in the log listener. 🤔

minwoox:
> Instead of putting the mount all over the place, is it possible to only do it once when handling the channel write future?

Well, that's what I did. 😄

anuraaga:
Yeah was trying to come up with an example of a user callback having the same issue - realized futures are generally completed on reads, not writes and are not related. I think a subscriber may still have a weird context when calling request and write, but not sure.

For just one location, I was thinking that since we only have one class that writes, we could probably handle this issue there?

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/HttpObjectEncoder.java#L35

But I don't know if it's a good idea. Before, it was still easy to reason about when context is mounted - HTTP invocation, RPC-framework callback (since no access to ctx), and log callback (I guess mostly for MDC). Now I don't really have any idea - I think some subscriber callbacks have it and others don't but it's very hard to reason about when done at such a low level.

Maybe we can go ahead and remove the check / exception when pushing context. I guess if a user uses something like our propagation examples, they shouldn't run into incorrect context. And if not, they'll probably have problems anyways with or without the check.

minwoox:
> For just one location, I was thinking that since we only have one class that writes, we could probably handle this issue there?

Unfortunately, it didn't happen there. 😭 It's happened when calling `ChannelHandlerContext.flush()` which means we cannot handle it in `HttpObjectEncoder`.

> Before, it was still easy to reason about when context is mounted - HTTP invocation, RPC-framework callback (since no access to ctx), and log callback (I guess mostly for MDC). Now I don't really have any idea - I think some subscriber callbacks have it and others don't but it's very hard to reason about when done at such a low level.

I agree. This could make it complicated.

> Maybe we can go ahead and remove the check / exception when pushing context.

I'm not sure about this. It helped me find out that this thing which is not going to happen after #2342. If we couldn't find this, in the worst situation, a user who attaches a callback which Is using the `RequestContext` in the thread-local, could see different context and it's a disaster.
So I think it's just not for the users but help us to find when something goes wrong. 🤔

anuraaga:
Can you maybe provide a code example of that mistake? Not sure if it's worse than other context propagation mistakes that result in lost context without an exception. I'm feeling that any callback that uses context needs be to a context aware callback, and if it isn't all bets are off.

minwoox:
>  I'm feeling that any callback that uses context needs be to a context aware callback, and if it isn't all bets are off.

Yes, I agree. 😄 We did change to return future now, it must be. (We don't know if the future is complete or now when adding callback. So the user must handle it.)

> Can you maybe provide a code example of that mistake?

Well it's just adding a decorator with not much thinking
```java
ctx.log().whenComplete().thenApply(log -> {
   RequestContext ctx = RequestContext.mapCurrent(...);
   ...
});
```
Of course, a user do not have to this and use the context aware future.
```java
ctx.log().whenComplete().thenApplyAsync(log -> {
    // do the job...
}, RequestContext.current().contextAwareExecutor);
```

> But I don't know if it's a good idea. Before, it was still easy to reason about when context is mounted - HTTP invocation, RPC-framework callback (since no access to ctx), and log callback (I guess mostly for MDC). Now I don't really have any idea - I think some subscriber callbacks have it and others don't but it's very hard to reason about when done at such a low level.

How about just popping rather replacing it and then put it back when the thread gets out of the try-with-resousce?
So it's mounted as it was except log callback.

anuraaga:
I might argue the first snippet is itself somewhat broken code since it uses the context in a callback that isn't context aware.

But yeah, if it's possible, the idea of just popping instead of replacing is easier to reason about and seems good to me.

minwoox:
> I might argue the first snippet is itself somewhat broken code since it uses the context in a callback that isn't context aware.

Yeah, it's broken and a user should never do that like that. 😉 

> But yeah, if it's possible, the idea of just popping instead of replacing is easier to reason about and seems good to me.

Then, I will do it. Thanks! 😄

minwoox:
Added internal API which removes the `RequestContext` temporarily. 😉

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
```diff
@@ -49,5 +49,18 @@ public static IllegalStateException newIllegalContextPushingException(
                 "unexpected thread or forgetting to close previous context.");
     }
 
+    /**
+     * Removes the {@link RequestContext} in the thread-local if exists and returns {@link SafeCloseable} which
+     * pushes it back to the thread-local.
```
trustin: How about mentioning the specific use case of this method? We could then remove the duplicate comments in this pull request.

trustin:
Thanks @minwoox and @anuraaga for good discussion.

minwoox:
> Thanks @minwoox and @anuraaga for good discussion.

@anuraaga is the best. 👍

core/src/main/java/com/linecorp/armeria/internal/RequestContextUtil.java
minwoox: Ah that's a good idea. Let me add it.

minwoox:
Thanks for reviewing!

PR #2420, opened 2020-01-28
Use modern features of Gradle instead of custom scripts where possible.

anuraaga:
Figured a new decade is a good time to rewrite the Gradle scripts :) Also a good time since changing Maven scopes is sort of a breaking change.

If this PR looks good I or someone else could copy the script directory changes into `gradle-scripts`. I'm not sure what repos other than central dogma that's used in though - these changes are somewhat, though not highly breaking I think - add `publish` to BOM, use `strictly` for downgrading dependencies). Projects should still be able to use `compile` if they want though not completely sure.

- Migrate from `java` to `java-library` plugin which is recommended for libraries
- Migrate from deprecated `compile` to `api` / `implementation`. Former is compile classpath, latter is runtime classpath - https://docs.gradle.org/current/userguide/java_plugin.html#tab:configurations
- Use `configureFeature` fo Gradle native support of optional dependencies
- Don't generate POM files. With above changes, Gradle correctly generates POM files
  - Except do need to strip out shaded dependencies since that's on us
- Replace spring dependency management with Gradle's `java-platform`
  - Overrides now use `strictly` version overriding, which is verified by Gradle so we don't need to verify it ourselves
- Switch to `compileOnly` for optional thrift dependency. It maps to `provided` maven scope, which is more appropriate for a required provided dependency
- Use Gradle's `java-platform` to define BOM
- Only evaluate runtime classpaths when creating shaded test runtime classpath. Compile classpath includes compile-only deps which should not be on the classpath when running tests (unless explicitly added). Normal `test` has this behavior and now `shadedTest` is a better approximation of it.

Result:
- Cleaner POMs. Our current POM generation seems to lack some sophistication and adds all exclude rules to all dependencies. Gradle attaches in a way that corresponds to the actual exclusions for example
```
    <dependency>
      <groupId>org.slf4j</groupId>
      <artifactId>slf4j-api</artifactId>
      <version>1.7.29</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>com.linecorp.armeria</groupId>
      <artifactId>armeria-grpc-protocol</artifactId>
      <version>0.97.1-SNAPSHOT</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>com.google.protobuf</groupId>
      <artifactId>protobuf-java-util</artifactId>
      <version>3.10.0</version>
      <scope>runtime</scope>
      <exclusions>
        <exclusion>
          <artifactId>error_prone_annotations</artifactId>
          <groupId>com.google.errorprone</groupId>
        </exclusion>
      </exclusions>
    </dependency>
```
- Less leakage into user compile classpath by specifying `implementation` dependencies which are necessary when compiling Armeria, but not user code
- Less code / 3rd party plugins by taking advantage of Gradle native functionality
- Probably sightly faster builds
- Most importantly? Fixing `thrift0.12` project not being actually used :-D


gradle/scripts/lib/common-dependencies.gradle
```diff
@@ -25,45 +25,96 @@ rootProject.ext {
...
+}
+
+def dependencyProjects = projectsWithFlags('dependencies')
+if (!dependencyProjects.isEmpty()) {
+    def dependencyProject = dependencyProjects[0]
+    configure(projectsWithFlags('java')) {
+        // We need to use afterEvaluate because there is no way to guarantee configuration.canBeResolved and
+        // canBeConsumed are set otherwise.
+        afterEvaluate {
+            evaluationDependsOn(dependencyProject.path)
```
trustin: Shouldn't we call `evaluationDependsOn(...)` from the outside of `afterEvaluate { .. }`?

gradle/scripts/lib/java-publish.gradle
```diff
@@ -2,168 +2,101 @@ configure(projectsWithFlags('publish', 'java')) {
...
-                    mainJarTask = tasks.trimShadedJar
-                } else if (tasks.findByName('shadedJar')) {
-                    mainJarFile = tasks.shadedJar.archivePath
-                    mainJarTask = tasks.shadedJar
-                } else {
-                    mainJarFile = tasks.jar.archivePath
-                    mainJarTask = tasks.jar
+configure(projectsWithFlags('publish')) {
+    publishing {
+        publications.each {publication->
```
trustin: nit: space after `{` and before `->`

gradle/scripts/lib/java.gradle
```diff
@@ -48,45 +59,12 @@ configure(projectsWithFlags('java')) {
...
-        from sourceSets.main.allSource
-    }
-
-    // Generate a javadoc JAR.
-    task javadocJar(type: Jar, dependsOn: tasks.javadoc) {
-        classifier = 'javadoc'
-        from javadoc.destinationDir
-    }
-
-    // Generate source/javadoc JARs only when publishing.
```
trustin: Do we now always generate source/javadoc JARs?

settings.gradle
```diff
@@ -40,11 +40,14 @@ includeWithFlags ':it:server',                'java', 'relocate'
 includeWithFlags ':it:spring:boot-tomcat',    'java', 'relocate'
 includeWithFlags ':it:spring:boot-tomcat8.5', 'java', 'relocate'
 includeWithFlags ':testing-internal',         'java', 'relocate'
-includeWithFlags ':thrift0.11',               'java', 'relocate'
+includeWithFlags ':thrift0.12',               'java', 'relocate'
```
trustin: Thanks! :sweat_smile:

settings.gradle
```diff
@@ -40,11 +40,14 @@ includeWithFlags ':it:server',                'java', 'relocate'
...
 
 // Unpublished non-Java projects
 includeWithFlags ':docs-client'
 
+// Unpublished dependency management project to create a virtual platform.
+includeWithFlags ':dependencies', 'dependencies'
```
trustin: Perhaps we could make this done in `settings-flags.gradle` and use the name and flag that's less likely to clash with user modules, e.g. `dependency-management`?

trustin:
How about:
- checking the current Gradle version in `prerequisite.gradle`?
- updating `README.md` if anything needs to be changed?

trustin:
No idea what this means, but getting this failure:

```
Exception in thread "main" java.lang.reflect.InvocationTargetException
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:386)
        at sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:401)
Caused by: java.util.zip.ZipException: error in opening zip file
        at java.util.zip.ZipFile.open(Native Method)
        at java.util.zip.ZipFile.<init>(ZipFile.java:225)
        at java.util.zip.ZipFile.<init>(ZipFile.java:155)
        at java.util.jar.JarFile.<init>(JarFile.java:166)
        at java.util.jar.JarFile.<init>(JarFile.java:103)
        at sun.net.www.protocol.jar.URLJarFile.<init>(URLJarFile.java:93)
        at sun.net.www.protocol.jar.URLJarFile.getJarFile(URLJarFile.java:69)
        at sun.net.www.protocol.jar.JarFileFactory.get(JarFileFactory.java:84)
        at sun.net.www.protocol.jar.JarURLConnection.connect(JarURLConnection.java:122)
        at sun.net.www.protocol.jar.JarURLConnection.getInputStream(JarURLConnection.java:152)
        at java.net.URL.openStream(URL.java:1068)
        at org.mortbay.jetty.alpn.agent.Premain.configureBootstrapClassLoaderSearch(Premain.java:115)
        at org.mortbay.jetty.alpn.agent.Premain.premain(Premain.java:97)
        ... 6 more
FAILURE: Build failed with an exception.
```

anuraaga:
Which gradle task is that?

anuraaga:
Nevermind, see you just copied from CI :)

gradle/scripts/lib/java.gradle
anuraaga: It's always generated for `assemble` (and so `build`). I find this more idiomatic and not a huge problem - it only affects a curent target so e.g., `:grpc:assemble` only builds javadoc for `grpc`, not `core`. And IntelliJ Gradle Runner doesn't build. I was a bit surprised that our CI wasn't building javadoc, and I think it's good to verify javadoc builds in CI too.

But not hard to restore the skipping behavior if you still prefer that let me know.

saml/build.gradle
```diff
@@ -1,12 +1,12 @@
...
-    compile 'net.shibboleth.utilities:java-support'
-    compile 'org.opensaml:opensaml-core'
-    compile 'org.opensaml:opensaml-saml-api'
-    compile 'org.opensaml:opensaml-saml-impl'
-    compile 'org.opensaml:opensaml-messaging-api'
-    compile 'org.opensaml:opensaml-messaging-impl'
-    compile 'org.opensaml:opensaml-soap-api'
-    compile 'org.opensaml:opensaml-soap-impl'
-    testCompile 'org.jsoup:jsoup'
+    api 'com.auth0:java-jwt'
```
anuraaga: I was somewhat surprised to find this in our public API, but looking at it realized it is tedious if we wanted to separate. Something to think about.

Also, I use `java-jwt` too and it's probably a fine choice, but worth giving a look at other libraries too I guess, maybe https://github.com/google/tink/blob/master/docs/JAVA-HOWTO.md

settings.gradle
anuraaga: Good idea!

gradle/scripts/lib/common-dependencies.gradle
anuraaga: Doh yeah

core/build.gradle
```diff
@@ -85,52 +85,49 @@ dependencies {
...
-    compile('io.prometheus:simpleclient_common') {
-        ext.optional = true
-    }
+    api 'io.micrometer:micrometer-core'
+    optionalApi 'io.micrometer:micrometer-registry-prometheus'
+    optionalApi 'io.dropwizard.metrics:metrics-core'
+    optionalApi 'io.prometheus:simpleclient_common'
 
     // Netty
     [ 'netty-transport', 'netty-codec-http2', 'netty-codec-haproxy', 'netty-resolver-dns' ].each {
```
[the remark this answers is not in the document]
anuraaga: But something to think about

anuraaga:
By the way, when you're comfortable with the code it will be great if you can check out the branch and run `./gradlew publishToMavenLocal` to check the POMs look good

anuraaga:
Build passed this time - hard to say whether I introduced some flakiness with this change or it's just Murphy's Law. But I think I only touched dependency resolution, not anything related to e.g. unzipping alpn, so leaning towards the latter (always convenient to do so :-D ).

testing-internal/src/main/java/com/linecorp/armeria/testing/internal/webapp/WebAppContainerTest.java
```diff
@@ -42,12 +42,11 @@
 import com.linecorp.armeria.client.ClientFactory;
 import com.linecorp.armeria.client.WebClient;
 import com.linecorp.armeria.common.AggregatedHttpResponse;
+import com.linecorp.armeria.common.HttpHeaderNames;
```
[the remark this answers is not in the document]
trustin: Thanks for fixing it. :bowing_man:

gradle/scripts/lib/java.gradle
trustin: OK as it is.

trustin:
@anuraaga Just checked the generated POMs and it seems to me that exclusions are not added at all.

anuraaga:
@trustin Oops, thanks for checking - it's because the exclusions got moved into `afterEvaluate` and stopped working

gradle/scripts/lib/common-dependencies.gradle
```diff
@@ -95,6 +95,16 @@ configure(dependencyManagementProject) {
 configure(projectsWithFlags('java')) {
     evaluationDependsOn(dependencyManagementProject.path)
 
+    configurations.configureEach { configuration ->
+        configuration.dependencies.whenObjectAdded { dep ->
```
trustin: This is nice. Good to know `whenObjectAdded`

trustin:
Now I see the exclusions in the POMs. Thanks!

However, if I run `./gradlew :core:dependencies`, I see the excluded dependencies are still there for some reason:

```
runtimeClasspath - Runtime classpath of source set 'main'.
...
+--- com.google.guava:guava -> 28.1-jre
|    +--- com.google.guava:failureaccess:1.0.1
|    +--- com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava
|    +--- com.google.code.findbugs:jsr305:3.0.2
|    +--- org.checkerframework:checker-qual:2.8.1 -> 2.10.0
|    +--- com.google.errorprone:error_prone_annotations:2.3.2
|    +--- com.google.j2objc:j2objc-annotations:1.3
|    \--- org.codehaus.mojo:animal-sniffer-annotations:1.18
...
```

As you see, `error_prone_annotations`, `j2objc-annotations` and `animal-sniffer-annotations` are still there. I guess some transitive dependencies revive them.. :thinking:

anuraaga:
Thanks for finding that. It looks like it's the transitive dependency from `org.reflections:reflections`, if I copy the exclusions there, it's gone.

I wonder what's the correct approach?
- Take another look to add `exclusions` to libraries like `org.reflections:reflections`
- Add exclusions to every library, which end up on every dependency in the POM, like before

I don't think I see POMs like the second pattern that much in the wild so think the former is better but not sure.

trustin:
> Add exclusions to every library

I think it makes sense to add exclusions to every library with transitive dependencies with exclusions. In this case, `reflections` needs to have `guava` exclusions.

anuraaga:
Cool - that's done :)

trustin:
I actually meant it has to be done automatically, because otherwise it will be pretty boring to figure out why a certain excluded dependency keeps showing up. :sweat_smile:

anuraaga:
@trustin Ah - I think it's doable but will be a bit complicated, do you think we can do it manually for now and I'll follow up with a PR to make that automatic?

trustin:
Sure. No problem.

jetty/build.gradle
```diff
@@ -1,7 +1,7 @@
 dependencies {
     // Jetty
-    compile 'org.eclipse.jetty:jetty-server'
+    api 'org.eclipse.jetty:jetty-server'
     ['jetty-webapp', 'jetty-annotations', 'apache-jsp', 'apache-jstl'].each {
-        testCompile "org.eclipse.jetty:$it"
+        api "org.eclipse.jetty:$it"
```
ikhoon: I wonder why this is not `testImplementation`?

anuraaga: Oops

ikhoon: Thanks :-)

Issue #1117, opened 2018-04-04
ClientDecoration order is unclear


anuraaga:
Even looking at the code, I believe it's in reverse-order of e.g., `decorator` calls to `ClientBuilder`, but I have trouble grokking to have 100% confidence. We should document the order in related Javadoc

PR #2422, opened 2020-01-29
Add content previewing client and service decorators

### LINKED ISSUE #2189: Content preview and compressed stream

minwoox:
Motivation:
Related #2189

We currently, store `HttpData`s to the `ContentPreviewer` right before writing to the socket to preview the content.
However, this makes the server decompress twice when there's an `EncodingService` and a user wants to preview decompressed content.
It's also, not the way we do for these kinds of things which are intercepting the requests and responses to manipulate or supplement some more logic.
We use decorators for that.

Modifications:
- Add `ContentPreviewingClient`, `ContentPreviewingService` and their builders.

Result:
- Close #2189
- (Breaking)
  - The methods in `ServerBuilder`, `VirtualHostBuilder`, `ServiceBindingBuilder`, `ClientBuilder` and `ClientOptionsBuilder` for setting content preview are gone.
  - You should use `ContentPreviewingClient` and `ContentPreviewingService` decorators instead.


core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewer.java
```diff
@@ -86,6 +86,13 @@ static ContentPreviewer disabled() {
         return ContentPreviewerAdapter.NOOP;
     }
 
+    /**
+     * Returns whether this {@link ContentPreviewer} is {@link #disabled()} or not.
+      */
```
anuraaga: Indent

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewer.java
```diff
@@ -86,6 +86,13 @@ static ContentPreviewer disabled() {
...
+    default boolean isDisabled() {
+        return this == ContentPreviewerAdapter.NOOP;
```
anuraaga: I think it's more idiomatic to `return false` here and have `NOOP` override with `return true`

core/src/main/java/com/linecorp/armeria/internal/logging/ContentPreviewerConfigurator.java
```diff
@@ -0,0 +1,129 @@
...
+    private final ContentPreviewerFactory responseContentPreviewerFactory;
+
+    /**
+     * Creates a new instance.
+     */
+    public ContentPreviewerConfigurator(@Nullable ContentPreviewerFactory requestContentPreviewerFactory,
+                                        @Nullable ContentPreviewerFactory responseContentPreviewerFactory) {
+        assert requestContentPreviewerFactory != null || responseContentPreviewerFactory != null;
+        this.requestContentPreviewerFactory =
+                requireNonNull(requestContentPreviewerFactory, "requestContentPreviewerFactory");
```
anuraaga: Remove `requireNonNull` for both they're `@Nullable`

core/src/main/java/com/linecorp/armeria/internal/logging/ContentPreviewerConfigurator.java
```diff
@@ -0,0 +1,129 @@
...
+                    final String status = headers.get(HttpHeaderNames.STATUS);
+                    if (ArmeriaHttpUtil.isInformational(status)) {
+                        return obj;
+                    }
+
+                    responseContentPreviewer = responseContentPreviewerFactory.get(ctx, headers);
+                    ctx.logBuilder().responseContentPreviewer(responseContentPreviewer);
+                    responseContentPreviewer.onHeaders(headers);
+                } else if (obj instanceof HttpData) {
+                    assert responseContentPreviewer != null;
```
anuraaga: Do we have some validation to make sure `ResponseHeaders` are first published object? I wonder if we are replacing a downstream HTTP protocol exception with a NPE here. I think it doesn't hurt much to just add the null check here to not worry.

core/src/main/java/com/linecorp/armeria/internal/logging/ContentPreviewerConfigurator.java
anuraaga: Though actually can we just go ahead and remove `@Nullable` throughout PR and use `ContentPreviewerFactory.disabled()` instead?

minwoox: Oops. 😅
Well, I was actually planning to remove `ContentPreviewerFactory.disabled()` and `ContentPreviewer.disabled()` because we usually use just `null` value for the disability.
The reason why we introduce `disabled()` is, I guess, because of `MappedContentPreviewerFactory` which iterates all of the factories until it finds a valid return value.
I think I want to remove that as well. 😅

`ContentPreviewer` can be created only after the request or response headers is available which means that we can pass the headers when we create `ContentPreviewer`. Then, we don't have to call `ContentPreviewer.onHeaders()` all the time.
In order to do that, we need to modify the methods which takes `Supplier<? extends ContentPreviewer>` as a parameter in `ContentPreviewerFactory`. For example, we should change `of(MediaType contentType, Supplier<? extends ContentPreviewer> supplier)` to `of(Function<? super HttpHeaders, ? extends ContentPreviewer> function)`. If we do that, I think we can remove `MappedContentPreviewerFactory`.
WDYT @anuraaga?

anuraaga: Sure that sounds good too

minwoox: Then, let me remove `requireNonNull` and use `@Nullable`.
I will do the refactoring of `ContentPreviewer` and `ContentPreviewerFactory` after this PR is merged because it could be a little complicated than I thought. 😆

minwoox:
I found out that I need sort of `responseContentPreviewDeferred`. 😅
Will fix this tomorrow.

minwoox:
This is ready. 😉

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewer.java
minwoox: Fixed. 😉

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1140,8 +1135,8 @@ private void endResponse0(@Nullable Throwable responseCause) {
 
     private void endResponse0(@Nullable Throwable responseCause, long responseEndTimeNanos) {
         final int flags;
-        if (responseCause == null && responseContentDeferred) {
-            flags = RequestLogProperty.FLAGS_RESPONSE_COMPLETE_WITHOUT_CONTENT;
+        if (responseCause == null && deferredFlags > 0) {
```
trustin: `>` -> `!=`

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1140,8 +1135,8 @@ private void endResponse0(@Nullable Throwable responseCause) {
...
+            flags = RequestLogProperty.FLAGS_RESPONSE_COMPLETE & ~deferredFlags;
```
trustin: 2 volatile reads - please extract into a single local variable.

core/src/main/java/com/linecorp/armeria/common/logging/RequestLog.java
```diff
@@ -166,19 +165,12 @@ default long totalDurationNanos() {
     /**
      * Returns the preview of response content of the {@link Response}.
      * Note that the content preview needs to be enabled when configuring a {@link Server} or a {@link Client}
-     * to use this functionality.
+     * by {@link ContentPreviewingService} or {@link ContentPreviewingClient} decorators respectively.
```
trustin: How about: `Note that a {@link Service} or a {@link Client} must be decorated with ... to enable the content preview.`

core/src/main/java/com/linecorp/armeria/server/logging/ContentPreviewingService.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Decorates an {@link HttpService} to preview the content of {@link Request}s and {@link Response}s.
+ *
+ * <p>Note that this decorator just sets {@link RequestLog#requestContentPreview()} and
+ * {@link RequestLog#responseContentPreview()}. You can get the previews using {@link RequestLogAccess}.
+ *
+ * <pre>{@code
+ * RequestLogAccess logAccess = ctx.log();
+ * logAccess.whenAvailable(RequestLogProperty.REQUEST_CONTENT_PREVIEW,
+ *                         RequestLogProperty.RESPONSE_CONTENT_PREVIEW).thenApply(...);
```
trustin: How about using `whenComplete()` to promote a good practice?

core/src/test/java/com/linecorp/armeria/client/HttpResponseDecoderTest.java
```diff
@@ -69,7 +70,7 @@ void confirmResponseStartAndEndInTheSameThread(SessionProtocol protocol)
         // This increases the execution duration of 'endResponse0' of the DefaultRequestLog,
         // which means that we have more chance to reproduce the bug if two threads are racing
         // for notifying RESPONSE_END to listeners.
-        builder.contentPreview(100);
+        builder.decorator(ContentPreviewingClient.builder().contentPreview(100).newDecorator());
         // In order to use a different thread to to subscribe to the response.
```
trustin: to to -> to

trustin:
What would happen if a user adds more than one `ContentPreview{Client,Service}`? Can we detect and prevent such situation?

minwoox:
> What would happen if a user adds more than one ContentPreview{Client,Service}? Can we detect and prevent such situation? e.g. check with as(Class) before decorating?
 
The decorator which calls `(request|response)ContentPreview` first is the winner in this case.
I thought there's a still chance that `RequestLogBuilder.requestContentPreview(str)` is called by another code, so just leaving it as it is okay and blaming the user for misuses. But it seems not. 🤔  Let me see what I can do.

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewingDecoratorBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.internal.ArmeriaHttpUtil;
+
+/**
+ * Builds a new content previewing decorator or its decorator function.
+ *
+ * @param <T> the type of this builder
+ */
+public abstract class ContentPreviewingDecoratorBuilder<T extends ContentPreviewingDecoratorBuilder<T>> {
```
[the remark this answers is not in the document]
minwoox: I thought it's okay because this class is public.
However, I should've moved this class under internal package and not using the type parameter. 😅

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -336,6 +322,21 @@ public RequestContext context() {
         return partiallyCompletedFuture;
     }
 
+    private void updateDeferredFlags(RequestLogProperty property) {
```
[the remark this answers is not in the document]
minwoox: `updateAvailability()` calls `removeSatisfiedFutures()` so I moved after that. 😄

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
minwoox: Oops. 😅

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -86,17 +89,14 @@
     @Nullable
     private UnmodifiableFuture<RequestLog> completedFuture;
 
-    private volatile boolean requestContentDeferred;
-    private volatile boolean responseContentDeferred;
+    private volatile int deferredFlags;
```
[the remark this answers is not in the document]
minwoox: Fixed.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -336,6 +322,21 @@ public RequestContext context() {
...
+                break;
+            }
+
+            if (deferredFlagsUpdater.compareAndSet(this, oldFlags, newFlags)) {
+                break;
+            }
+        }
+    }
+
     private void updateAvailability(RequestLogProperty property) {
```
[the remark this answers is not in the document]
minwoox: Renamed. 😉

core/src/main/java/com/linecorp/armeria/server/logging/ContentPreviewingService.java
minwoox: That's a good suggestion!

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -845,8 +844,9 @@ private void endRequest0(@Nullable Throwable requestCause) {
 
     private void endRequest0(@Nullable Throwable requestCause, long requestEndTimeNanos) {
         final int flags;
-        if (requestCause == null && requestContentDeferred) {
-            flags = RequestLogProperty.FLAGS_REQUEST_COMPLETE_WITHOUT_CONTENT;
+        final int deferredFlags = this.deferredFlags;
+        if (requestCause == null && deferredFlags != 0) {
```
trustin: `deferredFlags` doesn't need to be read when `requestCause` is not `null`.

core/src/main/java/com/linecorp/armeria/internal/logging/ContentPreviewerConfigurator.java
```diff
@@ -0,0 +1,174 @@
...
+        final RequestHeaders headers = req.headers();
+        final ContentPreviewer requestContentPreviewer = requestContentPreviewerFactory.get(ctx, headers);
+        if (requestContentPreviewer.isDisabled()) {
+            return req;
+        }
+
+        final RequestLogBuilder logBuilder = ctx.logBuilder();
+        logBuilder.deferRequestContentPreview();
+        requestContentPreviewer.onHeaders(headers);
+        req.completionFuture().handle((unused, unused1) -> {
```
trustin: Please merge master and use `whenComplete()`.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
minwoox: 🙆‍♂

core/src/main/java/com/linecorp/armeria/internal/logging/ContentPreviewerConfigurator.java
minwoox: Fixed. 😉

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -780,6 +785,11 @@ public String requestContentPreview() {
     @Override
     public void requestContentPreview(@Nullable String requestContentPreview) {
         if (isAvailable(RequestLogProperty.REQUEST_CONTENT_PREVIEW)) {
+            if (requestContentPreview != null) {
+                logger.warn("You try to set the request content preview twice: {}. " +
+                            " Do you configure content previewing decorator more than once?",
+                            requestContentPreview);
```
trustin: - Probably better logging only once?
- `{}.` -> `{}` ? (The resulting log message would be: `SomePreview.` looking like `.` is part of the preview)
- You try -> You tried
- Do you configure .. -> Did you apply .. ?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -1066,6 +1075,11 @@ public String responseContentPreview() {
     @Override
     public void responseContentPreview(@Nullable String responseContentPreview) {
         if (isAvailable(RequestLogProperty.RESPONSE_CONTENT_PREVIEW)) {
+            if (responseContentPreview != null) {
+                logger.warn("You try to set the response content preview twice: {}. " +
+                            " Do you configure content previewing decorator more than once?",
+                            responseContentPreview);
```

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
minwoox: Fixed. Thanks!

core/src/main/java/com/linecorp/armeria/client/logging/ContentPreviewingClientBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+        return new ContentPreviewingClient(delegate, requestContentPreviewerFactory(),
+                                           responseContentPreviewerFactory());
+    }
+
+    /**
+     * Returns a newly-created decorator that decorates an {@link HttpClient} with a new
+     * {@link ContentPreviewingClient} based on the properties of this builder.
+     */
+    public Function<? super HttpClient, ContentPreviewingClient> newDecorator() {
+        return this::build;
```
anuraaga: Do you think we should check state here too? It might give a user earlier feedback

core/src/test/java/com/linecorp/armeria/client/logging/ContentPreviewingClientTest.java
```diff
@@ -0,0 +1,139 @@
...
+            context = captor.get();
+        }
+
+        final RequestLog requestLog = context.log().whenComplete().join();
+        assertThat(requestLog.requestContentPreview()).isEqualTo("Armeria");
+        assertThat(requestLog.responseContentPreview()).isEqualTo("Hello Armeria!");
+    }
+
+    @Test
+    void contentPreviewIsDecodedInPreviewer() {
```
anuraaga: What is this testing? Or mostly for documentation of ordering semantics?

minwoox: Just wanted to show that the preview can be different depends on the orders it attached.

minwoox:
> I think we need to add in at least javadoc, maybe rstdoc, an explanation of the ordering.

Yeah, let me make a separate rstdoc for this. I was actually getting confused about the order of decorator while making the test cases. 😅 So I really thought that we needed a well documented separate page for that.

minwoox:
https://github.com/line/armeria/issues/1117
Self assigned. 😄

core/src/main/java/com/linecorp/armeria/client/logging/ContentPreviewingClientBuilder.java
minwoox: Which state do you mean?

anuraaga: I mean the `checkState` in `build` - I think usually that will be called much later than setting the decorator so could be confusing, but here it would be right away.

core/src/test/java/com/linecorp/armeria/client/logging/ContentPreviewingClientTest.java
anuraaga: Probably good to add a comment about that

core/src/main/java/com/linecorp/armeria/client/logging/ContentPreviewingClientBuilder.java
minwoox: Ah, that's a good point. Let me add that. 😄

core/src/test/java/com/linecorp/armeria/client/logging/ContentPreviewingClientTest.java
minwoox: Added. 😉

core/src/main/java/com/linecorp/armeria/server/ServiceConfig.java
```diff
@@ -60,32 +56,24 @@
      */
     ServiceConfig(Route route, HttpService service,
                   long requestTimeoutMillis, long maxRequestLength, boolean verboseResponses,
-                  ContentPreviewerFactory requestContentPreviewerFactory,
-                  ContentPreviewerFactory responseContentPreviewerFactory, AccessLogWriter accessLogWriter,
+                  AccessLogWriter accessLogWriter,
                   boolean shutdownAccessLogWriterOnStop) {
```
ikhoon: nit: Could join two lines?

core/src/main/java/com/linecorp/armeria/server/logging/ContentPreviewingServiceBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.function.Function;
+
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+import com.linecorp.armeria.common.logging.ContentPreviewingDecoratorBuilder;
+import com.linecorp.armeria.server.HttpService;
+
+/**
+ * Builds a new {@link ContentPreviewingService} or its decorator function.
+ */
+public final class ContentPreviewingServiceBuilder extends ContentPreviewingDecoratorBuilder {
```
ikhoon: Don't we need a package-private default constructor?

core/src/main/java/com/linecorp/armeria/client/logging/ContentPreviewingClientBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.function.Function;
+
+import com.linecorp.armeria.client.HttpClient;
+import com.linecorp.armeria.common.logging.ContentPreviewerFactory;
+import com.linecorp.armeria.common.logging.ContentPreviewingDecoratorBuilder;
+
+/**
+ * Builds a new {@link ContentPreviewingClient} or its decorator function.
+ */
+public final class ContentPreviewingClientBuilder extends ContentPreviewingDecoratorBuilder {
```
ikhoon: Don't we need a package-private default constructor?

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewingDecoratorBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+     *     <li>when it matches {@code text/*} or {@code application/x-www-form-urlencoded}</li>
+     *     <li>when its charset has been specified</li>
+     *     <li>when its subtype is {@code "xml"} or {@code "json"}</li>
+     *     <li>when its subtype ends with {@code "+xml"} or {@code "+json"}</li>
+     * </ul>
+     * @param length the maximum length of the preview
+     * @param defaultCharset the default charset used when a charset is not specified in the
+     *                       {@code "content-type"} header
+     */
+    public ContentPreviewingDecoratorBuilder contentPreview(int length, Charset defaultCharset) {
```
ikhoon: nit: `requireNonNull(defaulCharset, "defaultCharset");`?

core/src/main/java/com/linecorp/armeria/server/logging/ContentPreviewingServiceBuilder.java
minwoox: Oops we need that. Thanks!

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewingDecoratorBuilder.java
minwoox: Added!

trustin:
:tada::tada::tada:

minwoox:
Thanks, reviewers!

PR #2423, opened 2020-01-29
Don't expose SAML impl in public API.

anuraaga:
I noticed there is one SAML implementation API exposed in our public API. Given SAML's diligent separation of API and implementation, I guess it's better for us to also make sure to only expose the API.

### DIFF
--- a/examples/saml-service-provider/src/main/java/example/armeria/server/saml/sp/Main.java
+++ b/examples/saml-service-provider/src/main/java/example/armeria/server/saml/sp/Main.java
@@ -50,7 +50,7 @@ public class Main {
 
         // Specify information about your keystore.
         // The keystore contains two key pairs, which are identified as 'signing' and 'encryption'.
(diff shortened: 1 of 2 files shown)


examples/saml-service-provider/src/main/java/example/armeria/server/saml/sp/Main.java
```diff
@@ -50,7 +50,7 @@ private static SamlServiceProvider samlServiceProvider() throws IOException, Gen
 
         // Specify information about your keystore.
         // The keystore contains two key pairs, which are identified as 'signing' and 'encryption'.
-        final KeyStoreCredentialResolver credentialResolver =
+        final CredentialResolver credentialResolver =
```
minwoox: missing the import statement for this? 😄

trustin:
Thanks, @anuraaga !

PR #2424, opened 2020-01-29
Switch to JDK 13 for releasing and CI build.

anuraaga:
Also adds JRE exact version to paths so version can be updated without manually clearing CI cache

I found that JDK12 has this bug when generating javadoc

https://bugs.openjdk.java.net/browse/JDK-8222091

I sort of wonder how we've been releasing so far (do we release on Windows?), but I guess it's a good time to switch to JDK13 for release.

### DIFF
--- a/.appveyor.build.sh
+++ b/.appveyor.build.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 set -eo pipefail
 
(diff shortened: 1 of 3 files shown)


trustin:
> I sort of wonder how we've been releasing so far (do we release on Windows?)

I've been releasing on Linux. No idea it worked so far. :see_no_evil:

PR #2425, opened 2020-01-30
Use JUnit Jupiter's global timeout settings

trustin:
Motivation:

JUnit Jupiter provides various properties that controls the timeout
behavior globally, and we could use them instead of manually specifying
`Timeout` annotations with magic numbers.

Modifications:

- Disable timeout on debug by default.
- Set 60-second timeout for all testables and life cycle methods.
- Remove `TestUtil.withTimeout()`
- Update dependencies:
  - JUnit Platform 1.5.2 -> 1.6.0
  - JUnit 5.5.2 -> 5.6.0, 4.12 -> 4.13

Result:

No need to add `Timeout` annotation


trustin:
Special thanks to @KarboniteKream for inspiration.

PR #2427, opened 2020-01-30
Rename `*Future()` to `when*()` and `onDemand()` to `whenConsum…

trustin:
Motivation:

The recent revamp of `RequestLog` API (#2342) showed us a
`Future`-returning method can be named as `when*()` for better
readability:

    ctx.log().whenComplete().thenAccept(log -> ...);

We could apply the same technique to other places.

Modifications:

- Deprecate `completionFuture()` in favor of `whenComplete()`
- Deprecate `initialEndpointsFuture()` in favor of `whenReady()`
  - Also deprecated `awaitInitialEndpoints()`.
- Deprecate `onDemand()` in favor of `whenConsumed()`

Result:

- Improved readability


trustin:
To-do: Make `AwaitDemandFuture` unmodifiable.

Issue #1410, opened 2018-10-25
'cannot start a new stream with a DATA frame' error under load


trustin:
.. probably due to an unexpected disconnection.

```
17:06:32.883 [armeria-common-worker-epoll-2-11] WARN  c.l.armeria.server.HttpServerHandler - [id: 0x39d2ed0f, L:/127.0.0.1:44525 - R:/127.0.0.1:40232][h2c] Unexpected exception:
java.lang.IllegalStateException: cannot start a new stream 177893 with a DATA frame
        at com.linecorp.armeria.internal.Http2ObjectEncoder.doWriteData(Http2ObjectEncoder.java:92)
        at com.linecorp.armeria.internal.HttpObjectEncoder.writeData(HttpObjectEncoder.java:74)
        at com.linecorp.armeria.server.HttpResponseSubscriber.write0(HttpResponseSubscriber.java:295)
        at com.linecorp.armeria.server.HttpResponseSubscriber.lambda$write$1(HttpResponseSubscriber.java:286)
        at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:404)
        at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:326)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:897)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:844)
```

ikhoon:
Another thread dump, It is caused by GoAway frame
```java
"armeria-common-worker-nio-2-1@4812" daemon prio=5 tid=0x17 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
	  at com.linecorp.armeria.internal.Http2ObjectEncoder.doWriteData(Http2ObjectEncoder.java:93)
	  at com.linecorp.armeria.internal.HttpObjectEncoder.writeData(HttpObjectEncoder.java:85)
	  at com.linecorp.armeria.server.HttpResponseSubscriber.write(HttpResponseSubscriber.java:298)
	  at com.linecorp.armeria.server.HttpResponseSubscriber.onNext(HttpResponseSubscriber.java:227)
[...]

PR #2429, opened 2020-01-30
Fix 'cannot start a new stream with a DATA frame' error under load

ikhoon:
Motivation:
When a server under load receives `GOAWAY` frame from client,
the server try to close the given connection with the flowing step:
1) The server receives the `GOAWAY` frame.
2) Close all active streams before closing connection.
3) Flush `outboundBuffer` buffer.
4) The callback of `future.addListener` is called after successfully writing `nioBuffers` to channel.
https://github.com/line/armeria/blob/84403ba1424faf85ca883bb6a299705a3198c4bd/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java#L307
5) Request the next item in the `future.addListener`.
https://github.com/line/armeria/blob/84403ba1424faf85ca883bb6a299705a3198c4bd/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java#L333
6) Try to write the next item to the stream, but it was already closed on 2)

If try to write `HttpData` to the closed stream, it will throw 'cannot start a new stream with a DATA frame'.

Modification:
* Add `isWritable(id, streamId)` to `HttpObjectEncoder`
* Cancel subscription and propagate error

Result:
No more 'cannot start a new stream with a DATA frame' error when receiving GOAWAY with non empty `ChannelOutboundBuffer`.
Fixes: #1410


core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -321,6 +323,7 @@ private void write(HttpObject o, boolean endOfStream) {
                                 f.cause() instanceof ClosedChannelException &&
                                 responseEncoder instanceof Http1ObjectEncoder;
                 }
+                final boolean isWritable = responseEncoder.isWritable(req.id(), req.streamId());
```
trustin: Maybe better grouping with `final boolean isSuccess;`?

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -333,17 +336,28 @@ private void write(HttpObject o, boolean endOfStream) {
...
-                    subscription.request(1);
-                    return;
+                    if (isWritable) {
+                        subscription.request(1);
+                        return;
+                    }
+                }
+
+                ChannelFuture failedFuture = f;
+                if (isSuccess && !isWritable) {
```
trustin: Could be merged into the `if (isSuccess)` above, i.e.

```java
final ChannelFuture failedFuture;
if (isSuccess) {
    ...
    if (isWritable) {
        subscription.request(1);
        return;
    }

    if (reqCtx.sessionProtocol().isMultiplex()) {
        failedFuture = f.channel().newFailedFuture(ClosedPublisherException.get());
    } else {
        failedFuture = f.channel().newFailedFuture(ClosedSessionException.get());
    }
} else {
    failedFuture = f;
}
```

trustin:
Please inspect the test failures.

ikhoon:
> Please inspect the test failures.

Sure!

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -325,25 +328,43 @@ private void write(HttpObject o, boolean endOfStream) {
...
                     maybeLogFirstResponseBytesTransferred();
 
+                    if (state == State.DONE && !endOfStream) {
+                        // if state is DONE, the response should be ended with endOfStream.
+                        return;
+                    }
+
                     if (endOfStream && tryComplete()) {
                         logBuilder().endResponse();
+                        subscription.cancel();
```
ikhoon: Oops, this is miss pushed. 😱  Let me revert this.

thrift/src/test/java/com/linecorp/armeria/it/metric/DropwizardMetricsIntegrationTest.java
```diff
@@ -72,16 +67,13 @@ protected void configure(ServerBuilder sb) throws Exception {
...
 
-    @AfterClass
-    public static void closeClientFactory() {
+    @AfterAll
+    static void closeClientFactory() {
         clientFactory.close();
     }
 
-    @Rule
-    public final TestRule globalTimeout = new DisableOnDebug(new Timeout(30, TimeUnit.SECONDS));
```
mauhiz: This is just gone?

ikhoon: Now, we are using JUnit5 global timeout configar. https://github.com/line/armeria/pull/2425/files#diff-9c2d5d92bbbb9e98c96cf3f21ecedf97

mauhiz: oh, that is a nice feature!

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -325,25 +328,41 @@ private void write(HttpObject o, boolean endOfStream) {
...
+                    }
+
                     if (endOfStream && tryComplete()) {
                         logBuilder().endResponse();
                         reqCtx.log().whenComplete().thenAccept(reqCtx.accessLogWriter()::log);
                     }
 
-                    subscription.request(1);
-                    return;
+                    if (isWritable) {
```
minwoox: I realized that we might need this in `HttpRequestSubscriber` when we send more that one `HttpData`.

ikhoon: Good point! Let me look into it. 😀

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -289,6 +289,7 @@ public void onComplete() {
     }
 
     private void write(HttpObject o, boolean endOfStream) {
+        // if the first bytes transferred, the stream must be open
```
trustin: - if -> If
- transferred -> were transferred
- open -> open.

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -297,6 +296,16 @@ private void write(HttpObject o, boolean endOfStream, boolean flush) {
     }
 
     private void write0(HttpObject o, boolean endOfStream, boolean flush) {
+        // if the first bytes transferred, the stream must be open
```
trustin: Could you explain why?

trustin: The reason why I'm asking this is because the comment itself simply explains what the code does. It does not add much to the code without the comment in my opinion.

If we think there's something more to explain about the code, it has to be more detailed than this.

ikhoon: I agree with your optinion. I will add the detail explanation about 
* Why I use `loggedRequestFirstBytesTransferred` to check stream condition
* When this method could be called even if the stream is closed.

PR #2430, opened 2020-01-30
Make some `EndpointGroup`s implement `ListenableAsyncCloseable`

### LINKED ISSUE #1084: EndpointGroup design issues

trustin:
Motivation:

`EndpointGroup` does not extend `ListenableAsyncCloseable` but only
`AsyncCloseable`, because otherwise we would have to expose the methods
that do not blend well into `Endpoint`. However, it does not stop other
`EndpointGroup` implementations from implementing
`ListenableAsyncCloseable`.

Modifications:

- The following classes now implement `ListenableAsyncCloseable`:
  - `DynamicEndpointGroup`
  - `CompositeEndpointGroup`
  - `OrElseEndpointGroup`
- `HealthCheckedEndpointGroup` now closes itself automatically when its
  `ClientFactory` is closed.

Result:

- Easier to inspect the life cycle state of some `EndpointGroup`s.
- `HealthCheckedEndpointGroup` now closes itself automatically when its
  `ClientFactory` is closed.
  - Closes #1084

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/CompositeEndpointGroup.java
@@ -30,11 +30,14 @@ import com.linecorp.armeria.client.Endpoint;
 import com.linecorp.armeria.common.util.AbstractListenable;
 import com.linecorp.armeria.common.util.AsyncCloseable;
 import com.linecorp.armeria.common.util.AsyncCloseableSupport;
+import com.linecorp.armeria.common.util.ListenableAsyncCloseable;
 
 /**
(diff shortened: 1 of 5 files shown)


trustin:
Other `EndpointGroup`s than `HealthCheckedEndpointGroup` cannot be closed automatically when `ClientFactory` closes because they do not have any dependency on it.

PR #2436, opened 2020-01-31
Make requestStartTime set before a connection attempt

### LINKED ISSUE #2272: Request time does not take into account connection creation

trustin:
Motivation:

When a request involves a new connection attempt, `requestDuration` and
`totalDuration` of the request must include the time taken for
establishing a connection.

Modifications:

- Split `RequestLogBuilder.startRequest()` into `startRequest()` and
  `session()`, so that `requestStartTime` can be set as soon as
  possible.
  - `DefaultClientRequestContext` now calls `startRequest()` at its
    construction time.
- `ClientConnectionTimings` is now a first-class citizen of `RequestLog`.
  - Moved `ClientConnectionTimings` to `common.logging`.
  - Added `RequestOnlyLog.connectionTimings()`.
- Added `ClientRequestContextBuilder.connectionTimings()`.
- Miscellaneous:
  - Removed some constructors in `Default{Client,Service}RequestContext`
    for simplicity.

Result:

- Fixes #2272
- Client-side `RequestLog.requestStartTime()` is much more accurate now.
- `requestDuration` and `totalDuration` now include the time taken for
  creating a new connection.
- Easier to retrieve `ClientConnectionTimings`


trustin:
Thanks for reviewing.

Issue #2421, opened 2020-01-29
Provide `String` constants corresponding to `MediaType` constants


trustin:
From @tobias-

> Are there any plans for adding the more common types of media types as string constants for consumption with e.g. `@Produces` ? I know there's a pretty comprehensive list in `com.linecorp.armeria.common.MediaType`, but those are unusable by annotations.

Sounds like a good suggestion, as long as we can add some test cases that ensure the strings and `MediaType`s are in sync.

JunoJunho:
I want to handle this issue.

Is it okay just make pull request for this issue?

trustin:
Yes, please. Thanks a lot in advance, @JunoJunho!

trustin:
Perhaps we could create a new constant class `MediaTypeNames` and put strings there.

JunoJunho:
This can be closed?

trustin:
Yes. :smile:

PR #2438, opened 2020-01-31
Add MediaTypeNames

JunoJunho:
Motivation:

As described in #2421, armeria currently does not provide String type
constants for MediaType

Modifications:

Introduce new class for String Constants in MediaType, MediaTypeNames

Result:

Implement MediaTypeNames class and its test (MediaTypeNamesTest)


JunoJunho:
Let me check using full build

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test sync with MediaType and MediaTypeNames
+ */
+class MediaTypeNamesTest {
+
+    @Test
+    void matchMediaTypeToMediaTypeNames() throws Exception {
```
trustin: I think you can use reflection for completeness.

core/src/main/java/com/linecorp/armeria/common/MediaTypeNames.java
```diff
@@ -0,0 +1,155 @@
...
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  License for the specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package com.linecorp.armeria.common;
+
+public final class MediaTypeNames {
+
+    public static final String ANY_TYPE = "*/*";
```
trustin: Missing Javadoc. Could use the string in verbatim, e.g.

```
/**
 * {@value #ANY_TYPE}.
 */
// or..
/**
 * {@code "*/*"}.
 */
```

JunoJunho: For all constants? What your intention is showing the value in javadoc?

I think first suggestion is better because it seems good to later when editing this code.

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
```diff
@@ -0,0 +1,86 @@
...
+            final String mediaTypeName = mediaTypeNamesConstantsMap.get(mediaTypeEntry.getKey());
+            assertNotNull("MediaTypeName should be defined in MediaType constants", mediaTypeName);
+            assertEquals(mediaTypeName, mediaTypeEntry.getValue().toString());
+        }
+    }
+
+    // reflection
+    @SuppressWarnings("Guava")
+    private static <T> FluentIterable<Field> getConstantFields(Class<T> clazz) {
+        return FluentIterable.from(asList(clazz.getDeclaredFields())).filter(input -> {
```
minwoox: `FluentIterable` is a discouraged precursor by its Javadoc. How about using `Stream.of(clazz.getDeclaredFields()).filter...`?

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        final Map<String, MediaType> mediaTypeConstantsMap =
+                getConstantFieldMap(mediaTypeFields, MediaType.class);
+        final Map<String, String> mediaTypeNamesConstantsMap =
+                getConstantFieldMap(mediaTypeNamesFields, String.class);
+
+        for (Entry<String, MediaType> mediaTypeEntry : mediaTypeConstantsMap.entrySet()) {
+            final String mediaTypeName = mediaTypeNamesConstantsMap.get(mediaTypeEntry.getKey());
+            assertNotNull("MediaTypeName should be defined in MediaType constants", mediaTypeName);
+            assertEquals(mediaTypeName, mediaTypeEntry.getValue().toString());
```
minwoox: Could you use AssertJ?
`assertThat(mediaTypeName).isNotNull();`
`assertThat(mediaTypeName).isEqualTo(mediaTypeEntry.getValue().toString());`

core/src/main/java/com/linecorp/armeria/common/MediaTypeNames.java
```diff
@@ -0,0 +1,472 @@
...
+     */
+    public static final String JSON = "application/json";
+    /**
+     * {@value #JSON_PATCH}.
+     */
+    public static final String JSON_PATCH = "application/json-patch+json";
+    /**
+     * {@value #JSON_SEQ}.
+     */
+    public static final String JSON_SEQ = "application/json-seq";
```
ikhoon: Could you migrate `@Produces("application/json-seq")` to `@Produces(JSON_SEQ)` and other similar things in our code base?

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
minwoox: Found out that this code is from `MediaTypeTest`. How about making the method as package-private and reusing here?

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
```diff
@@ -0,0 +1,86 @@
...
+            return isPublic(modifiers) &&
+                   isStatic(modifiers) &&
+                   isFinal(modifiers) &&
+                   String.class.equals(input.getType());
+        });
+    }
+
+    // reflection
+    @SuppressWarnings("Guava")
+    private static <T> Map<String, T> getConstantFieldMap(FluentIterable<Field> iterable, Class<T> clazz) {
```
ikhoon: The `clazz` parameter seems unused.

How about using stream API & ImmutableMap instead of mutable `HashMap`?
```java
private static <T> Map<String, T> getConstantFieldMap(Iterable<Field> iterable) {
    return StreamSupport
            .stream(iterable.spliterator(), false)
            .collect(toImmutableMap(Field::getName, e -> {
                try {
                    final T cast = (T) e.get(null);
                    return cast;
                } catch (Exception ex) {
                    throw new RuntimeException(ex);
                }
            }));
}
```

core/src/main/java/com/linecorp/armeria/common/MediaTypeNames.java
JunoJunho: I applied them all but test code (I am not sure applying test code since some are applied some are not. It seems weird to me).

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
JunoJunho: Let me apply it.

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
JunoJunho: Yes, it is from `guava` test and it is no problem since they are used in verified way (in Guava).

I will make the method as package-private and re-using.

JunoJunho: But little-bit edit is applied (Class parameter) I think it is okay to me. If any issue, please tell me.

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
JunoJunho: Oh! I did not know that I can use without parameter. Let me apply this code snippet! Thanks!

core/src/main/java/com/linecorp/armeria/common/MediaTypeNames.java
trustin: When working with IDE, it will show the value of the constant in quick documentation, auto-completion hints, etc.

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
```diff
@@ -0,0 +1,67 @@
...
+        for (Entry<String, MediaType> mediaTypeEntry : mediaTypeConstantsMap.entrySet()) {
+            final String mediaTypeName = mediaTypeNamesConstantsMap.get(mediaTypeEntry.getKey());
+            assertThat(mediaTypeName).isNotNull();
+            assertThat(mediaTypeName).isEqualTo(mediaTypeEntry.getValue().toString());
+        }
+    }
+
+    private static <T> Map<String, T> getConstantFieldMap(Iterable<Field> iterable) {
+        return StreamSupport
+                .stream(iterable.spliterator(), false)
```
trustin: `Iterables.stream(iterable)` ?

JunoJunho: I could not find this method.

trustin: Ugh, sorry. I must be mistaken. :sweat_smile:

core/src/main/java/com/linecorp/armeria/server/annotation/ConsumesJson.java
```diff
@@ -20,11 +20,13 @@
...
+import com.linecorp.armeria.common.MediaTypeNames;
+
 /**
- * An alias for {@code @Consumes("application/json")}.
+ * An alias for {@code @Consumes(MediaTypeNames.JSON)}.
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.TYPE, ElementType.METHOD })
-@Consumes("application/json")
+@Consumes(MediaTypeNames.JSON)
```
ikhoon: Special thanks to this ❤️

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
anuraaga: In guava it's `Streams.stream` :)

core/src/test/java/com/linecorp/armeria/common/MediaTypeTest.java
```diff
@@ -116,20 +116,21 @@ public void testConstants_areUnique() {
 
     // reflection
     @SuppressWarnings("Guava")
-    private static FluentIterable<Field> getConstantFields() {
-        return FluentIterable.from(asList(MediaType.class.getDeclaredFields())).filter(input -> {
+    static <T, R> FluentIterable<Field> getConstantFields(Class<T> clazz, Class<R>... filterClazz) {
```
anuraaga: Though if we can I think it's good to migrate these to return `Stream`, we don't use `FluentIterable` anywhere else in our code

minwoox: > it's good to migrate these to return Stream

+1 😄

trustin: Mayne better leaving at it is, given the entire test case has been forked from Guava?

ikhoon:
@JunoJunho Could you merge the master branch? 😉

core/src/test/java/com/linecorp/armeria/common/MediaTypeTest.java
anuraaga: Ah yeah it's fine to leave it then, can just clean up the `Streams.stream` part.

JunoJunho:
@ikhoon merged :)

core/src/test/java/com/linecorp/armeria/common/MediaTypeNamesTest.java
trustin: @JunoJunho Could you fix this?

ikhoon: > In guava it's Streams.stream :)

Good to know 👍

JunoJunho: Let me handle this.

JunoJunho:
Let me check in local

trustin:
Thanks a lot, @JunoJunho !

minwoox:
:+1::+1::+1:

PR #2439, opened 2020-02-01
Public API clean up

trustin:
Related: #2360

- Removed `ExceptionFilter` which is not used anywhere.
- Added `StreamDecoderFactory.gzip()` and `deflate()` and hid
  `{Gzip,Deflate}StreamDecoderFactory`.
- Removed `AbstractConcurrencyLimitingClient.Deferred`.
- Deprecated `DefaultRequestLog` to hide it from public API later.
- Removed everything under 'structured' package.
- Cleaned up and move Jackson serializers for Thrift types to `common.thrift`.
- `DEFAULT` -> `ofDefault()`
- Minor Javadoc fixes here and there


thrift/src/main/java/com/linecorp/armeria/common/thrift/ThriftJacksonSerializers.java
```diff
@@ -0,0 +1,233 @@
...
+        gen.writeRawValue(serializeTBaseLike(protocol -> {
+            try {
+                value.write(protocol);
+            } catch (TException ex) {
+                throw new IllegalArgumentException(ex);
+            }
+        }));
+    }
+
+    static String serializeTBaseLike(Consumer<TProtocol> writer) {
```
ikhoon: nit: Could be `private`?

core/src/main/java/com/linecorp/armeria/client/encoding/GzipStreamDecoderFactory.java
```diff
@@ -22,7 +22,11 @@
 /**
  * A {@link StreamDecoderFactory} which supports the 'gzip' encoding.
  */
-public class GzipStreamDecoderFactory implements StreamDecoderFactory {
+final class GzipStreamDecoderFactory implements StreamDecoderFactory {
+
+    static final GzipStreamDecoderFactory INSTANCE = new GzipStreamDecoderFactory();
```
anuraaga: Maybe use enum singleton pattern for simple singletons like these

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/MeterIdPrefixFunctionFactory.java
```diff
@@ -27,10 +27,20 @@
...
-    MeterIdPrefixFunctionFactory DEFAULT = (type, serviceName) ->
-            MeterIdPrefixFunction.ofDefault("armeria." + requireNonNull(type, "type"))
-                                 .withTags("service", requireNonNull(serviceName, "serviceName"));
+    @Deprecated
+    MeterIdPrefixFunctionFactory DEFAULT = ofDefault();
+
+    /**
+     * Returns the default {@link MeterIdPrefixFunctionFactory} instance.
+     */
+    static MeterIdPrefixFunctionFactory ofDefault() {
```
anuraaga: This stopped being a singleton - probably not a big issue but just to make sure

zookeeper/src/main/java/com/linecorp/armeria/server/zookeeper/ZooKeeperUpdatingListener.java
```diff
@@ -25,15 +25,20 @@
...
 import com.linecorp.armeria.internal.zookeeper.ZooKeeperDefaults;
 import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.ServerListener;
 import com.linecorp.armeria.server.ServerListenerAdapter;
 import com.linecorp.armeria.server.ServerPort;
 
 /**
- * A ZooKeeper Server Listener. When you add this listener, server will be automatically registered
- * into the ZooKeeper.
+ * A {@link ServerListener} which (de)registers the current {@link Server} to
```
anuraaga: The following sentence explains deregistration and I think this reads better as just `which registers the current`..., it's the main point of the class.

core/src/main/java/com/linecorp/armeria/client/encoding/GzipStreamDecoderFactory.java
trustin: Good idea. Merged the two classes into one enum class.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/MeterIdPrefixFunctionFactory.java
trustin: Extracted into a separate singleton.

thrift/src/main/java/com/linecorp/armeria/common/thrift/ThriftJacksonSerializers.java
trustin: Ah, yeah. Thanks for pointing out.

ikhoon:
Still LGTM :-)

trustin:
Thanks for the reviews.

PR #2442, opened 2020-02-03
Make most public classes final

trustin:
.. so that we have more freedom to add new public methods to them
without breaking backward compatibility. Note that some classes are not
final yet because 1) it has a deprecated subtype or 2) it will be
package-private later.  They must be finalized once we delete all
deprecated classes before 1.0.0.

Related: #2360


trustin:
Need to update `HttpStreamReaderTest` not to use spies.

trustin:
Gave up and removed `final` from `ArmeriaMessage{Framer,Deframer}`. They are going to be annotated with `@UnstableApi` anyway. :man_shrugging:

trustin:
Let me merge this to move forward. Please feel free to let me know if something needs to be reverted.

PR #2444, opened 2020-02-03
Refactor ContentPreviewer

minwoox:
Motivation:
Related #2422

`ContentPreviewer` can only be created after the `RequestHeaders` or `ResponseHeaders` is available.
Therefore, I think we can remove `ContentPreviewer.onHeaders(...)` and just pass the headers when creating `ContentPreviewer` if it needs.

Modifications:
- `ContentPreviewerFactory` now has separate methods to return request and response `ContentPreviewer`s. 
- Add `ContentPreviewerFactoryBuilder`.
- Remove factory methods in `ContentPreviewer` and `ContentPreviewerFactory`.
  - Only `text(..)` is left in `ContentPreviewerFactory`
- Remove `ContentPreviewingClientBuilder` and `ContentPreviewingServiceBuilder`.
- (Deprecated)
  - The constructors of `MediaTypeSet` is deprecated.
    - Use `MediaTypeSet.of(...)` instead.

Result:
- Cleaner and more intuitive APIs hopefully.


core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactory.java
```diff
@@ -16,216 +16,262 @@
...
-     * Creates a new instance of {@link ContentPreviewerFactory} which creates a {@link ContentPreviewer}
-     * through {@code supplier} for the request/response with the {@code contentType}.
+     * The default HTTP content-type charset.
+     * See https://tools.ietf.org/html/rfc2616#section-3.7.1
      */
-    static ContentPreviewerFactory of(MediaType contentType, Supplier<? extends ContentPreviewer> supplier) {
-        return of(ImmutableMap.of(contentType, supplier));
+    static Charset defaultCharset() {
+        return ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET;
     }
```
trustin: I guess this should be moved somewhere else?

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactory.java
```diff
@@ -16,216 +16,262 @@
...
+        return (ctx, headers) -> ContentPreviewer.ofBinary(maxLength, producer, headers);
     }
 
     /**
-     * Returns a {@link ContentPreviewer}, given {@link RequestContext} and {@link HttpHeaders}.
+     * Returns a newly-created {@link ContentPreviewer} with the given {@link RequestContext} and
+     * {@link HttpHeaders}. Note that the returned {@link ContentPreviewer} can be
+     * {@link ContentPreviewer#disabled()}.
      */
     ContentPreviewer get(RequestContext ctx, HttpHeaders headers);
```
trustin: How about splitting this method into two:

```java
ContentPreviewer requestContentPreviewer(ctx, RequestHeaders headers);
ContentPreviewer responseContentPreviewer(ctx, RequestHeaders reqHeaders, ResponseHeaders resHeaders);
```

I guess the behavior of response previewer might be affected by a certain request header, e.g. `x-enable-content-preview: 1`.

minwoox: That's a good idea. I need more refactoring. 😉

core/src/main/java/com/linecorp/armeria/common/logging/AbstractContentPreviewer.java
```diff
@@ -0,0 +1,92 @@
...
+    private static ByteBuf duplicateData(HttpData httpData, int length) {
+        if (httpData instanceof ByteBufHolder) {
+            final ByteBuf content = ((ByteBufHolder) httpData).content();
+            if (content.readableBytes() == length) {
+                // No need to slice.
+                return content.retainedDuplicate();
+            }
+            return content.retainedSlice(content.readerIndex(), length);
+        } else {
+            return Unpooled.wrappedBuffer(httpData.array());
```
ikhoon: I think if length is less than `httpData.array()` size, the `aggregatedLength` will hold wrong value. 
For example:
```
// before
bufferListLength = 95
maxLength = 100
aggregatedLength = 95

// after
data.length = 10
length = 5
newAggregatedLength = 100
bufferListLength = 105
```

minwoox: Thanks! Just copied it without much thinking 😅.

minwoox:
This is WIP. Let me add more tests and Javadoc.

core/src/main/java/com/linecorp/armeria/client/logging/ContentPreviewingClient.java
```diff
@@ -49,48 +54,72 @@
...
-    private final ContentPreviewerConfigurator configurator;
+    private final ContentPreviewerFactory contentPreviewerFactory;
 
     /**
      * Creates a new instance that decorates the specified {@link HttpClient}.
      */
-    ContentPreviewingClient(HttpClient delegate,
-                            @Nullable ContentPreviewerFactory requestContentPreviewerFactory,
-                            @Nullable ContentPreviewerFactory responseContentPreviewerFactory) {
+    ContentPreviewingClient(HttpClient delegate, ContentPreviewerFactory contentPreviewerFactory) {
```
trustin: Could be `private`?

core/src/main/java/com/linecorp/armeria/common/MediaTypeSet.java
```diff
@@ -60,16 +60,36 @@
 
     private final MediaType[] mediaTypes;
 
+    /**
+     * Returns the {@link MediaTypeSet} which is composed of the specified {@link MediaType}s.
+     */
+    public static MediaTypeSet of(MediaType... mediaTypes) {
+        return of(ImmutableList.copyOf(requireNonNull(mediaTypes, "mediaTypes")));
```
trustin: Better calling the constructor directly after `copyOf()`, given `of()` calls `copyOf()` again?

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
```diff
@@ -0,0 +1,216 @@
...
+import io.netty.buffer.ByteBufUtil;
+
+/**
+ * A builder which builds a {@link ContentPreviewerFactory}.
+ */
+public final class ContentPreviewerFactoryBuilder {
+
+    //TODO(minwoox): Add setters for the seprate request and response previewer.
+
+    private final Builder<PreviewSpec> previewSpecsBuilder = ImmutableList.builder();
```
trustin: nit: `ImmutableList.Builder`

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
```diff
@@ -0,0 +1,216 @@
...
+        return (ctx, headers) -> {
+            final MediaType contentType = headers.contentType();
+            if (contentType == null) {
+                return false;
+            }
+            return mediaTypeSet.match(contentType) != null;
+        };
+    }
+
+    static BiFunction<? super HttpHeaders, ? super ByteBuf, String> hexDumpProducer() {
```
trustin: `private`?

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
```diff
@@ -0,0 +1,216 @@
...
+        this.defaultCharset = requireNonNull(defaultCharset, "defaultCharset");
+        return this;
+    }
+
+    /**
+     * Sets the specified {@link MediaTypeSet} to produce the text preview when the content type of the
+     * {@link RequestHeaders} or {@link ResponseHeaders} {@linkplain MediaTypeSet#match(MediaType) matches}
+     * the {@link MediaTypeSet}.
+     */
+    public ContentPreviewerFactoryBuilder text(MediaTypeSet mediaTypeSet) {
```
trustin: How about accepting `MediaType...` and `Iterable<MediaType>`? `MediaTypeSet` is `Iterable<MediaType>`, so the two methods will cover `MediaTypeSet` as well.

core/src/main/java/com/linecorp/armeria/common/MediaTypeSet.java
```diff
@@ -60,16 +60,36 @@
...
+    }
+
+    /**
+     * Returns the {@link MediaTypeSet} which is composed of the specified {@link MediaType}s.
+     */
+    public static MediaTypeSet of(Iterable<MediaType> mediaTypes) {
+        return new MediaTypeSet(ImmutableList.copyOf(requireNonNull(mediaTypes, "mediaTypes")));
```
trustin: How about checking if `mediaTypes` is already a `MediaTypeSet`?

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
minwoox: I was a little bit worried because it will make this builder have as twice methods.
But it seems not a big deal so let me make a change for that. Thanks!

core/src/main/java/com/linecorp/armeria/client/logging/ContentPreviewingClient.java
minwoox: Ah yes. 😄

core/src/main/java/com/linecorp/armeria/common/MediaTypeSet.java
minwoox: Yes, we should.

core/src/main/java/com/linecorp/armeria/common/MediaTypeSet.java
minwoox: That's good idea!

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
minwoox: Fixed!

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
minwoox: Added `@VisibleForTesting`

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
```diff
@@ -199,17 +240,17 @@ public ContentPreviewerFactory build() {
...
+            Iterable<MediaType> meidaTypes) {
         return (ctx, headers) -> {
             final MediaType contentType = headers.contentType();
             if (contentType == null) {
                 return false;
             }
-            return mediaTypeSet.match(contentType) != null;
+            return Streams.stream(meidaTypes).anyMatch(contentType::belongsTo);
         };
     }
```
trustin: How about keep accepting `MediaTypeSet`? The callers of this method could create a `MediaTypeSet` from the given `MediaType`s.

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * A builder which builds a {@link ContentPreviewerFactory}.
+ */
+public final class ContentPreviewerFactoryBuilder {
+
+    //TODO(minwoox): Add setters for the seprate request and response previewer.
+
+    private final ImmutableList.Builder<PreviewSpec> previewSpecsBuilder = ImmutableList.builder();
+    private int maxLength;
+    private Charset defaultCharset = ArmeriaHttpUtil.HTTP_DEFAULT_CONTENT_CHARSET;
```
anuraaga: I think we either need to 

- change this to UTF_8
- Have a special case for JSON which only supports UTF-8

I lean towards the former, I think users will have more gotchas with a default matching the ancient HTTP standard vs common practice nowadays, UTF-8 for everything.

core/src/main/java/com/linecorp/armeria/common/logging/PreviewSpec.java
```diff
@@ -0,0 +1,71 @@
...
+    BiPredicate<? super RequestContext, ? super HttpHeaders> predicate() {
+        return predicate;
+    }
+
+    PreviewMode mode() {
+        return mode;
+    }
+
+    @Nullable
+    Charset defaultCharset() {
```
anuraaga: Does a preview spec also need a default charset? I'm having trouble understanding how this is used.

core/src/main/java/com/linecorp/armeria/common/logging/DefaultContentPreviewFactory.java
```diff
@@ -0,0 +1,107 @@
...
+        if (contentType != null) {
+            final Charset charset = contentType.charset();
+            if (charset != null) {
+                return new TextContentPreviewer(maxLength, charset);
+            }
+
+            if ("text".equals(contentType.type()) ||
+                textSubTypes.contains(contentType.subtype()) ||
+                textSubTypeSuffixes.stream().anyMatch(contentType.subtype()::endsWith) ||
+                contentType.is(MediaType.FORM_DATA)) {
```
ikhoon: You don't need to address this preview. I just thought we can optimize this after releasing 1.0. 
`MediaType` class has `KNOWN_TYPES`. If the `contentType` is `KNOWN_TYPES` these predicate results are cachable with limited size. :-)

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
minwoox: > the ancient HTTP standard vs common practice nowadays

Yes, it's time to throw the ancient standard away. Thanks!

core/src/main/java/com/linecorp/armeria/common/logging/PreviewSpec.java
minwoox: Maybe it's too much to have different default charsets for different media types?
I think I can just remove this and use one default charset for every text preview when the charset is not specified in the content type. Thanks!

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
trustin: Modern clients almost always send content type with a `charset` attribute, so defaulting to ISO-8859-1 makes more sense in my opinion. The only exception is JSON, which is by definition UTF-8 if no charset is specified.

(+1 for latter therefore)

anuraaga: We seem to already have a strange dichotomy

- `StringRequestConverterFunction` and `StringResponseConverterFunction` have different defaults, former is ISO-8859-1 and latter is UTF-8. I don't see any reason for this other than we feel it's tremendously more convenient for responses to default to UTF-8, but this seems to apply to requests too. Why respect HTTP RFC for requests but not responses?
  - Also an example of requests that don't have charset applied even in modern times - manual debug requests from curl

- `spring-boot` defaults to UTF-8 for both. I think that's a good enough industry standard for us to follow (actually our annotated service is not a drop-in replacement for spring if we don't)

https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-application-properties.html
Search `spring.http.encoding.charset`
Confirmed in code
https://github.com/spring-projects/spring-boot/blame/711391cf2f27dd2637a1bb4d7847a2323b6c4e16/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/server/Encoding.java#L37
https://github.com/spring-projects/spring-boot/blob/711391cf2f27dd2637a1bb4d7847a2323b6c4e16/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/http/HttpMessageConvertersAutoConfiguration.java#L81

core/src/main/java/com/linecorp/armeria/common/logging/DefaultContentPreviewFactory.java
minwoox: Ah yes, it's probably a good idea to cache the result. But we have to check the performance because we have to do it in thread safely and the size of cached should be limited which means we have to use caffeine. 😉

core/src/main/java/com/linecorp/armeria/common/logging/LengthLimitingContentPreviewer.java
```diff
@@ -0,0 +1,104 @@
...
+        this.maxLength = maxLength;
+        inflatedMaxLength = inflateMaxLength(maxLength, charset);
+    }
+
+    private static int inflateMaxLength(int maxLength, @Nullable Charset charset) {
+        if (charset == null) {
+            return maxLength;
+        }
+        final long maxBytesPerChar = (long) Math.ceil(CharsetUtil.encoder(charset).maxBytesPerChar());
+        return (int) Long.min(Integer.MAX_VALUE, maxBytesPerChar * maxLength);
```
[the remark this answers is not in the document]
minwoox: Fixed. 😉

core/src/main/java/com/linecorp/armeria/common/logging/TextContentPreviewer.java
```diff
@@ -13,19 +13,25 @@
...
+        super(maxLength, charset);
+        this.charset = charset;
+    }
 
     @Override
-    public ContentPreviewer get(RequestContext ctx, HttpHeaders headers) {
-        return ContentPreviewer.disabled();
+    String produce(ByteBuf wrappedBuffer) {
+        return wrappedBuffer.toString(wrappedBuffer.readerIndex(),
+                                      wrappedBuffer.readableBytes(), charset);
```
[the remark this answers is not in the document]
minwoox: Just copied without much thinking as well. 😅

core/src/main/java/com/linecorp/armeria/common/logging/ContentPreviewerFactoryBuilder.java
trustin: I guess even ourselves are confused about the defaults... :sweat_smile: Let's just default to UTF-8 then.

minwoox: Fixed. 😉

core/src/main/java/com/linecorp/armeria/common/logging/DefaultContentPreviewFactory.java
ikhoon: Yes, we should prove optimized code with benchmarks. By the way I thought we might build this cache when class is loading so we don’t care about thread safety.

minwoox: How could we do that? I don't probably understand what you mean. Could you elaborate more please? 😄

ikhoon: Let me try this 😉and will give a feedback.

ikhoon: I thought this kind of cache. :-) Even though I didn't run a benchmark yet. 
```diff
diff --git a/core/src/main/java/com/linecorp/armeria/common/MediaType.java b/core/src/main/java/com/linecorp/armeria/common/MediaType.java
index 9875a1ae0..e7e504290 100644
--- a/core/src/main/java/com/linecorp/armeria/common/MediaType.java
+++ b/core/src/main/java/com/linecorp/armeria/common/MediaType.java
@@ -127,6 +127,10 @@ public final class MediaType {
 
     private static final Map<MediaType, MediaType> KNOWN_TYPES = Maps.newHashMap();
 
+    public static Map<MediaType, MediaType> knownTypes() {
+        return KNOWN_TYPES;
+    }
+
     private static MediaType createConstant(String type, String subtype) {
         MediaType mediaType =
                 addKnownType(new MediaType(type, subtype, ImmutableListMultimap.of()));
diff --git a/core/src/main/java/com/linecorp/armeria/common/logging/DefaultContentPreviewFactory.java b/core/src/main/java/com/linecorp/armeria/common/logging/DefaultContentPreviewFactory.java
index 6a4f2d665..405f69b45 100644
--- a/core/src/main/java/com/linecorp/armeria/common/logging/DefaultContentPreviewFactory.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/DefaultContentPreviewFactory.java
@@ -16,8 +16,12 @@
 
 package com.linecorp.armeria.common.logging;
 
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+
 import java.nio.charset.Charset;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.function.BiFunction;
 
 import com.google.common.collect.ImmutableList;
@@ -35,6 +39,12 @@ final class DefaultContentPreviewFactory implements ContentPreviewerFactory {
     private static final List<String> textSubTypes = ImmutableList.of("json", "xml");
     private static final List<String> textSubTypeSuffixes = ImmutableList.of("+json", "+xml");
 
+    private static final Map<MediaType, Boolean> previewableTextMediaTypes =
+            MediaType.knownTypes()
+                     .entrySet()
+                     .stream()
+                     .collect(toImmutableMap(Entry::getKey, entry -> isTextContent0(entry.getValue())));
+
     private final int maxLength;
     private final List<PreviewSpec> previewSpecs;
     private final Charset defaultCharset;
@@ -55,6 +65,22 @@ final class DefaultContentPreviewFactory implements ContentPreviewerFactory {
         return contentPreviewer(ctx, resHeaders);
     }
 
+    private static boolean isTextType(MediaType contentType) {
+        final Boolean isTextType = previewableTextMediaTypes.get(contentType);
+        if (isTextType != null) {
+            return isTextType;
+        } else {
+            return isTextContent0(contentType);
+        }
+    }
+
+    private static boolean isTextContent0(MediaType contentType) {
+        return "text".equals(contentType.type()) ||
+               textSubTypes.contains(contentType.subtype()) ||
+               textSubTypeSuffixes.stream().anyMatch(contentType.subtype()::endsWith) ||
+               contentType.is(MediaType.FORM_DATA);
+    }
+
     private ContentPreviewer contentPreviewer(RequestContext ctx, HttpHeaders headers) {
         for (PreviewSpec previewSpec : previewSpecs) {
             if (previewSpec.predicate().test(ctx, headers)) {
@@ -80,10 +106,7 @@ final class DefaultContentPreviewFactory implements ContentPreviewerFactory {
                 return new TextContentPreviewer(maxLength, charset);
             }
 
-            if ("text".equals(contentType.type()) ||
-                textSubTypes.contains(contentType.subtype()) ||
-                textSubTypeSuffixes.stream().anyMatch(contentType.subtype()::endsWith) ||
-                contentType.is(MediaType.FORM_DATA)) {
+            if (isTextType(contentType)) {
                 return new TextContentPreviewer(maxLength, defaultCharset);
             }
         }
```

minwoox: Ah, I was thinking caching every media type of request, but yes this is a good approach and we don't even need the benchmark because it's obviously winning.
As I left the comment, I will look into this later with other logic which handles text content type. Thanks!

minwoox:
Thanks for reviewing!

PR #2445, opened 2020-02-03
Add `UnstableApi` annotation and hide some internal classes

trustin:
See: #2360

- Added the `UnstableApi` annotation to indicate unstable API.
- Annotated some classes and packages with the `UnstableApi` annotation.
- Hid `DefaultRequestLog` from the public API completely.


trustin:
Please let me know if 1) I missed anything or 2) I added the annotation to a wrong target.

core/src/main/java/com/linecorp/armeria/common/util/UnstableApi.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2019 LINE Corporation
```
minwoox: nit: 2020. 😆

trustin: :rofl:

ikhoon:
Still LGTM

core/src/main/java/com/linecorp/armeria/client/ClientFactoryProvider.java
```diff
@@ -17,10 +17,12 @@
 package com.linecorp.armeria.client;
 
 import com.linecorp.armeria.common.SerializationFormat;
+import com.linecorp.armeria.common.util.UnstableApi;
 
 /**
  * Creates a new {@link ClientFactory} dynamically via Java SPI (Service Provider Interface).
  */
+@UnstableApi
```
anuraaga: Does this effectively mean we only officially support our official client implementations, and user-created custom ones may potentially have breaking changes? I'm ok with either but we may want to document that explicitly (I lean towards not marking this unstable though it seems pretty well-baked).

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -67,6 +67,7 @@
 /**
  * Default {@link ClientRequestContext} implementation.
  */
+@UnstableApi
```
anuraaga: Just want to confirm, I think it's possible to hide some of these completely, is the idea to mark as unstable first and then hide later?

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedDocServicePlugin.java
```diff
@@ -80,6 +81,7 @@
 /**
  * A {@link DocServicePlugin} implementation that supports the {@link AnnotatedService}.
  */
+@UnstableApi
```
anuraaga: This is already in `internal` package

core/src/main/java/com/linecorp/armeria/server/docs/DocServicePlugin.java
```diff
@@ -24,12 +24,14 @@
...
 import com.google.common.collect.ImmutableSet;
 
+import com.linecorp.armeria.common.util.UnstableApi;
 import com.linecorp.armeria.server.Service;
 import com.linecorp.armeria.server.ServiceConfig;
 
 /**
  * Generates the {@link ServiceSpecification}s of the supported {@link Service}s.
  */
+@UnstableApi
```
anuraaga: Should we add to `docs/package-inf.java`?

core/src/main/java/com/linecorp/armeria/unsafe/package-info.java
```diff
@@ -13,14 +13,14 @@
...
  * under the License.
  */
-
 /**
  * Utilities for working with {@link io.netty.buffer.ByteBuf} in an unsafe way. These can improve performance
  * when dealing with large buffers but require careful memory management or there will be memory leaks. Only use
  * these methods if you really know what you're doing.
  */
-
+@UnstableApi
```
anuraaga: Do we also need to add to the classes? (I noticed we did in `grpc-protocol`)

grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcDocServicePlugin.java
```diff
@@ -74,6 +75,7 @@
 /**
  * {@link DocServicePlugin} implementation that supports {@link GrpcService}s.
  */
+@UnstableApi
 public final class GrpcDocServicePlugin implements DocServicePlugin {
```
anuraaga: Guess we should move to `internal` package

grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactoryProvider.java
```diff
@@ -18,10 +18,12 @@
 
 import com.linecorp.armeria.client.ClientFactory;
 import com.linecorp.armeria.client.ClientFactoryProvider;
+import com.linecorp.armeria.common.util.UnstableApi;
 
 /**
  * {@link ClientFactoryProvider} that creates a {@link GrpcClientFactory}.
  */
+@UnstableApi
```
anuraaga: I wonder if it's ok for SPI implementations to be in `internal` package I guess so

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/Tomcat90ProtocolHandler.java
```diff
@@ -28,10 +28,13 @@
 import org.apache.coyote.UpgradeProtocol;
 import org.apache.tomcat.util.net.SSLHostConfig;
 
+import com.linecorp.armeria.common.util.UnstableApi;
+
 /**
  * A {@link ProtocolHandler} for Tomcat 8.5 and above.
  * Do not use; loaded and instantiated by Tomcat via reflection.
  */
+@UnstableApi
```
anuraaga: Is it worth having both `UnstableApi` and `Internal` to separate what definitely won't ever be stable?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryProvider.java
trustin: I'm somewhat unsure about the stability of this API although it stayed unchanged for a very long time. For instance, a user may want to build a `ClientFactory` which is built on top of `GrpcClientFactory` or `THttpClientFactory`.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: Is is possible to hide this class? Any user who wants to write a custom client implementation will need to use this API.

core/src/main/java/com/linecorp/armeria/internal/annotation/AnnotatedDocServicePlugin.java
trustin: Oops!

core/src/main/java/com/linecorp/armeria/server/docs/DocServicePlugin.java
trustin: No, because we have `DocService` there. :wink:

core/src/main/java/com/linecorp/armeria/unsafe/package-info.java
trustin: I did already.

tomcat/src/main/java/com/linecorp/armeria/server/tomcat/Tomcat90ProtocolHandler.java
trustin: I think I can just move this class to the internal package.

core/src/main/java/com/linecorp/armeria/unsafe/package-info.java
anuraaga: Oops not sure why I missed it before

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
anuraaga: If I'm not mistaken, only the constructor is ever needed. So we should be able to hide it by providing a factory in `*RequestContext`.

This class is literally what prompted me to file the issue about over-exposed APIs ;)

grpc/src/main/java/com/linecorp/armeria/client/grpc/GrpcClientFactoryProvider.java
trustin: The `internal` packages are getting more and more crowded. Let me send a follow-up PR which 1) reorganizes the `internal` packages to have the same structure with non-internal packages (common, client, server) and 2) move the SPI impls there.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
trustin: I'm not sure if it's a good idea to have a static factory method with such large number of parameters. Perhaps it's better as it is now, i.e. public class annotated with `@UnstableApi`? It's for advanced users only anyway.

anuraaga: I think it's still worth it. Though if we can merge the semantics with our current builders maybe even better not sure it's possible.

These classes leak a lot of implementation details, including a whole class `TimeoutController`. I think they're so core that we want to hide them as much as possible.

trustin: Just realized that `setTimeoutController` and `TimeoutController` can be package-private. I think we can hide `DefaultClientRequestContext` if we do something about `DefaultClientRequestContext.init(EndpointGroup)`. Not sure if it makes sense to move it to `ClientRequestContext`..

trustin: Let me keep thinking about how we can solve this. Meanwhile, let me leave this as it is and address it in another PR. At least `TimeoutController` is now removed from public API.

trustin:
Let me merge and continue working on the follow-up PR for https://github.com/line/armeria/pull/2445#discussion_r374561371 and maybe https://github.com/line/armeria/pull/2445#discussion_r374472760 if I find a good way.

trustin:
By the way, code coverage value seems way off. Maybe some regression related with recent Gradle scripts update?

anuraaga:
LGTM

Will take a look at the coverage issue.

trustin:
Thanks for reviewing. We are getting pretty close now :-)

Issue #2440, opened 2020-02-03
Add a static constructor for FailFastException


jrhee17:
We would like to be able to create mock tests for cases when circuitbreaker is opened.
However, it seems like `FailFastException` is `final`, and doesn't have any public static constructors.
It would be great if `FailFastException` had a method like `ClosedSessionException.get`

```
public static FailFastException get(CircuitBreaker circuitBreaker) {
    return new FailFastException(circuitBreaker);
}
```

Let me know if this is something you guys are open to considering 🙇

trustin:
Sounds good to me. I think it's just OK to make the constructor public, add Javadoc and add `requireNonNull()` on the parameter.

JunoJunho:
Can I handle this issue?

JunoJunho:
I just have a simple curiosity on the constructor's visibility.

Is there any reason to make the constructor as default?

PR #2446, opened 2020-02-03
Make FailFastException public

JunoJunho:
Motication:

User got a trouble on make test code when (s)he tried to do a test on
circuitBreaker and when it raises an exception (FailFastException).
Originally, to reproduce the exception, user needs to use Reflection to
make exception.

Modification:

Make FailFastException consturctor public, add java doc comment for it.
Make parameter checking for nonNull

Result:

public constructor for FailFastException and java doc

Issued by #2440

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/FailFastException.java
+++ b/core/src/main/java/com/linecorp/armeria/client/circuitbreaker/FailFastException.java
@@ -18,8 +18,6 @@ package com.linecorp.armeria.client.circuitbreaker;
 
 import static java.util.Objects.requireNonNull;
 
-import javax.annotation.Nonnull;
-
 import com.linecorp.armeria.common.Flags;
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/linecorp/armeria/client/circuitbreaker/FailFastException.java
```diff
@@ -27,7 +31,11 @@
 
     private final CircuitBreaker circuitBreaker;
 
-    FailFastException(CircuitBreaker circuitBreaker) {
+    /**
+     * Construct an {@link FailFastException} that a request has been failed by {@code circuitBreaker}.
+     */
+    public FailFastException(@Nonnull CircuitBreaker circuitBreaker) {
```
ikhoon: Armeria does not use `@Nonnull` since everything is non-null except `@Nullable`. You can check the detail definition in their `package-info.java`.
https://github.com/line/armeria/blob/master/CONTRIBUTING.md#use-nullable

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/FailFastException.java
```diff
@@ -27,7 +31,11 @@
 
     private final CircuitBreaker circuitBreaker;
 
-    FailFastException(CircuitBreaker circuitBreaker) {
+    /**
+     * Construct an {@link FailFastException} that a request has been failed by {@code circuitBreaker}.
```
trustin: How about just `Creates a new instance with the specified {@link CircuitBreaker}.`

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/FailFastException.java
JunoJunho: Ooops!

Thanks for pointing!

core/src/main/java/com/linecorp/armeria/client/circuitbreaker/FailFastException.java
JunoJunho: No need for `FastFailException`?

I updated the doc.

PR #2448, opened 2020-02-04
Introduce Pooled* versions of APIs which delegate to the stand…

anuraaga:
…ard ones while assuring all operations use pooled objects.

Currently APIs that use unsafe patterns, exposing reference counted objects, are mixed with the normal API. Because they're dangerous, we should hide them as much as possible, so this makes using the unsafe API explicit by requiring using the unsafe API entry points. Unsafe API users also get a QOL increase at the same time by adding the ability to use this API with try/resources.

See #1938 for prior discussion.


core/src/main/java/com/linecorp/armeria/internal/HttpRequestAggregator.java
```diff
@@ -14,22 +14,27 @@
  * under the License.
  */
 
-package com.linecorp.armeria.common;
+package com.linecorp.armeria.internal;
```
trustin: It's now `internal.common`.

trustin:
Looks good to me so far. Thoughts, @minwoox and @ikhoon ?

ikhoon:
I agreed. The API makes sense to me. 👍

anuraaga:
Have mostly finished implementing the new APIs for the client side and using them, need to backfill some tests and should be ready.

Removing old APIs related to pooled instead of deprecating since they're advanced anyways.

Some points wondering what people think about
- I was hoping `PooledHttpResponse` would use the same named methods, `aggregate` so it's clear that this returns pooled for everything, unlike a normal `HttpResponse`. But I found it affects the non-pooled case significantly, by making users have to use a return value like `CompletableFuture<? extends AggregatedHttpResponse>`. So went ahead and went back to the old names `aggregateWithPooledObjects`. Let me know if this makes sense.

- Currently we have just `withEndOfStream()` but several locations set `endOfStream` based on a `boolean`, not just always setting to true. I'm wondering if it makes sense to add `withEndOfStream(boolean)`, or add a separate factory instead.

- Was thinking of changing `SubscriptionOption` to a `AbstractOption` instead of enum so we can move `WITH_POOLED_OBJECTS` into `unsafe`. We could also move `EventExecutor` into a `SubscriptionOption` currently it's a bit weird that one option is a method argument and the others are not. But would such a change be overkill?

anuraaga:
@trustin @minwoox @ikhoon Any thoughts on some of the open questions?

trustin:
Oh, sorry. Will check this week. Was busy with a new release :sweat_smile:

anuraaga:
@trustin Happen to have any thoughts on this?

trustin:
Hey, did a first pass and looks like a really nice work. :clap: 

> So went ahead and went back to the old names `aggregateWithPooledObjects`. Let me know if this makes sense.

SGTM.

> I'm wondering if it makes sense to add `withEndOfStream(boolean)`, or add a separate factory instead.

Both sounds good to me. `withEndOfStream(boolean)` would be simpler to write though. 

> Was thinking of changing `SubscriptionOption` to a `AbstractOption` instead of enum so we can move `WITH_POOLED_OBJECTS` into `unsafe`. We could also move `EventExecutor` into a `SubscriptionOption` currently it's a bit weird that one option is a method argument and the others are not. But would such a change be overkill?

That's a good question. I'm somewhat concerned about the instantiation cost of the options though, although we can use singletons for boolean options. Probably worth a separate discussion issue?

minwoox:
> currently it's a bit weird that one option is a method argument and the others are not.

I agree. It's bit weird now.

anuraaga:
Thanks for the comments - it's always hard to get back after a while of pending but I should be able to work on this today and incorporate the feedback / brainstorm a little more.

core/src/main/java/com/linecorp/armeria/unsafe/ByteBufHttpData.java
```diff
@@ -20,30 +20,53 @@
...
+    private static final AtomicIntegerFieldUpdater<ByteBufHttpData>
+            closedUpdater = AtomicIntegerFieldUpdater.newUpdater(ByteBufHttpData.class, "closed");
+
+    static final ByteBufHttpData EMPTY = new ByteBufHttpData(
+            new EmptyByteBuf(UnpooledByteBufAllocator.DEFAULT), false);
+
+    /**
+     * Converts non-pooled {@link HttpData} into {@link PooledHttpData}.
+     */
+    static PooledHttpData convert(HttpData data) {
```
ikhoon: I think `from` is more suitable for our API style.
```suggestion
    static PooledHttpData from(HttpData data) {
```

anuraaga:
Really sorry for taking forever to get back to this, has been hard to find the concentration for such a complex change these weeks. Think I can finally wrap this up soon though.

I'm still missing unit tests but think the general API is ready for review. I found an unexpected aspect where for example `PooledHttpService` has two `serve` methods, one with normal request (mainly for use from rest of framework) and one with pooled. An alternative would mean every place we accept a service / client would need to be updated to accept the pooled version too which seems worse.

Anyways, let me know what you think of the current API structure and I'll fill in unit tests.

trustin:
No worries, @anuraaga. I appreciate that you didn't forget about this and made a good progress! Let me review it soon.

core/src/main/java/com/linecorp/armeria/unsafe/client/PooledHttpClient.java
```diff
@@ -0,0 +1,56 @@
...
+import com.linecorp.armeria.client.Endpoint;
+import com.linecorp.armeria.client.HttpClient;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.util.Unwrappable;
+import com.linecorp.armeria.unsafe.common.PooledHttpRequest;
+import com.linecorp.armeria.unsafe.common.PooledHttpResponse;
+
+/**
+ * Sends an {@link HttpRequest} to a remote {@link Endpoint}.
```
trustin, line 29: Could we add a little bit more information about pooling?

core/src/main/java/com/linecorp/armeria/unsafe/common/PooledHttpData.java
```diff
@@ -0,0 +1,133 @@
...
+        requireNonNull(buf, "buf");
+        if (!buf.isReadable()) {
+            return ByteBufHttpData.EMPTY;
+        }
+        return new ByteBufHttpData(buf, false);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
trustin, line 130: nit: probably OK to remove this?

core/src/main/java/com/linecorp/armeria/unsafe/client/PooledWebClient.java
```diff
@@ -0,0 +1,208 @@
...
+     * Creates a {@link PooledWebClient} that delegates to the provided {@link WebClient} for issuing requests.
+     */
+    static PooledWebClient of(WebClient delegate) {
+        return new DefaultPooledWebClient(delegate);
+    }
+
+    /**
+     * Sends the specified HTTP request.
+     */
+    PooledHttpResponse execute(HttpRequest req);
```
minwoox, line 48: Shouldn't we add `@Override` to all methods here?

core/src/main/java/com/linecorp/armeria/unsafe/common/PooledAggregatedHttpRequest.java
```diff
@@ -0,0 +1,50 @@
...
+        requireNonNull(delegate, "delegate");
+        if (delegate instanceof PooledAggregatedHttpRequest) {
+            return (PooledAggregatedHttpRequest) delegate;
+        }
+        return new DefaultPooledAggregatedHttpRequest(delegate);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
minwoox, line 44: Can remove this?

core/src/main/java/com/linecorp/armeria/unsafe/common/PooledAggregatedHttpResponse.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * An {@link AggregatedHttpResponse} using pooled buffers for the content. Make sure to call
+ * {@link AutoCloseable#close()} on this response or the {@code content} to release pooled resources.
+ */
+public interface PooledAggregatedHttpResponse extends AggregatedHttpResponse, SafeCloseable {
+
+    /**
+     * {@inheritDoc}
+     */
```
minwoox, line 30: ditto

core/src/main/java/com/linecorp/armeria/unsafe/common/PooledHttpData.java
```diff
@@ -0,0 +1,133 @@
...
+import com.linecorp.armeria.common.HttpData;
+import com.linecorp.armeria.common.util.SafeCloseable;
+import com.linecorp.armeria.unsafe.client.PooledWebClient;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
+import io.netty.buffer.Unpooled;
+
+/**
+ * A {@link HttpData} that is backed by a pooled {@link ByteBuf} for optimizing certain internal use cases. Not
```
minwoox, line 30: nit: An {@link HttpData}

core/src/main/java/com/linecorp/armeria/unsafe/common/PooledHttpData.java
```diff
@@ -0,0 +1,133 @@
...
+        requireNonNull(data, "obj");
+        if (data instanceof PooledHttpData) {
+            return (PooledHttpData) data;
+        }
+
+        return new ByteBufHttpData(Unpooled.wrappedBuffer(data.array()), data.isEndOfStream());
+    }
+
+    /**
+     * Converts the specified Netty {@link ByteBuf} into an {@link PooledHttpData}. The buffer is not copied;
```
minwoox, line 113: nit: a {@link PooledHttpData}

core/src/main/java/com/linecorp/armeria/unsafe/common/PooledHttpRequest.java
```diff
@@ -0,0 +1,103 @@
...
+import com.linecorp.armeria.common.stream.SubscriptionOption;
+import com.linecorp.armeria.common.util.EventLoopCheckingFuture;
+import com.linecorp.armeria.internal.common.HttpRequestAggregator;
+
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.PooledByteBufAllocator;
+import io.netty.util.concurrent.EventExecutor;
+
+/**
+ * A streamed HTTP/2 {@link Request} which returns pooled buffers.
```
minwoox, line 35: Should we use `HTTP` instead of `HTTP/2` because the client might use HTTP/1.1?

core/src/main/java/com/linecorp/armeria/unsafe/client/SimplePooledDecoratingHttpClient.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public final HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Exception {
+        return execute(ctx, PooledHttpRequest.of(req), delegate());
+    }
+
+    @Override
+    public PooledHttpResponse execute(ClientRequestContext ctx, PooledHttpRequest req) throws Exception {
+        return PooledHttpResponse.of(execute(ctx, req, delegate()));
```
minwoox, line 49: How about changing the order of methods parameters? `execute(delegate(), ctx, req))` looks natural to me

core/src/main/java/com/linecorp/armeria/unsafe/server/SimplePooledDecoratingHttpService.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public final HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
+        return serve(ctx, PooledHttpRequest.of(req), pooledDelegate);
+    }
+
+    @Override
+    public PooledHttpResponse serve(ServiceRequestContext ctx, PooledHttpRequest req) throws Exception {
+        return PooledHttpResponse.of(serve(ctx, req, delegate()));
```
minwoox, line 53: ditto

core/src/main/java/com/linecorp/armeria/unsafe/common/PooledHttpRequest.java
minwoox: Let's update both because we don't want to update again when HTTP/3 comes out. 😄

core/src/main/java/com/linecorp/armeria/client/unsafe/DefaultPooledHttpClient.java
```diff
@@ -0,0 +1,37 @@
...
+final class DefaultPooledHttpClient extends AbstractUnwrappable<HttpClient> implements PooledHttpClient {
+
+    DefaultPooledHttpClient(HttpClient delegate) {
+        super(delegate);
+    }
+
+    @Override
+    public PooledHttpResponse execute(ClientRequestContext ctx, PooledHttpRequest req) throws Exception {
+        // Always a wrapped non-pooled client, make sure it gets a normal request.
+        assert !(delegate() instanceof PooledHttpClient);
```
trustin, line 34: nit: How about moving this check to the constructor?

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledAggregatedHttpRequest.java
```diff
@@ -0,0 +1,50 @@
...
+        requireNonNull(delegate, "delegate");
+        if (delegate instanceof PooledAggregatedHttpRequest) {
+            return (PooledAggregatedHttpRequest) delegate;
+        }
+        return new DefaultPooledAggregatedHttpRequest(delegate);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
trustin, line 44: OK to remove?

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
```diff
@@ -20,12 +20,15 @@
...
 
+import com.linecorp.armeria.common.util.UnstableApi;
+
 /**
  * A complete HTTP message whose content is readily available as a single {@link HttpData}. It can be an
  * HTTP request or an HTTP response depending on what header values it contains. For example, having a
  * {@link HttpHeaderNames#STATUS} header could mean it is an HTTP response.
  */
-interface AggregatedHttpMessage {
+@UnstableApi
```
trustin, line 30: Perhaps OK to remove this?

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledAggregatedHttpResponse.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    PooledHttpData content();
+
+    /**
+     * {@inheritDoc}
+     */
```
trustin, line 36: No need to use `@inheritDoc` if you don't add anything to Javadoc

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledHttpData.java
```diff
@@ -0,0 +1,133 @@
...
+ * effects.
+ *
+ * <p>For example, <pre>{@code
+ *
+ * HttpResponse response = client.get("/");
+ * response.aggregateWithPooledObjects(ctx.alloc(), ctx.executor())
+ *     .thenApply(aggResp -> {
+ *         // try-with-resources here ensures the content is released if it is a ByteBufHttpData, or otherwise
+ *         // is a no-op if it is not.
+ *         try (HttpData content = aggResp.content()) {
```
trustin, line 59: This example needs an update because `HttpData` is not `Closeable` anymore.

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledHttpRequest.java
```diff
@@ -0,0 +1,119 @@
...
+    default CompletableFuture<AggregatedHttpRequest> aggregate(EventExecutor executor) {
+        return HttpRequest.super.aggregate(executor);
+    }
+
+    /**
+     * Converts this {@link PooledHttpRequest} to an unpooled {@link HttpRequest}. Only one of this
+     * {@link PooledHttpRequest} or the returned {@link HttpRequest} can be subscribed to, if both are attempted
+     * to be used, it will cause an error or bad behavior.
+     */
+    HttpRequest toUnpooled();
```
trustin, line 118: `toUnpooled` or `asUnpooled`?

core/src/main/java/com/linecorp/armeria/server/file/StreamingHttpFile.java
```diff
@@ -223,7 +224,7 @@ private void doRead(HttpResponseWriter res, T in, long offset, long end,
 
                     final HttpFileBuilder builder =
                             HttpFile.builder(array != null ? HttpData.wrap(array)
-                                                           : new ByteBufHttpData(buf, true),
+                                                           : PooledHttpData.wrap(buf).withEndOfStream(),
```
trustin, line 227: Question: Should we add `withEndOfStream(boolean)` which returns `this` if `endOfStream` did not change?

core/src/main/java/com/linecorp/armeria/server/unsafe/DefaultPooledHttpService.java
```diff
@@ -0,0 +1,45 @@
...
+final class DefaultPooledHttpService extends AbstractUnwrappable<HttpService> implements PooledHttpService {
+
+    DefaultPooledHttpService(HttpService delegate) {
+        super(delegate);
+    }
+
+    @Override
+    public PooledHttpResponse serve(ServiceRequestContext ctx, PooledHttpRequest req) throws Exception {
+        // Always a wrapped non-pooled service, make sure it gets a normal request.
+        assert !(delegate() instanceof PooledHttpService);
```
trustin, line 37: nit: Should we check this in the constructor?

core/src/test/java/com/linecorp/armeria/internal/common/DefaultHttpRequestTest.java
```diff
@@ -53,19 +54,19 @@
...
         final Thread mainThread = Thread.currentThread();
         final DefaultHttpRequest req = new DefaultHttpRequest(RequestHeaders.of(HttpMethod.GET, "/foo"));
-        final CompletableFuture<AggregatedHttpRequest> future;
+        final CompletableFuture<? extends AggregatedHttpRequest> future;
 
         // Practically same execution, but we need to test the both case due to code duplication.
         if (executorSpecified) {
             if (withPooledObjects) {
-                future = req.aggregateWithPooledObjects(
+                future = PooledHttpRequest.of(req).aggregateWithPooledObjects(
```
trustin, line 62: Could use `aggregate()` instead of `aggregateWithPooledObjects`?

core/src/test/java/com/linecorp/armeria/internal/common/DefaultHttpRequestTest.java
```diff
@@ -53,19 +54,19 @@
...
                         CommonPools.workerGroup().next(), PooledByteBufAllocator.DEFAULT);
             } else {
                 future = req.aggregate(CommonPools.workerGroup().next());
             }
         } else {
             if (withPooledObjects) {
-                future = req.aggregateWithPooledObjects(PooledByteBufAllocator.DEFAULT);
+                future = PooledHttpRequest.of(req).aggregateWithPooledObjects(PooledByteBufAllocator.DEFAULT);
```
trustin, line 69: Ditto

core/src/main/java/com/linecorp/armeria/client/unsafe/DefaultPooledHttpClient.java
anuraaga: I put it here more as documentation than anything so think it's better here. But if the comment is enough I can remove it?

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpMessage.java
anuraaga: Yeah intended to mark that users shouldn't need to implement this but guess it doesn't hurt either

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledHttpRequest.java
anuraaga: We have others like `toHttpResponse` and I think I followed that. Not sure if either is better 😅

core/src/main/java/com/linecorp/armeria/server/file/StreamingHttpFile.java
anuraaga: Yeah seems good

core/src/test/java/com/linecorp/armeria/internal/common/DefaultHttpRequestTest.java
anuraaga: Don't think so since this is `withPooledObjects` branch

anuraaga:
Build is finally green, missing some test coverage especially on client side, but one other big change I'm thinking is to add `subscribeWithPooledObjects` and somehow hide the subscriber option from users. Currently, for the same method `subscribe`, the pooled version has different behavior than the base classs and I guess it's a bit dangerous. If the API could have been entirely clean with `aggregate` it could have been ok but since we need `aggregateWIthPooledObjects` anyways I think it's ok to do similar to subscribe. If the pooled classes have same behavior as base class for the nonpooled methods, I can remove `toUnpooled` too. Any thoughts?

trustin:
> Any thoughts?

Sounds good to me (for both ideas). What do you think, @minwoox and @ikhoon?

core/src/main/java/com/linecorp/armeria/client/unsafe/DefaultPooledHttpClient.java
trustin: I don't mind. Fine with having it here.

anuraaga:
Ok finally have the API in an ok state I think. 

- Unsafe APIs are completely gone from the normal packages. Split `SubscriptionOption` into a normal and internal version for this
- Users can use the unsafe `PooledWebClient`, `SimplePooledDecoratingService` etc type methods to have access to the unsafe API all the time if they want to
- Users can opt-in on a one-off basis using `PooledHttpRequest.of` etc

While having lots of blahWithPooledObjects methods isn't ideal, I think it's ok and using `@Deprecated` to lead users to the correct methods should prevent using the wrong one most of the time I think. Let me know if this final API looks OK. Will backfill remaining tests.

anuraaga:
Ah also, one lazy point was not copying in `drainAll` to the unsafe APIs - it just seemed like so much more API surface and I filed #2782 since there didn't seem to be so much utility. If it's mostly for testing maybe it makes sense to move it out. Otherwise let me know and I'll go ahead and copy them in too with `drainAllWithPooledObjects`

trustin:
Maybe (or not) off-topic, but speaking of API surface, what would happen in terms of pooled objects if we make `StreamingMessage` implement `Flux`? I previous talked about switching to Reactor completely, but maybe extending it might be a more reasonable & less breaking approach.

anuraaga:
I think I previously also was interested it, after playing with `Flux` less so though :) Soo many methods make always make it very boring at best, tedious at worst, finding the method I want to actually use to solve a given problem...

But if it's useful it still makes sense, but for the task here of separating safe and unsafe API, do you see anything `Flux` would get it?

anuraaga:
Think this is finally ready for action!

trustin:
> But if it's useful it still makes sense, but for the task here of separating safe and unsafe API, do you see anything `Flux` would get us?

No, I was just curious if you have anything in mind in terms of releasing pooled objects properly with all those chaining methods in `Flux`. Didn't have anything specific in mind.

anuraaga:
@trustin Thanks - using Flux, and some of its disposal systems, could make it easier for a user to use the unsafe APIs. I think the APIs here are still the good entry points though for both non-Flux and Flux users :)

core/src/main/java/com/linecorp/armeria/common/unsafe/ByteBufHttpData.java
```diff
@@ -169,11 +181,22 @@ public String toString() {
...
 
     @Override
-    public ByteBufHttpData withEndOfStream() {
-        return new ByteBufHttpData(buf, true);
+    public ByteBufHttpData withEndOfStream(boolean endOfStream) {
+        if (endOfStream == this.endOfStream) {
+            return this;
+        }
+        return new ByteBufHttpData(buf, endOfStream);
+    }
```
ikhoon, line 193: This is not normal behavior, however a `IllegalReferenceCountException` could be raised by:
```java
final ByteBuf byteBuf = Unpooled.wrappedBuffer("hello".getBytes());
final PooledHttpData pooledHttpData1 = PooledHttpData.wrap(byteBuf);
pooledHttpData1.close();
final PooledHttpData polledHttpData2 = pooledHttpData1.withEndOfStream(true);
// io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
polledHttpData2.close();
```
This method does not copy the `ByteBufHttpData.closed`.

core/src/main/java/com/linecorp/armeria/server/unsafe/SimplePooledDecoratingHttpService.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Execute the {@code req} with the given {@code ctx}.
+     *
+     * @see SimpleDecoratingHttpService#serve(ServiceRequestContext, HttpRequest)
+     */
+    protected abstract HttpResponse serve(
+            PooledHttpService delegate, ServiceRequestContext ctx, PooledHttpRequest req)
+            throws Exception;
```
trustin, line 61: Ditto - two extension points that may be confusing

thrift/src/main/java/com/linecorp/armeria/internal/client/thrift/THttpClientDelegate.java
```diff
@@ -139,21 +141,24 @@ public RpcResponse execute(ClientRequestContext ctx, RpcRequest call) throws Exc
...
+                          .aggregateWithPooledObjects(ctx.eventLoop(), ctx.alloc());
 
             future.handle((res, cause) -> {
                 if (cause != null) {
                     handlePreDecodeException(ctx, reply, func, Exceptions.peel(cause));
                     return null;
                 }
 
-                try {
+                try (SafeCloseable unused = res) {
```
trustin, line 161: This looks funny, just sayin' 😆

thrift/src/main/java/com/linecorp/armeria/server/thrift/THttpService.java
```diff
@@ -352,20 +353,20 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
...
-                                               MediaType.PLAIN_TEXT_UTF_8,
-                                               Exceptions.traceText(cause));
-                } else {
-                    errorRes = HttpResponse.of(HttpStatus.INTERNAL_SERVER_ERROR);
-                }
-                responseFuture.complete(errorRes);
-                return null;
-            }
-
+        PooledHttpRequest.of(req).aggregateWithPooledObjects(
```
trustin, line 356: This makes me wonder which is better - `req.toPooled()` vs. `PooledHttpRequest.of(req)`. Should we make our users mention the type explicitly just because it's an advanced API? 🤔

core/src/main/java/com/linecorp/armeria/client/unsafe/SimplePooledDecoratingHttpClient.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Execute the {@code req} with the given {@code ctx}.
+     *
+     * @see SimpleDecoratingHttpClient#execute(ClientRequestContext, HttpRequest)
+     */
+    protected abstract HttpResponse execute(
+            PooledHttpClient client, ClientRequestContext ctx, PooledHttpRequest req)
+            throws Exception;
```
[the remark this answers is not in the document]
anuraaga: Ah I think I meant to make the other one `final` too. Goal of this method was to remove some boilerplate in user code

- `PooledHttpResponse.of(`
- `(PooledHttpClient) delegate()`

If the other is final is it ok?

core/src/main/java/com/linecorp/armeria/common/stream/SubscriptionOption.java
```diff
@@ -19,27 +19,22 @@
...
  * @see StreamMessage#subscribe(Subscriber, EventExecutor, SubscriptionOption...)
  */
-public enum SubscriptionOption {
-
-    /**
-     * To receive the pooled {@link ByteBuf} and {@link ByteBufHolder} as is, without making a copy.
-     * If you don't know what this means, do not specify this when you subscribe the {@link StreamMessage}.
-     */
-    WITH_POOLED_OBJECTS,
+public interface SubscriptionOption {
```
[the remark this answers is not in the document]
anuraaga: Since the original options were effectively booleans (we only compare equality with an enum) I figured it's ok to use this pattern which replicates it. `AbstractOption` has a lot of complexity because it supports arbitrary types, but I don't think we need it here, it would just make the code in general more verbose / slower.

core/src/main/java/com/linecorp/armeria/common/unsafe/ByteBufHttpData.java
anuraaga: Yup good idea

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledHttpStreamMessage.java
```diff
@@ -0,0 +1,156 @@
...
+    @Deprecated
+    default void subscribe(Subscriber<? super HttpObject> subscriber) {
+        StreamMessage.super.subscribe(subscriber);
+    }
+
+    /**
+     * Requests to start streaming data to the specified {@link Subscriber} without pooled objects. When
+     * operating on {@link PooledHttpStreamMessage} this should be avoided.
+     *
+     * @deprecated Use {@link #subscribe(Subscriber, SubscriptionOption...)}.
```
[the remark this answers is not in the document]
anuraaga: Thanks!

thrift/src/main/java/com/linecorp/armeria/internal/client/thrift/THttpClientDelegate.java
anuraaga: Does give some credit to go's `defer` statement pattern :)

thrift/src/main/java/com/linecorp/armeria/server/thrift/THttpService.java
anuraaga: Yeah thought about this but lean towards `PooledHttpRequest.of` - a bit more verbose, but it's the only way to keep the concept of unsafe completely out of the unsafe packages. I think this is valuable.

core/src/main/java/com/linecorp/armeria/client/unsafe/SimplePooledDecoratingHttpClient.java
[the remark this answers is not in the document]
anuraaga: Ah speaking of this, I had noticed the same problem in `AbstractHttpService` while working on this but forgot to fix. Added a commit that makes `serve` final in it. Think it makes sense (I think it had some use way back in the day before `HttpService` interface but not anymore)

thrift/src/main/java/com/linecorp/armeria/server/thrift/THttpService.java
trustin: I see. No problem.

core/src/main/java/com/linecorp/armeria/common/stream/SubscriptionOption.java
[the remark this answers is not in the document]
trustin: Ah, yeah. Forgot about the value part. 😅

core/src/main/java/com/linecorp/armeria/client/unsafe/SimplePooledDecoratingHttpClient.java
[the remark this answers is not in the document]
trustin: Good point. Let me add the `breaking change` label, then.

core/src/main/java/com/linecorp/armeria/common/unsafe/ByteBufHttpData.java
```diff
@@ -169,11 +181,22 @@ public String toString() {
 
     @Override
     public InputStream toInputStream() {
-        return new ByteBufInputStream(buf.retainedDuplicate(), true);
+        return new ByteBufInputStream(buf.duplicate(), false);
```
trustin, line 184: Could you explain this change? `retainedDuplicate()` generates less garbage.

anuraaga: I think we discussed this a long time ago, and ended up with it in https://github.com/line/armeria/pull/1938#issue-302381652

With `HttpData` closeable, it's simple for a user to reason about a reference needing to be contained in a single block and we're not working about `InputStream` moving out of the block, in other words, we don't want to create a new reference here that the user will have to close independenty of the `HttpData`.

I don't think there's any relation to garbage, it's just about whether a user can transfer the `InputStream` to a different scope and close it manually.

trustin: Ah, OK. That's fine then. Thanks for elaboration. :bow:

core/src/main/java/com/linecorp/armeria/client/unsafe/SimplePooledDecoratingHttpClient.java
```diff
@@ -0,0 +1,60 @@
...
+    public final PooledHttpResponse execute(ClientRequestContext ctx, PooledHttpRequest req) throws Exception {
+        return PooledHttpResponse.of(execute(delegate(), ctx, req));
+    }
+
+    /**
+     * Execute the {@code req} with the given {@code ctx}.
+     *
+     * @see SimpleDecoratingHttpClient#execute(ClientRequestContext, HttpRequest)
+     */
+    protected abstract HttpResponse execute(
```
minwoox, line 57: question: Shoudln't it better to make this return `PooledHttpResponse`?

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledHttpData.java
```diff
@@ -0,0 +1,146 @@
...
+ *                 throw new IllegalArgumentException("Bad JSON: " + content.toStringUtf8());
+ *             }
+ *         }
+ *     });
+ *
+ * }</pre>
+ *
+ * <p>In this example, it is the initial {@code try (HttpData content = ...} that ensures the data is released.
+ * Calls to methods on {@link HttpData} will all work and can be called any number of times within this block.
+ * If called after the block, or a manual call to {@link PooledHttpData#close}, these methods will fail or
```
minwoox, line 75: nit: `#close}` -> `#close()}`

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledHttpData.java
```diff
@@ -0,0 +1,146 @@
...
+ *                 return OBJECT_MAPPER.readValue(content.toInputStream(), Foo.class);
+ *             } catch (IOException e) {
+ *                 throw new IllegalArgumentException("Bad JSON: " + content.toStringUtf8());
+ *             }
+ *         }
+ *     });
+ *
+ * }</pre>
+ *
+ * <p>In this example, it is the initial {@code try (HttpData content = ...} that ensures the data is released.
```
minwoox, line 73: nit: `HttpData` -> `PooledHttpData`

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledAggregatedHttpRequest.java
```diff
@@ -0,0 +1,47 @@
...
+package com.linecorp.armeria.common.unsafe;
+
+import static java.util.Objects.requireNonNull;
+
+import com.linecorp.armeria.common.AggregatedHttpRequest;
+import com.linecorp.armeria.common.util.SafeCloseable;
+
+/**
+ * An {@link AggregatedHttpRequest} using pooled buffers for the content. Make sure to call
+ * {@link AutoCloseable#close()} on this response or the {@code content} to release pooled resources.
```
minwoox, line 26: nit: on this request

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledHttpRequest.java
```diff
@@ -0,0 +1,113 @@
...
+        requireNonNull(delegate, "delegate");
+        if (delegate instanceof PooledHttpRequest) {
+            return (PooledHttpRequest) delegate;
+        }
+        return new DefaultPooledHttpRequest(delegate);
+    }
+
+    /**
+     * Aggregates this request. The returned {@link CompletableFuture} will be notified when the content and
+     * the trailers of the response are received fully.
```
minwoox, line 52: nit: response -> request

core/src/main/java/com/linecorp/armeria/client/unsafe/DefaultPooledHttpClient.java
anuraaga: Don't think it matters so much - this is just a check against programming issue, the constructor is always called with something that isn't a `PooledHttpClient` My 20 days old comment still stands, I put it mainly as documentation ;)

core/src/main/java/com/linecorp/armeria/client/unsafe/SimplePooledDecoratingHttpClient.java
anuraaga: `PooledHttpResponse` is mainly a way to unlock the unsafe API but the semantics work the same as `HttpResponse` otherwise. I think it's restrictive to return `PooledHttpResponse` here without much benefit (if a user returns `PooledHttpResponse` there will be a no-op `instanceof` check

minwoox: Thanks for the explanation. 😄

trustin:
Seeing a build failure on Windows:

```
PooledWebClientTest > aggregateWithPooledObjects_eventExecutorAndAlloc() FAILED
    java.util.concurrent.CompletionException: org.opentest4j.AssertionFailedError: 
    Expecting:
     <false>
    to be equal to:
     <true>
    but was not.
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
        at java.base/java.util.concurrent.CompletableFuture.uniAcceptNow(CompletableFuture.java:757)
        at java.base/java.util.concurrent.CompletableFuture.uniAcceptStage(CompletableFuture.java:731)
        at java.base/java.util.concurrent.CompletableFuture.thenAccept(CompletableFuture.java:2179)
        at com.linecorp.armeria.client.unsafe.PooledWebClientTest.aggregateWithPooledObjects_eventExecutorAndAlloc(PooledWebClientTest.java:97)
        Caused by:
        org.opentest4j.AssertionFailedError: 
        Expecting:
         <false>
        to be equal to:
         <true>
        but was not.
            at jdk.internal.reflect.GeneratedConstructorAccessor55.newInstance(Unknown Source)
            at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
            at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500)
            at com.linecorp.armeria.client.unsafe.PooledWebClientTest.lambda$aggregateWithPooledObjects_eventExecutorAndAlloc$2(PooledWebClientTest.java:105)
            at java.base/java.util.concurrent.CompletableFuture.uniAcceptNow(CompletableFuture.java:753)
            ... 3 more
```

anuraaga:
Hmm - seems like a flake, but can't repro even running 1000x times :( Do you see anything that might cause that callback not to be called on the expected thread?

trustin:
I was able to reproduce the problem by repeating 10000 times. Failed at 5000+th try.

trustin:
When aggregation is finished really quickly and thus a complete future is returned by `aggregate*()` methods, `thenApply()` will be invoked in the caller thread (`"main"` in this case), which is an expected behavior.

trustin:
Should the future returned by `aggregate*()` call its callbacks from the executor thread, even if the future is done already? I think it's less surprising.

trustin:
However, it means increased round-trip due to an additional `*Async()` call.

trustin:
I think the problem (is it?) I found is not really related with this pull request. Let me just remove the event loop thread check and merge.

trustin:
@anuraaga Would you mind updating the PR description so it can be used as a commit message? (Motivation/Modifications/Result)

anuraaga:
Updated. It is a bug though I think. `CompletableFuture` APIs also have the behaviu where when the executor isn't specified, it's an arbitrary thread but when it's specified in a `*Async` form it's always that executor. When the executor has semantic meaning like a context aware executor this is critical.

core/src/main/java/com/linecorp/armeria/common/unsafe/PooledHttpData.java
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * <p>Any time you receive a {@link PooledHttpData} it will have a single reference that must be released -
+ * failure to release the reference will result in a memory leak and poor performance. You must make sure to do
+ * this by calling {@link PooledHttpData#close()}, usually in a try-with-resources structure to avoid side
+ * effects.
+ *
+ * <p>For example, <pre>{@code
+ *
+ * HttpResponse response = client.get("/");
+ * response.aggregateWithPooledObjects(ctx.alloc(), ctx.executor())
```
adriancole, line 55: this line drifted I think

PR #2449, opened 2020-02-04
Clean up `ServerExtension` and `ServerRule`

trustin:
Motivations:

`ServerExtension` and `ServerRule` needs to:

- Provide the methods that return `Endpoint`
- Return `URI` instead of `String` for URI-returning methods

A user also does not always need to specify path when getting a URI.

Modifications:

- Added `endpoint(SessionProtocol)`, `httpEndpoint()` and
  `httpsEndpoint()`
- Added `uri()`, `httpUri()` and `httpsUri()` methods that do not
  require a path and do return a `URI` instead of a `String`.
  - Also deprecated the old methods that requires a path.
- Added `socketAddress(SessionProtocol)`

Result:

- Hopefully cleaner API (once the deprecated methods are gone.)


trustin:
No idea why, but JaCoCo keeps failing.

trustin:
Let me merge this. Please let me know if you think we have to revive the `uri()` methods that return HTTP *or* HTTPS URIs. I deprecated them because they seem too ambiguous.

PR #2452, opened 2020-02-05
Reorganize the `internal` packages

trustin:
Related: #2445
Motivation:

As we hide many stuff from our public API, the `internal` package is
getting more and more crowded.

Modifications:

- Reorganize the `internal` packages so that the internal classes are
  always placed `internal.{common,client,server,...}.*`, following the
  structure of the non-internal classes.

Result:

- Less cognitive load for devs
- Got some room for moving classes to the `internal` packages


core/src/main/java/com/linecorp/armeria/internal/HttpHeaderSubscriber.java
```diff
@@ -1,91 +0,0 @@
...
-import com.linecorp.armeria.common.stream.CancelledSubscriptionException;
-
-/**
- * A {@link Subscriber} that completes the {@link CompletableFuture} which has taken as the argument in the
- * constructor with {@link HttpHeaders}. The {@link HttpHeaders} contains a status that is not informational.
- * If subscription is finished before subscribing a status, this will complete the future with
- * the {@link HttpHeaders#of()}.
- */
-public final class HttpHeaderSubscriber
-        implements Subscriber<HttpObject>, BiConsumer<Void, Throwable> {
```
trustin: This class has been removed since it's not used anywhere.

trustin:
Let me merge this pull request once build passes, since this is just mechanical package reorg restricted to the internal packages.

trustin:
Note that this PR does not include https://github.com/line/armeria/pull/2445#discussion_r374561371 - Let me send a follow-up PR soon.

PR #2453, opened 2020-02-05
Hide SPI implementations from the public API

trustin:
.. because they are not meant to be used by users.

Also:

- Made `GrpcService` an interface.
  - Extracted the implementation into `FramedGrpcService`.
  - Added `GrpcService.isFramed()`.
- Made `UnframedGrpcService` implement `GrpcService`.
- `GrpcServiceBuilder.build()` now returns a `GrpcService`.
- Added `ThriftService.supportedSerializationFormats()` for consistency
  with `GrpcService`.
  - Deprecated `allowedSerializationFormats()`

Related issue: #2360


grpc/src/main/java/com/linecorp/armeria/server/grpc/FramedGrpcService.java
```diff
@@ -0,0 +1,393 @@
...
+final class FramedGrpcService extends AbstractHttpService implements GrpcService {
+
+    private static final Logger logger = LoggerFactory.getLogger(FramedGrpcService.class);
+
+    /**
+     * Returns a new {@link GrpcServiceBuilder}.
+     */
+    public static GrpcServiceBuilder builder() {
+        return new GrpcServiceBuilder();
+    }
```
ikhoon: Question: Do we need this `builder()` method? This class is package-private and this method doesn't use anywhere.

trustin: Fixed

PR #2454, opened 2020-02-05
Do not use type parameter for builder self type in public API

trustin:
Motivation:

Builder self type parameter is very useful when writing builders with
type hierarchy, but it often results in bad-looking Javadoc, because the
return type is not shown as the actual return type but just as a type
parameter name, such as `B` or `SELF`.

Modifications:

- Remove self type parameters from all public builder types and
  explicitly override the return types.

Result:

- More intuitive Javadoc
- Devs need to be cautious when adding a new method to builders.


trustin:
Will fix checkstyle errors later.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -588,29 +588,48 @@ public ServerBuilder http1MaxChunkSize(int http1MaxChunkSize) {
...
     /**
      * Sets the amount of time to wait after calling {@link Server#stop()} for
      * requests to go away before actually shutting down.
      *
      * @param quietPeriod the number of milliseconds to wait for active
-     *     requests to go end before shutting down. {@link Duration#ZERO} means
-     *     the server will stop right away without waiting.
+     *                    requests to go end before shutting down. {@link Duration#ZERO} means
+     *                    the server will stop right away without waiting.
      * @param timeout the number of milliseconds to wait before shutting
```
minwoox: nit: I think we need to change this as well because it's duration.

PR #2456, opened 2020-02-06
Add EncodingServiceBuilder

minwoox:
Motivation:

`EncodingService` has only public constructors to use as decorator so a user has to cast the type explicitly.
Also, it's not the way we do for decorating. We use builders and `Function`s.

Modifications:
- Add `EncodingServiceBuilder`
- (Breaking)
  - Change the `encodableRequestHeadersPredicate` takes `RequestHeaders`
  - Remove `HttpEncodingService` completely.

Result:
- Cleaner API.


spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/internal/spring/ArmeriaConfigurationUtil.java
```diff
@@ -163,9 +164,12 @@ public static void configureServerWithArmeriaSettings(ServerBuilder server, Arme
 
         final ArmeriaSettings.Compression compression = settings.getCompression();
         if (compression != null && compression.isEnabled()) {
+            final long parsed = parseDataSize(compression.getMinResponseSize());
+            final int minBytesToForceChunkedAndEncoding = parsed > Integer.MAX_VALUE ? Integer.MAX_VALUE
+                                                                                     : (int) parsed;
             server.decorator(contentEncodingDecorator(compression.getMimeTypes(),
                                                       compression.getExcludedUserAgents(),
-                                                      parseDataSize(compression.getMinResponseSize())));
+                                                      minBytesToForceChunkedAndEncoding));
```
minwoox: We really need to dedup this configuration code. 😅

core/src/main/java/com/linecorp/armeria/server/encoding/EncodingService.java
```diff
@@ -39,66 +37,37 @@
...
-
-    private static final Predicate<HttpHeaders> DEFAULT_ENCODABLE_REQUEST_HEADERS_PREDICATE =
-            headers -> true;
-
-    private static final int DEFAULT_MIN_BYTES_TO_FORCE_CHUNKED_AND_ENCODING = 1024;
+public final class EncodingService extends SimpleDecoratingHttpService {
 
     private final Predicate<MediaType> encodableContentTypePredicate;
-    private final Predicate<HttpHeaders> encodableRequestHeadersPredicate;
+    private final Predicate<RequestHeaders> encodableRequestHeadersPredicate;
```
ikhoon: Don't we need `? super` here?

core/src/main/java/com/linecorp/armeria/server/encoding/EncodingServiceBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+        requireNonNull(encodableRequestHeadersPredicate, "encodableRequestHeadersPredicate");
+        this.encodableRequestHeadersPredicate = encodableRequestHeadersPredicate;
+        return this;
+    }
+
+    /**
+     * Sets the specified minimum length to force chunked encoding. The {@link HttpResponse} is encoded only
+     * when the content is variable, which means the {@link ResponseHeaders} does not have
+     * {@code "Content-Length"} header, or the length of the content exceeds the specified length.
+     * The default is {@value DEFAULT_MIN_BYTES_TO_FORCE_CHUNKED_AND_ENCODING}.
```
ikhoon: `DEFAULT_MIN_BYTES_TO_FORCE_CHUNKED_AND_ENCODING` is a private field.

dropwizard/src/main/java/com/linecorp/armeria/dropwizard/ArmeriaConfigurationUtil.java
```diff
@@ -429,9 +433,18 @@ private static URL getURL(String resourceLocation) throws FileNotFoundException
...
     private static void configureCompression(ServerBuilder serverBuilder, Compression compression) {
         if (compression.isEnabled()) {
+            final int minBytesToForceChunkedAndEncoding;
+            final String minResponseSize = compression.getMinResponseSize();
+            if (minResponseSize == null) {
+                minBytesToForceChunkedAndEncoding = DEFAULT_MIN_BYTES_TO_FORCE_CHUNKED_AND_ENCODING;
+            } else {
+                final long parsed = parseDataSize(minResponseSize);
+                minBytesToForceChunkedAndEncoding = parsed > Integer.MAX_VALUE ? Integer.MAX_VALUE
+                                                                               : (int) parsed;
```
ikhoon: `Ints.saturatedCast(parsed)`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -187,9 +187,11 @@ public WebServer getWebServer(HttpHandler httpHandler) {
 
         final Compression compression = getCompression();
         if (compression != null && compression.getEnabled()) {
+            final long minResponseSize = compression.getMinResponseSize().toBytes();
             sb.decorator(contentEncodingDecorator(compression.getMimeTypes(),
                                                   compression.getExcludedUserAgents(),
-                                                  compression.getMinResponseSize().toBytes()));
+                                                  minResponseSize > Integer.MAX_VALUE ? Integer.MAX_VALUE
+                                                                                      : (int) minResponseSize));
```
ikhoon: Ditto: `Ints.saturatedCast(minResponseSize)`?

spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaReactiveWebServerFactory.java
```diff
@@ -261,9 +263,12 @@ private void configureArmeriaService(ServerBuilder sb, ArmeriaSettings settings)
 
         final ArmeriaSettings.Compression compression = settings.getCompression();
         if (compression != null && compression.isEnabled()) {
+            final long parsed = parseDataSize(compression.getMinResponseSize());
+            final int minBytesToForceChunkedAndEncoding = parsed > Integer.MAX_VALUE ? Integer.MAX_VALUE
+                                                                                     : (int) parsed;
```
ikhoon: Ditto

core/src/main/java/com/linecorp/armeria/server/encoding/EncodingServiceBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+public final class EncodingServiceBuilder {
+
+    // TODO(minwoox) consider this condition to align with the default text preveiwer.
+    private static final Predicate<MediaType> defaultEncodableContentTypePredicate =
+            contentType -> Stream.of(MediaType.ANY_TEXT_TYPE,
+                                     MediaType.APPLICATION_XML_UTF_8,
+                                     MediaType.JAVASCRIPT_UTF_8,
+                                     MediaType.JSON_UTF_8)
+                                 .anyMatch(contentType::is);
```
trustin: Could create a `MediaTypeSet` and reuse it, instead of creating a new stream list again and again.

core/src/main/java/com/linecorp/armeria/server/encoding/EncodingServiceBuilder.java
trustin: It's OK to be a private field. `@value` will inline the value.

minwoox: Yes it does not have to be. 😉

core/src/main/java/com/linecorp/armeria/server/encoding/EncodingService.java
minwoox: Added. 😉

core/src/main/java/com/linecorp/armeria/server/encoding/EncodingServiceBuilder.java
minwoox: Let me address this once #2444 is merged. 😉

minwoox: Oops realized that the `MediaTypeSet` cannot have a media type which has wildcard. So just use Set and iterate on it.

trustin: Oh no! OK.

core/src/main/java/com/linecorp/armeria/server/encoding/EncodingServiceBuilder.java
ikhoon: Ah, I didn't know that, I've learned a lot. Thanks!

trustin:
Thanks, @minwoox

PR #2435, opened 2020-01-31
Add HttpStatus.isSuccess() and other boolean getters

mercanil:
This changeset added shortcut methods to HttpStatus class

PR #2459, opened 2020-02-06
Clean up `HttpStatus.is*()` and their tests

trustin:
This changeset is a follow-up of #2435


trustin:
@mercanil PTAL

core/src/main/java/com/linecorp/armeria/common/HttpStatus.java
```diff
@@ -496,35 +496,35 @@ public boolean isContentAlwaysEmpty() {
      * Returns whether the {@link HttpStatus} is an information, with a status code of 1XX.
      */
     public boolean isInformational() {
-        return codeClass().equals(HttpStatusClass.INFORMATIONAL);
+        return codeClass == HttpStatusClass.INFORMATIONAL;
```
trustin: Realized that we already pre-determined the class of all status instances, so I optimized it like this.

mercanil:
Thanks!

PR #2460, opened 2020-02-06
Remove Service.decorate(Class<R>)

minwoox:
Motivation:
`Service..decorate(Class<R>)` will not be used anymore after #2456 is merged.
Also, I think it's somewhat disparate from our APIs.

Modification:
- Remove `Service.decorate(Class<R>)`

Result:
- Cleaner API.


minwoox:
This PR is made twice. I think it's a bug in Github. 😉

PR #2461, opened 2020-02-06
Remove Service.decorate(Class<R>)

minwoox:
Motivation:
`Service.decorate(Class<R>)` will not be used anymore after #2456 is merged.
Also, I think it's somewhat disparate from our APIs because we use `newDecorator(...)` or builders for decorating.

Modification:
- Remove `Service.decorate(Class<R>)`

Result:
- Cleaner API.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/Service.java
+++ b/core/src/main/java/com/linecorp/armeria/server/Service.java
@@ -18,8 +18,6 @@ package com.linecorp.armeria.server;
 
 import static java.util.Objects.requireNonNull;
 
-import java.lang.reflect.Constructor;
-
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


minwoox:
Send this PR to listen to the opinions.
I can just make this method deprecated or just close this PR.

ikhoon:
FYI: Related documentation on this feature - https://github.com/line/armeria/blob/bd28eb2499c9149613098de17bc1479919dc3fd7/site/src/sphinx/server-decorator.rst#L83

trustin:
+1 for deprecation and then removal

minwoox:
> +1 for deprecation and then removal

Then, just let me close this and deprecate in #2456 😄

PR #2464, opened 2020-02-06
Fix a bug where making IPv6 DnsQuestion when it's not supported

minwoox:
Motivation:
If there's no network interfaces that support IPv6, we must not send a DNS query for that.

Modification:
- Check the network interfaces in `DnsAddressEndpointGourp` not to make IPv6 DNS query.

Result:
- `DnsAddressEndpointGourp` does not send DNS queries for AAAA when IPv6 is not available.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.armeria.client.endpoint.dns;
 
+import static com.linecorp.armeria.internal.client.DnsUtil.anyInterfaceSupportsIpV6;
 import static com.linecorp.armeria.internal.client.DnsUtil.extractAddressBytes;
 
 import java.util.List;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/dns/DnsAddressEndpointGroup.java
```diff
@@ -105,7 +106,7 @@ public static DnsAddressEndpointGroupBuilder builder(String hostname) {
             String hostname, @Nullable ResolvedAddressTypes resolvedAddressTypes) {
 
         if (resolvedAddressTypes == null) {
-            if (NetUtil.isIpV4StackPreferred()) {
+            if (NetUtil.isIpV4StackPreferred() || !anyInterfaceSupportsIpV6()) {
```
anuraaga: I think this code has been around for a long time - does it mean it's not a regression? It seems like it used to work though, but possibly an interaction with something else rather than a regression. Just want to check.

minwoox: Yeah, I might make this too impromptu. Let me dig this more tomorrow. 😅

trustin: @anuraaga, it now skips link-local addresses as well. Shall we merge, release and see if the problem goes away?

anuraaga: Yeah let's try

minwoox:
I will be trying to figure out why it does not happen in 0.95.

A notable change is that, in 0.95 the DNS resolver does not cache at all but `RefreshingResolver` cache the result and reuse it every time.
Also, I'm suspecting that the order of sending DNS query is reversed at a certain point.

PR #371, opened 2017-01-16
Fix graceful shutdown (#363)

trustin:
Motivation:

Graceful shutdown has been broken since the streaming rewrite. It only
worked for HTTP/1 traffic.

Modifications:

- Rename GracefulShutdownHandler to GracefulShutdownSupport
  - Make it an abstract class and provide two implementations; the
    default one and the disabled one
  - GracefulShutdownSupport is not a ChannelHandler anymore but just a
    utility class
- HttpServerHandler now requires GracefulShutdownSupport to count the
  number of pending responses.
- Add more tests and improve the existing ones in
  GracefulShutdownIntegrationTest

Result:

- Graceful shutdown should work as expected.
- Fixes #363

PR #2466, opened 2020-02-06
Fix a bug where `pendingResponses` meter is not updated

trustin:
... when graceful shutdown is disabled.

Motivation:

When graceful shutdown is disabled, `pendingResponses` meter stays at
`0`, because `DisabledGracefulShutdownSupport.pendingResponses()` always
returns `0`.

Modifications:

- Use `LongAdder` to keep track of the number of pending responses.
- Minor optimizations:
  - Do not add unnecessary callbacks when handling a request.
  - Remove the `boolean` flags at the cost of losing 1-nanosecond
    precision, which is OK because we check if quiet period is over
    every 100 ms.

Result:

- `pendingResponses` meter yields a correct value even when graceful
  shutdown is disabled.
- Potentially less contention under load, thanks to `LongAdder`.
- Tad bit less garbage due to less number of callbacks.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/GracefulShutdownSupport.java
+++ b/core/src/main/java/com/linecorp/armeria/server/GracefulShutdownSupport.java
@@ -19,7 +19,7 @@ package com.linecorp.armeria.server;
 import java.time.Duration;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.LongAdder;
 
 import com.google.common.base.Ticker;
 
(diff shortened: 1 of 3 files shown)


trustin:
See #371 for nostalgia :smile:

core/src/main/java/com/linecorp/armeria/server/GracefulShutdownSupport.java
```diff
@@ -100,65 +97,54 @@ boolean completedQuietPeriod() {
...
+                durationNanos = currentTimeNanos - shutdownStartTimeNanos;
             }
 
-            return duration >= quietPeriodNanos;
+            return durationNanos >= quietPeriodNanos;
+        }
+
+        private long readTicker() {
+            // '| 1' makes sure this method never returns 0.
+            return ticker.read() | 1;
```
ikhoon: Nice idea. 👍👍👍

PR #2471, opened 2020-02-07
Preserve service binding order when building routers

ikhoon:
Motivation:
Annotated services are lazily bond to `VirtualHost` to apply `AnnotatedServiceExtensions` to them. #2316
It causes a change in the order of original service binding defined by the user.
This is critical if a user uses a global fallback service.
In the following example, the user added the fallback service in the last.
However, the fallback will be added first and annotated services will be added later.

```java
sb.annotatedService("/api/v0/", new Object() {
    @Get("regex:/projects/(?<projectName>[^/]+)")
     ...
});
sb.annotatedService("/api/v0/", new Object() {
    @Get("/users/me")
    ...
});
sb.serviceUnder("/", (ctx, req) -> HttpResponse.of("fallback"));
```

Modifications:
* Make ServiceConfigBuilder implement ServiceConfigSetters
* Remove unused getters in ServiceConfigBuilder
* Merge `addAnnotatedServiceBindingBuilder` and  `addServiceConfigBuilder` into `addServiceConfigSetters`
* Remove `annotatedServiceBindingBuilders` and delegate it to `virtualHostTemplate`

Result:
Fix regression in service bindings.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/ServiceConfigBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/ServiceConfigBuilder.java
@@ -18,13 +18,16 @@ package com.linecorp.armeria.server;
 
 import static java.util.Objects.requireNonNull;
 
+import java.time.Duration;
+import java.util.function.Function;
+
 import javax.annotation.Nullable;
 
 import com.google.common.base.MoreObjects;
 
 import com.linecorp.armeria.server.logging.AccessLogWriter;
 
-final class ServiceConfigBuilder {
+final class ServiceConfigBuilder implements ServiceConfigSetters {
 
     private final Route route;
     private final HttpService service;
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/server/VirtualHostBuilder.java
```diff
@@ -917,17 +911,24 @@ VirtualHost build(VirtualHostBuilder template) {
...
+                    if (cfgSetters instanceof VirtualHostAnnotatedServiceBindingBuilder) {
+                        return ((VirtualHostAnnotatedServiceBindingBuilder) cfgSetters)
+                                .buildServiceConfigBuilder(extensions).stream();
+                    } else if (cfgSetters instanceof AnnotatedServiceBindingBuilder) {
+                        return ((AnnotatedServiceBindingBuilder) cfgSetters)
+                                .buildServiceConfigBuilder(extensions).stream();
+                    } else if (cfgSetters instanceof ServiceConfigBuilder) {
+                        return Stream.of((ServiceConfigBuilder) cfgSetters);
+                    } else {
+                        throw new Error(); // Should not reach here.
```
mercanil: Error is too generic in here may be we should throw more specific Exception in here

ikhoon: This is an internal API so users must not encounter this error. But, let me improve the error message.

core/src/main/java/com/linecorp/armeria/server/ServiceConfigBuilder.java
```diff
@@ -44,47 +47,47 @@
         this.service = requireNonNull(service, "service");
     }
 
-    @Nullable
-    Long requestTimeoutMillis() {
-        return requestTimeoutMillis;
+    @Override
+    public ServiceConfigSetters requestTimeout(Duration requestTimeout) {
```
trustin: You sometimes return `ServiceConfigSetters` and sometimes `ServiceConfigBuilder` in this class. It doesn't hurt to return `ServiceConfigSetters` because it's package-private, but perhaps it's better returning `ServiceConfigBuilder` for consistency.

ikhoon: Oops.. Thanks for pointing!

core/src/test/java/com/linecorp/armeria/server/ServiceRoutingTest.java
```diff
@@ -0,0 +1,55 @@
...
+import com.linecorp.armeria.server.annotation.Get;
+import com.linecorp.armeria.server.annotation.Param;
+import com.linecorp.armeria.testing.junit.server.ServerExtension;
+
+class ServiceRoutingTest {
+    @RegisterExtension
+    static ServerExtension server = new ServerExtension() {
+        @Override
+        protected void configure(ServerBuilder sb) throws Exception {
+            sb.annotatedService("/api/v0/", new Object() {
```
minwoox: nit: let's do this after adding the below annotated service so that two tri routers are made which is the regression case.

heowc:
Nice Job!!
Sorry for making me confused. 😅

ikhoon:
> Sorry for making me confused. 😅

Don't worry. We are always welcome your contribution! ♥️

PR #2483, opened 2020-02-13
Allow `ArmeriaRetrofitBuilder` to configure `ClientOption`

### LINKED ISSUE #2481: Add more options to `ArmeriaRetrofitBuilder`
To build Armeria Retrofit with a specified `ClientFactory`, a user should create a `WebClient` which is a little verbose:
```java
ClientFactory factory = ...;
// Before 0.98.0
new ArmeriaRetrofitBuilder(clientFactory);
// After 0.98.0
ArmeriaRetrofit.builder(WebClient.builder(...).clientFactory(factory)...);
```
Perhaps, we can provide more setters to build Armeria Retrofit more conveniently.

ikhoon:
Motivation:
If a user wants to add `ClientFactory`, he/she should have to create `WebClient`.
It is convenient to configure `ClientOption` when building `ArmeriaRetrofit`.

Modifications:
* Make `AbstractClientOptionsBuilder` public.
* Allow creating `ClientOption` with the specified options.
* Lazily build `baseWebClient` and `nonBaseClientFactory` with Retrofit client options.

Result:
You can now directly configure `ClientOption` when building `ArmeriaRetrofit`.
Fixes #2481

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
@@ -13,7 +13,7 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-package com.linecorp.armeria.internal.client;
+package com.linecorp.armeria.client;
 
 import static java.util.Objects.requireNonNull;
 
@@ -27,17 +27,6 @@ import java.util.function.Supplier;
 
 import javax.annotation.Nullable;
 
-import com.linecorp.armeria.client.ClientDecoration;
-import com.linecorp.armeria.client.ClientDecorationBuilder;
-import com.linecorp.armeria.client.ClientFactory;
-import com.linecorp.armeria.client.ClientOption;
-import com.linecorp.armeria.client.ClientOptionValue;
-import com.linecorp.armeria.client.ClientOptions;
-import com.linecorp.armeria.client.DecoratingHttpClientFunction;
-import com.linecorp.armeria.client.DecoratingRpcClientFunction;
-import com.linecorp.armeria.client.Endpoint;
-import com.linecorp.armeria.client.HttpClient;
-import com.linecorp.armeria.client.RpcClient;
 import com.linecorp.armeria.client.endpoint.EndpointGroup;
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpHeaders;
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
@@ -30,17 +30,23 @@ import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpHeadersBuilder;
 import com.linecorp.armeria.common.RequestId;
+import com.linecorp.armeria.common.util.UnstableApi;
 
-class AbstractClientOptionsBuilder {
+/**
+ * A skeletal builder implementation for {@link ClientOptions}.
+ * Note: This class is meant for internal use only.
+ */
+@UnstableApi
+public class AbstractClientOptionsBuilder {
 
     private final Map<ClientOption<?>, ClientOptionValue<?>> options = new LinkedHashMap<>();
     private final ClientDecorationBuilder decoration = ClientDecoration.builder();
     private final HttpHeadersBuilder httpHeaders = HttpHeaders.builder();
 
     /**
-     * Creates a new instance with the default options.
+     * Creates a new instance.
      */
-    AbstractClientOptionsBuilder() {}
+    protected AbstractClientOptionsBuilder() {}
 
     /**
      * Creates a new instance with the specified base options.
@@ -120,9 +126,9 @@ class AbstractClientOptionsBuilder {
     /**
      * Sets the timeout of a socket write attempt.
      *
-     * @deprecated Use {@link #writeTimeout(Duration)}.
-     *
      * @param writeTimeout the timeout. {@code 0} disables the timeout.
+     *
+     * @deprecated Use {@link #writeTimeout(Duration)}.
      */
     @Deprecated
     public AbstractClientOptionsBuilder defaultWriteTimeout(Duration writeTimeout) {
@@ -132,9 +138,9 @@ class AbstractClientOptionsBuilder {
     /**
      * Sets the timeout of a socket write attempt in milliseconds.
      *
-     * @deprecated Use {@link #writeTimeoutMillis(long)}.
-     *
      * @param writeTimeoutMillis the timeout in milliseconds. {@code 0} disables the timeout.
+     *
+     * @deprecated Use {@link #writeTimeoutMillis(long)}.
      */
     @Deprecated
     public AbstractClientOptionsBuilder defaultWriteTimeoutMillis(long writeTimeoutMillis) {
@@ -162,9 +168,9 @@ class AbstractClientOptionsBuilder {
     /**
      * Sets the timeout of a response.
      *
-     * @deprecated Use {@link #responseTimeout(Duration)}.
-     *
      * @param responseTimeout the timeout. {@code 0} disables the timeout.
+     *
+     * @deprecated Use {@link #responseTimeout(Duration)}.
      */
     @Deprecated
     public AbstractClientOptionsBuilder defaultResponseTimeout(Duration responseTimeout) {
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
```diff
@@ -30,17 +30,23 @@
 import com.linecorp.armeria.common.HttpHeaders;
 import com.linecorp.armeria.common.HttpHeadersBuilder;
 import com.linecorp.armeria.common.RequestId;
+import com.linecorp.armeria.common.util.UnstableApi;
 
-class AbstractClientOptionsBuilder {
+/**
+ * A skeletal builder implementation for {@link ClientOptions}.
+ * Note: This class is meant for internal use only.
```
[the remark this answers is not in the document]
ikhoon: I tried to move to the internal package first but `AbstractClientOptionsBuilder` access package-private method in `ClientDecoration`
https://github.com/line/armeria/blob/6178d35f4b26ae8fce31ede5d1b848aa3e2a80a6/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java#L101-L102

I thought making this public with `@Unstable` annotation is better than change the method to `public`.
https://github.com/line/armeria/blob/8a67c5823ab3cfd1f4a27683409c638091093d3f/core/src/main/java/com/linecorp/armeria/client/ClientDecoration.java#L78-L84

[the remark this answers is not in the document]
trustin: As a follow-up of this pull request, how about adding `ClientDecorationBuilder`, so that `AbstractClientOptionsBuilder` can mutate/build a `ClientDecoration` without accessing package-private methods?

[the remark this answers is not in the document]
trustin: Then, we could move `AbstractrClientOptionsBuilder` to `internal`.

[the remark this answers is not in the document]
minwoox: > I tried to move to the internal package first but AbstractClientOptionsBuilder access package-private method in ClientDecoration

Ah, I didn't notice that. 😅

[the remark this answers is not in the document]
ikhoon: > how about adding `ClientDecorationBuilder`, so that AbstractClientOptionsBuilder can mutate/build a ClientDecoration without accessing package-private methods?

IIUC, you were meant to add `addClientDecoration(ClientDecoration)` to [`ClientDecorationBuilder`](https://github.com/line/armeria/blob/8a67c5823ab3cfd1f4a27683409c638091093d3f/core/src/main/java/com/linecorp/armeria/client/ClientDecorationBuilder.java#L28)?

Sorry, I don't thoroughly understand your direction. 😅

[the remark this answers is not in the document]
trustin: Oh, I completely forgot we have `ClientDecorationBuilder` already. :rofl: 
If it's enough with adding just `public ClientDecorationBuilder.add(ClientDecoration)`, then how about handling it in this PR?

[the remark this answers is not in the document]
ikhoon: Sure, will handle it in this PR. :-)

core/src/main/java/com/linecorp/armeria/internal/client/AbstractClientOptionsBuilder.java
```diff
@@ -380,7 +386,7 @@ protected final ClientOptions buildOptions(boolean includeDefault) {
         if (includeDefault) {
             return ClientOptions.of(optValArray);
         } else {
-            return ClientOptions.of(ClientOptions.EMPTY, optValArray);
+            return ClientOptions.of(ClientOptions.empty(), optValArray);
```
trustin: Hmm, I'm not sure if it's a good idea to expose `empty()` in the public API. How about this:

- Remove `EMPTY` and `empty()`
- Change the signiture of `buildOptions()` to:
  - `buildOptions(@Nullable ClientOptions baseOptions)` (We pass `WebClient.options()` in this PR, or `null` for other cases.)
- Use `ClientOptions.of(baseOptions, optValArray)` (if `null`, use `ClientOptions.of(optValArray)`.

ikhoon: > I'm not sure if it's a good idea to expose empty()

That was my considering point. `buildOptions(@Nullable ClientOptions baseOptions)` sounds good.
Thanks for reviewing 👍

trustin:
I think we can't extend a class in the `internal` package. Perhaps the only way to fix this is moving it back to the public API? I guess we don't really need to annotate it with `@UnstableApi` though, because a user might want to extend it to build some builder?

ikhoon:
Oops.. OK to move back to public without `@UnstableApi` 😀

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -411,4 +412,26 @@ public ConnectionPoolListener connectionPoolListener() {
...
     }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    protected <T extends AbstractOptionValue<?, ?>> T mergeValue(T oldValue, T newValue) {
+        if (oldValue.option() == ClientFactoryOption.CHANNEL_OPTIONS) {
+            final ClientFactoryOptionValue<Map<ChannelOption<?>, Object>> castOldValue =
+                    (ClientFactoryOptionValue<Map<ChannelOption<?>, Object>>) oldValue;
+            final ClientFactoryOptionValue<Map<ChannelOption<?>, Object>> castNewValue =
+                    (ClientFactoryOptionValue<Map<ChannelOption<?>, Object>>) newValue;
```
trustin: Could use `newValue.value()` and `oldValue.value()` instead of calling `.value()` many times.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -411,4 +412,26 @@ public ConnectionPoolListener connectionPoolListener() {
...
+            final ImmutableMap.Builder<ChannelOption<?>, Object> builder =
+                    ImmutableMap.builderWithExpectedSize(castNewValue.value().size());
```
trustin: - Probably better using `oldValue`'s size + `newValue`'s size? Not always optimal, but expansion is more expensive.
- Could just return the `oldValue` if `newValue` is empty (or `newValue` is `oldValue` is empty.)

core/src/main/java/com/linecorp/armeria/client/ClientOptions.java
```diff
@@ -208,6 +209,30 @@ private ClientOptions(ClientOptions baseOptions, ClientOptions additionalOptions
         super(baseOptions, additionalOptions);
     }
 
+    @Override
+    protected <T extends AbstractOptionValue<?, ?>> T mergeValue(T oldValue, T newValue) {
+        if (oldValue.option() == DECORATION) {
+            final ClientDecoration oldDecoration = (ClientDecoration) oldValue.value();
+            final ClientDecoration newDecoration = (ClientDecoration) newValue.value();
```
trustin: Could return oldDecoration or newDecoration if newDecoration or oldDecoration is empty respectively.

core/src/main/java/com/linecorp/armeria/client/ClientOptions.java
```diff
@@ -208,6 +209,30 @@ private ClientOptions(ClientOptions baseOptions, ClientOptions additionalOptions
...
+                                                            .add(oldDecoration)
+                                                            .add(newDecoration)
+                                                            .build();
+            @SuppressWarnings("unchecked")
+            final T cast = (T) DECORATION.newValue(merged);
+            return cast;
+        } else if (oldValue.option() == HTTP_HEADERS) {
+            final HttpHeaders oldHeaders = (HttpHeaders) oldValue.value();
+            final HttpHeaders newHeaders = (HttpHeaders) newValue.value();
+            final HttpHeaders merged = oldHeaders.toBuilder().setObject(newHeaders).build();
```
trustin: Ditto - Could optimize when either is empty.

trustin:
If you decided to add `abstract mergeValue`, how about adding `abstract filterValue` as well and removing the constructor parameter function?

ikhoon:
> how about adding abstract `filterValue` as well and removing the constructor parameter function?

I was considering it, but not 100% sure. Thanks for clarifying. 😀

core/src/main/java/com/linecorp/armeria/client/ClientOptions.java
```diff
@@ -192,19 +177,70 @@ private static HttpHeaders filterHttpHeaders(HttpHeaders headers) {
...
+        }
+        return optionValue;
+    }
+
+    @Override
+    protected <T extends AbstractOptionValue<?, ?>> T mergeValue(T oldValue, T newValue) {
+        if (oldValue.option() == DECORATION) {
+            final ClientDecoration oldDecoration = (ClientDecoration) oldValue.value();
+            final ClientDecoration newDecoration = (ClientDecoration) newValue.value();
+            if (oldDecoration.decorators().isEmpty() && oldDecoration.rpcDecorators().isEmpty()) {
```
trustin: How about adding `ClientDecoration.isEmpty()` and simplifying here?

ikhoon: Oh, that sounds nice.

ikhoon:
Thanks for reviewing 😀

trustin:
Nice work, @ikhoon :tada:

PR #2485, opened 2020-02-13
Fix `InvalidKeySpecException` when using `PKCS#5` keys

trustin:
Motivation:

Armeria 0.98.x raises an `InvalidKeySpecException` when a user specifies
a `PKCS#5` private key via `ServerBuilder.tls()` or `{Server,ClientFactory}Builder.tlsCustomizer()`.

Modifications:

- Make sure `BouncyCastleKeyFactoryProvider` hack is applied for all TLS
  configuration path.
- Miscellaneous:
  - Create client-side `SslContext` upfront, so that TLS configuration
    errors are detected early, even without making a connection attempt.
- Re-generated test certificate and private key pair.

Result:

- A user can launch an Armeria server with a `PKCS#5` private key again,
  which was previously possible up to 0.97.x.


trustin:
It'd be awesome if we can add support for PKCS12 keystore, too, before 1.0 as well. :thinking:

trustin:
Will look into the build failure later.

trustin:
Thanks for reviewing.

PR #2492, opened 2020-02-14
Do not export `null` values in `RequestContextExporter`

trustin:
Motivation:

`RequestContextExporter` exports a property whose value is `null`, but
this can lead to a `NullPointerException` in the user code, because a
user often assumes a `Map` contains non-`null` values only.

Modifications:

- Do not put a `null`-value entry when exporting the properties in
  `RequestContextExporter`.

Result:

- A user can safely assume that the map returned by
  `RequestContextExporter` always have non-`null` entries only.


minwoox:
Should remove [this](https://github.com/line/armeria/blob/master/logback/src/test/java/com/linecorp/armeria/common/logback/RequestContextExportingAppenderTest.java#L266) 😉

trustin:
Thanks for reviewing.

Issue #2321, opened 2019-12-13
Client-side proxy support


trustin:
Related: #2320 

We could add support for various proxy protocols to our clients:

- HTTP CONNECT method
- SOCKS
- HAProxy
- Different from the above 3, but usual HTTP proxies

trustin:
@jrhee17 What about this issue? You asked me about a client-side issue that would help you understand client-side internals and thus might be it.

kojilin:
Curious about this issue, so we just need to add proxy option in ClientFactoyOption and add ProxyHandler here https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java#L102 ?

jrhee17:
btw feel free to work on this issue if you need to use it 😄 I think I will need at least a month to start this issue if I tackle it

trustin:
@kojilin I guess so, although I didn't try it by myself. :smile:

jrhee17:
@kojilin did you want to work on this? 👀  I think I can start working on this if not.

kojilin:
@jrhee17 oh, no problem :)

kojilin:
To avoid confusing, yes, you can work on this issue.

jrhee17:
Thanks! 🙇👍😄

trustin:
Created a follow-up issue: https://github.com/line/armeria/issues/2589

PR #2496, opened 2020-02-16
Client-side proxy support

jrhee17:
Fixes: #2321 

Client-side proxy support for SOCKS4, SOCKS5, CONNECT Protocol

**Motivation**

Add support for client-side proxy.

**Modification**

- Add a `ProxyConfig` class which serves as a base class for proxy configurations.
- Add a `ClientFactory.proxyConfig` method which accepts a `ProxyConfig` object containing proxy related configuration.
- Add a `DisabledProxyConfig` which represents the default -- don't apply client-side proxy

**Misc**
- I will be adding relative documentation after code review is complete as the api may change
- I have decided to add regular HTTP proxy, HaProxy in a separate issue to avoid having a PR that is too big.,
- I am slightly concerned that server side also has a PROXY `SessionProtocol` which may mislead users to think server side will support all of the implemented client-side proxies. I would like some opinions about this.


trustin:
Netty doesn't have client-side support for PROXY protocol. How about starting from SOCKS?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -398,6 +403,13 @@ public MeterRegistry meterRegistry() {
         return get(ClientFactoryOption.METER_REGISTRY);
     }
 
+    /**
+     * TODO: add javadoc comment.
+     */
+    public Optional<? extends ProxyHandler> getProxyHandler() {
+        return get(ClientFactoryOption.PROXY_HANDLER);
```
kojilin: Just curious if it's good to expose Netty proxy handler directly to the user is good or not.
Especially if we start to support Forward Proxy, we may need to add one more netty handler in the middle. Then the proxy handler is not suitable for that case.

JFYI my quick investigation https://github.com/kojilin/armeria/commit/6cbd1e4170bdb76aa848649bcc3598a70ddcb42a#diff-fd04617458eb8556240b7e1a48959f8bR349

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -118,6 +124,14 @@ protected void initChannel(Channel ch) throws Exception {
                                                       .get(ChannelOption.CONNECT_TIMEOUT_MILLIS);
     }
 
+    private static void addSocksProxyHandlerIfPossible(
+            ChannelPipeline pipeline, Optional<? extends ProxyHandler> proxyHandler) {
+        if (proxyHandler.isPresent() && (proxyHandler.get() instanceof Socks4ProxyHandler ||
```
kojilin: Maybe 1 more HttpProxyHandler for CONNECT?

jrhee17: I think `HttpProxyHandler` should be added after ssl is applied (at least that's what the netty examples were doing 🤔 )

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
jrhee17: Good point.   Was also thinking about this yesterday for a while.
Each proxy will require different parameters (connectionTimeout, user, password, etc.), so the choice was either to expose netty apis, or to create a wrapper.

I'm still unsure about this though.   (I haven't really thought about Forward Proxy..)
I'll take a look at your draft later 🙏 🙇 👍

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
kojilin: You mean between client and proxy using https? If we use HttpProxyHandler directly, we may support nomral HTTP CONNECT one I think.

--
Also can you tell me the link, just curious about which example.

jrhee17: https://github.com/netty/netty/blob/5c458c9a98d4d3d0345e58495e017175156d624f/handler-proxy/src/test/java/io/netty/handler/proxy/ProxyHandlerTest.java#L360

Didn't really take a thorough look though..

kojilin:
About test, I didn't find suitable http proxy java implementation, maybe we can use testcontainers with squid?(But there is no precedent in armeria)

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
kojilin: So it looks like 2 cases.
If we want to provide https proxy, I think as you said we need to put before HttpProxyHandler. 
But there is another https between client and the real server I think.

jrhee17: Ah,   I see.  You're right in that there are 2 cases.
This may complicate the api a bit (we may need to add another ssl handler if we want to support `Http[s]ProxyHandler` 🤔 )

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -257,4 +257,11 @@ public ConnectionPoolListener connectionPoolListener() {
     public MeterRegistry meterRegistry() {
         return get(ClientFactoryOption.METER_REGISTRY);
     }
+
+    /**
+     * The {@link Proxy} which contains the proxy configuration.
+     */
+    public Proxy getProxy() {
```
trustin: `proxy()` for consistency

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -118,6 +125,40 @@ protected void initChannel(Channel ch) throws Exception {
                                                       .get(ChannelOption.CONNECT_TIMEOUT_MILLIS);
     }
 
+    private void applyProxy(Channel ch, Proxy proxy, SslContext sslCtx) {
```
trustin: `configureProxy` ?

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -118,6 +125,40 @@ protected void initChannel(Channel ch) throws Exception {
...
+                final String username = proxy.userName();
+                final String password = ((ConnectProxy) proxy).password();
+                if (username == null || password == null) {
+                    proxyHandler = new HttpProxyHandler(proxy.proxyAddress());
+                } else {
+                    proxyHandler = new HttpProxyHandler(proxy.proxyAddress(), username, password);
+                }
+                break;
+            default:
+                logger.warn("Unknown proxy type not applied: {}.", proxy.proxyType());
```
trustin: `"{} Ignoring unknown proxy type: {}", ch, proxy.proxyType()` ?

core/src/main/java/com/linecorp/armeria/client/Proxy.java
```diff
@@ -0,0 +1,209 @@
...
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.client.ProxyBuilder.ConnectProxyBuilder;
+import com.linecorp.armeria.client.ProxyBuilder.Socks4ProxyBuilder;
+import com.linecorp.armeria.client.ProxyBuilder.Socks5ProxyBuilder;
+
+/**
+ * Contains base configuration for proxy related settings used in {@link ClientFactory}.
+ */
+public class Proxy {
```
trustin: - `final`
- Probably `ProxyConfig` is better?
- Please implement `toString()` for this class and its subclasses.

core/src/main/java/com/linecorp/armeria/client/Proxy.java
```diff
@@ -0,0 +1,209 @@
...
+
+    private static final long USE_DEFAULT_TIMEOUT_MILLIS = -1;
+    static final Proxy DEFAULT =
+            new Proxy(ProxyType.NONE, new InetSocketAddress(0), USE_DEFAULT_TIMEOUT_MILLIS);
```
trustin: Could we also add a static singleton factory method `ofDefault()` and `none()`?

core/src/main/java/com/linecorp/armeria/client/Proxy.java
```diff
@@ -0,0 +1,209 @@
...
+     * Creates a {@code Proxy} configuration for CONNECT protocol.
+     * @param proxyAddress The proxy address.
+     * @param connectTimeoutMillis The connection timeout for connecting to the proxy server.
+     */
+    public static ConnectProxyBuilder connect(InetSocketAddress proxyAddress, long connectTimeoutMillis) {
+        checkArgument(connectTimeoutMillis >= 0);
+        return new ConnectProxyBuilder(requireNonNull(proxyAddress), connectTimeoutMillis);
+    }
+
+    ProxyType proxyType() {
```
trustin: `type()`

core/src/main/java/com/linecorp/armeria/client/Proxy.java
```diff
@@ -0,0 +1,209 @@
...
+    @Nullable
+    public String userName() {
+        return userName;
+    }
+
+    void setUserName(@Nullable String userName) {
+        this.userName = userName;
+    }
+
+    enum ProxyType {
```
trustin: Please move to the top level.

core/src/main/java/com/linecorp/armeria/client/Proxy.java
```diff
@@ -0,0 +1,209 @@
...
+        NONE,
+        SOCKS4,
+        SOCKS5,
+        CONNECT,
+    }
+
+    /**
+     * Contains SOCKS4 proxy configuration used in {@link ClientFactory}.
+     */
+    public static class Socks4Proxy extends Proxy {
```
trustin: These subtypes could be moved to the top level.

core/src/main/java/com/linecorp/armeria/client/Proxy.java
```diff
@@ -0,0 +1,209 @@
...
+    private final long connectTimeoutMillis;
+
+    @Nullable
+    private String userName;
+
+    Proxy(ProxyType proxyType, InetSocketAddress proxyAddress, long connectTimeoutMillis) {
+        this.proxyType = proxyType;
+        this.proxyAddress = proxyAddress;
+        this.connectTimeoutMillis = connectTimeoutMillis;
+    }
```
trustin: instance members should appear after the static factory methods.

core/src/main/java/com/linecorp/armeria/client/ProxyBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+        this.connectTimeoutMillis = connectTimeoutMillis;
+    }
+
+    @Nullable
+    protected String getUserName() {
+        return userName;
+    }
+
+    protected void setUserName(@Nullable String userName) {
+        this.userName = userName;
```
trustin: Global comment: userName -> username

core/src/main/java/com/linecorp/armeria/client/ProxyBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+
+import java.net.InetSocketAddress;
+
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.client.Proxy.ConnectProxy;
+import com.linecorp.armeria.client.Proxy.Socks4Proxy;
+import com.linecorp.armeria.client.Proxy.Socks5Proxy;
+
+abstract class ProxyBuilder {
```
trustin: `AbstractProxyBuilder`?

core/src/main/java/com/linecorp/armeria/client/ProxyBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+     */
+    public static final class ConnectProxyBuilder extends ProxyBuilder {
+        ConnectProxyBuilder(InetSocketAddress proxyAddress, long connectTimeoutMillis) {
+            super(proxyAddress, connectTimeoutMillis);
+        }
+
+        /**
+         * Set the proxy userName and password.
+         */
+        public ConnectProxyBuilder auth(String userName, String password) {
```
trustin: How about just splitting into `username()` and `password()`?

core/src/main/java/com/linecorp/armeria/client/Proxy.java
```diff
@@ -0,0 +1,209 @@
...
+    public static ConnectProxyBuilder connect(InetSocketAddress proxyAddress) {
+        return new ConnectProxyBuilder(requireNonNull(proxyAddress), USE_DEFAULT_TIMEOUT_MILLIS);
+    }
+
+    /**
+     * Creates a {@code Proxy} configuration for CONNECT protocol.
+     * @param proxyAddress The proxy address.
+     * @param connectTimeoutMillis The connection timeout for connecting to the proxy server.
+     */
+    public static ConnectProxyBuilder connect(InetSocketAddress proxyAddress, long connectTimeoutMillis) {
```
trustin: if connect timeout is optional, we can remove the factory methods with `connectTimeoutMillis` and add `connectTimeout(Duration)` and `connectTimeoutMillis(long)` to the builder class.

core/src/main/java/com/linecorp/armeria/client/Proxy.java
```diff
@@ -0,0 +1,209 @@
...
+            return password;
+        }
+
+        void setPassword(@Nullable String password) {
+            this.password = password;
+        }
+
+        void setUseSsl(boolean useSsl) {
+            this.useSsl = useSsl;
+        }
```
trustin: Make all non-builder classes purely immutable.

testing-internal/src/main/java/com/linecorp/armeria/internal/testing/DynamicBehaviorHandler.java
```diff
@@ -0,0 +1,79 @@
...
+
+import javax.annotation.Nullable;
+
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandler.Sharable;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+
+@Sharable
+public class DynamicBehaviorHandler extends ChannelDuplexHandler {
```
trustin: This seems unused but I guess you have a plan. :wink:

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
jrhee17: > we may need to add one more netty handler in the middle

Agreed.  I have added ssl support to CONNECT to support this claim.

Thanks for the investigation by the way, it actually helped a lot.  I actually couldn't use `mock-server` like you did since I wanted some negative test cases, but I had a fun day looking through the `mock-server` code 👍 
Also, I am getting the feeling support for regular forward proxy will be trickier.. (especially with HTTP2). I am working on a POC, so I'll maybe upload a PR separately if I get it sorted out.

core/src/main/java/com/linecorp/armeria/client/Proxy.java
jrhee17: Couldn't use `final` since other classes are inheriting this class

core/src/main/java/com/linecorp/armeria/client/Proxy.java
jrhee17: Decided to get rid of proxy specific `connectTimeoutMillis ` and use `ClientFactory.connectTimeoutMillis`.

core/src/main/java/com/linecorp/armeria/client/ProxyBuilder.java
jrhee17: Actually, as you may know `HttpProxyHandler` requires both `username`, `password` are either null or non-null.

If we decide to split this, I can add validation logic to the `build` method, but I felt this style of API would communicate this restraint to the user better. Let me know what you think 😉

testing-internal/src/main/java/com/linecorp/armeria/internal/testing/DynamicBehaviorHandler.java
jrhee17: It's used to control proxy server behavior while running integration tests. 🏃 

https://github.com/line/armeria/blob/5c79ac09cb59f0e6e463b32ffe4cd2e2f362010c/core/src/test/java/com/linecorp/armeria/client/proxy/ProxyClientIntegrationTest.java#L94

core/src/main/java/com/linecorp/armeria/client/proxy/AbstractProxyConfigBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+    protected AbstractProxyConfigBuilder(InetSocketAddress proxyAddress) {
+        this.proxyAddress = proxyAddress;
+    }
+
+    protected InetSocketAddress getProxyAddress() {
+        return proxyAddress;
+    }
+
+    @Nullable
+    protected String getUsername() {
```
ikhoon: Could be package-private?

core/src/main/java/com/linecorp/armeria/client/proxy/AbstractProxyConfigBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+        return username;
+    }
+
+    protected void setUsername(@Nullable String username) {
+        this.username = username;
+    }
+
+    @Nullable
+    protected String getPassword() {
```
ikhoon: Ditto

core/src/main/java/com/linecorp/armeria/client/proxy/AbstractProxyConfigBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+    @Nullable
+    private String username;
+    @Nullable
+    private String password;
+
+    protected AbstractProxyConfigBuilder(InetSocketAddress proxyAddress) {
+        this.proxyAddress = proxyAddress;
+    }
+
+    protected InetSocketAddress getProxyAddress() {
```
ikhoon: Ditto

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,98 @@
...
+ * Contains base configuration for proxy related settings used in {@link ClientFactory}.
+ */
+public abstract class ProxyConfig {
+
+    /**
+     * Creates a {@code ProxyConfig} configuration for SOCKS4 protocol.
+     * @param proxyAddress The proxy address.
+     */
+    public static Socks4ProxyConfigBuilder socks4(InetSocketAddress proxyAddress) {
+        return new Socks4ProxyConfigBuilder(requireNonNull(proxyAddress));
```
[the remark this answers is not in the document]
ikhoon: I think Armeria would not use the following API design. You might want to start with `builder()...` to build `ProxyConfig` pattern. 
```java
ProxyConfig.builder()
           .socks4(...)
           .username(...)
           .build()
```


Or How about making factory methods as we did in `Backoff`?
https://github.com/line/armeria/blob/bd28eb2499c9149613098de17bc1479919dc3fd7/core/src/main/java/com/linecorp/armeria/client/retry/Backoff.java#L50-L62

core/src/main/java/com/linecorp/armeria/client/ProxyBuilder.java
trustin: OK, then how about renaming to `credential()`?

core/src/main/java/com/linecorp/armeria/client/proxy/AbstractProxyConfigBuilder.java
trustin: .. and `final`?

core/src/main/java/com/linecorp/armeria/client/proxy/ConnectProxyConfig.java
```diff
@@ -0,0 +1,66 @@
...
+    @Nullable
+    public String password() {
+        return password;
+    }
+
+    /**
+     * Whether ssl is enabled.
+     */
+    public boolean useSsl() {
+        return useSsl;
```
trustin: `useTls`

core/src/main/java/com/linecorp/armeria/client/proxy/ConnectProxyConfigBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+        setPassword(requireNonNull(password));
+        return this;
+    }
+
+    /**
+     * Signifies whether to use ssl to connect to the proxy.
+     * If enabled, the ssl configurations for the {@link ClientFactory} will also be applied to the proxy.
+     */
+    public ConnectProxyConfigBuilder useSsl(boolean useSsl) {
+        this.useSsl = useSsl;
```
trustin: `useTls`

core/src/main/java/com/linecorp/armeria/client/proxy/NoopProxyConfig.java
```diff
@@ -0,0 +1,44 @@
...
+import java.net.InetSocketAddress;
+
+import javax.annotation.Nullable;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * Represents client-side proxy is disabled.
+ */
+public class NoopProxyConfig extends ProxyConfig {
```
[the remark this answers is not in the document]
trustin: - How about `DisabledProxyConfig` ?
- Missing `final`

core/src/main/java/com/linecorp/armeria/client/proxy/NoopProxyConfig.java
```diff
@@ -0,0 +1,44 @@
...
+
+    static final NoopProxyConfig NOOP_PROXY_CONFIG =
+            new NoopProxyConfig(new InetSocketAddress(0), null);
+
+    NoopProxyConfig(InetSocketAddress proxyAddress, @Nullable String username) {
```
trustin: Why do we need `username` here?

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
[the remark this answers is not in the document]
trustin: How about `builderForSocks4`, etc?

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,98 @@
...
+     */
+    public static ConnectProxyConfigBuilder connect(InetSocketAddress proxyAddress) {
+        return new ConnectProxyConfigBuilder(requireNonNull(proxyAddress));
+    }
+
+    /**
+     * Returns a {@code ProxyConfig} which signifies that proxy is disabled.
+     */
+    public static ProxyConfig noop() {
+        return NOOP_PROXY_CONFIG;
```
trustin: `disabled()`

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private final InetSocketAddress proxyAddress;
+
+    @Nullable
+    private final String username;
```
trustin: These fields are non-existent when disabled or when we add a config with different authentication mechanism such as SASL. Could you push them down to the subtypes and make this class an interface?

core/src/main/java/com/linecorp/armeria/client/proxy/Socks5ProxyConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import java.net.InetSocketAddress;
+
+import javax.annotation.Nullable;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * Contains SOCKS5 proxy related configuration.
+ */
+public class Socks5ProxyConfig extends ProxyConfig {
```
trustin: Missing `final`

core/src/main/java/com/linecorp/armeria/client/proxy/ConnectProxyConfig.java
```diff
@@ -0,0 +1,66 @@
...
+import java.net.InetSocketAddress;
+
+import javax.annotation.Nullable;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * Contains CONNECT proxy related configuration.
+ */
+public class ConnectProxyConfig extends ProxyConfig {
```
trustin: Missing `final`

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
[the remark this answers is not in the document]
trustin: Alternatively, we could give up builder API and just add various factory methods:

```java
public static Socks4ProxyConfig socks4(address);
public static Socks4ProxyConfig socks4(address, username);
public static Socks5ProxyConfig socks5(address);
public static Socks5ProxyConfig socks5(address, username, password);
public static ConnectProxyConfig connect(address);
public static ConnectProxyConfig connect(address, HttpHeaders?);
```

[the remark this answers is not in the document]
jrhee17: > start with builder()
It may be hard to express  the idea that only one of the proxy protocols can be selected with this. As @trustin suggested, we would have to have sth like `builderForSocks4` if we want to go this route.

> give up builder API and just add various factory methods
I initially gave up on this idea bc I don't like nullable parameters, but I do feel like this pattern may be slightly overkill.

Decided to go with factory methods to save users from typing the extra `build()` 😉

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
jrhee17: Actually I made `ProxyConfig` an abstract class with a default visibility constructor to prevent library users from extending on their own. Let me know if you have a better idea (or if an interface is just fine since we are leaving warning logs)

trustin: Ah, OK, then.

core/src/main/java/com/linecorp/armeria/client/proxy/ConnectProxyConfig.java
```diff
@@ -25,18 +27,41 @@
...
+                      "Username and password must either both be null or non-null.");
+        this.proxyAddress = proxyAddress;
+        this.username = username;
         this.password = password;
-        this.useSsl = useSsl;
+        this.useTls = useTls;
+    }
+
+    /**
+     * The configured proxy address.
```
trustin: Returns the ...

core/src/main/java/com/linecorp/armeria/client/proxy/ConnectProxyConfig.java
```diff
@@ -50,17 +75,17 @@ public String password() {
     /**
      * Whether ssl is enabled.
```
trustin: Returns whether TLS is enabled

core/src/main/java/com/linecorp/armeria/client/proxy/ConnectProxyConfig.java
```diff
@@ -25,18 +27,41 @@
...
+
+    /**
+     * The configured proxy address.
+     */
+    public InetSocketAddress proxyAddress() {
+        return proxyAddress;
+    }
+
+    /**
+     * The configured username.
```
trustin: Returns the ...

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -16,83 +16,86 @@
...
-
 import com.linecorp.armeria.client.ClientFactory;
 
 /**
  * Contains base configuration for proxy related settings used in {@link ClientFactory}.
  */
 public abstract class ProxyConfig {
 
+    ProxyConfig() {
+    }
```
trustin: Please move down.

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -16,83 +16,86 @@
...
 
     /**
-     * Creates a {@code ProxyConfig} configuration for SOCKS5 protocol.
+     * Creates a {@code ProxyConfig} configuration for SOCKS4 protocol.
      * @param proxyAddress The proxy address.
+     * @param username The user name.
      */
-    public static Socks5ProxyConfigBuilder socks5(InetSocketAddress proxyAddress) {
-        return new Socks5ProxyConfigBuilder(requireNonNull(proxyAddress));
+    public static Socks4ProxyConfig socks4(InetSocketAddress proxyAddress, @Nullable String username) {
```
trustin: Perhaps better making `username` non-null, given that we have the version that does not require `username`?

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -16,83 +16,86 @@
...
-    private final InetSocketAddress proxyAddress;
-
-    @Nullable
-    private final String username;
-
-    ProxyConfig(InetSocketAddress proxyAddress, @Nullable String username) {
-        this.proxyAddress = proxyAddress;
-        this.username = username;
+    public static Socks5ProxyConfig socks5(
+            InetSocketAddress proxyAddress, @Nullable String username, @Nullable String password) {
```
trustin: Ditto - `username` and `password` could be non-null.

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -16,83 +16,86 @@
...
+     * @param proxyAddress The proxy address.
+     * @param username The user name.
+     * @param password The password.
+     * @param useTls Whether to use TLS to connect to the proxy.
      */
-    @Nullable
-    public String username() {
-        return username;
+    public static ConnectProxyConfig connect(
+            InetSocketAddress proxyAddress, @Nullable String username, @Nullable String password,
```
trustin: Ditto - `username` and `password` could be non-null.

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
jrhee17: Sorry, I always seem to forget this :P

trustin: No worries. We love you, fellow contributor. :heart:

core/build.gradle
```diff
@@ -106,7 +106,7 @@ dependencies {
     optionalApi 'io.prometheus:simpleclient_common'
 
     // Netty
-    [ 'netty-transport', 'netty-codec-http2', 'netty-codec-haproxy', 'netty-resolver-dns' ].each {
+    [ 'netty-transport', 'netty-codec-http2', 'netty-codec-haproxy', 'netty-resolver-dns', 'netty-handler-proxy' ].each {
```
trustin: `netty-handler-proxy` could be declared at `implementation` configuration? I know we need to move some other netty modules from `api` to `implementation`, but I thought this may be a good baby step.

core/src/main/java/com/linecorp/armeria/client/proxy/ConnectProxyConfig.java
```diff
@@ -0,0 +1,87 @@
...
+package com.linecorp.armeria.client.proxy;
+
+import java.net.InetSocketAddress;
+
+import javax.annotation.Nullable;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * Contains CONNECT proxy related configuration.
```
trustin: Could remove `Contains` and `related` from all `ProxyConfig` classes, i.e. `CONNECT proxy configuration`.

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,112 @@
...
+
+import static com.linecorp.armeria.client.proxy.DisabledProxyConfig.DISABLED_PROXY_CONFIG;
+import static java.util.Objects.requireNonNull;
+
+import java.net.InetSocketAddress;
+
+import com.linecorp.armeria.client.ClientFactory;
+
+/**
+ * Contains base configuration for proxy related settings used in {@link ClientFactory}.
```
trustin: `Base configuration for proxy settings used by {@link ClientFactory}`

trustin:
I see two potential follow-up items from https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html

- Auto-detect the default proxy settings from the system properties
- Add `ProxyConfigSelector` which is similar to `ProxySelector`, hopefully with a way to adapt an existing `ProxySelector` implementation into our `ProxyConfigSelector`.

Interested? :wink: Besides this, I think this pull request is pretty much done. Great job, @jrhee17!

testing-internal/src/main/java/com/linecorp/armeria/internal/testing/DynamicBehaviorHandler.java
```diff
@@ -0,0 +1,79 @@
...
+                customizerRef = writeCustomizer;
+        if (customizerRef != null) {
+            customizerRef.accept(ctx, msg, promise);
+        } else {
+            super.write(ctx, msg, promise);
+        }
+    }
+
+    @FunctionalInterface
+    public interface ThrowingTriConsumer<A,B,C> {
```
ikhoon: nit
```suggestion
    public interface ThrowingTriConsumer<A, B, C> {
```

testing-internal/src/main/java/com/linecorp/armeria/internal/testing/DynamicBehaviorHandler.java
```diff
@@ -0,0 +1,79 @@
...
+        void accept(A a, B b, C c) throws Exception;
+    }
+
+    @FunctionalInterface
+    public interface ThrowingBiConsumer<A,B> {
```
ikhoon:
```suggestion
    public interface ThrowingBiConsumer<A, B> {
```

testing-internal/src/main/java/com/linecorp/armeria/internal/testing/NettyServerExtension.java
```diff
@@ -0,0 +1,79 @@
...
+    private final EventLoopGroup workerGroup;
+    @Nullable
+    private Channel channel;
+
+    protected NettyServerExtension() {
+        bossGroup = new NioEventLoopGroup(1);
+        workerGroup = new NioEventLoopGroup(1);
+    }
+
+    public InetSocketAddress address() {
```
ikhoon: Could be `final`?

core/src/test/java/com/linecorp/armeria/client/proxy/ProxyClientIntegrationTest.java
```diff
@@ -0,0 +1,523 @@
...
+                if (!f.isSuccess()) {
+                    clientCtx.close();
+                    return;
+                }
+                backend = f.channel();
+                flush();
+            });
+        }
+
+        private void flush() {
```
ikhoon: nit: `writeToBackendAndFlush()` seems to be better for this method name. 😀

core/src/test/java/com/linecorp/armeria/client/proxy/ProxyClientIntegrationTest.java
```diff
@@ -0,0 +1,523 @@
...
+        final String username = "username";
+        SOCKS_DYNAMIC_HANDLER.setChannelReadCustomizer((ctx, msg) -> {
+            if (msg instanceof DefaultSocks4CommandRequest) {
+                assertThat(username.equals(((DefaultSocks4CommandRequest) msg).userId()));
+            }
+            ctx.fireChannelRead(msg);
+        });
+
+        final ClientFactory clientFactory = ClientFactory.builder().proxyConfig(
+                ProxyConfig.socks4(socksProxyServer.address(), username)).build();
```
ikhoon: nit: Indent for readability
```suggestion
        final ClientFactory clientFactory =
                ClientFactory.builder()
                             .proxyConfig(ProxyConfig.socks4(socksProxyServer.address(), username))
                             .build();
```

core/src/main/java/com/linecorp/armeria/client/proxy/DisabledProxyConfig.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package com.linecorp.armeria.client.proxy;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * Represents client-side proxy is disabled.
+ */
+public final class DisabledProxyConfig extends ProxyConfig {
```
minwoox: Let's remove public here then.

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public abstract class ProxyConfig {
+
+    /**
+     * Creates a {@code ProxyConfig} configuration for SOCKS4 protocol.
+     * @param proxyAddress The proxy address.
```
minwoox: Let's add an empty line before the parameter description.

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Returns a {@code ProxyConfig} which signifies that proxy is disabled.
+     */
+    public static ProxyConfig disabled() {
+        return DISABLED_PROXY_CONFIG;
+    }
+
+    ProxyConfig() {
```
minwoox: This class doesn't have any implementation. How about making this as an interface?

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -118,6 +128,41 @@ protected void initChannel(Channel ch) throws Exception {
                                                       .get(ChannelOption.CONNECT_TIMEOUT_MILLIS);
     }
 
+    private void configureProxy(Channel ch, ProxyConfig proxyConfig, SslContext sslCtx) {
+        if (proxyConfig instanceof DisabledProxyConfig) {
```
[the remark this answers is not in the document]
trustin: Or an enum, like ProxyType? E.g. ProxyType.SOCKS5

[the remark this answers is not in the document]
minwoox: Ah that's better. 😄

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
trustin: It's because we don't want a user to extend it.

minwoox: Ah, thanks!

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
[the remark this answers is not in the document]
jrhee17: So the reason I removed enum `ProxyType` previously was basically because I didn't like how when the end-user types `Proxy...`, intellij autocomplete would show `ProxyType`, but it doesn't really serve any purpose to the end-user.

For us, I definitely agree the `switch case` statement looks a lot cleaner 😄, so i guess this is a trade-off.

We can 1) keep as-is 2) revert back (I intentionally created a separate commit) 3)  move all proxy related classes to `com.linecorp.armeria.client` and keep `ProxyType` package private.

Let me know what you guys think 🙇

jrhee17:
Starting documentation of this feature 🚀 

> Interested? 😉

You already know it 😉✋🙆‍♂

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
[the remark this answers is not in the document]
trustin: SGTM :-)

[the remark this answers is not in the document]
minwoox: For me, I like the current code because I think it's much readable. 😉
Also, I don't think we don't have to move the package because:
- It doesn't harm to show one more class which starts with `Proxy...` because the IntelliJ shows them(there are so many classes already but we can know which one is from Armeria) with the package and I think the name, which is `ProxyType`, is clear to understand.

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Base configuration for proxy settings used by {@link ClientFactory}.
+ */
+public abstract class ProxyConfig {
+
+    /**
+     * Creates a {@code ProxyConfig} configuration for SOCKS4 protocol.
+     *
+     * @param proxyAddress The proxy address.
```
minwoox: nit but global comment:
```suggestion
     * @param proxyAddress The proxy address.
```

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOption.java
```diff
@@ -214,6 +215,12 @@
     public static final ClientFactoryOption<MeterRegistry> METER_REGISTRY =
             define("METER_REGISTRY", Metrics.globalRegistry);
 
+    /**
+     * The {@link ProxyConfig} which contains proxy related configuration.
+     */
+    public static final ClientFactoryOption<ProxyConfig> PROXY_CONFIG =
+            define("PROXY_CONFIG", ProxyConfig.disabled());
```
trustin: How about renaming to `ProxyConfig.direct()` and `DirectProxyConfig`, given that JDK uses `DIRECT`?

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
ikhoon: I finally catched the difference! 🤣
image

minwoox: Oops, What I meant is, `The` should start with a lowercase. :sweat_smile:

jrhee17: > I finally caught the difference! 

🦅👁.    😆😆😆

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -110,15 +108,10 @@ public static ConnectProxyConfig connect(
...
-     * Returns a {@code ProxyConfig} which signifies that proxy is disabled.
+     * Returns a {@code ProxyConfig} which signifies that a proxy is absent.
      */
-    public static ProxyConfig disabled() {
-        return DISABLED_PROXY_CONFIG;
-    }
-
-    @Nullable
-    static String maskedStr(@Nullable String value) {
-        return value == null ? null : "****";
```
trustin: How about reviving this and using `omitNullValues()` for `ToStringHelper`? Showing `****` for a non-existent password may give a user a misconception that authentication is enabled.

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Returns a {@code ProxyConfig} which signifies that proxy is disabled.
+     */
+    public static ProxyConfig disabled() {
+        return DISABLED_PROXY_CONFIG;
+    }
+
+    @Nullable
+    static String maskedStr(@Nullable String value) {
+        return value == null ? null : "****";
```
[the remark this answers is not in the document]
trustin: See https://github.com/line/armeria/pull/2496#discussion_r391990212

core/src/main/java/com/linecorp/armeria/client/proxy/ProxyConfig.java
trustin: We could still return `****` when `username` is not `null`, though.

core/src/main/java/com/linecorp/armeria/client/HttpChannelPool.java
```diff
@@ -118,6 +127,43 @@ protected void initChannel(Channel ch) throws Exception {
...
+                    proxyHandler = new HttpProxyHandler(connectProxyConfig.proxyAddress());
+                } else {
+                    proxyHandler = new HttpProxyHandler(connectProxyConfig.proxyAddress(), username, password);
+                }
+                if (connectProxyConfig.useTls()) {
+                    ch.pipeline().addLast(sslCtx.newHandler(ch.alloc()));
+                }
+                break;
+            default:
+                logger.warn("{} Ignoring unknown proxy type: {}", ch, proxyConfig.getClass().getSimpleName());
```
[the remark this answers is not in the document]
jrhee17: 😅

trustin:
:tada::tada::tada:

PR #2499, opened 2020-02-17
Always abort a failed client request

trustin:
Motivation:

`ClientUtil.failAndGetFallbackResponse()` aborts a request
conditionally, but I don't see a reason to do so. Also,
`DefaultWebClient.execute()` passes a similar logic as a fallback
function, but it doesn't seem necessary anymore.

Modifications:

- Always abort the request and finish logging.

Result:

- Less chance of unsubscribed requests


trustin:
Thanks for reviewing, @ikhoon!

Issue #2378, opened 2020-01-06
Add CompletableFuture converters to armeria-rxjava


anuraaga:
I wrote my first code using `armeria-rxjava` and I was surprised that despite us having official integration, I still needed another library, `future-converter` to actually use it with Armeria, which heavily uses `CompletableFuture`. Should we expose these converters in `armeria-rxjava` itself? We could probably just shade that library in, or reimplement since it's not a lot of code.

ikhoon:
The `future-converter` is an interesting library. https://github.com/lukas-krecan/future-converter

ikhoon:
I'm working on this issue.

I tested [`FutureConverter.toSingle(cf)`](https://github.com/lukas-krecan/future-converter/blob/master/rxjava2-java8/src/main/java/net/javacrumbs/futureconverter/java8rx2/FutureConverter.java#L40) to reimplemente it in Armeria. It emits a `ProtocolViolationException` exception when I specify a `Scheduler` to `Single`.

```java
// Works
Single.just(10)
      .flatMap(i -> Single.just(i + 10))
      .subscribeOn(Schedulers.from(pool))
      .subscribe(i -> {}, ex -> logger.error("ex", ex));

// Not work
final CompletableFuture<Integer> completableFuture = new CompletableFuture<>();
FutureConverter.toSingle(completableFuture)
               .flatMap(i -> Single.just(i + 10))
               .subscribeOn(Schedulers.from(pool))
               .subscribe(i -> {}, ex -> logger.error("ex", ex));
completableFuture.complete(20);

// Works
final CompletableFuture<Integer> completableFuture = new CompletableFuture<>();
FutureConverter.toSingle(completableFuture)
               .flatMap(i -> Single.just(i + 10))
               .subscribe(i -> {}, ex -> logger.error("ex", ex));
completableFuture.complete(20);
```

The error messages:
```
io.reactivex.exceptions.ProtocolViolationException: Disposable already set!
	at io.reactivex.internal.disposables.DisposableHelper.reportDisposableSet(DisposableHelper.java:156)
	at io.reactivex.internal.disposables.DisposableHelper.setOnce(DisposableHelper.java:82)
	at io.reactivex.internal.operators.single.SingleFlatMap$SingleFlatMapCallback.onSubscribe(SingleFlatMap.java:66)
[...]

PR #2501, opened 2020-02-17
Support RxJava 3

ikhoon:
Motivation:
Recently RxJava 3 has been released.
https://github.com/ReactiveX/RxJava/releases/tag/v3.0.0
Armeria supports RxJava2 integration with `armeria-rxjava`.
This PR migrates `RequestContextAssembly` from RxJava2 to RxJava3.
RxJava 3 being based on Java 8 also supports seamless conversions between CompletionStage and Single, Maybe, Observable.

Modifications:
* Rename original `armeria-rxjava` to `armeria-rxjava2`
* Make `armeria-rxjava` support RxJava 3
* Use built-in converter methods for ObservableResponseConverterFunction
* Change `*Callable` to `*Supplier`
  https://github.com/ReactiveX/RxJava/pull/6511
* Remove `assemblyContext.push()` in `RequestContextScalarCallable*.get()`
  * Don't need to push `RequestContext` for scalar type such as `Maybe.just(T)`
* Delegate `reset()` method to `RequestContextConnectableFlowable`, `RequestContextConnectableObservable`
  https://github.com/ReactiveX/RxJava/wiki/What's-different-in-3.0#connectable-source-reset
* Migrate `examples/context-propagation/rxjava` to RxJava 3

Result:
You can now use RxJava 3 with Armeria.
Fixes: #2378


ikhoon:
@anuraaga @kojilin might be interested in this PR. 😉

rxjava/src/main/java/com/linecorp/armeria/rxjava/RequestContextScalarSupplierMaybe.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public T get() {
+        try (SafeCloseable ignored = assemblyContext.push()) {
+            return ((ScalarSupplier<T>) source).get();
+        }
+    }
```
ikhoon: The value of `ScalarSupplier`, which was `ScalarCallable` in RxJava 2, was captured when calling `Maybe.just(T)`.
I wonder do we need to push `RequestContext` here? If not, I think we could remove all `RequestContext*ScalarSupplier*". 🧐

rxjava/src/main/java/com/linecorp/armeria/rxjava/RequestContextAssembly.java
```diff
@@ -0,0 +1,243 @@
...
+        oldOnConnectableFlowableAssembly = RxJavaPlugins.getOnConnectableFlowableAssembly();
+        RxJavaPlugins.setOnConnectableFlowableAssembly(
+                compose(oldOnConnectableFlowableAssembly,
+                        new ConditionalOnCurrentRequestContextFunction<ConnectableFlowable>() {
+                            @Override
+                            ConnectableFlowable applyActual(
+                                    ConnectableFlowable cf,
+                                    RequestContext ctx) {
+                                return new RequestContextConnectableFlowable(
+                                        cf, ctx);
```
trustin: Could merge the two lines

rxjava/src/main/java/com/linecorp/armeria/rxjava/RequestContextAssembly.java
```diff
@@ -0,0 +1,243 @@
...
+    @SuppressWarnings("rawtypes")
+    @Nullable
+    @GuardedBy("RequestContextAssembly.class")
+    private static Function<? super ParallelFlowable, ? extends ParallelFlowable> oldOnParallelAssembly;
+
+    @GuardedBy("RequestContextAssembly.class")
+    private static boolean enabled;
+
+    private RequestContextAssembly() {
+    }
```
trustin: Could be moved after the public static methods

rxjava2/src/main/java/com/linecorp/armeria/server/rxjava2/package-info.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Provide a default {@link com.linecorp.armeria.server.annotation.ResponseConverterFunction}
```
trustin: Provides

rxjava/src/main/java/com/linecorp/armeria/internal/server/rxjava/package-info.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Provide a default {@link com.linecorp.armeria.server.annotation.ResponseConverterFunction}
```
trustin: Provides

rxjava/src/main/java/com/linecorp/armeria/rxjava/RequestContextScalarSupplierMaybe.java
minwoox: I think we don't have to push the context here. 😄

minwoox: On second thought, I think there a chance that a user specifies a class that extends `Maybe` and implements `ScalarSupplier`.
The specified class might need the context in `get()` method, so I think we need this.

ikhoon: Hmm... `ScalarSupplier` is the same as `Supplier` except it abandon `throws Throwable`. If a user trigger effects in `get()` method, it seems a wrong usage of `ScalarSupplier`. Because it should hold const and scalar value. 
https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/internal/fuseable/ScalarSupplier.java#L19 
I think the user has to use `Supplier` for generating value dynamically.

minwoox: Ah, yes. That's a good point. 👍

rxjava/src/main/java/com/linecorp/armeria/common/rxjava/RequestContextSupplierFlowable.java
```diff
@@ -0,0 +1,58 @@
...
+            } else {
+                source.subscribe(new RequestContextSubscriber<>(s, assemblyContext));
+            }
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public T get() throws Throwable {
+        try (SafeCloseable ignored = assemblyContext.push()) {
```
kojilin: As the comment above, do we need to remove here too?

ikhoon: This implements `Supplier<T>`, not `ScalarSupplier<T>`. So I think this is fine. 😀

trustin:
Thanks a lot, @ikhoon !

ikhoon:
Thanks for reviewing! 🙏

Issue #2408, opened 2020-01-17
Allow specifying the port (or nic) for Spring Boot Actuator


trustin:
`WebOperationService` is currently always bound to `Server`'s all ports. It'd be nice if there's a way to limit/disable `WebOperationService` for other ports than a specific port/remoteAddr/NIC.

Armeria currently is not capable of binding a service to a specific port only, but we could simply respond with 404 if the request came from a wrong port. Internally this could be implemented as a generic decorator that works for any `HttpService`.

As a workaround, a user can specify a custom route decorator:

```java
@Bean
public ArmeriaServerConfigurator secureActuator() {
    return builder -> {
		builder.routeDecorator()
               .pathPrefix("/internal")
               .build((delegate, ctx, req) -> {
                   final InetSocketAddress laddr = ctx.localAddress();
                   if (laddr.getPort() != INTERNAL_PORT) {
                       return HttpResponse.of(404);
                   } else {
                       return delegate.serve(ctx, req);
                   }
               });
    };
}
```

heowc:
In spring boot actuator, the base path of the provided actuator is `/actuator`. 
In addition, is there a need to include `internal/docs`, `internal/metrics` and `internal/healthcheck` in armeria?

trustin:
@heowc Yes. It'd be nice to provide this feature as a part of the core, too.

PR #2502, opened 2020-02-17
Add not to bind on the specific port(s) from Spring Boot (Actuator)

heowc:
Motivation:

- Related #2408 

Modifications:

- Copy `ArmeriaSetting` to `boot-actuator-autoconfigure`
- If `management.server.port` is set, protect all services under `/internal/` and actuator services on other ports.
- If `management.server.port` is set, support `@LocalManagementPort`
- Add test

Result:

- Fixes: #2408


spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
```diff
@@ -61,6 +61,9 @@
  *     mime-types: text/*, application/json
  *     excluded-user-agents: some-user-agent, another-user-agent
  *     min-response-size: 1KB
+ *   security:
+ *     enabled: true
+ *     ports: 8080
```
trustin: Question: Is it possible to use `management.server.port` instead?

heowc: It is not possible right now. However, if `armeria.security.enabled` is true, and ` armeria.security.ports` is not defined, I think it's a good approach to look at `management.server.port` and assign them.

heowc: Question: Would you like to remove the `armeria.security.ports` value instead of the optional value?

heowc: I tried `management.server.port` and it doesn't bind to that value. Do you want this property to be an additional binding?

trustin: > Question: Do you want to remove the `armeria.security.ports` value instead of the optional value?

Yes, we might want to add it later but not until there's a user request.

> I tried `management.server.port` and it doesn't bind to that value. Do you want this property to be an additional binding?

Yes, please.

heowc: Thanks for the reply. :smile:

We can apply security by checking `management.server.port` when using actuator. However, we're obscure which ports to secure `internal/docs`, `internal/healthcheck`, and `internal/metrics`.  :flushed:

spring/boot-autoconfigure/build.gradle
```diff
@@ -15,5 +15,7 @@ dependencies {
     testImplementation project(':grpc')
     testImplementation project(':thrift')
     testImplementation 'org.springframework.boot:spring-boot-starter-actuator'
-    testImplementation 'org.springframework.boot:spring-boot-starter-test'
+    testImplementation ('org.springframework.boot:spring-boot-starter-test') {
+        exclude group: 'org.junit.vintage', module: 'junit-vintage-engine'
+    }
```
heowc: I changed all test cases in `spring-boot-autoconfigure` to JUnit5. The test ran twice, showing unintended failures in certain situations.

trustin: > JUnit 5

IIRC, Spring Boot 1 does not support JUnit 5 officially, so we should stay with JUnit 4.

> I think we should never specify a certain port when testing. Can you just specify port `0` to use an ephemeral port?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
trustin: That's a good question. Perhaps we could default to all services under `/internal/` and make it configurable?

spring/boot-autoconfigure/build.gradle
heowc: > IIRC, Spring Boot 1 does not support JUnit 5 officially, so we should stay with JUnit 4.

Oops ... I couldn't confirm this.

> I think we should never specify a certain port when testing. Can you just specify port 0 to use an ephemeral port?

I'll check. Thank you. 😄

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
heowc: How about removing `armeria.security.ports` and adding `armeria.internal.port`?

trustin:
@heowc Now we can resume working on this. Thanks a lot!

heowc:
Yes! I'm preparing for this PR. :smile:

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
heowc: I've configured the default port(8001) if `armeria.security.enabled` is true, and have it protect all services under `/internal/` on other ports.

heowc: Question: I specified the default port, but when the user specified `management.server.port` I didn't remove the default port. Should I avoid specifying a default port?

trustin:
Ready for a review?

heowc:
@trustin, the answer was late. 😢 

I have one problem, so I will solve it and apply for a review.

In this situation, the default port and the management port are bound separately, so the management paths are missing from the default port.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSecurityAutoConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final Port DEFAULT_SECURITY_PORT = new Port().setPort(8001)
+                                                                .setProtocol(SessionProtocol.HTTP);
+
+    @Bean
+    ArmeriaServerConfigurator secureArmeriaServerConfig(ArmeriaSettings settings,
+                                                        ConfigurableEnvironment environment,
+                                                        @Value("${management.server.port:#{null}}")
+                                                        Integer actuatorPort,
+                                                        @Value("${management.endpoints.web.base-path:#{null}}")
+                                                        String actuatorBasePath) {
```
heowc, line 59: I fetched Spring Actuator related fields to [solve what I mentioned](https://github.com/line/armeria/pull/2502#issuecomment-631326462). I am not sure if this is a good way. what do you think?

trustin: I think it's OK. If there's any issue with this, we can fix it later.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
```diff
@@ -61,6 +61,8 @@
  *     mime-types: text/*, application/json
  *     excluded-user-agents: some-user-agent, another-user-agent
  *     min-response-size: 1KB
+ *   security:
+ *     enabled: true
```
trustin, line 65: Question: Should we expose this option or always enable it when a different management port was specified?

spring/boot-autoconfigure/src/test/resources/config/application-secureTest.yml
```diff
@@ -0,0 +1,5 @@
+armeria:
+  ports:
+    - port: 0
+  security:
+    enabled: true
```
trustin, line 5: Missing newline

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
heowc: I have modified some code. I think I can divide it into the following cases.
- `armeria.security.enabled` = false, -> nothing (not secure)
- `armeria.security.enabled` = true, -> port = 7000 (secure)
- `armeria.security.enabled` = false, `management.server.port` = 1111 -> port = 1111 (not secure)
- `armeria.security.enabled` = true, `management.server.port` = 1111 -> port = 1111 (secure)

spring/boot-autoconfigure/src/test/resources/config/application-secureTest.yml
heowc: Fixed

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSecurityAutoConfiguration.java
```diff
@@ -0,0 +1,127 @@
...
+import com.linecorp.armeria.spring.ArmeriaSettings.Port;
+
+/**
+ * Spring Boot {@link Configuration} that provides Armeria integration.
+ */
+@Configuration
+@EnableConfigurationProperties(ArmeriaSettings.class)
+public class ArmeriaSecurityAutoConfiguration {
+
+    private static final Port DEFAULT_SECURITY_PORT = new Port().setPort(8001)
```
minwoox, line 50: Question: Where is the `8001` from? Spring uses it?

minwoox: I think we shouldn't use the default port.
If a use forgets to disable the security, an attacker might use the default port to get the information of the server. What do you think?

heowc: I understood [here](https://github.com/line/armeria/pull/2502#discussion_r388317746) to mean adding a default port.

heowc: If so, should this feature be enabled only by adding a spring-actuator?

heowc:
@trustin @minwoox,

I had a misunderstanding and corrected it.
- Remove `armeria.security.enabled` property
- Remove the default port (8001)
- When using the actuator, if the user specifies `management.server.port`, a decorator is added to secure internal services.

spring/boot-actuator-autoconfigure/src/test/java/com/linecorp/armeria/spring/actuate/ArmeriaSpringActuatorAutoConfigurationTest.java
```diff
@@ -305,4 +288,52 @@ public void testOptions() {
...
+        private Server server;
+        @Inject
+        private ArmeriaSettings settings;
+
+        @Test
+        void normal() throws Exception {
+            server.activePorts().values().stream()
+                  .map(p -> p.localAddress().getPort())
+                  .forEach(port -> {
+                      final Integer statusCode = actuatorPort.equals(port) ? 200 : 404;
```
trustin, line 316: nit: Could just use `int`?

spring/boot-actuator-autoconfigure/src/test/java/com/linecorp/armeria/spring/actuate/ArmeriaSpringActuatorAutoConfigurationTest.java
```diff
@@ -305,4 +288,52 @@ public void testOptions() {
...
+                      assertStatus(port, "/actuator/loggers/" + TEST_LOGGER_NAME, statusCode);
+                      assertStatus(port, "/actuator/prometheus", statusCode);
+                      assertStatus(port, settings.getDocsPath(), statusCode);
+                      assertStatus(port, settings.getHealthCheckPath(), statusCode);
+                      assertStatus(port, settings.getMetricsPath(), statusCode);
+                  });
+        }
+    }
+
+    private static void assertStatus(Integer port, String url, Integer statusCode) {
```
trustin, line 328: Ditto - `int` should work?

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSettings.java
heowc: I removed it.

spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/ArmeriaSecurityAutoConfiguration.java
heowc: I removed it. :)

spring/boot-actuator-autoconfigure/src/test/java/com/linecorp/armeria/spring/actuate/ArmeriaSpringActuatorAutoConfigurationTest.java
heowc: Fixed cb9cf10

spring/boot-actuator-autoconfigure/src/test/java/com/linecorp/armeria/spring/actuate/ArmeriaSpringActuatorAutoConfigurationTest.java
heowc: Fixed cb9cf10

spring/boot-actuator-autoconfigure/src/main/java/com/linecorp/armeria/spring/actuate/ArmeriaSpringActuatorAutoConfiguration.java
```diff
@@ -196,6 +215,70 @@ ArmeriaServerConfigurator actuatorServerConfigurator(
...
+                addLocalManagementPortPropertyAlias(environment, port);
+                configureSecureDecorator(sb, port, properties.getBasePath(), armeriaSettings);
+            }
+        };
+    }
+
+    @Nullable
+    private static Port obtainManagementServerPort(@Nullable Integer port) {
+        return Optional.ofNullable(port)
+                       .filter(it -> it.equals(0))
```
minwoox, line 237: What happens if the management port number is not `0`?

heowc: Oops... I fixed this. 😨

spring/boot-actuator-autoconfigure/src/main/java/com/linecorp/armeria/spring/actuate/ArmeriaSpringActuatorAutoConfiguration.java
```diff
@@ -196,6 +215,67 @@ ArmeriaServerConfigurator actuatorServerConfigurator(
...
+            configureSecureDecorator(sb, port, properties.getBasePath(), armeriaSettings);
+        };
+    }
+
+    private static Port obtainManagementServerPort(Integer port) {
+        int actualPort = requireNonNull(port, "port");
+        if (actualPort == 0) {
+            actualPort = SocketUtils.findAvailableTcpPort();
+        }
+        return new Port().setPort(actualPort).setProtocol(SessionProtocol.HTTP);
```
ikhoon, line 237: Question: Don't we need to support HTTPS? The upstream seems to support TLS/SSL configuration.
https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-management-specific-ssl

If that is worthy to handle, I think you can do it in the following PR. 😉

heowc: Can I set up another certificate on a specific port in Armeria? I think this is a bit vague.

Anyway, I felt the need to deal with [this](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-disabling-http-endpoints).

ikhoon: > Can I set up another certificate on a specific port in Armeria? I think this is a bit vague.

I thought that the actuator shares the Server's certification.
We cannot perfectly handle the feature, now. Please ignore this.

> Anyway, I felt the need to deal with this. WDYT?

That sounds good. 👍

heowc: Thanks :)

trustin:
:tada: Thanks, @heowc !

minwoox:
👍👍👍👍👍

PR #2503, opened 2020-02-17
Make `Throttling*Service` responses configurable

max904-github:
…nfigurable):

- added AbstractThrottlingService.strategy() method to provide access to the strategy set to the throttling service.
- added ThrottlingStrategy.failureStatus() method to provide status for throttled requests.
- adjusted ThrottlingService and RpcThrottlingService to leverage ThrottlingStrategy.failureStatus().
- adjusted various constructor methods to support failureStatus parameter.
- modified unit tests to cover the failure status.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 LINE Corporation
+ * Copyright 2020 LINE Corporation
  *
  * LINE Corporation licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
@@ -47,9 +47,9 @@ public abstract class AbstractThrottlingService<I extends Request, O extends Res
                                         ThrottlingRejectHandler<I, O> rejectHandler) {
         super(delegate);
         this.strategy = requireNonNull(strategy, "strategy");
-        this.responseConverter = requireNonNull(responseConverter);
-        this.acceptHandler = acceptHandler;
-        this.rejectHandler = rejectHandler;
+        this.responseConverter = requireNonNull(responseConverter, "responseConverter");
+        this.acceptHandler = requireNonNull(acceptHandler, "acceptHandler");
+        this.rejectHandler = requireNonNull(rejectHandler, "rejectHandler");
     }
 
     @Override
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
@@ -20,8 +20,6 @@ import static java.util.Objects.requireNonNull;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
 
-import javax.annotation.Nullable;
-
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Response;
 import com.linecorp.armeria.common.util.Exceptions;
@@ -37,55 +35,32 @@ public abstract class AbstractThrottlingService<I extends Request, O extends Res
 
     private final ThrottlingStrategy<I> strategy;
     private final Function<CompletionStage<? extends O>, O> responseConverter;
+    private final ThrottlingAcceptHandler<I, O> acceptHandler;
+    private final ThrottlingRejectHandler<I, O> rejectHandler;
 
     /**
      * Creates a new instance that decorates the specified {@link Service}.
      */
     protected AbstractThrottlingService(Service<I, O> delegate, ThrottlingStrategy<I> strategy,
-                                        Function<CompletionStage<? extends O>, O> responseConverter) {
+                                        Function<CompletionStage<? extends O>, O> responseConverter,
+                                        ThrottlingAcceptHandler<I, O> acceptHandler,
+                                        ThrottlingRejectHandler<I, O> rejectHandler) {
         super(delegate);
         this.strategy = requireNonNull(strategy, "strategy");
         this.responseConverter = requireNonNull(responseConverter);
+        this.acceptHandler = acceptHandler;
+        this.rejectHandler = rejectHandler;
     }
-
-    protected ThrottlingStrategy<I> strategy() {
-        return strategy;
-    }
-
-    /**
-     * Invoked when {@code req} is not throttled. By default, this method delegates the specified {@code req} to
-     * the {@link #delegate()} of this service.
-     * @param ctx request context
-     * @param req {@link Request} to process
-     * @return {@link Response} returned after successful processing
-     * @throws Exception when something wrong happens
-     */
-    protected O onSuccess(ServiceRequestContext ctx, I req) throws Exception {
-        return delegate().serve(ctx, req);
-    }
-
-    /**
-     * Invoked when {@code req} is throttled. This method responds with the
-     * {@link ThrottlingStrategy#failureStatus()} status.
-     * @param ctx request context
-     * @param req throttled {@link Request}
-     * @param cause exception thrown during execution of
-     *        {@link ThrottlingStrategy#accept(ServiceRequestContext, Request)}
-     * @return {@link Response} returned on throttled request
-     * @throws Exception when something wrong happens
-     */
-    protected abstract O onFailure(ServiceRequestContext ctx, I req, @Nullable Throwable cause)
-            throws Exception;
 
     @Override
     public O serve(ServiceRequestContext ctx, I req) throws Exception {
         return responseConverter.apply(
-                strategy.accept(ctx, req).handleAsync((accept, thrown) -> {
+                strategy.accept(ctx, req).handleAsync((accept, cause) -> {
                     try {
-                        if (thrown != null || !accept) {
-                            return onFailure(ctx, req, thrown);
+                        if (cause != null || !accept) {
+                            return rejectHandler.handleRejected(delegate(), ctx, req, cause);
                         }
-                        return onSuccess(ctx, req);
+                        return acceptHandler.handleAccepted(delegate(), ctx, req);
                     } catch (Exception e) {
                         return Exceptions.throwUnsafely(e);
                     }
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
@@ -22,7 +22,6 @@ import java.util.function.Function;
 
 import javax.annotation.Nullable;
 
-import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Response;
 import com.linecorp.armeria.common.util.Exceptions;
@@ -49,17 +48,31 @@ public abstract class AbstractThrottlingService<I extends Request, O extends Res
         this.responseConverter = requireNonNull(responseConverter);
     }
 
+    protected ThrottlingStrategy<I> strategy() {
+        return strategy;
+    }
+
     /**
      * Invoked when {@code req} is not throttled. By default, this method delegates the specified {@code req} to
      * the {@link #delegate()} of this service.
+     * @param ctx request context
+     * @param req {@link Request} to process
+     * @return {@link Response} returned after successful processing
+     * @throws Exception when something wrong happens
      */
     protected O onSuccess(ServiceRequestContext ctx, I req) throws Exception {
         return delegate().serve(ctx, req);
     }
 
     /**
-     * Invoked when {@code req} is throttled. By default, this method responds with the
-     * {@link HttpStatus#SERVICE_UNAVAILABLE} status.
+     * Invoked when {@code req} is throttled. This method responds with the
+     * {@link ThrottlingStrategy#failureStatus()} status.
+     * @param ctx request context
+     * @param req throttled {@link Request}
+     * @param cause exception thrown during execution of
+     *        {@link ThrottlingStrategy#accept(ServiceRequestContext, Request)}
+     * @return {@link Response} returned on throttled request
+     * @throws Exception when something wrong happens
      */
     protected abstract O onFailure(ServiceRequestContext ctx, I req, @Nullable Throwable cause)
             throws Exception;
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
@@ -32,23 +32,12 @@ abstract class AbstractThrottlingServiceBuilder<I extends Request, O extends Res
     private ThrottlingAcceptHandler<I, O> acceptHandler;
     private ThrottlingRejectHandler<I, O> rejectHandler;
 
-    AbstractThrottlingServiceBuilder(ThrottlingStrategy<I> strategy) {
+    AbstractThrottlingServiceBuilder(ThrottlingStrategy<I> strategy,
+                                     ThrottlingRejectHandler<I, O> defaultRejectHandler) {
         this.strategy = requireNonNull(strategy, "strategy");
-        acceptHandler = requireNonNull(defaultAcceptHandler(), "defaultAcceptHandler");
-        rejectHandler = requireNonNull(defaultRejectHandler(), "defaultRejectHandler");
+        acceptHandler = Service::serve;
+        rejectHandler = requireNonNull(defaultRejectHandler, "defaultRejectHandler");
     }
-
-    /**
-     * Provides default request accept handler.
-     */
-    ThrottlingAcceptHandler<I, O> defaultAcceptHandler() {
-        return Service::serve;
-    }
-
-    /**
-     * Provides default request reject handler.
-     */
-    abstract ThrottlingRejectHandler<I, O> defaultRejectHandler();
 
     final ThrottlingStrategy<I> getStrategy() {
         return strategy;
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.armeria.server.throttling;
+
+import static java.util.Objects.requireNonNull;
+
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.common.Request;
+import com.linecorp.armeria.common.Response;
+import com.linecorp.armeria.server.Service;
+
+/**
+ * Builds a new {@link AbstractThrottlingService}.
+ * @param <I> type of the request
+ * @param <O> type of the response
+ */
+abstract class AbstractThrottlingServiceBuilder<I extends Request, O extends Response> {
+
+    @Nullable
+    private ThrottlingStrategy<I> strategy;
+    private ThrottlingAcceptHandler<I, O> acceptHandler;
+    private ThrottlingRejectHandler<I, O> rejectHandler;
+
+    AbstractThrottlingServiceBuilder() {
+        acceptHandler = requireNonNull(defaultAcceptHandler(), "defaultAcceptHandler");
+        rejectHandler = requireNonNull(defaultRejectHandler(), "defaultRejectHandler");
+    }
+
+    /**
+     * Provides default request accept handler.
+     */
+    protected ThrottlingAcceptHandler<I, O> defaultAcceptHandler() {
+        return Service::serve;
+    }
+
+    /**
+     * Provides default request reject handler.
+     */
+    protected abstract ThrottlingRejectHandler<I, O> defaultRejectHandler();
+
+    /**
+     * Sets {@link ThrottlingStrategy}.
+     */
+    protected void setStrategy(ThrottlingStrategy<I> strategy) {
+        this.strategy = requireNonNull(strategy, "strategy");
+    }
+
+    protected ThrottlingStrategy<I> getStrategy() {
+        if (strategy == null) {
+            throw new IllegalStateException("no " + ThrottlingStrategy.class.getSimpleName() + " was added.");
+        }
+        return strategy;
+    }
+
+    /**
+     * Sets {@link ThrottlingAcceptHandler}.
+     */
+    protected void setAcceptHandler(
+            ThrottlingAcceptHandler<I, O> acceptHandler) {
+        this.acceptHandler = requireNonNull(acceptHandler, "acceptHandler");
+    }
+
+    protected ThrottlingAcceptHandler<I, O> getAcceptHandler() {
+        return acceptHandler;
+    }
+
+    /**
+     * Sets {@link ThrottlingRejectHandler}.
+     */
+    protected void setRejectHandler(
+            ThrottlingRejectHandler<I, O> rejectHandler) {
+        this.rejectHandler = requireNonNull(rejectHandler, "rejectHandler");
+    }
+
+    protected ThrottlingRejectHandler<I, O> getRejectHandler() {
+        return rejectHandler;
+    }
+}
--- a/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 LINE Corporation
+ * Copyright 2020 LINE Corporation
  *
  * LINE Corporation licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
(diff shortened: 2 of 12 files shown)


core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingStrategy.java
```diff
@@ -126,4 +152,11 @@ protected ThrottlingStrategy(@Nullable String name) {
     public String name() {
         return name;
     }
+
+    /**
+     * Returns the failure status to be used for the throttled requests.
+     */
+    protected HttpStatus failureStatus() {
```
anuraaga: Instead of adding just this knob to strategy, how about adding `ThrottlingFailureHandler` similar to our auth service?

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/auth/AuthService.java#L75

By letting it return an arbitrary response, we get more flexibility (including implementing the x-ratelimit headers if we add a token throttler).

max904-github: Ok, I can try that if all reviewers in agreement on that.
I have one problem with this proposed design though - its inability to pass the context directly from an `authorizer` to a `handler`. I experienced that in several places, including `AuthService` and `ThrottlingService`. In either case I ended up using `ServiceRequestContext`. While in case of `ThrottlingService` it was ok (but not very pretty) since I could set various `x-ratelimit` headers using `ctx.addAdditionalResponseHeader()` at `ThrottlingStrategy.accept()`, which sort of defeated the purpose of the error handler. In case of OAuth2 implementation, it was even trickier - token validation can fail in a number of ways and the RFC instructs to properly inform the client by composing specific JSon error response. Therefore setting up the headers to the context inside the `authorize()` was not very useful. What I needed then was an ability to pass some custom set of artefacts from an `authorizer` to a `handler`, including error response code, oauth2 error code and some additional elements none of which were headers.
Ideally, the interfaces like `Authorizer.authorize()` and `ThrottlingStrategy.accept()` should not be binary and should allow responding with some custom context.

anuraaga: Ah that's a good point. I think passing the context into the handler makes sense and we probably want to do so for auth service too. For here, I was expecting the interface for the handler to match the signature of `onFailure` exactly.

max904-github: Yes, I understand the idea of the handler. Could you recommend something on the context passing solution?

anuraaga: I may be missing the concern so let me know - I was thinking if the signature of the handler takes a `ServiceRequestContext` and returns a `HttpResponse`, which is not what the with service currently does, the context issue is solved since, using auth as an example, both the authorizer and handler have the context. So if here if both strategy has context (already does) and handler has it, it works. Let me know if this misses something.

max904-github: Assuming that there are two parties - the authorizer and the handler (consider auth example), where one party (the authorizer) makes decisions based on complex logic (token validation) and the other party (the handler) fulfills the decisions made.
Let's say the authorizer validates the token, and as a result, needs to pass following elements to the handler: HTTP error response code, OAuth error code, some reference data on the token (e.g. token expiry date). The latter two will have to be put into JSON response body. How such context gets passed with the proposed design?

max904-github: Assuming that there are two parties - the authorizer and the handler (consider auth example), where one party (the authorizer) makes decisions based on complex logic (token validation) and the other party (the handler) fulfills the decisions made.
Let's say the authorizer validates the token, and as a result, needs to pass following elements to the handler: HTTP error response code, OAuth error code, some reference data on the token (e.g. token expiry date). The latter two will have to be put into JSON response body. How such context gets passed with the proposed design?

trustin: Via `RequestContext.attr()`? Not sure if it's nice, though.

max904-github: Made a second commit - Implemented `ThrottlingReject`/`AcceptHandler` approach, similar to `AuthService`, as suggested by @anuraaga, to fully customize throttling response rejection/acceptance.
Introduced builders for `ThrottlingService` and `ThrottlingRpcService`.
Changed default to `TOO_MANY_REQUESTS` only for `HTTP`. `RPC` still uses `SERVICE_UNAVAILABLE`.

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+
+    AbstractThrottlingServiceBuilder() {
+        acceptHandler = requireNonNull(defaultAcceptHandler(), "defaultAcceptHandler");
+        rejectHandler = requireNonNull(defaultRejectHandler(), "defaultRejectHandler");
+    }
+
+    /**
+     * Provides default request accept handler.
+     */
+    protected ThrottlingAcceptHandler<I, O> defaultAcceptHandler() {
```
trustin: All protected methods in this class could be package-private and `final`.

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+ * @param <O> type of the response
+ */
+abstract class AbstractThrottlingServiceBuilder<I extends Request, O extends Response> {
+
+    @Nullable
+    private ThrottlingStrategy<I> strategy;
+    private ThrottlingAcceptHandler<I, O> acceptHandler;
+    private ThrottlingRejectHandler<I, O> rejectHandler;
+
+    AbstractThrottlingServiceBuilder() {
```
trustin: Could we require a strategy at construction time? e.g. `ThrottlingRpcService.builder(strategy)...build()`

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright 2018 LINE Corporation
```
trustin: nit: 2020

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingRpcServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+            ThrottlingAcceptHandler<RpcRequest, RpcResponse> acceptHandler) {
+        setAcceptHandler(acceptHandler);
+        return this;
+    }
+
+    /**
+     * Sets {@link ThrottlingRejectHandler}.
+     */
+    public ThrottlingRpcServiceBuilder onRejectedRequest(
+            ThrottlingRejectHandler<RpcRequest, RpcResponse> rejectHandler) {
```
trustin: Could be `ThrottlingRejectHandler<? super RpcRequest, ? extends RpcResponse>`

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingRpcServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+    public ThrottlingRpcServiceBuilder strategy(ThrottlingStrategy<RpcRequest> strategy) {
+        setStrategy(strategy);
+        return this;
+    }
+
+    /**
+     * Sets {@link ThrottlingAcceptHandler}.
+     */
+    public ThrottlingRpcServiceBuilder onAcceptedRequest(
+            ThrottlingAcceptHandler<RpcRequest, RpcResponse> acceptHandler) {
```
trustin: Could be `ThrottlingAcceptHandler<? super RpcRequest, ? extends RpcResponse>`

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+            throw new IllegalStateException("no " + ThrottlingStrategy.class.getSimpleName() + " was added.");
+        }
+        return strategy;
+    }
+
+    /**
+     * Sets {@link ThrottlingAcceptHandler}.
+     */
+    protected void setAcceptHandler(
+            ThrottlingAcceptHandler<I, O> acceptHandler) {
```
trustin: Could be `ThrottlingAcceptHandler<? super I, ? extends O>`

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+
+    protected ThrottlingAcceptHandler<I, O> getAcceptHandler() {
+        return acceptHandler;
+    }
+
+    /**
+     * Sets {@link ThrottlingRejectHandler}.
+     */
+    protected void setRejectHandler(
+            ThrottlingRejectHandler<I, O> rejectHandler) {
```
trustin: Could be `ThrottlingRejectHandler<? super I, ? extends O>`

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingAcceptHandler.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * A callback which is invoked to handle accepted (successful) requests indicated by {@link ThrottlingStrategy}.
+ *
+ * @param <I> the type of incoming {@link Request}. Must be {@link HttpRequest} or {@link RpcRequest}.
+ * @param <O> the type of outgoing {@link Response}. Must be {@link HttpResponse} or {@link RpcResponse}.
+ *
+ * @see ThrottlingServiceBuilder#onAcceptedRequest(ThrottlingAcceptHandler)
+ */
+@FunctionalInterface
+public interface ThrottlingAcceptHandler<I extends Request, O extends Response> {
```
anuraaga: Just wondering, any thoughts on

- Removing accept handler (what can we expect other than just running the delegate?)
- Remove `delegate` from reject handler (what can we expect other than not running delegate?)

I think it's fine to have them too but it does add some complexity which might not ever be needed.

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+        return this;
+    }
+
+    /**
+     * Provides default throttling reject behaviour for {@link HttpRequest}.
+     * Returns an {@link HttpResponse} with {@link HttpStatus#TOO_MANY_REQUESTS}.
+     */
+    @Override
+    protected ThrottlingRejectHandler<HttpRequest, HttpResponse> defaultRejectHandler() {
+        return (delegate, ctx, req, cause) -> HttpResponse.of(HttpStatus.TOO_MANY_REQUESTS);
```
anuraaga: Let's move lambdas to singleton constants

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingRpcServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        return this;
+    }
+
+    /**
+     * Provides default throttling reject behaviour for {@link RpcRequest}.
+     * Responds with {@link HttpStatusException} with {@code 503 Service Unavailable}.
+     */
+    @Override
+    protected ThrottlingRejectHandler<RpcRequest, RpcResponse> defaultRejectHandler() {
+        return (delegate, ctx, req, cause) ->
```
anuraaga: Ditto

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Builds a new {@link AbstractThrottlingService}.
+ * @param <I> type of the request
+ * @param <O> type of the response
+ */
+abstract class AbstractThrottlingServiceBuilder<I extends Request, O extends Response> {
+
+    @Nullable
+    private ThrottlingStrategy<I> strategy;
+    private ThrottlingAcceptHandler<I, O> acceptHandler;
```
anuraaga: I think instead of having `default*Handler` methods we can initialize these to defaults instead of `null`. Doesn't hurt to add the accept handler to the constructor to be called with `super` in child builder.

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingAcceptHandler.java
trustin: > Remove delegate from reject handler (what can we expect other than not running delegate?)
Perhaps delegating and then just counting?

anuraaga: I'd expect anything counting-related would be in the strategy. But not sure.

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingRpcServiceBuilder.java
max904-github: This is how it's done in AuthService, which I followed, as suggested earlier. This new suggestion conflicts with SimpleDecoratingService.

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingRpcServiceBuilder.java
max904-github: This is how it's done in AuthService, which I followed, as suggested earlier. This new suggestion conflicts with SimpleDecoratingService.

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
max904-github: This is how it's done in AuthService, which I followed, as suggested earlier. This new suggestion conflicts with SimpleDecoratingService.

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
max904-github: This is how it's done in AuthService, which I followed, as suggested earlier. This new suggestion conflicts with SimpleDecoratingService.

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
max904-github: Fixed

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
max904-github: Updated

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingAcceptHandler.java
max904-github: Kept the way it is, similar to AuthService.

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingServiceBuilder.java
max904-github: Moved

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingRpcServiceBuilder.java
max904-github: Moved

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
max904-github: Made all methods package private and made internal getters/setters final.

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
max904-github: Kept the way it is

anuraaga: Why?

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingAcceptHandler.java
anuraaga: Looking at `AuthService` was mainly for inspiration, and may be good to follow as is, or maybe can be tweaked to improve the API here it's all case-by-case. I think it may make sense to remove delegate from `AuthService` too after a second look, though that's another topic.

Here, there is even more to worry about since the strategy will actually be counting and making throttling decisions. So an accept handler that doesn't call the delegate stands to mess up the strategy quite easily, as would a failure handler that does call it. Decoupling the execution of a service with the generation of an error response seems cleaner here unless there are other concerns.

Anyways, the goal here is not to follow `AuthService` pattern word for word but to come up with a clean API for throttling.

trustin: I’m slightly in favor of keeping the API as it is. A handle might want to insert some response header, schedule some task or increase a counter on accept/reject.

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
trustin: Calling super with the default handlers sounds better to me.

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingAcceptHandler.java
anuraaga: Pretty sure the strategy can do all of these too - but it's ok to keep it anyways. Let's make sure the javadoc is clear about the expected use of delegate, especially on failure where it will usually cause a problem.

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
max904-github: > Why?
Because this is my design choice.
Defaults are implementation specific and those aren't parameters (which they would become when passing them in the constructor). A descendant is capable of defining its own defaults, and in case of "accept" it's obligated to do so.
Let me know if it violates any rules.

anuraaga: While there are some concepts defined as rules in our contributing guidelines, many are just principles and we use code review to share to contributors. In this case, we use methods as an indicator of our user API and prefer not to use methods for passing parameters to a base class (I'll have to disagree that default behavior isn't a "parameter"). These default handlers should be constants, and there isn't much value in having all subclasses manage the boilerplate to make them constants. Every project will have its own style and level of rigor and that's what we have code review for.

max904-github: Defaults are defining the behavior of the each implementation - `ThrottlingServiceBuilder` and `ThrottlingRpcServiceBuilder`. Such behavior defined by the base class - `AbstractThrottlingServiceBuilder`. `defaultRejectHandler()` method is not part of the class hierarchy public API. And no one suggest/forces to create boilerplate anywhere.
Inheritance and this kind of non-public behavior enforcement used widely in Java programming, including Armeria. You can take `com.linecorp.armeria.client.retry.AbstractBackoff` hierarchy and its `doNextDelayMillis()` method.

core/src/test/java/com/linecorp/armeria/server/throttling/ThrottlingServiceTest.java
```diff
@@ -46,8 +47,13 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req)
     public ServerRule serverRule = new ServerRule() {
         @Override
         protected void configure(ServerBuilder sb) throws Exception {
-            sb.service("/http-never", SERVICE.decorate(ThrottlingService.newDecorator(never())));
             sb.service("/http-always", SERVICE.decorate(ThrottlingService.newDecorator(always())));
+            sb.service("/http-never", SERVICE.decorate(ThrottlingService.newDecorator(never())));
+            sb.service("/http-never-custom", SERVICE.decorate(
+                    ThrottlingService.newDecorator(ThrottlingStrategy.of((ctx, req) -> completedFuture(false)),
```
ikhoon: Why don't we add a test for a custom reject handler?

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingService.java
```diff
@@ -38,41 +35,32 @@
...
     protected AbstractThrottlingService(Service<I, O> delegate, ThrottlingStrategy<I> strategy,
-                                        Function<CompletionStage<? extends O>, O> responseConverter) {
+                                        Function<CompletionStage<? extends O>, O> responseConverter,
+                                        ThrottlingAcceptHandler<I, O> acceptHandler,
+                                        ThrottlingRejectHandler<I, O> rejectHandler) {
         super(delegate);
         this.strategy = requireNonNull(strategy, "strategy");
         this.responseConverter = requireNonNull(responseConverter);
+        this.acceptHandler = acceptHandler;
+        this.rejectHandler = rejectHandler;
```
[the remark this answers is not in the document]
max904-github: Fixed

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Builds a new {@link AbstractThrottlingService}.
+ * @param <I> type of the request
+ * @param <O> type of the response
+ */
+abstract class AbstractThrottlingServiceBuilder<I extends Request, O extends Response> {
+
+    @Nullable
+    private final ThrottlingStrategy<I> strategy;
```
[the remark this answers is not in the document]
max904-github: Fixed

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Provides default request reject handler.
+     */
+    abstract ThrottlingRejectHandler<I, O> defaultRejectHandler();
+
+    final ThrottlingStrategy<I> getStrategy() {
+        if (strategy == null) {
+            throw new IllegalStateException("no " + ThrottlingStrategy.class.getSimpleName() + " was added.");
+        }
```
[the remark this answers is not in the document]
max904-github: Fixed

core/src/test/java/com/linecorp/armeria/server/throttling/ThrottlingServiceTest.java
max904-github: Do you mean custom "accept" handler? If yes, I've added a new test for it.

ikhoon: Ah, Thanks!

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingRpcServiceBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+import com.linecorp.armeria.common.RpcRequest;
+import com.linecorp.armeria.common.RpcResponse;
+import com.linecorp.armeria.server.HttpStatusException;
+import com.linecorp.armeria.server.RpcService;
+import com.linecorp.armeria.server.Service;
+
+/**
+ * Builds a new {@link ThrottlingRpcService}.
+ */
+public class ThrottlingRpcServiceBuilder extends AbstractThrottlingServiceBuilder<RpcRequest, RpcResponse> {
```
trustin: nit: `final`
```suggestion
public final class ThrottlingRpcServiceBuilder
        extends AbstractThrottlingServiceBuilder<RpcRequest, RpcResponse> {
```

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingServiceBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.server.HttpService;
+import com.linecorp.armeria.server.Service;
+
+/**
+ * Builds a new {@link ThrottlingService}.
+ */
+public class ThrottlingServiceBuilder extends AbstractThrottlingServiceBuilder<HttpRequest, HttpResponse> {
```
trustin: nit: `final`
```suggestion
public final class ThrottlingServiceBuilder
        extends AbstractThrottlingServiceBuilder<HttpRequest, HttpResponse> {
```

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingRpcServiceBuilder.java
max904-github: Fixed

core/src/main/java/com/linecorp/armeria/server/throttling/ThrottlingServiceBuilder.java
max904-github: Fixed

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
max904-github: Fixed

core/src/main/java/com/linecorp/armeria/server/throttling/AbstractThrottlingServiceBuilder.java
minwoox: Then, we should fix `AuthService` first. 😄
I must be forgotten when I was doing #2323. But don't worry, after this is merged, I will work on it. Thanks!

PR #2508, opened 2020-02-19
Abort with `ClosedStreamException` on `RST_STREAM`

trustin:
Motivation:

When an HTTP/2 client resets the current stream, Armeria server's
request will be aborted with Netty `Http2Exception`, leaking
implementation detail.

Modifications:

- Abort with `ClosedStreamException` instead of `Http2Exception`.

Result:

- Better user experience

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java
+++ b/core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java
@@ -44,6 +44,7 @@ import com.linecorp.armeria.common.Flags;
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.common.stream.AbortedStreamException;
 import com.linecorp.armeria.common.stream.CancelledSubscriptionException;
+import com.linecorp.armeria.common.stream.ClosedStreamException;
 
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelException;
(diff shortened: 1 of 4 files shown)


anuraaga:
Can you confirm this code is unaffected?

https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/internal/common/grpc/GrpcStatus.java#L106

If we can make sure Http2Exception is never thrown by Armeria we could also reconsider splitting `api` dependencies and `implementation` dependencies for Netty in the gradle config - it's literally the class that made me feel it's not worth splitting.

trustin:
@anuraaga That's a good point. Updated the conversion logic. 

However, we can't move Netty to `implementation` because we use `AttributeKey`, `AsciiString` and `ByteBuf`, etc.

anuraaga:
Yeah we can't move them all but at first I hoped for a limited subset of Netty APIs that aren't strictly implementation detail. But once it becomes something so detailed as `codec-http2`, I gave up. Now it might make sense though

trustin:
Ah yeah, that makes sense. I think we can hide some of them.

trustin:
@anuraaga Created a separate issue for that: https://github.com/line/armeria/issues/2509

trustin:
Thanks for quick reviews :heart:

Issue #2493, opened 2020-02-14
Confusing behavior when constructing clients


anuraaga:
We've added some client creation shortcuts in #2396 but I wonder if it hasn't resulted in some spaghetti.

I feel that this code doesn't produce the intended result

```java
ClientFactory factory = ClientFactory.builder().useHttp2Preface(false).build();
ClientOptions options = ClientOptions.builder().responseTimeoutMillis(timeout).build();  // Ends up being essentially equivalent to default client options, but I thought I'm only specifying response timeout
WebClient.builder(...)
  .factory(factory) // Want to use custom factory
  .options(options) // But custom factory gets overridden by ClientFactory.ofDefault(), which is highly surprising
  .build();
```

Probably related to #2483 (though I found it coincidentally, heh)

anuraaga:
I think we should remove factory from `WebClientBuilder` if we want to keep it on `ClientOptions`. It's too hard to grok this relationship otherwise.

trustin:
Such pitfall exists in all other properties in the builder. The reason it became a problem is because `factory` was not an option before. If `useHttp2Preface` was not an option before, a similar problem would occur as well. How about just updating release note to add some warning about this?

anuraaga:
I generally use defaults so haven't run into this in much code but did in Zipkin and I don't see a reason users won't continue to run into it.

Perhaps `ClientOptions.Builder` should initialize as empty, not defaults? The code snippet above is too surprising to just leave to javadoc or release notes IMO.

trustin:
I understand your concern and I agree that the problem stems from the idea that all option properties in `AbstractOptions` must be present at construction time. Let's fix this before 1.0.

anuraaga:
Also, I find it unintuitive understanding what is a property of a subclass, safe from overriding from clientoptions (path, etc) and what isn't. I guess this may be an argument for composition vs inheritance. But as long as the above surprising snippet gets fixed I'm ok.

trustin:
> Also, I find it unintuitive understanding what is a property of a subclass, safe from overriding from clientoptions (path, etc) and what isn't. 

Could you elaborate on this?

anuraaga:
Currently, when using the builder, I have to think about whether a method on the builder is safe to call before `options`. This is because they all exist on the class but only some are overridden by options.

But yeah if we change the behavior so calling options doesn't completely clobber, it should be ok.

trustin:
Ah, OK. Thanks for explanation. Makes sense.

PR #2516, opened 2020-02-19
Exclude default options when merging two `Client*Options`

ikhoon:
Motivation:
After Armeria 0.98.0, `WebClientBuilder` and `ClientBuilder` can set `ClientOptions` and `ClientFactory` easily.
Unfortunately, the `factory` set to the client gets overridden by `ClientOption.FACTORY` of `options`
even though a user did not specify the factory in `ClientOption`.

```java
WebClient.builder()
         .fatory(factory) // Set custom factory
         .options(options) // The custom factory gets overridden by ClientFactory.ofDefault()
         .build();
Clients.builder(uri)
       .factory(factory)
       .options(options)
       .build();
```
Becuase the `ClientOption` has `ClientOption.FACTORY` by default.
It confuses users who use the client builders.

Modifications:
* Do not set `DEFAULT` options when building `Client*Options`
* Make `AbstractOptions` implement `Iterable`
* `Client*Options.get*` has a priority in order of own and `DEFAULT`
* Breaking
  * `asMap()` in `ClientOptions` and `ClientFactoryOptions` does not return default options anymore.
  * `ClientOption.DEFAULT` has been removed in favor of `ClientOption.of()`

Result:
`AbstractOption`s set by users don't get overridden by default options.
Fixes: #2493

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
@@ -60,12 +60,7 @@ public class AbstractClientOptionsBuilder {
      */
     public AbstractClientOptionsBuilder options(ClientOptions options) {
         requireNonNull(options, "options");
-        for (ClientOption<Object> option : options.options()) {
-            final Object value = options.getOrNull(option, false);
-            if (value != null) {
-                option(option, value);
-            }
-        }
+        options.forEach(this::option);
         return this;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
@@ -60,7 +60,7 @@ public class AbstractClientOptionsBuilder {
      */
     public AbstractClientOptionsBuilder options(ClientOptions options) {
         requireNonNull(options, "options");
-        for (ClientOption<Object> option: ClientOption.OPTIONS) {
+        for (ClientOption<Object> option : options.options()) {
             final Object value = options.getOrNull(option, false);
             if (value != null) {
                 option(option, value);
--- a/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
@@ -60,7 +60,12 @@ public class AbstractClientOptionsBuilder {
      */
     public AbstractClientOptionsBuilder options(ClientOptions options) {
         requireNonNull(options, "options");
-        options.asMap().values().forEach(this::option);
+        for (ClientOption<Object> option: ClientOption.OPTIONS) {
+            final Object value = options.getOrNull(option, false);
+            if (value != null) {
+                option(option, value);
+            }
+        }
         return this;
     }
 
--- a/core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
@@ -20,7 +20,6 @@ import static java.util.Objects.requireNonNull;
 
 import java.net.InetSocketAddress;
 import java.util.Collections;
-import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
@@ -186,9 +185,6 @@ public final class ClientFactoryOptions extends AbstractOptions {
         return channelOptions;
     }
 
-    @Nullable
-    private Map<ClientFactoryOption<?>, ClientFactoryOptionValue<?>> valueMap;
-
     private ClientFactoryOptions(ClientFactoryOptionValue<?>... options) {
         super(options);
     }
@@ -219,10 +215,6 @@ public final class ClientFactoryOptions extends AbstractOptions {
      * @throws NoSuchElementException if no value is set for the specified {@link ClientFactoryOption}.
      */
     public <T> T get(ClientFactoryOption<T> option) {
-        final T ownValue = getOwnValue(option);
-        if (ownValue != null) {
-            return ownValue;
-        }
         return get(this, DEFAULT, option);
     }
 
@@ -234,20 +226,11 @@ public final class ClientFactoryOptions extends AbstractOptions {
      */
     @Nullable
     public <T> T getOrNull(ClientFactoryOption<T> option) {
-        final T ownValue = getOwnValue(option);
-        if (ownValue != null) {
-            return ownValue;
-        }
         return getOrNull(this, DEFAULT, option);
     }
 
     @Nullable
     <T> T getOrNull(ClientFactoryOption<T> option, boolean includeDefault) {
-        final T ownValue = getOwnValue(option);
-        if (ownValue != null) {
-            return ownValue;
-        }
-
         if (includeDefault) {
             return getOrNull(this, DEFAULT, option);
         } else {
@@ -263,25 +246,7 @@ public final class ClientFactoryOptions extends AbstractOptions {
      */
     public <T> T getOrElse(ClientFactoryOption<T> option, T defaultValue) {
         requireNonNull(defaultValue, "defaultValue");
-        final T ownValue = getOwnValue(option);
-        if (ownValue != null) {
-            return ownValue;
-        }
         return getOrElse(this, DEFAULT, option, defaultValue);
-    }
-
-    @SuppressWarnings("unchecked")
-    @Nullable
-    private <T> T getOwnValue(ClientFactoryOption<T> option) {
-        if (valueMap == null) {
-            return null;
-        }
-
-        final ClientFactoryOptionValue<?> optionValue = valueMap.get(option);
-        if (optionValue != null) {
-            return (T) optionValue.value();
-        }
-        return null;
     }
 
     /**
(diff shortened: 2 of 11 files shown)


core/src/main/java/com/linecorp/armeria/client/AbstractClientOptionsBuilder.java
```diff
@@ -60,7 +60,12 @@ protected AbstractClientOptionsBuilder(ClientOptions options) {
      */
     public AbstractClientOptionsBuilder options(ClientOptions options) {
         requireNonNull(options, "options");
-        options.asMap().values().forEach(this::option);
+        for (ClientOption<Object> option : options.options()) {
+            final Object value = options.getOrNull(option, false);
```
trustin: This is not very efficient because we have to do a lookup for each option. How about making `AbstractOptions` implement `Iterable<AbstractOptionValue>`?

    public abstract class AbstractOptions<T extends AbstractOptionValue<?, ?>>
            implements Iterable<T>

ikhoon: That sounds good. 👍

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -99,13 +99,16 @@
...
     };
 
-    private static final ClientFactoryOptions DEFAULT = new ClientFactoryOptions(DEFAULT_OPTIONS);
+    @VisibleForTesting
+    static final ClientFactoryOptions DEFAULT = new ClientFactoryOptions(DEFAULT_OPTIONS);
+
+    private static final ClientFactoryOptions EMPTY = new ClientFactoryOptions();
 
     /**
      * The default {@link ClientFactoryOptions}.
```
trustin: `Returns an empty singleton {@link ClientFactoryOptions}.`

core/src/main/java/com/linecorp/armeria/common/util/AbstractOptions.java
```diff
@@ -19,31 +19,60 @@
...
 /**
  * A set of configuration options and their respective values.
  *
  * @see AbstractOption
  * @see AbstractOptionValue
  */
-public abstract class AbstractOptions {
+public abstract class AbstractOptions<T extends AbstractOptionValue<?, ?>> implements Iterable<T> {
+
+    protected static <V> V get(AbstractOptions<?> first, AbstractOptions<?> second, AbstractOption<V> option) {
```
trustin: - Could we make sure `first` and `second` have the same type parameter?
- Missing Javadoc

core/src/main/java/com/linecorp/armeria/common/util/AbstractOptions.java
```diff
@@ -188,37 +201,14 @@ protected AbstractOptions(AbstractOptions baseOptions, AbstractOptions additiona
...
+        return Iterators.unmodifiableIterator(valueMap.values().iterator());
     }
 
     @Override
     public String toString() {
-        return toString(asMap0().values());
+        return toString(valueMap.values());
     }
 
     static String toString(Collection<?> values) {
```
trustin: Not necessarily part of this PR, but how about rewriting this method with `MoreObjects.toStringHelper`?

core/src/test/java/com/linecorp/armeria/client/ClientFactoryOptionsTest.java
```diff
@@ -0,0 +1,133 @@
...
+    @Test
+    void allDefaultOptionsArePresent() throws Exception {
+        @SuppressWarnings("rawtypes")
+        final Set<ClientFactoryOption> options = getAllPublicStaticFinal(ClientFactoryOption.class);
+        final Set<ClientFactoryOption<?>> defaults = Streams.stream(ClientFactoryOptions.DEFAULT)
+                                                            .map(AbstractOptionValue::option)
+                                                            .collect(toImmutableSet());
+
+        assertThat(defaults).isEqualTo(options);
+        assertThat(Iterables.size(ClientFactoryOptions.of())).isZero();
```
trustin: Could use: https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractIterableAssert.html#isEmpty() ?

core/src/test/java/com/linecorp/armeria/client/WebClientBuilderTest.java
```diff
@@ -0,0 +1,87 @@
...
+        assertThat(clientOptions.get(ClientOption.RESPONSE_TIMEOUT_MILLIS)).isEqualTo(200);
+        final ClientFactory clientFactory = clientOptions.factory();
+        assertThat(clientFactory.options().get(ClientFactoryOption.HTTP1_MAX_CHUNK_SIZE)).isEqualTo(100);
+    }
+
+    @Test
+    void keepLastFactory_by_options() {
+        final ClientFactory optionClientFactory = ClientFactory.builder()
+                                                    .option(ClientFactoryOption.HTTP1_MAX_CHUNK_SIZE, 200)
+                                                    .build();
```
trustin: Indentation

core/src/test/java/com/linecorp/armeria/client/WebClientBuilderTest.java
```diff
@@ -0,0 +1,87 @@
...
+        final ClientOptions options = ClientOptions.of(ClientOption.RESPONSE_TIMEOUT_MILLIS.newValue(200L),
+                                                       ClientOption.FACTORY.newValue(optionClientFactory));
+
+        final ClientFactory factory = ClientFactory.builder()
+                                                    .option(ClientFactoryOption.HTTP1_MAX_CHUNK_SIZE, 100)
+                                                    .build();
```
trustin: Indentation

core/src/test/java/com/linecorp/armeria/client/WebClientBuilderTest.java
```diff
@@ -0,0 +1,87 @@
...
+        assertThat(clientOptions.get(ClientOption.RESPONSE_TIMEOUT_MILLIS)).isEqualTo(200);
+        final ClientFactory clientFactory = clientOptions.factory();
+        assertThat(clientFactory.options().get(ClientFactoryOption.HTTP1_MAX_CHUNK_SIZE)).isEqualTo(200);
+    }
+
+    @Test
+    void keepLastFactory_by_factory() {
+        final ClientFactory optionClientFactory = ClientFactory.builder()
+                                                    .option(ClientFactoryOption.HTTP1_MAX_CHUNK_SIZE, 200)
+                                                    .build();
```
trustin: Indentation

core/src/test/java/com/linecorp/armeria/client/WebClientBuilderTest.java
```diff
@@ -0,0 +1,87 @@
...
+        final ClientOptions options = ClientOptions.of(ClientOption.RESPONSE_TIMEOUT_MILLIS.newValue(200L),
+                                                       ClientOption.FACTORY.newValue(optionClientFactory));
+
+        final ClientFactory factory = ClientFactory.builder()
+                                                    .option(ClientFactoryOption.HTTP1_MAX_CHUNK_SIZE, 100)
+                                                    .build();
```
trustin: Indentation

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
```diff
@@ -223,7 +226,16 @@ private ClientFactoryOptions(ClientFactoryOptions baseOptions,
      */
     @Nullable
     public <T> T getOrNull(ClientFactoryOption<T> option) {
-        return getOrNull0(option);
+        return getOrNull(this, DEFAULT, option);
+    }
+
+    @Nullable
+    <T> T getOrNull(ClientFactoryOption<T> option, boolean includeDefault) {
```
trustin: Unused?

core/src/main/java/com/linecorp/armeria/client/ClientOptions.java
```diff
@@ -262,7 +254,16 @@ private ClientOptions(ClientOptions baseOptions, ClientOptions additionalOptions
      */
     @Nullable
     public <T> T getOrNull(ClientOption<T> option) {
-        return getOrNull0(option);
+        return getOrNull(this, DEFAULT, option);
+    }
+
+    @Nullable
+    <T> T getOrNull(ClientOption<T> option, boolean includeDefault) {
```
trustin: Unused?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryOptions.java
ikhoon: Oops..

core/src/test/java/com/linecorp/armeria/client/ClientFactoryOptionsTest.java
ikhoon: Oh, Nice tip!

ikhoon:
The PR description is updated.

trustin:
Seems like `asMap()` methods are gone? How about resurrecting them for backward compatibility?

ikhoon:
> Seems like asMap() methods are gone? How about resurrecting them for backward compatibility?

My considering point is that we should include `DEFAULT` or just returns own options as map. I wonder other folks' opinions. 😀

trustin:
Own options only? Because iterator() is implemented that way

anuraaga:
Just a thought, what about doing the defaulting of options at the option layer itself, rather than group of options layer?

It seemed like it might be nice if instead of having a group of defaults here

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/ClientOptions.java#L78

each option has its default, defined here

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/client/ClientOption.java#L80

Then the options look a lot like our `Flags` and reduce some coupling between option values and a group of options.

trustin:
@anuraaga That sounds good to me. I have some idea for improving how options work. Let me make the suggested change in that PR.

trustin:
Thanks, @ikhoon !

PR #2517, opened 2020-02-20
Do not propagate the first child's `requestCause()` for nested logs

trustin:
Motivation:

With `RetryingClient` involved, a `RequestLog` can have child
`RequestLog`s. Currently, the parent `RequestLog`'s request-side
properties are fetched from the first child, and response-side
properties from the last child. See: https://line.github.io/armeria/advanced-structured-logging.html#nested-log

However, if the first child request fails with an
`UnprocessedRequestException` (e.g. due to a connection refused error),
the first child's `requestCause()`, which is non-null, is propagated to
the parent log. As a result, even if the request eventually succeeded
thanks to retries, the parent log's `requestCause()` will be non-null,
making `MetricCollectingClient` consider the request as a failure.

Modifications:

- Always use `null` as a parent `requestCause()`.
- Miscellaneous:
  - Fixed a bug where `DefaultRequestLog.endRequest()` does not set
    `requestHeaders` properly.

Result:

- `MetricCollectingClient` correctly records the metrics of the request
  which succeeded after a retry due to a 'connection refused' error.
- `RequestLog.requestHeaders()` has some headers set even for the
  requests that failed early, i.e. without writing anything.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithMetricsTest.java
+++ b/core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithMetricsTest.java
@@ -127,7 +127,7 @@ class RetryingClientWithMetricsTest {
     @Test
     void retryingThenMetricCollectingWithConnectionRefused() throws Exception {
         // The first request will fail with an UnprocessedException and
-        // the second request will succeed with 404.
+        // the second request will succeed with 200.
         final EndpointGroup group = EndpointGroup.of(Endpoint.of("127.0.0.1", 1),
                                                      server.httpEndpoint());
         final WebClient client = WebClient.builder(SessionProtocol.HTTP, group)
@@ -170,7 +170,7 @@ class RetryingClientWithMetricsTest {
     @Test
     public void metricCollectingThenRetryingWithConnectionRefused() throws Exception {
         // The first request will fail with an UnprocessedException and
-        // the second request will succeed with 404.
+        // the second request will succeed with 200.
         final EndpointGroup group = EndpointGroup.of(Endpoint.of("127.0.0.1", 1),
                                                      server.httpEndpoint());
         final WebClient client =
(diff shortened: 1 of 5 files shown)


core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithMetricsTest.java
```diff
@@ -0,0 +1,192 @@
...
+                    .containsEntry("foo.requests#count{http.status=200,method=GET,result=failure}", 0.0)
+                    .containsEntry("foo.requests#count{http.status=500,method=GET,result=success}", 0.0)
+                    .containsEntry("foo.requests#count{http.status=500,method=GET,result=failure}", 2.0);
+        });
+    }
+
+    @Test
+    void retryingThenMetricCollectingWithConnectionRefused() throws Exception {
+        // The first request will fail with an UnprocessedException and
+        // the second request will succeed with 404.
```
ikhoon: nit: Maybe `succeed with 200`?

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithMetricsTest.java
```diff
@@ -0,0 +1,192 @@
...
+            assertThat(MoreMeters.measureAll(meterRegistry))
+                    .containsEntry("foo.requests#count{http.status=200,method=GET,result=success}", 1.0)
+                    .containsEntry("foo.requests#count{http.status=200,method=GET,result=failure}", 0.0);
+        });
+    }
+
+    @Test
+    public void metricCollectingThenRetryingWithConnectionRefused() throws Exception {
+        // The first request will fail with an UnprocessedException and
+        // the second request will succeed with 404.
```
ikhoon: nit: Maybe `succeed with 200`?

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithMetricsTest.java
trustin: Oops! :sweat_smile:

trustin:
Thanks for reviewing, @ikhoon

trustin:
Note: This issue was reported by @andrey-tpt via our Slack workspace. Thanks, @andrey-tpt!

PR #2520, opened 2020-02-21
Fix a bug where `RequestContextExporter` exports wrong properties

### LINKED ISSUE #197: DocService: Handle '@example' tag in Thrift IDL
A comment block in a Thrift IDL could contain an `@example` tag which shows some example TTEXT requests a user can play with:

```
/**
 * This operation does this.
 * ...
 * @example
 * {
 *   "paramName": "paramValue",
 *   "paramName2": 42,
 *   ...
 * }
 * @example { .. second example ... }
 */
```

The examples could be shown in the debug form of the `DocService`.

trustin:
Motivation:

- `RequestContextExporter.state()` does not consider the case where the
  current context has a root context. As a result, a `ClientRequestContext`
  may reuse the state of its root context.
- `RequestContextExporter` does not take custom attributes into account
  when determining if the properties needs to be re-exported. As a
  result, custom attributes will not be re-exported even if a user
  changes any of their values.

Modifications:

- Use `ClientRequestContext.ownAttr()` to check whether the state object
  exists.
- Compare attribute values as well when determining if properties need
  to be re-exported.
- Handle the case where attribute or header disappears.
- Miscellaneous:
  - Implemented `toString()` for:
    - `DefaultAttributeMap`
    - `FakeChannel` in `AbstractRequestContextBuilder`
  - Fixed `StackOverflowError` in `RouteDecoratingService.toString()`

Result:

- When using `RequestContextExporter` or `RequestContextExportingAppender`,
  the properties are exported properly.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
@@ -241,26 +241,27 @@ public final class RequestContextExporter {
 
         final State state = state(ctx);
         final RequestLogAccess log = ctx.log();
-
-        boolean needsUpdate;
+        boolean needsUpdate = false;
 
         // Needs to update if availabilityStamp has changed.
+        // Also updates `State.availabilityStamp` while checking.
         final int availabilityStamp = log.availabilityStamp();
         if (state.availabilityStamp != availabilityStamp) {
             state.availabilityStamp = availabilityStamp;
             needsUpdate = true;
-        } else {
-            needsUpdate = false;
         }
 
         // Needs to update if any attributes have changed.
+        // Also updates `State.attrValues` while scanning.
         if (attrs != null) {
             assert state.attrValues != null;
             for (int i = 0; i < attrs.length; i++) {
-                final Object oldValue = state.attrValues[i];
                 final Object newValue = ctx.attr(attrs[i].key);
                 state.attrValues[i] = newValue;
-                needsUpdate |= !Objects.equals(oldValue, newValue);
+                if (!needsUpdate) {
+                    final Object oldValue = state.attrValues[i];
+                    needsUpdate = !Objects.equals(oldValue, newValue);
+                }
             }
         }
 
--- a/core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
@@ -275,68 +275,68 @@ public final class RequestContextExporter {
         return state.clone();
     }
 
-    private void export(Map<String, String> out, RequestContext ctx, RequestLog log) {
+    private void export(State state, RequestContext ctx, RequestLog log) {
         // Built-ins
         if (builtInProperties.containsAddresses()) {
-            exportAddresses(out, ctx);
+            exportAddresses(state, ctx);
         }
         if (builtInProperties.contains(SCHEME)) {
-            exportScheme(out, ctx, log);
+            exportScheme(state, ctx, log);
         }
         if (builtInProperties.contains(REQ_DIRECTION)) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
```diff
@@ -235,15 +240,37 @@ public boolean containsBuiltIn(BuiltInProperty property) {
...
+        final int availabilityStamp = log.availabilityStamp();
+        if (state.availabilityStamp != availabilityStamp) {
+            state.availabilityStamp = availabilityStamp;
+            needsUpdate = true;
+        } else {
+            needsUpdate = false;
+        }
+
+        // Needs to update if any attributes have changed.
+        if (attrs != null) {
```
anuraaga: `!needsUpdate && `

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
```diff
@@ -235,15 +240,37 @@ public boolean containsBuiltIn(BuiltInProperty property) {
...
+            assert state.attrValues != null;
+            for (int i = 0; i < attrs.length; i++) {
+                final Object oldValue = state.attrValues[i];
+                final Object newValue = ctx.attr(attrs[i].key);
+                state.attrValues[i] = newValue;
+                needsUpdate |= !Objects.equals(oldValue, newValue);
```
anuraaga:
```
if (!equals) {
  needsUpdate = true;
  break;
}
```

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
```diff
@@ -235,15 +240,37 @@ public boolean containsBuiltIn(BuiltInProperty property) {
...
 
         final State state = state(ctx);
-        final RequestLog log = ctx.log().partial();
-        final int availabilities = log.availabilityStamp();
-        if (availabilities != state.availabilities) {
-            state.availabilities = availabilities;
-            export(state, ctx, log);
+        final RequestLogAccess log = ctx.log();
+
+        boolean needsUpdate;
```
anuraaga: I guess easier to read if this is `false` and remove below `else`.

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
trustin: Let me make it clear that this loop also updates `state.attrValues` while scanning.

anuraaga: Ah :)

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
trustin: Can't break out of the loop because it needs to update `state.attrValues` with the new values.

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
```diff
@@ -537,54 +565,60 @@ private void exportRpcResponse(Map<String, String> out, RequestLog log) {
...
-    private void exportAttributes(Map<String, String> out, RequestContext ctx) {
+    private void exportAttributes(State state) {
         if (attrs == null) {
             return;
         }
 
-        for (ExportEntry<AttributeKey<?>> e : attrs) {
-            putOtherProperty(out, e, ctx.attr(e.key));
+        assert state.attrValues != null;
+        for (int i = 0; i < attrs.length; i++) {
```
ikhoon: nit 
```suggestion
        for (int i = 0; i < numAttrs; i++) {
```

trustin:
Thanks for reviewing, folks.

Issue #2513, opened 2020-02-19
`RequestContextExporter` does not work with `AttributeKey` created with `newInstance()`


trustin:
Reported by @andrewoma:

> it seems you have to create your key in the form of
> ```
> AttributeKey.valueOf(Context.class, "FOO")
> ```
> The `AttributeKey.newInstance(Context.class)` form isn't supported and that took me quite some debugging to realise

(in `RequestContextExporter`)

trustin:
Another problem which could be resolved together:

> Another gotcha of the custom attributes - you can't map the same attribute twice with different string functions, e.g.
> ```xml
> <export>attrs.request_id:com.example.Context#KEY:com.example.ContextRequestIdString</export>
> <export>attrs.trace_id:com.example.Context#KEY:com.example.ContextTraceIdString</export>
> ```
> Only the `attrs.request_id` is output

ikhoon:
Let me take a look. 😀

ikhoon:
>> The `AttributeKey.newInstance(Context.class)` form isn't supported and that took me quite some debugging to realise

Had chat with  @andrewoma, ans the exact API usage was `AttributeKey.newInstance(Context.class.getName())`

trustin:
`newInstance()` fails when there is an `AttributeKey` with the same name. When you use `RCEA`, it will use `AttributeKey.valueOf()` to instantiate an `AttributeKey`, which may happen before a user's `newInstance()` call. As a result, `newInstance()` will trigger an `IllegalStateException` and thus a user will see an unexpected behavior.

The workaround for this issue is for the user to use `valueOf()` rather than `newInstance()`.

trustin:
[This issue](https://github.com/line/armeria/issues/2513#issuecomment-588593453) has been resolved in #2521, though. Thanks for reporting, @andrewoma! :bowing_man:

PR #2521, opened 2020-02-21
Allow `RequestContextExporter` to export an `AttributeKey` with diffe…

ikhoon:
…rent aliases

Motivation:
When a user tries to export an attribute with different aliases, the later one will be ignored.
For example:
```xml
<export>attrs.request_id:com.example.Context#KEY:com.example.ContextRequestIdString</export>
<export>attrs.trace_id:com.example.Context#KEY:com.example.ContextTraceIdString</export>
```
The `attrs.trace_id` is ignored and only the `attrs.request_id` is output.
Because the `ExportEntry` overrides `equals` that only compares `ExportEntry.key`. 
When it is added to `RequestContextExporterBuilder.attrs`, the later `ExportEntry` is ignored.

Modifications:
- Change `ExportEntry.equals` to compare `key` and `exportKey`

Result:
You can now export an `Attribute` with different aliases.
Fixes: The second issue in #2513

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
+++ b/core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporter.java
@@ -613,7 +613,7 @@ public final class RequestContextExporter {
 
         @Override
         public int hashCode() {
-            return key.hashCode();
+            return key.hashCode() * 31 + exportKey.hashCode();
         }
 
         @Override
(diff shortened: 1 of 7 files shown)


ikhoon:
> Thanks! It could have been much easier to test at RequestContextExporterTest level, though.

Ah, That's a good idea.👍 Added more tests.

PR #2528, opened 2020-02-24
Push the current context when logging requests and responses

trustin:
Motivation:

When `Logging{Client,Service}` logs requests and responses via
`LoggingDecorators`, the context must be pushed so that
`RequestContextExportingAppender` picks up the context and export the
properties to the logging framework's MDC.

Modifications:

- Push the current context before logging anything in
  `LoggingDecorators`.
- Updated `Logging{Client,Service}Test` to:
  - Test if the context is pushed.
  - Test without using mocks excessively.

Result:

- `Logging{Client,Service}` now works well with
  `RequestContextExportingAppender` or any other logging subsystem that
  relies on thread-local variables.


trustin:
This should fix the problem reported by @andrewoma at our Slack workspace. Thanks again for reporting. :heart:

PR #2530, opened 2020-02-24
Make `HttpTimestampSupplier` work with negative clock value

trustin:
Motivation:

The initial value of `nextUpdateNanos` is `0`, which can make
`HttpTimestampSupplier` generate an empty string if `System.nanoTime()`
returns a negative value.

Modifivations:

- Initialize `nextUpdateNanos` with the current `System.nanoTime()`
  value.
- Improve testability of `HttpTimestampSupplier` so that we do not need
  to busy-loop while testing.

Result:

- `HttpTimestampSupplier` works even when `System.nanoTime()` returns a
  negative value.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/internal/common/util/HttpTimestampSupplierTest.java
+++ b/core/src/test/java/com/linecorp/armeria/internal/common/util/HttpTimestampSupplierTest.java
@@ -17,43 +17,79 @@
 package com.linecorp.armeria.internal.common.util;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.awaitility.Awaitility.await;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.clearInvocations;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/linecorp/armeria/internal/common/util/HttpTimestampSupplierTest.java
```diff
@@ -17,43 +17,79 @@
...
+    @Test
+    void normal() {
+        final AtomicReference<Instant> instantHolder = new AtomicReference<>(TIME0);
+        final AtomicLong nanoTimeHolder = new AtomicLong(TimeUnit.MILLISECONDS.toNanos(-500));
 
-    private HttpTimestampSupplier supplier;
+        // Create delegating mock suppliers.
+        // Note that we can't use method references or lambda expressions because they generate final classes.
+        @SuppressWarnings({ "Convert2Lambda", "Anonymous2MethodRef" })
+        final Supplier<Instant> instantSupplier = spy(new Supplier<Instant>() {
```
anuraaga: Why use spy instead of mock?

trustin: No idea. Now using mock.

PR #2533, opened 2020-02-26
Fix Subscriber not to throw an exception

### LINKED ISSUE #2475: Handle exceptions thrown by `Subscriber`
An exception raised by a `Subscriber`'s `onNext()` or other handler methods can cause an unexpected connection drop. The following stack trace shows the case where an exception raised by `onNext()` is propagated to `Http2ResponseDecoder` which triggers an HTTP/2 connection error:

```
HTTP/2 connection error:
io.netty.handler.codec.http2.Http2Exception: failed to consume a HEADERS frame
	at io.netty.handler.codec.http2.Http2Exception.connectionError(Http2Exception.java:117)
	at com.linecorp.armeria.client.Http2ResponseDecoder.onHeadersRead(Http2ResponseDecoder.java:200)
	at com.linecorp.armeria.client.Http2ResponseDecoder.onHeadersRead(Http2ResponseDecoder.java:213)
	at io.netty.handler.codec.http2.Http2FrameListenerDecorator.onHeadersRead(Http2FrameListenerDecorator.java:48)
	at io.netty.handler.codec.http2.Http2EmptyDataFrameListener.onHeadersRead(Http2EmptyDataFrameListener.java:63)
	at io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder$FrameReadListener.onHeadersRead(DefaultHttp2ConnectionDecoder.java:373)
	at io.netty.handler.codec.http2.DefaultHttp2FrameReader$1.processFragment(DefaultHttp2FrameReader.java:457)
	at io.netty.handler.codec.http2.DefaultHttp2FrameReader.readHeadersFrame(DefaultHttp2FrameReader.java:464)
	at io.netty.handler.codec.http2.DefaultHttp2FrameReader.processPayloadState(DefaultHttp2FrameReader.java:254)
	at io.netty.handler.codec.http2.DefaultHttp2FrameReader.readFrame(DefaultHttp2FrameReader.java:160)
	at io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder.decodeFrame(DefaultHttp2ConnectionDecoder.java:174)
	at io.netty.handler.codec.http2.DecoratingHttp2ConnectionDecoder.decodeFrame(DecoratingHttp2ConnectionDecoder.java:63)
	at io.netty.handler.codec.http2.Http2ConnectionHandler$FrameDecoder.decode(Http2ConnectionHandler.java:378)
	at io.netty.handler.codec.http2.Http2ConnectionHandler.decode(Http2ConnectionHandler.java:438)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:498)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:437)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:377)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:363)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:355)
	at io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:227)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:377)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:363)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:355)
	at io.netty.handler.flush.FlushConsolidationHandler.channelRead(FlushConsolidationHandler.java:152)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:377)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:363)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:355)
	at io.netty.handler.timeout.IdleStateHandler.channelRead(IdleStateHandler.java:286)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:377)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:363)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:355)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:377)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:363)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
	at io.netty.channel.epoll.AbstractEpollStreamChannel$EpollStreamUnsafe.epollInReady(AbstractEpollStreamChannel.java:792)
	at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:475)
	at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:378)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at java.base/java.lang.Thread.run(Thread.java:830)
Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 2
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3720)
	at java.base/java.lang.String.substring(String.java:1909)
	at com.linecorp.armeria.client.endpoint.healthcheck.HttpHealthChecker$HealthCheckResponseSubscriber.updateLongPollingSettings(HttpHealthChecker.java:250)
	at com.linecorp.armeria.client.endpoint.healthcheck.HttpHealthChecker$HealthCheckResponseSubscriber.onNext(HttpHealthChecker.java:189)
	at com.linecorp.armeria.client.endpoint.healthcheck.HttpHealthChecker$HealthCheckResponseSubscriber.onNext(HttpHealthChecker.java:149)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.notifySubscriberWithElements(DefaultStreamMessage.java:344)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.notifySubscriber0(DefaultStreamMessage.java:322)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.notifySubscriber(DefaultStreamMessage.java:255)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.addObjectOrEvent(DefaultStreamMessage.java:241)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.addObject(DefaultStreamMessage.java:161)
	at com.linecorp.armeria.common.stream.AbstractStreamMessageAndWriter.tryWrite(AbstractStreamMessageAndWriter.java:74)
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.tryWrite(DefaultStreamMessage.java:65)
	at com.linecorp.armeria.client.DecodedHttpResponse.tryWrite(DecodedHttpResponse.java:55)
	at com.linecorp.armeria.client.HttpResponseDecoder$HttpResponseWrapper.tryWrite(HttpResponseDecoder.java:227)
	at com.linecorp.armeria.client.Http2ResponseDecoder.onHeadersRead(Http2ResponseDecoder.java:197)
	... 41 common frames omitted
```

We must catch any exceptions raised by `Subscriber` and abort the `StreamMessage` with the caught exception.

minwoox:
Motivation:
According to https://github.com/reactive-streams/reactive-streams-jvm#2.13,
a subscriber must not throw an exception but just call `Subscription.cancel()`.
Some of our subscribers violated this rule so we should fix them.

Modifications:
- Call `Subscription.cancel()` instead of throwing an exception in `Subscriber`s

Result:
- `Subscriber` does not throw an exception anymore.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
@@ -236,8 +236,9 @@ final class HttpRequestSubscriber implements Subscriber<HttpObject>, ChannelFutu
     @Override
     public void onNext(HttpObject o) {
         if (!(o instanceof HttpData) && !(o instanceof HttpHeaders)) {
-            throw newIllegalStateException(
-                    "published an HttpObject that's neither Http2Headers nor Http2Data: " + o);
+            fail(new IllegalArgumentException(
+                    "published an HttpObject that's neither Http2Headers nor Http2Data: " + o));
+            return;
         }
 
         boolean endOfStream = o.isEndOfStream();
@@ -246,7 +247,8 @@ final class HttpRequestSubscriber implements Subscriber<HttpObject>, ChannelFutu
                 if (o instanceof HttpHeaders) {
                     final HttpHeaders trailers = (HttpHeaders) o;
                     if (trailers.contains(HttpHeaderNames.STATUS)) {
-                        throw newIllegalStateException("published a trailers with status: " + o);
+                        fail(new IllegalArgumentException("published a trailers with status: " + o));
+                        return;
                     }
                     // Trailers always end the stream even if not explicitly set.
                     endOfStream = true;
@@ -373,11 +375,5 @@ final class HttpRequestSubscriber implements Subscriber<HttpObject>, ChannelFutu
 
         this.timeoutFuture = null;
         return timeoutFuture.cancel(false);
-    }
-
-    private IllegalStateException newIllegalStateException(String msg) {
-        final IllegalStateException cause = new IllegalStateException(msg);
-        fail(cause);
-        return cause;
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
@@ -174,6 +174,10 @@ final class HttpHealthChecker implements AsyncCloseable {
 
         @Override
         public void onNext(HttpObject obj) {
+            // There's no chance that an exception is raised from the underlying logic, so we don't do
+            // try catch not to propagate the exception to the publisher.
+            // https://github.com/reactive-streams/reactive-streams-jvm#2.13
+
             if (closeable.isClosing()) {
                 subscription.cancel();
                 return;
--- a/core/src/main/java/com/linecorp/armeria/common/stream/NeverInvokedSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/NeverInvokedSubscriber.java
@@ -18,12 +18,8 @@ package com.linecorp.armeria.common.stream;
 
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 final class NeverInvokedSubscriber<T> implements Subscriber<T> {
-
-    private static final Logger logger = LoggerFactory.getLogger(NeverInvokedSubscriber.class);
 
     private static final NeverInvokedSubscriber<Object> INSTANCE = new NeverInvokedSubscriber<>();
 
@@ -34,22 +30,21 @@ final class NeverInvokedSubscriber<T> implements Subscriber<T> {
 
     @Override
     public void onSubscribe(Subscription s) {
-        s.cancel();
-        logger.warn("onSubscribe({}) is invoked.", s);
+        throw new IllegalStateException("onSubscribe(" + s + ')');
     }
 
     @Override
     public void onNext(T t) {
-        logger.warn("onNext({}) is invoked.", t);
+        throw new IllegalStateException("onNext(" + t + ')');
     }
 
     @Override
     public void onError(Throwable t) {
-        logger.warn("onError() is invoked with:", t);
+        throw new IllegalStateException("onError(" + t + ')', t);
     }
 
     @Override
     public void onComplete() {
-        logger.warn("onComplete() is invoked.");
+        throw new IllegalStateException("onComplete()");
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/common/stream/NeverInvokedSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/NeverInvokedSubscriber.java
@@ -18,8 +18,12 @@ package com.linecorp.armeria.common.stream;
 
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 final class NeverInvokedSubscriber<T> implements Subscriber<T> {
+
+    private static final Logger logger = LoggerFactory.getLogger(NeverInvokedSubscriber.class);
 
     private static final NeverInvokedSubscriber<Object> INSTANCE = new NeverInvokedSubscriber<>();
 
@@ -30,21 +34,22 @@ final class NeverInvokedSubscriber<T> implements Subscriber<T> {
 
     @Override
     public void onSubscribe(Subscription s) {
-        throw new IllegalStateException("onSubscribe(" + s + ')');
+        s.cancel();
+        logger.warn("onSubscribe({}) is invoked.", s);
     }
 
     @Override
     public void onNext(T t) {
-        throw new IllegalStateException("onNext(" + t + ')');
+        logger.warn("onNext({}) is invoked.", t);
     }
 
     @Override
     public void onError(Throwable t) {
-        throw new IllegalStateException("onError(" + t + ')', t);
+        logger.warn("onError() is invoked with:", t);
     }
 
     @Override
     public void onComplete() {
-        throw new IllegalStateException("onComplete()");
+        logger.warn("onComplete() is invoked.");
     }
 }
--- a/core/src/main/java/com/linecorp/armeria/internal/server/ResponseConversionUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/server/ResponseConversionUtil.java
@@ -29,6 +29,8 @@ import javax.annotation.Nullable;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.HttpHeaders;
@@ -40,6 +42,8 @@ import com.linecorp.armeria.common.ResponseHeaders;
  * A utility class which helps to send a streaming {@link HttpResponse}.
  */
 public final class ResponseConversionUtil {
+
+    private static final Logger logger = LoggerFactory.getLogger(ResponseConversionUtil.class);
 
     /**
      * Returns a new {@link HttpResponseWriter} which has a content converted from the collected objects.
@@ -248,7 +252,9 @@ public final class ResponseConversionUtil {
                 return;
             }
             if (!trailers.isEmpty()) {
-                writer.write(trailers);
+                if (!writer.tryWrite(trailers)) {
+                    logger.warn("Failed to write a trailers: {}", trailers);
+                }
             }
             writer.close();
         }
--- a/core/src/main/java/com/linecorp/armeria/internal/server/ResponseConversionUtil.java
+++ b/core/src/main/java/com/linecorp/armeria/internal/server/ResponseConversionUtil.java
@@ -252,9 +252,7 @@ public final class ResponseConversionUtil {
                 return;
             }
             if (!trailers.isEmpty()) {
-                if (!writer.tryWrite(trailers)) {
-                    logger.warn("Failed to write a trailers: {}", trailers);
-                }
+                writer.tryWrite(trailers);
             }
             writer.close();
         }
--- a/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
@@ -62,9 +62,9 @@ final class HttpResponseSubscriber extends DefaultTimeoutController implements S
 
     private static final Logger logger = LoggerFactory.getLogger(HttpResponseSubscriber.class);
 
-    private static final AggregatedHttpResponse INTERNAL_SERVER_ERROR_MESSAGE =
+    private static final AggregatedHttpResponse internalServerErrorResponse =
             AggregatedHttpResponse.of(HttpStatus.INTERNAL_SERVER_ERROR);
-    private static final AggregatedHttpResponse SERVICE_UNAVAILABLE_MESSAGE =
+    private static final AggregatedHttpResponse serviceUnavailableResponse =
             AggregatedHttpResponse.of(HttpStatus.SERVICE_UNAVAILABLE);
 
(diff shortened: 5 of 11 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/NeverInvokedSubscriber.java
```diff
@@ -30,21 +34,22 @@
 
     @Override
     public void onSubscribe(Subscription s) {
-        throw new IllegalStateException("onSubscribe(" + s + ')');
+        s.cancel();
+        logger.warn("onSubscribe({}) is invoked.", s);
```
trustin: I would say none of these methods should be invoked. The exceptions that were thrown in this classes are like an assertion, i.e. better leaving as is?

core/src/main/java/com/linecorp/armeria/internal/server/ResponseConversionUtil.java
```diff
@@ -248,7 +252,9 @@ public void onComplete() {
                 return;
             }
             if (!trailers.isEmpty()) {
-                writer.write(trailers);
+                if (!writer.tryWrite(trailers)) {
+                    logger.warn("Failed to write a trailers: {}", trailers);
```
trustin: Probably better not logging, because a response can be closed by a remote peer at any time?

ikhoon:
I found 2 places where they throw `Exception` and could be called by `Subscriber.onNext(T)`. 
https://github.com/line/armeria/blob/e3d18350b6062cc6a6ed5bd27c16db2c67c7ead0/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessageDuplicator.java#L957-L959

https://github.com/line/armeria/blob/2a41b71a10151f18cd82c638e9770129353c91ee/spring/boot-webflux-autoconfigure/src/main/java/com/linecorp/armeria/spring/web/reactive/ArmeriaHttpClientResponseSubscriber.java#L129-L131

minwoox:
> I found 2 places where they throw Exception and could be called by Subscriber.onNext(T).

So both cases are handled. Right?

minwoox:
> > I found 2 places where they throw Exception and could be called by Subscriber.onNext(T).

As I talked to in the offline, both cases are handled. 😄

core/src/main/java/com/linecorp/armeria/common/stream/NeverInvokedSubscriber.java
minwoox: Sounds good. Let me revert this class. 😉

core/src/main/java/com/linecorp/armeria/internal/server/ResponseConversionUtil.java
minwoox: Removed that. Thanks!

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -246,7 +247,8 @@ public void onNext(HttpObject o) {
                 if (o instanceof HttpHeaders) {
                     final HttpHeaders trailers = (HttpHeaders) o;
                     if (trailers.contains(HttpHeaderNames.STATUS)) {
-                        throw newIllegalStateException("published a trailers with status: " + o);
+                        fail(new IllegalArgumentException("published a trailers with status: " + o));
```
trustin: I guess `IllegalStateException` is better in this class because the exception does not go to the caller.

core/src/main/java/com/linecorp/armeria/server/HttpResponseSubscriber.java
```diff
@@ -198,9 +200,10 @@ public void onNext(HttpObject o) {
                 if (o instanceof HttpHeaders) {
                     final HttpHeaders trailers = (HttpHeaders) o;
                     if (trailers.contains(HttpHeaderNames.STATUS)) {
-                        throw newIllegalStateException(
+                        failAndRespond(new IllegalArgumentException(
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
trustin: Never mind. It's good as it is.

minwoox: Had a chat with @trustin and I slightly prefer to leave this as it is because the user will see this exception from `RequestLog` even though the exception does not go to the caller right away.
But I'm not so strong on this. 😆 Any idea? @anuraaga, @ikhoon

minwoox: Okay then, let me just use `IllegalArgumentException`. 😉

core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HttpHealthChecker.java
```diff
@@ -174,6 +174,10 @@ public void onSubscribe(Subscription subscription) {
 
         @Override
         public void onNext(HttpObject obj) {
+            // There's no chance that an exception is raised from the underlying logic, so we don't do
```
anuraaga: These comments seem very noisy. Do we need them? I feel like just leaving it to a `SafeSubscriber` abstract class that always adds try / catch is better than adding these comments that reduce the readability of the code.

minwoox:
I was going to make a separate PR to handle exceptions thrown from the subscriber. But it just takes a little more time because I found that `DefaultStreamMessage` and `FixedStreamMessage` behave a little differently when abort is called. So I was fixing it first and then handle the exceptions.

https://github.com/reactive-streams/reactive-streams-jvm/issues/107
This is the thread that argues to include exceptions to the spec but the result is not as we do know now. I think it's worth reading through.

> These comments seem very noisy. Do we need them? I feel like just leaving it to a SafeSubscriber abstract class that always adds try / catch is better than adding these comments that reduce the readability of the code.

I left it to let the reviewer know that I actually search through all the subscribers. I will be happy to remove those comments after all review are done.
I'm not sure if it's a good idea to add `SafeSubscriber` because I think try catch should do by the publisher when an exception is raised to clean up the resources.

anuraaga:
Thanks for the link! This comment seems very important

https://github.com/reactive-streams/reactive-streams-jvm/issues/107#issuecomment-54142897

It seems like we need to augment `Subscriber` and `Publisher` in our public API with Armeria-specific types, which can allow exceptions if we want (probably do) and relax other requirements that hinder performance. It seems like reactor / rxjava does this.

https://projectreactor.io/docs/core/release/api/reactor/core/CoreSubscriber.html
http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/FlowableSubscriber.html

If `Subscriber` is not meant for users, than strict requirements like not allowing exceptions does make sense, and it's important for us to follow them when we use it as our primary API for subscribers. But as long as we expose `Subscriber` to users, it seems futile for us to have a pattern of "never throwing exceptions", as we've already found since we are leaving `NeverInvokedSubscriber` alone. It would be very unhelpful for a user to see a `SpecViolation("reactive streams 2.13 - don't throw exceptions")` then the actual exception they threw.

> I left it to let the reviewer know that I actually search through all the subscribers. I will be happy to remove those comments after all review are done.

Yeah would lean towards not having them, or maybe we can define an annotation that we put on `Subscriber` implementations like `NeverThrows` or something. Though as per above, I wonder if we should even be implementing `org.reactivestreams.Subscriber` and not have our own

minwoox:
> Yeah would lean towards not having them, or maybe we can define an annotation that we put on Subscriber implementations like NeverThrows or something. Though as per above, I wonder if we should even be implementing org.reactivestreams.Subscriber and not have our own

Having a marker annotation sounds good to me. I wonder they want to have it though because they stated it's illegal to throw an exception from a subscriber. 🤔 

> It seems like we need to augment Subscriber and Publisher in our public API with Armeria-specific types, which can allow exceptions if we want (probably do) and relax other requirements that hinder performance.

So what's the expected behavior for a Publisher to handle exceptions? I was trying to call `onError()` with the `IllegalStateException` which holds the thrown exception from `onSubscribe()` and `onNext()`. (I have to think about more on thrown exceptions from `onComplete()` and `onError()`. Probably just will log\ except for the fatal-ones.)

> But as long as we expose Subscriber to users, it seems futile for us to have a pattern of "never throwing exceptions", as we've already found since we are leaving NeverInvokedSubscriber alone.

Yes, exactly right. The user could implement bad subscribers so we should make our publisher handle this. Except that I mentioned earlier, is there anything we should do more? 🤔

anuraaga:
> So what's the expected behavior for a Publisher to handle exceptions?

The spec seems to say you should wrap in `IllegalStateException` with a method quoting the reactive streams spec. But this doesn't seem very user-friendly to me since users don't care about reactive streams spec they just want to see the exception. For example, a possible `onNext` is

```java
onNext(Request r) {
  backendResponse = backendClient(r).aggregate().join();
  if (backendResponse.isSuccess()) {
    subscription.cancel();
    response.complete(HttpResponse.of(OK));
  }
}
```

I think it's more useful for a user if `onError` is called with the actual backend client error, but if we follow the spec, it needs to be wrapped in a spec violation error. I don't consider this bad code by the user so don't think our publisher should either.

But if we follow reactive-streams spec, we can't do that - so it seems reactive-streams suggests having custom types. I guess like we have `StreamMessage`, we can have `StreamSubscriber` in Armeria. `StreamSubscriber` doesn't have restrictions on errors and is the primary API for users to use. And then when interoping with another reactive streams library like rxjava, we could delegate using a `InteropSubscriber` which handles wrapping exceptions with protocol violation or logging.

minwoox:
> But this doesn't seem very user-friendly to me since users don't care about reactive streams spec they just want to see the exception.

I agree. 😄
So you think we should call `onError()` with the exception thrown from `onNext` instead of wrapping it, right?
Then, it's the publisher that relaxes the rule not calling `onError` with wrapping the exception, so we might not need `StreamSubscriber`?
Perhaps just adding a description to `StreamMessage` which explains about it is enough? 🤔

anuraaga:
> Perhaps just adding a description to StreamMessage which explains about it is enough?

Yeah I guess this seems good too. But in that case, do we need to change our subscribers? I think we only ever use them with `StreamMessage`, not other reactive frameworks.

minwoox:
> But in that case, do we need to change our subscribers?

Do you mean this PR? If so, I think abiding by the spec doesn't harm. 🤔

anuraaga:
> Do you mean this PR? If so, I think abiding by the spec doesn't harm. 

I guess so - let's make sure our logging decorators don't end up logging spec violations instead of real ones because of the change :)

minwoox:
> don't end up logging spec violations instead of real ones because of the change 

Yes, I will make a follow-up PR. 😉
Thanks a lot for your opinion and for leading me to the right direction. 👍

trustin:
Thanks, @minwoox and reviewers!

minwoox:
Thanks for reviewing!

PR #2536, opened 2020-02-26
Update Brave from 5.9.5 to 5.10.1

adriancole:
there will be a brave 5.10.1 shortly, just raising for feedback as there is at least one question area.

### DIFF
--- a/brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
+++ b/brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
@@ -16,6 +16,10 @@
 
 package com.linecorp.armeria.client.brave;
 
+import brave.http.HttpRequestParser;
+import brave.http.HttpResponse;
+import brave.http.HttpResponseParser;
+import brave.propagation.TraceContext;
 import java.net.SocketAddress;
 
 import com.linecorp.armeria.client.ClientRequestContext;
@@ -25,11 +29,9 @@ import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.internal.common.brave.SpanTags;
 
 import brave.SpanCustomizer;
-import brave.http.HttpAdapter;
-import brave.http.HttpClientParser;
 
 /**
- * Default implementation of {@link HttpClientParser}.
+ * Default implementation of {@link HttpRequestParser} and {@link HttpResponseParser} for clients.
  * This parser adds some custom tags and overwrites the name of span if {@link RequestLog#requestContent()}
  * is {@link RpcRequest}.
  * The following tags become available:
@@ -42,7 +44,7 @@ import brave.http.HttpClientParser;
  *   <li>address.local</li>
  * </ul>
  */
-final class ArmeriaHttpClientParser extends HttpClientParser {
+final class ArmeriaHttpClientParser implements HttpRequestParser, HttpResponseParser {
 
     private static final ArmeriaHttpClientParser INSTANCE = new ArmeriaHttpClientParser();
 
@@ -53,8 +55,10 @@ final class ArmeriaHttpClientParser extends HttpClientParser {
     private ArmeriaHttpClientParser() {}
 
     @Override
-    public <T> void request(HttpAdapter<T, ?> rawAdapter, T req, SpanCustomizer customizer) {
-        super.request(rawAdapter, req, customizer);
+    public void parse(brave.http.HttpRequest request, TraceContext context, SpanCustomizer span) {
+        HttpRequestParser.DEFAULT.parse(request, context, span);
+
+        final Object req = request.unwrap();
         if (!(req instanceof ClientRequestContext)) {
             return;
         }
@@ -66,39 +70,41 @@ final class ArmeriaHttpClientParser extends HttpClientParser {
             return;
         }
 
-        customizer.tag(SpanTags.TAG_HTTP_HOST, httpReq.authority())
-                  .tag(SpanTags.TAG_HTTP_URL, httpReq.uri().toString());
+        span.tag(SpanTags.TAG_HTTP_HOST, httpReq.authority())
+            .tag(SpanTags.TAG_HTTP_URL, httpReq.uri().toString());
     }
 
     @Override
-    public <T> void response(HttpAdapter<?, T> rawAdapter, T res, Throwable error, SpanCustomizer customizer) {
-        super.response(rawAdapter, res, error, customizer);
+    public void parse(HttpResponse response, TraceContext context, SpanCustomizer span) {
+        HttpResponseParser.DEFAULT.parse(response, context, span);
+
+        final Object res = response.unwrap();
         if (!(res instanceof ClientRequestContext)) {
             return;
         }
 
         final ClientRequestContext ctx = (ClientRequestContext) res;
         final RequestLog log = ctx.log().ensureComplete();
-        customizer.tag(SpanTags.TAG_HTTP_PROTOCOL, ClientRequestContextAdapter.protocol(log));
+        span.tag(SpanTags.TAG_HTTP_PROTOCOL, ClientRequestContextAdapter.protocol(log));
 
         final String serFmt = ClientRequestContextAdapter.serializationFormat(log);
         if (serFmt != null) {
-            customizer.tag(SpanTags.TAG_HTTP_SERIALIZATION_FORMAT, serFmt);
+            span.tag(SpanTags.TAG_HTTP_SERIALIZATION_FORMAT, serFmt);
         }
 
         final SocketAddress raddr = ctx.remoteAddress();
         if (raddr != null) {
-            customizer.tag(SpanTags.TAG_ADDRESS_REMOTE, raddr.toString());
+            span.tag(SpanTags.TAG_ADDRESS_REMOTE, raddr.toString());
         }
 
         final SocketAddress laddr = ctx.localAddress();
         if (laddr != null) {
-            customizer.tag(SpanTags.TAG_ADDRESS_LOCAL, laddr.toString());
+            span.tag(SpanTags.TAG_ADDRESS_LOCAL, laddr.toString());
         }
 
         final String name = log.name();
         if (name != null) {
-            customizer.name(name);
+            span.name(name);
         }
     }
 }
--- a/brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
+++ b/brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
@@ -16,10 +16,6 @@
 
 package com.linecorp.armeria.client.brave;
 
-import brave.http.HttpRequestParser;
-import brave.http.HttpResponse;
-import brave.http.HttpResponseParser;
-import brave.propagation.TraceContext;
 import java.net.SocketAddress;
 
 import com.linecorp.armeria.client.ClientRequestContext;
@@ -29,6 +25,10 @@ import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.internal.common.brave.SpanTags;
 
 import brave.SpanCustomizer;
+import brave.http.HttpRequestParser;
+import brave.http.HttpResponse;
+import brave.http.HttpResponseParser;
+import brave.propagation.TraceContext;
 
 /**
  * Default implementation of {@link HttpRequestParser} and {@link HttpResponseParser} for clients.
--- a/brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
+++ b/brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
@@ -58,12 +58,12 @@ final class ArmeriaHttpClientParser implements HttpRequestParser, HttpResponsePa
     public void parse(brave.http.HttpRequest request, TraceContext context, SpanCustomizer span) {
         HttpRequestParser.DEFAULT.parse(request, context, span);
 
-        final Object req = request.unwrap();
-        if (!(req instanceof ClientRequestContext)) {
+        final Object unwrapped = request.unwrap();
+        if (!(unwrapped instanceof ClientRequestContext)) {
             return;
         }
 
-        final ClientRequestContext ctx = (ClientRequestContext) req;
+        final ClientRequestContext ctx = (ClientRequestContext) unwrapped;
         final HttpRequest httpReq = ctx.request();
         if (httpReq == null) {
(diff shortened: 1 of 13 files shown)


brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -89,7 +89,9 @@ public String header(String name) {
 
         @Override
         public void header(String name, String value) {
-            headersBuilder.set(name, value);
+            if (headersBuilder != null) {
```
adriancole: when serving the request from the response side. I'm thinking we ignore header mutations..

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -113,6 +115,7 @@ public long startTimestamp() {
     @SuppressWarnings("ClassNameSameAsAncestorName")
     private static final class HttpClientResponse extends brave.http.HttpClientResponse {
         private final RequestLog log;
+        private brave.http.HttpClientRequest request;
```
adriancole: is being lazy here (and in server) over-optimization from POV of armeria?

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
adriancole: we could before, just htis will make that more explicit https://github.com/openzipkin/brave/pull/1103

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -113,6 +115,7 @@ public long startTimestamp() {
...
 
         HttpClientResponse(RequestLog log) {
```
anuraaga: I was sort of expecting this to accept `HttpClientRequest` as a parameter and passed in when initializing here

https://github.com/line/armeria/blob/master/brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java#L162

I guess it does mean mutations would be present but is it a problem either way?

But anyways, reexposing, with the lazy instantiation, seems fine to me too.

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
anuraaga: Think it's good.

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
adriancole: this is part of me not knowing the lifecycle. I'm not sure if the instance could be different by the time we reach response time. is that something we cam know?

anuraaga: Instance is the same. In other words, the only difference between passing the originally created `HttpClientRequest` here or recreating it here is the value of `headersBuilder`, everything else should be exactly the same since it's the same `ClientRequestContext`.

adriancole: ok in that case, let's reuse it. thanks!

adriancole:
fingers crossed

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -159,9 +160,9 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
                 }
             }
 
-            final HttpClientResponse response = ClientRequestContextAdapter.asHttpClientResponse(log);
+            final HttpClientResponse response = ClientRequestContextAdapter.asHttpClientResponse(log, request);
             try (SafeCloseable ignored = ctx.push()) {
-                handler.handleReceive(response, log.responseCause(), span);
+                handler.handleReceive(response, response.error(), span);
```
anuraaga: Unrelated to this PR so just food for thought / another PR but just noticed it - since we pass the span to the handler, so we need to push the context (e.g. mount the scope)?

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
```diff
@@ -44,9 +44,9 @@
     @SuppressWarnings("ClassNameSameAsAncestorName")
     private static final class HttpClientRequest extends brave.http.HttpClientRequest {
         private final ClientRequestContext ctx;
-        private final RequestHeadersBuilder headersBuilder;
+        @Nullable private final RequestHeadersBuilder headersBuilder;
```
anuraaga: I think we can revert the headers related changes now that we just serve the same request object

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
```diff
@@ -98,9 +100,9 @@ public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exc
                 // If the client timed-out the request, we will have never sent any response data at all.
             }
 
-            final HttpServerResponse response = ServiceRequestContextAdapter.asHttpServerResponse(log);
+            final HttpServerResponse response = ServiceRequestContextAdapter.asHttpServerResponse(log, request);
             try (SafeCloseable ignored = ctx.push()) {
```
anuraaga: Ditto

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceTest.java
```diff
@@ -148,7 +148,7 @@ void scopeDecorator() throws Exception {
 
         // check service name
         assertThat(span.localServiceName()).isEqualTo(TEST_SERVICE);
-        assertThat(scopeDecoratorCallingCounter.get()).isEqualTo(3);
+        assertThat(scopeDecoratorCallingCounter.get()).isEqualTo(1);
```
anuraaga: I guess neither 3 nor 1 mean much to me. We probably should add a comment explaining what this is actually asserting semantically.

anuraaga: Or we can remove if it's too strongly tied to implementation details to be helpful.

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
```diff
@@ -53,8 +55,10 @@ static ArmeriaHttpClientParser get() {
     private ArmeriaHttpClientParser() {}
 
     @Override
-    public <T> void request(HttpAdapter<T, ?> rawAdapter, T req, SpanCustomizer customizer) {
-        super.request(rawAdapter, req, customizer);
+    public void parse(brave.http.HttpRequest request, TraceContext context, SpanCustomizer span) {
```
trustin: nit: `span` could be renamed back to `customizer` or `spanCustomizer`?

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
```diff
@@ -66,39 +70,41 @@ private ArmeriaHttpClientParser() {}
...
+            .tag(SpanTags.TAG_HTTP_URL, httpReq.uri().toString());
     }
 
     @Override
-    public <T> void response(HttpAdapter<?, T> rawAdapter, T res, Throwable error, SpanCustomizer customizer) {
-        super.response(rawAdapter, res, error, customizer);
+    public void parse(HttpResponse response, TraceContext context, SpanCustomizer span) {
+        HttpResponseParser.DEFAULT.parse(response, context, span);
+
+        final Object res = response.unwrap();
```
trustin: nit: `res` -> `unwrapped`?

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerParser.java
```diff
@@ -51,23 +53,27 @@ private ArmeriaHttpServerParser() {
     }
 
     @Override
-    public <T> void request(HttpAdapter<T, ?> rawAdapter, T req, SpanCustomizer customizer) {
-        super.request(rawAdapter, req, customizer);
+    public void parse(brave.http.HttpRequest request, TraceContext context, SpanCustomizer span) {
+        HttpRequestParser.DEFAULT.parse(request, context, span);
+
+        final Object req = request.unwrap();
```
trustin: nit: `req` -> `unwrapped`

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerParser.java
```diff
@@ -51,23 +53,27 @@ private ArmeriaHttpServerParser() {
     }
 
     @Override
-    public <T> void request(HttpAdapter<T, ?> rawAdapter, T req, SpanCustomizer customizer) {
-        super.request(rawAdapter, req, customizer);
+    public void parse(brave.http.HttpRequest request, TraceContext context, SpanCustomizer span) {
```
trustin: nit: `span` -> `customizer` or `spanCustomizer`

brave/src/main/java/com/linecorp/armeria/server/brave/BraveService.java
```diff
@@ -75,7 +76,8 @@ private BraveService(HttpService delegate, HttpTracing httpTracing) {
 
     @Override
     public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
-        final Span span = handler.handleReceive(ServiceRequestContextAdapter.asHttpServerRequest(ctx));
+        final HttpServerRequest request = ServiceRequestContextAdapter.asHttpServerRequest(ctx);
```
trustin: `request` -> `braveReq` to avoid confusion?

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
```diff
@@ -159,9 +160,9 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
                 }
             }
 
-            final HttpClientResponse response = ClientRequestContextAdapter.asHttpClientResponse(log);
+            final HttpClientResponse response = ClientRequestContextAdapter.asHttpClientResponse(log, request);
```
trustin: nit: `request` -> `braveReq` and `response` -> `braveRes` to avoid confusion

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
anuraaga: Personally good with span, `Span`, `MutableSpan`, and `SpanCustomizer` are basically the same concept.

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceTest.java
adriancole: yeah good point.. the reduction by two is not scoping during parsing :) I'll make a comment that the presence of 1 is for downstream handlers. usually I check some interceptor can read the context vs counting, though in this case counting was a good thing as it shows the better performance we should expect

brave/src/main/java/com/linecorp/armeria/client/brave/ClientRequestContextAdapter.java
adriancole: agreed

brave/src/main/java/com/linecorp/armeria/client/brave/BraveClient.java
adriancole: brave 5.11 doesn't mount the scope anymore intentionally, to improve performance (as scope handlers dangling off a thread local can cost a bit). instead the parsers get an arg for the context. the old implementation of the parser will indeed scope, but as handleSend/handleReceive are callbacks (iotw cul-de-sacs, not interceptors which would surround user code), we don't have any use case for scoping anymore internally.

anuraaga: Cool - yeah I think that means we can recover the request context push around the callbacks too that is there to allow brave to scope off the threadlocal but if it's not doing that it should be unnecessary.

adriancole: great idea

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
trustin: OK. No need to rename, then. :-)

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceTest.java
trustin: Maybe `.isGreaterThanZero()` then?

brave/src/main/java/com/linecorp/armeria/client/brave/ArmeriaHttpClientParser.java
adriancole: yeah we fixed the name to span upstream. naming parameter based on type was laborious

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerParser.java
adriancole: choosing not to unless you think this is very important

adriancole:
thanks for the review @anuraaga and @trustin !

brave/src/main/java/com/linecorp/armeria/server/brave/ArmeriaHttpServerParser.java
adriancole: ditto on opting out, but if you really really want it to I can change it

brave/src/test/java/com/linecorp/armeria/server/brave/BraveServiceTest.java
adriancole: added isOne as incrementing would be a bug unless intentional

adriancole:
ps here's the release notes on the brave bump https://github.com/openzipkin/brave/releases/tag/5.10.1

trustin:
:clap::clap::clap:

adriancole:
fyi here's sleuth wiring for the parsers https://github.com/spring-cloud/spring-cloud-sleuth/pull/1575

Issue #2535, opened 2020-02-26
Undeprecate `ctx.setRequest/ResponseTimeout()`?


trustin:
We got some internal inquiry about what should a user do instead of `ServiceRequestContext.setRequestTimeoutMillis(long)`. IIUC the current API, it has to be:

```java
final requestStartTimeMillis = 
    ctx.log().ensureAvailable(RequestLogProperty.REQUEST_START_TIME)
             .requestStartTimeMillis();
ctx.setRequestTimeoutAt(Instant.ofEpochMillis(requestStartTimeMillis + timeout);
```

... which is not ideal.

On the other hand, I found a bug where the current `setRequestTimeoutMillis()` implementation uses `extendRequestTimeout()`, which means the timeout will not be set when there's no timeout scheduled yet.

ikhoon:
> On the other hand, I found a bug where the current setRequestTimeoutMillis() implementation uses extendRequestTimeout(), which means the timeout will not be set when there's no timeout scheduled yet.

Oops, Let me check this.

anuraaga:
I remember suggesting not having `extend` and only having `set` since `set`(`get` + `delta`) seems reasonable compact and very readable. But didn't get any feedback on that. I'd still recommend it.

I think it's also good to confirm with the user though whether they actually want `setTimeout`. I do strongly believe that 90% of users actually want `setTimeoutAfter` which we have.

anuraaga:
Sorry for accidentally closing issue >< On mobile it's just one mistouch when wanting to scroll the page.

KarboniteKream:
> I think it's also good to confirm with the user ...

We don't _really_ need `setTimeout()`, as our current logic is implemented in a decorator, which is executed at the start of the request, so the behavior should be virtually the same as `setTimeoutAfter()`.

anuraaga:
Just wondering, if you had never seen the `setTimeout` method before (I'm guessing you did while upgrading to latest version of Armeria and cleaning up deprecations), would you have found using `setTimeoutAfter` awkward?

I've seen several cases where users use `setTimeout` wrongly assuming it sets it based on the current time creating a bug and haven't heard the opposite yet - I'm wondering how many users really want `setTimeout`.

Either way, having a fourth way of setting timeout seems too much, so if we want it, lean towards deprecating `extend` in favor of a good old fashioned `set(get+delta)`.

trustin:
> But didn't get any feedback on that. I'd still recommend it.

[...]

PR #2537, opened 2020-02-27
Fix a bug where `ctx.set*Timeout()` does not schedule a timeout

ikhoon:
…eout*`

Motivation:

When a timeout is not scheduled for a request, a new timeout set by
`set{Request,Response}Timeout` does not work well.
See #2535

Modifications:

- Make `set{Request,Response}Timeout` call `set{Request,Reponse}TimeoutAfterMillis`
  if a timeout is not scheduled yet.

Result:

The `set{Request,Response}Timeout*` properly set the new timeout when no timeout is scheduled.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
@@ -457,9 +457,10 @@ public final class DefaultClientRequestContext
                       "responseTimeoutMillis: %s (expected: >= 0)", responseTimeoutMillis);
         if (responseTimeoutMillis == 0) {
             clearResponseTimeout();
+            return;
         }
 
-        if (this.responseTimeoutMillis == 0 && responseTimeoutMillis > 0) {
+        if (this.responseTimeoutMillis == 0) {
             setResponseTimeoutAfterMillis(responseTimeoutMillis);
             return;
         }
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -459,6 +459,11 @@ public void setResponseTimeoutMillis(long responseTimeoutMillis) {
             clearResponseTimeout();
         }
 
+        if (this.responseTimeoutMillis == 0 && responseTimeoutMillis > 0) {
```
trustin: `responseTimeoutMillis > 0` is always true at this point.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -339,6 +339,11 @@ public void setRequestTimeoutMillis(long requestTimeoutMillis) {
             clearRequestTimeout();
         }
 
+        if (this.requestTimeoutMillis == 0 && requestTimeoutMillis > 0) {
```
trustin: Ditto

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
ikhoon: Oops..

trustin:
Thanks, @ikhoon!

PR #2538, opened 2020-02-27
Require request path when building a client

### LINKED ISSUE #2525: Allow `*ClientBuilder.path()` even when created with a URI
Currently, the `path()` call in the following code will throw an exception:

```java
WebClient.builder("http://127.0.0.1:8080")
         .path("/foo")
```

.. because `path()` is currently allowed only when a builder is created with an `Endpoint`.

However, I feel like we don't really need to prohibit `path()` but could just concatenate, e.g.

```java
WebClient.builder("http://127.0.0.1:8080")
         .path("/foo") // URI becomes http://127.0.0.1:8080/foo

WebClient.builder("http://127.0.0.1:8080/foo/")
         .path("/bar") // URI becomes http://127.0.0.1:8080/foo/bar
``

We need to define the case of multiple `.path()` calls though:

```java
// /foo/baz or /foo/bar/baz ?
WebClient.builder("http://127.0.0.1:8080/foo/")
         .path("/bar")
         .path("/baz")
```

I'm somewhat in favor of concatenating multiple `.path()` calls, inserting/deduplicating slashes between them automatically. Thoughts?

/cc @mauhiz

trustin:
Related: #2525

Motivation:

The behavior of the `path()` method in client builder classes is
confusing. It throws an exception when the builder was created with a
URI while it doesn't when created with an `EndpointGroup`.

Modifications:

- Removed `path()` method from the builder classes.
- Added various factory methods that require a path to the clients and
  client builders.
- Added various factory methods that accepts `String` in lieu of
  `Scheme` or `SessionProtocol` for convenience.
- Slightly changed the behavior of `ClientBuilder.serializationFormat()`,
  so it doesn't throw an exception even when the builder was created
  with a URI.
- Added a variant of `ClientBuilder.serializationFormat()` that accept a
  `String`.

Result:

- Less confusion.
- More convenience.
- Closes #2525

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
@@ -91,7 +91,7 @@ public final class ClientBuilder extends AbstractClientOptionsBuilder {
     }
 
     /**
-     * Sets the {@link SerializationFormat} of the client from the specified {@code String}.
+     * Sets the {@link SerializationFormat} of the client from the specified {@code format}.
      */
     public ClientBuilder serializationFormat(String format) {
         return serializationFormat(SerializationFormat.of(requireNonNull(format, "format")));
--- a/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
@@ -30,7 +30,6 @@ import com.linecorp.armeria.client.endpoint.EndpointGroup;
 import com.linecorp.armeria.common.RequestId;
 import com.linecorp.armeria.common.Scheme;
 import com.linecorp.armeria.common.SerializationFormat;
-import com.linecorp.armeria.common.SessionProtocol;
 
 /**
  * Creates a new client that connects to the specified {@link URI} using the builder pattern. Use the factory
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -67,96 +66,43 @@
...
+        }
+        uri = null;
         this.endpointGroup = endpointGroup;
+        this.path = path;
+        this.scheme = scheme;
     }
 
     /**
-     * Sets the {@code path} of the client.
+     * Sets the {@link SerializationFormat} of the client from the specified {@code String}.
```
ikhoon: nit:
```suggestion
     * Sets the {@link SerializationFormat} of the client from the specified {@code format}.
```

core/src/main/java/com/linecorp/armeria/client/WebClient.java
```diff
@@ -312,15 +352,50 @@ static WebClientBuilder builder(URI uri) {
...
+    static WebClientBuilder builder(String protocol, EndpointGroup endpointGroup) {
+        return builder(SessionProtocol.of(requireNonNull(protocol, "protocol")), endpointGroup);
+    }
+
     /**
      * Returns a new {@link WebClientBuilder} created with the specified {@link SessionProtocol}
      * and base {@link EndpointGroup}.
      *
+     * @throws IllegalArgumentException if the {@code protocol} is not one of the fields
+     *                                  in {@link SessionProtocol}
```
ikhoon: nit: I don't think all `SessionProtocal` is supported. 
How about: 
```suggestion
     * @throws IllegalArgumentException if the {@code protocol} is not one of the fields
     *                                  in {@link SessionProtocol#httpValues()} or {@link SessionProtocol#httpsValues()}

```

core/src/main/java/com/linecorp/armeria/client/WebClient.java
```diff
@@ -312,15 +352,50 @@ static WebClientBuilder builder(URI uri) {
...
+    /**
+     * Returns a new {@link WebClientBuilder} created with the specified {@link SessionProtocol},
+     * base {@link EndpointGroup} and path.
+     *
      * @throws IllegalArgumentException if the {@code sessionProtocol} is not one of the fields
      *                                  in {@link SessionProtocol}
      */
-    static WebClientBuilder builder(SessionProtocol sessionProtocol, EndpointGroup endpointGroup) {
-        return new WebClientBuilder(sessionProtocol, endpointGroup);
+    static WebClientBuilder builder(SessionProtocol sessionProtocol, EndpointGroup endpointGroup,
```
ikhoon: nit: Change `sessionProtocol` to `protocol` for consistency?
```suggestion
    static WebClientBuilder builder(SessionProtocol protocol, EndpointGroup endpointGroup,
```

core/src/main/java/com/linecorp/armeria/client/WebClient.java
```diff
@@ -312,15 +352,50 @@ static WebClientBuilder builder(URI uri) {
...
+    }
+
+    /**
+     * Returns a new {@link WebClientBuilder} created with the specified {@code protocol}.
+     * base {@link EndpointGroup} and path.
+     *
+     * @throws IllegalArgumentException if the {@code protocol} is not one of the fields
+     *                                  in {@link SessionProtocol}
+     */
+    static WebClientBuilder builder(String protocol, EndpointGroup endpointGroup, @Nullable String path) {
```
ikhoon: How about making `path` non-null? Because this builder already have 
`builder(SessionProtocol protocol, EndpointGroup endpointGroup)`.

trustin: Made `path` non-null. :bowing_man:

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofit.java
```diff
@@ -78,14 +104,40 @@ public static ArmeriaRetrofitBuilder builder(URI baseUrl) {
...
     public static ArmeriaRetrofitBuilder builder(SessionProtocol protocol, EndpointGroup endpointGroup) {
+        return builder(protocol, endpointGroup, null);
+    }
+
+    /**
+     * Returns a new {@link ArmeriaRetrofitBuilder} that builds a client that sends requests to
+     * the specified {@link EndpointGroup} using the specified {@link SessionProtocol} and {@code path}.
+     */
+    public static ArmeriaRetrofitBuilder builder(String protocol, EndpointGroup endpointGroup,
+                                                 @Nullable String path) {
```
ikhoon: Shouldn't we update this too? 😀

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofit.java
```diff
@@ -78,14 +104,40 @@ public static ArmeriaRetrofitBuilder builder(URI baseUrl) {
...
+        return builder(SessionProtocol.of(requireNonNull(protocol, "protocol")), endpointGroup, path);
+    }
+
+    /**
+     * Returns a new {@link ArmeriaRetrofitBuilder} that builds a client that sends requests to
+     * the specified {@link EndpointGroup} using the specified {@link SessionProtocol} and {@code path}.
+     */
+    public static ArmeriaRetrofitBuilder builder(SessionProtocol protocol, EndpointGroup endpointGroup,
+                                                 @Nullable String path) {
```
ikhoon: Ditto

retrofit2/src/main/java/com/linecorp/armeria/client/retrofit2/ArmeriaRetrofit.java
trustin: Fixed :sweat_smile:

core/src/main/java/com/linecorp/armeria/client/ClientBuilder.java
```diff
@@ -67,96 +66,43 @@
...
-     * Sets the {@code path} of the client.
+     * Sets the {@link SerializationFormat} of the client from the specified {@code format}.
      */
-    public ClientBuilder path(String path) {
-        ensureEndpointGroup();
-        requireNonNull(path, "path");
-        checkArgument(path.startsWith("/"), "path: %s (expected: an absolute path starting with '/')", path);
-        this.path = path;
-        return this;
+    public ClientBuilder serializationFormat(String format) {
```
anuraaga: Similar to path, shouldn't this also go away since scheme is a required parameter?

minwoox: +1 if possible.

trustin: Done

PR #2539, opened 2020-02-27
Fix StreamMessage to handle abort() correctly

minwoox:
Motivation:
Let's see following example:
```java
StreamMessageAndWriter<Object> stream = new DefaultStreamMessage<>();
stream.write(data);
stream.close();
stream.subscribe(mySubscriber);
stream.abort();
```
In this case, `onError()` or `onComplete()`, which `onNext(data)` is called previously, should be called. But, currently, just `onComplete()` is invoked.

Modifications:
- If `StreamMessage.abort()` or `cancel()` is called, the `StreamMessage` handles the subscriber with the event and ignores the `close()` call.

Result:
- You now get `AbortedStreamException` error when the stream is aborted after closed.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
@@ -133,12 +133,6 @@ abstract class AbstractStreamMessage<T> implements StreamMessage<T> {
     abstract void cancel();
 
     /**
-     * Callback invoked to notify a {@link Subscriber} of a {@link CloseEvent}. The
-     * {@link AbstractStreamMessage} needs to ensure the notification happens on the correct thread.
-     */
-    abstract void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEvent event);
-
-    /**
      * Invoked after an element is removed from the {@link StreamMessage} and before
      * {@link Subscriber#onNext(Object)} is invoked.
      *
@@ -276,7 +270,7 @@ abstract class AbstractStreamMessage<T> implements StreamMessage<T> {
 
     static final class CloseEvent {
         @Nullable
-        private final Throwable cause;
+        final Throwable cause;
 
         CloseEvent(@Nullable Throwable cause) {
             this.cause = cause;
--- a/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
@@ -22,7 +22,6 @@ import static com.linecorp.armeria.common.stream.StreamMessageUtil.containsWithP
 import static java.util.Objects.requireNonNull;
 
 import java.util.List;
-import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
 
 import javax.annotation.Nullable;
@@ -169,36 +168,6 @@ abstract class AbstractStreamMessage<T> implements StreamMessage<T> {
             o = PooledObjects.toUnpooled(o);
         }
         return o;
-    }
-
-    /**
-     * Helper method for the common case of cleaning up all elements in a queue when shutting down the stream.
-     */
-    void cleanupQueue(SubscriptionImpl subscription, Queue<Object> queue) {
-        final Throwable cause = ClosedStreamException.get();
-        for (;;) {
-            final Object e = queue.poll();
-            if (e == null) {
-                break;
-            }
-
-            try {
-                if (e instanceof CloseEvent) {
-                    notifySubscriberOfCloseEvent(subscription, (CloseEvent) e);
-                    continue;
-                }
-
-                if (e instanceof CompletableFuture) {
-                    ((CompletableFuture<?>) e).completeExceptionally(cause);
-                }
-
-                @SuppressWarnings("unchecked")
-                final T obj = (T) e;
-                onRemoval(obj);
-            } finally {
-                ReferenceCountUtil.safeRelease(e);
-            }
-        }
     }
 
     /**
--- a/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
+++ b/core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
@@ -242,7 +242,7 @@ public class DefaultStreamMessage<T> extends AbstractStreamMessageAndWriter<T> {
             case CLOSED:
                 // close() has been called before cancel() or abort() is called.
                 // We just push the new CloseEvent so that SUCCESSFUL_CLOSE, which was pushed by close(), is
-                // ignored and cancel() or abort() call works?
+                // ignored and cancel() or abort() call has effect.
                 if (setState(State.CLOSED, State.CLEANUP)) {
                     addObjectOrEvent(newCloseEvent(cause));
                     return;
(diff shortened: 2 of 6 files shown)


core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -216,13 +240,12 @@ private void cancelOrAbort(Throwable cause) {
 
         switch (state) {
             case CLOSED:
-                // close() has been called before cancel(). There's no need to push a CloseEvent,
-                // but we need to ensure the completionFuture is notified and any pending objects
-                // are removed.
+                // close() has been called before cancel() or abort() is called.
+                // We just push the new CloseEvent so that SUCCESSFUL_CLOSE, which was pushed by close(), is
+                // ignored and cancel() or abort() call works?
```
trustin: Question mark?

core/src/test/java/com/linecorp/armeria/server/HttpServerHandlerTest.java
```diff
@@ -103,6 +104,6 @@ void httpResponseExceptionIsNotLoggedAsRequestCause() {
         await().untilAsserted(() -> {
             assertThat(logHolder.get().requestHeaders().path()).isEqualTo("/httpResponseException");
         });
-        assertThat(logHolder.get().requestCause()).isNull();
+        assertThat(logHolder.get().requestCause()).isInstanceOf(AbortedStreamException.class);
```
minwoox: The request is not subscribed but aborted in [`HttpServerHandler`](https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/HttpServerHandler.java#L416) so the cause is changed.
Should I change not to log this exception when the request is not subscribed? 🤔

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -361,8 +384,9 @@ private boolean notifyAwaitDemandFuture() {
     }
 
     private void handleCloseEvent(SubscriptionImpl subscription, CloseEvent o) {
-        setState(State.OPEN, State.CLEANUP);
-        notifySubscriberOfCloseEvent(subscription, o);
+        if (setState(State.CLOSED, State.CLEANUP)) {
```
minwoox: So the concept now is that, whatever the event which changes to the CLEANUP state, will notify the subscriber with it.

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -402,6 +426,41 @@ private boolean setState(State oldState, State newState) {
     }
 
     private void cleanup() {
-        cleanupQueue(subscription, queue);
+        final Throwable cause = ClosedStreamException.get();
```
trustin: `cause` could be instantiated only when necessary.

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -402,6 +426,41 @@ private boolean setState(State oldState, State newState) {
...
+        for (;;) {
+            final Object e = queue.poll();
+            if (e == null) {
+                break;
+            }
+
+            try {
+                // Just skip SUCCESSFUL_CLOSE because cancel() or abort() is called so
+                // we need to handle the subscriber with the event.
```
trustin:
```java
// Just skip SUCCESSFUL_CLOSE because being in cleanup() means one of the following:
// - All elements have been consumed already
// - cancel() or abort() has been invoked after successful close().
```

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
minwoox: -_-;

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
minwoox: It's a lot better. Thanks!

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
minwoox: Ah fixed. 😄

core/src/main/java/com/linecorp/armeria/common/stream/AbstractStreamMessage.java
```diff
@@ -171,36 +170,6 @@ T prepareObjectForNotification(SubscriptionImpl subscription, T o) {
...
-
-            try {
-                if (e instanceof CloseEvent) {
-                    notifySubscriberOfCloseEvent(subscription, (CloseEvent) e);
-                    continue;
-                }
-
-                if (e instanceof CompletableFuture) {
-                    ((CompletableFuture<?>) e).completeExceptionally(cause);
-                }
```
minwoox: There was a bug here. we should continue here. 😱

core/src/test/java/com/linecorp/armeria/server/HttpServerHandlerTest.java
trustin: Hmm. Most `GET` request handler will not subscribe at all. Perhaps we should handle the case where it is certain that request (or response) body will be empty?

trustin: However, such logic isn't part of the `StreamMessage` implementations. Could probably be done in the places that call `abort()`?

minwoox: > Perhaps we should handle the case where it is certain that request (or response) body will be empty?

That's a nice idea. `EmptyFixedStreamMessage` works in the same way. Let me make a change for that. Thanks!

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -194,44 +215,56 @@ private void doRequest(long n) {
...
     void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEvent event) {
-        // Always called from the subscriber thread.
+        if (subscription.needsDirectInvocation()) {
+            notifySubscriberOfCloseEvent0(subscription, event);
+        } else {
+            subscription.executor().execute(() -> notifySubscriberOfCloseEvent0(subscription, event));
+        }
+    }
+
+    void notifySubscriberOfCloseEvent0(SubscriptionImpl subscription, CloseEvent event) {
```
ikhoon: Could be `private`?

core/src/main/java/com/linecorp/armeria/common/stream/DefaultStreamMessage.java
```diff
@@ -194,44 +215,56 @@ private void doRequest(long n) {
...
+            // It the state was CLOSED, close() or close(cause) has been called before cancel() or abort()
+            // is called. We just ignore the previously pushed event and deal with CANCELLED_CLOSE.
+            final SubscriptionImpl subscription = this.subscription;
+            assert subscription != null;
+            notifySubscriberOfCloseEvent(subscription, CANCELLED_CLOSE);
+        }
     }
 
-    @Override
     void notifySubscriberOfCloseEvent(SubscriptionImpl subscription, CloseEvent event) {
```
ikhoon: Could be `private`?

minwoox: Fixed. :)

ikhoon:
Still LGTM 😀

minwoox:
Thanks for reviewing!

PR #2543, opened 2020-02-28
Remove, hide or unstablize deprecated API

trustin:
- Removed all deprecated API if possible.
- Hid all deprecated API if possible.
- Some notable changes worth mentioning:
  - `DefaultHttp{Request,Response}` has been moved to `internal.common`.
  - Renamed `DefaultRpcResponse` to `CompletableRpcResponse` and left only
    the default constructor.
  - Removed `http` from `RequestContextExporter` and its related classes,
    e.g. `httpRequestHeaders` -> `requestHeaders`
  - Fixed bugs where `RequestContextAware{Future,Promise,ProgressivePromise}`
    does not return itself but its delegate.
    - `removeListener(s)` now throws `UnsupportedOperationException`,
      because there's no way to remove the wrapped listener by reference.
  - Moved `RequestContext.makeContextAware(GenericFutureListener)` to
    `RequestContextAwareFutureListener` and `RequestContextAwareProgressiveFutureListener`.
  - Fixed a bug where `RequestContextExporter.attributes()` throws an exception or a map with incorrect aliases.


trustin:
Note that I did not remove:
- `*RequestContext.set*Timeout*()` because we might undeprecate them soon.
- `MoreNamingConventions` to give users more time to migrate.

ikhoon:
Still LGTM :-)

core/src/main/java/com/linecorp/armeria/common/logging/RequestContextExporterBuilder.java
```diff
@@ -316,7 +221,7 @@ public RequestContextExporterBuilder addKeyPattern(String keyPattern) {
         }
 
         if (exportKey == null) {
-            exportKey = components[0].substring(PREFIX_ATTRS.length());
+            exportKey = components[0];
```
ikhoon: 👍

trustin:
Thanks for reviewing!

PR #2548, opened 2020-03-04
Do not add additionalTrailers to ResponseHeaders

### LINKED ISSUE #2544: Trailers is added to ResponseHeaders when there's no content
Related: https://github.com/line/armeria/pull/2372#discussion_r386025977

I found out that the additional trailers is added to `ResponseHeaders` when the content of the response is empty.
I think we implemented this because of gRPC service that sends trailers only when in certain circumstances. https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java#L323

However, I think a user may want to send the response with just `ResponseHeaders` and trailers because it does not violate the [spec](https://tools.ietf.org/html/rfc7540#section-8.1).

Perhaps, we can just add the trailers to `ResponseHeaders` in the gRPC service only and remove the trailers from ctx so that `HttpResponseSubscriber` is able to send `ResponseHeaders` and trailers when it should.

/cc @anuraaga, @jyblue

minwoox:
Motivation:
Currently, we combine `additionalResponseTrailers` to the `ResponseHeaders` if the `EOS` of the `ResponseHeaders` is true.
I think we do this because, in gRPC, there's a chance that only trailers is sent to the client so we combine them.
However, the response can consist of just only `ResponseHeaders` and trailers. (i.e no data)
So I think we can just combine the trailers to `ResponseHeaders` in gRPC especially, and let the user send the trailers without data.

Modifications:
- Add `mutateAdditionalResponse(Headers|Trailers)` which takes a `Consumer`
- Add `mutateAdditionalRequestHeaders` which takes a `Consumer`
- Throw an exception when `HttpResponse` and `AggregatedHttpResponse` is created with the `ResponseHeaders` whose status is informational.

Result:
- Close #2544
- You now can send the trailers even when the content is empty.
- (Breaking)
  - Methods adding and removing additional headers and trailers are gone from `ClientRequestContext` and `ServiceRequestContext` except:
    - `addAdditionalResponseHeader()`
    - `setAdditionalResponseHeader()`
    - `addAdditionalResponseTrailer()`
    - `setAdditionalResponseTrailer()`
    - `addAdditionalRequestHeader()`
    - `setAdditionalRequestHeader()`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -26,6 +26,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Iterator;
 import java.util.Map.Entry;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
@@ -34,6 +35,7 @@ import javax.annotation.Nullable;
 import com.linecorp.armeria.client.endpoint.EndpointGroup;
 import com.linecorp.armeria.common.ContentTooLargeException;
 import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpHeadersBuilder;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponseWriter;
 import com.linecorp.armeria.common.Request;
@@ -632,34 +634,9 @@ public interface ClientRequestContext extends RequestContext {
     HttpHeaders additionalRequestHeaders();
 
     /**
-     * Sets a header with the specified {@code name} and {@code value}. This will remove all previous values
-     * associated with the specified {@code name}.
-     * The header will be included when a {@link Client} sends an {@link HttpRequest}.
-     */
-    void setAdditionalRequestHeader(CharSequence name, Object value);
-
-    /**
-     * Clears the current header and sets the specified {@link HttpHeaders} which is included when a
-     * {@link Client} sends an {@link HttpRequest}.
-     */
-    void setAdditionalRequestHeaders(Iterable<? extends Entry<? extends CharSequence, ?>> headers);
-
-    /**
-     * Adds a header with the specified {@code name} and {@code value}. The header will be included when
-     * a {@link Client} sends an {@link HttpRequest}.
-     */
-    void addAdditionalRequestHeader(CharSequence name, Object value);
-
-    /**
-     * Adds the specified {@link HttpHeaders} which is included when a {@link Client} sends an
-     * {@link HttpRequest}.
-     */
-    void addAdditionalRequestHeaders(Iterable<? extends Entry<? extends CharSequence, ?>> headers);
-
-    /**
-     * Removes all headers with the specified {@code name}.
+     * Mutates the {@link HttpHeaders} which is included when a {@link Client} sends an {@link HttpRequest}.
      *
-     * @return {@code true} if at least one entry has been removed
+     * @param mutator the {@link Consumer} that mutates the additional request headers
      */
-    boolean removeAdditionalRequestHeader(CharSequence name);
+    void mutateAdditionalRequestHeaders(Consumer<HttpHeadersBuilder> mutator);
 }
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -629,12 +629,26 @@ public interface ClientRequestContext extends RequestContext {
     void setMaxResponseLength(long maxResponseLength);
 
     /**
-     * Returns an {@link HttpHeaders} which is included when a {@link Client} sends an {@link HttpRequest}.
+     * Returns an {@link HttpHeaders} which will be included when a {@link Client} sends an {@link HttpRequest}.
      */
     HttpHeaders additionalRequestHeaders();
 
     /**
-     * Mutates the {@link HttpHeaders} which is included when a {@link Client} sends an {@link HttpRequest}.
+     * Sets a header with the specified {@code name} and {@code value}. This will remove all previous values
+     * associated with the specified {@code name}.
+     * The header will be included when a {@link Client} sends an {@link HttpRequest}.
+     */
+    void setAdditionalRequestHeader(CharSequence name, Object value);
+
+    /**
+     * Adds a header with the specified {@code name} and {@code value}. The header will be included when
+     * a {@link Client} sends an {@link HttpRequest}.
+     */
+    void addAdditionalRequestHeader(CharSequence name, Object value);
+
+    /**
+     * Mutates the {@link HttpHeaders} which will be included when a {@link Client} sends
+     * an {@link HttpRequest}.
      *
      * @param mutator the {@link Consumer} that mutates the additional request headers
      */
--- a/core/src/main/java/com/linecorp/armeria/client/Clients.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Clients.java
@@ -498,9 +498,9 @@ public final class Clients {
      * import static com.linecorp.armeria.common.HttpHeaderNames.AUTHORIZATION;
      * import static com.linecorp.armeria.common.HttpHeaderNames.USER_AGENT;
      *
-     * try (SafeCloseable ignored = withHttpHeaders(headersBuilder -> {
-     *     headersBuilder.set(HttpHeaders.AUTHORIZATION, myCredential)
-     *                   .set(HttpHeaders.USER_AGENT, myAgent);
+     * try (SafeCloseable ignored = withHttpHeaders(builder -> {
+     *     builder.set(HttpHeaders.AUTHORIZATION, myCredential)
+     *            .set(HttpHeaders.USER_AGENT, myAgent);
      * })) {
      *     client.executeSomething(..);
      * }
@@ -515,7 +515,7 @@ public final class Clients {
      *      })) {
      *     for (String secret : secrets) {
      *         try (SafeCloseable ignored2 = withHttpHeaders(builder -> {
-     *                  builder.set(AUTHORIZATION, secret)
+     *                  builder.set(AUTHORIZATION, secret);
      *              })) {
      *             // Both USER_AGENT and AUTHORIZATION will be set.
      *             client.executeSomething(..);
--- a/core/src/main/java/com/linecorp/armeria/client/Clients.java
+++ b/core/src/main/java/com/linecorp/armeria/client/Clients.java
@@ -25,6 +25,7 @@ import javax.annotation.Nullable;
 
 import com.linecorp.armeria.client.endpoint.EndpointGroup;
 import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpHeadersBuilder;
 import com.linecorp.armeria.common.Request;
 import com.linecorp.armeria.common.Scheme;
 import com.linecorp.armeria.common.SerializationFormat;
@@ -386,12 +387,14 @@ public final class Clients {
      * }
      * }</pre>
      *
-     * @see #withHttpHeaders(Function)
+     * @see #withHttpHeaders(Consumer)
      */
     public static SafeCloseable withHttpHeader(CharSequence name, String value) {
         requireNonNull(name, "name");
         requireNonNull(value, "value");
-        return withHttpHeaders(headers -> headers.toBuilder().set(name, value).build());
+        return withHttpHeaders(headersBuilder -> {
+            headersBuilder.set(name, value);
+        });
     }
 
     /**
@@ -422,12 +425,14 @@ public final class Clients {
      * }
      * }</pre>
      *
-     * @see #withHttpHeaders(Function)
+     * @see #withHttpHeaders(Consumer)
      */
     public static SafeCloseable withHttpHeader(CharSequence name, Object value) {
         requireNonNull(name, "name");
         requireNonNull(value, "value");
-        return withHttpHeaders(headers -> headers.toBuilder().setObject(name, value).build());
+        return withHttpHeaders(headersBuilder -> {
+            headersBuilder.setObject(name, value);
+        });
     }
 
     /**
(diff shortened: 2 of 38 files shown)


core/src/main/java/com/linecorp/armeria/server/ServiceRequestContext.java
```diff
@@ -564,19 +565,21 @@ default String pathParam(String name) {
...
     HttpHeaders additionalResponseTrailers();
 
+    /**
+     * Returns the {@link HttpHeaders} which is set using {@code (add|set)AdditionalResponseTrailer} and
+     * clears the {@link HttpHeaders} so that the {@link HttpHeaders} is not sending twice when
+     * a {@link Service} completes an {@link HttpResponse}.
+     *
+     * <p>This is used only when the trailers is sent as a {@link ResponseHeaders} in gRPC.
+     */
+    HttpHeaders getAndRemoveAdditionalResponseTrailers();
```
anuraaga: I don't think we should add a gRPC only API to request context interface. Actually I was expecting a gRPC only implementation to look a little hacky hence encouraging a client option or something similarly at the client level instead :)

anuraaga: That being said if we can expose this through an internal API that relies on downcast to `DefaultRequestContext` it's still ok

minwoox: Ah, Thanks. Yes, I was a little worried about it but just didn't make sure. 😉 Let me remove this.

grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java
```diff
@@ -534,6 +531,12 @@ static HttpHeaders statusToTrailers(
                          Base64.getEncoder().encodeToString(proto.toByteArray()));
         }
 
+        if (ctx instanceof DefaultServiceRequestContext) {
```
anuraaga: I forget if request context is unwrappable but if so better to use unwrap.

minwoox: It's not unwrappable so I guess just using `instanceof` is the best I can do. 😆

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -40,6 +41,9 @@
      * Creates a new HTTP response.
      *
      * @param statusCode the HTTP status code
+     *
+     * @throws IllegalArgumentException if the specified {@code statusCode} is
+     *                                  {@link HttpStatus#isInformational()}.
```
trustin: How about using `{@linkplain}`?

```suggestion
     * @throws IllegalArgumentException if the specified {@code statusCode} is
     *                                  {@linkplain HttpStatus#isInformational() informational}.
```

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -49,9 +53,13 @@ static AggregatedHttpResponse of(int statusCode) {
      * Creates a new HTTP response.
      *
      * @param status the HTTP status
+     *
+     * @throws IllegalArgumentException if the specified {@link HttpStatus} is
+     *                                  {@link HttpStatus#isInformational()}.
```
trustin:
```suggestion
     * @throws IllegalArgumentException if the specified {@link HttpStatus} is
     *                                  {@linkplain HttpStatus#isInformational() informational}.
```

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -159,6 +185,9 @@ static AggregatedHttpResponse of(HttpStatus status, MediaType mediaType, HttpDat
      * Creates a new HTTP response with empty content.
      *
      * @param headers the HTTP headers
+     *
+     * @throws IllegalArgumentException if the status of the specified {@link ResponseHeaders} is
+     *                                  {@link HttpStatus#isInformational()}.
```
trustin:
```suggestion
     * @throws IllegalArgumentException if the status of the specified {@link ResponseHeaders} is
     *                                  {@linkplain HttpStatus#isInformational() informational}.
```

core/src/main/java/com/linecorp/armeria/common/AggregatedHttpResponse.java
```diff
@@ -198,12 +233,18 @@ static AggregatedHttpResponse of(ResponseHeaders headers, HttpData content, Http
...
      */
     static AggregatedHttpResponse of(Iterable<ResponseHeaders> informationals, ResponseHeaders headers,
                                      HttpData content, HttpHeaders trailers) {
 
         requireNonNull(informationals, "informationals");
         requireNonNull(headers, "headers");
+        final HttpStatus status = headers.status();
+        checkArgument(!status.isInformational(), "status: %s (expected: a non-1xx status", status);
```
trustin: Missing closing parenthesis:
```suggestion
        checkArgument(!status.isInformational(), "status: %s (expected: a non-1xx status)", status);
```

core/src/main/java/com/linecorp/armeria/common/HttpStatus.java
```diff
@@ -377,14 +377,10 @@ public static HttpStatus valueOf(String statusText) {
...
      * Returns {@code true} if the content of the response for the specified status code is expected to
-     * be always empty (1xx, 204, 205 and 304 responses.)
+     * be always empty (204, 205 and 304 responses.)
      */
     @SuppressWarnings("checkstyle:OverloadMethodsDeclarationOrder")
     public static boolean isContentAlwaysEmpty(int statusCode) {
-        if (HttpStatusClass.INFORMATIONAL.contains(statusCode)) {
-            return true;
-        }
-
```
trustin: Why was this removed?

core/src/main/java/com/linecorp/armeria/internal/server/annotation/AnnotatedServiceFactory.java
```diff
@@ -279,12 +279,6 @@ private static HttpStatus defaultResponseStatus(Method method) {
         setAdditionalHeader(defaultTrailers, method, "trailer", methodAlias, "method",
                             AdditionalTrailer.class, AdditionalTrailer::name, AdditionalTrailer::value);
 
-        if (defaultHeaders.status().isContentAlwaysEmpty() && !defaultTrailers.isEmpty()) {
-            logger.warn("A response with HTTP status code '{}' cannot have a content. " +
-                        "Trailers defined at '{}' might be ignored.",
-                        defaultHeaders.status().code(), methodAlias);
```
trustin: Shouldn't we warn because it's true in HTTP/1?

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -496,33 +498,33 @@ public HttpHeaders additionalResponseHeaders() {
...
     public void addAdditionalResponseHeaders(Iterable<? extends Entry<? extends CharSequence, ?>> headers) {
         requireNonNull(headers, "headers");
-        updateAdditionalResponseHeaders(additionalResponseHeadersUpdater,
-                                        builder -> builder.addObject(headers));
+        updateAdditionalHeaders(additionalResponseHeadersUpdater,
+                                builder -> builder.addObject(headers));
     }
 
-    private void updateAdditionalResponseHeaders(
+    private void updateAdditionalHeaders(
```
trustin: Not sure why we need to rename this method. Could you explain?

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -534,66 +536,50 @@ private void updateAdditionalResponseHeaders(
         }
     }
 
-    @Override
-    public boolean removeAdditionalResponseHeader(CharSequence name) {
-        return removeAdditionalResponseHeader(additionalResponseHeadersUpdater, name);
-    }
```
trustin: I don't get why this method has to be removed. Doesn't a user sometimes need to remove a header from the additional response headers? e.g. A user can write a decorator that removes an additional response header added by other service.

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -534,66 +536,50 @@ private void updateAdditionalResponseHeaders(
...
+    /**
+     * Returns the {@link HttpHeaders} which is set using {@code (add|set)AdditionalResponseTrailer} and
+     * clears the {@link HttpHeaders} so that the {@link HttpHeaders} is not sending twice when
+     * a {@link Service} completes an {@link HttpResponse}.
+     *
+     * <p>This is used only when the trailers is sent as a {@link ResponseHeaders} in gRPC.
+     */
+    public HttpHeaders getAndRemoveAdditionalResponseTrailers() {
+        return additionalResponseTrailersUpdater.getAndSet(this, HttpHeaders.of());
+    }
```
trustin: We will not need this method at all if we have `mutateAdditionalResponseTrailers()`:

```java
HttpHeaders oldTrailers = ctx.additionalResponseTrailers();
ctx.mutateAdditionalResponseTrailers(mutator -> mutator.clear());
```

core/src/main/java/com/linecorp/armeria/common/HttpStatus.java
minwoox: Because I think whether the content is empty or not is decided by the response headers, not the informational headers.
Informational headers will have the following response headers, so shouldn't we look at the response headers to make a decision if the following content will be empty or not?

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
minwoox: Let me revert this. Bad decision. 😆

core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
minwoox: I thought the situation is never going to happen but I guess I was wrong. 😆
And I really like the idea of providing the API to mutate headers. So we can have this behavior again. Thanks!

core/src/main/java/com/linecorp/armeria/internal/server/annotation/AnnotatedServiceFactory.java
minwoox: That's a good point. Let me put it back with a little change.

minwoox:
> Perhaps we could replace {add,set,remove}Additional{Request,Response}{Header,Trailer}*() with mutateAdditional{Request,Response}{Headers,Trailers}()?

Fixed. I really like the idea of providing this simple API instead of a bunch of APIs. PTAL. 🙇

core/src/main/java/com/linecorp/armeria/client/Clients.java
```diff
@@ -477,7 +477,7 @@ public static SafeCloseable withHttpHeaders(
         requireNonNull(headerManipulator, "headerManipulator");
         return withContextCustomizer(ctx -> {
             final HttpHeaders manipulatedHeaders = headerManipulator.apply(ctx.additionalRequestHeaders());
-            ctx.setAdditionalRequestHeaders(manipulatedHeaders);
+            ctx.mutateAdditionalRequestHeaders(mutator -> mutator.add(manipulatedHeaders));
```
minwoox: In this class, we have 3 different `withHttpHeader` methods and I'm wondering if we can do the same approach here. Or at least we can deprecate this method and add a new method that takes a `Consumer`.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -632,34 +634,9 @@ ClientRequestContext newDerivedContext(RequestId id, @Nullable HttpRequest req,
     HttpHeaders additionalRequestHeaders();
 
     /**
-     * Sets a header with the specified {@code name} and {@code value}. This will remove all previous values
-     * associated with the specified {@code name}.
-     * The header will be included when a {@link Client} sends an {@link HttpRequest}.
-     */
-    void setAdditionalRequestHeader(CharSequence name, Object value);
```
trustin: How about reviving two methods - `{add,set}AdditionalRequestHeader(CharSequence, Object)` - since it's used pretty commonly?

core/src/main/java/com/linecorp/armeria/client/Clients.java
trustin: It's pretty common to specify a single header, so I'd prefer keeping them all.

core/src/main/java/com/linecorp/armeria/common/HttpStatus.java
trustin: Doesn't it still mean that informational headers have no content?

core/src/main/java/com/linecorp/armeria/client/Clients.java
minwoox: Then just how about changing this method?
```java
// before
public static SafeCloseable withHttpHeaders(
            Function<? super HttpHeaders, ? extends HttpHeaders> headerManipulator) {
// after
public static SafeCloseable withHttpHeaders(Consumer<HttpHeadersBuilder> headerManipulator) {
```
So that a user does not have to `toBuilder` and `build`?

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Yes, let me put this back. 😉

core/src/main/java/com/linecorp/armeria/client/Clients.java
trustin: OK!

core/src/main/java/com/linecorp/armeria/common/HttpStatus.java
trustin: Had a quick chat with @minwoox and we chose to keep the modification in this PR, because we can't tell if the response will have no content from an informational status.

core/src/main/java/com/linecorp/armeria/client/Clients.java
minwoox: Done! 😉

ikhoon:
Still LGTM :-)

trustin:
Thanks a lot, @minwoox !

minwoox:
Thanks for reviewing! 😉

PR #2549, opened 2020-03-04
Add documentation creating WebClient using HealthCheckedEndpointGroup

minwoox:
I recently heard that some of our users specified the original endpoint group not the `HealthCheckecEndpointGroup` when building a `WebClient`.
So I thought we need to reinforce the documentation.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
+++ b/core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
@@ -73,8 +73,8 @@ import io.netty.util.concurrent.Future;
  *                                   .retryInterval(Duration.ofSeconds(10))
  *                                   .build();
  *
- * // You must specify healthCheckedGroup when building a WebClient, otherwise the client would send the
- * // requests to the unhealthy endpoints.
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/linecorp/armeria/client/endpoint/healthcheck/HealthCheckedEndpointGroup.java
```diff
@@ -61,6 +61,23 @@
...
+ *
+ * // Decorate the EndpointGroup with HealthCheckedEndpointGroup
+ * // that sends HTTP health check requests to '/internal/l7check' every 10 seconds.
+ * HealthCheckedEndpointGroup healthCheckedGroup =
+ *         HealthCheckedEndpointGroup.builder(originalGroup, "/internal/l7check")
+ *                                   .protocol(SessionProtocol.HTTP)
+ *                                   .retryInterval(Duration.ofSeconds(10))
+ *                                   .build();
+ *
+ * // You must specify healthCheckedGroup when building a WebClient, otherwise the client would send the
```
anuraaga: `, otherwise health checking will not be enabled.`

We can't predict the future of whether there are unhealthy endpoints or not so should just state the facts.

minwoox: That's a good point. Fixed. 😄

PR #2557, opened 2020-03-05
Do not leak ping tasks in `HealthCheckService`

trustin:
Motivation:

When long-polling and periodic ping task are enabled together, which is
so by default, `PingTask` instances are not cleaned up properly, leading
to memory leak.

Modifications:

- Use `handle()` instead of `exceptionally()` to cancel `PingTask` and
  `TimeoutTask`.

Result:

- Memory leak is gone.


trustin:
I guess we need to release 0.98.4 that's 0.98.3 + this fix. We need more time to release 0.99.0 because we really don't want to release 0.100.0. :sweat_smile:

mauhiz:
What versions are affected by the issue this PR fixes?

minwoox:
It's `0.98.0` ~ `0.98.3`.
It's introduced from https://github.com/line/armeria/pull/2392

PR #2560, opened 2020-03-06
Copy RequestLog properties to child in RetryingClient

### LINKED ISSUE #2551: RequestLog is not propagated to child RequestLog when created
When `orginalCtx.newDerivedContext()` is called, if the `RequestLog` of the `orginalCtx` has certain `RequestLogProperty`s already (e.g. `RequestLog.name()`), the `RequestLog` of the derived context should have same `RequestLogProperty`s.

minwoox:
Motivation:
When retrying, certain properties of `RequestLog` such as `name()`,  `requestContent()` and `requestContentPreview()` can be copied to the child context, so they are used.

Modifications:
- Copy `name()`, `SerializationFormat`, `requestContent` and `requestContentPreview` manually when creating derived ctx in `RetryingClient`.
  - This is not happening in `newDerivedContext` because we don't know which properties should be copied when deriving.
  - Copy `responseContent` and `responseContentPreview` if `deferResponseContent(|Preview)` is called.

Result:
- Close #2551
- You can now retry depending on the content of the request.


core/src/main/java/com/linecorp/armeria/client/retry/AbstractRetryingClient.java
```diff
@@ -270,11 +274,41 @@ protected static ClientRequestContext newDerivedContext(ClientRequestContext ctx
...
+            if (name != null) {
+                logBuilder.name(name);
+            }
+        }
+
+        final RequestLogBuilder parentLogBuilder = ctx.logBuilder();
+        if (parentLogBuilder.isDeferRequestContentSet()) {
+            logBuilder.deferRequestContent();
+        }
+        parentLog.whenAvailable(RequestLogProperty.REQUEST_CONTENT).thenApply(requestLog -> {
```
minwoox: I wonder if I have to copy the `RESPONSE_CONTENT` as well. 🤔

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -357,6 +357,8 @@ default ClientRequestContext newDerivedContext(RequestId id,
      * Creates a new {@link ClientRequestContext} whose properties and {@link Attribute}s are copied from this
      * {@link ClientRequestContext}, except having different {@link Request}, {@link Endpoint} and its own
      * {@link RequestLog}.
+     *
+     * <p>Note that if you want to copy the properties of the {@link RequestLog}, you should do it by yourself.
```
trustin: How about simplifying like the following?

    Note that this method does not copy the {@link RequestLog} properties to the derived
    context.

core/src/main/java/com/linecorp/armeria/client/retry/AbstractRetryingClient.java
trustin: I think so. How about copying `SESSION` and `SCHEME` as well?

minwoox: I think `SESSION` and `SCHEME` are not available at this point.
They are rather copied from child to parent.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
minwoox: Fixed. 😉

trustin:
Could we also copy `SESSION` and `SCHEME`?

core/src/main/java/com/linecorp/armeria/client/retry/AbstractRetryingClient.java
```diff
@@ -270,11 +274,57 @@ protected static ClientRequestContext newDerivedContext(ClientRequestContext ctx
...
         } else {
-            return ctx.newDerivedContext(id, req, rpcReq);
+            derived = ctx.newDerivedContext(id, req, rpcReq);
         }
+
+        final RequestLogAccess parentLog = ctx.log();
+        final RequestLog partial = parentLog.partial();
+        final RequestLogBuilder logBuilder = derived.logBuilder();
+        // serializationFormat is always not null, so this is fine.
+        logBuilder.serializationFormat(partial.serializationFormat());
```
ikhoon: I wonder that could we move the code of log propopagation to `DefaultRequestLog.addChild(RequestLogAccess)`?

minwoox: As I said before, `addChild` can be called not just by `RetryingClient`, but other places because it's a public API.
We do not know which properties we have to copy from the parent to the child.

ikhoon: I understood, thanks for the explanation! 😀

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithLoggingTest.java
```diff
@@ -114,11 +118,38 @@ void retryingThenLogging() {
...
+                                              return delegate.execute(ctx, req);
+                                          })
                                           .build();
-        assertThat(client.get("/hello").aggregate().join().contentUtf8()).isEqualTo("hello");
+        try (ClientRequestContextCaptor captor = Clients.newContextCaptor()) {
+            assertThat(client.get("/hello").aggregate().join().contentUtf8()).isEqualTo("hello");
+            final RequestLogBuilder logBuilder = captor.get().logBuilder();
+            logBuilder.requestContentPreview("baz");
+
+            await().untilAsserted(() -> assertThat(logResult.size()).isEqualTo(successLogIndex));
```
ikhoon: nit:
```suggestion
            await().untilAsserted(() -> assertThat(logResult).hasSize(successLogIndex));
```

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithLoggingTest.java
```diff
@@ -114,11 +118,38 @@ void retryingThenLogging() {
...
+            TimeUnit.SECONDS.sleep(1);
+            // The last log is not complete because the parent log doesn't set the response content yet.
+            assertThat(logResult.size()).isEqualTo(successLogIndex);
```
ikhoon:
```suggestion
            assertThat(logResult).hasSize(successLogIndex);
```

core/src/test/java/com/linecorp/armeria/client/retry/RetryingClientWithLoggingTest.java
minwoox: Thanks, fixed. 😉

PR #232, opened 2016-08-17
Add DeferredStreamMessage and DeferredHttpResponse

trustin:
Motivation:

Sometimes you cannot prepare your HttpResponse or StreamMessage instance
when returning from Service.serve() or Client.execute(). It will be
useful to have a placeholder StreamMessage implementation whose stream
is actually delegated from another StreamMessage.

Modifications:

- Add DeferredStreamMessage and DeferredHttpRequest

Result:

Easier to write a Service or a Client that delays its execution.

PR #2562, opened 2020-03-09
Fix AuthSuccessHandler to use HttpRequest and HttpResponse

minwoox:
Motivation:
Related: #2323

Result:
- (Breaking) `AuthSuccessHandler` and `AuthFailureHandler` now does not have type parameters.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/auth/AuthServiceBuilder.java
+++ b/core/src/main/java/com/linecorp/armeria/server/auth/AuthServiceBuilder.java
@@ -37,8 +37,8 @@ public final class AuthServiceBuilder {
 
     @Nullable
     private Authorizer<HttpRequest> authorizer;
-    private AuthSuccessHandler<HttpRequest, HttpResponse> successHandler = Service::serve;
-    private AuthFailureHandler<HttpRequest, HttpResponse> failureHandler = (delegate, ctx, req, cause) -> {
+    private AuthSuccessHandler successHandler = Service::serve;
+    private AuthFailureHandler failureHandler = (delegate, ctx, req, cause) -> {
         if (cause != null) {
             AuthService.logger.warn("Unexpected exception during authorization.", cause);
         }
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/linecorp/armeria/server/auth/AuthServiceBuilder.java
```diff
@@ -154,7 +154,7 @@ public AuthServiceBuilder onSuccess(AuthSuccessHandler<HttpRequest, HttpResponse
      * By default, an exception thrown during authorization is logged at WARN level (if any) and a
      * {@code 401 Unauthorized} response will be sent.
      */
-    public AuthServiceBuilder onFailure(AuthFailureHandler<HttpRequest, HttpResponse> failureHandler) {
+    public AuthServiceBuilder onFailure(AuthFailureHandler failureHandler) {
```
minwoox: I have a question about this. Because I have changed to this to use `HttpRequest` only,
(I should've done this in #232) is there any chance that a user might not want to return different status than 401?
If so, I was wondering if we can guide them to use `HttpResponseException` or `HttpStatusException`.
And remove these handlers and just add sort of `AuthorizingStrategy` with a builder, so that a user just can return true or false.

anuraaga: Not sure I understand the question but I don't think we need a new strategy since we have `Authorizer`. These handlers seem clearer to me than throwing exceptions in the authorizer to me, and should have better performance.

minwoox: Ah yes, I should've made it clear before I ask a question. I was sort of thinking like:
- Change `Authorizer` to `AuthorizingStrategy`
  - The new strategy can have multiple `Authorizer`s inside to check. 
- Remove `AuthSuccessHandler` and `AuthFailureHandler`.

All these changes are based on that there's not so much chance that a user wants to send a different response other than 401, but I guess not. 😆
So yes, let's keep this as it is.

minwoox: Found the issue for this. https://github.com/line/armeria/issues/506

minwoox:
I will do the rest of the work after #2503 is merged. 😉

PR #2564, opened 2020-03-09
Do not warn blocking future operations when done

trustin:
Motivation:

Blocking future operations never block when the future is done, and thus
we should not warn about such cases.

Modifications:

- Do not warn when the future being checked is done.
- Update and clean up the test case
- Make `CompletableRpcResponse` and `AbstractRequestContextAwareFuture`
  extend `EventLoopCheckingFuture` to reduce code duplication.
- Move `EventLoopCheckingUtil` back into `EventLoopCheckingFuture`.
- Miscellaneous:
  - Keep the reported threads in a concurrent map.

Result:

- No false positive warnings
- Less code duplication
- Less contention


trustin:
Thanks for reviewing!

PR #2568, opened 2020-03-10
Add documentation for the order of client decoration

minwoox:
Related: #1117

To-do:
Add documentation for server side

### DIFF
--- a/site/src/sphinx/client-decorator.rst
+++ b/site/src/sphinx/client-decorator.rst
@@ -145,12 +145,12 @@ request is converted into the HTTP request. The following diagram describes it:
 .. uml::
 
     @startditaa
-    +--------+ req +-----------+ req +-----------+                       +-----------+ req +-----------+
-    |        |---->|           |---->|           |----> RPC to HTTP ---->|           |---->|           |--=->
-    | Thrift |     | #1 RPC    |     | #2 RPC    |                       | #1 HTTP   |     | #2 HTTP   |
-    | Client | res | decorator | res | decorator |                       | decorator | res | decorator |
-    |        |<----|           |<----|           |<---- HTTP to RPC <----|           |<----|           |<-=--
-    +--------+     +-----------+     +-----------+                       +-----------+     +-----------+
+    +--------+ req +-----------+ req +-----------+                     +-----------+ req +-----------+
+    |        |---->|           |---->|           |---> RPC to HTTP --->|           |---->|           |----=->
+    | Thrift |     | #1 RPC    |     | #2 RPC    |                     | #1 HTTP   |     | #2 HTTP   |
+    | Client | res | decorator | res | decorator |                     | decorator | res | decorator |
+    |        |<----|           |<----|           |<--- HTTP to RPC <---|           |<----|           |<-=----
+    +--------+     +-----------+     +-----------+                     +-----------+     +-----------+
     @endditaa
 
 If the decorator modifies the response (e.g. :api:`DecodingClient`) or spawns more requests
--- a/site/src/sphinx/client-decorator.rst
+++ b/site/src/sphinx/client-decorator.rst
@@ -67,6 +67,85 @@ If your decorator is expected to be reusable, it is recommended to define a new 
     // Using a lambda expression:
     cb.decorator(delegate -> new AuditClient(delegate));
 
+The order of decoration
+-----------------------
+
+The decorators are executed in reverse order of the insertion. The following example shows which order
+the decorators are executed by printing the messages.
+
+.. code-block:: java
+
+    ClientBuilder cb = Clients.builder(...);
+    cb.decorator((delegate, ctx, req) -> {
+        System.err.println("Thirdly, executed.");
+        ...
+    });
+    // No matter decorator() or option() is used, decorators are executed in reverse order.
+    cb.option(ClientOption.DECORATION, ClientDecoration.of((delegate, ctx, req) -> {
+        System.err.println("Secondly, executed");
+        ...
+    });
+    cb.decorator((delegate, ctx, req) -> {
+        System.err.println("Firstly, executed.");
+        ...
+    });
+
+If the client is a Thrift client and RPC decorators are inserted, HTTP decorators and RPC decorators are
+separately grouped and executed in reverse order of the insertion:
+
+.. code-block:: java
+
+    ClientBuilder cb = Clients.builder(...);
+
+    // RPC decorator #2.
+    cb.rpcDecorator((delegate, ctx, rpcReq) -> {
+        System.err.println("Secondly, executed.");
+        ...
+    });
+
+    // HTTP decorator #4.
+    cb.decorator((delegate, ctx, httpReq) -> {
+        System.err.println("Fourthly, executed.");
+        ...
+    });
+
+    // RPC decorator #1.
+    cb.rpcDecorator((delegate, ctx, rpcReq) -> {
+        System.err.println("Firstly, executed.");
+        ...
+    });
+
+    // HTTP decorator #3.
+    cb.decorator((delegate, ctx, httpReq) -> {
+        System.err.println("Thirdly, executed.");
+        ...
+    });
+
+An RPC request is converted into an HTTP request before it's sent to a server.
+So RPC decorators are inserted before the RPC request is converted and HTTP decorators are inserted after the
+request is converted into the HTTP request.
+
+If the decorator modifies the response (e.g. :api:`DecodingClient`) or spawns more requests
+(e.g. :api:`RetryingClient`), the consequence is different depending on the order of the decorators.
+Let's look at the following example that :api:`DecodingClient` and :api:`ContentPreviewingClient`
+are used together:
+
+.. code-block:: java
+
+    import com.linecorp.armeria.client.encoding.DecodingClient;
+    import com.linecorp.armeria.client.logging.ContentPreviewingClient;
+
+    ClientBuilder cb = Clients.builder(...);
+    // Don't do this. ContentPreviewingClient should be inserted after DecodingClient.
+    cb.decorator(ContentPreviewingClient.newDecorator(1000));
+    cb.decorator(DecodingClient.newDecorator());
+
+:api:`DecodingClient` decodes the content of HTTP responses.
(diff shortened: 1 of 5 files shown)


minwoox:
1583833025048@2x

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,113 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+
+.. code-block:: java
+
+    import com.linecorp.armeria.client.encoding.DecodingClient;
+    import com.linecorp.armeria.client.logging.ContentPreviewingClient;
+
+    ClientBuilder cb = Clients.builder(...);
+    // Don't do this. ContentPreviewingClient should be inserted after DecodingClient.
+    cb.decorator(ContentPreviewingClient.newDecorator(1000));
+    cb.decorator(DecodingClient.newDecorator());
```
ikhoon: Why don't we show the right example over the wrong one? Because many users might want to copy and paste.

```suggestion
    cb.decorator(DecodingClient.newDecorator());
    // ContentPreviewingClient should be inserted after DecodingClient.
    cb.decorator(ContentPreviewingClient.newDecorator(1000));
```

minwoox: Fixed. 😉

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,113 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+
+The following diagram describes how an HTTP request and HTTP response are gone through decorators:
+
+.. uml::
+
+    @startditaa
+    +-----------+  req  +-----------+  req  +-----------+  req  +------------------+    req    +--------------+
+    |           |------>|           |------>|           |------>|                  |---------->|              |
+    | WebClient |       | #1        |       | #2        |       | Armeria          |           |    Server    |
+    |           |  res  | decorator |  res  | decorator |  res  | Netwokring Layer |    res    |              |
```
jrhee17: `Networking`

minwoox: 😱 😅

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,113 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+    });
+
+    // #1 RPC decorator.
+    cb.rpcDecorator((delegate, ctx, rpcReq) -> {
+        System.err.println("Firstly, executed.");
+        ...
+    });
+
+An RPC request is converted into an HTTP request before it's sent to a server.
+So RPC decorators are inserted before the RPC request is converted and HTTP decorators are inserted after the
```
trustin: So -> Therefore,

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,113 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+request is converted into the HTTP request. The following diagram describes it:
```
trustin: ..., as described in the following diagram:

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,113 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+    +--------+ req +-----------+ req +-----------+                     +-----------+ req +-----------+
+    |        |---->|           |---->|           |---> RPC to HTTP --->|           |---->|           |----=->
+    | Thrift |     | #1 RPC    |     | #2 RPC    |                     | #1 HTTP   |     | #2 HTTP   |
+    | Client | res | decorator | res | decorator |                     | decorator | res | decorator |
+    |        |<----|           |<----|           |<--- HTTP to RPC <---|           |<----|           |<-=----
+    +--------+     +-----------+     +-----------+                     +-----------+     +-----------+
+    @endditaa
+
+If the decorator modifies the response (e.g. :api:`DecodingClient`) or spawns more requests
+(e.g. :api:`RetryingClient`), the consequence is different depending on the order of the decorators.
```
trustin: the consequence is ... -> the outcome may be ?

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,113 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+
+    ClientBuilder cb = Clients.builder(...);
+    cb.decorator(ContentPreviewingClient.newDecorator(1000));
+    // ContentPreviewingClient should be inserted after DecodingClient.
+    cb.decorator(DecodingClient.newDecorator());
+
+:api:`DecodingClient` decodes the content of HTTP responses.
+:api:`ContentPreviewingClient` is :ref:`content-previewing` of the HTTP response by setting it to the
+:api:`RequestLog`. Because it's inserted after :api:`DecodingClient`, which means that the response content
+is set before it's decoded, you will see the encoded response content preview.
```
trustin: How about adding the opposite case?

```rst
If the two decorators are added in the opposite order, you will get the decoded preview because :api:`DecodingClient` is evaluated first.
```

site/src/sphinx/client-decorator.rst
minwoox: Fixed. 😉

site/src/sphinx/client-decorator.rst
minwoox: Fixed.

site/src/sphinx/client-decorator.rst
minwoox: Fixed.

site/src/sphinx/client-decorator.rst
minwoox: Fixed. I rephrased a little bit because I changed the order of decoration. 😉

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,116 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+    @startditaa
+    +-----------+  req  +-----------+  req  +-----------+  req  +------------------+    req    +--------------+
+    |           |------>|           |------>|           |------>|                  |---------->|              |
+    | WebClient |       | #1        |       | #2        |       | Armeria          |           |    Server    |
+    |           |  res  | decorator |  res  | decorator |  res  | Networking Layer |    res    |              |
+    |           |<------|           |<------|           |<------|                  |<----------|              |
+    +-----------+       +-----------+       +-----------+       +------------------+           +--------------+
+    @endditaa
+
+If the client is a Thrift client and RPC decorators are inserted, HTTP decorators and RPC decorators are
```
anuraaga: I haven't really used RPC decorators before since don't use thrift. I'm surprised by this behavior, and understand it but it's not very intuitive to have implicit grouping.

Would it be worth changing the API a little to make the grouping explicit?

```java
cb.decorators().add()...

cb.rpcDecorators().add()...
```

minwoox: That's a good suggestion.
The example code that I picked up was really ugly and I think there's a still chance that a user does it in the same way:
```java
cb.decorators().add(d1);
cb.rpcDecorators().add(d2);
cb.decorators().add(d3);
```
And mostly, a user would have separate methods to set decorators if it gets complicated, so I'm not sure it's worth changing. 🤔 (But it looks good to me.)

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,116 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+
+    // #2 decorator
+    cb.decorator((delegate, ctx, req) -> {
+        System.err.println("Secondly, executed.");
+        ...
+    });
+
+    // #1 decorator.
+    // No matter decorator() or option() is used, decorators are executed in reverse order of the insertion.
+    cb.option(ClientOption.DECORATION, ClientDecoration.of((delegate, ctx, req) -> {
```
ikhoon: I found `ClientDecoration.of(...)` does not accept `DecoratingHttpClientFunction`.
https://github.com/line/armeria/blob/6f21290a3e6e3bb2af22d8a9061d472dcabda545/core/src/main/java/com/linecorp/armeria/client/ClientDecoration.java#L43

ikhoon: And how about making the method body complete for the detail example?
```java
// #2 decorator
cb.decorator((delegate, ctx, req) -> {
    System.err.println("Secondly, executed.");
    final HttpResponse res = delegate.execute(ctx, req);
    System.err.println("Thirdly, executed.");
    return res;
});

// #2 decorator
cb.option(ClientOption.DECORATION, ClientDecoration.of(delegate -> (ctx, req) -> {
    System.err.println("Firstly, executed");
    final HttpResponse res = delegate.execute(ctx, req);
    System.err.println("Fourthly, executed.");
    return res;
}));
```

ikhoon: > I found ClientDecoration.of(...) does not accept DecoratingHttpClientFunction.

Oops, you've already added it to this PR 😅

ikhoon: Had chat with @minwoox, we are going to focus on the exeution orders.
If any questions or requests on this feature, we will improve our documentation. :-)

site/src/sphinx/client-decorator.rst
```diff
@@ -67,6 +67,116 @@ If your decorator is expected to be reusable, it is recommended to define a new
...
+    });
+
+    // #1 RPC decorator.
+    cb.rpcDecorator((delegate, ctx, rpcReq) -> {
+        System.err.println("Firstly, executed.");
+        ...
+    });
+
+An RPC request is converted into an HTTP request before it's sent to a server.
+Therefore, RPC decorators are inserted before the RPC request is converted and HTTP decorators are inserted
```
ikhoon: How about using `apply` instead of `insert`
For example:
```suggestion
Therefore, RPC decorators are applied before the RPC request is converted and HTTP decorators are applied
```

minwoox: That's a good suggestion. Fixed. 😉

trustin:
Thanks, @minwoox !

minwoox:
Thanks, reviewers!

PR #2570, opened 2020-03-11
Add all HTTP methods to route when the method is empty

minwoox:
Motivation:
Check out this code:
```java
sb.decorator((delegate, ctx, req) -> {
    System.err.println(2);
});
sb.routeDecorator().pathPrefix("/").build((delegate, ctx, req) -> {
    System.err.println(1);
});
sb.decoratorUnder("/", (delegate, ctx, req) -> {
    System.err.println(3);
});
```
The executed order of decorators are a little messed becuase `routeDecorator()` has all `HttpMethod`s set, but others don't.
If we set `HttpMethods.knownMethods()` when there's no method set for a route, the execution order of decorator will be the order they inserted.

Modifications:
- Add `HttpMethod.knownMethods()` to the `Route` if it's empty.
  - However, the logger name and metric don't include it because it's verbose.

Result:
- Right decoration order.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
+++ b/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
@@ -59,8 +59,8 @@ final class DefaultRoute implements Route {
                  List<RoutingPredicate<QueryParams>> paramPredicates,
                  List<RoutingPredicate<HttpHeaders>> headerPredicates) {
         this.pathMapping = requireNonNull(pathMapping, "pathMapping");
-        checkArgument(!methods.isEmpty(), "methods is empty.");
-        this.methods = Sets.immutableEnumSet(requireNonNull(methods, "methods"));
+        checkArgument(!requireNonNull(methods, "methods").isEmpty(), "methods is empty.");
+        this.methods = Sets.immutableEnumSet(methods);
         this.consumes = ImmutableSet.copyOf(requireNonNull(consumes, "consumes"));
         this.produces = ImmutableSet.copyOf(requireNonNull(produces, "produces"));
         this.paramPredicates = ImmutableList.copyOf(requireNonNull(paramPredicates, "paramPredicates"));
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
```diff
@@ -244,7 +244,9 @@ private static String generateLoggerName(String prefix, Set<HttpMethod> methods,
                                              List<RoutingPredicate<HttpHeaders>> headerPredicates) {
         final StringJoiner name = new StringJoiner(".");
         name.add(prefix);
-        if (!methods.isEmpty()) {
+
+        // Skip if the methods is empty or it's knownMethods because it's verbose.
+        if (!methods.isEmpty() && methods != HttpMethod.knownMethods()) {
```
ikhoon: (Maybe) don't we need to update `methods.isEmpty()` condition in `apply(RoutingContext routingCtx)`?
https://github.com/line/armeria/blob/3bc7bbd9363fef1a3b6d42569222ee5a4d69679d/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java#L99

minwoox: Indeed! Thanks.

core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java
```diff
@@ -58,6 +59,7 @@
                  List<RoutingPredicate<QueryParams>> paramPredicates,
                  List<RoutingPredicate<HttpHeaders>> headerPredicates) {
         this.pathMapping = requireNonNull(pathMapping, "pathMapping");
+        checkArgument(!methods.isEmpty(), "methods is empty.");
         this.methods = Sets.immutableEnumSet(requireNonNull(methods, "methods"));
```
trustin: You have to check null first.

minwoox: Oops 😅

PR #2571, opened 2020-03-11
Revamp `RequestContext.set{Request,Response}Timeout()`

### LINKED ISSUE #2535: Undeprecate `ctx.setRequest/ResponseTimeout()`?

ikhoon:
…ions.

Motivation:

The new timeout API was introduced in Armeria 0.98.0 and `set*Timeout{Millis}` has been deprecated. #2343
However, we got some internal inquiries about how to replace the `setTimeout{Millis}`.
It is hard to implement with the existing API.
On the other hand, some of the users are confused about the new timeout API.
See #2535 for the detail discussion.
I think we need to clean and revamp the timeout API before 1.0

Modifications:

- Undeprecate `set*Timeout{Millis}`
- Deprecate `set*TimeoutAt{Millis}` without a replacement
- Deprecate `set*TimeoutAfter{Millis}` in favor of `setTimeout{Millis}(TimeoutMode.FROM_NOW, ...)`
- Deprecate `extend*Timeout{Millis}` in favor of `setTimeout{Millis}(TimeoutMode.EXTEND, ...)`
- Add TimeoutMode enum
- Migrate deprecated API
- Use default method to deduplicate code

Result:

- `set*Timeout{Millis}` is revived.
- You can now schedule a timeout with `TimeoutMode`.
- Fixes #2535

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
@@ -439,9 +439,9 @@ public interface ClientRequestContext extends RequestContext {
      *
      * <table>
      * <tr><th>Timeout mode</th><th>description</th></tr>
-     * <tr><td>{@link TimeoutMode#FROM_NOW}</td>
+     * <tr><td>{@link TimeoutMode#SET_FROM_NOW}</td>
      *     <td>Sets a given amount of timeout from the current time.</td></tr>
-     * <tr><td>{@link TimeoutMode#FROM_START}</td>
+     * <tr><td>{@link TimeoutMode#SET_FROM_START}</td>
      *     <td>Sets a given amount of timeout since the current {@link Response} began processing.</td></tr>
      * <tr><td>{@link TimeoutMode#EXTEND}</td>
      *     <td>Extends the previously scheduled timeout by the given amount of timeout.</td></tr>
@@ -451,13 +451,13 @@ public interface ClientRequestContext extends RequestContext {
      * <pre>{@code
      * ClientRequestContext ctx = ...;
      * // Schedules a timeout from the start time of the response
-     * ctx.setResponseTimeoutMillis(TimeoutMode.FROM_START, 2000);
+     * ctx.setResponseTimeoutMillis(TimeoutMode.SET_FROM_START, 2000);
      * assert ctx.responseTimeoutMillis() == 2000;
-     * ctx.setResponseTimeoutMillis(TimeoutMode.FROM_START, 1000);
+     * ctx.setResponseTimeoutMillis(TimeoutMode.SET_FROM_START, 1000);
      * assert ctx.responseTimeoutMillis() == 1000;
      *
      * // Schedules timeout after 3 seconds from now.
-     * ctx.setResponseTimeoutMillis(TimeoutMode.FROM_NOW, 3000);
+     * ctx.setResponseTimeoutMillis(TimeoutMode.SET_FROM_NOW, 3000);
      *
      * // Extends the previously scheduled timeout.
      * long oldResponseTimeoutMillis = ctx.responseTimeoutMillis();
@@ -475,7 +475,7 @@ public interface ClientRequestContext extends RequestContext {
      * Note that the specified {@code responseTimeoutMillis} must be positive.
      * This value is initially set from {@link ClientOption#RESPONSE_TIMEOUT_MILLIS}.
      * This method is a shortcut for
-     * {@code setResponseTimeoutMillis(TimeoutMode.FROM_NOW, responseTimeoutMillis)}.
+     * {@code setResponseTimeoutMillis(TimeoutMode.SET_FROM_NOW, responseTimeoutMillis)}.
      *
      * <p>For example:
      * <pre>{@code
@@ -487,7 +487,7 @@ public interface ClientRequestContext extends RequestContext {
      * @param responseTimeoutMillis the amount of time allowed in milliseconds from now
      */
     default void setResponseTimeoutMillis(long responseTimeoutMillis) {
-        setResponseTimeoutMillis(TimeoutMode.FROM_NOW, responseTimeoutMillis);
+        setResponseTimeoutMillis(TimeoutMode.SET_FROM_NOW, responseTimeoutMillis);
     }
 
     /**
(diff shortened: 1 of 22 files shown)


core/src/main/java/com/linecorp/armeria/common/util/TimeoutMode.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.linecorp.armeria.common.util;
+
+import com.linecorp.armeria.common.Request;
+
+/**
+ * The timeout mode.
+ */
+public enum TimeoutMode {
```
anuraaga: Any thoughts on my comment here?

https://github.com/line/armeria/issues/2535#issuecomment-593757305

It prevents having adjective (from) and verb (extend) in the same enum which looks weird.

ikhoon: That's a good suggestion. 👍 I know that our previous discussion does not reach a decision.
I'm still very open to this enum. 
The word meaning of `NOW` and `CURRENT` are similar. So I hesitated to use `CURRENT` a little bit.

How about adding two kinds of API.(I'm not sure this is better.)
```java
public TimeoutMode {
    NOW,
    START
}
// Schedules timeout from specific timeout mode
void setTimeout(TimeoutMode, Duration)
// Extends the previous scheduled timeout
void setTimeoutAfter(Duration)
```

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -444,17 +485,57 @@ ClientRequestContext newDerivedContext(RequestId id, @Nullable HttpRequest req,
      * @param responseTimeoutMillis the amount of time allowed in milliseconds from
      *                              the beginning of the response
      *
-     * @deprecated Use {@link #extendResponseTimeoutMillis(long)}, {@link #setResponseTimeoutAfterMillis(long)},
-     *             {@link #setResponseTimeoutAtMillis(long)} or {@link #clearResponseTimeout()}
      */
-    @Deprecated
-    void setResponseTimeoutMillis(long responseTimeoutMillis);
+    default void setResponseTimeoutMillis(long responseTimeoutMillis) {
+        setResponseTimeoutMillis(TimeoutMode.FROM_START, responseTimeoutMillis);
```
trustin: Perhaps the sensible default would be `TimeoutMode.FROM_NOW`?

core/src/main/java/com/linecorp/armeria/common/util/TimeoutMode.java
trustin: How about `FROM_OLD_TIMEOUT` or `FROM_PREV_TIMEOUT` instead of `EXTEND` then? Or we could make the two verbs, like `SET_FROM_NOW` and `SET_FROM_START`. I think the latter sounds better.

core/src/main/java/com/linecorp/armeria/internal/common/TimeoutScheduler.java
```diff
@@ -61,6 +66,16 @@ public void clearTimeout() {
...
     }
 
+    public void setTimeoutMillis(TimeoutMode mode, long timeoutMillis) {
+        if (mode == FROM_NOW) {
+            setTimeoutAfterMillis(timeoutMillis);
+        } else if (mode == FROM_START) {
+            setTimeoutMillis(timeoutMillis);
+        } else if (mode == EXTEND) {
+            extendTimeoutMillis(timeoutMillis);
+        }
```
trustin: Could use `switch-case`

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
ikhoon: Oops... thanks for pointing it out!🙏

core/src/main/java/com/linecorp/armeria/common/util/TimeoutMode.java
anuraaga: Making all verbs with e.g. `SET_*` sounds good.

ikhoon: `SET_*` is explicit, but we alread have `set` verb in method name.
```java
ctx.setTimeout(TimeoutMode.SET_FROM_NOW, Duration.ofSeconds(1))
```

Maybe we could remove `set` from method name?
```java
ctx.timeout(TimeoutMode.SET_FROM_NOW, Duration.ofSeconds(1))
ctx.timeout(TimeoutMode.SET_FROM_START, Duration.ofSeconds(1))
ctx.timeout(TimeoutMode.EXTEND, Duration.ofSeconds(1))
```

anuraaga: Removing `set` from method names SGTM

core/src/main/java/com/linecorp/armeria/internal/common/TimeoutScheduler.java
```diff
@@ -61,6 +63,20 @@ public void clearTimeout() {
...
+            case SET_FROM_START:
+                setTimeoutMillis(timeoutMillis);
+                break;
+            case EXTEND:
+                extendTimeoutMillis(timeoutMillis);
+                break;
+        }
+    }
+
     public void setTimeoutMillis(long timeoutMillis) {
```
minwoox: this could be private

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -439,45 +439,16 @@ public void clearResponseTimeout() {
...
-    @Override
-    public void setResponseTimeoutAfterMillis(long responseTimeoutMillis) {
-        timeoutScheduler.setTimeoutAfterMillis(responseTimeoutMillis);
-    }
-
-    @Override
-    public void setResponseTimeoutAfter(Duration responseTimeout) {
-        setResponseTimeoutAfterMillis(requireNonNull(responseTimeout, "responseTimeout").toMillis());
+    public void setResponseTimeoutMillis(TimeoutMode mode, long responseTimeoutMillis) {
+        timeoutScheduler.setTimeoutMillis(mode, responseTimeoutMillis);
```
minwoox: nit: null check

trustin:
Thanks, @ikhoon!

PR #2572, opened 2020-03-12
Do not warn on unexpected exceptions in annotated services

trustin:
Motivation:

An annotated sevice currently logs all exceptions except the following
by default:

- `IllegalArgumentException`
- `HttpStatusException`
- `HttpResponseException`

However, it logs some ignorable socket-related exceptions, such as
`ClosedSessionException`.

Modifications:

- Check the exception with `Exceptions.isExpected()` before logging.

Result:

- Reduced amount of meaningless log messages


trustin:
Thanks for reviewing.

PR #2577, opened 2020-03-12
Allow `UUID` to be bound to annotated service parameters

### LINKED ISSUE #2573: UUID as annotated rest service parameters
We use a lot of `UUID` as parameters in paths. As Armeria doesn't support converting `String` -> `UUID` in annotated rest services, we get a lot of duplicated code and variable scope pollution.

I *believe* this to be a pretty simple fix, but the question is if there are more simple types that also needs parsing. `Instant` is not one of those imho, but the opposite could be argued.

ikhoon:
Motivation:
`UUID` is commonly used to represent a unique ID.
However `UUID` is not included in the supported types by default.

Modifications:

- Add UUID to `supportedElementTypes` in `AnnotatedServiceTypeUtil`

Result:
You can now bind `UUID` to path variables or query parameters.
Fixes #2573


trustin:
Perhaps documentation needs an update as well?

ikhoon:
Sure! will update it :-)

PR #2580, opened 2020-03-13
Specify why on `Closed{Session,Stream}Exception`

trustin:
Motivation:

It is sometimes hard to know why a connection or a stream has been
closed or reset.

Modifications:

- Allow a user specify a message or a cause when creating a
  `Closed{Session,Stream}Exception`.
- Specify a message or a cause when creating `Closed{Session,Stream}Exception`
  when applicable.
- Removed redundant instantiation of `ClosedSessionException`s.
- Fixed a bug where `ClosedSessionException` is raised when
  `ClosedStreamException` is expected.

Result:

- More information when diagnosing a `Closed{Session,Stream}Exception`
- Fixed a bug where `ClosedSessionException` is raised when
  `ClosedStreamException` is expected.

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
@@ -22,6 +22,7 @@ import static com.linecorp.armeria.common.SessionProtocol.H2C;
 import static com.linecorp.armeria.common.stream.SubscriptionOption.WITH_POOLED_OBJECTS;
 import static java.util.Objects.requireNonNull;
 
+import java.io.IOException;
 import java.net.SocketAddress;
 import java.util.concurrent.ScheduledFuture;
 
@@ -54,6 +55,7 @@ import io.netty.handler.codec.http2.Http2ConnectionPrefaceAndSettingsFrameWritte
 import io.netty.handler.codec.http2.Http2Settings;
 import io.netty.handler.ssl.SslCloseCompletionEvent;
 import io.netty.handler.ssl.SslHandshakeCompletionEvent;
+import io.netty.util.Attribute;
 import io.netty.util.AttributeKey;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.Promise;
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
```diff
@@ -322,34 +324,53 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
             channelPool.connect(channel.remoteAddress(), H1C, sessionPromise);
         } else {
             // Fail all pending responses.
-            final Throwable throwable;
-            if (ctx.channel().hasAttr(PENDING_EXCEPTION)) {
-                throwable = ctx.channel().attr(PENDING_EXCEPTION).get();
+            final HttpResponseDecoder responseDecoder = this.responseDecoder;
+            final Throwable pendingException;
```
minwoox: We can just set `pendingException` here and use it.

core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
```diff
@@ -322,34 +324,53 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+                pendingException = null;
             }
-            failUnfinishedResponses(throwable);
+
             // Cancel the timeout and reject the sessionPromise just in case the connection has been closed
             // even before the session protocol negotiation is done.
             sessionTimeoutFuture.cancel(false);
-            sessionPromise.tryFailure(throwable);
+            if (!sessionPromise.isDone()) {
+                sessionPromise.tryFailure(pendingException != null ? pendingException
```
minwoox: then we can just set `pendingException` here.

core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
```diff
@@ -322,34 +324,53 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+        return ClosedSessionException.get();
+    }
+
+    static void setPendingException(ChannelHandlerContext ctx, Throwable cause) {
+        final Attribute<Throwable> attr = ctx.channel().attr(PENDING_EXCEPTION);
+        final Throwable previousCause = attr.get();
+        if (previousCause == null) {
+            attr.set(cause);
+        } else if (previousCause != cause) {
+            previousCause.addSuppressed(cause);
```
minwoox: Is there any chance that the `previousCause` is not suppressible?

trustin: It's possible but we need to preserve the type of the original exception, such as `UnprocessedRequestException`.

minwoox: Yes couldn't think of that. :sweat_smile:

trustin: Now just logging, instead of suppressing.

PR #2587, opened 2020-03-16
Make `RequestContextAwareLogger` push context

trustin:
Motivation:

`RequestContextAwareLogger` currently does not push the context when
logging, which means `RequestContextExporter` will not be able to export
the properties to MDC.

Modifications:

- Push the context every time `RequestContextAwareLogger` logs.
- Use `TemporaryThreadLocals` when building a log message.

Result:

- `RequestContextAwareLogger` pushes the context as expected by
  `RequestContextExporter`, and thus the properties are exported to MDC.


trustin:
Thanks for quick reviews.

PR #2591, opened 2020-03-16
Reimplement `RequestContext.toString()`

trustin:
- Use `TemporaryThreadLocals` to reduce `StringBuilder` reallocation.
- Add `RequestId`s
- Move the slow path to a separate method so that fast path can be
  inlined.
- Add `TextFormatter.socketAddress()` and `inetAddress()`
- Miscellaneous:
  - Fix method ordering.
  - Fix indentation
  - Fix for a potential `NPE` on Windows in
    `DefaultServiceRequestContext.{local,remote}Address()`

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
@@ -560,7 +560,6 @@ public final class DefaultClientRequestContext
             return strVal;
         }
         return toStringSlow();
-
     }
 
     private String toStringSlow() {
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
@@ -565,7 +565,8 @@ public final class DefaultClientRequestContext
 
     private String toStringSlow() {
         // Prepare all properties required for building a string representation,
-        // to follow our convention of not calling other methods while appending to TemporaryThreadLocals.
+        // so that we don't have a chance of building two Strings using one StringBuilder
+        // provided by TemporaryThreadLocals.
         final Channel ch = channel();
         final String creqId = id().shortText();
         final String sreqId = root() != null ? root().id().shortText() : null;
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
@@ -563,9 +563,9 @@ public final class DefaultClientRequestContext
     }
 
     private String toStringSlow() {
-        // Prepare all properties required for building a string representation,
-        // so that we don't have a chance of building two Strings using one StringBuilder
-        // provided by TemporaryThreadLocals.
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/linecorp/armeria/server/DefaultServiceRequestContext.java
```diff
@@ -485,48 +488,45 @@ void setRequestTimeoutController(TimeoutController requestTimeoutController) {
...
+        final String authority = virtualHost().defaultHostname();
+        final String path = path();
+        final String method = method().name();
+
+        // Build the string representation.
+        final StringBuilder buf = TemporaryThreadLocals.get().stringBuilder();
+        buf.append("[sreqId=").append(sreqId)
+           .append(", chanId=").append(chanId);
+
+        if (!Objects.equal(caddr, raddr.getAddress())) {
```
anuraaga: Use JDK `Objects.equals()`

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -548,42 +556,46 @@ public RequestLogBuilder logBuilder() {
...
+        return toStringSlow();
 
-        final StringBuilder buf = new StringBuilder(107);
-        buf.append("[C]");
+    }
 
-        // Prepend the current channel information if available.
+    private String toStringSlow() {
+        // Prepare all properties required for building a string representation,
+        // so we do not have any chance of nesting the use of TemporaryThreadLocals.
```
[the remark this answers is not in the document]
trustin: Hmm, I meant not using `TemporaryThreadLocals` while using it in nested `toString()`s.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -548,42 +556,47 @@ public RequestLogBuilder logBuilder() {
...
         }
+        return toStringSlow();
 
-        final StringBuilder buf = new StringBuilder(107);
-        buf.append("[C]");
+    }
 
-        // Prepend the current channel information if available.
+    private String toStringSlow() {
+        // Prepare all properties required for building a string representation,
```
anuraaga: This is less clear again. I think we have a convention of not calling any methods during thread local building because it's not clear whether that method may reinitialize the buffer or not. toString is of course a common culprit but not necessarily the only one.

Actually the javadoc rule of thumb doesn't seem strong enough, it's not really an issue of recursion and just about whether other methods may initialize a buffer so the general rule of thumb is to call no other methods unless they accept a `StringBuilder` as a parameter.
https://github.com/line/armeria/blob/9061e241961c806951dd855be65fd871131a1835/core/src/main/java/com/linecorp/armeria/internal/common/util/TemporaryThreadLocals.java#L47

trustin: Updated again, this time `TemporaryThreadLocals` as well.

trustin:
Let me merge this since we need a new release real soon and the updated Javadoc seems pretty clear to me. Please feel free to open a follow-up issue or PR to improve further.

PR #2592, opened 2020-03-16
Change CountingSampler to accept float instead of double

KarboniteKream:
Motivation:

`CountingSampler` is currently using `double` for probability, while [Brave](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/sampler/CountingSampler.java#L60) (where this class was originally forked from) is using `float`. This causes issues when passing in floats like `0.01f`, which is converted to `double` as `0.009999999776482582`. This causes `percent` to be `0`.

Modification:

- Change `CountingSampler` to accept a `float`.

Result:

When passing in the probability of `0.01f`, the percentage is correctly calculated as `1`.

### DIFF
--- a/core/src/test/java/com/linecorp/armeria/common/util/CountingSamplerTest.java
+++ b/core/src/test/java/com/linecorp/armeria/common/util/CountingSamplerTest.java
@@ -48,7 +48,7 @@ public class CountingSamplerTest {
     @Test
     public void testSamplingRatePercentageRounding() {
         assertThat(CountingSampler.create(0.01f)).isInstanceOfSatisfying(CountingSampler.class, sampler -> {
-            assertThat(sampler.sampleDecisions.cardinality()).isEqualTo(10);
+            assertThat(sampler.sampleDecisions.cardinality()).isEqualTo(1);
         });
     }
 
--- a/core/src/test/java/com/linecorp/armeria/common/util/CountingSamplerTest.java
+++ b/core/src/test/java/com/linecorp/armeria/common/util/CountingSamplerTest.java
@@ -36,22 +36,28 @@ import org.junit.jupiter.api.Test;
 
 public class CountingSamplerTest {
     @Test
-    public void testSamplingRateMinimumLimit() throws Exception {
-        assertThatThrownBy(() -> CountingSampler.create(-1.99)).isInstanceOf(IllegalArgumentException.class);
+    public void testSamplingRateMinimumLimit() {
+        assertThatThrownBy(() -> CountingSampler.create(-1.99f)).isInstanceOf(IllegalArgumentException.class);
     }
 
     @Test
-    public void testSamplingRateMaximumLimit() throws Exception {
-        assertThatThrownBy(() -> CountingSampler.create(1.01)).isInstanceOf(IllegalArgumentException.class);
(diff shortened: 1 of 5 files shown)


core/src/test/java/com/linecorp/armeria/common/util/CountingSamplerTest.java
```diff
@@ -36,22 +36,29 @@
...
+    public void testSamplingRateMaximumLimit() {
+        assertThatThrownBy(() -> CountingSampler.create(1.01f)).isInstanceOf(IllegalArgumentException.class);
     }
 
     @Test
-    public void testNeverSampledSampler() throws Exception {
-        assertThat(CountingSampler.create(0.0)).isSameAs(Sampler.never());
+    public void testSamplingRatePercentageRounding() {
+        assertThat(CountingSampler.create(0.01f)).isInstanceOfSatisfying(CountingSampler.class, sampler -> {
+            assertThat(sampler.sampleDecisions.cardinality()).isEqualTo(10);
```
minwoox: Isn't this 1?

KarboniteKream: Just noticed it, thanks.

core/src/test/java/com/linecorp/armeria/common/util/CountingSamplerTest.java
```diff
@@ -36,22 +36,29 @@
...
+    public void testSamplingRateMaximumLimit() {
+        assertThatThrownBy(() -> CountingSampler.create(1.01f)).isInstanceOf(IllegalArgumentException.class);
     }
 
     @Test
-    public void testNeverSampledSampler() throws Exception {
-        assertThat(CountingSampler.create(0.0)).isSameAs(Sampler.never());
+    public void testSamplingRatePercentageRounding() {
+        assertThat(CountingSampler.create(0.01f)).isInstanceOfSatisfying(CountingSampler.class, sampler -> {
+            assertThat(sampler.sampleDecisions.cardinality()).isEqualTo(1);
```
minwoox: nit: you could just do `isOne()`.

anuraaga:
Do you think it's worth validating the parameter only has two decimal places - I think `BigDecimal.BigDecimal.scale()` is an easy way and we wouldn't even need to multiply by 100 can just use the unscaledValue. A parameter with more that two decimals will always drop precision.

KarboniteKream:
@anuraaga You mean something like `BigDecimal.valueOf(probability).setScale(2, RoundingMode.HALF_UP).unscaledValue()`?

trustin:
Let me merge this for now. Please continue the discussion and create a follow-up PR if necessary.

anuraaga:
Similar but I think instead of `setScale` we could throw an error if the input scale is greater than two to indicate that precision will be lost. Not a big deal just an idea.

KarboniteKream:
I think it's doable, but it might not be so short. For example `BigDecimal.valueOf(0.01f).scale()` returns 18, because it will already try to do some conversion, so we would need to handle cases like this.

anuraaga:
Ah thanks for checking, yeah that won't work well since it only accepts `double` so has the same widening problem. I guess instead just using `Float.toString` is the only precise way. Maybe overkill 😅

PR #2597, opened 2020-03-17
Assign stream ID right before writing RequestHeaders

minwoox:
Motivation:
When sending requests asynchronously, the requests might not be sending in order due to DNS resolution and HTTP/2 multiplexing.
If it happens after stream ID is assigned, the request, which tries to create a new stream after a stream whose ID is bigger is created, couldn't be used.

Modifications:
- Assign the stream ID right before writing `RequestHeaders`.

Result:
- You no longer see `ClosedStreamException` due to the stream creation reversal

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
+++ b/core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
@@ -16,6 +16,8 @@
 
 package com.linecorp.armeria.client;
 
+import static com.linecorp.armeria.client.HttpSessionHandler.MAX_NUM_REQUESTS_SENT;
+
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
@@ -45,7 +47,6 @@ import com.linecorp.armeria.internal.common.RequestContextUtil;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
-import io.netty.channel.EventLoop;
 import io.netty.handler.codec.http2.Http2Error;
 import io.netty.util.ReferenceCountUtil;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -122,43 +129,53 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
+            return;
+        }
 
-        final EventLoop eventLoop = ch.eventLoop();
+        final HttpSession session = HttpSession.get(ch);
+        id = session.getAndIncrementNumRequestsSent();
+        if (!session.canSendRequest() || id >= MAX_NUM_REQUESTS_SENT) {
+            responseDecoder.disconnectWhenFinished();
+            // No need to send RST because we didn't send any packet and this will be disconnected anyway.
+            fail(new UnprocessedRequestException(ClosedSessionException.get()));
```
trustin: How about adding a message to the exception? Could use different message for the two cases (`!session.canSendRequest()` and `id >= MAX_NUM_REQUESTS_SENT`)

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -298,7 +322,7 @@ private void cancelSubscription() {
         subscription.cancel();
     }
 
-    private void failAndRespond(Throwable cause) {
+    private void failAndWriteResetIfActive(Throwable cause) {
```
trustin: I think `failAndReset()` is enough. Whether a channel is active or not doesn't matter much because the RST_STREAM frame would not be sent even if we didn't have `if (ch.isActive())`.

core/src/main/java/com/linecorp/armeria/client/HttpSession.java
```diff
@@ -63,6 +63,11 @@ public void retryWithH1C() {
 
         @Override
         public void deactivate() {}
+
+        @Override
+        public int getAndIncrementNumRequestsSent() {
+            return 0;
```
trustin: Maybe better returning `MAX_NUM_REQUESTS_SENT`?

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
minwoox: Thanks fixed. 😄

core/src/main/java/com/linecorp/armeria/client/HttpSession.java
minwoox: Right!

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
minwoox: Specified the exception messages. Thanks!

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -218,7 +242,12 @@ public void onNext(HttpObject o) {
     @Override
     public void onError(Throwable cause) {
         isSubscriptionCompleted = true;
-        failAndRespond(cause);
+        if (id >= 0) { // onSubscribe is called.
+            failAndReset(cause);
+        } else {
+            // No need to send RST because we didn't send any packet.
+            fail(cause);
```
ikhoon: Don't we need to wrap this cause with `UnprocessedRequestException?

minwoox: Yes, right. Thanks!

core/src/main/java/com/linecorp/armeria/client/HttpSessionHandler.java
```diff
@@ -146,37 +146,41 @@ public boolean canSendRequest() {
...
 
-        if (numRequestsSent >= MAX_NUM_REQUESTS_SENT) {
-            responseDecoder.disconnectWhenFinished();
-            return false;
-        } else {
-            return true;
-        }
+    @Override
+    public int getAndIncrementNumRequestsSent() {
+        return ++numRequestsSent;
```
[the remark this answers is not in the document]
minwoox: Fixed. 😉

core/src/main/java/com/linecorp/armeria/client/HttpRequestSubscriber.java
```diff
@@ -218,7 +242,12 @@ public void onNext(HttpObject o) {
     @Override
     public void onError(Throwable cause) {
         isSubscriptionCompleted = true;
-        failAndRespond(cause);
+        if (id >= 0) { // onSubscribe is called.
+            failAndReset(cause);
+        } else {
+            // No need to send RST because we didn't send any packet.
+            fail(new UnprocessedRequestException(cause));
```
ikhoon: Question: We don't need to care about the type of cause just in case the cause is instance of `UnprocessedRequestException` already? 
If the cause can not be `UnprocessedRequestException`, skip this review 😀

minwoox: I think that's not going to happen. If so, I should unwrap the cause and call `failAndReset` on line 246. I think it's not worth. 😄

ikhoon: Thanks for the explanation! 🙏

minwoox:
Thanks for reviewing!

PR #2599, opened 2020-03-17
Provide a way to access a parent log from a child log

### LINKED ISSUE #2588: Provide a way to get RequestId of the parent ctx while retrying
It will be good if a user is able to get to know the `RequestId` of the parent `ClientRequestContext` from the child when retrying.

ikhoon:
Motivation:

`RequestId` is used for tracking the request and the response pair.
A request can be executed multiple times by `RetryingClient`.
It is difficult to group and trace all retried requests without the request ID of the original request.
By accessing the parent log from the current request context, we are able to log the parent and current request ID pair.

Modifications:

- Add `parent()` to `RequestLogAccess`
- Include the request ID of parent log to `toString()` in `DefaultClientRequestContext`

Results:

- You can now access a parent log from a child log.
- `DefaultClientRequest.toString()` contains a parent request ID if available.
- Fixes #2588

### DIFF
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
@@ -31,6 +31,7 @@ import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.RequestContextWrapper;
 import com.linecorp.armeria.common.RequestId;
 import com.linecorp.armeria.common.RpcRequest;
+import com.linecorp.armeria.common.logging.RequestLogAccess;
 import com.linecorp.armeria.server.ServiceRequestContext;
 
 import io.netty.util.AttributeKey;
--- a/core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
+++ b/core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
@@ -31,7 +31,6 @@ import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.RequestContextWrapper;
 import com.linecorp.armeria.common.RequestId;
 import com.linecorp.armeria.common.RpcRequest;
-import com.linecorp.armeria.common.logging.RequestLogAccess;
 import com.linecorp.armeria.server.ServiceRequestContext;
 
 import io.netty.util.AttributeKey;
--- a/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
+++ b/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
@@ -559,36 +559,24 @@ public final class DefaultClientRequestContext
 
     @Override
     public String toString() {
-        if (strVal == null) {
-            strVal = toStringSlow();
-            if (channel() != null) {
-                strValAvailabilities |= STR_CHANNEL_AVAILABILITY;
-            }
-            if (log().parent() != null) {
-                strValAvailabilities |= STR_PARENT_LOG_AVAILABILITY;
-            }
+        final Channel ch = channel();
+        final RequestLogAccess parent = log().parent();
+        final short newAvailability =
+                (short) ((ch != null ? STR_CHANNEL_AVAILABILITY : 0) |
+                         (parent != null ? STR_PARENT_LOG_AVAILABILITY : 0));
+        if (strVal != null && strValAvailabilities == newAvailability) {
             return strVal;
         }
 
-        boolean dirty = false;
-        if (channel() != null && (strValAvailabilities & STR_CHANNEL_AVAILABILITY) == 0) {
-            dirty = true;
-            strValAvailabilities |= STR_CHANNEL_AVAILABILITY;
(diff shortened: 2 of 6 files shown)


core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
```diff
@@ -31,6 +31,7 @@
 import com.linecorp.armeria.common.RequestContextWrapper;
 import com.linecorp.armeria.common.RequestId;
 import com.linecorp.armeria.common.RpcRequest;
+import com.linecorp.armeria.common.logging.RequestLogAccess;
```
trustin: Checkstyle?

core/src/main/java/com/linecorp/armeria/common/logging/DefaultRequestLog.java
```diff
@@ -420,6 +428,11 @@ private boolean isDeferredFlagSet(RequestLogProperty property) {
     public void addChild(RequestLogAccess child) {
         checkState(!hasLastChild, "last child is already added");
         requireNonNull(child, "child");
+
+        if (child instanceof DefaultRequestLog) {
+            ((DefaultRequestLog) child).parent = this;
```
trustin: Could you add `checkState` that makes sure that `parent` is `null`?

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -556,18 +556,30 @@ public RequestLogBuilder logBuilder() {
...
     @Override
     public String toString() {
+        final String clientRequestIdStr = toStringClientRequestId();
         if (strVal != null) {
-            return strVal;
+            return clientRequestIdStr + strVal;
         }
-        return toStringSlow();
+        return clientRequestIdStr + toStringSlow();
+    }
```
trustin: This will create a new `clientReqIdStr` every time. How about adding some state variable to minimize the regeneration of strings? e.g. You could create a small flag that keeps whether 1) `channel` is available and 2) `log.parent` is available, so `strVal` is regenerated when the flag changes.

ikhoon: Oh, that sounds nice. Let me address it!

core/src/main/java/com/linecorp/armeria/client/ClientRequestContextWrapper.java
ikhoon: Oops..

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -556,30 +559,37 @@ public RequestLogBuilder logBuilder() {
...
-        if (strVal != null) {
-            return clientRequestIdStr + strVal;
+        if (strVal == null) {
+            strVal = toStringSlow();
+            if (channel() != null) {
+                strValAvailabilities |= STR_CHANNEL_AVAILABILITY;
+            }
+            if (log().parent() != null) {
+                strValAvailabilities |= STR_PARENT_LOG_AVAILABILITY;
+            }
```
trustin: How about simplifying like this:

```java
final Channel ch = channel();
final RequestLogAccess parent = log().parent();
final short newAvailability = (ch != null ? ... : 0) | (parent != null ? ... : 0);
if (strVal != null && strValAvailabilities == newAvailability) {
    return strVal;
}

strValAvailabilities = newAvailability;
return toStringSlow(ch, parent);
```

Keep in mind that `toStringSlow()` does this:

```java
        if (ch != null) {
            this.strVal = strVal;
        }
```

.. and you must change it to make it work properly. Add tests properly so that this kind of mistake doesn't slip in.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
```diff
@@ -578,6 +588,9 @@ private String toStringSlow() {
         // Build the string representation.
         final StringBuilder buf = TemporaryThreadLocals.get().stringBuilder();
         buf.append("[creqId=").append(creqId);
+        if (parent != null) {
+            buf.append(", pcreqId=").append(pcreqId);
+        }
         if (sreqId != null) {
             buf.append(", sreqId=").append(sreqId);
         }
```
trustin: How about just `preqId`?

trustin: I meant renaming `pcreqId` to `preqId`, just in case it wasn't clear.

ikhoon: I thought the `preqId` too. `preqId` sounds good to me. Let's go with `preqId`.

core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java
ikhoon: Thanks for the gentle suggestion. I've learned a lot. 🙏

trustin:
Thanks a lot, @ikhoon !

ikhoon:
Thanks for reviewing!

Issue #2054, opened 2019-09-08
Migrate TSlint to ESlint according to deprecation of TSlint


taejs:
https://github.com/palantir/tslint/issues/4534

TSLint is officially announced to be deprecated in 2019,
Now docs-client uses TSLint, so I think it would be great to migrate to ESLint.

ikhoon:
That's a good idea. I’d like to listen to other folks opinion. :-)

anuraaga:
Yup I think it's a good idea - it's been in my mind's backlog for a while.

PR #2604, opened 2020-03-18
Switch from TSLint to ESLint

trustin:
Related issues: #2054

Motivation:

TSLint is officially announced to be deprecated in 2019.

Modifications:

- Replace TSLint with ESLint
- Place `eslintrc.js` in `settings/eslint` so that other frontend
  modules (e.g. Gatsby) can use it as well.

Result:

- New and shiny linting


docs-client/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+const defaultConfig = require('../settings/eslint/eslintrc');
+
+// Ignore 'webpack.config.ts' so that it doesn't complain that
+// some dependencies must be moved from devDependencies to dependencies.
```
anuraaga: We could also just move everything into `dependencies`. For apps not published to NPM it's one reasonable convention, which is followed by `create-react-app`

trustin:
/cc @taejs @anuraaga 

The new ESLint settings are not same with our old rules. As a result, I see many errors. I think I'm gonna fix them all in this PR, but please let me know if you think it's better disabling a certain rule.

settings/eslint/eslintrc.js
```diff
@@ -0,0 +1,66 @@
+const commonParserOptions = {
+  ecmaFeatures: {
+    jsx: true,
+  },
+  ecmaVersion: '2020',
+  sourceType: 'module',
+};
+
+const commonRules = {
+  'prettier/prettier': [
```
anuraaga: Let's put these in `.prettierrc.js` since it will have better support with IDEs

settings/eslint/eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+    'error',
+    {
+      singleQuote: true,
+      jsxSingleQuote: true,
+      trailingComma: 'all',
+    },
+  ],
```
trustin: This is the only customization I did except applying the recommended presets below.

settings/eslint/eslintrc.js
```diff
@@ -0,0 +1,66 @@
+const commonParserOptions = {
+  ecmaFeatures: {
```
anuraaga: Are these options needed? I've never seen them defined without losing any linting

settings/eslint/eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+module.exports = {
+  globals: {
+    Atomics: 'readonly',
+    SharedArrayBuffer: 'readonly',
+  },
+  ignorePatterns: ['node_modules/', 'public/', 'build/', '!.*.js', '!.*.json'],
+  // Settings for .js, .jsx and .json
+  parser: 'babel-eslint',
```
anuraaga: Maybe better to leave this as default

settings/eslint/eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+    {
+      singleQuote: true,
+      jsxSingleQuote: true,
+      trailingComma: 'all',
+    },
+  ],
+};
+
+module.exports = {
+  globals: {
```
anuraaga: Instead of globals, does this work?

```js
env: {
  browser: true, // Since we run in browser
  jest: true, // Since we wish we had unit tests
}
```

docs-client/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+const defaultConfig = require('../settings/eslint/eslintrc');
```
anuraaga: I think instead of javascript, it could be better to just use eslint's default functionality with `extends`

trustin: I tried, but it seems like it fails to load the configs in the `extends` section, e.g. It fails to find `airbnb`. I guess it looks for the configs relatively from `../settings/eslint/` rather than from `.`.

docs-client/.eslintrc.js
trustin: That would be convenient, but I'm not sure if it's a good idea to move webpack and its plugins to `dependencies` because they are used only when building obviously.

settings/eslint/eslintrc.js
trustin: Seems to work. Thanks.

settings/eslint/eslintrc.js
trustin: Yeah, the default parser works fine. Cargo-culted from `eslint-plugin-json` :man_facepalming:

settings/eslint/eslintrc.js
trustin: Seems to work without `ecmaFeatures`

trustin:
@anuraaga wrote:
> I find disabling prefer-destructuring, no-continue, and tweaking no-restricted-syntax to be particularly important.

If we disable `prefer-destructuring`, should we also disable `react/destructuring-assignment`?

settings/eslint/eslintrc.js
trustin: Done. Thanks for suggestion.

docs-client/.eslintrc.js
anuraaga: Technically everything is only when building since we build a bundle rather than publishing anything - I think that's why create-react-app follows that convention. But it's ok if it's easy to ignore files too

settings/eslint/eslintrc.js
```diff
@@ -0,0 +1,82 @@
...
+
+const commonRules = {
+  'class-methods-use-this': 'off',
+  'eslint-comments/no-unused-disable': 'error',
+  'import/export': 'off',
+  'import/no-cycle': 'off',
+  'import/no-named-as-default': 'off',
+  'import/no-useless-path-segments': 'off',
+  'import/prefer-default-export': 'off',
+  'prefer-destructuring': 'off',
```
anuraaga: Ah yeah it's true probably need the react version as well when disabling this.

settings/prettier/prettierrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  singleQuote: true,
+  jsxSingleQuote: true,
```
anuraaga: Ah just noticed this - I think this is a pretty rare configuration since JSX is basically XML. Up to you but I think most projects use double quote in JSX.

trustin: Isn't JSX not an XML? It allows things like `className={myClassName}`. I thought it's good for consistency because otherwise it would mean two different single quote rules in a single .jsx file. What do you think?

anuraaga: Good point that it's not true XML. But I've never seen a project use this flag, including our two main dependencies material-ui and react. So I'd say it's not conventional to use jsxSingleQuote. Considering the quotes are used in totally different ways (const foo = 'bar' vs `<Component foo="bar">`, I guess it's rare to cause confusion.

docs-client/.eslintrc.js
trustin: Merged into `dependencies` for simplicity

settings/prettier/prettierrc.js
trustin: Removed `jsxSingleQuote: true` and reverted auto-fixes

trustin:
@anuraaga Merging `devDependencies` into `dependencies` increased the size of `web-licenses.txt`. Better reverting?

anuraaga:
@trustin Ah yeah forgot about that file - yeah better to split then :)

trustin:
Build now passes. :sparkles:

anuraaga:
LGTM - sorry for back and forth!

trustin:
Thanks for reviewing. Very educational experience.

PR #2608, opened 2020-03-18
Make `Client` and `Service` decorator always override `as()` or `serviceAdded()`

### LINKED ISSUE #1958: Refuse to use a decorating `Client` and `Service` if `as()` or `serviceAdded()` is not overriden.
A common mistake a user makes when writing a decorator is:

```java
sb.service("/svc", myService.decorate(delegate -> (ctx, req) -> {
    ...
}));
```

The above decorator is not correct because the returned service does not override the default `as()` and `serviceAdded()` implementation. A correct usage is:

```java
sb.service("/svc", myService.decorate((delegate, ctx, req) -> {
});
// or
sb.service("/svc", myService.decorate(delegate -> new SimpleDecoratingService(delegate) {
    ...
    ...
}));
```

We can detect such a mistake by checking if the decorator overrides the default `as()` and `serviceAdded()` implementation using Java reflection API and raise an exception.

ikhoon:
Motivation:

If a user writes a decorator with curried lambda style,
the given decorator will not override `as()` or `serviceAdded()`. #1958

```java
sb.service("/svc", myService.decorate(delegate -> (ctx, req) -> {
    ...
}));
```
The decorated service will not call `ServiceCallbackInvoker.invokeServiceAdded()` properly.

Modifications:

- Add `DecoratorUtil` for validating decorators using reflection.
- Add the validation logic to `{Http,Rpc}Service.decorate()`, `ClientDecoration.*decorate()`, `RouteDecoratingService()`.

Result:

You can now get `IllegalArgumentException` when the specified decorator does not override `as()` or `serviceAdded()`
Fixes #1958


anuraaga:
Is it not possible to define an interface with `as` to catch at compile time?

ikhoon:
> Is it not possible to define an interface with as to catch at compile time?

That sounds good. I'm not sure, but let me try whether it is possible. 😀

trustin:
I've just realized this stops a user from using `Function.identity()` or other creative use of functions, like returning a completely different `Client` or `Service` which is not a decorator.

That means, I think we can't validate anything this way. :sob:

ikhoon:
> other creative use of functions, like returning a completely different Client or Service which is not a decorator.

Ah... I don’t think of that. That is important point. 
I will close this PR. If we have another idea on this, will make a PR. 😀

trustin:
> Is it not possible to define an interface with as to catch at compile time?

Then a user can't write a service using a lambda expression.

anuraaga:
@trustin Yeah I thought if the return value of the decorator lambda was an interface with `as` (which `SimpleDecoratingService` would implement) it'd be checked and allow the syntax in the attached issue. But anyways, sounds like there are other issues 😅

PR #2610, opened 2020-03-19
Provide a way to customize context propagation using its own storage

### LINKED ISSUE #2514: Consider adding hooks back to `RequestContext`
This is a follow-up discussion of https://github.com/line/armeria/pull/2375

It seems pretty common for users to set/clear a thread-local variable based on the life cycle of `RequestContext, e.g.

```kotlin
serviceContext.onEnter { _ ->
    MDC.put(correlationIdKey, context.correlationId)
    MDC.put(requestIdKey, context.requestId)
}
serviceContext.onExit { _ ->
    MDC.remove(correlationIdKey)
    MDC.remove(requestIdKey)
}
```

This is doable via `RequestContextExportingAppender`, but in some cases like the above, callbacks can be much simpler and more efficient. (See the thread from @andrewoma at: https://line-armeria.slack.com/archives/C1NGPBUH2/p1582090225136400)

minwoox:
Motivation:
Related https://github.com/line/armeria/pull/2375#issuecomment-571028023
It will be good if we provide a way to customize the context storage so that a user can do the scope management by him/herself.

Modifications:
- Add `RequestContextStorage`.
- Add `RequestContextStorageProvier` so that a user customizes the `ContextStorage` using Java SPI

Result:
- You can, now, create your own storage to store `RequestContext`.
- Close #2514


minwoox:
I need to add Javadoc, but I wanted to get some feedback especially for naming before I finish. PTAL

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -560,6 +561,8 @@ private ClientFactoryOptions buildOptions() {
      * Returns a newly-created {@link ClientFactory} based on the properties of this builder.
      */
     public ClientFactory build() {
+        // To initialize the context storage when the factory is built not the first request is sent.
```
anuraaga: Asserts are often disabled, it's better to make it more explicit I think - can add an `init` method and call it in a static initializer instead.

core/src/main/java/com/linecorp/armeria/client/ClientRequestContext.java
```diff
@@ -322,24 +323,25 @@ static ClientRequestContextBuilder builder(RpcRequest request, URI uri) {
      */
     @Override
     default SafeCloseable push() {
-        final RequestContext oldCtx = RequestContextThreadLocal.getAndSet(this);
+        final ContextStorage contextStorage = RequestContextUtil.storage();
```
anuraaga: Not sure if it's better but `RequestContextUtil` could still expose the same static API as before without returning `ContextStorage`

core/src/main/java/com/linecorp/armeria/common/ContextStorage.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Push.
+     */
+    @Nullable
+    <T extends RequestContext> T push(RequestContext toPush);
+
+    /**
+     * Pop.
+     */
+    void pop(@Nullable RequestContext toRestore);
```
anuraaga: Think this is fine as is but for reference I notice grpc accepts a reference to the current context as well, maybe it helps catch bugs.

https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.Storage.html#detach-io.grpc.Context-io.grpc.Context-

core/src/main/java/com/linecorp/armeria/common/ContextStorageThreadLocal.java
```diff
@@ -0,0 +1,59 @@
...
+
+import io.netty.util.concurrent.FastThreadLocal;
+import io.netty.util.internal.InternalThreadLocalMap;
+
+/**
+ * A {@link ContextStorage} that uses thread-local to store {@link RequestContext}.
+ * Override this.
+ */
+@UnstableApi
+public class ContextStorageThreadLocal implements ContextStorage {
```
anuraaga: Usually type of implementation is an adjective so `ThreadLocalContextStorage`

core/src/main/java/com/linecorp/armeria/common/ContextStorageThreadLocal.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.annotation.Nullable;
+
+import com.linecorp.armeria.common.util.UnstableApi;
+
+import io.netty.util.concurrent.FastThreadLocal;
+import io.netty.util.internal.InternalThreadLocalMap;
+
+/**
+ * A {@link ContextStorage} that uses thread-local to store {@link RequestContext}.
+ * Override this.
```
anuraaga: Instead of exposing this class let's return it from `static ContextStorage default()`. A user can hook by implementing a storage that delegates to the default.

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
```diff
@@ -48,13 +55,45 @@
     private static final Set<Thread> REPORTED_THREADS =
             Collections.newSetFromMap(new MapMaker().weakKeys().makeMap());
 
+    private static final ContextStorage contextStorage;
+
+    static {
+        final List<ContextStorageProvider> providers = Streams.stream(
```
anuraaga: ImmutableList.copyOf

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
```diff
@@ -48,13 +55,45 @@
...
+                ServiceLoader.load(ContextStorageProvider.class)).collect(toImmutableList());
+
+        if (providers.isEmpty()) {
+            contextStorage = new ContextStorageThreadLocal();
+        } else {
+            final ContextStorageProvider provider = providers.get(0);
+            if (providers.size() > 1) {
```
anuraaga: I think it's helpful to have a flag to to set the desired implementation class name to help with this situation without needing to hack with the jar

anuraaga: Actually maybe it's fine to crash instead of warn too since it's so dangerous to use a wrong storage.

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
```diff
@@ -48,13 +55,45 @@
...
+                logger.warn("Found more than one {}. Only the first provider is used: {}, providers: {}",
+                            ContextStorageProvider.class.getSimpleName(), provider, providers);
+            }
+            ContextStorage temp;
+            try {
+                temp = provider.newContextStorage();
+            } catch (Throwable t) {
+                logger.warn("Failed to create context storage. provider: {}", provider, t);
```
anuraaga: Think it's better to crash than fallback. As long as there is a flag to set the storage class (so could set it to default if really desired) it seems safer than a possibly ignored warning.

core/src/main/java/com/linecorp/armeria/common/ContextStorage.java
minwoox: Ah yes, I didn't add it because we recommend using `push()` with `try-with-resources`, but there's no harm to add the current context. Will add it, thanks!

core/src/main/java/com/linecorp/armeria/common/ContextStorageThreadLocal.java
minwoox: Ah that's a good suggestion. Thanks!

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
minwoox: Isn't it okay since it calls `.collect(toImmutableList());`?

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
minwoox: > I think it's helpful to have a flag to to set the desired implementation class name to help with this situation without needing to hack with the jar

So you mean just providing a flag instead of SPI? or both?

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
minwoox: So you mean adding `RequestContextUtil.init()` and call it from the static initializer of `ClientFactory`? If `init()` doesn't do anything, isn't it disabled as well?

minwoox: Ah, you meant making `contextStorage` not final and initializing in `init()`.

anuraaga: Ah nah I meant an empty method - it would still trigger classload and is a but clearer at the call site `init()` as a verb rather than just `storage()`

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
anuraaga: Yeah but less code / complexity :)

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
anuraaga: Both - the flag would be when the SPI is ambiguous but usually there'd just be 0-1 SPI providers and no need for flag.

core/src/main/java/com/linecorp/armeria/server/ServerBuilder.java
```diff
@@ -1476,6 +1477,8 @@ public Server build() {
                 enableServerHeader, enableDateHeader, requestIdGenerator), sslContexts);
 
         serverListeners.forEach(server::addListener);
+        // To initialize the context storage at the server start time not when the first request is received.
+        assert RequestContextUtil.storage() != null;
```
trustin: assert -> explicit initialization call

minwoox:
@anuraaga and @trustin, I updated. PTAL. 🙇
Will update the commit message later.

core/src/main/java/com/linecorp/armeria/client/ClientFactoryBuilder.java
```diff
@@ -560,6 +561,7 @@ private ClientFactoryOptions buildOptions() {
      * Returns a newly-created {@link ClientFactory} based on the properties of this builder.
      */
     public ClientFactory build() {
+        RequestContextUtil.init();
```
anuraaga: I'd put this in a `static` block in this class instead of here. Mostly same affect but clearer it's a one-time initialization

core/src/main/java/com/linecorp/armeria/common/ContextStorage.java
```diff
@@ -0,0 +1,94 @@
...
+ * >                 }
+ * >                 storage.pop(current, toRestore);
+ * >             }
+ * >             ...
+ * >      }
+ * > }
+ * }</pre>
+ */
+@UnstableApi
+public interface ContextStorage {
```
trustin: Should this be `RequestContextStorage`?

core/src/main/java/com/linecorp/armeria/common/ContextStorage.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Returns the default {@link ContextStorage} which stores the {@link RequestContext} in the thread-local.
+     */
+    static ContextStorage ofDefault() {
```
trustin: Maybe `threadLocal()`, because the default storage changes based on flag/SPI?

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
```diff
@@ -48,6 +58,57 @@
...
+    public static final ContextStorage defaultContextStorage = new ThreadLocalContextStorage();
+
+    private static final ContextStorage contextStorage;
+
+    static {
+        final List<ContextStorageProvider> providers = ImmutableList.copyOf(
+                ServiceLoader.load(ContextStorageProvider.class));
+        final String contextStorageFqcn = Flags.contextStorage();
+        if (!providers.isEmpty()) {
+            if (providers.size() > 1) {
```
anuraaga: Ah I was thinking instead of loading directly from the flag, we'd use the flag to disambiguate among the different providers in this case. I guess we can do both though

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
```diff
@@ -48,6 +58,57 @@
...
+                throw new IllegalStateException("Found more than one " +
+                                                ContextStorageProvider.class.getSimpleName() + ". providers:" +
+                                                providers);
+            }
+
+            final ContextStorageProvider provider = providers.get(0);
+            if (!contextStorageFqcn.isEmpty()) {
+                throw new IllegalStateException("Found " + provider + " and " + contextStorageFqcn +
+                                                ". Which one do you want to use?");
```
trustin: Shouldn't we always prefer what's specified in the flag? How about also logging the actual storage provider type?

core/src/main/java/com/linecorp/armeria/internal/common/ThreadLocalContextStorage.java
```diff
@@ -16,67 +16,46 @@
...
 import javax.annotation.Nullable;
 
+import com.linecorp.armeria.common.ContextStorage;
 import com.linecorp.armeria.common.RequestContext;
 
 import io.netty.util.concurrent.FastThreadLocal;
 import io.netty.util.internal.InternalThreadLocalMap;
 
-public final class RequestContextThreadLocal {
+class ThreadLocalContextStorage implements ContextStorage {
```
trustin: `final`?

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -414,6 +416,18 @@ public static boolean verboseResponses() {
         return VERBOSE_RESPONSES;
     }
 
+    /**
+     * Returns the fully qualified class name of {@link ContextStorage}.
+     *
+     * <p>The default value of this flag is an empty string, which means {@link ContextStorage#ofDefault()}
+     * is used. Specify the {@code -Dcom.linecorp.armeria.contextStorage=<FQCN>} JVM option where
+     * {@code FQCN} is the impletation of {@link ContextStorage} (e.g. {@code com.mycom.CustomContextStorage})
```
trustin: implementation

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
minwoox: Ah, I guess I totally misunderstood. 🤣
So you meant when multiple providers exist (e.g `com.x.AProvider`, `io.x.BProvider`...), if the flag is `io.x.BProvider`, then we should choose the `BProvider` class. Right?

If so, I want to get rid of the current way and follow it because, I think, there would be a case who wants to use two different JARs that have their own `RequestContextStorage`.
So I think we should choose one of them instead of throwing an exception. 😉

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
minwoox: Let me remove this. 😄

core/src/main/java/com/linecorp/armeria/common/ContextStorage.java
minwoox: Will change it. Thanks!

core/src/main/java/com/linecorp/armeria/common/ContextStorage.java
minwoox: That's a good point. Let me make a change. Thanks!

core/src/main/java/com/linecorp/armeria/common/Flags.java
minwoox: Oops. 😅

core/src/main/java/com/linecorp/armeria/common/RequestContextStorage.java
```diff
@@ -0,0 +1,97 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.armeria.common;
+
+import static com.linecorp.armeria.internal.common.RequestContextUtil.threadLocalRequestContextStorage;
```
anuraaga: I think `ThreadLocalContextStorage` could be moved to this package to avoid importing / exposing

core/src/main/java/com/linecorp/armeria/internal/common/ThreadLocalRequestContextStorage.java
```diff
@@ -16,67 +16,46 @@
...
-    /**
-     * Sets the specified {@link RequestContext} in the thread-local.
-     */
-    public static void set(RequestContext ctx) {
-        requireNonNull(ctx, "ctx");
-        context.set(ctx);
+    @Override
+    public void pop(RequestContext current, @Nullable RequestContext toRestore) {
+        requireNonNull(current, "current");
+        final RequestContext contextInThreadLocal = context.get();
```
anuraaga: Think you can optimize a but getting the map for both get / set like above.

core/src/main/java/com/linecorp/armeria/common/Flags.java
```diff
@@ -414,6 +417,19 @@ public static boolean verboseResponses() {
...
+    /**
+     * Returns the fully qualified class name of {@link RequestContextStorageProvider} that is used to choose
+     * when multiple {@link RequestContextStorageProvider}s exist.
+     *
+     * <p>The default value of this flag is an empty string, which means only one
+     * {@link RequestContextStorageProvider} must be found via Java SPI. If there are more than one,
+     * specify the {@code -Dcom.linecorp.armeria.requestContextStorageProvider=<FQCN>} JVM option to
+     * choose the {@link RequestContextStorageProvider}.
+     */
+    public static String requestContextStorageProvider() {
```
trustin: I wonder if returning `null` makes more sense when no provider is specified.

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
```diff
@@ -48,6 +58,66 @@
...
+                        }
+                    }
+                }
+                if (provider == null) {
+                    throw new IllegalStateException(
+                            providerFqcn + " does not match any " +
+                            RequestContextStorageProvider.class.getSimpleName() + ". providers: " + providers);
+                }
+            } else {
+                provider = providers.get(0);
```
trustin: Could you log the type of the provider in this branch, so that what custom provider was chosen? For other branches, `Flag` will log.

core/src/main/java/com/linecorp/armeria/common/RequestContextStorage.java
minwoox: Oops. 😅

core/src/main/java/com/linecorp/armeria/internal/common/ThreadLocalRequestContextStorage.java
minwoox: Done!

core/src/main/java/com/linecorp/armeria/common/Flags.java
minwoox: Done!

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
minwoox: Added logging after `requestContextStorage` is created.

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
```diff
@@ -74,21 +138,69 @@ public static IllegalStateException newIllegalContextPushingException(
...
 
-        return () -> RequestContextThreadLocal.set(oldCtx);
+        pop(oldCtx, null);
+        return () -> requestContextStorage.push(oldCtx);
+    }
+
+    /**
+     * Pops the current {@link RequestContext} in the storage and pushes back the specified {@code toRestore}.
+     */
+    public static void pop(RequestContext current, @Nullable RequestContext toRestore) {
```
ikhoon: How about renaming this `pop` to `replace`? Because It is hart to predict the behavior when I saw `pop` method with arguments.
```java
requestContextStorage.pop(current, toRestore)
```
However when I think this method name is `replace`, then I can understand the effect of the API.
```java
requestContextStorage.replace(current, toRestore)
```

minwoox: As we discussed, let's keep it as it is.

core/src/main/java/com/linecorp/armeria/common/ThreadLocalRequestContextStorage.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.annotation.Nullable;
+
+import io.netty.util.concurrent.FastThreadLocal;
+import io.netty.util.internal.InternalThreadLocalMap;
+
+final class ThreadLocalRequestContextStorage implements RequestContextStorage {
+
+    private static final FastThreadLocal<RequestContext> context = new FastThreadLocal<>();
+
+    static final ThreadLocalRequestContextStorage INSTANCE = new ThreadLocalRequestContextStorage();
```
anuraaga: Think you could use enum Singleton pattern if you want.

minwoox: Ah, will do. 😄

core/src/main/java/com/linecorp/armeria/internal/common/RequestContextUtil.java
```diff
@@ -48,6 +57,61 @@
...
+            } else {
+                provider = providers.get(0);
+            }
+
+            try {
+                requestContextStorage = provider.newStorage();
+            } catch (Throwable t) {
+                throw new IllegalStateException("Failed to create context storage. provider: " + provider, t);
+            }
+            logger.info("{} is used to create the request context storage: {}",
```
trustin: I actually meant not logging when:

- using the default provider (because it may be noisy.)
- specified via the flag (because it's logged there already.)

i.e. we can log in the `else {}` block above.

minwoox: I thought it might be useful to show the information of the created request context storage after it's fully created. (i.e. no exception when calling `newStorage()`).

minwoox: But I guess showing the created storage context is too verbose. Fixed. Thank you!

minwoox:
Thanks for reviewing!

PR #2612, opened 2020-03-19
Allow monitoring `Executor` and Reactor `Scheduler` together with Prometheus

### LINKED ISSUE #2605: Failed executor meter registration with using reactor's scheduler metric
**Situation**

When I use reactor and enable scheduler metric (https://projectreactor.io/docs/core/release/api/reactor/core/scheduler/Schedulers.html#enableMetrics--), I got exception like below 

```
Caused by: java.lang.IllegalArgumentException: Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'executor_seconds' containing tag keys [name, reactor_scheduler_id]. The meter you are attempting to register has keys [name].
```

When I check in reactor, it registered executor metric with additional tag key `reactor.scheduler.id` (https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/scheduler/SchedulerMetricDecorator.java#L70). meanwhile, Armeria registered executor metric without additional metric (https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/ServerConfig.java#L162)

So there is conflict. (I don't know why Prometheus has the constrain same meter should have same tags..)

**Suggestion**

In my case, although I don't use ArmeriaBlockingTaskExecutor, but the metric is registered. I was wondering if there would be flags whether using ArmeriaBlockingTaskExecutor

**Full stack trace** 

```
Caused by: java.lang.IllegalArgumentException: Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'executor_seconds' containing tag keys [name, reactor_scheduler_id]. The meter you are attempting to register has keys [name].
 at io.micrometer.prometheus.PrometheusMeterRegistry.lambda$collectorByName$9(PrometheusMeterRegistry.java:382)
 at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1932)
 at io.micrometer.prometheus.PrometheusMeterRegistry.collectorByName(PrometheusMeterRegistry.java:369)
 at io.micrometer.prometheus.PrometheusMeterRegistry.newTimer(PrometheusMeterRegistry.java:175)
 at io.micrometer.core.instrument.MeterRegistry.lambda$timer$2(MeterRegistry.java:271)
 at io.micrometer.core.instrument.MeterRegistry.getOrCreateMeter(MeterRegistry.java:576)
 at io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:529)
 at io.micrometer.core.instrument.MeterRegistry.timer(MeterRegistry.java:269)
 at io.micrometer.core.instrument.Timer$Builder.register(Timer.java:477)
 at io.micrometer.core.instrument.composite.CompositeTimer.registerNewMeter(CompositeTimer.java:140)
 at io.micrometer.core.instrument.composite.CompositeTimer.registerNewMeter(CompositeTimer.java:31)
 at io.micrometer.core.instrument.composite.AbstractCompositeMeter.add(AbstractCompositeMeter.java:66)
 at java.base/java.lang.Iterable.forEach(Iterable.java:75)
 at java.base/java.util.Collections$SetFromMap.forEach(Collections.java:5581)
 at io.micrometer.core.instrument.composite.CompositeMeterRegistry.lambda$new$0(CompositeMeterRegistry.java:65)
 at io.micrometer.core.instrument.composite.CompositeMeterRegistry.lock(CompositeMeterRegistry.java:184)
 at io.micrometer.core.instrument.composite.CompositeMeterRegistry.lambda$new$1(CompositeMeterRegistry.java:65)
 at io.micrometer.core.instrument.MeterRegistry.getOrCreateMeter(MeterRegistry.java:585)
 at io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:529)
 at io.micrometer.core.instrument.MeterRegistry.timer(MeterRegistry.java:269)
 at io.micrometer.core.instrument.Timer$Builder.register(Timer.java:477)
 at io.micrometer.core.instrument.MeterRegistry.timer(MeterRegistry.java:396)
 at io.micrometer.core.instrument.internal.TimedScheduledExecutorService.<init>(TimedScheduledExecutorService.java:44)
 at io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics.monitor(ExecutorServiceMetrics.java:126)
 at io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics.monitor(ExecutorServiceMetrics.java:140)
 at com.linecorp.armeria.server.ServerConfig.<init>(ServerConfig.java:159)
 at com.linecorp.armeria.server.ServerBuilder.build(ServerBuilder.java:1623)
 at com.linecorp.armeria.spring.web.reactive.ArmeriaReactiveWebServerFactory.getWebServer(ArmeriaReactiveWebServerFactory.java:205)
 at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext$ServerManager.<init>(ReactiveWebServerApplicationContext.java:230)
 at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext$ServerManager.get(ReactiveWebServerApplicationContext.java:248)
 at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext.createWebServer(ReactiveWebServerApplicationContext.java:91)
 at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext.onRefresh(ReactiveWebServerApplicationContext.java:78)
 ... 8 common frames omitted
```

ikhoon:
Motivation:

When a user wants to register Reactor Scheduler to Micrometer Prometheus Registry,
the tags registered by Micrometer `ExecutorServiceMetrics` and Reactor `SchedulerMetricDecorator` are different. 
Micrometer Prometheus Registry throws an `IllegalArgumentException` if the same meter has different tags.

See #2605 for detail information.

Modifications:

- Fork `ExecutorServiceMetrics` and `Timed*` from Micrometer 1.3.6
- Make `ExecutorServiceMetrics` accept a custom metric prefix
- Make the instance of `blockTaskExecutor` checked by all of `Timed*Executor*`
- Breaking
  - Remove `aremria` from executor name, `armeriaBlockingTaskExecutor` -> `blockingTaskExecutor`

Result:

- No more conflict of meter names.
- You can now register and monitor Reactor `Scheduler` with Prometheus in Armeria Server.
- Fixes #2605


core/src/main/java/com/linecorp/armeria/common/metric/ExecutorServiceMetrics.java
```diff
@@ -0,0 +1,280 @@
...
+                                                   String executorServiceName, Tag... tags) {
+        return monitor(registry, executor, executorServiceName, asList(tags));
+    }
+
+    private void monitor(MeterRegistry registry, @Nullable ThreadPoolExecutor tp) {
+        if (tp == null) {
+            return;
+        }
+
+        FunctionCounter.builder("armeria.executor.completed", tp, ThreadPoolExecutor::getCompletedTaskCount)
```
trustin: If we are putting this class to `common`, it has to be generic, i.e. The caller should be able to specify the prefix, rather than hard-coded with `armeria.executor`.

core/src/main/java/com/linecorp/armeria/common/metric/ExecutorServiceMetrics.java
```diff
@@ -0,0 +1,280 @@
...
+ * Monitors the status of executor service pools. Does not record timings on operations executed in the
+ * {@link ExecutorService}, as this requires the instance to be wrapped.
+ * Timings are provided separately by wrapping the executor service with {@link TimedExecutorService}.
+ *
+ * @author Jon Schneider
+ * @author Clint Checketts
+ * @author Johnny Lim
+ */
+@NonNullApi
+@NonNullFields
```
trustin: Please use our annotation

core/src/main/java/com/linecorp/armeria/common/metric/TimedExecutor.java
```diff
@@ -0,0 +1,71 @@
...
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+
+/**
+ * An {@link Executor} that is timed.
+ */
+@UnstableApi
+public class TimedExecutor implements Executor {
```
trustin: How about hiding everything but `ExecutorServiceMetrics` to minimize the amount of forked public API?

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -158,7 +158,9 @@
                                    gracefulShutdownQuietPeriod, "gracefulShutdownQuietPeriod");
 
         requireNonNull(blockingTaskExecutor, "blockingTaskExecutor");
-        if (!(blockingTaskExecutor instanceof TimedExecutorService)) {
+        if (!(blockingTaskExecutor instanceof TimedScheduledExecutorService ||
+              blockingTaskExecutor instanceof
+                      io.micrometer.core.instrument.internal.TimedScheduledExecutorService)) {
```
trustin: Could you explain why we are not using `TimedExecutorService` here?

ikhoon: After rethinking, it is my mistake.  
I think The `blockingTaskExecutor` instance should be checked by all of `TimedScheduledExecutorService`, `TimedExecutorService` and `TimedExecutorService`.
Because there is no inheritance between `Timed*Executor*`s.
Please let me know if I missed something. 😀

core/src/main/java/com/linecorp/armeria/common/metric/TimedExecutor.java
ikhoon: As you mentioned https://github.com/line/armeria/pull/2612#discussion_r395122626, if `ExecutorServiceMetrics` is moved to `common`, we need to move to `TimedExecutor` to `common` to remove `public` modifier. 
If it's fine, I can make it package private. 😀

ikhoon:
> Could you also fork the test cases?

Sure! 👍

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
trustin: SGTM

core/src/main/java/com/linecorp/armeria/common/metric/ExecutorServiceMetrics.java
ikhoon: I thought we temporarily use it. But specifying the prefix sounds good. 👍

core/src/main/java/com/linecorp/armeria/common/metric/ExecutorServiceMetrics.java
ikhoon: Removed.

anuraaga:
Wondering, instead of forking micrometer so extensively, does it make sense to instead have a workaround which detects if reactor is on the class path and adds the reactor tag in our metrics if it is? Or just not put this class in common even with the unstable API since it's clearly temporary?

Upstream seems to have a clear bug where tag customization is allowed without name customization, and I think we could get a fix / release fairly quickly.

trustin:
> does it make sense to instead have a workaround which detects if reactor is on the class path and adds the reactor tag in our metrics if it is?

I'm not sure it will work for other users who monitors `Executor`s with different tag sets.

> Or just not put this class in common even with the unstable API since it's clearly temporary?

That works for me, too.

ikhoon:
> > Or just not put this class in common even with the unstable API since it's clearly temporary?

> That works for me, too.

OK, will move these class to `internal.common.metric`.

ikhoon:
And I will try to make a PR to upstream. 😀

core/src/main/java/com/linecorp/armeria/internal/common/metric/ExecutorServiceMetrics.java
```diff
@@ -0,0 +1,322 @@
...
+
+    @Nullable
+    private final ExecutorService executorService;
+    private final String metricPrefix;
+    private final Iterable<Tag> tags;
+
+    private ExecutorServiceMetrics(@Nullable ExecutorService executorService, String executorServiceName,
+                                   String metricPrefix, Iterable<Tag> tags) {
+        this.executorService = executorService;
+        this.metricPrefix = metricPrefix;
```
minwoox: Just wondering if we could check this is an empty string or not, so metric doesn't look `.completed`

ikhoon:
> Thanks for moving to internal! FYI filed issue here [micrometer-metrics/micrometer#1919](https://github.com/micrometer-metrics/micrometer/issues/1919)

@anuraaga I opened PR 😀https://github.com/micrometer-metrics/micrometer/pull/1920

core/src/main/java/com/linecorp/armeria/internal/common/metric/ExecutorServiceMetrics.java
ikhoon: That sounds good. 👍

ikhoon: Added for `executorServiceName` and `metricPrefix`

core/src/main/java/com/linecorp/armeria/server/ServerConfig.java
```diff
@@ -158,9 +157,10 @@
                                    gracefulShutdownQuietPeriod, "gracefulShutdownQuietPeriod");
 
         requireNonNull(blockingTaskExecutor, "blockingTaskExecutor");
-        if (!(blockingTaskExecutor instanceof TimedExecutorService)) {
-            blockingTaskExecutor = ExecutorServiceMetrics.monitor(meterRegistry, blockingTaskExecutor,
-                                                                  "armeriaBlockingTaskExecutor");
+        if (!ExecutorServiceMetrics.isMonitoredExecutor(blockingTaskExecutor)) {
+            blockingTaskExecutor =
+                    ExecutorServiceMetrics.monitor(meterRegistry, blockingTaskExecutor,
+                                                   "armeriaBlockingTaskExecutor", "armeria.executor");
```
trustin: Now that we have our own meter name, could we remove `armeria` from the executor name?

ikhoon: Yes! 😎

ikhoon:
The PR description is updated 😀

ikhoon:
Thanks for reviewing! 😀

PR #2614, opened 2020-03-20
Fix intermittent test failure in SubscriberThrowingExceptionTest

minwoox:
The test failure was because we cannot specify the executor when `PublisherBasedStreamMessage` subscribes the publisher.


trustin:
Thanks, @minwoox !

